PR #40, opened 2020-02-01
Fix thumbnails for W-2 uploads that aren't "variable"

tdooner:
As I learned [in this article][1]...

A "variable" upload is one that ActiveStorage can create a variant of
(images) whereas PDFs are not transformable and are instead considered
"previewable" (i.e. the preview is a thumbnail of the first page).

Combined, "variants" and "previews" are considered "representations" of
the object, which is what we want for the thumbnail. We can use the
`#representable?` method to determine if a representation exists, so the
page will render either way.

As a fallback for when the representation doesn't exist, I found an icon
of a paper clip.

[1]: https://prograils.com/posts/rails-5-2-active-storage-new-approach-to-file-uploads

### DIFF
--- a/app/views/questions/w2s/edit.html.erb
+++ b/app/views/questions/w2s/edit.html.erb
@@ -20,7 +20,13 @@
               <% @documents.each do |document| %>
 
                 <li class="doc-preview">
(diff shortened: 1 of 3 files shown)


tdooner:
I tested this locally and here's what it looks like:

image

PR #42, opened 2020-02-01
Sign out of ID.me

bengolder:
When someone signs out of our application, they should also be signed out of ID.me. Otherwise, their browser session will still be able to sign back into our app without entering credentials.

This overrides the default SessionsController::destroy action. When a
user signs out, this will sign them out of our app and then out
of ID.me

Anytime we want to redirect the user to ID.me and then have the user
return to our app, we need to use the same OmniAuth strategy callback
path. This means that requests redirected back to us from ID.me will
need to either go through the "happy path" of `#idme` or the "error
path" of `#failure` on the `OmniauthCallbacksController`.

In either of those actions, we may need to reroute the request based on
context provided, such as params that we asked ID.me to include the
callback request.


spec/rails_helper.rb
```diff
@@ -1,23 +1,18 @@
...
 require "capybara/rspec"
 require "selenium/webdriver"
 require "webdrivers"
 Dir[Rails.root.join("spec/support/**/*.rb")].each { |f| require f }
 Dir[Rails.root.join("lib/strategies/**/*.rb")].each { |f| require f }
 
-
-Capybara.javascript_driver = :selenium_chrome_headless
-Capybara.server = :puma, { Silent: true }
-Capybara.default_max_wait_time = 0.2
```
tdooner: I suspect we will need something like these again when we want to try `js: true` again... but until then this makes sense to remove. ✅

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -63,15 +63,45 @@
...
+        let(:params) do
+          { logout: "success" }
+        end
+
+        before do
+          request.env["omniauth.error.type"] = :invalid_credentials
+        end
+
+        it "redirects to root path" do
+          get :failure, params: params
```
bengolder: add the expectation!

PR #43, opened 2020-02-04
Install poppler-utils for PDF previewing

tdooner:
This is required for Rails to be able to render a preview thumbnail of
an uploaded PDF document.

Also added: ignore "tmp" directory when building docker images. This
will speed up the build process locally where the "tmp" directory can
get pointlessly large.


tdooner:
Is this how to install system packages on our aptible machines?

bengolder:
@tdooner here is a ticket I separated out for this: https://www.pivotaltracker.com/story/show/171085353

PR #45, opened 2020-02-05
Add "additional documents" page

tdooner:
This page is a copy-paste of the W2sController, and it asks for any
additional documents the user might have.

https://www.pivotaltracker.com/story/show/170906879


app/models/document.rb
```diff
@@ -22,6 +22,8 @@ class Document < ApplicationRecord
   validates :document_type, inclusion: { in: DOCUMENT_TYPES }
   validates :intake, presence: true
 
+  scope :of_type, ->(type) { where(document_type: type) }
```
bengolder: nice!

PR #47, opened 2020-02-07
Add Mailing Address & Notification Opt-In pages

tdooner:
🔗 Pivotal: https://www.pivotaltracker.com/story/show/170289811

First commit adds the Mailing Address page:
* Add columns on `intake` for the mailing address
* Pre-fill values for the form based off ID.me authentication
* Use default form styling for now

Second commit adds the Notification Opt-In page:
* Add sms/email_notification_opt_in fields to `User` model
* Add associated controller / form / view to render the page
* Create a `form-group__stacked-checkboxes` to visually style the
  checkboxes


app/controllers/questions/mailing_address_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    private
+
+    def section_title
+      "Personal Information"
+    end
+
+    def illustration_path; end
+
+    def set_default_mailing_address
```
jenny-heath: another option may be to set these fields on the intake when we create it (since we'll already have them)

app/controllers/questions/notification_preference_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Questions
+  class NotificationPreferenceController < QuestionsController
+    private
```
jenny-heath: these methods aren't private on our other questions controllers - let's try to be consistent by either making them all private or not

app/forms/mailing_address_form.rb
```diff
@@ -0,0 +1,7 @@
+class MailingAddressForm < QuestionsForm
```
jenny-heath: re: above comment about when to pull address from user to intake - we could also do that in this form in the `existing_attributes` class method

app/views/questions/mailing_address/edit.html.erb
```diff
@@ -0,0 +1,18 @@
+<% content_for :page_title, "Mailing Address" %>
+
+<% content_for :form_card do %>
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+    <h2>Welcome <%= current_user.first_name %>! We'll send your final tax return to this address:</h2>
```
jenny-heath: this should be h1 - you can still style it as an h2 (using the class `h2`)

app/views/questions/mailing_address/edit.html.erb
```diff
@@ -0,0 +1,18 @@
+<% content_for :page_title, "Mailing Address" %>
```
jenny-heath: this should mirror the main question / h1 on the form

app/views/questions/notification_preference/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :page_title, "How may we contact you about your taxes?" %>
+
+<% content_for :form_card do %>
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card" } do |f| %>
+    <h2>How may we contact you about your taxes?</h2>
```
jenny-heath: h1

app/forms/mailing_address_form.rb
```diff
@@ -0,0 +1,7 @@
...
+  set_attributes_for :intake, :street_address, :city, :state, :zip_code
+
+  def save
+    @intake.update(attributes_for(:intake))
+  end
```
jenny-heath: add validations for zip, presence of fields, etc.

app/controllers/questions/questions_controller.rb
```diff
@@ -78,4 +78,4 @@ def form_class
       end
     end
   end
-end
```
jenny-heath: add analytics for successful posts and validation errors

jenny-heath: also add analytics for notification opt-ins and how often people change their address

spec/controllers/questions/mailing_address_controller_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      it "uses the mailing address from the intake" do
+        get :edit
+        expect(response.body).not_to include(user.street_address)
+        expect(response.body).to include(intake.street_address)
+      end
+    end
+  end
+
+  describe "#update" do
+    context "with valid params" do
```
jenny-heath: add a test for invalid params

spec/controllers/questions/notification_preference_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      it "updates the user's notification preferences" do
+        expect(user.sms_notification_opt_in).to eq("unfilled")
+        expect(user.email_notification_opt_in).to eq("unfilled")
+
+        post :update, params: params
+
+        user.reload
+        expect(user.sms_notification_opt_in).to eq("yes")
```
jenny-heath: test the no case - just to make sure we aren't leaving things unfilled (since it is a checkbox)

could just change one of these two to "no"

app/controllers/questions/mailing_address_controller.rb
tdooner: I like that idea, since we'll end up with `Intake` records that always have these fields set, which is a simpler assumption.

app/controllers/questions/notification_preference_controller.rb
tdooner: Ah yeah, consistency is good. I think it's a general Rails idiom to make non-action methods in controllers `private` (e.g. anything except new/index/edit/etc.), so I'd advocate for the `private` version in general. But since that's a refactor/computer gardening, it seems good to just make this non-private.

bengolder: That's a good point Tom, I'll put a chore in the backlog for converting all non-action methods on controllers to private methods.

app/controllers/questions/mailing_address_controller.rb
bengolder: ✅

app/controllers/questions/notification_preference_controller.rb
bengolder: ✅

PR #49, opened 2020-02-09
SPIKE: Send outgoing SMS using Zendesk

bengolder:
I was able to send an outgoing text message with this code!

Zendesk will not text end users by default, even when you publicly comment on a ticket.
https://support.zendesk.com/hc/en-us/articles/235638348#ariaid-title4

You need to have a trigger in place that will send text notifications to the user under some condition, such as having a specific tag. I used `outgoing_sms`.

Once the trigger is in place, if you create a ticket with the tag and a public comment, it will send the comment as an outgoing text message.

There are some remaining questions:
- should we create new tickets and close them?
- should the trigger send texts when the ticket is updated with new public replies?
- how should we handle email?
- is there a danger in having a tag for outgoing sms if the user doesn't have a phone number? Do we need to restrict the tags to the type of contact info available?
- should we look for existing tickets before creating new ones? Do we want single continuous text threads?


app/services/zendesk_spouse_verification_service.rb
```diff
@@ -0,0 +1,39 @@
+class ZendeskSpouseVerificationService
+  include ZendeskServiceHelper
+  include ZendeskPartnerHelper
+
+  def initialize(verification_request)
+    @verification_request = verification_request
+    @intake = verification_request.intake
+  end
+
+  def create_verification_request_ticket_requester
```
jenny-heath: my understanding is that this requester is the spouse (has the spouse's contact info) - i wonder whether the naming of this method will be confusing in the future because intuitively the person who requested the verification is the primary spouse

PR #50, opened 2020-02-11
Fix grey sidebar when uploading files

tdooner:
This is a weird Chrome-only bug: when clicking the "Select a file"
button, the page expands due to the `width: 100%` rule in the
`.document-upload .form__documentuploader` selector. I can't explain
why, it only happens in Chrome, and toggling the `width: 100%` rule off
and back on again fixes the page.

The problem is avoided by ensuring a new stacking context is created in
the `.file-upload` container.


tdooner:
(here's the bug:)

Kapture 2020-02-10 at 16 58 33

PR #51, opened 2020-02-11
Request consent after identify verification

bengolder:
- adds a consent page after the ID.me successful sign in.
- stores IP address and datetime of consent on the user model
- only exists for the primary user


app/assets/stylesheets/templates/_consent-form.scss
```diff
@@ -0,0 +1,14 @@
+.consent-form {
+  .checkbox {
+    @extend .button--wide;
```
bengolder: copy in the rules for `button--wide`

PR #54, opened 2020-02-12
Add Documents Overview Page

tdooner:
Ready for review now!

image

A couple outstanding design questions remain:
1. What is the destination for the "I have another document to upload". Right now, it goes all the way back to W-2's upload -- is that right?
2. Do we want to adjust the design given the appearance of the existing classes in the app? (Edit button is bigger than expected, button--wide is not wide enough)
3. Copy is inconsistent: Do users "upload", "share", or "file" documents.

Some non-design questions also:
1. The "return to previous page" behavior proved to be more complicated than I thought due to the fact that we'd have to pass through the parameter when the user deletes an existing file or uploads a new file. This is possible, but it would mean multiple params being passed around in some cases. I decided to skip this for now. Are we ok with that?
2. Are any analytics necessary for this feature?

### DIFF
--- a/spec/controllers/questions/documents_overview_controller_spec.rb
+++ b/spec/controllers/questions/documents_overview_controller_spec.rb
@@ -33,10 +33,11 @@ RSpec.describe Questions::DocumentsOverviewController do
         ]
       end
 
-      it "includes the sections" do
+      it "includes the sections with links to the document types" do
         get :edit
 
         expect(response.body).to include("W-2")
+        expect(response.body).to include(w2s_questions_path)
         expect(response.body).to include(documents[0].upload.filename.to_s)
       end
     end
--- a/spec/controllers/questions/documents_overview_controller_spec.rb
+++ b/spec/controllers/questions/documents_overview_controller_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe Questions::DocumentsOverviewController do
+  render_views
+
+  let(:intake) { create :intake }
+
+  before do
+    allow(subject).to receive(:user_signed_in?).and_return(true)
+    allow(subject).to receive(:current_intake).and_return intake
+  end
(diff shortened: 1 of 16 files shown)


bengolder:
Thank you Tom! Here's some thoughts for the questions you raised:

For each of the design questions, let's make tickets in tracker to get them answered. You can pull them down into the design debt section of the icebox.

For the non-design questions, I trust your judgement on the complexity of the params, but I am curious to learn more so I can judge whether it should be on our roadmap or not, and how to assess the workload vs. priority if people really want it.

I don't believe we need any dedicated analytics for this feature. Any major action from the user would already be captured by our default "page_view" event on GET requests.

tdooner:
Okay, this is ready for review.

* I created a new Button variant called `button--inline-action` for the purposes of these "Edit" button type of things
* I created a couple design tickets for the copy (https://www.pivotaltracker.com/story/show/171248635) and button styling (https://www.pivotaltracker.com/story/show/171248620)
* I did another quick spike on the returning URL, by attempting to implement it by passing arbitrary URL params through the document upload and back. That would be so we could pass a param like `?return_to_overview=true` when a user clicks the "edit" button. It's hard to pass all arbitrary params through though, since we'd have to embed them in the upload form. Perhaps an easier implementation is to just forward that single param, `return_to_overview`, if it exists. How does that sound?

spec/controllers/questions/documents_overview_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      intake.documents = documents
+      intake.save
+    end
+
+    context "with no document uploads" do
+      let(:documents) { [] }
+
+      it "displays an empty message" do
+        get :edit
+        expect(response.body).to include("No documents were uploaded.")
```
tdooner: update with zero state

PR #57, opened 2020-02-13
Add encryption to SSN in User model

tdooner:
We very much want to store the SSN encrypted for our users' security.

This commit adds the `attr_encrypted` gem we use on GCF, as well as a
decryption key to our credentials.

https://www.pivotaltracker.com/story/show/171220558


db/migrate/20200213213234_add_encrypted_ssn_fields_to_user.rb
```diff
@@ -0,0 +1,27 @@
...
+    remove_column :users, :ssn
+  end
+
+  def down
+    add_column :users, :ssn, :string
+
+    # Copy all encrypted SSNs to the unencrypted `ssn` column.
+    User.find_each do |u|
+      previous_ssn = u.ssn
+      u.update_column(:ssn, previous_ssn)
```
bengolder: let's remove the decryption here

PR #58, opened 2020-02-14
Add a bunch of document types

jenny-heath:
- include dependent names on ssn/itin form page and student account statement page
- slightly modify generator


app/models/document.rb
```diff
@@ -18,6 +18,33 @@ class Document < ApplicationRecord
...
+    "1099-K" => Questions::Form1099ksController,
+    "1099-MISC" => Questions::Form1099miscsController,
+    "1099-R" => Questions::Form1099rsController,
+    "1099-S" => Questions::Form1099ssController,
+    "1099-SA" => Questions::Form1099sasController,
+    "1099-SSDI" => Questions::Form1099ssdisController,
+    "1099G" => Questions::Form1099gsController,
+    "5498-SA" => Questions::Form5498sasController,
+    "IraStatement" => Questions::FormIrastatementsController,
+    "property_tax_statement" => Questions::FormPropertyTaxStatementsController,
```
bengolder: the capitalization is inconsistent here. I'd be inclined to change these keys to be "IRA Statement" & "Property Tax Statement" so we have good string to use for labeling these types of documents. I'm concerned that the inconsistencies could become an issue down the road when we query documents by type.

app/forms/form1095as_form.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
bengolder: I'm tempted to refactor all these forms into one form, because it looks like the only thing that differs is a single string.

PR #59, opened 2020-02-16
Only show relevant document upload pages to users

bengolder:
### [`d8a835e`](https://github.com/codeforamerica/vita-min/pull/59/commits/d8a835ede4570e01cadd38bc4641464dc99d872a) Hide non-relevant doc type pages

- Adds `.show?` methods to most of the document controllers, along with tests for each.
 - Updates feature specs to pass
- cleans out some unnecessary comments in the code

### [`9532d4c`](https://github.com/codeforamerica/vita-min/pull/59/commits/9532d4c9a52c0d333a909c35d312c618a3daab2c) Move QuestionNavigation to lib folder
- Just moves QuestionNavigation out of `app/forms/` and into `app/lib/`

### [`f08b191`](https://github.com/codeforamerica/vita-min/pull/59/commits/f08b191141b68feef37514d15a396b904dc603ed) List recommended docs on doc overview
- Overview page now only displays relevant document types.
- Adds a new `DocumentNavigation` object, which contains the type to controller hash, as well as methods to get an array of doc types or doc upload controllers for a specific intake.
- Adds new section to routes for defining document routes separate from questions. Doc pages now live under the `/documents/` URL path instead of `/questions/`. For example `/documents/w2s` instead of `/questions/w2s`
- puts document controllers and forms under a `Documents` module and folder (and removes them from `questions/` folder). 
- Document controllers can look up their document type using the navigation object, so the document type look up is now stored on the navigation object, rather than on both model and controller.
- `AdditionalInfoController` now uses `next_path` to go to the first relevant doc upload page.

### [`58e15c6`](https://github.com/codeforamerica/vita-min/pull/59/commits/58e15c6ad28ef405059c9222522dc30083de0559) Refactor document controllers to use the same form
- Refactors the base document upload controller to use one form, a document upload form. Deletes all the other forms for document uploads.

**Note:** I feel like we could use more refactoring of the base `QuestionsController`, the `DocumentUploadQuestionController`, `QuestionForm` and both of the `Navigation` classes. The "seams" between them feel awkward and confusing to override. They are too tightly coupled.


lib/generators/templates/document_upload_controller.template
```diff
@@ -1,12 +1,4 @@
-# frozen_string_literal: true
-
 module Questions
```
bengolder: change the module!

spec/controllers/documents/overview_controller_spec.rb
```diff
@@ -37,7 +38,7 @@
         get :edit
 
         expect(response.body).to include("W-2")
-        expect(response.body).to include(w2s_questions_path)
+        expect(response.body).to include(w2s_documents_path)
         expect(response.body).to include(documents[0].upload.filename.to_s)
       end
     end
```
bengolder: Add a spec for displaying certain doc types based on answers

PR #60, opened 2020-02-19
Optional demographic questions

jenny-heath:
9a090fe adds the db columns
9529652 changes the question layout (but even so a lot of the pages (outside of this PR as well) still have awkward spacing IMO)
2d77496 adds all of the feature specs, controllers, forms, and views - haven't added unit tests and would like an opinion on how necessary we think this is right now
0dec5ea fixes the demographic race question form so that it only saves either the selected checkboxes or prefer not to answer, depending on whether the user clicks 'continue' or 'skip question'
29bf299 adds the spouse questions because i forgot to before
27c8f75 adds model validation and make the html better (removing unnecessary attributes and improving structure)
84f1f3e fills out the pdf
0547217 fixes the database columns that were spelled "africa_american"


app/views/questions/demographic_disability/edit.html.erb
```diff
@@ -0,0 +1,23 @@
...
+            { value: :yes, label: "Yes" },
+            { value: :no, label: "No" },
+          ]
+        )
+    %>
+    <button class="button button--primary button--wide" type="submit">
+      Continue
+    </button>
+    <%= f.button :submit, name: "#{f.object_name}[demographic_disability]", value: :prefer_not_to_answer, class: "button button--wide", "aria-describedby": "demographic_disability_question" do %>
+      Skip Question
```
jenny-heath: change all of these to "Skip question"

PR #64, opened 2020-02-21
We tried to get all of the docs into one comment....it looks interesting

jenny-heath:

### DIFF
--- a/app/helpers/zendesk_service_helper.rb
+++ b/app/helpers/zendesk_service_helper.rb
@@ -99,10 +99,22 @@ module ZendeskServiceHelper
   def append_file_to_ticket(ticket_id:, filename:, file:, comment: "", fields: {})
     raise MissingTicketIdError if ticket_id.blank?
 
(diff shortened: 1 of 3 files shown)


app/helpers/zendesk_service_helper.rb
```diff
@@ -99,10 +99,22 @@ def create_ticket(subject:, requester_id:, group_id:, body:, fields: {})
...
 
+    append_multiple_files_to_ticket(
+      ticket_id: ticket_id,
+      file_list: [{ filename: filename, file: file }],
+      comment: comment,
+      fields: fields
+    )
+  end
+
+  # file_list is an Array of Hashes with keys :file and :filename.
```
bengolder: for this case, would it make sense to just pull the filename from each File object? I'm not sure if that's possible by default, but it seems like a reasonable API to expect.

bengolder:
Made a ticket for refactoring this: https://www.pivotaltracker.com/story/show/171415422

PR #67, opened 2020-02-25
Add VITA eligibility page

jenny-heath:
f32183c creates the eligibility and offboarding pages
  - adds `over_income_limit` column
  - creates new intake when eligibility form submitted, saves id to session
  - routes user appropriately depending on answers

15c24d9 associates new intake with user after sign in
  - only associates new intake if user is new
  - when user is not new or user is spouse, deletes new intake
  - removes intake id from session no matter who is signing in
  - fixes eligibility question bc it was missing a condition

555e40b moves saving of source and referrer from idme callback to eligibility controller

1f1c5cc remove redundant questions later in flow

Co-authored-by: Laura Kogler <lkogler@codeforamerica.org>


app/controllers/questions/eligibility_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def update
+      if [:had_farm_income, :had_rental_income, :income_over_limit].any? { |attr| form_params[attr] == "yes" }
+        @intake = Intake.create(form_params)
+        @intake.update(source: source, referrer: referrer)
+        session[:intake_id] = @intake.id
+        redirect_to maybe_ineligible_path
+      else
+        @intake = Intake.create(source: source, referrer: referrer)
```
tdooner: do we want to create it with `form_params` here too so the Intake has `no` values in these fields instead of `unfilled`?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -18,6 +18,14 @@ def idme
       return redirect_to spouse_identity_questions_path(missing_spouse: "true")
     end
 
+    if is_returning_consenting_user || is_returning_nonconsenting_user || is_new_spouse
+      if session[:intake_id]
+        intake_from_session = Intake.find(session[:intake_id])
+        intake_from_session.destroy! if intake_from_session.users.count == 0
```
tdooner: nice `if` statement to guard against accidentally deleting an intake!

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -18,6 +18,14 @@ def idme
       return redirect_to spouse_identity_questions_path(missing_spouse: "true")
     end
 
+    if is_returning_consenting_user || is_returning_nonconsenting_user || is_new_spouse
```
tdooner: thoughts on simplifying this to:

`if session[:intake_id] && (is_returning_user || is_new_spouse)`

?

Also- as an idea- you could create a new variable for this situation:

```ruby
# up above:
should_delete_duplicate_intake = session[:intake_id] && (is_returning_user || is_new_spouse)

# right here:
if should_delete_duplicate_intake
  # ...
end
```

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -122,6 +157,20 @@
...
+        before do
+          session[:intake_id] = intake_from_session.id
+        end
+
+        it "clears the intake id from the session and deletes the intake" do
+          get :idme, params: { spouse: "true" }
+          expect(session[:intake_id]).to be_nil
+          expect(Intake.exists?(intake_from_session.id)).to eq false
+        end
+      end
```
tdooner: nice tests!

PR #68, opened 2020-02-25
Add spouse consent page

tdooner:
This branch adds a consent page that's shown after a spouse
authenticates for the first time. This is the same as the primary user
consent page (they now share a partial), which now includes the
consenting user's full name to allow the user to differentiate.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -7,18 +7,21 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     is_new_user = !@user.persisted?
     is_consenting_user = @user.consented_to_service_yes?
     is_returning_user = !is_new_user
+    is_returning_spouse = @user.is_spouse && is_returning_user
 
     is_new_primary_user = !has_spouse_param && is_new_user
     is_new_spouse = has_spouse_param && is_new_user
-    is_returning_consenting_user = is_returning_user && is_consenting_user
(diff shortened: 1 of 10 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -7,18 +7,21 @@ def idme
...
+    is_returning_consenting_spouse = is_returning_spouse && is_consenting_user
+    is_returning_nonconsenting_spouse = is_returning_spouse && !is_consenting_user
     is_primary_but_expected_spouse = (@user == current_user && has_spouse_param)
 
     if is_primary_but_expected_spouse
       return redirect_to spouse_identity_questions_path(missing_spouse: "true")
     end
 
-    should_delete_duplicate_intake = session[:intake_id] && (is_returning_consenting_user || is_returning_nonconsenting_user || is_new_spouse)
+    should_delete_duplicate_intake = session[:intake_id] && (is_returning_user || is_new_spouse)
```
tdooner: @jenny-heath I simplified this because the line got very long. Does this logic look correct to you?

jenny-heath: i think not quite the same? it looks like `is_returning_consenting_user` and `is_returning_nonconsenting_user` both check that it is not a returning spouse. that said, we should probably delete duplicate intakes for returning spouses anyway?

tdooner: cool. ben and I looked at this and yeah, we should delete the duplicate intakes anyway. Thanks!

PR #71, opened 2020-02-26
WIP: add feelings survey

jenny-heath:


app/controllers/questions/feelings_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Questions
+  class FeelingsController < QuestionsController
+    skip_before_action :require_sign_in
+    layout "question"
+
+    def current_intake
+      @intake
+    end
```
bengolder: It might make sense to put this back on the ApplicationController, and pull it from the session, so we can use it both in the EligibilityController, and here. 
perhaps something like

```ruby
def current_intake
  @intake ||= Intake.find_by_id(session[:intake_id])
end
```

PR #72, opened 2020-02-27
Final copy changes ticket

tdooner:
This should be ready to go except for three questions about specific copy in the ticket. (Those three items could be addressed subsequently.)

https://www.pivotaltracker.com/story/show/171243681


app/forms/notification_preference_form.rb
```diff
@@ -8,4 +9,12 @@ def save
   def self.existing_attributes(intake)
     HashWithIndifferentAccess.new(intake.primary_user.attributes)
   end
+
+  private
+
+  def validate_one_method_selected
+    unless sms_notification_opt_in == "yes" || email_notification_opt_in == "yes"
+      errors.add(:base, "We need a way to ask you a question if there is a problem with your taxes.")
```
jenny-heath: @tdooner  A suggestion for better copy: "We need a way to get in touch in order to help you."

tdooner: Cool, thanks for the suggestion. I will carry the suggestion into the Tracker story and update the copy here in a follow-up commit if we decide to choose it.

tdooner:
I went down a little rabbit hole about the HTML <title> attribute, the problem is that we don't have this behavior in the "question" layout. (We do have it in the "yes_no_question" layout.)

I created a computer gardening story to ensure that all pages have appropriate <title> elements with form errors:
https://www.pivotaltracker.com/story/show/171530047

app/forms/notification_preference_form.rb
tdooner: (I decided to just change the copy. It's in https://github.com/codeforamerica/vita-min/commit/a2c7a51e8677047db326f98c2902c8e6616fa7a5. Generally, I'm trying to minimize the amount of deviation from well-specified Tracker stories, which this one seemed to be. But this change seems pretty unlikely to be surprising when acceptance testing, and positive for users so :shipit: .)

PR #80, opened 2020-03-05
Update logic for conditional "Get Started" button

tdooner:
We want the "Get Started" button to appear conditionally as described in https://www.pivotaltracker.com/story/show/171530714. Also the `utm_state` param should be stored in the session & sent to Mixpanel.


app/views/public_pages/home.html.erb
```diff
@@ -18,14 +18,14 @@
...
-        <% if Rails.env.production? %>
-          <%= link_to vita_providers_path, id: "firstCta", class: "button button--primary" do %>
-            Find a location
-          <% end %>
-        <% else %>
+        <% if enable_online_intake? %>
           <%= link_to question_path(QuestionNavigation.first), id: "firstCta", class: "button button--primary" do %>
             Get started
           <% end %>
+        <% else %>
```
jenny-heath: i think we decided that the cta button at the bottom of the page should behave in the same way as this one

spec/controllers/public_pages_controller_spec.rb
```diff
@@ -7,22 +7,37 @@
...
+
+      context "when online intake is not enabled" do
+        before do
+          allow_any_instance_of(PublicPagesHelper).to receive(:enable_online_intake?).and_return(false)
+        end
+
+        it "does not link to the first questions page" do
+          get :home
+
+          expect(response.body).not_to include "Get started"
```
jenny-heath: should we also check that a link to location search is present?

app/views/public_pages/home.html.erb
tdooner: missed that, thanks!

spec/controllers/public_pages_controller_spec.rb
tdooner: cool yeah, I'll beef these up a bit. That would have made me realize there was a second button.

app/helpers/public_pages_helper.rb
```diff
@@ -0,0 +1,7 @@
+module PublicPagesHelper
+  def enable_online_intake?
+    return true unless Rails.env.production?
+
+    ENV['ENABLE_ONLINE_INTAKE'].present? && session[:source] != 'propel'
```
jenny-heath: for `session[:source]` we should instead use the getter method, `source` that's available on the ApplicationController, just so that we have a consistent way to get the source.

jenny-heath: ^ that was ben, i missed this

jenny-heath:
Ben mentioned that it would be good to remove the part where it checks whether it's in production - that way there will be fewer differences between prod and other environments (and it should work the same way) 

so let's just check the env var and the source

app/helpers/public_pages_helper.rb
tdooner: there were some complications here, but good idea on the general approach of re-using that method. I had to call it as `controller.source` from within the helper, and also it was a bit tricky in test because the TestController class that Rspec sets up for us doesn't inherit from ApplicationController. fyi @bengolder  see https://github.com/codeforamerica/vita-min/pull/80/commits/b2f3896e3fc68ad02a9c5d8cc4ea31bbea4c7225 for the diff if you're curious.

PR #82, opened 2020-03-06
[WIP?] Allow users to return to pages later in the flow

tdooner:
**WIP because:** I'm not sure about the `callback_url` change in ID.me strategy, and removing the test because it didn't easily work. I wish I could have had more time to get the test working but I had to go.

---
For a returning user, we want them to be able to return to a later page,
e.g. the /documents/additional-documents after they log in. This
accomplishes it by adding a `next_path` param to the URL which is
carried through the "Sign in with ID.me" button and the OmniAuth
callback.

The change to the OmniAuth strategy (removing the overridden
`callback_url` method) is to return to the superclass's behavior of
including the query string when generating the callback URL. See:
https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb#L443-L445

As for testing, I added a couple light controller tests but the real
heavy lifter here is a surprisingly-existent feature spec that seemed to
be testing this behavior (but with a strange expectation). I updated the
expectation to match the intended behavior, and it now works.

https://www.pivotaltracker.com/story/show/171505178

### DIFF
--- a/spec/controllers/documents_controller_spec.rb
+++ b/spec/controllers/documents_controller_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe DocumentsController, type: :controller do
         delete :destroy, params: { id: document.id }
       end.not_to change(Document, :count)
 
-      expect(response).to redirect_to(identity_questions_path)
+      expect(response).to redirect_to(identity_questions_path(after_login: document_path(document)))
     end
 
     context "with an authenticated user" do
--- a/spec/controllers/questions/job_count_controller_spec.rb
+++ b/spec/controllers/questions/job_count_controller_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Questions::JobCountController do
       it "redirects to ID.me login page" do
         get :edit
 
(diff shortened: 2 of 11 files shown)


spec/controllers/documents_controller_spec.rb
```diff
@@ -10,7 +10,7 @@
         delete :destroy, params: { id: document.id }
       end.not_to change(Document, :count)
 
-      expect(response).to redirect_to(identity_questions_path)
+      expect(response).to redirect_to(include(identity_questions_path))
```
bengolder: add the param to the expectation.

spec/controllers/questions/job_count_controller_spec.rb
```diff
@@ -17,7 +17,7 @@
       it "redirects to ID.me login page" do
         get :edit
 
-        expect(response).to redirect_to identity_questions_path
+        expect(response).to redirect_to(include(identity_questions_path))
```
bengolder: add param to expectation

PR #83, opened 2020-03-09
Follow up docs page

jenny-heath:
Some things to note:
- the page is in the docs navigation list because there's a lot of functionality there that didn't seem worth trying to fill in the gaps for
- the show method just returns false so that it never shows up in the flow
- i added the send_requested_documents controller to the routes by adding it to the list of docs controllers that's iterated over - would like input on whether this is a good way to have done that

### DIFF
--- a/app/controllers/documents/overview_controller.rb
+++ b/app/controllers/documents/overview_controller.rb
@@ -15,7 +15,10 @@ module Documents
     end
 
     def recommended_document_types
-      DocumentNavigation.new(self).types_for_intake(current_intake)
+      document_types = DocumentNavigation.new(self).types_for_intake(current_intake)
+      include_requested_documents = @documents.where(document_type: "Requested").exists?
+      document_types += ["Requested"] if include_requested_documents
+      document_types
     end
   end
 end
--- a/spec/controllers/documents/requested_documents_controller_spec.rb
+++ b/spec/controllers/documents/requested_documents_controller_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe Documents::RequestedDocumentsController do
+  render_views
+
+  let(:intake) { create :intake }
+
+  before do
+    allow(subject).to receive(:user_signed_in?).and_return(true)
+    allow(subject).to receive(:current_intake).and_return intake
+  end
+
+  describe "#edit" do
+    context "with existing document uploads" do
+      it "assigns the documents to the form" do
+        doc = create :document, :with_upload, document_type: "Requested", intake: intake
+
+        get :edit
+
+        expect(assigns(:documents)).to include(doc)
+      end
+    end
+  end
+
+  describe "#next_path" do
+    it "returns send requested documents path" do
+      result = subject.next_path
+
+      expect(result).to eq send_requested_documents_documents_path
+    end
+  end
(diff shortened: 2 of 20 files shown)


app/controllers/documents/overview_controller.rb
```diff
@@ -15,7 +15,10 @@ def section_title
     end
 
     def recommended_document_types
-      DocumentNavigation.new(self).types_for_intake(current_intake)
+      document_types = DocumentNavigation.new(self).types_for_intake(current_intake)
+      include_requested_documents = @documents.any? { |doc| doc.document_type == "Requested" }
```
bengolder: this might a little more efficient as a query instead of an map block, for example, `@documents.where(document_type: "Requested").exists?`

config/routes.rb
```diff
@@ -33,7 +33,7 @@
 
   resources :documents, controller: :documents do
     collection do
-      DocumentNavigation.all_controllers.uniq.each do |controller_class|
+      (DocumentNavigation.all_controllers + [Documents::SendRequestedDocumentsController]).uniq.each do |controller_class|
```
bengolder: This logic (appending the requested docs controller) should be handled by the Navigation class, not the routes file. `.all_controllers` is intended to include any controllers, even if they are not in the main flow.

spec/controllers/documents/requested_documents_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      it "renders the thumbnails" do
+        doc = create :document, :with_upload, document_type: "Requested", intake: intake,
+          upload_path: document_path
+
+        expect { get :edit }.not_to raise_error
+        expect(response.body).to include('document_bundle.pdf')
+      end
+    end
+  end
```
bengolder: While totally fine to include it, just wanted to note that I don't feel like we need to test this for every doc upload type page. Only for a representative one.

spec/features/web_intake/requested_documents_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.feature "Client uploads a requested document" do
+  before do
+    allow_any_instance_of(Users::SessionsController).to receive(:idme_request).and_return(
+      user_idme_omniauth_callback_path(logout: "success")
+    )
+  end
```
bengolder: Why do we need to set up the logout for this test? Are we logging them out?

spec/services/zendesk_follow_up_docs_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        comment: <<~DOCS
+          The client added requested follow-up documents:
+          * #{requested_docs[0].upload.filename}
+          * #{requested_docs[1].upload.filename}
+        DOCS
+      )
+      requested_docs.each do |doc|
+        doc.reload
+        expect(doc.zendesk_ticket_id).to eq 34
+      end
```
bengolder: Should this be a separate "it" assertion?

spec/features/web_intake/requested_documents_spec.rb
jenny-heath: nope! my bad!

PR #85, opened 2020-03-14
Send event when "sign in with ID.me" clicked

jenny-heath:
Things I've tried:

**Overwrite the controller action**
- the link on the 'Sign in with ID.me' button is `user_idme_omniauth_authorize_path` which according to the routes maps to `users/omniauth_callbacks#passthru`
- `passthru` is defined on `Devise::OmniauthCallbacksController` which is what our `Users::OmniauthCallbacksController` inherits from
- i tried overwriting `passthru` in our callbacks controller:
```
def passthru
  puts 'BLOOP'
  super
end
```
it did not appear to work

**Overwrite the `request_phase`**
- i cannot honestly say i remember what the request phase is
- all i know is that it gets called when we do the ID.me sign in
- i overwrote it in the strategy:
```
def request_phase
  puts 'BLOOP!!'
  super
end
```
- this appears to work but now that i'm not in a controller i don't know how to access `send_mixpanel_event`

**next steps**
my next thought was to have the button link to another method which could send the even and then redirect to `idme_authorize` (as happens in `Users::OmniauthCallbacksController#failure`) but then i remembered that the sign in button's method is 'post', which seems to mean that i should not do this. however. i have a question. why can we redirect to `idme_authorize` for the spouse? another question. what is the plural of 'spouse'? 'spice'?

**UPDATE**
i have done what i outlined in the 'next steps'. there are still failing specs. i haven't figured out why they are failing.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,6 +1,11 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include IdmeAuthenticatable
 
+  def idme_sign_in(after_login: nil)
+    send_mixpanel_event(event_name: "click_sign_in")
+    redirect_to idme_authorize(after_login: after_login)
+  end
+
   def idme
     has_spouse_param = params["spouse"] == "true"
     has_after_login_param = params["after_login"].present?
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -71,6 +71,8 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def failure
     error_type = request.env["omniauth.error.type"]
     error = request.env["omniauth.error"]
+    send_mixpanel_event(event_name: "authentication_failure", data: { error: error, error_type: error_type })
+
     if error_type == :access_denied
       # the user did not grant permission to view their info
       redirect_to identity_needed_path
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,9 +1,9 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include IdmeAuthenticatable
 
-  def idme_sign_in(after_login: nil)
+  def idme_sign_in
     send_mixpanel_event(event_name: "click_sign_in")
-    redirect_to idme_authorize(after_login: after_login)
+    redirect_to idme_authorize(after_login: params[:after_login])
   end
 
   def idme
--- a/lib/strategies/idme.rb
+++ b/lib/strategies/idme.rb
@@ -11,11 +11,6 @@ module OmniAuth
         :token_url => "https://#{DOMAIN}/oauth/token"
       }
 
-      def request_phase
-        puts 'BLOOP!!'
-        super
-      end
-
       def raw_info
         access_token.options[:mode] = :query
         @raw_info ||= access_token.get("api/public/v3/attributes.json").parsed
--- a/lib/strategies/idme.rb
+++ b/lib/strategies/idme.rb
@@ -11,6 +11,11 @@ module OmniAuth
         :token_url => "https://#{DOMAIN}/oauth/token"
       }
 
+      def request_phase
+        puts 'BLOOP!!'
+        super
+      end
+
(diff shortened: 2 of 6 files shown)


lib/strategies/idme.rb
```diff
@@ -11,6 +11,11 @@ class IdMe < OmniAuth::Strategies::OAuth2
         :token_url => "https://#{DOMAIN}/oauth/token"
       }
 
+      def request_phase
```
jenny-heath: @bengolder i have failed ☹️

bengolder:
@jenny-heath happy to help with this, but if we don't connect in the next hour, could you document more about what you tried and didn't work?

bengolder:
Overriding `#passthru` on the Sessions controller did not help. The method was never called.

bengolder:
It's unclear where to intervene in order to capture the click.

bengolder:
Some ways to work through this:
1. Set a breakpoint inside library and figure out the callstack in order to better understand what is being called in what order.
    - perhaps inside the ID.me strategy
    - perhaps in something that the Callbacks controller inherits from.
2. create separate method in callbacks controller and make new route to go to it and then redirect to the right place.

tdooner:
I took a quick look and [you're not the only one that hit this](https://github.com/omniauth/omniauth/issues/925). It appears `passthru` was deprecated/removed [seven years ago](https://github.com/heartcombo/devise/issues/2386#issuecomment-17592473) by devise. Great 🤦‍♂ .

I think the best option is the next step you described (also ben's item #2) -- to make our own passthru action (maybe called something else? idk) that we link to and then it redirects to the actual place.

One other (less-good) option would have been to use the Omniauth's `before_request_phase` configuration setting which is not documented anywhere but [seems to be called right when we need it](https://github.com/omniauth/omniauth/blob/c2380ae848ce4e0e39b4bb94c5b8e3fd0a544825/lib/omniauth/strategy.rb#L209). Although this approach could work, the callback is called outside of the Rails controller context so we wouldn't necessarily be able to use the existing mixpanel tracking logic. May as well just make another controller action for it instead.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,6 +1,11 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include IdmeAuthenticatable
 
+  def idme_sign_in(after_login: nil)
```
tdooner: Won't `after_login` be in the `params` here (instead of being passed in as an argument to the controller method?)

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,6 +1,11 @@
...
+    send_mixpanel_event(event_name: "click_sign_in")
+    redirect_to idme_authorize(after_login: after_login)
```
tdooner: I think this should be `user_idme_omniauth_authorize_path(after_login: params[:after_login])` to match the previous behavior? That will allow us to stub out the behavior in feature specs (which I believe is the reason they are failing right now -- they are actually trying to request the id.me OAuth flow via the feature specs web server)

jenny-heath:
UPDATE: Tom and I paired on this a little. It's looking increasingly difficult to go about it in this way. The state of things when we started: I had created an intermediate controller action called `idme_sign_in` that was linked to from the sign in button which sent the mixpanel event and redirected to `idme_authorize`. This functioned but caused feature specs to fail because the mocking didn't work anymore.

Findings:
- redirecting to `user_idme_omniauth_authorize_path` from `idme_sign_in` didn't work. It went to a 404 page that said "Not found. Authentication passthru."
- we tried switching the link on the button back to `user_idme_omniauth_authorize_path` and overriding the `passthru` method on the devise controller. we discovered that `passthru` only gets called if it's a get request (not a post), which explains the [404](https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/app/controllers/devise/omniauth_callbacks_controller.rb#L6-L8). the problem was that if we changed the method on the button to a get, the mocking broke again and we got the same feature specs failing
- in trying to address the problem of the feature spec mocking, we found that this worked for some tests:
```
allow_any_instance_of(Users::OmniauthCallbacksController).to receive(:idme_authorize).and_return(
    user_idme_omniauth_callback_path
)
```
but is unlikely to work for all (see `auth_failure_spec`)

Conclusion:
We are now considering a front end option and will look into the GCF repo to see whether they have any click tracking that uses js. Thoughts?

PR #86, opened 2020-03-16
Direct users to ID.me sign up page instead of sign in

jenny-heath:
- also for spouse

### DIFF
--- a/app/controllers/concerns/idme_authenticatable.rb
+++ b/app/controllers/concerns/idme_authenticatable.rb
@@ -27,6 +27,7 @@ module IdmeAuthenticatable
       logout_params.merge!({
         scope: idme_scope,
         response_type: "code",
+        op: "signup",
       })
     end
 
--- a/lib/strategies/idme.rb
+++ b/lib/strategies/idme.rb
@@ -7,7 +7,7 @@ module OmniAuth
 
       option :client_options, {
         :site => "https://#{DOMAIN}",
-        :authorize_url => "https://#{DOMAIN}/oauth/authorize",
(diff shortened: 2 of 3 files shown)


spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -309,7 +309,7 @@
           expect(redirect_params["state"]).to eq "1234"
           expect(redirect_params["scope"]).to eq "ial2"
           expect(redirect_params["response_type"]).to eq "code"
-          expect(redirect_params["redirect_uri"]).to eq "http://test.host/users/auth/idme/callback?spouse=true"
```
bengolder: we don't need to include the param on our callback url, only for the "authorize" redirect to ID.me

bengolder: Did you add this expectation because ID.me automatically includes it?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -79,7 +79,7 @@ def failure
       redirect_to root_path
     elsif error_type == :invalid_credentials && params["logout"] == "primary"
       # We have signed out the primary user and need to verify the spouse
-      redirect_to idme_authorize(spouse: "true")
```
bengolder: This should be moved line 30 in `idme_authenticatable.rb`

spec/controllers/users/omniauth_callbacks_controller_spec.rb
jenny-heath: lol yes sorry

PR #90, opened 2020-03-18
Skip Spouse Authentication

bengolder:
2170035 Add button to skip spouse authentication 
94605a5 Create & store spouse auth token
4096062 Update Zendesk comments to note when spouse has not authenticated (includes some extra cruft sorry)
4525689 Create new spouse auth page and match token to intake
7d72dac Send spouse to 'you did it' page after they consent
80e3a8e Send all the stuff to zendesk if it's already been sent and needs to be updated with spouse

### DIFF
--- a/app/controllers/questions/spouse_consent_controller.rb
+++ b/app/controllers/questions/spouse_consent_controller.rb
@@ -12,8 +12,7 @@ module Questions
 
     def next_path
       if session[:authenticate_spouse_only]
-        session[:authenticate_spouse_only] = nil
-        spouse_auth_complete_path
+        verify_spouse_done_path
       else
         super
       end
--- a/app/controllers/questions/spouse_consent_controller.rb
+++ b/app/controllers/questions/spouse_consent_controller.rb
@@ -3,5 +3,14 @@ module Questions
     def self.show?(intake)
       intake.filing_joint_yes?
     end
+
+    def next_path
+      if session[:authenticate_spouse_only]
+        session[:authenticate_spouse_only] = nil
+        spouse_auth_complete_path
+      else
+        super
+      end
+    end
   end
 end
--- a/app/controllers/questions/spouse_consent_controller.rb
+++ b/app/controllers/questions/spouse_consent_controller.rb
@@ -4,6 +4,12 @@ module Questions
       intake.filing_joint_yes?
     end
 
+    def after_update_success
+      if session[:authenticate_spouse_only]
+        SendSpouseAuthDocsToZendeskJob.perform_later(current_intake.id)
+      end
+    end
+
     def next_path
       if session[:authenticate_spouse_only]
         session[:authenticate_spouse_only] = nil
--- a/app/controllers/spouse_auth_only_controller.rb
+++ b/app/controllers/spouse_auth_only_controller.rb
@@ -5,13 +5,8 @@ class SpouseAuthOnlyController < ApplicationController
   helper_method :illustration_path
 
   def show
-    token = params[:token]
-    if token
-      intake = Intake.where(spouse_auth_token: params[:token]).first
-    end
-    if !token || !intake
-      return redirect_to not_found_path
-    end
+    intake = Intake.where.not(spouse_auth_token: nil).where(spouse_auth_token: params[:token]).first
+    return redirect_to verify_spouse_not_found_path unless intake.present?
 
     session[:intake_id] = intake.id
     session[:authenticate_spouse_only] = true
--- a/app/controllers/spouse_auth_only_controller.rb
+++ b/app/controllers/spouse_auth_only_controller.rb
@@ -1,0 +1,27 @@
+class SpouseAuthOnlyController < ApplicationController
+  layout "question"
+
+  helper_method :section_title
+  helper_method :illustration_path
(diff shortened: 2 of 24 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -27,17 +28,29 @@ def idme
       return redirect_to params["after_login"]
     end
 
-    should_delete_duplicate_intake = session[:intake_id] && (is_returning_user || is_new_spouse)
+    # A new intake is created before sign in so that we can save answers to questions that are asked before sign in.
+    # If a returning user signs in, we want to delete this new intake because the user will already have an intake.
+    # TODO: We also want to do this for a new spouse - why??
```
jenny-heath: can someone remind me why we delete duplicate intake for spouse? why is there a duplicate intake when the spouse is signing in?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -27,17 +28,29 @@ def idme
...
+    should_delete_duplicate_intake = session[:intake_id] && (is_returning_user || is_new_spouse) && !used_spouse_auth_only_link
     if should_delete_duplicate_intake
       intake_from_session = Intake.find(session[:intake_id])
       intake_from_session.destroy! if intake_from_session.users.count == 0
       session[:intake_id] = nil
     end
 
     if is_new_spouse
+      if used_spouse_auth_only_link
+        # TODO: could alternatively use Intake.find_by_id(session[:intake_id]) here just to be sure
```
jenny-heath: opinions? what do we think are the odds that someone is already signed in on the device where this person is visiting the authorize spouse link? and that the person signed in is not the primary? that's the only case in which this would go wrong

app/controllers/questions/spouse_consent_controller.rb
```diff
@@ -3,5 +3,14 @@ class SpouseConsentController < ConsentController
     def self.show?(intake)
       intake.filing_joint_yes?
     end
+
+    def next_path
+      if session[:authenticate_spouse_only]
+        session[:authenticate_spouse_only] = nil
```
jenny-heath: do we need to do this? i'm not sure when session stuff gets cleared

app/jobs/send_spouse_auth_docs_to_zendesk_job.rb
```diff
@@ -0,0 +1,14 @@
+class SendSpouseAuthDocsToZendeskJob < ApplicationJob
+  queue_as :default
+
+  def perform(intake_id)
+    intake = Intake.find(intake_id)
+
+    if intake.completed_intake_sent_to_zendesk
```
jenny-heath: we don't have an equivalent for `completed_intake_sent_to_zendesk` for sending the spouse documents. i don't know that it's necessary but wanted to call that out in case someone had an opinion

app/services/zendesk_intake_service.rb
```diff
@@ -149,6 +149,26 @@ def send_final_intake_pdf
     output
   end
 
+  def send_intake_pdf_with_spouse
+    comment_body = <<~BODY
+      Updated 13614-c from online intake - added spouse signature and contact
```
jenny-heath: the completed intake comment includes a macro that sends a notification to clients. should we have one here?

app/services/zendesk_intake_service.rb
```diff
@@ -149,6 +149,26 @@ def send_final_intake_pdf
...
+
+      Client's provided interview preferences: #{@intake.interview_timing_preference}
```
jenny-heath: i included the interview preferences and additional info here because they are included in the comment for the completed intake - maybe redundant but seemed good to include since it's not on the form

app/controllers/spouse_auth_only_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  helper_method :illustration_path
+
+  def show
+    token = params[:token]
+    if token
+      intake = Intake.where(spouse_auth_token: params[:token]).first
+    end
+    if !token || !intake
+      return redirect_to not_found_path
+    end
```
bengolder: A little more concise might be:

```ruby
    intake = Intake.where(spouse_auth_token: params[:token]).first
    return redirect_to not_found_path unless intake.present?
```

jenny-heath: i think it would have to be `params[:token].present? && intake.present?` because `params[:token]` could be `nil` in which case it would find an intake where `spouse_auth_token` is `nil`?

app/controllers/questions/spouse_consent_controller.rb
jenny-heath: moved to spouse auth complete method

app/jobs/send_spouse_auth_docs_to_zendesk_job.rb
jenny-heath: seems unlikely

app/services/zendesk_intake_service.rb
jenny-heath: asked kelly

PR #91, opened 2020-03-19
Change copy on dependents page

jenny-heath:
- split VISA question into separate checkbox group
- invert north american resident question


app/views/questions/had_dependents/edit.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :form_question, "Did #{you_or_spouse(@form.intake)} have any dependents in 2019?" %>
-<% content_for :form_help_text, "This generally means a child or aging parent that you live with and financially support." %>
+<% content_for :form_question, "Would #{you_or_spouse(@form.intake)} like to claim anyone for 2019?" %>
```
tdooner: I'm not sure about using "you or your spouse" here... the dependents are dependents of them both collectively, not either one of them. Maintaining the previous behavior seems fine to me, so not a blocker but something to think about.

PR #92, opened 2020-03-19
Add ajaxMixpanelEvents to track link clicks

tdooner:
Any link with a `data-track-click` property will have its click events
intercepted by this handler, which will send an AJAX POST to a new
controller action that sends the event to Mixpanel.

Of particular note is that the click handler will allow the user to continue regardless of whether the AJAX request succeeds, fails, or hits a 200ms timeout. (I chose 200ms to be as imperceptible as possible.)

### `AjaxMixpanelEventsController`
This controller's single action, `#create` will accept a valid POST and then send a Mixpanel event in response.

### `window.MixpanelData`
On every page, we add some data about the page to the Javascript `window` object. it looks like this:
```javascript
{controller_action: "Questions::IdentityController#edit", full_path: "/questions/identity?spouse=true"}
```
### `ajaxMixpanelEvents` function

On page load, we look for any elements with the `data-track-click` property. When they are clicked, we use the value of `data-track-click` to send a POST to the `AjaxMixpanelEventsController`.

```html
<!-- sends a "click_whatever" event to Mixpanel -->
<a href="/whatever" data-track-click="whatever">Click meee!</a>
```

### DIFF
--- a/app/assets/javascripts/ajax_mixpanel_events.js
+++ b/app/assets/javascripts/ajax_mixpanel_events.js
@@ -1,0 +1,35 @@
+var ajaxMixpanelEvents = (function() {
+  var init = function() {
+    $("[data-track-click]").on("click", function (e, options = {}) {
+      if (options.trackedEvent === true) {
+        return;
+      }
+
+      e.preventDefault();
+      e.stopPropagation();
+
+      var clickedElement = $(e.target);
+      var eventData = {
+        event_name: "clicked_link",
+        controller_action: window.mixpanelData.controller_action,
+        full_path: window.mixpanelData.full_path,
+        data: {
+          link_text: clickedElement.text().trim(),
+        }
+      };
+
+      $.ajax({
+        type: "POST",
+        url: "/ajax_mixpanel_events",
+        timeout: 200,
+        data: eventData,
+      })
+      .then(() => clickedElement.trigger("click", {trackedEvent: true}))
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/ajax_mixpanel_events.js
```diff
@@ -0,0 +1,35 @@
...
+      e.stopPropagation();
+
+      var clickedElement = $(e.target);
+      var eventData = {
+        event_name: "clicked_link",
+        controller_action: window.mixpanelData.controller_action,
+        full_path: window.mixpanelData.full_path,
+        data: {
+          link_text: clickedElement.text().trim(),
+        }
```
bengolder: It doesn't look link this is adding "idme" to the event data or the event name. The event name submitted should be "click_idme_signin"

app/assets/javascripts/ajax_mixpanel_events.js
```diff
@@ -0,0 +1,35 @@
...
+      };
+
+      $.ajax({
+        type: "POST",
+        url: "/ajax_mixpanel_events",
+        timeout: 200,
```
bengolder: I would put this number higher. The risk of moving on too early is that we mistake timeouts as a lack of interest in signing in.

bengolder: This timed out for me locally

PR #94, opened 2020-03-23
Change copy on ID.me sign in page

jenny-heath:


app/views/questions/identity/edit.html.erb
```diff
@@ -6,21 +6,29 @@
         <main role="main">
           <div class="form-card">
             <h1 class="form-question">
-              Sign in
+              First, let’s get some basic information.
```
bengolder: Would be helpful to make sure this is reflected in the `<head> > <title>` too

PR #95, opened 2020-03-23
Add mixpanel_data to intake

jenny-heath:


app/controllers/application_controller.rb
```diff
@@ -102,6 +102,7 @@ def send_mixpanel_event(event_name:, data: {})
       intake_source: current_intake&.source,
       intake_referrer: current_intake&.referrer,
       intake_referrer_domain: current_intake&.referrer_domain,
+      **(current_intake.present? ? current_intake.mixpanel_data : {})
```
tdooner: Hmm, using the double splat operator is a using a new pattern for merging hashes - we could extend the existing pattern of using `Hash#merge` instead with something like this (on the next line):

```ruby
default_data.merge!(current_intake.mixpanel_data) if current_intake.present?
```


Also, something to consider would be to move the other `current_intake` fields above into the `mixpanel_data` method (the `source`, `referrer`, and `referrer_domain`)

app/models/intake.rb
```diff
@@ -257,4 +257,13 @@ def external_id
 
     ["intake", id].join("-")
   end
+
+  def mixpanel_data
+    {
+      primary_filer_age_at_end_of_tax_year: primary_user&.age_end_of_tax_year.to_s,
+      spouse_age_at_end_of_tax_year: spouse&.age_end_of_tax_year.to_s,
+      primary_filer_disabled: had_disability,
+      spouse_disabled: spouse_had_disability,
```
tdooner: Nice, this will submit "unfilled" / "yes" / "no", right?

app/models/user.rb
```diff
@@ -118,4 +118,8 @@ def ssn_last_four
   def opted_into_notifications?
     sms_notification_opt_in_yes? || email_notification_opt_in_yes?
   end
+
+  def age_end_of_tax_year
+    intake.tax_year - Date.parse(birth_date).year if birth_date
```
tdooner: a style nitpick: I find the `if birth_date` guard clause combined with the implicit return to be a bit nonintuitive due to the low associativity of `if` by the ruby parser. Mentally, I parse the statement as

```ruby
intake.tax_year - (Date.parse(birth_date).year if birth_date)
```
which is not how ruby does it (and doesn't work). I think it's more clear to explicitly return for cases we aren't expecting the math to work:

```ruby
def age_end_of_tax_year
  return unless birth_date.present?

  intake.tax_year - Date.parse(birth_date).year
end
```

Do you agree?

spec/models/user_spec.rb
```diff
@@ -63,6 +63,14 @@
     end
   end
 
+  describe "#age_end_of_tax_year" do
+    let(:user) { build :user, birth_date: "1990-04-21" }
+
+    it "returns their age at the end of 2019" do
```
tdooner: Since birth_date can be nil on intakes before we get ID.me back, can you also add a test for that case?

app/controllers/application_controller.rb
jenny-heath: sorry, i'm completely missing the `mixpanel_data` method? is it in this file?

app/models/intake.rb
jenny-heath: yup!

app/models/user.rb
jenny-heath: i think that looks better! the `if birth_date` was a quick fix for a broken test haha

what is "low associativity"?

tdooner: Ah, I might have used the wrong word, "precedence" might be the right one. I was going for the concept explained here: https://www.tinfoilsecurity.com/blog/ruby-demystified-and-vs except for "if" rather than "and"/"or".

It's a fancy way to describe the concept of where parentheses are automatically added to an expression. For example

```ruby
> a = true && false      # parsed as: a = (true && false)
> a
true

> b = true and false      # parsed as: (b = true) && false
> b
false
```

jenny-heath: ohhh got it - thanks!

app/models/intake.rb
```diff
@@ -270,4 +270,16 @@ def get_or_create_spouse_auth_token
...
+      intake_source: source,
+      intake_referrer: referrer,
+      intake_referrer_domain: referrer_domain,
+    }
```
bengolder: I want to piggy back on this to ask for a few more key metrics, if you don't mind:
- whether they have dependents
- whether they are filing joint
- whether they have any dependent under 6 (we already have a method for this I think)
- number of dependents (as a string, otherwise mixpanel will butcher the number)

app/models/intake.rb
```diff
@@ -270,4 +270,23 @@ def get_or_create_spouse_auth_token
...
+    {
+      intake_source: source,
+      intake_referrer: referrer,
+      intake_referrer_domain: referrer_domain,
+      primary_filer_age_at_end_of_tax_year: primary_user&.age_end_of_tax_year.to_s,
+      spouse_age_at_end_of_tax_year: spouse&.age_end_of_tax_year.to_s,
+      primary_filer_disabled: had_disability,
+      spouse_disabled: spouse_had_disability,
+      had_dependents: dependents.count > 0 ? "yes" : "no",
+      number_of_dependents: dependents.count.to_s,
```
tdooner: one last thing -- for performance sake. The `count` method executes a SQL query every time, so the above two lines are going to do 2 queries on every page view. You should use the `dependents.size` method instead:

https://apidock.com/rails/ActiveRecord/Associations/CollectionAssociation/size

This method will run the `COUNT(*)` SQL only if it needs to, which it shouldn't because the `dependents` array is loaded above (in the `dependents_under_6` expression) so it can just count the number of items in the array.

app/models/intake.rb
```diff
@@ -270,4 +270,23 @@ def get_or_create_spouse_auth_token
     update(spouse_auth_token: new_token)
     new_token
   end
+
+  def mixpanel_data
+    dependents_under_6 = dependents.any? { |dependent| dependent.age_at_end_of_tax_year < 6 }
+    had_earned_income = had_a_job? || had_wages_yes? || had_self_employment_income?
```
bengolder: Should this be `had_self_employment_income_yes?` like `had_wages_yes?`

jenny-heath: oops good catch

PR #96, opened 2020-03-24
add action_mailer default_url_options to demo, staging, prod config

jenny-heath:


bengolder:
@jenny-heath can you add staging too?

PR #101, opened 2020-03-25
Save visitor id from cookies to intake for tracking

jenny-heath:


app/controllers/application_controller.rb
```diff
@@ -12,12 +12,19 @@ def include_google_analytics?
...
   def set_visitor_id
-    return if cookies[:visitor_id].present?
-    cookies.permanent[:visitor_id] = SecureRandom.hex(26)
+    if cookies[:visitor_id].present?
+      visitor_id = cookies[:visitor_id]
+    else
+      visitor_id = SecureRandom.hex(26)
+      cookies.permanent[:visitor_id] = visitor_id
+    end
+    if current_intake.present? && current_intake.persisted? && current_intake.visitor_id.blank?
```
bengolder: Do you think this could be simplified?

```
unless current_intake&.visitor_id.present?
```

jenny-heath: this unfortunately doesn't work because it will also try to update the visitor id when there is no `current_intake` - we have to check that it exists and doesn't have a visitor id, and also that it's persisted because this gets called as a before action on like everything (which we may want to change)

PR #103, opened 2020-03-26
Show spouse name placeholder if spouse unverified

bengolder:


spec/models/intake_spec.rb
```diff
@@ -375,6 +375,19 @@
...
+      let(:intake) do
+        create :intake, was_full_time_student: "yes", spouse_was_full_time_student: "yes"
+      end
+      before do
+        create :user, intake: intake, first_name: "Henrietta", last_name: "Huckleberry"
+      end
+
+      it "Shows a placeholder for the spouse name" do
+        expect(intake.student_names).to eq(["Henrietta Huckleberry", "Your spouse"])
+      end
```
tdooner: nit: spacing before `before` and the capitalized "Shows"

app/models/intake.rb
```diff
@@ -245,10 +245,14 @@ def any_students?
       dependents.where(was_student: "yes" ).any?
   end
 
+  def spouse_name_or_placeholder
+    spouse.present? ? spouse.full_name : "Your spouse"
```
tdooner: fun rails tip: this exists to handle this pattern, should you so choose:

```ruby
spouse.full_name.presence || "Your Spouse"
```

bengolder: @tdooner that works for spouse presence or just full name presence?

bengolder: looks like that's for full_name. I changed it to

```ruby
spouse&.full_name || "Your spouse"
```

spec/models/intake_spec.rb
bengolder: done, thanks!

app/models/intake.rb
tdooner: Sorry yeah, I misread this. That looks good.

PR #104, opened 2020-03-26
Track question answers

jenny-heath:


app/controllers/questions/mailing_address_controller.rb
```diff
@@ -4,7 +4,7 @@ def section_title
       "Household Information"
     end
 
-    def custom_tracking_data
+    def tracking_data
       {
         mailing_address_same_as_idme_address: current_intake.address_matches_primary_user_address?
       }
```
bengolder: I feel like it would be harmless to add `state` here or we should just send that on intake for all events. Also, i checked mixpanel, and `mailing_address_same_as_idme_address` has been false every time.

jenny-heath: added to intake mixpanel data!

PR #112, opened 2020-03-30
Amend set_visitor_id to get visitor_id from current intake first

jenny-heath:
- write up brief job to do backfill

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,12 @@ class ApplicationController < ActionController::Base
   end
 
   def set_visitor_id
-    if cookies[:visitor_id].present?
+    return if current_intake&.visitor_id.present? && current_intake&.visitor_id == cookies[:visitor_id]
+
+    if current_intake&.visitor_id.present?
+      cookies.permanent[:visitor_id] = current_intake.visitor_id
+      return
+    elsif cookies[:visitor_id].present?
       visitor_id = cookies[:visitor_id]
     else
       visitor_id = SecureRandom.hex(26)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,11 +12,8 @@ class ApplicationController < ActionController::Base
   end
 
   def set_visitor_id
-    return if current_intake&.visitor_id.present? && current_intake&.visitor_id == cookies[:visitor_id]
-
     if current_intake&.visitor_id.present?
       cookies.permanent[:visitor_id] = current_intake.visitor_id
-      return
     elsif cookies[:visitor_id].present?
       visitor_id = cookies[:visitor_id]
     else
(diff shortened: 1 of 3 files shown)


app/jobs/backfill_visitor_ids_job.rb
```diff
@@ -0,0 +1,11 @@
+class BackfillVisitorIdsJob < ApplicationJob
+  queue_as :default
+
+  def perform
```
jenny-heath: was thinking this could be done in the console but didn't want to do it on my own! let me know what you think is the best way to do this

app/controllers/application_controller.rb
```diff
@@ -12,7 +12,12 @@ def include_google_analytics?
   end
 
   def set_visitor_id
-    if cookies[:visitor_id].present?
+    return if current_intake&.visitor_id.present? && current_intake&.visitor_id == cookies[:visitor_id]
```
jenny-heath: added an early return because this happens as a before action on like everything?

app/controllers/application_controller.rb
```diff
@@ -12,7 +12,12 @@ def include_google_analytics?
...
+
+    if current_intake&.visitor_id.present?
```
jenny-heath: also realized that we were checking cookies and not overwriting visitor_id on intakes but not checking intakes first - does this seem right? happy to talk it over in a call because it's a little confusing imo

app/controllers/application_controller.rb
tdooner: seems like a win from a code formatting perspective. I'm not sure the performance implications are very large because there aren't any database calls in the logic that this skips.

app/controllers/application_controller.rb
tdooner: okay, I wrote out a big thing here but I'm not going to post it because I realized I don't have context.

I suspect the logic here can be simplified if we set the `visitor_id` when the Intake is created, in `Questions::FeelingsController`. This would be nice because it's also the same as how we set `source` and `referrer`.

I think this logic would be simpler, I think it could be accomplished like this:

```ruby
def set_visitor_id
  if current_intake&.visitor_id.present?
    cookies.permanent[:visitor_id] = current_intake.visitor_id
  else
    cookies.permanent[:visitor_id] ||= SecureRandom.hex(26)
  end
end
```

But yeah, I am very likely missing some context or behavior that needs to happen here. Happy to chat.

app/jobs/backfill_visitor_ids_job.rb
tdooner: Definitely big 👍 for not doing things in the console. Going through code review is a good idea, and can prevent typos in production that can affect user data. I think a job makes sense for this kind of thing, also just a simple class like I did here worked pretty well: https://github.com/codeforamerica/vita-min/commit/2fb2916e5fa7c19946182a56e476af7d50fbc45c

app/controllers/application_controller.rb
jenny-heath: hm true! still not great with performance stuff haha. i think i'll remove the line so the whole thing is a little less confusing

PR #114, opened 2020-03-31
Update content on ID.me sign in page

jenny-heath:
the CSS was probably more complicated than it needed to be tho


app/assets/stylesheets/application.scss
```diff
@@ -196,3 +201,9 @@ body {
     }
   }
 }
+
+.inherit-link-color {
```
tdooner: based on the name of this class I would expect it to be a class applied to the link itself (instead of its parent). would that approach work here? I think it's also a bit more intuitive to have these "helper classes" apply directly to the elements they affect, whenever possible

app/views/questions/identity/edit.html.erb
```diff
@@ -11,27 +11,47 @@
...
             </div>
 
-            <p>
-              We are partnering with ID.me to protect your identity and ensure that your personal information is accurate and secure.
-            </p>
-
-            <p>
-              You will need to upload your ID and take a selfie, so we can be sure it’s you! It usually takes about 5 minutes to complete.
-            </p>
+            <div class="step">
```
tdooner: do we use BEM naming for CSS styles? I guess this is not relevant to this commit, but a BEM naming convention for this would be `steps__step`, which would also allow the SCSS to be un-nested from within `steps`.

jenny-heath: ah, i think we haven't been the most diligent about sticking to a naming system - i'd love to go in and clean up some things like this one of these days...

PR #117, opened 2020-04-01
Add a list of exceptions to ignore for Sentry

tdooner:
All but `ActionController::UnknownFormat` came from [GCF's sizeable
list][1]. The UnknownFormat error happened because someone tried to
request a JSON version of our homepage.

[1]: https://github.com/codeforamerica/gcf-backend/blob/master/config/initializers/sentry.rb


bengolder:
I'm hesitant to merge. It would be good to revisit this after more discussion with the group.

tdooner:
@bengolder I updated this to only include the ActionController::UnknownFormat exception we explicitly ignored in Sentry. I would advocate we start using this list whenever in Sentry we are pretty sure we never want to hear about an error again.

bengolder:
Souds good, thanks Tom!

PR #119, opened 2020-04-01
Backfill filing_years in DB and ZD, add validation to backfill form

jenny-heath:


app/forms/backtaxes_form.rb
```diff
@@ -1,7 +1,21 @@
...
+
+  def at_least_one_year
+    chose_one = needs_help_2016 == "yes" ||
+      needs_help_2017 == "yes" ||
+      needs_help_2018 == "yes" ||
+      needs_help_2019 == "yes"
+    unless chose_one
+      errors.add(:needs_help_2016, "Please pick at least one year.")
+    end
+    chose_one
```
tdooner, line 19: FYI - no need to return false here for the validation to work. The object will be invalid if any errors are added to it.

app/lib/zendesk_filing_years_backfill.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def update_db
+      never_answered = Intake.where(needs_help_2016: "unfilled", needs_help_2017: "unfilled", needs_help_2018: "unfilled", needs_help_2019: "unfilled")
+      never_answered.each { |intake| intake.update(needs_help_2019: "yes") }
+    end
+
+    def update_zendesk_tickets
+      Intake.find_each do |intake|
+        service = ZendeskIntakeService.new(intake)
+        service.update_filing_years
```
tdooner, line 20: Good call using the `ZendeskIntakeService` for this. It is how I should have done the other backfill (if I had used it in my previous backfill I wouldn't have needed the `include ZendeskServiceHelper` part at the top of here)

app/services/zendesk_intake_service.rb
```diff
@@ -63,6 +63,20 @@ def create_intake_ticket
     )
   end
 
+  def update_filing_years
```
tdooner, line 66: the intent is that this method will later be removed, right? I think maybe we should use a pattern of putting `backfill` in the name just to make it clear in the future. I assume you'll clean this up once it's done though so no worries for this commit.

app/lib/zendesk_filing_years_backfill.rb
jenny-heath: oops i forgot to take that out - good catch

PR #122, opened 2020-04-02
Add incoming webhook for Zendesk

bengolder:
Implements: https://www.pivotaltracker.com/story/show/171982764

This adds a new controller for receiving incoming JSON POSTs from Zendesk.
It requires Basic HTTP Authentication. The credentials have been added to the credentials file.
The scope for the current ticket is only to have an endpoint that is ready to receive the events. A later feature will begin to take action in response to the incoming JSONs.

I've tested this manually with Zendesk, using ngrok, to make sure that the current test webhook is configured to send and the app can successfully receive the POSTs.


app/controllers/zendesk_webhook_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  private
+
+  def authenticate
+    authenticate_or_request_with_http_basic do |name, password|
+      expected_name = EnvironmentCredentials.dig(:zendesk_webhook_auth, :name)
+      expected_password = EnvironmentCredentials.dig(:zendesk_webhook_auth, :password)
+      ActiveSupport::SecurityUtils.secure_compare(name, expected_name) &&
+        ActiveSupport::SecurityUtils.secure_compare(password, expected_password)
+    end
+  end
```
tdooner, line 18: curious, did you consider using `http_basic_authenticate_with`?
https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Basic.html

(This looks fine too. Good idea to use the `secure_compare`.)

spec/controllers/zendesk_webhook_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+require "rails_helper"
+
+RSpec.describe ZendeskWebhookController, type: :controller do
+  let(:valid_auth_credentials) do
+    ActionController::HttpAuthentication::Basic.encode_credentials("testname", "n1c3!")
+  end
+
+  let(:invalid_auth_credentials) do
+    ActionController::HttpAuthentication::Basic.encode_credentials("l33tH4x0r","k3rn3l_k!d")
```
tdooner, line 9: lol

app/controllers/zendesk_webhook_controller.rb
bengolder: @tdooner I actually did that at first, but it didn't work with tests :/ . `http_basic_authenticate_with` expects a username and password and is evaluated when the class is loaded, which prevented me from mocking out a fake username and password inside the test (the class is loaded before the `before` block is run). I needed it to evaluate the username & password at runtime, so went for the block-based implementation.

PR #125, opened 2020-04-02
Route returns by source param

jenny-heath:


spec/services/zendesk_intake_service_spec.rb
```diff
@@ -307,6 +309,73 @@
   end
 
   describe "#new_ticket_group_id" do
+    context "when there is a source parameter" do
+      context "when there is also a state" do
+        let(:source) { "uwkc" }
+        let(:state) { "ak" }
```
bengolder, line 315: two other specs that would help:
- when there is a souce but it does not match any organization (for example: "propel")
- when there is a source that is longer than just the organization code (but contains the org code in the beginning, for example: "uwkc_flyer")

app/services/zendesk_intake_service.rb
```diff
@@ -86,6 +86,33 @@ def update_filing_years
...
+    elsif source.start_with?("ia")
+      EitcZendeskInstance::ONLINE_INTAKE_IA_AL
+    elsif source.start_with?("goodwillsr")
+      EitcZendeskInstance::ONLINE_INTAKE_GWISR
+    elsif source.start_with?("fc")
+      EitcZendeskInstance::ONLINE_INTAKE_FC
+    elsif source.start_with?("uwco")
+      EitcZendeskInstance::ONLINE_INTAKE_UW_CENTRAL_OHIO
+    end
+  end
```
bengolder, line 113: potentially, this could be handled with hash lookup, and then maybe the spec should just test how it works with the hash rather than testing each indvidual partner. This might get crazy with dozens of partners.

PR #131, opened 2020-04-03
Ignore agent replies to SMS threads

bengolder:
This should quiet down the Sentry errors we were receiving from being unable to parse incoming messages.

https://www.pivotaltracker.com/story/show/172153973

It was blowing up whenever a Zendesk agent replied to an SMS thread. Now it should ignore that situation, but still raise an error if it gets something unexpected.

### DIFF
--- a/app/controllers/email_controller.rb
+++ b/app/controllers/email_controller.rb
@@ -5,17 +5,18 @@ class EmailController < ApplicationController
   ZENDESK_SMS_REGEX = /〒\nphone: \+(?<phone_number>[0-9]{11})\nticket_id: (?<ticket_id>[0-9]*)\nbody: (?<body>.+)\n〶/m
 
   def create
-    raise StandardError unless sent_to_known_email_hook?
+    unless sent_to_known_email_hook?
+      raise StandardError.new("Email webhook received invalid email to #{params[:to]}")
+    end
 
-    body = params[:text]
-    match = body.match(ZENDESK_SMS_REGEX)
+    unless valid_update_from_user? || from_agent_replying_to_thread?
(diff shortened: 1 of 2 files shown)


app/controllers/email_controller.rb
```diff
@@ -5,17 +5,18 @@ class EmailController < ApplicationController
...
-    raise StandardError unless sent_to_known_email_hook?
-
-    body = params[:text]
-    match = body.match(ZENDESK_SMS_REGEX)
+    unless sent_to_known_email_hook?
+      raise StandardError.new("Email webhook received invalid email to #{params[:to]}")
+    end
 
-    raise "Could not parse incoming message (#{params[:subject]}) from: #{params[:from]}" unless match
+    unless valid_update_from_user? || from_agent_replying_to_thread?
```
tdooner, line 12: the conditionals here confused me a bit, specifically this one in comparison to the next conditional (switching between `unless` and `if`). I think I personally find the combination more readable if this one uses the `if` variant 

> if !valid_update_from_user? && !from_agent_replying_to_thread?

this helps make the logic more clear the cases that cause the error vs the cases that cause the logic below to happen.

bengolder: Thank you for the input, it's a good point. I agree that the conditionals are a bit confusing. I'm going to leave it because:
1. I feel like both types of conditionals are a bit confusing and slightly prefer the `unless`. But happy to talk more to reach a consensus on situations like this.
2. I want the whole thing to go away and become a webhook!

tdooner: Makes sense, yeah! I think there is some unavoidable confusion since there's tricky logic to capture. Thanks for considering!

PR #133, opened 2020-04-06
Add lograge gem for JSON logging

tdooner:
This is a start of JSON logging -- initially to output one JSON line per
request. We add a hash of additional info to aid in debugging based off
the logs.

We mostly followed this guide:
https://docs.datadoghq.com/logs/log_collection/ruby/

We also discovered that Raven allows us to add extra context to our error
events, so we'll send some non-PII that will help us debugging.

https://www.pivotaltracker.com/story/show/172084932


tdooner:
Tests passed, so we're going to merge and see what this looks like in demo.

PR #135, opened 2020-04-07
Don't use TaggedLogging in deployment environments

tdooner:
We want to use JSON based logs, so we need to remove the
ActiveSupport::TaggedLogging wrapper so the line is output as full JSON.

Example log line before:

    I, [2020-04-07T00:05:57.908012 #5]  INFO -- : [...] [...] [...] {"method":"GET","path":"/"...

After:

    {"method":"GET","path":"/","format":"html","controller":"PublicPagesController...


tdooner:
I tested this locally pretty well, I'm going to merge this and observe carefully in demo.

PR #136, opened 2020-04-07
Add `request_id` to Sentry context

tdooner:
This will help us find errors that happened and correlate between Kibana
and Sentry.


tdooner:
Jenny and I looked at this!

PR #139, opened 2020-04-08
Remove section titles

kayline:
- no section titles for any questions
- adjust styling so we don't have vestigial whitespace

[#172107063]


app/views/dependents/_form.html.erb
```diff
@@ -10,10 +10,6 @@
     <div class="grid">
       <div class="grid__item width-three-fourths shift-one-eighth">
         <%= yield :notices %>
-        <h3 class="section-title">
-          <%= image_tag("#{section_title.parameterize}.svg", alt: "") %>
```
tdooner, line 14: Can we remove these SVGs now if they aren't in use anywhere?

PR #143, opened 2020-04-10
Update notifications page

jenny-heath:


app/assets/stylesheets/atoms/_form-elements.scss
```diff
@@ -12,4 +12,11 @@
   @media screen and (min-width: $tablet-up) {
     width: 25em;
   }
+}
+
+.text-input--small-label {
+  .form-question {
+    font-size: 1.9rem;
+    margin-bottom: 1rem;
+  }
```
tdooner, line 21: it's a shame to add a class like this for just one page -- does it need it?

jenny-heath: well, the size of the text otherwise is the size of a yes no question heading, which is to say huge. i think it needs it but maybe there's a better way to do it?

PR #145, opened 2020-04-10
Add "Personal Info" page for name & state

tdooner:
note: I'm not sure about the need for the new state column. I'll check and update accordingly.
___
This adds a simple page for name & state at the beginning of the flow.

I debated putting state in the existing `state` column (populated by
/questions/mailing-address), but decided to make a new column in case
the mailing address page will remain.


tdooner:
Screenshot of Page:
image

Page with validation errors:
image

PR #147, opened 2020-04-13
Skip oversize uploads #172293223

kayline:
If a file is too big, adds the following note to the comment:

"The file [filename] could not be uploaded because it exceeds the maximum size of [file_size_limit]MB."

If multiple files exceed the limit, the message will be repeated for each file name.

The size limit depends on the Zendesk instance for the ticket.

### DIFF
--- a/app/helpers/zendesk_service_helper.rb
+++ b/app/helpers/zendesk_service_helper.rb
@@ -31,6 +31,9 @@ module ZendeskServiceHelper
     "Pacific/Honolulu" => "Hawaii",
   }.freeze
 
+  # Zendesk caps uploads at 20MB
+  MAXIMUM_UPLOAD_SIZE = 20000000
+
   def client
     @client ||= instance.client
   end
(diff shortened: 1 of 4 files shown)


jenny-heath:
looks good! i'd just make sure to test this locally if you can

bengolder:
Could we add the document type to the comment as well?

app/helpers/zendesk_service_helper.rb
```diff
@@ -169,6 +169,22 @@ def append_multiple_files_to_ticket(ticket_id:, file_list:, comment: "", fields:
     success
   end
 
+  def append_file_or_add_comment(file, ticket, comment)
+    if file[:file].size < file_size_limit
+      ticket.comment.uploads << file
+    else
+      comment.concat("\n\nThe file #{file[:filename]} could not be uploaded because it exceeds the maximum size of #{file_size_limit/1000000}MB.")
```
bengolder, line 176: I think it would be helpful to include the document type here as well, if possible.

kayline: We no longer have a reference to the document at this point, just the file blob and filename. We could try to find the document on the intake with an upload matching the filename, but that would rely on clients never uploading two files with the same name for different documents.

There will be a list of filenames with their document type directly above any error comments, so I think it's ok not to state the document type again in the oversize error message. I agree it would be helpful to explicitly list it twice, but I don't see a straightforward path there. Merging for now to get the basic skip functionality in place, can noodle with the error message more in future if it seems needed?

PR #148, opened 2020-04-14
ID guidance page

jenny-heath:


app/controllers/documents/intro_controller.rb
```diff
@@ -1,16 +1,11 @@
 module Documents
-  class IntroController < Questions::QuestionsController
+  class IntroController < DocumentUploadQuestionController
```
jenny-heath, line 2: i don't know why this wasn't a document upload controller to begin with, but i made this change because it doesn't require us to define `next_path` anymore, the one on `DocumentUploadQuestionController` works. but maybe i'm overlooking something?

app/lib/document_navigation.rb
```diff
@@ -32,7 +32,8 @@ class DocumentNavigation
     "Requested" => Documents::RequestedDocumentsController,
   }.freeze
   BEFORE_CONTROLLERS = [
-    Documents::IntroController
+    Documents::IntroController,
+    Documents::IdGuidanceController,
```
jenny-heath, line 36: makes me nervous adding to this pattern 😬

bengolder: Agreed. It's not ideal. This would be good to discuss together.

PR #150, opened 2020-04-14
 Always send E164 format phone number to Zendesk 

tdooner:
See: https://www.pivotaltracker.com/story/show/172319194

We want to use the E164 format (e.g. "+15105551234") for phone numbers
to ensure maximum ability to send users text messages via Twilio.

The first commit adds a new method to drop off and user models called
`standardized_phone_number` which provides the phone number in this
format, and updates the two call sites of
`ZendeskServiceHelper#find_or_create_end_user` to use this standardized
number.

The second commit adds a script to backfill E164 phone numbers in Zendesk based on
Intake and DropOffs.


app/models/intake_site_drop_off.rb
```diff
@@ -109,10 +109,16 @@ def phone_number=(value)
...
+  # Returns the phone number formatted for user display, e.g.: "(510) 555-1234"
   def formatted_phone_number
     Phonelib.parse(phone_number).local_number
   end
 
+  # Returns the phone number in the E164 standardized format, e.g.: "+15105551234"
+  def standardized_phone_number
+    Phonelib.parse(phone_number).e164
+  end
+
```
bengolder: Thank you for adding these clarifying comments! Super helpful!

app/models/intake_site_drop_off.rb
```diff
@@ -109,10 +109,16 @@ def phone_number=(value)
...
 
+  # Returns the phone number formatted for user display, e.g.: "(510) 555-1234"
   def formatted_phone_number
     Phonelib.parse(phone_number).local_number
   end
 
+  # Returns the phone number in the E164 standardized format, e.g.: "+15105551234"
+  def standardized_phone_number
+    Phonelib.parse(phone_number).e164
+  end
```
bengolder: If `phone_number` is `nil`, this would just return `nil`, correct?

app/models/intake_site_drop_off.rb
```diff
@@ -116,7 +116,7 @@ def formatted_phone_number
 
   # Returns the phone number in the E164 standardized format, e.g.: "+15105551234"
   def standardized_phone_number
-    Phonelib.parse(phone_number).e164
+    Phonelib.parse(phone_number, "US").e164
   end
```
bengolder: This seems like an improvement. Should it be a module or an external class? Should we use this to improve our phone number parsing in our drop off forms or anywhere else?

app/models/intake_site_drop_off.rb
tdooner: Yep, though I realized I didn't explicitly add a test for that. I'll push one up in this branch just to be sure.

app/models/intake_site_drop_off.rb
tdooner: Agreed. I think we could perhaps make a phone number wrapper class that includes these responsibilities as well as the /questions/phone-number edit page's "Add the country code if necessary" logic. I'll make a gardening story for it.

PR #152, opened 2020-04-15
Add SSN/ITIN doc uploads for primary and spouse

jenny-heath:
this has changed rather a lot since it was approved:
- the ID upload pages have been consolidated
- the SSN/ITIN pages have been consolidated with each other AND with the dependent SSN/ITIN page
- both pages have conditional copy based on whether they are filing jointly or not

single filers will see:
image
image

couples will see:
image
image


app/lib/document_navigation.rb
```diff
@@ -4,6 +4,8 @@ class DocumentNavigation
   DOCUMENT_CONTROLLERS = {
     "ID" => Documents::IdsController,
     "Spouse ID" => Documents::SpouseIdsController,
+    "Primary SSN or ITIN" => Documents::PrimarySsnItinsController,
```
tdooner, line 7: As far as I can tell this is the first time we're introducing the copy "Primary" to something that the user would see (in the docs intro list)... Should we keep this as just "SSN or ITIN" to avoid the case where the user isn't filing joint and would be confused about it?

app/views/documents/spouse_ssn_itins/edit.html.erb
```diff
@@ -0,0 +1,3 @@
+<% content_for :form_question, "Attach a photo of your spouse's Social Security Card or ITIN" %>
+
+<% content_for :form_help_text, "The IRS requires us to see a valid SSN or ITIN for tax preparation services." %>
```
tdooner, line 3: should we put "for your spouse" somewhere in here?

app/lib/document_navigation.rb
jenny-heath: ah so i thought about this - there's already an "SSN or ITIN" and it's for dependents. We could either: A) rename the existing document type and do a backfill, B) combine all of the SSN/ITIN pages (wouldn't allow us to know whose is whose) C) leave this with "Primary" and rename the primary ID page to have "Primary" also for consistency

app/views/documents/spouse_ssn_itins/edit.html.erb
jenny-heath: could! the ticket didn't specify but that seems helpful

bengolder:
handle the case of not having a spouse name. There is code on the intake for this.

PR #154, opened 2020-04-16
Remove id.me references from zendesk comments

jenny-heath:


app/controllers/documents/ids_controller.rb
```diff
@@ -3,6 +3,7 @@ class IdsController < DocumentUploadQuestionController
     def edit
       @title = "Attach a photo of your ID card"
       @help_text = "The IRS requires us to see a current drivers license, passport, or state ID."
+      # TODO: replace this with name from intake
```
jenny-heath, line 6: this is one of a few places where we get the client's name from the user record. is it part of this story to replace these with the new name column on intake?

app/services/zendesk_intake_service.rb
```diff
@@ -21,6 +21,7 @@ def instance_eitc?
 
   def create_intake_ticket_requester
     # returns the Zendesk ID of the created user
+    # TODO: move notification opt in columns to intake from user so contact info can also move
```
jenny-heath, line 24: this would require more changes to the notification page. also potentially a backfill of the opt-in columns if we are too move the column from user to intake?

app/services/zendesk_sms_service.rb
```diff
@@ -10,18 +10,18 @@ def instance
   end
 
   def handle_inbound_sms(phone_number:, sms_ticket_id:, message_body:)
-    users = User.where(phone_number: phone_number)
+    intakes = Intake.where(phone_number: phone_number)
```
jenny-heath, line 13: this change should be coupled with the change to `create_intake_ticket_requester`

bengolder:
@jenny-heath I'm closing this since we dealt with this in another branch

PR #159, opened 2020-04-21
Audit presence of title on every page

jenny-heath:
- add title to pages that don't have one
- add errors and title to question layout
- fix error_summary on form not that it was used anyway

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -5,7 +5,7 @@ class Form
 
   def error_summary(exclude_keys: [])
     if errors.present?
-      visible_errors = errors.messages.select{ |key, _| exclude_keys.include? key }
+      visible_errors = errors.messages.select{ |key, _| exclude_keys.exclude? key }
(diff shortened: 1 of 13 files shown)


app/views/layouts/question.html.erb
```diff
@@ -1,3 +1,10 @@
+<% content_for :page_title do %>
+  <% if @form&.errors.present? %>
```
tdooner, line 2: I think `if @form.error_summary.present?` would DRY up the error-visibility logic a little bit, but I don't see an issue with it as written. (it looks like we don't use the `exclude_keys` option anywhere, anyway)

app/forms/form.rb
```diff
@@ -5,7 +5,7 @@ class Form
 
   def error_summary(exclude_keys: [])
     if errors.present?
-      visible_errors = errors.messages.select{ |key, _| exclude_keys.include? key }
+      visible_errors = errors.messages.select{ |key, _| exclude_keys.exclude? key }
```
tdooner, line 8: lol, good catch. I guess we don't actually use `exclude_keys` anywhere?

PR #160, opened 2020-04-21
Remove non-standard duplicate phone numbers from Zendesk

tdooner:
As discovered in [the story][1], the presence of duplicate non-E164
phone numbers in Zendesk causes the text messages to error in sending.

This backfill will remove any non-standard phone numbers to hopefully
prevent this error going forward. Any phone numbers that have been
manually changed in Zendesk will not be deleted.

[1]: https://www.pivotaltracker.com/story/show/172390122


tdooner:
Yep, good question. I couldn't find any by clicking around the Fake tickets but I modified one fake ticket to be able to to test it: https://eitc.zendesk.com/agent/tickets/4332

Also I will be ensuring that other phone numbers are not removed (e.g. this ticket: https://eitc.zendesk.com/agent/tickets/4211)

PR #164, opened 2020-04-21
Replace user data with intake for zendesk

bengolder:


app/models/intake.rb
```diff
@@ -433,6 +429,17 @@ def zendesk_instance
     end
   end
 
+  def contact_info_filtered_by_preferences
+    contact_info = {}
+    contact_info[:phone_number] = phone_number if sms_notification_opt_in_yes?
```
tdooner, line 434: I would suggest copying the `standardized_phone_number` implementation from `User` and using it here. I know we validate phone numbers when we're saving them in the PhoneNumberForm, but using the `standardized_phone_number` method here could prevent bugs down the road by ensuring we're always sending the standardized number to Zendesk.

bengolder: @tdooner good catch, thank you!

PR #166, opened 2020-04-21
Add custom error pages to encourage users to contact us

tdooner:
We want to add better, non-default-Rails, error pages. By setting the
`config.exceptions_app` to our own application, we can handle these
errors.

This removes the 422 error page -- unprocessible entity -- since we
don't use it.

Screenshots:
500 error -
image

404 error -
image


jenny-heath:
@tdooner we actually have quite a few 422 errors. It might make sense to show the 500 page in those cases.
https://logs.vitataxhelp.org/app/kibana#/discover/c3012240-79be-11ea-bba6-edc4c6f1eb3d?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(jsonlog.path,jsonlog.status,jsonlog.error),filters:!(),index:'10c45f90-33df-11ea-bba6-edc4c6f1eb3d',interval:auto,query:(language:lucene,query:'jsonlog.status:422'),sort:!('@timestamp',desc))

tdooner:
Good call, I wonder why we get so many of those. I'll make it go to the server error too!

PR #167, opened 2020-04-22
Convert document-type / controller hash in DocumentNavigation to an ordered array-based list

bvandgrift:
resolves PT#172417481

this is one approach, and would lend itself to easily allowing for multiple types of controllers to handle a single document type.


bengolder:
Closing because this was implemented in another PR. Thank you @bvandgrift !

PR #168, opened 2020-04-22
Refactor DocumentNavigation to use an array

tdooner:
This flattens the objects in DocumentNavigation back into one big array,
and moves the responsibility for determining document type onto the
subclass controllers. Overview pages (which don't upload documents) can
return `nil`.


bengolder:
@tdooner Closing because you included this in your other PR.

PR #170, opened 2020-04-22
Move pages in flow to a post-ID.me world

tdooner:
(this is built atop https://github.com/codeforamerica/vita-min/pull/169, that branch should be merged first)

You have no idea how satisfying it was to delete the line `Questions::IdentityController` from the QuestionsNavigation.

This PR reorganizes the flow to a post ID.me world according to the comprehensive (and incredibly useful) spreadsheet here: https://docs.google.com/spreadsheets/d/10_2fQ_3WRKaUnbqG1xkv43JlmYqVeXjX2UJhoL2ChmM/edit#gid=0

Getting the tests to pass was the hardest part, and unraveling that is where most this diff comes from. I disabled a bunch of test that had to do with ID.me -- assuming we will go through and remove them more methodically later.

I tried to leave a clean commit history, I recommend reading the messages one-by-one.


app/forms/mailing_address_form.rb
```diff
@@ -12,11 +12,14 @@ def save
...
-    else
-      attributes = intake.primary_user.attributes
-    end
+    attributes = if intake.street_address.present?
+                   intake.attributes
+                 elsif intake.primary_user.present?
+                   intake.primary_user.attributes
+                 else
+                   {}
+                 end
```
bengolder, line 21: This is great. We could also set state to state_of_residence at this point too, just to make things easier for the user.

app/lib/document_navigation.rb
```diff
@@ -1,107 +1,81 @@
...
-
-  def index
-    all_controllers.index(@current_controller.class)
-  end
-
-  def seek(list)
-    list.detect do |controller_class|
-      controller_class.show?(@current_controller.current_intake)
-    end
+    controllers[(current_index + 1)..-1].find { |c| c.show?(intake) }
```
bengolder, line 79: This is terse but dense! Non-blocking, but I think this would be more readable if broken up into separate statements.

app/lib/document_navigation.rb
```diff
@@ -1,107 +1,81 @@
...
-      DOCUMENT_CONTROLLERS.key(controller_class)
+    def document_types_for_intake(intake)
+      controllers
+        .find_all { |c| c.show?(intake) }
+        .map(&:document_type)
+        .compact
+    end
+
+    def document_controller_for_type(document_type)
+      controllers.find { |c| c.document_type == document_type }
```
bengolder, line 65: Love that these became class methods!

app/lib/document_navigation.rb
```diff
@@ -1,107 +1,81 @@
...
     def controllers
-      DOCUMENT_CONTROLLERS.values
+      FLOW
     end
 
-    def all_controllers
-      BEFORE_CONTROLLERS + controllers + AFTER_CONTROLLERS
+    def first_for_intake(intake)
+      controllers.find { |c| c.show?(intake) }
     end
```
bengolder, line 55: "for_intake" was implied before. I like that you made it explicit

spec/controllers/questions/mailing_address_controller_spec.rb
```diff
@@ -25,13 +16,6 @@
       expect(response).to be_successful
     end
 
-    context "when mailing address on the intake is not present" do
-      it "pre-fills mailing address from the user" do
-        get :edit
-        expect(response.body).to include(user.street_address)
-      end
-    end
-
```
bengolder, line 34: I feel like if we are retaining the pre-fill functionality, we should keep the test. I'm okay either including or excluding user-based pre-fill, we should just test if including.

app/controllers/documents/requested_documents_later_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Documents
   class RequestedDocumentsLaterController < DocumentUploadQuestionController
     before_action :handle_session, only: :edit
-    skip_before_action :require_sign_in
+    skip_before_action :require_intake
```
bengolder, line 4: we should remove this line. This controller skipped sign in but does require an intake.

app/controllers/documents/send_requested_documents_later_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Documents
   class SendRequestedDocumentsLaterController < DocumentUploadQuestionController
-    skip_before_action :require_sign_in
+    skip_before_action :require_intake
```
bengolder, line 3: We should remove this line as well. This one does require an intake.

app/controllers/questions/identity_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Questions
   class IdentityController < QuestionsController
-    skip_before_action :require_sign_in
+    skip_before_action :require_intake
```
bengolder, line 3: We should remove this line too.

app/controllers/questions/feelings_controller.rb
```diff
@@ -1,6 +1,5 @@
 module Questions
   class FeelingsController < QuestionsController
-    skip_before_action :require_sign_in
```
bengolder, line 3: Does this controller need `skip_before_action :require_intake`?

spec/controllers/questions/job_count_controller_spec.rb
```diff
@@ -4,20 +4,19 @@
   let(:intake) { create :intake }
 
   before do
-    allow(subject).to receive(:user_signed_in?).and_return(true)
     allow(subject).to receive(:current_intake).and_return(intake)
   end
 
   describe "#edit" do
     context "when user not signed in" do
```
bengolder, line 11: --> "when user doesn't have an intake"

spec/features/web_intake/requested_documents_spec.rb
```diff
@@ -1,7 +1,8 @@
 require "rails_helper"
 
 RSpec.feature "Client uploads a requested document" do
-  scenario "client goes to the follow up documents link provided by their preparer" do
+  # TODO: What do we do about requested documents post-idme?
```
bengolder, line 4: We have already implemented token-based requested docs

app/lib/document_navigation.rb
bengolder: @bvandgrift: This could be a class-level constant.

spec/controllers/questions/identity_controller_spec.rb
```diff
@@ -4,12 +4,12 @@
   render_views
 
   describe "#edit" do
-    context "with a next_path param in the URL" do
+    xcontext "with a next_path param in the URL" do
```
bengolder, line 7: would this fail if run?

spec/controllers/questions/spouse_identity_controller_spec.rb
```diff
@@ -33,11 +33,11 @@
     end
 
     context "with missing_spouse param true" do
-      it "displays a message" do
+      xit "displays a message" do
```
bengolder, line 36: would this pass if run?

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe Users::OmniauthCallbacksController do
+RSpec.xdescribe Users::OmniauthCallbacksController do
```
bengolder, line 3: would this run? Could we skip particular unit specs within this instead of the whole thing?

app/helpers/documents_helper.rb
```diff
@@ -1,16 +1,24 @@
 module DocumentsHelper
-  MUST_HAVE_DOC_TYPES= [
+  ALREADY_SUBMITTED_DOC_TYPES = [
     "ID",
     "SSN or ITIN",
+    "Selfie"
+  ]
```
bengolder, line 6: Add "Other" document type to this list

spec/controllers/questions/identity_controller_spec.rb
tdooner: yep, it fails because there's no route anymore.

spec/controllers/questions/spouse_identity_controller_spec.rb
tdooner: nope, also fails for routes.

spec/controllers/users/omniauth_callbacks_controller_spec.rb
tdooner: yep. I just disabled the spouse auth ones, those were the only ones that failed.

app/controllers/documents/requested_documents_later_controller.rb
tdooner: @bengolder - I am pretty sure we need to skip the `require_intake` behavior of this controller, but not the `SendRequestedDocumentsLaterController`, since this controller  performs its own management of `session[:intake_id]`, and thus `current_intake` will be defined by the `handle_session` before_action, and will be available in the `Send...` controller.

PR #172, opened 2020-04-22
adds guard, coverage, updates README with how to use

bvandgrift:


tdooner:
It looks like Guard can also be used within Rubymine to speed up test runs! https://stackoverflow.com/questions/11996124/is-it-impossible-to-use-guard-with-rubymine/12000765#12000765

tdooner:
Ok, I verified that it doesn't break in Rubymine. Also, 99.25% coverage? Wow.

PR #178, opened 2020-04-23
Remove sign in stubs #172466279

kayline:
Also removes `send_additional_info_pdf` behavior


spec/models/intake_spec.rb
```diff
@@ -248,58 +235,6 @@
...
-    end
-
-    context "with one user" do
-      let!(:primary) { create :user, intake: intake }
-
-      it "returns that one user" do
-        expect(intake.primary_user).to eq primary
-      end
-    end
-  end
```
bengolder, line 283: I don't think we should remove these tests for the associations yet. We may need to use these to debug issues that arise after we make the change.

PR #181, opened 2020-04-23
Store instance #172461634

kayline:
Adds columns for `zendesk_instance_domain` and `zendesk_group_id` on Intake, and saves values to those columns the first time we calculate the value.

Includes script to backfill these values for all existing Intakes.
`IntakeInstanceDomainAndGroupIdBackfill.backfill_instance_domains!`
`IntakeInstanceDomainAndGroupIdBackfill.backfill_group_ids!`

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -404,7 +404,15 @@ class Intake < ApplicationRecord
     ].compact
   end
 
-  def zendesk_group_id
+  def get_or_create_zendesk_group_id
+    return zendesk_group_id if zendesk_group_id.present?
+
+    group_id = determine_zendesk_group_id
+    self.update(zendesk_group_id: group_id)
+    group_id
+  end
+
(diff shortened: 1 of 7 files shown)


app/models/intake.rb
```diff
@@ -345,14 +347,6 @@ def get_or_create_spouse_auth_token
...
   end
 
-  def get_or_create_requested_docs_token
-    return requested_docs_token if requested_docs_token.present?
-
-    new_token = SecureRandom.urlsafe_base64(8)
-    update(requested_docs_token: new_token, requested_docs_token_created_at: Time.now)
-    new_token
-  end
-
```
bengolder, line 355: I was super confused by this deletion at first. It looks like this was a duplicate method, correct?

kayline: Yes, the rebase duplicated it for some reason

PR #184, opened 2020-04-24
Write script to backfill any requested docs links that are missing

jenny-heath:
- we moved the place where the field is populated and some people may have as a result not had a link set on their ticket

Co-authored-by: Jonathan Greenberg <jgreenberg@codeforamerica.org>


spec/lib/zendesk_phone_number_backfill_spec.rb
```diff
@@ -1,71 +1,65 @@
 require "rails_helper"
 
-RSpec.describe ZendeskPhoneNumberBackfill do
-  describe '.update_zendesk!' do
-    let(:fake_zendesk_user) do
-      double(ZendeskAPI::User, id: 1234569879, phone: "5105551234", email: "testy@example.com", name: "Testy McTestuser")
-    end
-    let(:user_properties) { {} }
+RSpec.describe ZendeskRequestedDocsLinkBackfill do
```
bengolder: Excellent, but file should prob be renamed

PR #187, opened 2020-04-24
Trust home faq

loumoore:
This change contains
- homepage refresh, including link to new FAQ page
- new FAQ page

Some smaller updates included
- fix minor layout/ui issues relative to homepage design
- update partner logo images
- defer zendesk loading for performance improvement
- change default font to work around an active Chrome bug that messes with the display of large system fonts (https://bugs.chromium.org/p/chromium/issues/detail?id=1057654&q=font%20macos%20catalina&can=2)


app/assets/stylesheets/components/_buttons.scss
```diff
@@ -7,6 +7,7 @@
   background-color: black;
   border-color: black;
   box-shadow: none;
+  padding: $s35/2 4rem;
```
bengolder, line 10: there's a small risk that this could make some odd buttons elsewhere. `--primary` & `--cta` are mainly intended as color changes that could be applied to different size buttons. I'm not sure if we have any small `button--primary` but would be good to separate padding & anything with absolute sizing from the color styles.

app/assets/stylesheets/atoms/_variables.scss
```diff
@@ -1,5 +1,8 @@
 // Fonts
-$font-system:         	-apple-system, BlinkMacSystemFont,
+$font-gyr:         	    "Helvetica Neue", -apple-system, BlinkMacSystemFont,
                         "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell",
-                        "Fira Sans", "Droid Sans", "Helvetica Neue",
-                        sans-serif;
\ No newline at end of file
+                        "Fira Sans", "Droid Sans", sans-serif;
```
bengolder: Thanks for fixing this!

app/assets/stylesheets/atoms/_variables.scss
```diff
@@ -1,5 +1,8 @@
...
+
+// Colors
+$color-green-money: #C6F9EC;
```
bengolder, line 7: Best color name 💸

app/assets/stylesheets/templates/_about-us.scss
```diff
@@ -34,17 +34,6 @@
...
-    display: flex;
-    flex-wrap: wrap;
-    justify-content: space-between;
-    align-items: center;
-    margin: 0 -0.5em;
-    &__listing {
-      padding-bottom: 1em;
-      margin: 0 .5em;
-    }
-  }
```
bengolder, line 47: Thanks for pulling this into it's own file!

app/views/layouts/application.html.erb
```diff
@@ -113,6 +113,15 @@
...
     <!-- Start of eitc Zendesk Widget script -->
+    <script type="text/javascript">
+        window.zESettings = {
+            webWidget: {
+                chat: {
+                    connectOnPageLoad: false
+                }
+            }
+        };
+    </script>
```
bengolder, line 124: Awesome, thank you! Can't wait for the performance boost!

app/views/public_pages/faq.html.erb
```diff
@@ -0,0 +1,131 @@
...
+        </p>
+
+        <h2 class="h3 spacing-above-35">
+          Help getting started
+        </h2>
+
+        <div class="accordion">
+          <button href="#" class="accordion__button" aria-expanded="true" aria-controls="a1">
+            What should I do if I haven’t filed in several years?
+          </button>
```
bengolder, line 22: Nonblocking, but if these were `h3` tags wrapped in `a` tags, it would improve keyboard navigation and thereby any assistive tech navigation, including screen readers.

app/assets/stylesheets/components/_buttons.scss
loumoore: moving it into .button instead (to solve the "--primary & --cta are mainly intended as color" issue) and moving it out of this global context into the homepage specific .body--home selector in application.scss (to solve the "could make some odd buttons elsewhere issue")

i was assuming the wider buttons were desired in the design generally but this is definitely safer and can always be expanded later. thanks!

app/views/public_pages/faq.html.erb
loumoore: ah, i was going by the honeycrisp examples, assuming given the aria attributes that it had already been optimized for accessibility. 

it's easy enough to change this with just a small piece of additional css for the accordion molecule, so I'm submitting an update to do that. I tested with the WAVE plugin before and after this change, and while I must admit I'm not totally sure what I should be looking for, I observed:
1) no errors with the button approach or the a>h3 approach
2) able to use tab to navigate and use the enter key to open the FAQ question with both approaches too
3) the a>h3 approach introduces an orange border highlight on the h3s when you navigate which I assume is a desired improvement for accessibility. however, it also introduces an extra tab press to navigate from question to question, and the space key, which also opened questions with the button approach, no longer works. i'm not sure if those things are problems or just..different

i'm going to go ahead and submit this update on the assumption that the orange border thingy is a plus and the other changes I observed are not a big deal.

PR #191, opened 2020-04-27
Add source mapping for UWCCR, Branches, and Hispanic Unity

jonsgreen:
- Move specs to shared examples for easier testing using actual group id
for a bit better verification with ids from Pivotal Tracker
- Possibly more idiomatic use of find/detect instead of each for
matching group_id
- enable focused running of specs

(#172511772) [Tuesday 4/28] Enable routing for UWCCR, Branches, and
Hispanic Unity


jonsgreen:
Note: May need to confirm that the source will come through correctly for "RefundDay-B" and "RefundDay-H" correctly with the dash.

PR #195, opened 2020-04-28
changes heading and help copy, updates feature spec

bvandgrift:
the story was unclear as to whether the body should be above or below the heading. below made the most sense to me.


bengolder:
merging

PR #197, opened 2020-04-28
[Do Not Merge] WIP questionable refactoring of required document submit w/form object

jonsgreen:
- call edit to instantiate needed variables and render edit so that form
errors are not lost
- create new RequiredDocumentUploadForm form object inheriting from
DocumentTypeUploadForm to handle document validation before next step
- use class instance variable for scoped_attributes to allow inheritance
for RequiredDocumentUploadForm
- use private error helper methods from cfa_styleguide for rendering
consistent errors in views; quite funky

(#172510637) Refactor Required Document Validation


app/controllers/documents/document_upload_question_controller.rb
```diff
@@ -9,30 +9,38 @@ def edit
...
+        @form.save
         after_update_success
         track_document_upload
+        if @form.try(:next_step)
+          redirect_to next_path and return
+        end
       end
 
-      redirect_to action: :edit
+      edit
```
jonsgreen, line 28: This is clearly not ideal and suggests that some larger refactoring is required such that instance variables required for rendering the edit template will be instantiated from all inherited controllers.

app/controllers/documents/document_upload_question_controller.rb
```diff
@@ -9,30 +9,38 @@ def edit
...
+      self.class.form_class.name.underscore
     end
 
     def self.form_class
-      DocumentTypeUploadForm
+      if DocumentNavigation::REQUIRED_DOCUMENT_TYPES.include?(document_type)
+        RequiredDocumentUploadForm
+      else
+        DocumentTypeUploadForm
+      end
```
jonsgreen, line 43: This is one of the less than ideal consequences of created a child class for the RequiredDocumentUploadForm. Creating a different controller for required document upload forms would be an option or just combining the logic into the DocumentTypeUploadForm which was the approach that @bengolder suggested in an original spike that we did with @jenny-heath.

app/forms/document_type_upload_form.rb
```diff
@@ -1,4 +1,7 @@
 class DocumentTypeUploadForm < QuestionsForm
+  def self.scoped_attributes
+    @@scoped_attributes ||= {}
+  end
```
jonsgreen, line 4: This was another unfortunate addition as a consequence of the RequireDocumentUploadForm inheriting from the DocumentTypeUploadForm and wanting to have the scoped_attributes set by `set_attribute_for` to be shared. Again just sticking with the DocumentTypeUploadForm is probably the way to go.

app/views/layouts/document_upload.html.erb
```diff
@@ -40,8 +40,12 @@
           <%# document upload form %>
           <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, id: "file-upload-form", class: "spacing-below-60" do |f| %>
             <div class="document-upload">
-              <div class="file-upload">
-                <%= f.file_field(:document, class: "form__documentuploader file-input", data: { "upload-immediately" => true}) %>
+              <div class="file-upload form-group <%= f.send(:error_state, @form, :document) %>">
+                <%= f.file_field(
+                  :document,
+                  class: "form__documentuploader file-input#{f.send(:error_state, @form, :document)}",
+                  data: { "upload-immediately" => true}
```
jonsgreen, line 47: Unfortunately since this is not using a form builder public helper method I am calling a private method to get the error styling into the inputs. Maybe investing some time to create an enhanced cfa_file_field in VitaMinFormBuilder would help clean this up?

app/views/layouts/document_upload.html.erb
```diff
@@ -52,34 +56,15 @@
                     Take a picture
                   </span>
                 <% end %>
+                <%= f.send(:errors_for, @form, :document)&.html_safe %>
```
jonsgreen, line 59: Again private method called...boo.

app/views/layouts/document_upload.html.erb
```diff
@@ -52,34 +56,15 @@
...
-              <%= image_tag "checkmark--white.svg", alt: "" %>
-              I'm done for now
-            <% end %>
-            <%= link_to next_path, class: "button button--wide button--icon", "data-track-click": "dont_have_doc", "data-track-attribute-document_type": "#{document_type}" do %>
-              <%= image_tag "crossmark--black.svg", alt: "" %>
-              I don't have this document
-            <% end %>
-          <% end %>
-
+          <%= render "#{@form.class.name.underscore}_actions" %>
```
jonsgreen, line 67: So this was the main motivation for having a different RequireDocumentUploadForm object to reduce the logic in the view. Probably not worth the other costs mentioned above unless they can be resolved cleanly.

spec/controllers/documents/w2s_controller_spec.rb
```diff
@@ -118,11 +118,11 @@
         }
       end
 
-      it "redirects back to the W2s page" do
+      it "renders back the W2s page" do
         expect do
           post :update, params: params
         end.not_to raise_error
-        expect(response).to redirect_to w2s_documents_path
+        expect(response).to render_template(:edit)
```
jonsgreen, line 125: This was a good indication of being at least somewhat on the right track with the refactoring since we are now at least rendering the edit template with errors rather than redirecting.

jonsgreen:
Closing this PR and moving the story to the computer gardening section of icebox perhaps to be revisited at some future time when appropriate. Hopefully the comments contained will give some indication of the pain points run into so far with this incomplete refactoring to a form object to handle required document validation.

PR #198, opened 2020-04-29
Add partner logos to "Chat with us" page

jenny-heath:
This includes adding the vita partner model, which puts the code at a part-way point to the new method of routing

Co-authored-by: Ben Vandgrift <bvandgrift@codeforamerica.org>


db/migrate/20200429190910_add_partner_information_to_intake.rb
```diff
@@ -0,0 +1,15 @@
+class AddPartnerInformationToIntake < ActiveRecord::Migration[6.0]
+  def change
+    add_column :intakes, :vita_partner_name, :string
+
+    reversible do |dir|
```
jenny-heath, line 5: interesting - is this a backfill combined with a migration? never seen anything like this

app/models/intake.rb
```diff
@@ -396,6 +410,21 @@ def filing_years
...
 
+  def assign_vita_partner!
+    # NOTE: this MUST be called before create_intake_ticket is called.
+    return if vita_partner.present?
+    # given the information on the intake
+    get_or_create_zendesk_group_id
+    get_or_create_zendesk_instance_domain
+    if zendesk_group_id
+      # this will not be true for intakes to the UWTSA instance
+      # TODO: log this!
```
jenny-heath, line 421: what are we logging?

app/models/vita_partner.rb
```diff
@@ -0,0 +1,18 @@
+# == Schema Information
+#
+# Table name: vita_partners
+#
+#  id                      :bigint           not null, primary key
+#  display_name            :string
+#  drop_off_code           :string
+#  logo_url                :string
+#  name                    :string           not null
+#  referral_code           :string
```
bengolder, line 10: might be helpful to rename this to `source_parameter_prefix` to make it consistent with other language throughout the code.

db/migrate/20200428220200_create_vita_partners.rb
```diff
@@ -0,0 +1,17 @@
+class CreateVitaPartners < ActiveRecord::Migration[6.0]
+  def change
+    create_table :vita_partners do |t|
+      t.string :name, null: false
+      t.string :display_name
+      t.string :logo_url
```
bengolder, line 6: might make sense to call this `logo_filename` or `logo_path` if it's just storing the name of the image file. I would generally expect a _url property to be a full URL

PR #201, opened 2020-04-29
Fixes places where we display 2018 to show the year prior to the most recent filing year

jenny-heath:


app/controllers/documents/prior_tax_returns_controller.rb
```diff
@@ -5,7 +5,7 @@ def self.show?(intake)
     end
 
     def self.document_type
-      "2018 Tax Return"
+      "Prior Year Tax Return"
```
bengolder, line 8: Changing document type strings may break things. We should either accompany this with a data migration or separate out the type key from the display name.

bengolder: In this case, it's an improvement, so I would lean towards a data migration.

PR #202, opened 2020-04-30
Rescue job errors and send explicitly to Sentry with intake_ticket_id

jonsgreen:
- Create with_raven_context ApplicationJob helper for catching and
sending errors to Sentry with extra context
- Use shared examples for testing background jobs with errors in specs

(#172583987) Improve Sentry error reporting context to include zendesk
ticket ID

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,2 +1,8 @@
 class ApplicationJob < ActiveJob::Base
+
+  def with_raven_context(extra)
+    yield
+  rescue => exception
+    Raven.capture_exception(exception, extra: extra)
+  end
 end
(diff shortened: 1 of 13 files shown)


jonsgreen:
Note: I was a bit nervous about using the Raven.user_context not being completely sure how thread safe delayed job workers are. Tom was 90% sure it would be safe but this seemed like a safer approach and I kind of liked only bothering with Raven if an exception truly happened. Definitely open to feedback!

jonsgreen:
@bengolder: @bvandgrift and I are have a simple rake task brewing that will throw an error with extra details going though `with_raven_context`. We can just test that on demo and/or if we were given the sentry_dsn or the master.key then we could first test it from development. 

I also just want to make sure that you saw that there are some automated tests already in the PR. Raven is obviously mocked but I don't think you are asking me to actually write some sort of integration test that actually hits Sentry.

PR #208, opened 2020-05-01
Consolidated error notification #172616663

bvandgrift:
# Catching Zendesk Errors / Notifying Everything

Adds `ConsolidatedTraceHelper`, which will distribute error notifications and additional context to both the logs and to sentry. This is then integrated into the Zendesk Intake Ticket Job, notifying on unexpected behavior.

This also includes @jonsgreen 's work on `with_raven_context`, pulling it into a cohesive whole.

* Add spec for handling unexpected behavior when creating an intake ticket
* Notify and bail if a ticket requester id can't be created, preventing an error due to a lack of requester when creating a ticket. this enqueues the ticket creation job, and sets a time delay on the current job after 1 minute.
* adds similar guards to intake pdf attachment
* adds similar guards to Document Attachment and other jobs requiring a ticket

## TODO

* ~Add similar guards to Pdf Attachment~
* ~Add similar guards to Document Attachment~
* ~Possibly move notification into the controllers~, instead of the jobs (decided against)

PR #210, opened 2020-05-02
Use Raven.extra_context api to add details to exception

jonsgreen:
- Adds FailingJob for testing purposes

(#172583987) Improve Sentry error reporting context to include zendesk
ticket ID

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,8 +1,10 @@
 class ApplicationJob < ActiveJob::Base
 
-  def with_raven_context(extra)
+  def with_raven_context(extra_context={})
     yield
   rescue => exception
-    Raven.capture_exception(exception, extra: extra)
+    Raven.extra_context(extra_context)
+    raise exception
   end
+
 end
--- a/app/jobs/failing_job.rb
+++ b/app/jobs/failing_job.rb
@@ -1,0 +1,9 @@
+class FailingJob < ApplicationJob
+  queue_as :default
+
+  def perform(test_data)
+    with_raven_context(test_data: test_data) do
+      raise "forced test exception with context"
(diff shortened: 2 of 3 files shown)


app/jobs/failing_job.rb
```diff
@@ -0,0 +1,9 @@
+class FailingJob < ApplicationJob
+  queue_as :default
+
+  def perform(test_data)
+    with_raven_context(test_data: test_data) do
+      raise "forced test exception with context"
+    end
+  end
+end
```
jonsgreen, line 9: I am not sure if we still need to include this but it was helpful to test on staging. Let me know what you think?

app/jobs/application_job.rb
```diff
@@ -1,8 +1,10 @@
 class ApplicationJob < ActiveJob::Base
 
-  def with_raven_context(extra)
+  def with_raven_context(extra_context={})
     yield
   rescue => exception
-    Raven.capture_exception(exception, extra: extra)
+    Raven.extra_context(extra_context)
```
jonsgreen, line 6: So I think this is likely the favored API for adding context. I was a bit concerned about thread safety but was relieved to see that DelayedJob clears the Context after each exception: https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/delayed_job.rb#L45. Also since we are only adding the context when we know there is an exception makes the risk even smaller.

jonsgreen:
The result of a forced failed error can be reviewed [here](https://sentry.io/organizations/codeforamerica/issues/1642436118/?project=1848308&query=is%3Aunresolved)

jonsgreen:
This will be replaced by work done in https://github.com/codeforamerica/vita-min/pull/208

PR #211, opened 2020-05-02
At capacity page #172511461

jonsgreen:


spec/controllers/public_pages_controller_spec.rb
```diff
@@ -83,33 +83,4 @@
...
-
-    it 'renders when not at capacity (in at_capacity mode)' do
-      # TODO: should this even render when not at capacity, or should
-      # it redirect to somewhere else?
-
-      get :at_capacity
-
-      expect(response).to be_successful
-    end
-  end
```
kayline: Not clear on the reasons for deleting these tests. Do we have any other tests that check that these pages exist and can be rendered by the controller? Non-blocking question.

PR #213, opened 2020-05-05
Adds RequestedDocumentsLater to list of required document types

jonsgreen:
- (#172660949) Require at least one requested document


jonsgreen:
Note: using the shared view code for required documents changes the text of the button/links to just be "Continue". Hopefully that makes sense enough in this context but otherwise we will need to make some changes to accommodate different text.

app/lib/bank_details_pdf.rb
```diff
@@ -1,3 +1,5 @@
+require 'mini_magick'
+
```
bengolder, line 2: This looks unrelated, why was it included?

jonsgreen: Yes, it is unrelated. My test failed so I think it must be a failure that only happens when run in a certain order. Generally the suite passes. I probably should remove it and maybe we should research this separately. I am curious if that spec fails for you if you run it individually .

bengolder: Yes tests are passing for me

PR #215, opened 2020-05-05
Allow vita partner association to be overwritten with assign_vita_partner!

jenny-heath:
- users should be able to go back and change their state

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -435,11 +435,13 @@ class Intake < ApplicationRecord
   end
 
   def assign_vita_partner!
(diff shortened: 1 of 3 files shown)


bengolder:
upon reading this, I noticed some things we would want to test for. The test should be included as part of the unit tests for the PersonalInfoController or the PersonalInfoForm. 

- We should test that submitting with an existing intake that does not have a ZD ticket ID on the personal info page assigns a different partner.
- We should test that submitting with an existing intake that already has a ZD ticket ID retains the existing partner assignment.

PR #216, opened 2020-05-05
adds partner source codes and states

bvandgrift:
this PR implements the decisions from [pivotal story #172323019](https://www.pivotaltracker.com/story/show/172323019), captured in the source tree in [doc/arch/2020-05-07-refactor-zendesk-routing-logic.md](https://github.com/codeforamerica/vita-min/blob/add-partner-codes-and-states/doc/arch/2020-05-07-refactor-zendesk-routing-logic.md)

it adds structure for adding new vita partners, SourceParameter (source codes), State, and their relationship to VitaPartners, cleans up some testing, provisions from YML files, and adds data capture during the partner assignment process. for all the delicious details, read the linked above.

additionally, the source parameter codes (referral codes) now match exactly, rather than doing a fuzzy match on the front of the string. [rationale and decision here](https://github.com/codeforamerica/vita-min/blob/add-partner-codes-and-states/doc/2020-05-12-exact-matching-on-source-parameter.md).

at this time, it's ready for a review. it could still use a peek into the production console to ensure that no source parameters are in use that aren't accounted for in the YML file, and the YML files checked for correctness by a few more eyes.

For production console, something like:

```ruby
Intake.pluck(:source).uniq 
```

...would provide a list of sources submitted by clients.


lib/tasks/upsert_vita_partners.rake
```diff
@@ -1,32 +1,9 @@
...
-    puts "beginning partner upsert using environment: #{Rails.env}"
-    # TODO: don't leave the next line in
-    # load yaml file
-    partners = YAML.load_file('db/vita_partners.yml')['vita_partners']
-    partners.each do |datum|
-      partner = VitaPartner.find_by(zendesk_group_id: datum['zendesk_group_id'])
-      partner.present? ? update_partner(partner, datum) : create_partner(datum)
-    end
+    include VitaPartnerImporter
+    upsert_vita_partners
```
tdooner, line 7: Is there a reason to not make these classes (so you could just call `VitaPartnerImporter.upsert_vita_partners`)? I think mixing in the methods makes me a bit less comfortable that I understand the full effect of what methods are defined in which scopes.

app/lib/state_importer.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def upsert_states(yml = STATE_YAML)
+    say "beginning state upsert using environment: #{Rails.env}"
+    states = YAML.load_file(yml)['states']
+    State.destroy_all
+    states.each do |datum|
+      state = State.find_by(abbreviation: datum['abbreviation'])
+      state.present? ? State.update!(name: datum['name']) : State.create!(abbreviation: datum['abbreviation'], name: datum['name'])
+    end
```
tdooner, line 21: Since this will be run on live data as part of the deploy, it should be in a transaction so that there is no possibility of an inconsistent state during the import (and especially if the import fails for some reason).

```ruby
State.transaction do
  State.destroy_all
  states.each do ... end
end
```

app/lib/vita_partner_importer.rb
```diff
@@ -0,0 +1,61 @@
...
+    partners = YAML.load_file(yml)['vita_partners']
+    partners.each do |datum|
+      states = datum.delete('states')
+      sources = datum.delete('source_parameters')
+      partner = VitaPartner.find_by(zendesk_group_id: datum['zendesk_group_id'])
+      partner = partner.present? ? update_partner(partner, datum) : create_partner(datum)
+      refresh_partner_sources(partner, sources)
+      refresh_partner_states(partner, states)
+      say "  => done"
+    end
```
tdooner, line 26: Transaction here would be ideal as well.

db/vita_partners.yml
```diff
@@ -51,71 +78,95 @@ vita_partners:
...
     zendesk_instance_domain: eitc
     zendesk_group_id: "360009267673"
     display_name: United Way of Greater Richmond and Petersburg
-    source_parameter: uwvp
-    logo_path: "partner-logos/uwvp.jpg"
+    source_parameters:
+      - uwvp
+    states:
+      - VA
+    logo_path: ""
```
tdooner, line 118: Looks like a bunch of these logo_path's got deleted in the restructuring

app/models/intake.rb
```diff
@@ -511,19 +511,20 @@ def determine_zendesk_instance_domain
 
   private
 
+  ##
+  # at load time, maps the source codes to the appropriate associated
+  # group id for speedy lookup
+  SOURCE_MATCHERS = Hash[SourceCode.all.map { |s| [s.code, s.vita_partner.zendesk_group_id] }]
```
tdooner, line 517: Caching this in a constant is a bit of an unnecessary optimization, IMHO, because of how infrequently we actually invoke the routing logic. That said, I think it's unlikely to cause confusion because the constant would be reloaded properly in development.

Also: We could implement longest prefix matching by sorting the SourceCode query by the code's length. (E.g. route `RefundDay-C` to Catalyst despite the `RefundDay` source code also existing.)

db/vita_partners.yml
```diff
@@ -51,71 +78,95 @@ vita_partners:
...
+      - RefundDay-C
     logo_path: ""
   - name: RefundDay
     zendesk_instance_domain: eitc
     zendesk_group_id: "360009657014"
     display_name: RefundDay
-    source_parameter:
-    logo_path: "partner-logos/refund-day.png"
+    source_parameters:
+      - RefundDay
```
tdooner, line 162: Yeah, this is a prefix of the other RefundDay sources, so I would want to have a bit more confidence that the routing logic properly works for the other ones if you're going to add this one as part of this branch.

app/lib/state_importer.rb
jenny-heath: is it an "upsert" if you destroy_all?

app/lib/vita_partner_importer.rb
```diff
@@ -0,0 +1,61 @@
...
+  def upsert_vita_partners(yml = VITA_PARTNERS_YAML)
+    say "beginning partner upsert using environment: #{Rails.env}"
+
+    partners = YAML.load_file(yml)['vita_partners']
+    partners.each do |datum|
+      states = datum.delete('states')
+      sources = datum.delete('source_parameters')
+      partner = VitaPartner.find_by(zendesk_group_id: datum['zendesk_group_id'])
+      partner = partner.present? ? update_partner(partner, datum) : create_partner(datum)
+      refresh_partner_sources(partner, sources)
```
jenny-heath, line 23: bit confused about the difference between source_parameters and source_codes here...

app/models/intake.rb
```diff
@@ -513,19 +513,20 @@ def determine_zendesk_instance_domain
...
+    code_match&.first&.last
   end
 
   def group_id_for_state
-    EitcZendeskInstance::GROUP_ID_TO_STATE_LIST_MAPPING.each do |group_id, state_list|
-      return group_id if state_list.include? state_of_residence
-    end
+    state = State.find_by(abbreviation: state_of_residence&.upcase || state&.upcase)
+    return state.vita_partners.first.zendesk_group_id if state.present? && !state.vita_partners.empty?
+    # TODO: 'load-balancing'
```
jenny-heath, line 529: is that part of this story? if not i don't think we need this comment

spec/models/intake_spec.rb
```diff
@@ -543,7 +543,19 @@
   end
 
   describe "#mixpanel_data" do
-    let!(:vita_partner) { create :vita_partner, name: "test_partner", zendesk_group_id: EitcZendeskInstance::ONLINE_INTAKE_UW_TSA }
+    let(:state_of_residence) { 'CA' }
+    let(:state) { State.find_by!(abbreviation: state_of_residence) }
+    let(:vita_partner) do
+      partner = state.vita_partners.first
+      return partner if partner.present?
```
jenny-heath, line 550: is this check here because we don't know what models will exist or be associated in the test environment? not sure when it will return here and when it will go on to create the vita partner

spec/rails_helper.rb
```diff
@@ -72,6 +72,18 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+  config.before(:suite) do
+    # we include this because at present (2020-05-06) Vita Partners
+    # are yaml-loaded read-only domain information, as are states.
+    # TODO: when Vita Partners are editable, this should be removed
+    # and fixed tests adjusted
+    extend StateImporter
```
jenny-heath, line 80: i guess Tom's above comment about doing this instead of `StateImporter.upsert_states` also applies here

spec/rails_helper.rb
```diff
@@ -121,3 +133,10 @@ def omniauth_idme_success
     }
   })
 end
+
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
+end
```
jenny-heath, line 142: not a clue what this does

db/vita_partners.yml
```diff
@@ -3,119 +3,170 @@ vita_partners:
...
+    states:
+      - CA
+      - AK
   - name: Tax Help Colorado (Piton Foundation)
     zendesk_instance_domain: eitc
     zendesk_group_id: "360009415854"
     display_name: Tax Help Colorado
-    source_parameter: thc
-    logo_path: ""
+    source_parameters:
```
jenny-heath, line 16: should this be `source_codes`?

lib/tasks/upsert_vita_partners.rake
bvandgrift: There is! Since the class doesn't encapsulate any information, there's no need for the class overhead.

app/lib/state_importer.rb
bvandgrift: +1 very true.

app/lib/vita_partner_importer.rb
bvandgrift: Yup.

app/models/intake.rb
bvandgrift: That's a solid idea. I wonder, however, if that wouldn't inadvertently obscure shorter codes that had been introduced ill-advisedly. Thinking out loud -- partner A's code is `orange`, partner B's is `ora`. partner B doesn't know about partner A's code, so they issue a referral-code `orange-juice`, counting on the `ora` part to redirect the client to them. In sorting by the longest first, client is misdirected to partner A instead of B. Sorting by the shortest code first, the opposite problem happens every time partner A's code comes in.

So sorting by longest has the least problems by default, but it doesn't solve for some cases. I think the rule here should be that a partner source code can't be a subset of another partner's code.

* partner A: united-way-

app/lib/intake_instance_domain_and_group_id_backfill.rb
```diff
@@ -12,7 +12,7 @@ def self.backfill_instance_domains!
   end
 
   def self.backfill_group_ids!
-    Intake.where(zendesk_group_id: nil).each do |intake|
+    Intake.where(vita_partner_group_id: nil).each do |intake|
       populate_group_id(intake)
     end
     nil
```
bengolder, line 18: I don't believe we are using this file for anything. These changes are harmless, but we can also delete this file.

lib/tasks/terraform.rake
```diff
@@ -0,0 +1,6 @@
+namespace :db do
+  desc 'tear down and rebuild the database'
+  task terraform: [:drop, :create, :migrate, :upsert_vita_partners] do
+    puts "database rebuilt!"
```
kayline, line 4: Does this also need to insert states now?

app/models/intake.rb
```diff
@@ -439,34 +442,25 @@ def year_before_most_recent_filing_year
...
-      raise "unable to determine VITA Partner from zendesk group id: [#{group_id}]" unless partner.present?
-      update(vita_partner_id: partner.id, vita_partner_name: partner.name)
-    end
-  end
+    # we only check whether it has been persisted, and do not save it
+    return nil if zendesk_instance == UwtsaZendeskInstance
 
-  def get_or_create_zendesk_group_id
-    return zendesk_group_id if zendesk_group_id.present?
+    partner, routing_criteria, routing_value = partner_for_source || partner_for_state || partner_for_overflow
```
bengolder, line 452: Exciting to see that we are getting all these values back! It might be easier to maintain or work on these in the future with a more explicit output, such as a hash or a minimal "PartnerRoutingResult" class.

app/models/intake.rb
```diff
@@ -439,34 +442,25 @@ def year_before_most_recent_filing_year
...
 
-    group_id = determine_zendesk_group_id
-    self.update(zendesk_group_id: group_id)
-    group_id
-  end
+    raise "partner not found!" unless partner # this shouldn't happen unless the data is horked!
```
bengolder, line 454: love that it would blow up here without a result!

app/models/intake.rb
bengolder: Non-blocking, but given that we only want two types of results from these methods (`nil` or `Array` of three specific things) think it would be helpful to at least comment if not enforce with a class

db/migrate/20200505214308_create_source_parameters.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSourceParameters < ActiveRecord::Migration[6.0]
+  def change
+    create_table :source_parameters do |t|
+      t.string :code
```
bengolder, line 4: might be good to enforce uniqueness on the code column

app/models/intake.rb
```diff
@@ -519,20 +515,28 @@ def current_ticket_status
...
+
+    [partner, "source_parameter", source]
   end
 
-  def group_id_for_state
-    EitcZendeskInstance::GROUP_ID_TO_STATE_LIST_MAPPING.each do |group_id, state_list|
-      return group_id if state_list.include? state_of_residence
-    end
+  def partner_for_state
+    state = State.find_by(abbreviation: state_of_residence&.upcase || state&.upcase)
```
bengolder, line 528: having two different cases (uppercase for state models, lowercase for intake records) I think bring some risk of bugs in the future. I think we should be consistent across all the models, but don't have an opinion on which one we use. Should we convert everything to uppercase, convert states to lowercase, or risk it?

app/models/intake.rb
jonsgreen: A light weight Struct can also be a nice in between option until a full blown class with logic is needed.

db/vita_partners.yml
```diff
@@ -3,131 +3,184 @@ vita_partners:
...
+      - WA
     logo_path: ""
   - name: Campaign for Working Families
     zendesk_instance_domain: eitc
     zendesk_group_id: "360009220253"
     display_name: Campaign for Working Families
-    source_parameter: cwf
-    logo_path: "partner-logos/cwf.png"
+    source_parameters:
+      - cwf
```
tdooner, line 49: also add: cwf-mfp

lib/tasks/terraform.rake
bvandgrift: it does.

app/models/intake.rb
bvandgrift: yes. i've removed the `STATE_OPTIONS` structure from `app/helpers/states.rb`, replacing with a dynamically generated list (without messing with the rest of the helpers). also, a backfill rake task `backfill:upcase_states` has been added to upcase all state field references on `Intake`.

bengolder: Would we need to run this backfill during deploy? Is it already set up to run automatically when we deploy this feature? Is there a risk if we run it manually after we deploy?

app/models/intake.rb
```diff
@@ -439,34 +442,26 @@ def year_before_most_recent_filing_year
...
-      update(vita_partner_id: partner.id, vita_partner_name: partner.name)
-    end
-  end
+    # we only check whether it has been persisted, and do not save it
+    return nil if zendesk_instance == UwtsaZendeskInstance
 
-  def get_or_create_zendesk_group_id
-    return zendesk_group_id if zendesk_group_id.present?
+    # this is a RouteOptions struct, defined below
+    route_options = partner_for_source || partner_for_state || partner_for_overflow
```
bengolder, line 453: Totally non-blocking, but "options" implies to me that this is sn object used for configuring something else or something that offers a choice between different things. It seems like RoutingResult or a name that implies it is the output of routing logic might be clearer.

app/models/source_parameter.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+#  fk_rails_...  (vita_partner_id => vita_partners.id)
+#
+class SourceParameter < ApplicationRecord
+  belongs_to :vita_partner
+
+  validates_presence_of :vita_partner_id
+  validates_presence_of :code
+  validates_uniqueness_of :code
+end
```
bengolder, line 26: This looks great

app/models/intake.rb
bvandgrift: good suggestion, and one to be done after we get the rest of it merged in.

PR #219, opened 2020-05-06
[SPIKE removing anonymous intakes] Excludes anonymous intakes from current_intake except req docs flow

jonsgreen:
- Adds a feature spec for current bug scenario of user trying to access
main intake flow from incomplete requested docs flow

(#172627601) Spike on improving how we handle Anonymous Intakes for
returning sessions


bengolder:
I converted this to a "draft". Recently saw this feature after Ben V used it. And it seems perfect for a use case like this! Let's use draft PRs more often for WIP/do-not-merge situations.

PR #220, opened 2020-05-06
Deploy on git tag #170722758

bvandgrift:
## Goals

Story: https://www.pivotaltracker.com/n/projects/2409240/stories/170722758

95% complete: at current configuration, when this branch is merged into master (it has a tag), and master pushed, CI should deploy the branch to `staging`, using the `deploy_to_aptible--production` job. If this is successful, replace the staging git remote in that task (in `.circleci/config.yml`) with the production git remote.

Documentation is linked from README, and alternatively [here](https://github.com/codeforamerica/vita-min/blob/deploy-on-git-tag-%23170722758/doc/environments-and-deployment.md)

Placeholders in that doc could be filled in.


bengolder:
One question, does it make a difference that the tag format is `version-X.X.X` vs. `vX.X.X`?
The [examples I see in github](https://developer.github.com/v3/repos/releases/#input) typically just show the lowercase `v` and it looks like this PR looks for a `version-`.

bvandgrift:
@bengolder since we're using a regex match, `/v*/ feels riskier to me than `/version-/`. so, my gut said go with the longer thing.

PR #222, opened 2020-05-06
Only link sms messages to open ZD tickets

kayline:
- attach a comment to the sms ticket if there are no open related tickets

### DIFF
--- a/spec/services/zendesk_sms_service_spec.rb
+++ b/spec/services/zendesk_sms_service_spec.rb
@@ -47,10 +47,10 @@ describe ZendeskSmsService do
     end
 
     context "when there are users, drop offs and Zendesk tickets associated with this phone number" do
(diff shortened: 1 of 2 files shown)


spec/services/zendesk_sms_service_spec.rb
```diff
@@ -108,7 +108,7 @@
           body here
         BODY
 
-        ["1", "2", "3", "4"].map do |ticket_id|
+        [1, 2, 3, 4].map do |ticket_id|
```
bvandgrift, line 111: nitpick: here and on line 185, prefer `each` to `map`, as `map` implies a transformation in order to return a new array, whereas `each` returns the original enumerable. it doesn't really matter, but 'each' makes clear that no transform is happening, just iteration.

bengolder:
I'm going to go ahead and merge, because I would love to have these errors cleared tomorrow morning. @kayline Feel free to add another commit if there is anything you'd like to change based on comments.

PR #224, opened 2020-05-07
Update copy and styles on successfully submitted page

jenny-heath:
Screen Shot 2020-05-07 at 4 36 02 PM

without intake (if user refreshes page)
Screen Shot 2020-05-07 at 4 34 04 PM

mobile:
Screen Shot 2020-05-07 at 4 37 20 PM


bengolder:
@jenny-heath when you have time, could you add a screenshot to this PR or the pivotal tracker ticket?

bengolder:
@jenny-heath I think we should add an expectation to the feature specs that the final page includes "Your confirmation number is"

jenny-heath:
@bengolder it turned out to be a bit complicated to add that line to the feature specs because `intake_ticket_id` is assigned in a delayed job so it doesn't exist when this page is rendered. i got it to work by adding this line:
`allow_any_instance_of(Intake).to receive(:intake_ticket_id).and_return(9876)`
but thought that was a little janky and cluttered up the test a bit so i've left it out for now. let me know if you think this is an important test though and we can discuss / add something later.

PR #230, opened 2020-05-08
Adds file with help with options for moving forward or back to start

jonsgreen:
- adds button-group class style for spacing multiple buttons

(#172725759) Add "filing with help" signpost page


app/views/questions/file_with_help/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      This is the right option for you if:
+    </p>
+
+    <ul class="list--checkmark">
+      <li>You need assistance with tax filing</li>
+      <li>Your tax situation is confusing/complicated</li>
+      <li>You’re nervous to file completely on your own</li>
+    </ul>
+    <div class="button--group">
+      <%= link_to feelings_questions_path, class: "button text--centered" do %>
```
jonsgreen, line 23: I was not sure whether to hard code this to the previous page or to literally link back in case this page will be accessed from multiple pages. This keeps people form going to the wrong place if they access this page using the url. As it is now it will need to be changed if we change the flow to this page.

app/views/questions/file_with_help/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <p>
+      You selected our full service option. We’ll have you answer a series of questions about your tax situation (to the best of your ability) and have a tax specialist review everything with you.
+    </p>
+
+    <p>
+    <strong>This is the right option for you if:</strong>
+    </p>
+
+    <ul class="checkmark-list">
+      <li>
```
jonsgreen, line 18: This could be made into some sort of partial or helper but in the end I did not think that much would be gained.

app/views/questions/file_with_help/edit.html.erb
bengolder: I think a `link_to(:back) do` would likely make sense here, essentially just acting close to a browser back button. I've noticed in some future mocks that people may arrive to this page from more than one page.

app/views/questions/file_with_help/edit.html.erb
bengolder: we may be able to use other examples of a list of items with an icon for each

jonsgreen: @bengolder Are you suggesting this as a future refactoring when we need it or something to prepare for now?

bengolder: @jonsgreen non-blocking

PR #231, opened 2020-05-11
appends '(Fake User)' to users created/searched in demo and staging

bvandgrift:
PT inception: https://www.pivotaltracker.com/story/show/172570012

- list of environments in ZendeskServiceHelper
- specs added for support function and environments
- helper method `with_environment` added to prevent duplication

### DIFF
--- a/app/helpers/zendesk_service_helper.rb
+++ b/app/helpers/zendesk_service_helper.rb
@@ -63,7 +63,7 @@ module ZendeskServiceHelper
     end
 
     search_string = ""
-    search_string += "name:\"#{name}\" " if name.present?
+    search_string += "name:\"#{qualify_user_name(name)}\" " if name.present?
     search_string += "phone:#{phone}" if phone.present?
     results = search_zendesk_users(search_string)
     if exact_match
@@ -84,7 +84,7 @@ module ZendeskServiceHelper
   def create_end_user(name:, **attributes)
     # for a list of possible valid attributes, see:
     #   https://developer.zendesk.com/rest_api/docs/support/users#json-format-for-end-user-requests
-    client.users.create(name: name, verified: true, **attributes)
+    client.users.create(name: qualify_user_name(name), verified: true, **attributes)
   end
 
   def find_or_create_end_user(name, email, phone, exact_match: false, time_zone: nil)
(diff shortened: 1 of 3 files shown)


app/helpers/zendesk_service_helper.rb
```diff
@@ -63,7 +63,7 @@ def find_end_user(name, email, phone, exact_match: false)
     end
 
     search_string = ""
-    search_string += "name:\"#{name}\" " if name.present?
+    search_string += "name:\"#{qualify_user_name(name)}\" " if name.present?
```
bengolder, line 66: I don't think we should include the "(Fake User)" string in the search. But open to being convinced otherwise.

app/helpers/zendesk_service_helper.rb
```diff
@@ -98,6 +98,14 @@ def find_or_create_end_user(name, email, phone, exact_match: false, time_zone: n
...
   end
 
+  def qualified_environments
+    %w[development demo staging]
+  end
+
+  def qualify_user_name(name)
+    qualified_environments.include?(Rails.env) ? "#{name} (Fake User)" : name
+  end
+
```
bengolder, line 108: non-blocking but I wonder if there is a more specific name we could use for these methods. I wouldn't know what `qualify` means without having read the ticket. For example: `zendesk_testing_environments`, or `tagged_user_name` or `scoped_user_name` or even `zendesk_requester_name`?

spec/support/with_environment.rb
```diff
@@ -0,0 +1,11 @@
...
+# temporarily changes the Rails environment
+def with_environment(env_name)
+  prev = Rails.env
+  Rails.env = env_name
+
+  yield
+
+ensure
+  Rails.env = prev
+end
```
bengolder, line 11: Would this differ in behavior from the approach we currently use?
```
allow(Rails).to receive(:env).and_return("production".inquiry)
```
We probably want to convert to a string and then inquiry to ensure that existing methods of checking environment wouldn't fail, such as `Rails.env.production?`

app/helpers/zendesk_service_helper.rb
bvandgrift: I suppose the underlying question is whether or not Zendesk quoted name searches will find a partial match. If they can, that may introduce other problems ("john smith" vs "john smithfield"). I was assuming that the quotes in line 66 meant we wanted something like an exact match. If that's the case, then having the (Fake User) string present is required to match at all, since it's in the name of the Zendesk user on their end.

spec/support/with_environment.rb
bvandgrift: Using `allow` stubs a specific function for the duration of the subsequent tests. It doesn't feel thorough to me. I don't have any confidence that invisible dependencies are being met.

By setting the .env directly, it changes the environment (and brings along all the rest of the functionality). So setting .env to 'foobar' means that Rails.env.foobar? works as expected. Since at any given time, I'm not certain what underlying structure is depending on the `env`, I preferred to set it at the top level so no discrepancies would pop up.

PR #236, opened 2020-05-12
Add United Way Capital Region (PA) and Greenville County

tdooner:
Adding them only based on source parameter at first to allow them to
test.


bengolder:
@tdooner have you checked that the logos look okay in context, on the chat-with-us page?

app/helpers/eitc_zendesk_instance.rb
```diff
@@ -58,6 +60,8 @@ class EitcZendeskInstance
     "refundday-h" => ONLINE_INTAKE_HU_FL,
     hispanicunity: ONLINE_INTAKE_HU_FL,
     uwfm: ONLINE_INTAKE_UW_FRESNO_MADERA,
+    uwcrpa: ONLINE_INTAKE_UWCRPA,
+    uwgc: ONLINE_INTAKE_GREENVILLE,
```
bengolder, line 64: Let's use the following variable names for consistency:
```
ONLINE_INTAKE_UW_CRPA,
ONLINE_INTAKE_UW_GREENVILLE,
``

tdooner: Will do!

PR #238, opened 2020-05-13
Adjusts styling of two buttons for desktop

jonsgreen:
Signed-off-by: Ben Golder <bgolder@codeforamerica.org>


jonsgreen:
We ended up removing the second button so this is not needed.

PR #245, opened 2020-05-14
Adds Already Filed page after File With Help

jonsgreen:
- Adds message to ticket body when already_filed is "yes"

(#172399045) Add "have you already filed?" question


app/views/questions/already_filed/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<% content_for :form_question, "Before we start, have you already filed for 2019?" %>
+<% content_for :form_help_text, "We’re asking because you may have filed with another service or you may have an ammendment to your 2019 return." %>
```
jenny-heath, line 2: amendment has one "m"

PR #247, opened 2020-05-15
Filing for stimulus #172323097

jonsgreen:
This PR builds on top of https://github.com/codeforamerica/vita-min/tree/already-filed-page-%23172399045


app/controllers/questions/stimulus_recommendation_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Questions
```
jonsgreen, line 1: So this is perhaps being a bit lazy since this really does not use much of the QuestionsController since there really isn't a question or form on the page. I also considered making this an action in the PublicPagesController but that did not seem quite right either. It could be just its own stand alone controller inheriting from application controller using the default layout.

I did think that perhaps some special tracking might be desired at some point to see if clients are going to the FAQ page from here in which case it might make sense to make the button a form. Of course that could be easily tracked in other ways as well.

app/services/zendesk_intake_service.rb
```diff
@@ -164,7 +164,7 @@ def send_final_intake_pdf
       Online intake form submitted and ready for review. The taxpayer was notified that their information has been submitted. (automated_notification_submit_confirmation)
 
       Client's provided interview preferences: #{@intake.interview_timing_preference}
-
+      #{"Client is filing for Economic Impact Payment support" if @intake.filing_for_stimulus_yes?}
```
jonsgreen, line 167: This was my best guess for where the requirement in the story: "Should add a comment with the 13614-C "client is filing for Economic Impact Payment support". It may very well be that something else was intended?

app/controllers/questions/filing_for_stimulus_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def next_path
+      if current_intake.filing_for_stimulus_yes? && current_intake.already_filed_yes?
+        stimulus_recommendation_path
+      else
+        super
+      end
+    end
+
+    def track_question_answer
```
jonsgreen, line 13: This seemed like the easiest way to fill Lou's suggestion: "we should log responses to this in mixpanel as a custom event so we can easily report on what % of people are electing stimulus". If it is important to have the usual "question_answered" event as well then I could just call super above.

app/controllers/questions/stimulus_recommendation_controller.rb
jenny-heath: we talked about it and decided to put it in public pages because that's the pattern we use for the eligibility offboarding path

app/controllers/questions/filing_for_stimulus_controller.rb
jenny-heath: `track_question_answer` is probably sufficient for this

app/services/zendesk_intake_service.rb
jenny-heath: let's add it to the first place where we upload the 13614c also

PR #248, opened 2020-05-15
Set up datadog APM tracer in staging for testing in the Aptible envir…

loumoore:
…onment


loumoore:
I set up a new Aptible app in staging only that's running the Datadog agent in a container. 

I also tested that this Datadog configuration works locally in dev environment (I disabled it by default assuming not everyone will want to run the DD agent locally, but you could try it out by editing the initializer to enable it in development).

So the purpose of this PR is to allow me to verify that this is working in staging. The initializer only enables the datadog tracer in staging so it should be a no-op in demo and prod. Once this has gone to staging and I've verified it's working there, I will set up Datadog agents in demo and prod and turn it on everywhere.

To start, this will give us APM monitoring, which means we'll get performance traces on pages and be able to see where we have http errors, slow queries, slow pages, etc. But it is also just a first step in getting the agent set up in aptible. From here we can add more types of metric collection with Dogstatsd for custom metrics and we can even point logs at it to try to get everything in one place in Datadog.

PR #250, opened 2020-05-15
Add Zendesk CLI for adding partners/users

tdooner:
An automated process for adding Zendesk partners and users will help us:
1) add new partners more efficiently
2) ensure consistency in Zendesk configuration, preventing configuration
   mistakes
3) quickly handle updates for partners

This adds a quick command line interface using Thor (similar to GCF's
various command line utils). This is a work in progress. It implements
the correct logic in Zendesk related to the groups that the users should
be in, for simple cases (no sub-sites/non-coalition), and we will need
more complicated logic soon. There may also be a couple small bugs in
the script since it is not ideal to run against production Zendesk and
there is not a sandbox option yet.

Usage:

Create a partner's configuration on Zendesk:
```bash
bin/zendesk import_partner
```

Create Zendesk users for a partner:
```bash
bin/zendesk import_users --partner "United Way of Greater Newark" --csv_path ~/Downloads/uwgn_users.csv
```


tdooner:
I rewrote it a bit today to process the spreadsheet line by line in a big chunk of unbroken logic. Although it's a bit more daunting that way, the control flow is incredibly clear and it's very easy to reason about what's happening.

bengolder:
@tdooner it wasn't clear if this was ready for review, based on the branch title and "This is a work in progress". Can you convert it from a draft when you feel like the code is ready, & sufficiently tested?

tdooner:
@bengolder - Check it out now. I wrote a bunch of tests for `ImportUsers` since it is high importance to verify behavior. I didn't write any tests for the `ImportPartner` command since it doesn't do any access control. Let me know if you see any issues that would block merge!

PR #252, opened 2020-05-15
Security fixes

loumoore:


app/views/public_pages/stimulus.html.erb
```diff
@@ -20,7 +20,7 @@
           broken down some common questions to help you out.
         </p>
         <p class="hint last">
-          The IRS will post all key information on <%=link_to 'https://irs.gov/eip', target: '_blank', "data-track-click": "stimulus-irs-eip", "data-track-attribute-position": "intro" do %>IRS.gov/eip<% end %>
+          The IRS will post all key information on <%=link_to 'https://irs.gov/eip', target: '_blank', rel: "noopener", "data-track-click": "stimulus-irs-eip", "data-track-attribute-position": "intro" do %>IRS.gov/eip<% end %>
```
bengolder, line 23: There is some risk that this might block click tracking. Lou, have you already checked?

loumoore: i will verify before we merge. 

noopener is different from noreferrer, it should only prevent the opened window from having access to the opening window. but could call to verify this.

loumoore: yup still works

loumoore:
after looking this over one more time, i'm reverting the subresources integrity changes for now. it seems possible for these resources to change when google/zendesk update them without there being anything malicious going on, and the likelihood of them being hacked and having their javascript replaced seems low enough we can wait and make sure this is a good idea.

PR #256, opened 2020-05-18
Filing for stimulus #172323097

jonsgreen:


app/lib/question_navigation.rb
```diff
@@ -5,6 +5,7 @@ class QuestionNavigation
     Questions::FileWithHelpController,
     Questions::AlreadyFiledController,
     Questions::FilingForStimulusController,
+    Questions::StimulusRecommendationController, #possible off-boarding from flow
```
jenny-heath, line 8: nice note!

PR #263, opened 2020-05-20
add datadog metrics api module and instrument first metric

loumoore:
Added the module for reporting to the datadog api and a spec for it, plus just the most straightforward-to-name metric (zendesk.ticket.created) so we can validate it working while we finalize the naming convention


app/lib/datadog_api.rb
```diff
@@ -0,0 +1,48 @@
...
+    @configuration ||= Configuration.new
+  end
+
+  def self.configure
+    yield configuration
+  end
+
+  def self.client
+    @statsd ||= Dogapi::Client.new(configuration.api_key)
+  end
```
kayline, line 29: Since we are using the API client and not DogStatsD, I might change the name of the variable here. Not a big deal.

loumoore: oops you're right good call, will change

PR #264, opened 2020-05-20
Remove partner routing constants from EitcZendeskInstance

tdooner:
To prevent drift from our new source of truth (db/vita_partners.yml),
I've removed the constants from the EitcZendeskInstance file.

This is ready for review.


spec/models/intake_spec.rb
```diff
@@ -841,46 +841,47 @@
...
         end
       end
     end
 
     context "with state routing" do
-      shared_examples "state-level routing" do |state_criteria, zendesk_group_id, zendesk_instance|
+      shared_examples "state-level routing" do |state_criteria, partner_name|
         context "given a state" do
           let(:state) { state_criteria } # might not be necessary?
+          let(:partner) { VitaPartner.find_by!(name: partner_name) }
```
bvandgrift, line 854: Adding an exception here--especially given that it's a shared example set--is an excellent choice.

spec/models/intake_spec.rb
```diff
@@ -841,46 +841,47 @@
...
+      it_behaves_like "state-level routing", "TN", "Impact America (Save First) Mississippi River States", "eitc"
+      it_behaves_like "state-level routing", "AR", "Impact America (Save First) Mississippi River States", "eitc"
+      it_behaves_like "state-level routing", "MS", "Impact America (Save First) Mississippi River States", "eitc"
+      it_behaves_like "state-level routing", "NV", "Nevada Free Taxes Coalition", "eitc"
+      it_behaves_like "state-level routing", "TX", "Foundation Communities", "eitc"
+      it_behaves_like "state-level routing", "AZ", "United Way of Tuscon and Southern Arizona", "eitc"
+      it_behaves_like "state-level routing", "VA", "United Way of Greater Richmond and Petersburg", "eitc"
+      it_behaves_like "state-level routing", "FL", "RefundDay", "eitc"
+      it_behaves_like "state-level routing", "NM", "Tax Help New Mexico", "eitc"
+      it_behaves_like "state-level routing", "XX", "United Way of Tuscon and Southern Arizona", "eitc"
```
bvandgrift, line 884: This line will likely cause a problem when more overflow providers are added, but definitely not blocking.

PR #268, opened 2020-05-22
instrument zendesk custom metrics

loumoore:
Added instrumentation to start sending more custom metrics to zendesk using the naming convention outlined here (see final comment): https://www.pivotaltracker.com/story/show/172898761

Metrics we wanted to track are covered in these tickets:
https://www.pivotaltracker.com/story/show/172846685
https://www.pivotaltracker.com/story/show/172846520
https://www.pivotaltracker.com/story/show/172846528


spec/services/zendesk_intake_service_spec.rb
```diff
@@ -211,6 +218,13 @@
           data: intake.mixpanel_data.merge(ticket_status.mixpanel_data)
         )
       end
+
+      it "sends a datadog metric" do
```
jonsgreen, line 222: Perhaps a shared examples would DRY this up and also make it easier to add/test other jobs/services in the future?

loumoore: good call. i added a ticket to do that as a follow up.

PR #270, opened 2020-05-23
Zendesk Admin Portal ##4579468

bengolder:
For the Zendesk Admin MVP Epic: 
https://www.pivotaltracker.com/epic/show/4579468

Brief:
https://docs.google.com/document/d/1V432EftQ4x3nAbq8zhsLpUgg61Hmoz5V_qmplTOYkpc/edit


spec/support/shared_examples/a_protected_zendesk_ticket_page.rb
```diff
@@ -0,0 +1,40 @@
+shared_examples :a_protected_zendesk_ticket_page do |action: :show|
```
tdooner, line 1: interesting syntax there, I didn't realize you could have a block argument with a default value...

app/assets/stylesheets/components/_document-listing.scss
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+  &__details {
+    margin-left: 90px;
+  }
+  &__filename {
+    margin-bottom: 0;
+  }
+  &__upload-time {
+  }
```
tdooner, line 33: remove the empty CSS class?

app/helpers/zendesk_authentication_helper.rb
```diff
@@ -0,0 +1,25 @@
+module ZendeskAuthenticationHelper
```
tdooner, line 1: Seems fine to me, though. Wondering why you chose against the base class approach? These helpers are not reusable outside the controller environment anyways (since they rely on `send_data` method and `current_user`)

spec/support/shared_examples/a_protected_zendesk_ticket_page.rb
bengolder: Also worth noting: I used a symbol to define the shared example because I get frustrated when I can't easily navigate between shared examples and the specs that use them. By making it a symbol, RubyMine is able to quickly identify all usages of the shared example.

app/views/zendesk/tickets/show.html.erb
```diff
@@ -0,0 +1,84 @@
...
+            </h1>
+
+            <div class="document-overview-list">
+              <%= image_tag "paper.svg", alt: "", class: "document-overview-list__icon" %>
+              <h2 class="h3 document-overview-list__heading">13614-C Intake Form</h2>
+              <ul class="document-overview-list__filenames">
+                <% @intakes.each do |intake| %>
+                  <li class="document-listing">
+                    <%= link_to pdf_zendesk_intake_path(intake.id, filename: "13614c_#{intake.name_for_filename}.pdf") do %>
+                      <%= "13614c_#{intake.name_for_filename}.pdf" %>
```
kayline, line 31: Non-blocking, but because it's common for people to re-start the app, sometimes more than once, many tickets have an duplicate intake that doesn't have a 13614c. Not sure how that will behave here.

bengolder: It would show a mostly empty 13614-C

PR #273, opened 2020-05-26
Adds source, referrer, visitor_id to diy_intake

jenny-heath:
- fixes state tracking on intake to read from state_of_residence
- adds diy_intake_id to request_details for logs

[#173016392] Track request info on DIY intake and DIY intake id on request details
Co-authored-by: Jonathan Greenberg <jgreenberg@codeforamerica.org>

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,26 +9,34 @@ class ApplicationController < ActionController::Base
     current_user&.intake || Intake.find_by_id(session[:intake_id])
   end
 
+  def current_diy_intake
+    DiyIntake.find_by_id(session[:diy_intake_id])
+  end
+
   def include_google_analytics?
     false
   end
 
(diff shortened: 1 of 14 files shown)


app/controllers/application_controller.rb
```diff
@@ -9,26 +9,34 @@ def current_intake
...
+    DiyIntake.find_by_id(session[:diy_intake_id])
+  end
+
   def include_google_analytics?
     false
   end
 
+  def visitor_record
+    current_intake
+  end
```
bengolder, line 22: This is an interesting abstraction. It might be something that we could expand beyond just our handling of the visitor_id. A generic "key model" or "navigation record" or something could be used to make all our flow, forms, and other things more modular and less dependent on a specific type of model. For example, what if the personal info form could work on any model that had preferred name and state of residence fields?

PR #280, opened 2020-05-28
Add script to create a deploy tag

tdooner:
This script standardizes the release process. It:

1. Fetches the latest status of the repos so you are up to date when
   deploying.
2. Provides a full list of what will be deployed.
3. Creates a template message for the tag

Tags created with this script show up in Github like this:
https://github.com/codeforamerica/vita-min/releases/tag/version-1.0.10

create-deploy-tag script

### DIFF
--- a/bin/create-deploy-tag
+++ b/bin/create-deploy-tag
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -euo pipefail
+
+production_git_remote=$(git remote -v | grep 'vita-min-prod.git' | head -n1 | awk '{ print $1 }')
+origin_git_remote=$(git remote -v | grep 'github.com:codeforamerica/vita-min' | head -n1 | awk '{ print $1 }')
+
+# Get the latest tags and branches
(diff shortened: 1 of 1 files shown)


jonsgreen:
@tdooner This looks really sweet and helpful. I am impressed with your bash skills! We could consider using this cli tool: https://hub.github.com/ which actually will create the big release message in Github here: https://github.com/codeforamerica/vita-min/releases. I would enjoy pairing with you on it if you want since you would help me refresh my shell scripting.

bengolder:
@tdooner The process we have in place is to use Github's release interface and to write a summary of what is going out. We already had some scripts that Ben V wrote for automating this, but the CLI produces a less desirable result than a manual release in the Github interface. I'd prefer to stick with Github for now, but thank you.

bengolder:
If we can use cli tools that result in an equivalent level of clarity & documentation to the manual Github GUI process, that sounds like something that would be useful to others.

tdooner:
Sounds good, seems like using the `gh` gem might be the way to go to make the full release notes visible. I'll close this and follow along with Ben's attempt.

PR #281, opened 2020-05-29
Adds banking info page accessing to Zendesk user with ticket access

jonsgreen:
- refactors zendesk admin views into admin.html.erb layout
- adds admin-details components stylesheet including link colors from
figma styles.

[#173044379] Add page for viewing banking info

### DIFF
--- a/app/views/zendesk/intakes/banking_info.html.erb
+++ b/app/views/zendesk/intakes/banking_info.html.erb
@@ -1,0 +1,28 @@
+<% content_for :page_title, "#{@intake.primary_full_name}, ##{zendesk_ticket_id}" %>
+<% content_for :body_class, "admin-details" %>
+<% content_for :card do %>
+  <div class="header">
+    <h1 class="h2">
+      <%= @intake.primary_full_name %>
+    </h1>
+    <p>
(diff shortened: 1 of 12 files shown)


app/views/zendesk/intakes/banking_info.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <h1 class="h2">
+      <%= @intake.primary_full_name %>
+    </h1>
+    <p>
+      <%= link_to "##{zendesk_ticket_id}",
+        "https://eitc.zendesk.com/agent/tickets/#{zendesk_ticket_id}" %>
+    </p>
+  </div>
+
+  <div class="info-details">
```
kayline, line 14: Could this class name be made more specific? Maybe `bank-info-details`? I worry that as the CSS gets more complicated this will be hard to keep track of.

jonsgreen: I was a bit unsure about where this css would go and was trying to make something that could be applied on other detail pages. I am going to leave this since the next story (currently working on) is going to change the styling for these pages and so hopefully things will get clarified there.

PR #282, opened 2020-05-30
Adds table styles for admin documents list page

jonsgreen:
- adds DocumentPresenter for wrapping collection of documents with
convenient helper methods for view
- adds link to bank info page
- adds notes for documents of uncommon file types and large byte size
- adds purple styling colors for links
- adds Zendesk logo to OAuth sigin in page
- adds admin.scss template stylesheet

[#173044379] Add page for viewing banking info


app/assets/stylesheets/components/_document-table.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    &.upload-date-column {
+      width: 129px;
+    }
+
+    &.notes-column {
+      width: 250px;
+    }
+  }
```
bengolder, line 22: might be clearer as 
```scss
.document-column {
  &__doc-type {}
  &__link {}
  &__date {}
  &__notes {}
```

app/assets/stylesheets/components/_document-table.scss
```diff
@@ -0,0 +1,48 @@
...
+
+  td {
+
+    &.document-notes {
+      color: $color-gold-accent;
+    }
```
bengolder, line 28: name-wise, I'm having trouble guessing that these are. I would assume that the notes column is not always yellow. Is this for warnings only?

app/assets/stylesheets/components/_document-table.scss
```diff
@@ -0,0 +1,48 @@
...
+
+      &:hover {
+        color: #5C5C5C;
+      }
+    }
+
+    img.notes {
+      position: relative;
+      top: 4px;
+    }
```
bengolder, line 46: naming this with a class would be helpful. perhaps `icon--inline-notice`?

app/assets/stylesheets/templates/_admin.scss
```diff
@@ -0,0 +1,44 @@
...
+  div.header {
+
+    * {
+      margin: 0;
+      margin-bottom: 1.0rem;
+    }
+
+    h1 {
+      margin-bottom: 1.0rem;
+    }
```
bengolder, line 12: These look like redundant rules. keep in mind that we do have spacing classes that you can add to html.

app/assets/stylesheets/templates/_admin.scss
```diff
@@ -0,0 +1,44 @@
...
+
+      a {
+        color: $color-purple;
+        font-size: 1.9rem;
+        font-weight: bold;
+        &:hover {
+          color: #8E8EEF;
+        }
+      }
+    }
```
bengolder, line 24: a class would be helpful to understand what is being styled

app/assets/stylesheets/templates/_admin.scss
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  div.info-details {
+    margin-top: 6.0rem;
+
+    .h3 {
+      margin-bottom: 1.0rem;
+    }
+  }
```
bengolder, line 38: usually it's best to avoid tag selectors and stick with classes. so for example, `.info-details` instead of `div.info-details`.

app/assets/stylesheets/components/_document-table.scss
kayline: The notes column only contains warnings about large file size or uncommon file type. The mock explicitly titles this column "Notes".

app/assets/stylesheets/components/_document-table.scss
kayline: Good call, will do.

PR #286, opened 2020-06-02
Revert Taxslayer FSA link back to plain white page

tdooner:
Upon further analysis, this page does a POST to start a session, and
accounts created via the login page without this POST behavior do not
seem to be associated with the free FSA product.

This branch also reverts the launch of the admin portal, to give us some more time to prepare the communications to partners before it launches.


tdooner:
(Waiting to merge this until we have confirmed the fix with Courtney)

PR #289, opened 2020-06-03
Cron job for unsent docs #173063073

kayline:
Creates a cron job to run a task that checks for unsent documents and adds comments to tickets when unsent documents are found.

### DIFF
--- a/app/services/unsent_documents_service.rb
+++ b/app/services/unsent_documents_service.rb
@@ -1,5 +1,3 @@
-
-
 class UnsentDocumentsService
   include ZendeskServiceHelper
   include Rails.application.routes.url_helpers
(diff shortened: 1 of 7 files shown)


app/services/unsent_documents_service.rb
```diff
@@ -0,0 +1,34 @@
+
```
jonsgreen, line 1: Silly little knit but are the extra lines here intentional?

app/services/unsent_documents_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    Intake.where.not(intake_ticket_id: nil).find_each(batch_size: 100) do |intake|
+      unsent_docs = intake.documents.where(zendesk_ticket_id: nil).where("created_at < ?", 15.minutes.ago)
+      if unsent_docs.present?
+        comment_body = <<~BODY
+          New client documents are available to view: #{zendesk_ticket_url(id: intake.intake_ticket_id)}
+          Files uploaded:
+          #{unsent_docs.map {|d| "* #{d.upload.filename} (#{d.document_type})"}.join("\n")}
+        BODY
+
+        append_comment_to_ticket(
```
jonsgreen, line 22: I wonder if we should add the zendesk_ticket_url to the custom field as well? Might be a question for @bengolder or   Kelly.

bengolder: Yes we should, that is a good catch. but non-blocking

PR #290, opened 2020-06-03
Adds backfill for adding client document field to older ZD tickets

jonsgreen:
[#173143583] Backfill Link to Client Documents

### DIFF
--- a/app/lib/backfill/ticket_client_document_link.rb
+++ b/app/lib/backfill/ticket_client_document_link.rb
@@ -1,0 +1,43 @@
+module Backfill
+  class TicketClientDocumentLink
+    include ZendeskServiceHelper
+    include Rails.application.routes.url_helpers
+
+    def call
+      Intake
+        .includes(:documents)
+        .where.not(documents: { id: nil })
+        .where.not(intake_ticket_id: nil)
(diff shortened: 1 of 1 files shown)


app/lib/backfill/ticket_client_document_link.rb
```diff
@@ -0,0 +1,43 @@
...
+    include Rails.application.routes.url_helpers
+
+    def call
+      Intake
+        .includes(:documents)
+        .where.not(documents: { id: nil })
+        .where.not(intake_ticket_id: nil)
+        .select(:intake_ticket_id).distinct
+        .pluck(:intake_ticket_id)
+        .each do |ticket_id|
```
kayline, line 13: Maybe better to use `find_each` for memory reasons, even with the pluck?

jonsgreen: I think that pluck just returns an array so find_each is not possible.

I don't think there will be memory concerns since it won't be a very large array and most of all it is not instantiating large ActiveRecord objects.

jonsgreen: But definitely good that we are all more aware of this risk in different contexts!

PR #291, opened 2020-06-04
Release from production branch #173115497

bvandgrift:
Two primary efforts:

1. replace the tagged-deploy circleci workflow with a workflow that deploys on a commit to the production branch
2. modifies @tdooner 's script into a release script, integrating the `hub` tool for rich release messaging


bin/release
```diff
@@ -0,0 +1,103 @@
...
+
+Describe the release--include high-level changes, _markdown accepted_
+
+* #123123123 - story/feature description
+* short description of change
+* another change
+
+# Full list of changes (remove this line and all following before saving):
+$(git log --pretty="format:# * %s (%an)" ${latest_version}...${origin_git_remote}/master)
+# DON'T FORGET TO SAVE THE FILE
```
bengolder, line 68: Anything we don't want displayed in the markdown we should turn into an html comment.

PR #292, opened 2020-06-04
Add CASH Campaign of MD

combinatorist:


db/vita_partners.yml
```diff
@@ -261,3 +261,12 @@ vita_partners:
   source_parameters:
   - LABC
   logo_path: partner-logos/louisville-abc.png
+- name: CASH Campaign of MD
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360010772553'
+  display_name: CASH Campaign of MD
```
tdooner, line 267: I think we should spell out most abbreviations like "MD" in the display name for users. E.g. see what I did for MASSCAP above.

"CASH" seems fine since it's a word even if it technically stands for something.

combinatorist:
Gotcha - wasn't sure what that comment about the URL meant. I'll reset the branch.

combinatorist:
I agree about merging our when it comes to clients, @tdooner (slack thread), but it doesn't look like I have permissions. Can you do this one, with me taking full responsibility (or get me perms)?

tdooner:
I'll try to get you permission and merge this one for you.

PR #293, opened 2020-06-05
Finish translating gyr #173051001

jonsgreen:


app/assets/stylesheets/application.scss
```diff
@@ -133,6 +133,10 @@ body {
   }
 }
 
+.main-header .toolbar__item {
+  color: white;
+}
+
```
loumoore, line 139: fyi, this is for the language switcher link that is commented out until we turn spanish on in production. i think we added some new header links for zendesk admin while i was working on this too but i couldn't log in at /zednesk/sign-in in dev to make sure it plays nice with the header changes i made.

app/controllers/documents/ids_controller.rb
```diff
@@ -1,12 +1,8 @@
 module Documents
   class IdsController < DocumentUploadQuestionController
     def edit
-      @title = "Attach a photo of your ID card"
-      @help_text = "The IRS requires us to see a current drivers license, passport, or state ID."
       @names = [current_intake.primary_full_name]
       if current_intake.filing_joint_yes?
-        @title = "Attach photos of ID cards"
-        @help_text = "The IRS requires us to see a current drivers license, passport, or state ID for you and your spouse."
```
loumoore, line 9: all of this kind of logic is now handled by i18n pluralization rules

app/helpers/application_helper.rb
```diff
@@ -14,9 +14,4 @@ def round_meters_up_to_5_mi(meters)
     (meters / five_miles_in_meters).ceil * 5
   end
 
-  def you_or_spouse(intake)
-    return "you or your spouse" if intake.filing_joint_yes?
-
-    "you"
-  end
```
loumoore, line 21: this is also replaced with i18n pluralization rules in views. embedding string fragments like this inside of longer phrases tends to be a no-no for i18n.

app/models/diy_intake.rb
```diff
@@ -42,7 +42,7 @@ def issue_token
   end
 
   def start_filing_url
-    Rails.application.routes.url_helpers.diy_start_filing_url(token)
+    Rails.application.routes.url_helpers.diy_start_filing_url(:token => token)
```
loumoore, line 45: fyi, this is a common thing to watch out for when adding a locale scope to the app to support i18n. the first argument to paths is now always locale, so if you want to pass another arg like the token here or an id, it needs to be explicitly named

app/models/intake.rb
```diff
@@ -390,7 +390,7 @@ def get_or_create_requested_docs_token
   end
 
   def requested_docs_token_link
-    "#{Rails.application.routes.url_helpers.root_url}documents/add/#{get_or_create_requested_docs_token}"
+    "#{File.join(Rails.application.routes.url_helpers.root_url, "")}documents/add/#{get_or_create_requested_docs_token}"
```
loumoore, line 393: makes sure there's a trailing / if needed.

app/assets/stylesheets/application.scss
jonsgreen: Good to know, thanks.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -18,6 +19,7 @@ def failure
     error = request.env["omniauth.error"]
     if error_type == :access_denied
       # the user did not grant permission to view their info
+      # TODO: i18n
```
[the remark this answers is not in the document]
jonsgreen: Do others have opinions on whether to translate?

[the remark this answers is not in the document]
jenny-heath: we're gonna hold off on this for now (signing in locally is proving complicated) and try to get the branch merged first

PR #296, opened 2020-06-08
Add logo for Granite United Way

tdooner:


tdooner:
I verified it looks good in the browser.

PR #299, opened 2020-06-09
Update UW Bay Area routing to not go to CA Online Intake

tdooner:
The entry in `vita_partners.yml` for United Way Bay Area actually
corresponded to the "Online Intake" group for the whole California
coalition.

Per partner support ticket #24957, we should update the routing such
that:
* "UWBA" source routing goes to UWBA (AK state routing remains with UWBA)
* "UWCA" source routing and CA state routing goes to CA Online Intake

I also reordered the YAML file so all the United Way of California
groups are next to each other.

One subtlety here is that our partner upsert behavior uses the partner's
group ID as a primary key, so the record for "United Way Bay Area" will
be renamed to "[United Way California] Online Intake", and a new record
for "United Way Bay Area" will be created. This might cause a bit of a
metrics incompatibility if we are looking at the partner name for "uwba"
source code routing. Ultimately, this seems pretty minor to me though.


tdooner:
@bvandgrift tagging you for review here since this is the first time we're going to rename a partner. I audited the upsert behavior and I think it will work fine (since it uses `group id` to match and updates the existing record) but I want to tag you for visibility anyhow.

tdooner:
Thanks. (I made one last minute update to the partner group for UWBA). Answers:

* The intakes that have already been routed don't need to be updated. The California partners work closely together and have been able to manually re-route the intakes that came through.
* The metrics over time will have a discontinuity in partner if we're looking at the "uwba" source code or "CA" state routing. But the group IDs should still be right. I don't think we use those metrics for anything critical at the moment, so I can't think of a reason for this to block or change our approach.

PR #302, opened 2020-06-09
Fix "FAQ Link" on /tax-questions

tdooner:
It wasn't HTML-safe, so was outputting the escaped version of the HTML.


jenny-heath:
jk specs are failing...

tdooner:
Oh, duh. Okay, there's the version that should work.

PR #321, opened 2020-06-16
Progress bar for intake questions #172721835

kayline:


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   before_action :redirect_to_getyourrefund, :set_visitor_id, :set_source, :set_referrer, :set_utm_state, :set_sentry_context, :check_maintenance_mode, :check_at_capacity
   after_action :track_page_view
   around_action :switch_locale
-  helper_method :include_google_analytics?, :current_intake
+  helper_method :include_google_analytics?, :current_intake, :show_progress?
```
bengolder, line 7: Might it be better to only include this for QuestionsControllers and DocumentControllers? I don't believe we have a need to measure or display progress outside of those.

app/services/intake_progress_calculator.rb
```diff
@@ -8,14 +8,20 @@ class IntakeProgressCalculator
...
       controller_step.show?(intake)
     end
-    current_index = current_flow.index(controller)
-    ((current_index + 1)/current_flow.count.to_f * 100).round
+    current_index = steps_for_intake.index(controller)
+    completed_steps = current_index + 1 # Everything that the client has done
+
+    index_of_possible_steps = POSSIBLE_STEPS.index(controller)
+    possible_remaining_steps = POSSIBLE_STEPS[index_of_possible_steps + 1..-1].length # Everything the client could potentially do
+
```
jonsgreen, line 24: Nice! I am glad that you went ahead and made this improvement on the calculation.

app/services/intake_progress_calculator.rb
```diff
@@ -0,0 +1,27 @@
+class IntakeProgressCalculator
+
+  starting_index = QuestionNavigation::FLOW.index(Questions::BacktaxesController)
+  question_steps = QuestionNavigation::FLOW[starting_index..-1]
+  doc_overview_index = question_steps.index(Questions::OverviewDocumentsController)
+  POSSIBLE_STEPS = question_steps.insert(doc_overview_index + 1, *DocumentNavigation::FLOW)
+
+  def get_progress(controller, intake)
```
jonsgreen, line 8: @kayline I had wanted to tell you that I only meant to make this an instance method if we ended up breaking up the code into smaller private methods. This still could be done in which case it can be nice to have an `initialize` method for setting instance variables for controller and intake that can be used everywhere. However, as the code is now I think this could just be a class method in which case you could call it with `IntakeProgressCalculator.get_progress(controller, intake)`. Does that make sense? If not we can chat about it.

app/controllers/application_controller.rb
kayline: Other controllers in the app use the layouts that include the progress bar (`questions` and `yes-no`), and those controllers need access to `show_progress` in their views so they don't try to display the progress bar.

app/services/intake_progress_calculator.rb
kayline: That makes sense, I can update it

PR #326, opened 2020-06-18
Move validation messagetranslations to config/locales errors key

jonsgreen:
[#173111774] Turn on Spanish translations

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -14,22 +14,70 @@ en:
     errors:
       at_least_one_year: Please pick at least one year.
       need_phone_number: Please enter a cell phone number.
-    validators:
-      answer_presence: Please answer or click "Skip question"
-      email_address_confirmation_match: Please double check that the email addresses match.
-      email_address_valid: Please enter a valid email address.
-      generic_presence: Can't be blank.
-      job_count_presence: Please select a number of jobs.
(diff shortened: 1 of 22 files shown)


config/locales/en.yml
```diff
@@ -14,22 +14,70 @@ en:
...
-      phone_number_valid: Please enter a valid phone number.
-      preferred_name_presence: Please enter your preferred name.
-      primary_first_name_presence: Please enter your first name.
-      primary_last_four_ssn_length: Please enter the last four digits of your SSN or ITIN.
-      primary_last_name_presence: Please enter your last name.
-      spouse_first_name_presence: Please enter your spouse's first name.
-      spouse_last_four_ssn_length: Please enter the last four digits of your spouse's SSN or ITIN.
-      spouse_last_name_presence: Please enter your spouse's last name.
-      state_of_residence_inclusion: Please select a state from the list.
+  errors:
```
loumoore, line 17: I like this approach ++

Just FYI: for GCF we put things under the "activerecord.errors.messages" namespace exclusively, with key names closer to what I had before (flat and custom rather than nested by attribute and following convention). but while that can reduce string duplication in the yml files, it also requires some funky code to set the correct error messages in many cases. string duplication is cheap, transifex handles it cleanly, so this seems preferable. and i can't recall any other reason to prefer that other approach.

PR #332, opened 2020-06-22
Redirect Impact America routing to FC and UWGC

combinatorist:
Per Annelise's request in https://eitc.zendesk.com/agent/tickets/34253:

1. SC state routing - move to UW Greenville (I wrote him to confirm this is ok, but if you don't hear from me by Monday, please go ahead and make the change)
2. Deactivate source code routing for Impact America (clients will then be routed by state) - https://www.getyourrefund.org/?s=IA
3. Send TN, MS, AR to Foundation Communities/ the national site pool (not sure how you have that working)

### DIFF
--- a/db/vita_partners.yml
+++ b/db/vita_partners.yml
@@ -65,9 +65,6 @@ vita_partners:
   - fc
   states:
   - TX
-  - TN
-  - AR
-  - MS
   accepts_overflow: true
   logo_path: ''
 - name: United Way of Central Ohio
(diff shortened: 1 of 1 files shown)


db/vita_partners.yml
```diff
@@ -73,6 +65,9 @@ vita_partners:
   - fc
   states:
   - TX
+  - TN
+  - AR
+  - MS
```
combinatorist, line 70: I routed these states to Foundation Communities, but is there a better way to share them across the "national pool"?

tdooner: I think it might be slightly preferable to keep the `states` list here to be only the states we've explicitly promised a partner, rather than making sure to list all of the states. Any missing states will be routed via the `accepts_overflow: true` option, which will go to FC anyway.

combinatorist: Agreed. I made those changes, but I cannot merge, so I think you'll need to.

PR #335, opened 2020-06-23
Move FL routing to City of Miami

tdooner:
Per Gregory, our new City of Miami partner will become the new default
recipient of all Florida returns.

Please hold off on merging this until I am able to double-check with Gregory about it, but I think we should be good here.


tdooner:
We determined this was not the expectation of the partner, so I'm closing this out. Returns will be manually assigned from RefundDay coalition to the City of Miami.

PR #338, opened 2020-06-23
Adds redirect to home page with message when going back after req docs

jonsgreen:
- Silence Puma when running javascript feature specs

[#172969038] Clients throw errors using back button after submitting requested
documents

### DIFF
--- a/app/controllers/documents/send_requested_documents_later_controller.rb
+++ b/app/controllers/documents/send_requested_documents_later_controller.rb
@@ -5,10 +5,17 @@ module Documents
 
     def edit
       documents_request = DocumentsRequest.find_by(id: session[:documents_request_id])
-      intake = documents_request.intake
-      documents_request.documents.update_all(intake_id: intake.id)
-      SendRequestedDocumentsToZendeskJob.perform_later(intake.id)
-      redirect_to documents_requested_documents_success_path
+      if documents_request.nil?
+        redirect_to(
+          root_path,
+          alert: t("controllers.send_requested_documents_later_controller.not_found")
+        )
+      else
+        intake = documents_request.intake
+        documents_request.documents.update_all(intake_id: intake.id)
+        SendRequestedDocumentsToZendeskJob.perform_later(intake.id)
+        redirect_to documents_requested_documents_success_path
+      end
     end
 
     def success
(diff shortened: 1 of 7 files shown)


app/controllers/documents/send_requested_documents_later_controller.rb
```diff
@@ -5,10 +5,17 @@ class SendRequestedDocumentsLaterController < DocumentUploadQuestionController
...
     def edit
       documents_request = DocumentsRequest.find_by(id: session[:documents_request_id])
-      intake = documents_request.intake
-      documents_request.documents.update_all(intake_id: intake.id)
-      SendRequestedDocumentsToZendeskJob.perform_later(intake.id)
-      redirect_to documents_requested_documents_success_path
+      if documents_request.nil?
+        redirect_to(
+          root_path,
+          alert: t("controllers.send_requested_documents_later_controller.not_found")
```
bengolder, line 11: I believe this will style it as a yellow or red banner, as if it were an error or warning. However, the content sounds more like a success confirmation. Should we be using alert or notice?

spec/features/web_intake/requested_documents_token_spec.rb
```diff
@@ -45,4 +45,35 @@
...
+  context "with forgery protection" do
+    around do |example|
+      ActionController::Base.allow_forgery_protection = true
+      Capybara.raise_server_errors = false
+      example.run
+      Capybara.raise_server_errors = true
+      ActionController::Base.allow_forgery_protection = false
+    end
+
+    scenario "client completes requested docs and gets message when trying to go back", :js do
```
bengolder, line 58: We should be able to test this without Javascript. Is the intention here to test the javascript functionality?

spec/support/navigation_helpers.rb
```diff
@@ -0,0 +1,3 @@
+def go_back
+  page.evaluate_script('window.history.back()')
+end
```
bengolder, line 3: How does this relate to the PR?

spec/features/web_intake/requested_documents_token_spec.rb
kayline: I tried to make these tests work without JS for about an hour and made no progress. We'll try to catch you for an in-person conversation but I don't think this needs to block release of the bugfix.

PR #342, opened 2020-06-24
greenville not ready for sc yet

combinatorist:


combinatorist:
@tdooner, who should I confirm with? This came from Annelise.

(fixing the tests now)

PR #347, opened 2020-06-25
Fixes bugs related to requested docs

jonsgreen:
- replaces documents_request_id in session when new token param is
requested
- corrects the documents_remove_requested_document_path signature to
pass document id in a hash to generate correct route

[#173504865] Multiple requested docs links leak across session
[#173504893 Removing uploaded requested document throws 404

Signed-off-by: Jenny Heath <jheath@codeforamerica.org>

### DIFF
--- a/app/controllers/documents/requested_documents_later_controller.rb
+++ b/app/controllers/documents/requested_documents_later_controller.rb
@@ -62,7 +62,7 @@ module Documents
     private
 
     def destroy_document_path(document)
-      documents_remove_requested_document_path(document)
+      documents_remove_requested_document_path(id: document)
     end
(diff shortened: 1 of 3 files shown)


app/controllers/documents/requested_documents_later_controller.rb
```diff
@@ -87,11 +87,16 @@ def current_session_or_home
     end
 
     def handle_session
-      return if session[:documents_request_id].present?
+      return if session[:documents_request_id].present? &&
```
jenny-heath, line 90: non-blocking: consider moving this into a function name that describes what's happening

PR #348, opened 2020-06-25
distribute overflow returns evenly across available overflow partners

loumoore:
allow multiple overflow partners, and assign to them based on the intake's id modulo the number of available partners. this will distribute overflow intakes evenly in a round-robin fashion across however many overflow partners we have. 

this change also re-enables UWTSA as a second overflow partner along with FC per the story [here](https://www.pivotaltracker.com/story/show/173505584)

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -533,8 +533,12 @@ class Intake < ApplicationRecord
   end
 
   def partner_for_overflow
-    # when we have more than one partner that accepts overflow, this should balance the load
-    partner = VitaPartner.find_by(accepts_overflow: true)
+    # assign overflow intakes to available overflow partners evenly by using the intake id modulo the number of
+    # available overflow partners
+    partners = VitaPartner.where(accepts_overflow: true)
+    return nil if partners.empty?
+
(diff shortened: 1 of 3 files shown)


jonsgreen:
@loumoore This looks good. I assume that we don't have to be concerned about distributing the load in a more nuanced or weighted way at this time. Also I just want to make sure you saw this brief that @bvandgrift wrote up last week for if/when something more automated is needed for adjusting flow based on capacity: https://docs.google.com/document/d/1VRCAMD6_Q5V0qbZUs50gzJRpfqIlhwOS__DZQ4Mh1uU/edit.

loumoore:
@jonsgreen i did take a look at that! quickly came across it in the code when i was searching for where this routing was happening which was cool.

and yeah, i was just hoping to implement something simple that wouldn't get in the way of future refactoring here, after discussing some of the overflow challenges with Kelly and Annelise yesterday afternoon. This simple load balancing should work well enough to make it work a lot better than it does right now, and we can of course return to the original functionality if needed by simply going back to only having one partner set as an overflow.

thanks for the review!

PR #349, opened 2020-06-25
update MASSCAP to include BTH and match ZD

combinatorist:


combinatorist:
No idea why this PR is failing, especially after the second commit: https://github.com/codeforamerica/vita-min/pull/349/files/483756601338c801b05cc7137bf29dd3f3b1dbe2. I wonder if it won't let us rename groups here.

combinatorist:
nm - we're using group name in the spec. I'll fix it.

db/vita_partners.yml
```diff
@@ -220,51 +220,73 @@ vita_partners:
...
   zendesk_instance_domain: eitc
   zendesk_group_id: '360010415253'
   display_name: Massachusetts Association for Community Action (MASSCAP)
   source_parameters:
   - MC-WOR
   logo_path: ''
+- name: "[MA/BTH] ABCD"
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360011283393'
+  display_name: 'Boston Tax Help'
```
tdooner, line 271: source paramater: BTHC-ABCD

db/vita_partners.yml
```diff
@@ -220,51 +220,73 @@ vita_partners:
...
+  logo_path: ''
+- name: "[MA/BTH] AB APAC"
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360011284774'
+  display_name: 'Boston Tax Help'
```
tdooner, line 276: source parameter: BTHC-AB-APAC

db/vita_partners.yml
```diff
@@ -220,51 +220,73 @@ vita_partners:
...
+  logo_path: ''
+- name: "[MA/BTH] Roxbury & Jamaica Plain"
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360011284494'
+  display_name: 'Boston Tax Help'
```
tdooner, line 281: source parameter: BTHC-ROXBURY

tdooner:
Side note: when you merge this can you use the "Squash" option? I don't think the intermediate commits add much useful context in this case.

combinatorist:
@tdooner, I actually removed the BTH source codes because Brian said they weren't ready to be routed (except StreetCred which wants it urgently). Doesn't this immediately turn them on? I suppose they could just avoid advertising those codes, but the cat may be out of the bag - IDK.

PR #365, opened 2020-07-01
add logo and ZD group for ncc

combinatorist:


db/vita_partners.yml
```diff
@@ -419,3 +419,8 @@ vita_partners:
   source_parameters:
   - greaternashville
   logo_path: partner-logos/uw-greater-nashville.png
+- name: New Community Corporation
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360011727473'
+  display_name: New Community Corporation
+  logo_path: partner-logos/ncc-logo.png
```
tdooner, line 426: no unique URL?

combinatorist: They didn't when I made the draft the PR. In fact, despite this being "urgent" last week, they still haven't loaded their users to their sheet.

PR #374, opened 2020-07-02
update north texas area and pocono mountains united way logos

combinatorist:


db/vita_partners.yml
```diff
@@ -387,7 +387,7 @@ vita_partners:
   display_name: North Texas Area United Way
   source_parameters:
   - ntauw
-  logo_path: partner-logos/ntauw.jpg
```
tdooner, line 390: remove the old one?

PR #376, opened 2020-07-06
Add Central City Neighborhood Partners

GAV07:


db/vita_partners.yml
```diff
@@ -428,3 +428,10 @@ vita_partners:
   source_parameters:
   - uwpm
   logo_path: 'partner-logos/uwpm-logo.png'
+- name: Central City Neighborhood Partners [Los Angeles Ventanilla VITA site]
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360011733194'
+  display_name: Central City Neighborhood Partners
+  source_parameters:
+  - laccnp
+  logo_path: 'partner-logos/CCNP-logo.jpeg'
```
tdooner, line 437: it looks like you didn't include this file in the commit 😅

tdooner: Oh, I see, it's included in your other commit. But the filenames don't match - this calls for a filename with a dash in it, but the filename has an underscore in it.

GAV07: Yeah I was trying to do them straight on github but I'll just stick to the terminal so I can do it right lol.

PR #378, opened 2020-07-06
Anonymized csv admin download #172206916

jonsgreen:
Screen Shot 2020-07-06 at 3 52 09 PM

[anonymized-intakes-2020-07-06.csv.zip](https://github.com/codeforamerica/vita-min/files/4880955/anonymized-intakes-2020-07-06.csv.zip)


app/services/anonymized_intake_csv_service.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  private
+
+  def csv_row(intake)
+    CSV_FIELDS.map { |field| intake.send(field) }
+  end
+
+  class AnonymizedCSVIntake < SimpleDelegator
+
+    def dependent_count
```
jonsgreen, line 87: These methods are not terribly performant but I think it would take some clever SQL/Arel to avoid. I thought it might not be worth the effort and complication at this time since this will be happening as a background job but it may be worth considering.

bengolder: We can optimize that later if it is a problem.

spec/features/zendesk/anonymized_intake_csv_viewing_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      # redirected to sign in page
+      expect(page).to have_text "Sign in with Zendesk"
+
+      OmniAuth.config.mock_auth[:zendesk] = auth_hash
+      click_link "Sign in with Zendesk"
+
+      expect(page).to have_content("You are not authorized to access that page")
+    end
+
+  end
```
bengolder, line 60: I don't think we need this feature spec

spec/features/zendesk/anonymized_intake_csv_viewing_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    filename = extract.upload.attachment.filename.to_s
+    extract_row = find("table tbody tr").text
+    expect(extract_row)
+      .to eq([filename, extract.record_count, extract.run_at].join(" "))
+
+    click_on filename
+    expect(page.response_headers).to include("Content-Type" => "text/csv")
+    expect(page.response_headers["Content-Disposition"]).to include("attachment")
+  end
```
bengolder, line 43: I'd prefer to cover all this functionality in controller specs and avoid adding a feature spec. Non-blocking though. I just want to keep our test suite fast.

spec/jobs/anonymized_intake_csv_job_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      before do
+        described_class.perform_now
+      end
+
+      it "calls the service" do
+        expect(AnonymizedIntakeCsvService).to have_received(:new).with(nil)
+        expect(fake_intake_csv_service).to have_received(:store_csv)
+      end
+    end
+  end
```
bengolder, line 37: It looks like there is no difference in functionality here. What is the value of this spec?

spec/features/zendesk/anonymized_intake_csv_viewing_spec.rb
jonsgreen: @bengolder I did end up removing this feature spec and moved some of what could be easily tested in the controller spec

PR #386, opened 2020-07-08
collect all group pages in ZD cli

combinatorist:
No idea what I'm doing or how ruby syntax really works, but I adapted some of your code @tdooner and it was able to find the second page of groups!

Feel free to tear it up if my ham-fisted implementation raises any other concerns or breaks any style conventions etc.

### DIFF
--- a/app/cli/zendesk_cli/import_users.rb
+++ b/app/cli/zendesk_cli/import_users.rb
@@ -188,7 +188,9 @@ class ZendeskCli
     # Load a cache of all Zendesk groups so we can easily search for a given
     # group by name.
     def zendesk_groups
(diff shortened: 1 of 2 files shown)


combinatorist:
Unfortunately, I don't have access and don't know what test this failing, but it's not urgent so I'll just mark it as a draft for now.

tdooner:
@combinatorist - Cool, thanks for this. It's a bit of a pain to test the block-based `all!` implementation so I just took a stab in https://github.com/codeforamerica/vita-min/pull/386/commits/bad045dea64531303fa13b347079bdc97fd874ff. Thanks for fixing. I'll merge when tests pass.

PR #393, opened 2020-07-09
DIY Page Update

bengolder:


config/locales/en.yml
```diff
@@ -108,9 +108,9 @@ en:
     find_vita_site: Find a VITA site near you
     free_file_html: If you'd like to file your taxes without the assistance of our staff and services, use the %{wizard_link} to find a free way to file your taxes online.
     free_file_lookup_tool: IRS Free File Lookup Tool
+    get_help_html: Get&nbsp;help
```
jenny-heath, line 111: i assume this is to prevent wrapping? is that even something we need to worry about with a fixed width button?

jenny-heath:
things to review with Anu:
- centered button on secondary cta slab (mobile)
- padding on secondary cta slab
- width of grid items in slab (2/3 vs 3/4)

PR #396, opened 2020-07-10
Adds DIY button to maybe ineligible page

jonsgreen:
[#172927473] Update freefile offboarding links to the DIY homepage link

Screen Shot 2020-07-10 at 4 19 12 PM

Screen Shot 2020-07-10 at 4 22 16 PM

Screen Shot 2020-07-10 at 4 23 16 PM

### DIFF
--- a/app/views/public_pages/maybe_ineligible.html.erb
+++ b/app/views/public_pages/maybe_ineligible.html.erb
@@ -14,17 +14,30 @@
 
           <div class="offboarding--help-text">
             <p><%=t("views.public_pages.maybe_ineligible.info") %></p>
-            <p><%=link_to t("general.submit_question"), "mailto:hello@getyourrefund.org" %></p>
+            <p><%=t("views.public_pages.maybe_ineligible.alternatives") %></p>
           </div>
           <div class="offboarding--help-text">
-            <p class="offboarding--subtitle"><%=t("general.other_options") %></p>
-            <p>
-              <%=t("general.free_file_html", :wizard_link => link_to(t("general.free_file_lookup_tool"), "https://apps.irs.gov/app/freeFile/jsp/wizard.jsp", target: "_blank", rel: "noopener", "data-track_click": "freefile")) %>
+            <p class="offboarding--subtitle">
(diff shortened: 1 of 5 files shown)


app/views/public_pages/maybe_ineligible.html.erb
```diff
@@ -14,17 +14,30 @@
...
+              <% end %>
+            </ul>
           </div>
 
           <div class="buttons--vertical-stack">
-            <%= link_to t("general.find_vita_site"), vita_providers_path, class: "button button--primary button--wide text--centered" %>
+            <%= link_to t("views.public_pages.maybe_ineligible.diy_button"),
+              diy_file_yourself_path,
+              class: "button button--primary button--wide text--centered",
+              "data-track-click": "click_diy_offboarding"
```
bengolder, line 39: `click_` is automatically prepended by `data-track-click`, so this will create an event called `click_click_diy_offboarding`. You can take `click_` off the attribute value.

config/i18n-tasks.yml
```diff
@@ -102,6 +102,7 @@ ignore_unused:
   - 'date.formats.*'
   - 'general.*'
   - 'errors.*'
+  - 'views.public_pages.maybe_ineligible.diy_criteria_list.{item_1,item_2,item_3,item_4}'
```
bengolder, line 105: Is there a reason we should ignore these?

jonsgreen: @bengolder I was trying to avoid duplication in the view by iterating [here](https://github.com/codeforamerica/vita-min/pull/396/files#diff-d23db4b6c12dc20c355b3387ac27349eR28). Unfortunately i18n-tasks is not clever enough to see that those interpolated keys are being used so the spec fails. I am happy to just be explicit in the view if you would prefer to avoid having this ignored in the yml?

bengolder: explicit in the view sounds preferable to me

jonsgreen:
This was merged directly into master while reviewing local changes with Ben Golder

PR #403, opened 2020-07-14
WIP: Record client efforts

bengolder:
Questions for a reviewer
- Should the effort types be constants or symbols so that it's easier to refer to them elsewhere in the codebase? 
- Are the presence & belongs to validations excessive or the right amount?

### DIFF
--- a/app/models/client_effort.rb
+++ b/app/models/client_effort.rb
@@ -1,0 +1,53 @@
+# == Schema Information
+#
+# Table name: client_efforts
+#
+#  id              :bigint           not null, primary key
+#  effort_type     :string           not null
+#  made_at         :datetime         not null
+#  responded_to_at :datetime
+#  response_type   :string
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  intake_id       :bigint           not null
+#  ticket_id       :bigint           not null
+#
+# Indexes
+#
+#  index_client_efforts_on_intake_id  (intake_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (intake_id => intakes.id)
+#
+class ClientEffort < ApplicationRecord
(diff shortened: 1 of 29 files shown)


app/models/client_effort.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+#  index_client_efforts_on_intake_id  (intake_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (intake_id => intakes.id)
+#
+class ClientEffort < ApplicationRecord
+  EFFORT_TYPES = %w(
+    consented
```
alaroia-cfa, line 25: I took a look (with you!) at this, and we could (I think) use an ActiveRecord Enum, where we store the actual strings in the database. This sounds like it adds no value, but if we also define the keys this way:

```
enum effort_type: string_map_from_keys [:consented, :completed_intake_questions, ...]
```

then RubyMine can auto-complete from those symbols.

It also adds syntax sugar, like, `client_effort.effort_sent_sms?` or `client_effort.sent_sms?` (depending on if we want the prefix).

That's one idea. Not a showstopper.

app/models/client_effort.rb
```diff
@@ -0,0 +1,53 @@
...
+    public_reply
+    phone_call
+    status_change
+  )
+
+  belongs_to :intake
+
+  validates_presence_of :ticket_id
+  validates_presence_of :made_at
+  validates_presence_of :effort_type
```
alaroia-cfa, line 47: I'm basically always a fan of making fields required first, then un-required later if you ever need to do that. +1 to this as-is.

bengolder:
I'm going to close this because we shelved this work.

bengolder:
but I'm not going to delete the branch yet!

PR #404, opened 2020-07-14
Send documents needed to zendesk

jenny-heath:
9128e88 Refactors document type recommendations to be in models instead of view helper
40ea4ff Actually fills out the zendesk field


app/models/intake.rb
```diff
@@ -528,6 +528,22 @@ def triaged_from_stimulus?
     triage_source.present? && triage_source.class == StimulusTriage
   end
 
+  def must_have_document_types
```
alaroia-cfa, line 531: Non-blocking suggestion: Since these two methods seem to be the same except for `might_have_doc_type?` vs `must_have_doc_type?`, is there a way to unify them a little? Honestly asking; happy as-is with the code.

app/views/documents/intro/edit.html.erb
```diff
@@ -18,37 +18,33 @@
           </p>
 
           <div class="document-overview-list">
-            <% if any_must_have_docs?(recommended_document_types) %>
+            <% if current_intake.must_have_document_types.present? %>
```
alaroia-cfa, line 21: This is a very fun line of code to read!

spec/models/intake_spec.rb
```diff
@@ -1125,4 +1124,44 @@
       it { expect(intake.triaged_from_stimulus?).to be_falsey }
     end
   end
+
+  describe "#must_have_documents" do
+    let(:intake) { create(:intake) }
+
+    before do
+      allow(DocumentNavigation).to receive(:document_types_for_intake).with(intake).and_return(["1095-A", "Selfie", "W-2"])
```
alaroia-cfa, line 1132: It's great that this PR mocks out `DocumentNavigation` -- really helps with clarity around separation of concerns.

app/models/intake.rb
jenny-heath: hm, that occurred to me but i couldn't think of anything off the top of my head - maybe that refactor can happen as part of [this story](https://www.pivotaltracker.com/story/show/173349198)...

PR #406, opened 2020-07-16
Racine and Suncoast Logos

GAV07:
Updating partner logos


db/vita_partners.yml
```diff
@@ -464,11 +464,18 @@ vita_partners:
   display_name: United Way of Racine County
   source_parameters:
   - unitedwayracine
-  logo_path: ''
+  logo_path: 'VITA & UWRC Logo - Horizontal RGB Lockup.png'
```
tdooner, line 467: This needs to start with `partner-logos/`. Also - it would be best to rename these files to not contain any special characters or spaces.

GAV07: No problem Tom will do.

PR #408, opened 2020-07-21
Racine and Suncoast (revised update)

GAV07:
Updated db and photos with corrected names.


bengolder:
Closing because we added the commit elsewhere

PR #409, opened 2020-07-23
Restrict file types #173943188

jenny-heath:
harder than i thought it would be


app/controllers/documents/document_upload_question_controller.rb
```diff
@@ -21,9 +21,11 @@ def update
         form_saved = @form.save
         after_update_success
         track_document_upload
+        redirect_to action: :edit
+      else
+        track_validation_error
+        render :edit
```
jenny-heath, line 27: this follows the same pattern as the questions controller - would love context on why we weren't doing this in the first place and whether there's a problem with it. the reason i changed it was that rendering errors in the usual way doesn't work if you redirect

app/validators/file_type_allowed_validator.rb
```diff
@@ -0,0 +1,15 @@
+class FileTypeAllowedValidator < ActiveModel::EachValidator
+  # The list of allowed types is security sensitive because it may allow files with script
+  # capabilities to be uploaded leading to XSS attacks. Please be sure to know what you are
+  # doing when modifying this list.
+  VALID_FILE_TYPES = [".jpg", ".jpeg", ".pdf", ".png", ".heic", ".bmp", ".txt", ".tiff", ".gif"]
+
+  def validate_each(record, attr_name, value)
+    return if !value
```
jenny-heath, line 8: according to the tests we're okay with `nil` documents ?

app/validators/file_type_allowed_validator.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    extension = File.extname(value.tempfile.path)
```
jenny-heath, line 10: i concluded that this was the best way to get the extension by studying the test file uploads and also putting a breakpoint in and testing locally to see what `value` is here. it appears to be something of type `ActionDispatch::Http::UploadedFile` not a document, so our original tactic of getting `.upload.attachment.filename.extension` didn't work. there could be a better way to do this!

app/views/layouts/document_upload.html.erb
```diff
@@ -61,6 +61,15 @@
...
               </div>
             </div>
+
+            <% if @form.errors[:document].any? %>
+              <div class="form-group form-group--error">
+                <p class="text--error"><i class="icon-warning"></i>
+                  <%= @form.errors.full_messages.first %>
+                </p>
+              </div>
+            <% end %>
```
jenny-heath, line 71: `file_field` doesn't seem to include errors in the way we do on our other form fields (see `VitaMinFormBuilder`) so i had to add them in

spec/controllers/documents/w2s_controller_spec.rb
```diff
@@ -56,7 +57,7 @@
         allow(subject).to receive(:send_mixpanel_event).and_return(true)
       end
 
-      it "appends the W-2 documents to the intake and rerenders :edit without redirecting" do
```
jenny-heath, line 59: the expectation says it redirects so no idea why this was the description...

spec/fixtures/attachments/test-pattern.html
```diff
@@ -0,0 +1,235 @@
+<!DOCTYPE html>
```
jenny-heath, line 1: i downloaded our FAQ page because i was in a hurry to get things to work but there's probably a better fixture to put here - i really just need the .html extension

app/validators/file_type_allowed_validator.rb
```diff
@@ -0,0 +1,15 @@
+class FileTypeAllowedValidator < ActiveModel::EachValidator
+  # The list of allowed types is security sensitive because it may allow files with script
+  # capabilities to be uploaded leading to XSS attacks. Please be sure to know what you are
+  # doing when modifying this list.
```
bengolder, line 4: A more helpful comment might be:

> Do not add file types that can execute code into this list (for example html, js, exe)

app/validators/file_type_allowed_validator.rb
bengolder: We may be able to use `value.content_type` or `value.original_filename` as well. You can also just call `value.path` I think. https://api.rubyonrails.org/v6.0.3.2/classes/ActionDispatch/Http/UploadedFile.html

bengolder: jpeg is also a valid extension for jpg

app/views/layouts/document_upload.html.erb
bengolder: Would be good to add screenshot

PR #410, opened 2020-07-23
set up the DIY landing page to support a messaging test based on source

loumoore:
this is a quick and dirty approach to testing some of the messaging suggestions from our comms firm.

if the source is one of two test variants that are set up in adwords, it will show different copy that was created to match the ads and test two different narratives.

there's an additional check on locale, even though the ads that will trigger the variants are all in english too.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -177,13 +177,18 @@ body {
       }
     }
 
+    .diy-header {
+      max-width: 450px;
+    }
+
     .diy-subheader {
       font-weight: 200;
-      max-width: 391px;
+      max-width: 450px;
     }
 
     .grid {
       position: relative;
+      min-height: 355px;
 
       @media screen and (max-width: $site-max-up) {
         position: static;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -177,13 +177,18 @@ body {
...
+
     .diy-subheader {
       font-weight: 200;
-      max-width: 391px;
+      max-width: 450px;
     }
 
     .grid {
       position: relative;
+      min-height: 355px;
```
loumoore, line 191: ensures the hero image isn't cut off at the bottom with short copy (the current live version is short)

app/assets/stylesheets/application.scss
```diff
@@ -177,13 +177,18 @@ body {
...
     }
 
+    .diy-header {
+      max-width: 450px;
+    }
+
     .diy-subheader {
       font-weight: 200;
-      max-width: 391px;
+      max-width: 450px;
```
loumoore, line 186: test copy is much longer and needed more space. original copy works fine with this too

app/assets/stylesheets/application.scss
bengolder: good catch!

PR #416, opened 2020-07-29
Remove only remaining usage of trace_error

jenny-heath:


app/helpers/consolidated_trace_helper.rb
```diff
@@ -115,6 +115,7 @@ def diy_intake_context(diy_intake)
     }
   end
 
+  # TODO: now that none of the below trace_ methods are used - can we delete them?
```
jenny-heath, line 118: not sure how much of the `consolidated_trace_helper` file we want to keep?

app/services/zendesk_intake_service.rb
```diff
@@ -30,10 +30,7 @@ def assign_requester
       return requester_id
     end
 
-    # failing all else, that's likely noteworthy
-    trace_error('ZendeskIntakeTicketJob failed to create a ticket requester',
-                intake_context(@intake))
-    return # ensure a nil return value
+    raise ZendeskServiceError.new("ZendeskIntakeTicketJob failed to create a ticket requester")
```
jenny-heath, line 33: just want to make sure i understand something - `trace_error` here included an `intake_context` but it looks like every place this method is called was wrapped in a `with_raven_context` so were those two things redundant?

PR #420, opened 2020-07-29
temple goodwill central city

GAV07:
db update for Temple University, Goodwill ORR, and Central City


db/vita_partners.yml
```diff
@@ -484,3 +484,24 @@ vita_partners:
...
+  display_name: Temple University
+  source_parameters:
+  - tuvs
+  logo_path: 'partner-logos/temple.jpg'
+- name: Goodwill of the Olympics and Rainier Region
+  zendesk_instance_domain: eitc
+  zendesk_group_id: '360012588394'
+  display_name: Goodwill of the Olympics and Rainier Region
+  source_parameters:
+  - orrr
```
bengolder, line 499: @GAV07 does `orrr` have the correct number of `r`s? it looks like the rest hints at just `orr`

db/vita_partners.yml
```diff
@@ -484,3 +484,24 @@ vita_partners:
...
+  logo_path: 'partner-logos/goodwill_orr.png'
```
bengolder, line 500: @GAV07 It looks like this image path does not match the file name

GAV07: Yes, I forgot to add the extra 'r' to the link

db/vita_partners.yml
GAV07: Yes it is correct although I am not sure why they would have added an extra r instead of just two.

GAV07: https://www.getyourrefund.org/?s=orrr (from [airtable](https://airtable.com/tbl5epnGDYQf8OAYZ/viwgXpuhMweqBKPIP/reck0uQ1GZjhbstX1?blocks=bipsOP1FkYFvEXxjd))

bengolder:
Closing because changes were moved into this commit:
https://github.com/codeforamerica/vita-min/commit/4b7d72d92926473a702f0ac660ee721dcbf5889b

PR #421, opened 2020-07-29
Add a test_ticket tag to new ZD tickets created outside of prod

jenny-heath:


spec/services/zendesk_intake_service_spec.rb
```diff
@@ -59,8 +59,10 @@
     DatadogApi.configure do |c|
       c.enabled = true
       c.namespace = "test.dogapi"
+      c.env = "production"
```
bengolder, line 62: Why was this line added?

bengolder: Was it needed to prevent breakage of the test?

jenny-heath: yes! because the test_ticket tag depends on the environment, i had to change all the places where env was set in the test to be consistent otherwise the test failed. it was actually a bit of a challenge because i didn't realize i had to unset it in the after block

PR #422, opened 2020-07-30
Add GYR Volunteer roles to Import Script

tdooner:
These roles have existed in Zendesk for a while now, and we've defined
processes for volunteers to get these levels of access. This is
documented in the [Access Matrix for GYR Volunteers][1].

[1]: https://docs.google.com/spreadsheets/d/1MySYITwcOmye2mLzusNV8TGArGPyrjoXasGGQlFxYPQ/edit#gid=0


tdooner:
Thank you both!

PR #426, opened 2020-08-06
Stop attaching drop off documents to zendesk ticket, show in doc portal instead

jenny-heath:
Co-authored-by: Dimitri DeFigueiredo <ddefigueiredo@codeforamerica.org>

### DIFF
--- a/app/controllers/zendesk/drop_offs_controller.rb
+++ b/app/controllers/zendesk/drop_offs_controller.rb
@@ -1,0 +1,22 @@
+module Zendesk
+  class DropOffsController < ApplicationController
+    include ZendeskAuthenticationControllerHelper
+    include FileResponseControllerHelper
+
+    before_action :require_zendesk_user, :set_drop_off, :require_ticket_access
+
+    def show
+      render_active_storage_attachment @drop_off.document_bundle
+    end
+
+    private
+
+    def set_drop_off
+      @drop_off = IntakeSiteDropOff.find(params[:id])
+    end
+
+    def zendesk_ticket_id
+      @drop_off.zendesk_ticket_id
+    end
+  end
+end
--- a/app/controllers/zendesk/tickets_controller.rb
+++ b/app/controllers/zendesk/tickets_controller.rb
@@ -3,6 +3,7 @@ module Zendesk
     include ZendeskAuthenticationControllerHelper
 
     before_action :require_zendesk_user, :require_ticket_access
+    helper_method :drop_off_upload_date
 
     layout "admin"
 
@@ -13,9 +14,15 @@ module Zendesk
         zendesk_instance_domain: EitcZendeskInstance::DOMAIN
       )
       @document_groups = DocumentPresenter.grouped_documents(@intakes)
+
+      @drop_offs = IntakeSiteDropOff.where(zendesk_ticket_id: zendesk_ticket_id)
     end
 
     private
+
+    def drop_off_upload_date(drop_off)
+      "#{ActionController::Base.helpers.time_ago_in_words(drop_off.created_at)} ago"
+    end
 
     def zendesk_ticket_id
       params[:id]
--- a/app/services/zendesk_drop_off_service.rb
+++ b/app/services/zendesk_drop_off_service.rb
@@ -1,5 +1,6 @@
 class ZendeskDropOffService
   include ZendeskServiceHelper
+  include Rails.application.routes.url_helpers
 
   # Group IDs
   ORGANIZATION_GROUP_IDS = {
@@ -20,7 +21,7 @@ class ZendeskDropOffService
     EitcZendeskInstance
   end
 
-  def create_ticket_and_attach_file
+  def create_ticket
     zendesk_user_id = find_or_create_end_user(@drop_off.name, @drop_off.email, @drop_off.standardized_phone_number)
     ticket = build_ticket(
       subject: @drop_off.name,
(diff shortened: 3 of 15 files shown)


app/controllers/zendesk/tickets_controller.rb
```diff
@@ -13,10 +14,16 @@ def show
         zendesk_instance_domain: EitcZendeskInstance::DOMAIN
       )
       @document_groups = DocumentPresenter.grouped_documents(@intakes)
+
+      @drop_offs = IntakeSiteDropOff.where(zendesk_ticket_id: zendesk_ticket_id)
```
jenny-heath, line 18: we considered modifying the `DocumentPresenter` to handle both types of records but it seemed like overkill since preparers are the ones uploading documents for drop off, all the extra info is probably less helpful. let us know if the code is too repetitive as a result!

app/controllers/zendesk/tickets_controller.rb
```diff
@@ -13,10 +14,16 @@ def show
...
     end
 
     private
 
+    def drop_off_upload_date(drop_off)
```
jenny-heath, line 23: exhibit A of not being totally DRY

spec/controllers/zendesk/tickets_controller_spec.rb
```diff
@@ -23,61 +20,97 @@
...
-                                             id: intake.id, filename: intake.consent_pdf_filename))
-          expect(response.body).to include(zendesk_document_path(id: id_document.id))
-          expect(response.body).to include(zendesk_document_path(id: w2_document.id))
-
-          expect(response.body).to include(pdf_zendesk_intake_path(id: duplicate_intake.id, filename: duplicate_intake.intake_pdf_filename))
-          expect(response.body).to include(consent_pdf_zendesk_intake_path(
-                                             id: duplicate_intake.id, filename: duplicate_intake.consent_pdf_filename))
-          expect(response.body).to include(zendesk_document_path(id: duplicate_id_document.id))
-          expect(response.body).to include(zendesk_document_path(id: duplicate_w2_document.id))
+        context "with duplicate linked intakes" do
```
jenny-heath, line 44: doesn't seem to handle new indentation very well (none of this changed)

spec/services/zendesk_drop_off_service_spec.rb
```diff
@@ -32,68 +32,22 @@
...
-              EitcZendeskInstance::SIGNATURE_METHOD => drop_off.signature_method,
-            }
-          ]
-        }
-      )
-      expect(result).to eq 2
-      expect(comment_uploads.first[:filename]).to eq "GaryGuava.pdf"
-      expect(fake_zendesk_ticket).to have_received(:save)
-    end
+    context "successfully creating ticket" do
```
jenny-heath, line 41: same with this :/

app/controllers/zendesk/drop_offs_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    private
+
+    def set_drop_off
+      @drop_off = IntakeSiteDropOff.find(params[:id])
+    end
+
+    def zendesk_ticket_id
+      @drop_off.zendesk_ticket_id
```
bengolder, line 19: we likely need the same Zendesk instance protection that we use on the documentscontroller. Moreso on this one if anything because there are more older tickets.

bengolder: but if we haven't stored the instance then don't worry about it.

app/controllers/zendesk/tickets_controller.rb
bengolder: I think that's fine, we can add if needed later. Good call.

app/controllers/zendesk/tickets_controller.rb
bengolder: makes sense for now

app/services/zendesk_drop_off_service.rb
```diff
@@ -37,15 +38,26 @@ def create_ticket_and_attach_file
           EitcZendeskInstance::SIGNATURE_METHOD => @drop_off.signature_method
       }
     )
-    attach_file_and_save_ticket(ticket)
+    success = ticket.save
+    raise ZendeskServiceHelper::ZendeskAPIError.new("Error creating drop off ticket: #{ticket.errors}") unless success
```
bengolder, line 42: We can just use `save!` to let the Zendesk Ruby API raise a good specific error, and then we wouldn't have to check for success.

app/services/zendesk_drop_off_service.rb
```diff
@@ -37,15 +38,26 @@ def create_ticket_and_attach_file
...
+
+    ticket.fields = { EitcZendeskInstance::LINK_TO_CLIENT_DOCUMENTS => zendesk_ticket_url(id: ticket.id) }
+    update_success = ticket.save
+    raise ZendeskServiceHelper::ZendeskAPIError.new(
+      "Error updating drop off ticket document link: #{ticket.errors}"
+    ) unless update_success
```
bengolder, line 48: Same, we can use `save!`

app/services/zendesk_drop_off_service.rb
```diff
@@ -37,15 +38,26 @@ def create_ticket_and_attach_file
...
 
   def append_to_existing_ticket
     ticket = ZendeskAPI::Ticket.find(client, id: @drop_off.zendesk_ticket_id)
     ticket.comment = { body: comment_body }
 
-    attach_file_and_save_ticket(ticket)
+    success = ticket.save
+    raise ZendeskServiceHelper::ZendeskAPIError.new("Error updating drop off ticket: #{ticket.errors}") unless success
+
+    success
```
bengolder, line 60: I think you can replace all of this with just `ticket.save!`

app/services/zendesk_drop_off_service.rb
bengolder: https://github.com/zendesk/zendesk_api_client_rb/wiki/Errors#resources

app/controllers/zendesk/drop_offs_controller.rb
jenny-heath: unfortunately we have not stored the instance (probably because UWTSA had their own system for valet)

PR #428, opened 2020-08-06
Remove deprecated credentials file

bengolder:
Just some leftover cleanup from our prior work separating credentials.


bengolder:
That test failure doesn't make any sense to me. I'm rerunning the workflow in case it is a bizarre fluke

bengolder:
@alaroia-cfa I think it was a fluke. The error made no sense—circle doesn't even use the credentials file

alaroia-cfa:
@bengolder I see that the tests now pass -- my approval remains the same.

PR #430, opened 2020-08-07
add a new, shorter, messaging variant to the diy page for testing

loumoore:


loumoore:
yup it's tested. there's just one knew else statement basically, with a new set of copy to test, and for each test there's a new source for facebook vs adwords since we're turning FB back on.

PR #432, opened 2020-08-07
Don't send 'Unknown' as time zone to Zendesk [#174120762]

alaroia-cfa:
This would close https://www.pivotaltracker.com/story/show/174120762

### DIFF
--- a/spec/services/zendesk_intake_service_spec.rb
+++ b/spec/services/zendesk_intake_service_spec.rb
@@ -67,6 +67,11 @@ describe ZendeskIntakeService do
 
   after do
     I18n.locale = I18n.default_locale
+    DatadogApi.instance_variable_set("@dogapi_client", nil)
+    DatadogApi.configure do |c|
+      c.env = "test"
+    end
+    allow(Rails).to receive(:env).and_return("test".inquiry)
   end
 
   describe "#client" do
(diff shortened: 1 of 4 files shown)


spec/services/zendesk_intake_service_spec.rb
```diff
@@ -67,6 +67,11 @@
 
   after do
     I18n.locale = I18n.default_locale
+    DatadogApi.instance_variable_set("@dogapi_client", nil)
+    DatadogApi.configure do |c|
+      c.env = "test"
+    end
+    allow(Rails).to receive(:env).and_return("test".inquiry)
```
bengolder, line 74: Thanks for moving this up!

spec/services/zendesk_intake_service_spec.rb
```diff
@@ -133,9 +138,9 @@
 
           it "passes the email address and not phone" do
             expect(service).to have_received(:create_or_update_zendesk_user).with(hash_including(
-              email: "cash@raining.money",
-              phone: nil,
-            ))
+                                                                                    email: "cash@raining.money",
```
alaroia-cfa, line 141: Looks like I introduced whitespace tragedies -- oops. I'll let it be for now.

PR #433, opened 2020-08-10
Add DIY intake export & create model to store these extracts

alaroia-cfa:
This pull request does very little, and since in the Pivotal Tracker story <https://www.pivotaltracker.com/story/show/173933108> we don't need a web UI for showing these, I think this is enough to get going.

Eager for feedback/review.

### DIFF
--- a/app/models/anonymized_diy_intake_csv_extract.rb
+++ b/app/models/anonymized_diy_intake_csv_extract.rb
@@ -1,0 +1,13 @@
+# == Schema Information
+#
+# Table name: anonymized_diy_intake_csv_extracts
+#
+#  id           :bigint           not null, primary key
+#  record_count :integer
+#  run_at       :datetime
+#  created_at   :datetime         not null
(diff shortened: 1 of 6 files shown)


bengolder:
This is fine as an incremental commit, but the end goal would be for this to produce a CSV that shows up in `AnonymizedIntakeCsvExtractsController` which should perhaps just be `AnonymizedCsvExtractsController`.

alaroia-cfa:
@bengolder I'll push this, but I think I'm confused about the requirements. I think the story says all we need do is pull one CSV and get it to the right person, so I don't know that we really need UI if we can accomplish it from `rails c`.

alaroia-cfa:
And @bengolder thanks for the review!

alaroia-cfa:
@bengolder Oh... maybe you didn't strictly approve it as a review.  Hmm. Oops. Well, I'll chat with you elsewhere then.

PR #434, opened 2020-08-11
Add job for running anonymized diy intake csv service

jenny-heath:
Re: canceling the job if it causes a problem - my research so far has not yielded many promising results. It doesn't look like ActiveJob gives you an easy way to stop a running job, but there are a number of gems that may. See [this thread (from 2014 - not ideal)](https://github.com/rails/rails/issues/17054) and this [stack overflow question about Resque](https://stackoverflow.com/questions/20097568/resque-jobs-how-to-stop-running-job)

### DIFF
--- a/app/services/anonymized_diy_intake_csv_service.rb
+++ b/app/services/anonymized_diy_intake_csv_service.rb
@@ -22,7 +22,7 @@ class AnonymizedDiyIntakeCsvService
     if @ids
       DiyIntake.where(id: @ids)
     else
-      DiyIntake.all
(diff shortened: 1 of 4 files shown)


app/services/anonymized_diy_intake_csv_service.rb
```diff
@@ -22,7 +22,7 @@ def records
     if @ids
       DiyIntake.where(id: @ids)
     else
-      DiyIntake.all
+      DiyIntake.where.not(ticket_id: nil)
```
bengolder, line 25: thanks, this is better!

PR #439, opened 2020-08-19
Truncate data before adding it to the session [#174371407]

alaroia-cfa:
I wasn't sure if it was really important to test all 3 ways `source` data comes into the app in the test suite, so I only tested one. Happy to have feedback.


alaroia-cfa:
Closing in favor of f7d404a461966dbb2f9b411c382a8036ad9a21ca (which I paired with bgolder on)

PR #440, opened 2020-08-19
Overwrite full intake in session at start

jenny-heath:
- fixes bug where client starts EIP and continues to full intake and gets stuck
- hard code file with help next_path
- create new intake every time full intake is started (backtaxes controller)

Co-authored-by: Dimitri DeFigueiredo <ddefigueiredo@codeforamerica.org>

### DIFF
--- a/app/controllers/questions/backtaxes_controller.rb
+++ b/app/controllers/questions/backtaxes_controller.rb
@@ -5,7 +5,7 @@ module Questions
     layout "question"
 
     def current_intake
-      super || Intake.new
+      Intake.new
     end
 
     private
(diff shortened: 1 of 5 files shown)


app/controllers/questions/backtaxes_controller.rb
```diff
@@ -19,28 +19,30 @@ def check_for_triage
...
-    # to intake
-    def assign_triage_source
+    # looks for a triage source and attaches to intake
+    def assign_triage_source(intake)
       return unless session.key?(:triage_source_id) && session.key?(:triage_source_type)
 
       triage_type = session.delete(:triage_source_type).constantize
       triage_source = triage_type.find(session.delete(:triage_source_id))
-      current_intake.update_attribute(:triage_source, triage_source)
+      intake.update_attribute(:triage_source, triage_source)
```
bengolder, line 37: Nice! allows for different types of intakes if needed. thanks!

spec/controllers/questions/file_with_help_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      before do
+        session[:intake_id] = intake.id
+      end
+
+      it "returns the backtaxes controller" do
+        expect(subject.next_path).to eq backtaxes_questions_path
+      end
+    end
+  end
+end
```
bengolder, line 23: excellent, thanks for spelling this out so clearly in the spec!

bengolder:
One last thing I forgot to check: do we have a unit test for backtaxes asserting that with an existing intake the session, we would overwrite it?

ddefigueiredo:
@bengolder do mean something other than [this test](https://github.com/codeforamerica/vita-min/blob/4a30fd59f31a8bdef759e78c66e49d04efa63e9d/spec/controllers/questions/backtaxes_controller_spec.rb#L45)?

spec/controllers/questions/backtaxes_controller_spec.rb
```diff
@@ -20,19 +20,36 @@
...
+          expect(intake.source).to eq "source_from_session"
+          expect(intake.referrer).to eq "referrer_from_session"
+          expect(intake.locale).to eq "en"
+          expect(intake.needs_help_2017).to eq "no"
+          expect(intake.needs_help_2018).to eq "yes"
+          expect(intake.needs_help_2019).to eq "yes"
+        end
+      end
+
+      context "with an existing intake in the session" do
```
jenny-heath, line 40: @bengolder this is the test that checks that the backtaxes controller overwrites an existing intake (dimitri also linked to it i think)

PR #441, opened 2020-08-19
Add intake & return status times & EIP status return times to anonymized CSV download

alaroia-cfa:
Resolves https://www.pivotaltracker.com/story/show/173933593

Partially-co-authored-by: Ben Golder <bgolder@codeforamerica.org>

### DIFF
--- a/app/services/anonymized_intake_csv_service.rb
+++ b/app/services/anonymized_intake_csv_service.rb
@@ -48,7 +48,7 @@ class AnonymizedIntakeCsvService
   end
 
   def generate_csv
-    CSV.generate(headers: CSV_HEADERS, write_headers: true) do |csv|
+    CSV.generate(headers: csv_headers, write_headers: true) do |csv|
       intakes.find_each(batch_size: 100) do |intake|
         csv << csv_row(decorated_intake(intake))
       end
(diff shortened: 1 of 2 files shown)


app/services/anonymized_intake_csv_service.rb
```diff
@@ -80,8 +80,39 @@ def decorated_intake(intake)
...
+        intake_status_timestamps[ticket_status.intake_status] = ticket_status.created_at
+      end
+
+      unless return_status_timestamps.key?(ticket_status.return_status)
+        return_status_timestamps[ticket_status.return_status] = ticket_status.created_at
+      end
+
+      unless eip_status_timestamps.key?(ticket_status.eip_status)
+        eip_status_timestamps[ticket_status.eip_status] = ticket_status.created_at
+      end
```
jenny-heath, line 108: i may be confused about something here - could there be multiple ticket statuses of a given type? if so, are you just recording the first one encountered? why?

spec/services/anonymized_intake_csv_service_spec.rb
```diff
@@ -45,6 +45,127 @@
...
+          intake_status: EitcZendeskInstance::INTAKE_STATUS_IN_REVIEW,
+        )
+      end
+
+      it "stores timestamps in the appropriate column" do
+        row_1_csv_data = csv[0]
+        expect(row_1_csv_data["Intake Status - Gathering Documents"]).to eq("2020-01-01 00:00:00 UTC")
+        expect(row_1_csv_data["Intake Status - In Review"]).to eq("2020-01-02 00:00:00 UTC")
+        row_2_csv_data = csv[1]
+        expect(row_2_csv_data["Intake Status - Gathering Documents"]).to be_nil
```
jenny-heath, line 83: it seems that in my previous comment i was wrong but i can't quite tell why...

app/services/anonymized_intake_csv_service.rb
alaroia-cfa: Good question. I'll push a follow-up commit to this PR that hopefully answers this. Happy to discuss.

spec/services/anonymized_intake_csv_service_spec.rb
alaroia-cfa: The answer is in `verified_change: false` on line 67

https://github.com/codeforamerica/vita-min/pull/441/files#diff-9b3aa8c374ae5eda5bb4a7e87050bb17R67

Ben told me this is important, but I can't quite figure out why, so I'm going to retain the behavior and document in a follow-up commit.

PR #442, opened 2020-08-19
Update new DIY ticket comment with related EIP intakes

jenny-heath:
- previously only specified related "full service" intakes when creating new DIY ticket

### DIFF
--- a/spec/services/zendesk_diy_intake_service_spec.rb
+++ b/spec/services/zendesk_diy_intake_service_spec.rb
@@ -90,14 +90,19 @@ describe ZendeskDiyIntakeService do
       expect(body).to include "send_diy_confirmation"
     end
 
-    context "with corresponding full service tickets" do
+    context "with corresponding intake tickets" do
       let(:intake_ticket_map) do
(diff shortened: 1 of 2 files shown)


spec/services/zendesk_diy_intake_service_spec.rb
```diff
@@ -90,17 +90,22 @@
...
-    context "with corresponding full service tickets" do
+    context "with corresponding intake tickets" do
       let(:intake_ticket_map) do
         { 99998 => double(ZendeskAPI::Ticket, id: "99998"),
           99997 => double(ZendeskAPI::Ticket, id: "99997") }
       end
       let!(:related_intakes) do
         intake_ticket_map.map do |ticket_id, _|
-          create(
+          ticket_id.even? ? create(
```
alaroia-cfa, line 100: This is fine, but a little excessively slick IMHO :)

No action required

jenny-heath: i was being lazy :(

alaroia-cfa: lazy and slick are sorta synonyms; it's not a showstopper

PR #444, opened 2020-08-21
full service landing page

loumoore:
so we can continue to do digital outreach for partners with capacity, in a more effective way than sending people to the homepage

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -80,8 +80,8 @@ en:
   forms:
     errors:
       at_least_one_year: Please pick at least one year.
-      need_phone_number: Please enter a cell phone number.
       need_one_communication_method: Please choose some way for us to contact you.
+      need_phone_number: Please enter a cell phone number.
   general:
     address: Address
     affirmative: 'Yes'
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -485,6 +485,13 @@ en:
...
         subheader: Click below for responses to frequently asked questions. Chat with us if you still have questions!
         title: FAQ
+      full_service_home:
+        header: Your money is waiting!
+        subheader1: Does your family make less than $56,000? You might qualify for a refund of up to $3,500.
+        subheader2: Millions of people qualify for government programs to get through hard times. We're offering a free service to help you get your share.
+        subheader3: Trained volunteers are ready to help you! Our single form signs you up for stimulus checks and tax credits. It takes less than an hour to fill out.
+        subheader4: Don't miss out!
+        title: File taxes with help for free!
       home:
```
loumoore, line 495: these are the only new strings. the other changes are because i ran i18n-tasks normalize, which i thought was the correct procedure, before committing. seems like we haven't always been doing it.

PR #448, opened 2020-08-25
update homepage banner and stimulus page to be clearer about stimulus…

loumoore:
… deadline

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -521,7 +521,7 @@ en:
         file:
           header: Get more of your refund dollars back by filing for free today!
         filing_deadline_banner:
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -521,7 +521,7 @@ en:
         file:
           header: Get more of your refund dollars back by filing for free today!
         filing_deadline_banner:
-          body_html: Unless you anticipate owing money this year, you can still file until October 15th without a penalty.
```
bengolder, line 524: It just occurred to me that this existing language is off. It implies that after October 15th you can incur penalties even if you don't owe money.

PR #455, opened 2020-08-27
Create VITA partners explicitly as needed in tests

alaroia-cfa:
I discovered I need to do this before I could cleanly create better tests for `upsert_vita_partners`. The reason is that `upsert_vita_partners` mutates the set of VITA partners, and other tests depend on the global test-wide state of which VITA partners exist.

I tried to make these changes as minimal as possible toward that goal.

I think this is an improvement. Feedback welcome.


bengolder:
Closing in favor of https://github.com/codeforamerica/vita-min/commit/a85343f012e366f865f283d5c96c55d0f786b982 on which we paired

PR #460, opened 2020-09-02
updates to stimulus FAQ to connect to our EIP flow

loumoore:


app/assets/stylesheets/atoms/_variables.scss
```diff
@@ -14,6 +14,7 @@ $color-lavender: #C2C2FF;
 $color-purple: #3d3de1;
 $color-default-purple: #4c2c92;
 $color-venice-blue: #03578c;
+$color-stimulus-blue: #89DDEF;
```
bengolder, line 17: Non-blocking, but in general it's good to make sure we have a vocabulary with design about colors. Is this what Anu would call it?

app/assets/stylesheets/components/_slabs.scss
```diff
@@ -24,6 +24,10 @@
   }
 }
 
+.slab--stimulus-cta {
+  background-color: $color-stimulus-blue;
+}
+
```
bengolder, line 30: Non-blocking, but for CSS cleanliness, it looks like we could be reusing css classes more effectively here, for example, the slab could use more classes as if they were "mixins", while keeping these css class definition more modular.
```html
<!-- this allows us to keep only one css definition per class -->
<div class="slab slab--cta slab--stimulus">
</div>
```

app/assets/stylesheets/templates/_stimulus.scss
```diff
@@ -1,12 +1,12 @@
...
   .hint {
-    color: #3D3DE1;
+    color: #03578C;
     font-size: 1.6rem;
     font-weight: 600;
     line-height: 2rem;
 
     a {
-      color: #3D3DE1;
+      color: #03578C;
```
bengolder, line 9: Obviously these were already in place without variables, but using color variables would improve maintainability here.

loumoore:
yup i'll squash. here's a screenshot:
image (2)

app/assets/stylesheets/components/_slabs.scss
loumoore: cool i'll change this. i chose not to do this initially because there is exactly one other usage of slab--cta so it seemed like a toss-up.

app/assets/stylesheets/atoms/_variables.scss
loumoore: Anu called it "stimulus blue" :)

app/assets/stylesheets/components/_slabs.scss
loumoore: done!

loumoore:
applied all the suggested code review changes

PR #463, opened 2020-09-08
Use Mailgun for production email sending

alaroia-cfa:


bengolder (approved):
Reviewed with Asheesh

PR #464, opened 2020-09-09
Rename CaseFile to Client

bengolder:


app/controllers/incoming_text_messages_controller.rb
```diff
@@ -3,16 +3,16 @@ class IncomingTextMessagesController < ApplicationController
 
   def create
     phone_number = Phonelib.parse(params["From"]).sanitized
-    case_file = CaseFile.where(phone_number: phone_number).or(CaseFile.where(sms_phone_number: phone_number)).first
-    unless case_file.present?
-      case_file = CaseFile.create!(phone_number: phone_number, sms_phone_number: phone_number)
+    client = Client.where(phone_number: phone_number).or(Client.where(sms_phone_number: phone_number)).first
```
alaroia-cfa, line 6: It'd be cool if the create were atomic. I see that this pull request is just doing a rename, so not a blocker for this pull request. That and we don't usually go to any lengths to make stuff atomic, so doubly not a blocker.

alaroia-cfa (approved):
Great! :ship: 💯 💥 ↗️

PR #465, opened 2020-09-09
Add display of outbound email messages

alaroia-cfa:
## Summary

Very, very incremental pull request to show outbound email messages (not that we actually make any such objects yet).

Also refactors the comms page's tests in a minor way.

Should be a quick review. Happy to answer questions or fix things.

## Screenshot

image

### DIFF
--- a/app/assets/stylesheets/components/_contact-record.scss
+++ b/app/assets/stylesheets/components/_contact-record.scss
@@ -1,6 +1,6 @@
 .contact-record {
   margin-bottom: 2rem;
-  &--outgoing_text_message {
+  &--outgoing_text_message, &--outgoing_email {
    margin-left: 3rem;
     .contact-record__heading {
       padding-left: 3rem;
@@ -18,6 +18,12 @@
       padding-left: 3rem;
       padding-right: 0.75rem;
       background-color: white;
+    }
+    .contact-record__subject::before {
+      content: "✉️ ";
+    }
+    .contact-record__subject {
+      color: $color-grey-dark;
     }
   }
 
--- a/app/models/outgoing_email.rb
+++ b/app/models/outgoing_email.rb
@@ -1,0 +1,42 @@
+# == Schema Information
+#
+# Table name: outgoing_emails
+#
+#  id         :bigint           not null, primary key
+#  body       :string           not null
+#  sent_at    :datetime         not null
+#  subject    :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/components/_contact-record.scss
```diff
@@ -19,6 +19,12 @@
       padding-right: 0.75rem;
       background-color: white;
     }
+    .contact-record__subject::before {
+      content: "✉️ ";
+    }
```
bengolder, line 24: If this is intended as semantic content to indicate that this is an email, I'd put it in the html. If it's purely a visual accent and does not add significant information, it's fine to keep it in a pseudoelement.

app/models/outgoing_email.rb
```diff
@@ -0,0 +1,42 @@
...
+class OutgoingEmail < ApplicationRecord
+  belongs_to :client
+  belongs_to :user
+  validates_presence_of :body
+  validates_presence_of :subject
+  validates_presence_of :sent_at
+
+  def contact_record_type
+    self.class.name.underscore.to_sym
+  end
```
bengolder, line 33: would be helpful to refactor into a shared concern across models. non-blocking

app/models/outgoing_email.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def datetime
+    sent_at
+  end
+
+  def author
+    user.name
+  end
```
bengolder, line 41: this implies we may want to preload all related authors when querying all emails.

spec/controllers/clients_controller_spec.rb
```diff
@@ -115,33 +115,43 @@
       end
 
       context "with existing contact history" do
+        let(:twilio_status) { nil }
         let!(:expected_contact_history) do
           [
             create(:outgoing_text_message, body: "Your tax return is great", sent_at: DateTime.new(2020, 1, 1, 0, 0, 1), client: client, twilio_status: twilio_status),
             create(:incoming_text_message, body: "Thx appreciate yr gratitude", received_at: DateTime.new(2020, 1, 1, 0, 0, 2), client: client),
+            create(:outgoing_email, body: "We are really excited to work with you", sent_at: DateTime.new(2020, 1, 1, 0, 0, 3), client: client),
           ]
```
bengolder, line 124: Note for a way we might improve these expectations, possibly create records in a different order from their expected listing and date. Perhaps create in reverse order and then reverse list?

bengolder (approved):
Awesome, looks great!

PR #466, opened 2020-09-10
Refactor send_message

bengolder:


app/controllers/clients_controller.rb
```diff
@@ -21,17 +21,8 @@ def show
...
-  def send_text
-    outgoing_text_message = OutgoingTextMessage.create(
-      client: Client.find(params[:client_id]),
-      body: params[:body],
-      sent_at: DateTime.now,
-      user: current_user,
-    )
-    SendOutgoingTextMessageJob.perform_later(outgoing_text_message.id)
-    redirect_to client_path(id: params[:client_id])
+    @contact_history = (@client.outgoing_text_messages + @client.incoming_text_messages).sort_by(&:datetime)
```
alaroia-cfa: Seems duplicated with line 22. IMHO you should remove one definition of `@contact_history = ...`

app/controllers/incoming_text_messages_controller.rb
```diff
@@ -1,5 +1,6 @@
 class IncomingTextMessagesController < ApplicationController
   include TwilioRequestable
+  process_twilio_actions
```
alaroia-cfa, line 3: I think web hooks are weird, and they're so weird they deserve to always be in their own controller, and therefore TwilioRequestable's code can & should be simpler and affect an entire controller at a time.

Web hooks are weird: You can't handle them with a regular `:update` Rails method because of the `put` vs. `post` distinction you have to handle in `routes.rb`. Plus, they bypass the Rails anti-cross-site-request-forgery protection, and they have their own authorization checking which is separate from our session handling.

Web hooks' weirdness merits putting them in their own controller. I think this makes code review easier. In the `OutgoingTextMessagesController`, it's not super-duper clear from reading it that `update` does the Twilio authorization check. By contrast, if there were a `TwilioWebhooksController` or `OutgoingTextMessagesWebhooksController`, it'd be crystal clear that 100% of the actions in there would be webhook-style.

If we did that, we could remove `process_twilio_actions` and have easier to review code.

Due to that, I'm -1 on `process_twilio_actions`, but open to discussing more.

app/controllers/outgoing_text_messages_controller.rb
```diff
@@ -1,8 +1,30 @@
 class OutgoingTextMessagesController < ApplicationController
   include TwilioRequestable
+  process_twilio_actions only: [:update]
+  before_action :redirect_to_client_unless_admin, only: :create
+
+  def create
```
alaroia-cfa, line 6: `create` looks great

app/controllers/outgoing_text_messages_controller.rb
```diff
@@ -1,8 +1,30 @@
 class OutgoingTextMessagesController < ApplicationController
   include TwilioRequestable
+  process_twilio_actions only: [:update]
```
alaroia-cfa, line 3: I think `update` having such differences in how it's handled makes it a bad fit to be in the same controller as `create`. Personally, I'd like to see `update` moved out of here.

app/controllers/outgoing_text_messages_controller.rb
```diff
@@ -1,8 +1,30 @@
...
+
+  def create
+    message = OutgoingTextMessage.new(outgoing_text_message_params)
+    if message.save
+      SendOutgoingTextMessageJob.perform_later(message.id)
+    end
+    redirect_to client_path(id: outgoing_text_message_params[:client_id])
+  end
 
   def update
```
alaroia-cfa, line 14: This is a duplicate of `text_status_callback` - if we're refactoring, can we pick one and remove the other?

config/locales/es.yml
```diff
@@ -178,6 +178,11 @@ es:
     file_type: Por favor, sube un tipo de documento válido. Los tipos aceptados incluyen jpg, pdf y png.
     zip: Por favor, introduzca un código postal válido de 5 dígitos.
   views:
+    clients:
```
alaroia-cfa, line 181: IMHO let's either decide to translate all of admin, or punt on that entirely. I'm worried we'll have a half-translated admin site that looks weird.

spec/controllers/outgoing_text_messages_controller_spec.rb
```diff
@@ -1,6 +1,57 @@
...
+
+        outgoing_text_message = OutgoingTextMessage.last
+        expect(outgoing_text_message.body).to eq "This is an outgoing text"
+        expect(outgoing_text_message.client).to eq client
+        expect(SendOutgoingTextMessageJob).to have_been_enqueued.with(outgoing_text_message.id)
+        expect(response).to redirect_to(client_path(id: client.id))
+      end
+    end
+  end
+
```
alaroia-cfa, line 54: Tests are 💯

alaroia-cfa (changes requested):
I think the complication of `process_twilio_actions` makes me conclude we're better-off having Twilio webhook actions not share controllers with non-webhook actions.

The rest looks fine.

Happy to discuss; you could change my mind, but you said you were curious if I thought it was a bad idea, so there you go :P :)

config/locales/es.yml
alaroia-cfa: DECIDED with @bengolder & @jenny-heath that we will i18n-ify all new strings in admin, and we're willing to leave English-language text in `es.yml` to get there fast, and Ben Golder might possibly translate some to Spanish but it's not something we need others to necessarily do.

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -0,0 +1,39 @@
+class TwilioWebhooksController < ApplicationController
```
bengolder, line 1: should we subclass the rails base controller then!!!???

PR #467, opened 2020-09-10
Handle incoming emails

bengolder:


app/assets/stylesheets/components/_contact-record.scss
```diff
@@ -1,5 +1,12 @@
 .contact-record {
   margin-bottom: 2rem;
+
+  &--incoming_text_message, &--outgoing_text_message {
+    .contact-record__body {
+      border-radius: 2rem;
+    }
+  }
```
bengolder, line 8: Just wanted to distinguish sms from email so I could see it easily

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+class MailgunWebhooksController < ApplicationController
+  skip_before_action :redirect_to_getyourrefund
+  skip_before_action :set_visitor_id
+  skip_before_action :set_source
+  skip_before_action :set_referrer
+  skip_before_action :set_utm_state
+  skip_before_action :set_sentry_context
+  skip_before_action :check_maintenance_mode
+  skip_after_action :track_page_view
+  skip_before_action :verify_authenticity_token
```
bengolder, line 12: would love to refactor this inheritance and all these before_actions. Also present in TwilioWebhooksController, and would be useful in ZendeskWebhooks

app/models/incoming_email.rb
```diff
@@ -0,0 +1,47 @@
...
+class IncomingEmail < ApplicationRecord
+  belongs_to :client
+
+  def contact_record_type
+    self.class.name.underscore.to_sym
+  end
+
+  def body
+    stripped_html&.html_safe || stripped_text || body_html&.html_safe || body_plain
+  end
```
bengolder, line 38: this is a little dangerous. Not sure how to better display html content.

spec/controllers/clients_controller_spec.rb
```diff
@@ -118,10 +118,11 @@
...
           [
-            create(:outgoing_text_message, body: "Your tax return is great", sent_at: DateTime.new(2020, 1, 1, 0, 0, 1), client: client, twilio_status: twilio_status),
-            create(:incoming_text_message, body: "Thx appreciate yr gratitude", received_at: DateTime.new(2020, 1, 1, 0, 0, 2), client: client),
+            create(:incoming_email, body_plain: "Me too! Happy to get every notification", received_at: DateTime.new(2020, 1, 1, 0, 0, 4), client: client),
             create(:outgoing_email, body: "We are really excited to work with you", sent_at: DateTime.new(2020, 1, 1, 0, 0, 3), client: client),
-          ]
+            create(:incoming_text_message, body: "Thx appreciate yr gratitude", received_at: DateTime.new(2020, 1, 1, 0, 0, 2), client: client),
+            create(:outgoing_text_message, body: "Your tax return is great", sent_at: DateTime.new(2020, 1, 1, 0, 0, 1), client: client, twilio_status: twilio_status),
+          ].reverse
         end
```
bengolder, line 126: I created these in reverse order to make sure it was sorting by datetime

app/models/incoming_email.rb
bengolder: should I only do plaintext for now?

app/controllers/mailgun_webhooks_controller.rb
bengolder: Should we?
1. do a base WebhookController that skips all the junk
2. have webhook controllers inherit from `ActionController::Base`
3. make a new base controller for all non-webhook controllers that does all these before actions

app/models/concerns/contact_record.rb
```diff
@@ -0,0 +1,7 @@
+module ContactRecord
+  extend ActiveSupport::Concern
```
bengolder, line 2: this line isn't strictly necessary, but adds functionality that could be useful in the future

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  before_action :valid_mailgun_post?
+
+  def create_incoming_email
+    sender_email = params["sender"]
```
alaroia-cfa, line 14: Is this the `Sender: ` email header? Or is it something computed by Mailgun? Or is it the "Envelope From"?

I think the "From:" header is a better here, but happy to discuss.

app/models/concerns/contact_record.rb
```diff
@@ -0,0 +1,7 @@
+module ContactRecord
```
bengolder, line 1: I got tired of naming mixins with the "Something-able" pattern

app/models/incoming_email.rb
alaroia-cfa: I vote just plaintext for now.

The safety depends how we render it in the template. If we're unsure, we can do just plaintext for now and come back to it imho.

app/controllers/clients_controller.rb
```diff
@@ -19,8 +19,12 @@ def create
...
   def show
     @client = Client.find(params[:id])
-    @contact_history = (@client.outgoing_text_messages + @client.incoming_text_messages + @client.outgoing_emails
-                       ).sort_by(&:datetime)
+    @contact_history = (
+      @client.outgoing_text_messages.includes(:user) +
+      @client.incoming_text_messages +
+      @client.outgoing_emails.includes(:user) +
+      @client.incoming_emails
+    ).sort_by(&:datetime)
```
bengolder, line 27: this will prefetch related user objects

app/services/mailgun_service.rb
```diff
@@ -0,0 +1,15 @@
+class MailgunService
+  class << self
+    def valid_post?(params)
+      signing_key = EnvironmentCredentials.dig(:mailgun, :webhook_signing_key)
```
alaroia-cfa, line 4: It'd be nice to have comments about which parts of this were specifically recommended by Mailgun, or which parts we chose for our reasons (and a link to their doc)

app/controllers/mailgun_webhooks_controller.rb
alaroia-cfa: Non-blocking!

alaroia-cfa (approved):

app/controllers/mailgun_webhooks_controller.rb
alaroia-cfa: Jenny says 1 or 2, preferably 2

I say 4 or 2, preferably 4, but 2 is also great

sounds like 2 :P

app/controllers/mailgun_webhooks_controller.rb
bengolder: They document it here: https://documentation.mailgun.com/en/latest/user_manual.html#routes

> sender of the message as reported by `MAIL FROM` during SMTP chat. Note: this value may differ from `From MIME header.

alaroia-cfa: decided, we'll go with sender for now

alaroia-cfa: comment with docs is nice

PR #468, opened 2020-09-11
Small follow-up refactors

bengolder:
- Webhooks inherit from ActionController::Base
- render incoming emails as plaintext
- Clarify mailgun api with comments


alaroia-cfa (approved):
yessss

PR #470, opened 2020-09-15
User invites

bengolder:


bengolder:
This adds translations for both devise and devise-invitable, and as a result it has many unused keys. If we remove the unused keys, we won't be able to use them later for devise when we use more features.
If we ignore the additional translation files, then it will complain that we have missing keys in any views we try to customize.

app/controllers/concerns/access_controllable.rb
```diff
@@ -0,0 +1,14 @@
+module AccessControllable
```
alaroia-cfa, line 1: It's great to have this be its own concern!

app/controllers/users/invitations_controller.rb
```diff
@@ -0,0 +1,53 @@
+class Users::InvitationsController < Devise::InvitationsController
+  include AccessControllable
+
+  # skip inherited methods
```
alaroia-cfa, line 4: Per pairing discussion, I'd like to see a comment clarifying this, maybe

```
# Devise::InvitationsController from devise-invitable uses some before_actions to validate
# data being passed in. We skip those and use our before_action methods to customize
# access control and error messages.
```

app/controllers/users/invitations_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      render :not_found, status: :not_found
+    end
+  end
+
+  def raw_invitation_token
+    return params[:invitation_token] if request.get?
+    update_resource_params[:invitation_token]
+  end
+
+  def get_user_by_invitation_token
```
alaroia-cfa, line 50: Per discussion when pairing, it'd be nice to have a comment about what this is and why it's needed.

app/models/user.rb
```diff
@@ -27,19 +33,29 @@
...
+#  fk_rails_...  (invited_by_id => users.id)
+#
 class User < ApplicationRecord
-  devise :database_authenticatable, :lockable, :validatable, :timeoutable, :trackable,
+  devise :database_authenticatable, :lockable, :validatable, :timeoutable, :trackable, :invitable,
          :omniauthable, omniauth_providers: [:zendesk]
 
   attr_encrypted :access_token, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
 
+  validates_presence_of :name
```
alaroia-cfa, line 57: I think it'd be good to validate there are no `User` records in prod/elsewhere with a blank `name`.

app/views/devise/invitations/edit.html.erb
```diff
@@ -0,0 +1,16 @@
+<% @main_heading = t(".title") %>
```
alaroia-cfa, line 1: This short syntax is so cool.

config/i18n-tasks.yml
```diff
@@ -102,8 +102,8 @@ ignore_unused:
   - 'date.formats.*'
   - 'general.*'
   - 'errors.*'
+  - '{devise,kaminari,will_paginate}.*'
```
alaroia-cfa, line 105: Non-blocking: I'd love to see the `devise` line not added to this `ignore_unused` block.

config/initializers/devise.rb
```diff
@@ -120,6 +120,55 @@
   # Send a notification email when the user's password is changed.
   # config.send_password_change_notification = false
 
+  # ==> Configuration for :invitable
+  # The period the generated invitation token is valid.
+  # After this period, the invited resource won't be able to accept the invitation.
+  # When invite_for is 0 (the default), the invitation won't expire.
+  # config.invite_for = 2.weeks
```
alaroia-cfa, line 127: I'd like to see this be some value by default, say 4 weeks.

config/initializers/devise.rb
```diff
@@ -120,6 +120,55 @@
...
+   config.validate_on_invite = true
+
+  # Resend invitation if user with invited status is invited again
+  # Default: true
+  # config.resend_invitation = false
+
+  # The class name of the inviting model. If this is nil,
+  # the #invited_by association is declared to be polymorphic.
+  # Default: nil
+   config.invited_by_class_name = 'User'
```
alaroia-cfa, line 156: Non-blocking: I think this indented by 3 spaces, but our usual style would have it indented by two.

config/locales/devise_invitable.en.yml
```diff
@@ -0,0 +1,31 @@
+en:
```
alaroia-cfa, line 1: I'd like to see this file go away, since we have its contents in `config/locales/en.yml`.

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+          user: {
+            name: "Cher Cherimoya",
+            password: "secret password",
+            password_confirmation: "secret password",
+            invitation_token: raw_invitation_token
+          }
+        }
+      end
+
+
```
alaroia-cfa, line 111: Non blocking - extra newline here

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+          user: {
+            name: "",
+            password: "secret password",
+            password_confirmation: "secret password",
+            invitation_token: raw_invitation_token
+          }
+        }
+      end
+
+      it "shows a validation error" do
```
alaroia-cfa, line 134: I think this test is not done yet

spec/factories/users.rb
```diff
@@ -27,17 +33,47 @@
...
+
+    factory :agent_user do
+      role { "agent" }
+    end
+
+    factory :invited_user do
+      association :invited_by, factory: :admin_user
+      invitation_created_at { 1.day.ago - 1.minute }
+      invitation_sent_at { 1.day.ago }
+      sequence(:invitation_token) do |n|
```
alaroia-cfa, line 70: Might not need the digest (though I know the tokens have to be different from each other)

alaroia-cfa: Might leave a comment explaining it's to overcome the uniqueness requirement

`# Creating a sequence because `invitation_token` is UNIQUE in the DB`

spec/features/invitations_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      click_on "Invite a new user"
+
+      # new invitation page
+      expect(page).to have_text "Send a new invitation"
+      fill_in "What is their name?", with: "Colleen Cauliflower"
+      fill_in "What is their email?", with: "colleague@cauliflower.org"
+      click_on "Send invitation email"
+
+      # back on the invitations page
+      within(".flash--notice") do
```
alaroia-cfa, line 27: `within()` is really cool and I haven't used it before

spec/support/shared_examples/access_controllable.rb
```diff
@@ -0,0 +1,53 @@
+shared_examples :a_get_action_that_redirects_anonymous_users_to_sign_in do |action:|
```
alaroia-cfa, line 1: Would be nice to have a comment at the top here

```
# These shared_examples use `params` from the place they're invoked (using `{}` by default).
```

PR #472, opened 2020-09-16
Skip document pages when source is 211intake

jenny-heath:
ticket says to make the ID documents not required but the reason for that is that the 211 assisters won't upload any documents. so, i decided to skip all of those pages to save them some time.

### DIFF
--- a/spec/controllers/documents/document_upload_question_controller_spec.rb
+++ b/spec/controllers/documents/document_upload_question_controller_spec.rb
@@ -36,4 +36,30 @@ RSpec.describe Documents::DocumentUploadQuestionController do
       end
     end
   end
+
+  describe ".show?" do
+    before do
+      module TestingClass
+        class ExampleDocumentUploadController < Documents::DocumentUploadQuestionController
+          def self.document_type; end
+        end
+      end
+    end
+
+    context "for intakes with 211 source" do
+      let(:intake) { create :intake, source: "211intake" }
+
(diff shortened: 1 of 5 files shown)


spec/controllers/documents/document_upload_question_controller_spec.rb
```diff
@@ -36,4 +36,24 @@
...
+    context "for intakes with any other source" do
+      let(:intake) { create :intake }
+
+      it "raises an error because it needs the document_type method ot be defined" do
+        expect {
+          described_class.show?(intake)
+        }.to raise_error NotImplementedError
+      end
+    end
+  end
```
bengolder, line 58: I don't understand this test. It does not set another source and it implies raising an error is the correct behavior here, but I don't understand why it is the correct behavior.

bengolder (approved):
Looks good and simple!
I'm surprised that the two changes are all that is need and I don't understand one of the specs, but trust that it works well due to the passing feature spec. If you feel like clarifying that would be helpful but isn't blocking.

spec/controllers/documents/document_upload_question_controller_spec.rb
jenny-heath: it's because this is a test for a controller that is always a parent and expects the inheriting controller to implement `document_type` - asheesh made a good point that it would be better to define a test class instead of just testing the document upload question controller itself so we did that

PR #473, opened 2020-09-17
Password login additions

bengolder:
- Add a sign in link to non-prod environments (in the footer)
- Style and internationalize the sign in form

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,2 +1,14 @@
 class Users::SessionsController < Devise::SessionsController
+  def new
+    super do |user|
+      # When warden and devise threw an authentication error (invalid password)
+      # They don't respect the locale param
+      # This allows us to detect devise's authentication error, enforce internationalization,
+      # and move the error message onto the user object in order to treat it like a validation error
+      if flash[:alert] == "Invalid email or password" # The default devise english failure message
+        flash.delete :alert
+        user.errors[:password] << I18n.t("controllers.users.sessions_controller.new.invalid_email_or_password")
+      end
+    end
+  end
 end
--- a/app/views/shared/_footer.html.erb
+++ b/app/views/shared/_footer.html.erb
@@ -60,7 +60,7 @@
         <% unless Rails.env.production? %>
           <li>
             <%= link_to new_user_session_path do %>
-              <%= t("general.sign_in") %>
+              <%= t("views.shared.footer.volunteer_sign_in") %>
             <% end %>
           </li>
         <% end %>
--- a/app/views/shared/_footer.html.erb
+++ b/app/views/shared/_footer.html.erb
@@ -56,6 +56,14 @@
             <%=t("views.public_pages.privacy_policy.title") %>
           <% end %>
         </li>
+
+        <% unless Rails.env.production? %>
+          <li>
+            <%= link_to new_user_session_path do %>
+              <%= t("general.sign_in") %>
+            <% end %>
+          </li>
+        <% end %>
       </ul>
 
     </div>
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -10,7 +10,7 @@ Devise.setup do |config|
 
   # ==> Controller configuration
   # Configure the parent class to the devise controllers.
-  # config.parent_controller = 'DeviseController'
+  config.parent_controller = 'BaseDeviseController'
 
   # ==> Mailer Configuration
   # Configure the e-mail address which will be shown in Devise::Mailer,
--- a/config/locales/devise.en.yml
+++ b/config/locales/devise.en.yml
@@ -9,10 +9,10 @@ en:
     failure:
       already_authenticated: "You are already signed in."
       inactive: "Your account is not activated yet."
-      invalid: "Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes."
+      invalid: "Invalid email or password"
       locked: "Your account is locked."
       last_attempt: "You have one more attempt before your account is locked."
-      not_found_in_database: "Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes."
+      not_found_in_database: "Invalid email or password"
       timeout: "Your session expired. Please sign in again to continue."
       unauthenticated: "You need to sign in or sign up before continuing."
       unconfirmed: "You have to confirm your email address before continuing."
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -151,6 +151,7 @@ en:
     return_search: Return to search
     search: Search
     security_html: Your information is secure and the information you provide will be handled in accordance with our %{privacy_link}.
+    sign_in: Sign in
     sign_out: Sign out
     skip_question: Skip question
     skip_questions: Skip questions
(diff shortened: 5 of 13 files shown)


app/views/shared/_footer.html.erb
```diff
@@ -56,6 +56,14 @@
...
           <% end %>
         </li>
+
+        <% unless Rails.env.production? %>
+          <li>
+            <%= link_to new_user_session_path do %>
+              <%= t("views.shared.footer.volunteer_sign_in") %>
+            <% end %>
+          </li>
+        <% end %>
```
bengolder, line 66: This allows demo users to use the sign in form

config/initializers/devise.rb
```diff
@@ -10,7 +10,7 @@
 
   # ==> Controller configuration
   # Configure the parent class to the devise controllers.
-  # config.parent_controller = 'DeviseController'
+  config.parent_controller = 'BaseDeviseController'
```
bengolder, line 13: This is necessary to use a different layout for the view

config/locales/devise.en.yml
```diff
@@ -9,10 +9,10 @@ en:
     failure:
       already_authenticated: "You are already signed in."
       inactive: "Your account is not activated yet."
-      invalid: "Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes."
+      invalid: "Invalid email or password"
       locked: "Your account is locked."
       last_attempt: "You have one more attempt before your account is locked."
-      not_found_in_database: "Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes."
+      not_found_in_database: "Invalid email or password"
```
bengolder, line 15: Restoring these to their original value.

config/routes.rb
```diff
@@ -106,7 +106,9 @@ def scoped_navigation_routes(context, navigation)
     resources :clients, only: [:show, :create]
     resources :outgoing_text_messages, only: [:create]
     resources :outgoing_emails, only: [:create]
-    devise_for :users, skip: :omniauth_callbacks
+    devise_for :users, skip: :omniauth_callbacks, controllers: {
+      sessions: "users/sessions"
+    }
```
bengolder, line 111: This allows us to user our own child controller

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -1,4 +1,32 @@
 require "rails_helper"
 
 RSpec.describe Users::SessionsController do
+  before do
+    request.env["devise.mapping"] = Devise.mappings[:user]
+  end
+
+  describe "#new" do
+    context "when the locale is not english and an invalid login alert is added by warden and the devise failure app" do
```
bengolder, line 9: Two problems with the sign in page:
1. It would only display the invalid login error message as a flash message.
2. It would not respect the locale for the invalid login error message.

This test ensures that we correct both of those issues.

config/initializers/devise.rb
alaroia-cfa: Cool!

config/locales/devise.en.yml
alaroia-cfa: I liked our customizations! :) Why reset them? If in support of the string compare below, well, OK, I guess that's life.

spec/controllers/users/sessions_controller_spec.rb
alaroia-cfa: For the locale, I'd prefer to see us figure out how to layer our locale data through so that Warden/Devise can find it, rather than render the template twice. It doesn't have to be a blocking request on this PR.

Good test either way.

app/controllers/users/sessions_controller.rb
```diff
@@ -1,2 +1,13 @@
 class Users::SessionsController < Devise::SessionsController
+  def new
```
alaroia-cfa, line 2: This is a little hacky... but OK.

I'd prefer to see a less hacky approach, but maybe that's not possible? Not blocking.

app/views/shared/_footer.html.erb
alaroia-cfa: ✨

config/locales/devise.en.yml
alaroia-cfa: Never mind -- I see our customizations live on below.

Strange to have the same key in both files. I guess the idea is that our definition shadows this definition. I'd prefer to only have the key in one file. Non-blocking.

app/controllers/users/sessions_controller.rb
```diff
@@ -1,2 +1,13 @@
...
+    super do |resource|
+      # When warden and devise threw an authentication error (invalid password)
```
alaroia-cfa, line 4: One alternative I can suggest:

https://stackoverflow.com/a/32655614 remarks that we could delete _all_ flash messages for new. I think that's better. What do you think? (Non-blocking.)

In fact, our Users::SessionsController could remove _all_ flash messages in general.

alaroia-cfa:
I think this is good and we could merge it! Feel free to do so!

I have some non-blocking questions and suggestions. Cheers!

alaroia-cfa (approved):

spec/controllers/users/sessions_controller_spec.rb
bengolder: Rendering the template twice? I don't see where that is happening.

app/controllers/users/sessions_controller.rb
```diff
@@ -1,2 +1,13 @@
...
+      # They don't respect the locale param
+      # This allows us to detect, enforce internationalization, and move the error message
+      if flash[:alert] == I18n.t("devise.failure.invalid", locale: :en) # The default english failure message
+        flash.delete :alert
+      end
+      resource.errors[:password] << I18n.t("devise.failure.invalid")
```
alaroia-cfa, line 10: This should be inside the conditional

alaroia-cfa: It'd be easier for me to understand if we use a different translation key for our message. Then conceptually,  we're looking for their string's presence & replacing it with our entirely different thing.

app/controllers/users/sessions_controller.rb
```diff
@@ -1,2 +1,13 @@
 class Users::SessionsController < Devise::SessionsController
+  def new
+    super do |resource|
+      # When warden and devise threw an authentication error (invalid password)
+      # They don't respect the locale param
+      # This allows us to detect, enforce internationalization, and move the error message
```
alaroia-cfa, line 6: Would love a comment we're looking for their message to detect the authentication error

config/locales/en.yml
```diff
@@ -16,6 +16,9 @@ en:
     formats:
       default: "%-m/%-d/%Y"
       long: "%B %-d, %Y"
+  devise:
+    failure:
+      invalid: Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes.
```
alaroia-cfa, line 21: Per `sessions_controller.rb` discussion, new keys here

spec/controllers/users/sessions_controller_spec.rb
alaroia-cfa: I was wrong, ignore this; thought we were re-executing the controller

alaroia-cfa (approved):

PR #474, opened 2020-09-17
Use Rails.logger explicitly

alaroia-cfa:
This simplifies ConsolidatedTraceHelper -- it's required because in 55009c8e410e7a53e255e97d089043f544d6e906 , we didn't realize we needed a `def logger` on `UnsentDocumentsService`.

Rather than add it, this proposes how it can be not required.

p.s. `ConsolidatedTraceHelper` != `ConsolidatedTranceHelper`


alaroia-cfa:
Closing in favor of 0952e84783dbb7965f58e30c35ea467ce50b2ded

PR #478, opened 2020-09-19
fix locale tracking in mixpanel

loumoore:
i verified the new test failed before the change and passed with the change, also stepped through it in the debugger.

filters were out of order, resulting in locale always being "en" in mixpanel


bengolder (approved):
Wow, thank you! Such an easy fix for something that was messing up a lot of data! Great catch Lou 🙏

PR #480, opened 2020-09-20
Work situations

loumoore:
consolidation of work related questions


loumoore:
i can squash on merge. forgot to re-run tests after making some copy tweaks.

app/views/questions/work_situations/edit.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <%= content_for(:page_title) %>
+    </h1>
+    <p>
+      <%=t('views.questions.work_situations.help_text') %>
+    </p>
+    <div class="form-card__stacked-checkboxes spacing-above-0">
+      <%= f.cfa_checkbox(:had_wages, t('views.questions.work_situations.options.had_wages_html', :count => current_intake.filing_joint_yes? ? 2 : 1), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:had_self_employment_income, t('views.questions.work_situations.options.had_self_employment_income_html', :count => current_intake.filing_joint_yes? ? 2 : 1), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:had_unemployment_income, t('views.questions.work_situations.options.had_unemployment_income_html', :count => current_intake.filing_joint_yes? ? 2 : 1), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:had_tips, t('views.questions.work_situations.options.had_tips', :count => current_intake.filing_joint_yes? ? 2 : 1), options: { checked_value: "yes", unchecked_value: "no" }) %>
```
bengolder, line 15: Would it be cleaner to add a new method to the Intake class?

```ruby
def filer_count
  current_intake.filing_joint_yes? ? 2 : 1
end
```

```ruby
t('views.questions.work_situations.options.had_wages_html', :count => current_intake.filer_count)
```

config/locales/en.yml
```diff
@@ -1218,6 +1218,24 @@ en:
...
         title: What was the year of your spouse's death?
+      work_situations:
+        help_text: Select all the options that apply.
+        options:
+          had_self_employment_income_html:
+            one: I was self-employed or worked as an independent contractor.<br><span class="checkbox-help">Collected a 1099-K form, 1099-MISC form, or Personal log</span>
+            other: My spouse or I was self-employed or worked as an independent contractor.<br><span class="checkbox-help">Collected a 1099-K form, 1099-MISC form, or Personal log</span>
+          had_tips:
+            one: I collected tips at work not included in my W-2
+            other: My spouse or I collected tips at work not included in a W-2
```
bengolder, line 1229: I like this use of the one/other functionality for joint filers!

app/lib/question_navigation.rb
```diff
@@ -85,12 +85,17 @@ class QuestionNavigation
...
     Questions::OtherStatesController,
+
+    # TODO: Old controllers being consolidated, to be deleted after WorkSituationsController is deployed
     Questions::WagesController,
     Questions::SelfEmploymentController,
     Questions::TipsController,
+    Questions::UnemploymentIncomeController,
+
+    # Work Situations
+    Questions::WorkSituationsController,
```
bengolder, line 96: Thanks for making the deploy less likely to disrupt a user!

app/controllers/questions/work_situations_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Questions
+  class WorkSituationsController < QuestionsController
+    def self.show?(intake)
+      intake.had_a_job?
+    end
```
bengolder, line 5: Does this have a unit test?

bengolder: None of the other controllers used this criteria to determine whether to show themselves, as far as I can tell. Did we trule want to gate all these questions on whether they indicated having 1 or more jobs?

bengolder: This would mean if someone said they didn't have a job, then we wouldn't ask them if they have unemployment income. It doesn't seem correct.

bengolder (changes requested):
Overall looks great, but I think the new logic for when to ask these questions should have a unit test and doesn't seem correct to me. Thanks Lou!

app/views/questions/work_situations/edit.html.erb
loumoore: yeah, i can refactor that. used widely though not just here, so will expand the ticket. maybe as a follow-up?

app/controllers/questions/work_situations_controller.rb
loumoore: i'll discuss your last point with Anu. this was the intent of the design, but sounds like it might not be right!

loumoore: thought i worked in a test too but will double check. might need to change this logic anyway now per your last point.

loumoore: the unit test was in work_situations_controller_spec, but you were correct about the logic issue with the unemployment question so i'm submitting a change that will remove the skip logic (and the test)

app/views/questions/work_situations/edit.html.erb
bengolder: That sounds good. I didn't realize it was a pattern used elsewhere.

bengolder (approved):
Thanks Lou!

PR #481, opened 2020-09-21
Add Alex and Sarah to git-pair

sniemeyer13:
Had to change email format because our domain is different

### DIFF
--- a/.pairs
+++ b/.pairs
@@ -1,17 +1,27 @@
 # .pairs - configuration for 'git pair'
 pairs:
   # <initials>: <Firstname> <Lastname>[; <email-id>]
-  jh: Jenny Heath; jheath
-  bg: Ben Golder; bgolder
-  td: Tom Dooner; tdooner
-  lk: Laura Kogler; lkogler
-  mtm: Molly T-M; mollyt
-  jg: Jonathan Greenberg; jgreenberg
-  bv: Ben Vandgrift; bvandgrift
-  dbd: Dimitri DeFigueiredo; ddefigueiredo
-  al: Asheesh Laroia; alaroia
-email:
-  prefix: pair
-  domain: codeforamerica.org
-  no_solo_prefix: true
+  jh: Jenny Heath
+  bg: Ben Golder
+  td: Tom Dooner
+  lk: Laura Kogler
+  mtm: Molly T-M
+  jg: Jonathan Greenberg
+  bv: Ben Vandgrift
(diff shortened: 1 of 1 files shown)


.pairs
```diff
@@ -1,17 +1,27 @@
...
-  prefix: pair
-  domain: codeforamerica.org
-  no_solo_prefix: true
+  jh: Jenny Heath
+  bg: Ben Golder
+  td: Tom Dooner
+  lk: Laura Kogler
+  mtm: Molly T-M
+  jg: Jonathan Greenberg
+  bv: Ben Vandgrift
```
jenny-heath, line 10: nonblocking but molly, jonathan, and ben vandgrift can all be removed from here if you want to make that change as well!

.pairs
```diff
@@ -1,17 +1,27 @@
...
+  dbd: Dimitri DeFigueiredo
+  al: Asheesh Laroia
+  sn: Sarah Niemeyer
+  as: Alex Sartan
+email_addresses:
```
jenny-heath, line 15: i imagine there's some documentation i haven't read that indicates `email` -> `email_addresses` works - if you were able to duet-commit then i think this looks good!

jenny-heath (approved):

.pairs
```diff
@@ -1,17 +1,27 @@
...
-  lk: Laura Kogler; lkogler
-  mtm: Molly T-M; mollyt
-  jg: Jonathan Greenberg; jgreenberg
-  bv: Ben Vandgrift; bvandgrift
-  dbd: Dimitri DeFigueiredo; ddefigueiredo
-  al: Asheesh Laroia; alaroia
-email:
-  prefix: pair
-  domain: codeforamerica.org
-  no_solo_prefix: true
```
jenny-heath, line 16: hm actually sorry for the extra comment but i'm not sure whether we still need prefix/no_solo_prefix ?

sniemeyer13: It's interesting because in this new format, it choses one author at random for the email address. (because there are different domains now, rather than putting "pair+david+sam@codeforamerica.org" it will simply chose "david@codeforamerica.org" or "sam@codeforamerica.org"). It will, however, put both members as the commit author. So we shouldn't need the no_solo_prefix anymore.

more details.. https://github.com/pivotal-legacy/git_scripts#git-pair-commit

.pairs
sniemeyer13: Got it!

PR #488, opened 2020-09-24
Move TicketStatus updates into job

bengolder:
I wasn't able to determine how to implement Asheesh's suggestion to limit these jobs to running one at a time.

I greatly reduced the detail of the mixpanel event tests in here. They were testing method calls across two other classes and it did not feel appropriate in a unit test. They are still reaching across one class and testing the way that ticket status calls its own instance method. I feel that our mixpanel service needs an overhaul to simplify it. I didn't want to address that in this PR, but I think we should address it before we add any significant amount of new analytics.


alaroia-cfa (approved):
w/r/t running them in serial:

It looks to me from e.g. https://stackoverflow.com/questions/9479318/delayed-job-multiple-parallel-queues that Delayed::Job always runs tasks in series! The consensus is you have to run multiple workers to run jobs in parallel. We only have 1 worker in prod, so that's that.

This is so great. So excited to see this happen as a background job.

Reviewed w/ @sartan

PR #489, opened 2020-09-24
Index ticket id, email, phones on Intakes

bengolder:
Our queries that use up the most time are those that look up intakes based on their zendesk ticket ids, email, or phone numbers.
This should speed up those queries a bit.


alaroia-cfa (approved):
Looks great; reviewed with @sartan 

Thanks for adding these indices. I'm hopeful it'll speed up the app without us doing any real thinking :D

PR #490, opened 2020-09-24
Add beta testers

bengolder:
Replaces all the admin-only restrictions with beta tester only restrictions.

Any users invited through the new interface will be marked as beta testers, including existing users.


alaroia-cfa (approved):
Hi @bengolder !

This is great. Glad to separate the idea of beta from admin!

Alex and I changed some test descriptions -- normally we would leave a comment in GitHub, but I got so excited when doing a review in RubyMine that I just made the changes on this branch.

Please feel free to merge!!!

PR #491, opened 2020-09-24
add gating question on sale of assets (stocks, bonds, real estate)

loumoore:


config/locales/en.yml
```diff
@@ -13,9 +13,6 @@ en:
       sessions_controller:
         new:
           invalid_email_or_password: Incorrect email or password. After 5 login attempts, accounts are locked for 30 minutes.
-  email:
-    user_message:
-      subject: "Update from GetYourRefund"
```
bengolder, line 18: Why did this get deleted? We'd have thought that any unused keys would be caught by the i18n specs already

app/controllers/questions/asset_sale_gate_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Questions
+  class AssetSaleGateController < TicketedQuestionsController
```
bengolder, line 2: Might be more consistent to just name this `AssetSaleController` or `SoldAssetsController` to be consistent with the rest of the controllers. We don't have any other explicit `Gate` controllers.

bengolder (approved):
Approved! with one small suggested non-blocking improvement.

app/controllers/questions/asset_sale_gate_controller.rb
loumoore: would love thoughts on naming convention as there will be more gate controllers, but still not sure if it'll be clearer to folks if they are explicitly named "*Gate" or if I try to come up with a descriptive name for each. Designs [here]([https://www.figma.com/file/KhjUvgH13LaUjOHR8sTuRm/Get-Your-Refund?node-id=3503%3A0])

loumoore: went ahead and did this, refactored all the things (controller + form) from AssetSalesGate to SoldAssets

PR #492, opened 2020-09-24
add 211 hotline notice to EIP overview page

loumoore:


app/views/questions/eip_overview/edit.html.erb
```diff
@@ -26,7 +26,14 @@
       </li>
     </ul>
 
-    <button class="button button--primary button--wide text--centered spacing-above-60" name="eip_overview_form[eip_only]" value="true" type="submit">
+    <div class="notice--warning text--small">
+      <p class="spacing-below-25 text--bold"><%=t("views.questions.eip_overview.hotline_notice.title") %></p>
+      <p><%=t("views.questions.eip_overview.hotline_notice.what") %></p>
+      <p><%=t("views.questions.eip_overview.hotline_notice.why") %></p>
+      <p class="text--underline"><%=t("views.questions.eip_overview.hotline_notice.number") %></p>
```
bengolder, line 33: I noticed that you underlined this. That would make it look like a link. You could make it into an actual link that allows a user to initiate a phone call.
```html
Call <a href="tel:+18443223639">+1 844 322 3639</a>
```

app/assets/stylesheets/application.scss
```diff
@@ -77,6 +77,10 @@ body {
   font-weight: bold;
 }
 
+.text--underline {
+  text-decoration: underline;
+}
+
```
bengolder, line 83: I see that we have other typographic styles in this file, but ideally, style classes like `.text--underline` should be in `atoms/typography` if possible. Non-blocking, but would be a small step to better CSS organization.

bengolder (approved):
approved with nonblocking ideas for possible improvements

app/views/questions/eip_overview/edit.html.erb
loumoore: oh that makes much more sense thanks

app/assets/stylesheets/application.scss
loumoore: ah yeah just saw the bold one here already..but will move all the typographic stuff to the right place, thanks!

PR #493, opened 2020-09-25
Small-batch, prefetched, bespoke cron queries

bengolder:
This reduces the scope of the initial cron job database query to only include intakes that have associated unsent documents.

Additionally, it reduces the batch size to 10, because why not work slowly and carefully, focusing only on a few intakes at a time? This is cron, no need to rush.

The existing query would iterate over every intake with an intake ticket id, which is a set that always grows over time, and I think that was reflected in the gradual increase of the cron job CPU utilization.

The new query is restricted only to the relevant superset (intakes with unsent docs) and then will figure out in the loop if those docs are too recent to worry about. We could incorporate this additional older unsent doc criteria (`documents.where("created_at < ?", 15.minutes.ago)`) into the query, but I was unsure of the syntax and I don't think we need to. These adjustments should be sufficient to avoid performance issues.

```sql
SELECT
    "intakes"."id" AS "t0_r0",
    -- ...
    "intakes"."zip_code" AS "t0_r170",
    "documents"."id" AS "t1_r0",
    "documents"."intake_id" AS "t1_r4",
    -- ...
    "documents"."zendesk_ticket_id" AS "t1_r6"
  FROM "intakes"
    INNER JOIN "documents"
    ON "documents"."intake_id" = "intakes"."id"
    WHERE ("intakes"."intake_ticket_id" IS NOT NULL) AND ("documents"."zendesk_ticket_id" IS NULL)
```

### DIFF
--- a/app/services/unsent_documents_service.rb
+++ b/app/services/unsent_documents_service.rb
@@ -6,10 +6,10 @@ class UnsentDocumentsService
     DatadogApi.increment("cronjob.documents.unsent.detect_and_notify")
 
     tickets_updated = 0
-    Intake.where.not(intake_ticket_id: nil).find_each(batch_size: 100) do |intake|
+    intakes_with_unsent_docs_query.find_each(batch_size: 10) do |intake|
       with_raven_context(intake_context(intake)) do
-        unsent_docs = intake.documents.where(zendesk_ticket_id: nil).where("created_at < ?", 15.minutes.ago)
-        if unsent_docs.present?
+        unsent_docs = intake.documents.where("created_at < ?", 15.minutes.ago)
+        if unsent_docs.exists?
           zendesk_service = ZendeskIntakeService.new(intake)
 
           ticket = zendesk_service.get_ticket(ticket_id: intake.intake_ticket_id)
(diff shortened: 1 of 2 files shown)


spec/services/unsent_documents_service_spec.rb
```diff
@@ -88,7 +88,7 @@
 
       it "sets the Raven exception context to the ticket ID when operating on the ticket ID" do
         service.detect_unsent_docs_and_notify
-        expect(service).to have_received(:with_raven_context).with(hash_including(ticket_id: 1))
+        expect(service).not_to have_received(:with_raven_context).with(hash_including(ticket_id: 1))
```
bengolder, line 91: This intake is no longer fed to the each loop! Because the its docs have already been sent, it is not included in the query.

app/services/unsent_documents_service.rb
```diff
@@ -6,10 +6,10 @@ def detect_unsent_docs_and_notify
     DatadogApi.increment("cronjob.documents.unsent.detect_and_notify")
 
     tickets_updated = 0
-    Intake.where.not(intake_ticket_id: nil).find_each(batch_size: 100) do |intake|
+    intakes_with_unsent_docs_query.find_each(batch_size: 10) do |intake|
       with_raven_context(intake_context(intake)) do
-        unsent_docs = intake.documents.where(zendesk_ticket_id: nil).where("created_at < ?", 15.minutes.ago)
-        if unsent_docs.present?
+        unsent_docs = intake.documents.where("created_at < ?", 15.minutes.ago)
+        if unsent_docs.exists?
```
loumoore, line 12: nice. TIL about the performance difference between present and exists!

loumoore (approved):
approved. all of the optimizations here make sense, thanks Ben!

PR #494, opened 2020-09-25
Link users to partners

bengolder:
This PR makes it so that a given User can belong to a VitaPartner.

- You can set an organization when inviting a user (they can't change it when they accept)
- You can edit an existing user's organization
- It lists the user's organization on the user index page.

### DIFF
--- a/app/models/vita_partner.rb
+++ b/app/models/vita_partner.rb
@@ -21,8 +21,13 @@ class VitaPartner < ApplicationRecord
   has_many :intakes
   has_and_belongs_to_many :states, association_foreign_key: :state_abbreviation
   has_many :source_parameters
+  has_many :users
 
   after_initialize :defaults
+
+  def self.select_input_options
+    all.collect { |v| [v.name, v.id] }
+  end
 
   def at_capacity?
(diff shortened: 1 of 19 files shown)


app/models/vita_partner.rb
```diff
@@ -21,9 +21,14 @@ class VitaPartner < ApplicationRecord
   has_many :intakes
   has_and_belongs_to_many :states, association_foreign_key: :state_abbreviation
   has_many :source_parameters
+  has_many :users
 
   after_initialize :defaults
 
+  def self.choices
```
bengolder, line 28: rename to select_options?

app/views/users/profile.html.erb
```diff
@@ -5,7 +5,10 @@
     <strong><%= t("general.email") %></strong> <%= current_user.email %>
   </p>
   <p>
-    <strong>Role:</strong> <%= current_user.role.capitalize %>
+    <strong><%= t("general.role") %>></strong> <%= current_user.role.capitalize %>
```
bengolder, line 8: &.capitalize ?

jenny-heath (approved):

jenny-heath:
approved with nonblocking comments!

PR #495, opened 2020-09-25
extend progress indicator to cover the entire flow

loumoore:
given we are seeing dropoff where it ends currently despite no significant ask of the client


app/services/intake_progress_calculator.rb
```diff
@@ -1,8 +1,11 @@
 class IntakeProgressCalculator
 
   starting_index = QuestionNavigation::FLOW.index(Questions::LifeSituationsController)
-  ending_index = QuestionNavigation::FLOW.index(Questions::AdditionalInfoController)
-  POSSIBLE_STEPS = QuestionNavigation::FLOW[starting_index..ending_index]
+  ending_index = QuestionNavigation::FLOW.index(Questions::SuccessfullySubmittedController)
+  question_steps = QuestionNavigation::FLOW[starting_index..ending_index]
+  doc_overview_index = question_steps.index(Questions::OverviewDocumentsController)
+
+  POSSIBLE_STEPS = question_steps.insert(doc_overview_index + 1, *DocumentNavigation::FLOW)
```
loumoore, line 8: this inserts the documents steps into the set of possible steps at the right position.

app/services/intake_progress_calculator.rb
```diff
@@ -22,9 +25,9 @@ def self.get_progress(controller, intake)
     current_index = steps_for_intake.index(controller)
     completed_steps = current_index + 1 # Everything that the client has done
 
-    index_of_possible_steps = POSSIBLE_STEPS.index(controller)
-    possible_remaining_steps = POSSIBLE_STEPS[index_of_possible_steps + 1..-1].length # Everything the client could potentially do
+    index_of_intake_steps = steps_for_intake.index(controller)
+    remaining_steps = steps_for_intake[index_of_intake_steps + 1..-1].length # Everything the client could potentially do
```
loumoore, line 29: this is a change in behavior i want to make sure is reviewed. this makes it so steps that we know won't be included based on the data we've collected won't be considered in the progress calculation. clearly it was intentional to include them in the past, but i can't understand why because the progress meter was less accurate. 

am i missing something?

spec/controllers/documents/w2s_controller_spec.rb
```diff
@@ -16,116 +16,5 @@
...
-
-        expect(MixpanelService).to have_received(:send_event).with(hash_including(
-          event_name: "validation_error",
-          data: hash_including({
-            invalid_document: true
-          })
-        ))
-      end
-    end
-  end
```
loumoore, line 129: the W2 form is never shown anymore, but all of these tests were left in place. i deleted rather than updating.

spec/services/intake_progress_calculator_spec.rb
```diff
@@ -10,23 +10,13 @@
...
-
-    it "includes all possible future steps until you reach the question" do
-      ever_married_yes = Intake.new(ever_married: :yes)
-      ever_married_no = Intake.new(ever_married: :no)
-      ever_married_unfilled = Intake.new
-      controller_before_question = Questions::IssuedIdentityPinController
-
-      expect(IntakeProgressCalculator.get_progress(controller_before_question, ever_married_yes)).to eq IntakeProgressCalculator.get_progress(controller_before_question, ever_married_no)
-      expect(IntakeProgressCalculator.get_progress(controller_before_question, ever_married_yes)).to eq IntakeProgressCalculator.get_progress(controller_before_question, ever_married_unfilled)
-    end
```
loumoore, line 30: this is related to the behavior change I commented about in app/services/intake_progress_calculator.rb above. i don't see why this is desirable behavior, but i may be missing something.

app/services/intake_progress_calculator.rb
loumoore: also, the reason to do this is that there are a ton of Document controllers in that flow so the status bar looks like you're way farther away from done than you actually are without this change.

but if this change is bad for some reason, i think the alternative would be to only add document controllers where the document type is needed_if_relevant? to the list of POSSIBLE_STEPS because there aren't that many of those.

spec/controllers/documents/w2s_controller_spec.rb
bengolder: These were meant to test basic functionality that is shared across document pages by putting them on one controller in particular. The specs should have been moved when the W-2 controller was deprecated. If possible, it would be helpful to move these specs to a different document controller, and perhaps add a comment that they are meant to test the behavior of a given document controller.

bengolder (approved):
Thanks for working to fix this issue!

Overall, I'm okay with sending this out. It takes some focus to reason about how it will behave for some of the intermediate pages, like document pages. I think it would be improved by an additional spec that better documents this behavior for intermediate steps (not just 0% and 100%), but I wouldn't block on that because I think I understand what it is doing enough to trust it.

However, before we merge I think it's important to make sure that this was manually checked in local development to verify that the percentages increase at a reasonable pace. Have you already done that?

spec/controllers/documents/w2s_controller_spec.rb
loumoore: ah gotcha. i believe w-2s were refactored into the new Employment screen, so that may be a natural place to add this back. thanks!

loumoore:
@bengolder i have stepped through it all manually watching the percentages. i agree some more intermediate automated testing would be valuable though given all the permutations of flows (211intake, intake, eip) and questions especially as we add more gating logic. 

i was having a hard time thinking of a way to test that wouldn't be annoyingly brittle...but i will think on it a bit more before i merge. 

i'm going to fix up that W2 test anyway first as well, and might be able to get some more robust testing of the progress indicator in place for a monday merge. thanks for looking it over!

spec/controllers/documents/w2s_controller_spec.rb
loumoore: turns out this was already done, the specs in question have already been translated to the new employment_controller_spec, they were just left in w2s_controller_spec too. so i think that part is ok to ship.

loumoore:
@bengolder doesn't really feel comprehensive but I realized I could add back the kind of test I had removed from intake_progress_calculator_spec and update it to essentially validate the opposite behavior. so, using the question about marriage status, i validate that 1) future optional questions are not included in the progress calculation until determined necessary and 2) the progress bar slows down when an answer results in new questions becoming necessary.

bengolder (approved):
Thanks Lou!

PR #496, opened 2020-09-25
Realtime spike #174714481

bengolder:


Gemfile
```diff
@@ -58,6 +58,7 @@ gem 'websocket-extensions', '>= 0.1.5'
 gem 'twilio-ruby', '~> 5.40.1'
 gem 'mailgun-ruby', '~>1.1.6'
 gem 'devise_invitable', '~> 2.0.0'
+gem 'webpacker'
```
sartan, line 61: Add webpacker to compile Javascript

app/views/clients/show.html.erb
```diff
@@ -11,8 +11,8 @@
     <h3>Phone for texting: <%= @client.sms_phone_number %></h3>
   </div>
   <h2>Contact History</h2>
-  <% if @contact_history.present? %>
-    <ul class="contact-history">
+    <ul id="contact-history" aria-live="polite">
```
sartan, line 14: DOM element `#contact-history` get replaced with new data as it streams in. For accessibility, [aria-live=\"polite\"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) defines a Live Region and notifies user of updates.

app/views/layouts/application.html.erb
```diff
@@ -34,6 +34,7 @@
         full_path: request.fullpath
       ).inspect.html_safe %>);
     </script>
+    <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
```
sartan: Renders a `<script>` tag for `/packs/js/application.js` which is the entry-point for the front-end portion of this implementation.

app/assets/stylesheets/components/_contact-history.scss
```diff
@@ -1,5 +1,8 @@
-.contact-history {
+#contact-history {
```
alaroia-cfa, line 1: We made this an ID to make it easier to select the element and replace it with data from the server.

app/assets/stylesheets/components/_contact-history.scss
```diff
@@ -1,5 +1,8 @@
...
   border-top: 2rem solid darken($color-grey-light, 50%);
   background-color: darken($color-grey-light, 10%);
   padding: 1rem;
 }
+#contact-history:focus {
```
alaroia-cfa, line 6: We attempted to add `:focus` styles. We were unable to trigger these when calling `.focus()` on the DOM element, but we suspect that's something that is easily fixable.

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -21,6 +21,10 @@ def create_incoming_text_message
       client: client,
     )
     head :ok
+
+  # To notify any users on the client's page that there is new contact info, add this
```
alaroia-cfa, line 25: This is where we would add code to notify web browsers of a change. If we were worried about load, this could peacefully occur in a job.

alaroia-cfa:
I'm going to close this pull request - https://hackmd.io/y_kxJODlQ6GdCCI2a34r2Q is more informative.

PR #498, opened 2020-09-26
add cache-control header to static assets

loumoore:
a first step towards better front end performance.


loumoore:
oops i didn't update my commit message from the previous one...this is a cache-control header change as the PR title suggests. i'll do this over.

PR #500, opened 2020-09-26
Link clients to vita partners

bengolder:
This simply adds a foreign key from a client to a vita_partner.
When a client is created from an intake, the intake's vita_partner_id is copied over to the client.


sartan (approved):
@alaroia-cfa and I reviewed it and we both think it looks great!

PR #503, opened 2020-09-29
Add more details to outgoing text message display

jenny-heath:
- adds method for formatted phone number
- adds method for formatted timestamp
- adds empty test contexts for other message types (incoming text, outgoing & incoming email)

image

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -36,4 +36,9 @@ class Client < ApplicationRecord
       vita_partner: intake.vita_partner,
     )
   end
+
+  # Returns the phone number formatted for user display, e.g.: "(510) 555-1234"
+  def formatted_phone_number
+    Phonelib.parse(phone_number).local_number
+  end
 end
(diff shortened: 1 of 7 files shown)


app/models/client.rb
```diff
@@ -36,4 +36,9 @@ def self.create_from_intake(intake)
       vita_partner: intake.vita_partner,
     )
   end
+
+  # Returns the phone number formatted for user display, e.g.: "(510) 555-1234"
+  def formatted_phone_number
+    Phonelib.parse(phone_number).local_number
+  end
```
bengolder, line 43: We have a bunch of these formatted_phone_number methods spread around the code base. It might make sense to create a view helper for this, thereby making a method that can be used in any template for any model.

```ruby
# in app/helpers
module PhoneNumberHelper
  def local_phone_format(phone_number)
    Phonelib.parse(phone_number).local_number
  end
end
```

```html.erb
<%# in any template %>
<%= local_phone_format(client.phone_number) %>
```

bengolder (approved):
Looks good, I appreciate the tests. We might consider refactoring some of this functionality into a view helper,  but I wouldn't block on that.

app/models/client.rb
jenny-heath: will do this as part of the next one!

PR #505, opened 2020-09-30
Add/change message details on contact history

jenny-heath:
image


app/models/incoming_email.rb
```diff
@@ -40,6 +40,6 @@ def datetime
   end
 
   def author
-    from
+    client.preferred_name
   end
```
bengolder, line 44: I think these changes are a little risky. If a client's contact info changes, or we have to resolve contact info between clients, this could give the false impression that a message was sent in the past by the current client in the present.
Zendesk did this with requesters, and it made it hard for us to debug when we ran into edge cases.
In other words, I don't think we should attribute messages to some piece of info that is mutable and we expect might change over time. If it just displays the "from" info that it pulled from the message itself, we prevent the risk of misattribution if we ever have to resolve weird edge cases involving contact info.

app/views/messages/index.html.erb
```diff
@@ -23,13 +23,31 @@
...
 
+              <% if contact_record.contact_record_type == :incoming_text_message %>
+                <span class="contact-record__phone_number">
+                  Text from <%= local_phone_number(contact_record.from_phone_number) %>
+                </span>
+              <% end %>
+
+              <% if contact_record.contact_record_type == :outgoing_email %>
+                <span class="contact-record__phone_number">
+                  Email to <%= contact_record.client.email_address %>
```
alaroia-cfa, line 41: I think we should adjust our outgoing email model so we store the email address we sent the message to, then show that here. I'll likely do that, just logging it.

app/helpers/phone_number_helper.rb
```diff
@@ -0,0 +1,8 @@
+module PhoneNumberHelper
```
alaroia-cfa, line 1: Awesome!

app/views/messages/index.html.erb
```diff
@@ -23,13 +23,31 @@
 
               <% if contact_record.contact_record_type == :outgoing_text_message %>
                 <span class="contact-record__phone_number">
-                  Text to <%= contact_record.client.formatted_phone_number %>
+                  Text to <%= local_phone_number(contact_record.client.phone_number) %>
```
alaroia-cfa, line 26: IMHO we should store the phone number we sent it to, at the time we send it, so we have a record of what happened, even if the client's phone number changed.

I'll probably make that change, just logging it here.

sartan:
Reviewed by @bengolder

PR #506, opened 2020-09-30
Make first comment public and add tags if given

bengolder:
We need the first comment to be public, otherwise all volunteers will see a weird warning when they open a new ticket.
image (1)

We apparently weren't properly setting tags during `create_ticket`, which I only learned while manually checking in development and seeing the result in Zendesk.


alaroia-cfa (approved):
@bengolder Huge thanks for writing and testing these changes. Superb.

PR #507, opened 2020-09-30
Add communication opt in tags to dropoffs

sniemeyer13:
Co-authored-by: Ben Golder <bgolder@codeforamerica.org>

### DIFF
--- a/app/services/zendesk_drop_off_service.rb
+++ b/app/services/zendesk_drop_off_service.rb
@@ -44,6 +44,7 @@ class ZendeskDropOffService
           EitcZendeskInstance::HSA => @drop_off.hsa,
           EitcZendeskInstance::INTAKE_SITE => intake_site_tag,
           EitcZendeskInstance::STATE => @drop_off.state,
+          EitcZendeskInstance::COMMUNICATION_PREFERENCES => communication_tags,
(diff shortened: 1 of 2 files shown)


app/services/zendesk_drop_off_service.rb
```diff
@@ -82,6 +83,13 @@ def comment_body
 
   private
 
+  def communication_tags
+    [
+      ("email_opt_in" if @drop_off.email.present?),
```
alaroia-cfa, line 88: Per the tests, we seem to always yield the array ["email_opt_in", "sms_opt_in"].

It would be nice if one of these tests validated our use of the `.present?`.

Not a showstopper.

alaroia-cfa (approved):
Good work! I would prefer to see the `.present?` tested clearly, but that's not a showstopper. Approved. Happy to review promptly if you want to address that.

app/services/zendesk_drop_off_service.rb
sniemeyer13: Great! We just pushed some changes that added tests for the individual presence of each tag.

alaroia-cfa: :D

PR #508, opened 2020-10-01
fix errors caused by progress calculations when a controller can be shown but show? doesn't evaluate to true

bengolder:
…


loumoore (approved):
reviewed with @bengolder !

PR #510, opened 2020-10-02
Documents page

jenny-heath:
Screen Shot 2020-10-01 at 4 43 16 PM
Screen Shot 2020-10-01 at 4 35 16 PM


app/controllers/documents_controller.rb
```diff
@@ -1,5 +1,20 @@
 class DocumentsController < ApplicationController
-  before_action :require_intake
+  include AccessControllable
+  include FileResponseControllerHelper
```
alaroia-cfa, line 3: IMHO within the FileResponseControllerHelper, we need to start using `.service_url` from ActiveStorage as part of resolving a security issue we found a little while back. I consider it non-blocking for this pull request for now.

Using `.service_url` would mean redirecting to a S3 signed URL, which would ensure that if the client uploaded a HTML file, it renders on a different domain than ours, which separates its Javascript from our origin.

app/controllers/documents_controller.rb
```diff
@@ -1,5 +1,20 @@
...
+    @documents = @client.documents
+  end
+
+  def show
+    @document = Document.find(params[:id])
+    render_active_storage_attachment @document.upload
+  end
 
   def destroy
     document = current_intake.documents.find_by(id: params[:id])
```
alaroia-cfa, line 20: ❓ I was initially confused to see that the intake-related `destroy` lives in the same controller as `client`-related `index` and `show`!

👍 It's also OK. I foresee confusion when we add a client document view delete button. We can resolve any such confusion down the road.

app/models/document.rb
```diff
@@ -28,6 +32,19 @@ class Document < ApplicationRecord
...
   belongs_to :documents_request, optional: true
   has_one_attached :upload
+
+  before_save :set_display_name
+
+  def set_display_name
+    return if display_name
+
+    if upload.present?
+      self.display_name = upload.attachment.filename
```
alaroia-cfa, line 45: Curious why we're not modifying the filename directly. https://stackoverflow.com/a/52957480 should do the trick for that. Non-blocking.

app/views/documents/index.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <th class="document-column__upload-date"><%=t(".upload_date") %></th>
+    </tr>
+    </thead>
+    <tbody>
+    <% if @documents.present? %>
+      <% @documents.each do |document| %>
+        <tr id="document-<%= document.id %>">
+          <td><%= document.document_type %></td>
+          <td>
+            <%= link_to document_path(id: document.id), target: "_blank" do %>
```
alaroia-cfa, line 22: Would be nice to toss `rel="noopener noreferrer"` in there, too, to avoid an arguably-minor confusion/security-issue vector https://sites.google.com/site/bughunteruniversity/nonvuln/phishing-with-window-opener

app/views/messages/index.html.erb
```diff
@@ -1,12 +1,7 @@
 <% content_for :page_title, "#{@client.preferred_name}, ##{@client.id}" %>
 
 <% content_for :card do %>
-  <div class="client-header">
-    <h1 class="h2">
-      <%= @client.preferred_name %>
-    </h1>
-    <h2>#<%= @client.id %></h2>
-  </div>
+  <%= render "clients/client_header" %>
```
alaroia-cfa, line 4: AWESOME

app/controllers/documents_controller.rb
```diff
@@ -1,5 +1,20 @@
...
+  before_action :require_intake, only: [:destroy]
+  layout "admin", only: [:index]
+
+  def index
+    @client = Client.find(params[:client_id])
+    @documents = @client.documents
+  end
+
+  def show
+    @document = Document.find(params[:id])
```
alaroia-cfa, line 15: One day we'll need to add access checking to this, but now is not that time.

alaroia-cfa (approved):
This is good! I had some confusion when reading it, but I also was able to understand it.

If I have time, I'll fix the `rel=noopener` suggestion, then merge it straight to master in the interest of latency-reduction.

alaroia-cfa:
I merged this as b8ed82f4f3133f353480721cb23e381766d49886 , which resolves this pull request. (The merge button was disabled in GitHub due to the conflicts in the history.)

app/controllers/documents_controller.rb
jenny-heath: a good point! we should probably namespace everything going into case management soon

app/models/document.rb
jenny-heath: we talked about this with ben! decided it made sense to preserve the original filename

PR #512, opened 2020-10-05
Edit document name #174938639

bytheway875:
Paired on with @jenny-heath 

Screen Shot 2020-10-05 at 8 48 59 AM
Screen Shot 2020-10-05 at 8 49 08 AM


app/assets/stylesheets/components/_document-table.scss
```diff
@@ -2,26 +2,45 @@
   table-layout: fixed;
   width: 904px;
```
bytheway875, line 4: I think there is some sort of prettification/formatter applied to the document that made some inadvertant changes. I can revert if its a problem.

app/helpers/documents_overview_helper.rb
```diff
@@ -1,5 +1,5 @@
 module DocumentsOverviewHelper
-  def edit_document_path(document_type)
+  def edit_document_path_for(document_type)
```
bytheway875, line 2: I had to change the name of this helper because the edit_document_path helper overwrote the default behavior of "edit_document_path" routing helper. Overwriting default helpers seems prone to odd behavior/bugs. It wasn't being used anywhere else, and I did need to use the default "edit_document_path" helper, so I changed the name of this helper.

config/routes.rb
```diff
@@ -36,7 +36,7 @@ def scoped_navigation_routes(context, navigation)
       end
     end
 
-    resources :documents, controller: :documents do
+    resources :documents, only: [:destroy], controller: :documents do
```
bytheway875, line 39: This route was, by default, creating all of the restful routes for documents in addition to all of the DocumentNavigation custom routes that were being created in the block. Based on other existing code, the creation of these default routes seemed inadvertent (only a destroy method in controller, later defines only: destroy on line 121)

config/routes.rb
```diff
@@ -119,6 +118,8 @@ def scoped_navigation_routes(context, navigation)
       resources :documents, only: [:index]
       resources :notes, only: [:index]
     end
+    resources :documents, only: [:edit, :update, :show]
```
bytheway875, line 121: I think it would make sense to namespace the admin/case management routes similar to how we do the Zendesk stuff... I ran into a couple namespace issues by having them unscoped, and had to make the decision of whether to nest the documents routes under clients in order to "force" some sort of namespacing, but that added unnecessary complexity to the forms. For now, I moved the documents routes that are case-management related down near the case management routes so that at least there is a logical separation in the routes file. Namespacing also has some other benefits for code/route organization and access control. Hopefully we can talk about name spacing in the gardening meeting!

app/models/client.rb
```diff
@@ -35,6 +35,7 @@ def self.create_from_intake(intake)
       phone_number: intake.phone_number,
       sms_phone_number: intake.sms_phone_number,
       vita_partner: intake.vita_partner,
+      documents: intake.documents,
```
bytheway875, line 38: This should copy the documents from the intake onto the client directly, which should work for now.

alaroia-cfa, line 38: Let's add a test to `spec/controllers/clients_controller_spec.rb`

config/routes.rb
alaroia-cfa: Let's ensure that this is nicely tested

alaroia-cfa: Tis

alaroia-cfa:
Approved! :D

PR #514, opened 2020-10-06
Don't react to SMS from invalid phone numbers

bengolder:


bengolder:
Added directly on master via separate commit

PR #515, opened 2020-10-06
CanCanCan Spike

bengolder:
# CanCanCan Gem Spike

[Also available on HackMD](https://hackmd.io/jjGatRUBReu-roKyuCnDMg?view)

Resource-based permissions in controllers. It grows well, and keeps access control organized, consistent, and readable.

### Centralizes unauthorized access handling. We can ensure that we always log unauthorized access attempts.

```ruby
  rescue_from CanCan::AccessDenied do |exception|
    # "attempted unauthorized access by user 3 to read Client"
    log "attempted unauthorized access by #{current_user} to #{exception.action} #{exception.subject}"
    respond_to do |format|
      format.html { head :forbidden }
    end
  end
```

### Fetches the right scope, by default

```ruby
class DocumentsController
  # checks for access a particular client, sets @client
  load_and_authorize_resource :client
  # checks for access to documents, sets @documents to only those documents user can access that are linked to client
  load_and_authorize_resource through: :client
end
```

- Centralized way to grab currently authorized resources.
- Does routine work in a helpful way, by fetching only resources that are available to the current user.
- Encourages us to scope nested resources under access to the parent model.

### Super flexible and facilitates more complex permissions logic

```ruby
class Ability
  include CanCan::Ability

  def initialize(user)
    can :manage, Client if user.is_super_admin?
    can :manage, Client, organization: user.organization
    # coalition
    can :manage, Client, organization: { parent_organization: { id: user.organization_id } }
    if user.is_greeter? || user.is_client_support?
      # basic greeter or client support check
      can :manage, Client, organization_id: user.supported_organizations
      # greeter supporting a coalition parent
      can :manage, Client, organization: { coalition_parent_id: user.supported_organizations }
    end

    if user.is_site_coordinator?
      can :manage, User, organization: user.organization
    end

    if user.is_coalition_owner?
      can :manage, Organization, coalition_parent_id: user.organization_id
    end
  end
end
```

We can proxy permission to do things through access to a client

```ruby
class MessagesController
  load_and_authorize_resource :client
  load_and_authorize_resource :outgoing_text_message, parent: false, through: :client
  load_and_authorize_resource :incoming_text_message, parent: false, through: :client
  load_and_authorize_resource :outgoing_email, parent: false, through: :client
  load_and_authorize_resource :incoming_email, parent: false, through: :client

  def index
    @contact_history = (
      @outgoing_text_messages.includes(:user) +
      @incoming_text_messages +
      @outgoing_emails.includes(:user) +
      @incoming_emails
    ).sort_by(&:datetime)
  end
end
```
- If we ever want to differentiate the ability to read, write, or delete based on any criteria, this gem would make it much easier to implement.
- If we ever need more complex permissions logic, such as storing permissions in the database or creating custom permissions logic (for example, limiting access to certain spans of time)

## Mixed
- Probably does not eliminate need to write controller specs to affirm correct access control or scope. But shared code for determining access might make access control specs easier to write.
- I don't think it's actually reducing the lines of code.

## Risks
- Always make sure that we check if the user is signed in before checking their abilities.
- If a controller is not clearly tied to a resource, we need to be careful to pin access to something helpful. Not clear how to tie permissions to things other than Models.
- Lack of transparency for querying access carries some risk of bugs or performance issues. But probably less risk than home-cooked code.

### Cascading Ability Sheets
Can statements are executed as a series of `OR` queries. Like CSS, subsequent `can` statements override previous statements.

## Multiple Ability Classes
A common way to implement CanCanCan permissions is to write a single `Ability` class and put all the permission logic inside of the single `initialize` method.
`current_ability` is a method that determines the currently instantiated Ability class. If we want to use multiple Ability classes, we'd have to overwrite `current_ability`.
I think we'd be most likely to separate abilities by _role_. For example, a `SiteCoordinatorAbility` or a `GreeterAbility`.

### Choosing a specific Ability class

```ruby
def current_ability
  @current_ability ||= MyAbility.new(current_user)
end
```

### Merging multiple Ability classes

```ruby
def current_ability
  @current_ability ||= MyAbility.new(current_user).merge(MyOtherAbility.new(current_user))
end
```

### A more complex example of ability structure
```
# file tree
app/
  abilities/
    base_ability.rb
    one_ability.rb  # inherits from base_ability
    other_ability.rb  # inherits from base_ability
    another_ability.rb  # inherits from base_ability
```

```ruby
def current_ability
  # build method can figure out which class to instantiate
  @current_ability ||= Abilities::BaseAbility.build(current_user, request)
end
```

## Further Reading

- [Defining Abilities: Best Practices](https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities:-Best-Practices)
- [Ensuring Authorization is run for a broad set of controllers](https://github.com/CanCanCommunity/cancancan/wiki/Ensure-Authorization)

### DIFF
--- a/app/controllers/outgoing_emails_controller.rb
+++ b/app/controllers/outgoing_emails_controller.rb
@@ -1,15 +1,15 @@
 class OutgoingEmailsController < ApplicationController
   include AccessControllable
 
-  before_action :require_sign_in, :require_beta_tester
+  before_action :require_sign_in
+  load_and_authorize_resource
 
   def create
-    email = OutgoingEmail.new(outgoing_email_params)
-    email.to = email.client.email_address
-    if email.save
-      OutgoingEmailMailer.user_message(outgoing_email: email).deliver_later
+    @outgoing_email.to = @outgoing_email.client.email_address
+    if @outgoing_email.save
+      OutgoingEmailMailer.user_message(outgoing_email: @outgoing_email).deliver_later
     end
-    redirect_to client_messages_path(client_id: email.client_id)
+    redirect_to client_messages_path(client_id: @outgoing_email.client_id)
   end
 
   private
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ gem 'websocket-extensions', '>= 0.1.5'
 gem 'twilio-ruby', '~> 5.40.1'
 gem 'mailgun-ruby', '~>1.1.6'
 gem 'devise_invitable', '~> 2.0.0'
+gem 'cancancan'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -114,6 +114,7 @@ GEM
       thor (~> 0.19)
     builder (3.2.4)
     byebug (11.1.3)
+    cancancan (3.1.0)
     capybara (3.33.0)
       addressable
       mini_mime (>= 0.1.3)
@@ -511,6 +512,7 @@ DEPENDENCIES
   axe-matchers
   bootsnap (>= 1.1.0)
   byebug
+  cancancan
   capybara (>= 2.15)
   cfa-styleguide!
   ddtrace
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -239,4 +239,10 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path_for(_user)
     user_profile_path
   end
+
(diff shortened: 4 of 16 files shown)


app/controllers/outgoing_emails_controller.rb
```diff
@@ -1,15 +1,15 @@
 class OutgoingEmailsController < ApplicationController
   include AccessControllable
 
-  before_action :require_sign_in, :require_beta_tester
+  before_action :require_sign_in
+  load_and_authorize_resource
```
alaroia-cfa, line 5: This is my absolute favorite

PR #518, opened 2020-10-07
Add date heading to message history

sartan:

image

### DIFF
--- a/app/assets/stylesheets/components/_contact-history.scss
+++ b/app/assets/stylesheets/components/_contact-history.scss
@@ -2,4 +2,15 @@
   border-top: 2rem solid darken($color-grey-light, 50%);
   background-color: darken($color-grey-light, 10%);
   padding: 1rem;
+
+  .contact-record__day-heading:first-of-type {
+    padding-top: inherit;
+  }
+
+  .contact-record__day-heading {
+    text-align: center;
+    color: $color-grey-dark;
+    padding-top: 0.5em;
+    padding-bottom: 0.5em;
+  }
 }
--- a/app/assets/stylesheets/components/_contact-history.scss
+++ b/app/assets/stylesheets/components/_contact-history.scss
@@ -2,15 +2,4 @@
   border-top: 2rem solid darken($color-grey-light, 50%);
   background-color: darken($color-grey-light, 10%);
   padding: 1rem;
-
-  .contact-record__day-heading:first-of-type {
-    padding-top: inherit;
-  }
-
-  .contact-record__day-heading {
-    text-align: center;
-    color: $color-grey-dark;
-    padding-top: 0.5em;
-    padding-bottom: 0.5em;
-  }
 }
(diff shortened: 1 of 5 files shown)


app/views/messages/index.html.erb
```diff
@@ -8,7 +8,12 @@
   <h2>Contact History</h2>
   <% if @contact_history.present? %>
     <ul class="contact-history">
+      <% last_date = nil %>
       <% @contact_history.each do |contact_record| %>
+        <% if contact_record.datetime.to_date != last_date %>
+          <li class="contact-record__day-heading"><%= if contact_record.datetime.to_date == Date.today then "Today" end %> <%= contact_record.datetime.to_date.strftime("%m/%d/%Y") %></li>
+          <% last_date = contact_record.datetime.to_date %>
+        <% end %>
```
bengolder, line 16: Realized that we may have to change this when we start displaying things in the user's timezone, but this makes sense since we're only displaying UTC right now.

bengolder: Shannon pairing with Ben + Sarah:

Have you considered group_by to group the records by date before heading into the view?

@records_by_day = @contact_history.group_by { |record| record.created_at.beginning_of_day }

{key is the datetime beginning of day => []}

app/views/messages/index.html.erb
```diff
@@ -8,7 +8,12 @@
   <h2>Contact History</h2>
   <% if @contact_history.present? %>
     <ul class="contact-history">
+      <% last_date = nil %>
       <% @contact_history.each do |contact_record| %>
+        <% if contact_record.datetime.to_date != last_date %>
+          <li class="contact-record__day-heading"><%= if contact_record.datetime.to_date == Date.today then "Today" end %> <%= contact_record.datetime.to_date.strftime("%m/%d/%Y") %></li>
```
bengolder, line 14: Putting "Today" into a translation string would help ensure we don't miss it later

app/assets/stylesheets/components/_contact-history.scss
```diff
@@ -2,4 +2,15 @@
...
+  .contact-record__day-heading:first-of-type {
+    padding-top: inherit;
+  }
+
+  .contact-record__day-heading {
+    text-align: center;
+    color: $color-grey-dark;
+    padding-top: 0.5em;
+    padding-bottom: 0.5em;
+  }
```
bengolder, line 15: Optional scss syntax that you can use to group elements:

```scss
.contact-record {
  &__day-heading { // same as .contact-record__day-heading
    &:first-of-type {
     }
  }
}
```

bengolder (approved):
(this is Ben, Shannon, & Sarah all reviewing together)
Approved! This looks great!

Appreciated the screenshot, which made it easier to review. We commented with some observations, none of them are blocking, in case you find them helpful.

Thanks!

app/views/messages/index.html.erb
sartan: We considered it, and we chose this approach for simplicity. Asheesh has a preference for flatter data structures that comes out of his time writing Python (they have a motto in Python land about his apparently, "flat is better than nested"). I think either approach works, so went with his preference.

PR #526, opened 2020-10-08
User can edit their timezone and timestamps show in user's selected zone

bytheway875:
Defaults to east coast time if browser can't recognize user's time. 

Mostly because it seems like West Coast people have a better understanding of time zones than East Coast people because most things default to East Coast time.


bytheway875:
@alaroia-cfa I've been tinkering with this failing spec for a while now and nothings jumping out -- it fails on CI but not locally. I'm wondering if we're either massaging the date we want to see as the group_by key value in a way thats causing inconsistency on CI? Or in the controller setup itself?

sniemeyer13:
Asheesh and I just pushed a fix to the failing test. 

It was not failing locally, so we attempted to recreate the failure by faking a different timezone on our local machine. To do this, we ran the spec with `TZ=Europe/London rspec /Users/sarahniemeyer/workspace/vita-min/spec/controllers/case_management/notes_controller_spec.rb`

According to the rails `in_time_zone` docs, time and dates in timezone methods act differently. 
```ruby
    #   Time.utc(2000).in_time_zone('Alaska') # => Fri, 31 Dec 1999 15:00:00 AKST -09:00
    #   Date.new(2000).in_time_zone('Alaska') # => Sat, 01 Jan 2000 00:00:00 AKST -09:00
```

PR #528, opened 2020-10-09
Upgrade rails to fix Hakiri warning

alaroia-cfa:

To validate locally that this fixes the CVE-2020-8264 warning, you can run:

```
bundle exec bundle-audit --ignore CVE-2015-9284
```

Note that we ignore CVE-2015-9284 because it's a software bug with a documented mitigation, and we already use that mitigation; see [Resolving CVE-2015-9284](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284).

We looked at the Puma changelog with a critical eye and concluded that we expect it to be peaceful for us. We'll also be able to validate that on demo.


bytheway875 (approved):
👯

PR #533, opened 2020-10-13
Track stimulus link clicks

sniemeyer13:


alaroia-cfa (approved):
👍 

Matches the story; often we do mixpanel stuff in the backend, but I see the story asked for it this way, so OK!

PR #534, opened 2020-10-13
Implement realtime messaging with ActionCable

sniemeyer13:
We plan on future clean-up commits, however realtime messaging is now working!

Check out our gif to see it working >>

out


sniemeyer13:
Please allow me to merge this.

app/channels/client_channel.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    unless current_ability.can? :read, @client
+      reject
+    end
+
+    stream_for @client
+  end
+
+  def self.broadcast_contact_record(contact_record)
+    broadcast_to(contact_record.client, [".message-list", ApplicationController.render(partial: 'shared/message_list_contact_record', locals: { contact_record: contact_record })])
```
bengolder, line 17: why does the backend need to know this CSS class? Couldn't the front end figure this out on its own?

app/controllers/case_management/outgoing_emails_controller.rb
```diff
@@ -11,6 +11,7 @@ def create
       if @outgoing_email.save
         OutgoingEmailMailer.user_message(outgoing_email: @outgoing_email).deliver_later
       end
+      ClientChannel.broadcast_contact_record(@outgoing_email)
```
bengolder, line 14: I like how simple and clear this is!

app/controllers/case_management/outgoing_text_messages_controller.rb
```diff
@@ -10,6 +10,7 @@ def create
       @outgoing_text_message.to_phone_number = @outgoing_text_message.client.sms_phone_number
       if @outgoing_text_message.save
         SendOutgoingTextMessageJob.perform_later(@outgoing_text_message.id)
+        ClientChannel.broadcast_contact_record(@outgoing_text_message)
```
bengolder, line 13: I love how we are broadcasting outgoing as well as incoming!

app/javascript/channels/client_channel.js
```diff
@@ -0,0 +1,20 @@
...
+
+  disconnected() {
+    console.log("We're disconnected!")
+  },
+
+  received(data) {
+    let selector = data[0];
+    let newContent = data[1];
+
+    document.querySelector(selector).innerHTML += newContent;
```
bengolder, line 18: I wouldn't expect this to take an arbitrary selector. But totally understand this preliminary to get things working.

app/javascript/channels/client_channel.js
```diff
@@ -0,0 +1,20 @@
+export function getChannelName(url) {
+  return { channel: "ClientChannel", id: url.match(/\/clients\/(\d*)\/messages/)[1] }
```
bengolder, line 2: was this autogenerated? if not, might be helpful to either link to an explanation of this regex or add comments to explain. Is this the recommended approach from Rails documentation?

app/javascript/channels/index.js
```diff
@@ -1,5 +1,9 @@
+// Load channel JS files within this directory and all subdirectories.
+// We require() them explicitly. By convention, they end in *_channel.js.
+require('./client_channel.js');
+
 function sayHello() {
-  return "hello!";
+    return "hello!";
 }
 
 module.exports = sayHello;
```
bengolder, line 9: I think we can delete `sayHello` now that we have actual JS to run :)

spec/channels/client_channel_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      it 'subscribes to a client' do
+        subscribe params
+
+        expect(subscription).to be_confirmed
+        expect(subscription).to have_stream_for(client)
+      end
+    end
+  end
+
+  context "publishing a message" do
```
bengolder, line 34: Would be more consistent with other specs if we said `describe "#broadcast_contact_record"`

spec/controllers/case_management/outgoing_emails_controller_spec.rb
```diff
@@ -41,6 +41,17 @@
...
+        context "real-time updates" do
+          before do
+            allow(ClientChannel).to receive(:broadcast_contact_record)
+          end
+
+          it "sends a real-time update to anyone on this client's page", active_job: true do
+            post :create, params: params
+            expect(ClientChannel).to have_received(:broadcast_contact_record).with(OutgoingEmail.last)
+          end
+        end
```
bengolder, line 54: I'd recommend moving the `before` statement into the parent context before (line 28) and just having this `it` as a separate `it` block in the parent context.

spec/controllers/case_management/outgoing_text_messages_controller_spec.rb
```diff
@@ -31,6 +31,17 @@
...
       end
+
+      context "real-time updates" do
+        before do
+          allow(ClientChannel).to receive(:broadcast_contact_record)
+        end
+
+        it "sends a real-time update to anyone on this client's page" do
+          post :create, params: params
+          expect(ClientChannel).to have_received(:broadcast_contact_record).with(OutgoingTextMessage.last)
```
bengolder, line 42: great! would also move this into the parent context.

spec/javascript/channels/client_channel.spec.js
```diff
@@ -0,0 +1,20 @@
+import {callback, getChannelName} from "channels/client_channel";
+import {describe} from "@jest/globals";
```
bengolder, line 2: not sure if we need this line

spec/javascript/channels/client_channel.spec.js
```diff
@@ -0,0 +1,20 @@
...
+import consumer from "../../../app/javascript/channels/consumer";
```
bengolder, line 3: do we need these lines?

spec/javascript/channels/client_channel.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        expect(document.body.innerHTML).toEqual(
+            '<div id="contact-history">old<p>unrelated</p><p id="#new-message">a message came in</p></div>'
+        );
+    });
+
+    test("can create a client channel based on id in url", () => {
+        expect(getChannelName("http://www.example.com/en/case_management/clients/14/messages")).toEqual(
+            {channel: "ClientChannel", id: "14"}
+        );
+    })
```
bengolder, line 19: might be helpful to label/structure tests around the separate methods they test

bengolder (approved):
Super Approve!!! Thank you!

app/channels/client_channel.rb
bytheway875: If .message-list is a class on the notes page (i think we used the same class on both pages), will this also append new client contacts (text and emails) to the notes page by accident?

sartan: @bytheway875 It looks like we're [scoping the subscription](https://github.com/codeforamerica/vita-min/blob/master/app/javascript/listeners/index.js#L11) to the messages page, so even though the javascript is included on all pages, the callback wouldn't be triggered.

PR #536, opened 2020-10-14
Add basic client info to client profile page

bytheway875:

### DIFF
--- a/app/controllers/case_management/clients_controller.rb
+++ b/app/controllers/case_management/clients_controller.rb
@@ -23,15 +23,6 @@ module CaseManagement
       redirect_to case_management_client_path(id: client.id)
     end
 
-    def show
-      @contact_history = (
-        @client.outgoing_text_messages.includes(:user) +
-        @client.incoming_text_messages +
-        @client.outgoing_emails.includes(:user) +
-        @client.incoming_emails
-      ).sort_by(&:datetime)
-      @outgoing_text_message = OutgoingTextMessage.new(client: @client)
-      @outgoing_email = OutgoingEmail.new(client: @client)
-    end
+    def show; end
   end
 end
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -454,6 +454,16 @@ class Intake < ApplicationRecord
     (most_recent_filing_year.to_i - 1).to_s if most_recent_filing_year.present?
   end
 
+  def marital_statuses
+    [
+        ("Married" if married_yes?),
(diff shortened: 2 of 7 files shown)


app/controllers/case_management/clients_controller.rb
```diff
@@ -23,15 +23,6 @@ def create
...
-      @contact_history = (
-        @client.outgoing_text_messages.includes(:user) +
-        @client.incoming_text_messages +
-        @client.outgoing_emails.includes(:user) +
-        @client.incoming_emails
-      ).sort_by(&:datetime)
-      @outgoing_text_message = OutgoingTextMessage.new(client: @client)
-      @outgoing_email = OutgoingEmail.new(client: @client)
-    end
+    def show; end
```
bengolder, line 26: Wow, good catch! We should have removed this before, thanks!

app/models/intake.rb
```diff
@@ -454,6 +454,16 @@ def year_before_most_recent_filing_year
...
   end
 
+  def marital_statuses
+    [
+        ("Married" if married_yes?),
+        ("Lived with spouse" if lived_with_spouse_yes?),
+        ("Legally separated #{separated_year}" if separated_yes?),
+        ("Divorced #{divorced_year}" if divorced_yes?),
+        ("Widowed #{widowed_year}" if widowed_yes?)
+    ].compact.presence || ["Single"]
```
bengolder, line 464: This is really clear, I like how readable this is. 

Also, I would have expected these to be translation strings. Is the a reason we wouldn't want to?

spec/controllers/case_management/clients_controller_spec.rb
```diff
@@ -76,10 +97,21 @@
...
+        expect(profile).to have_text(client.legal_name)
+        expect(profile).to have_text("Intake Language: " + client.intake.locale)
+        expect(profile).to have_text("Phone Language: " + client.intake.preferred_interview_language)
+        expect(profile).to have_text("2019, 2018")
+        expect(profile).to have_text(client.email_address)
+        expect(profile).to have_text(client.phone_number)
+        expect(profile).to have_text("Marital Status: Married, Lived with spouse")
+        expect(profile).to have_text("Filing Status: Filing jointly")
+        expect(profile).to have_text("Oakland, CA 94606")
+        expect(profile).to have_text("Spouse Contact Info")
```
bengolder, line 112: nice!

app/models/intake.rb
bytheway875: We just realized they should be!! Thanks :D

PR #538, opened 2020-10-15
Add filing years and language column to client list

bytheway875:
Filing Years Ticket: https://www.pivotaltracker.com/n/projects/2409240 
Language Ticket: https://www.pivotaltracker.com/n/projects/2409240

### DIFF
--- a/app/forms/interview_scheduling_form.rb
+++ b/app/forms/interview_scheduling_form.rb
@@ -6,7 +6,7 @@ class InterviewSchedulingForm < QuestionsForm
   end
 
   def language_options
-    I18n.backend.translations[I18n.locale][:general][:language].invert
+    I18n.backend.translations[I18n.locale][:general][:language_options].invert
   end
 
   def self.existing_attributes(intake)
--- a/app/views/case_management/clients/index.html.erb
+++ b/app/views/case_management/clients/index.html.erb
@@ -11,6 +11,7 @@
         <th scope="col" class="index-table__header"><%= t("general.name") %></th>
         <th scope="col" class="index-table__header"><%= t(".client_id_heading") %></th>
         <th scope="col" class="index-table__header"><%= t("general.partner") %></th>
+        <th scope="col" class="index-table__header"><%= t("general.filing_years") %></th>
         <th scope="col" class="index-table__header"><%= t("general.updated_at") %></th>
       </tr>
     </thead>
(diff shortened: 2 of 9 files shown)


app/forms/interview_scheduling_form.rb
```diff
@@ -6,7 +6,7 @@ def save
   end
 
   def language_options
-    I18n.backend.translations[I18n.locale][:general][:language].invert
+    I18n.backend.translations[I18n.locale][:general][:language_options].invert
```
bytheway875, line 9: I updated language list to language_options in order to define a translation for language directly.

bengolder: That's a helpful change

app/views/case_management/clients/index.html.erb
```diff
@@ -24,7 +26,9 @@
             <% end %>
           </th>
           <td class="index-table__cell"><%= client.id %></td>
-          <td class="index-table__cell"><%= client.intakes.exists? ? client.intakes.first.vita_partner.display_name : "No intake for this client" %></td>
+          <td class="index-table__cell"><%= client.intake&.filing_years&.join(", ") || "N/A" %></td>
+          <td class="index-table__cell"><%= client.intake&.vita_partner&.display_name || "No intake for this client" %></td>
+          <td class="index-table__cell"><%= client.intake&.locale ? t("general.language_options.#{client.intake&.locale}") : "N/A" %></td>
```
bengolder, line 31: Should these hardcoded strings have translation keys?

Reminder for the easy way to make new keys:

1. Write the key you want in the template: ` : t(".no_intake")`
2. Put your cursor on the key `".no_intake"` and hit option Enter and it will let you create a key in `en.yml`
3. Keep your cursor on the key and hit option Enter again and it will create another key in `es.yml`
4. Command click on the key will now let you jump to the English or Spanish key

spec/factories/intakes.rb
```diff
@@ -269,11 +269,13 @@
...
-        Intake.defined_enums.keys.each_with_object({}) do |key, hash|
-          intake[key] = Intake.send(key.pluralize).keys.sample
+        Intake.defined_enums.each_key do |key|
+          # only randomize values for keys that have not been supplied
+          unless intake[key]
+            intake[key] = Intake.send(key.pluralize).keys.sample
+          end
         end
       end
     end
```
bengolder, line 279: Totally appreciate that you did the work to keep this code from breaking. I think we should probably just delete it though. Sarah & I are both inclined to remove randomization from the factories.

bengolder (approved):
Looks great! Maybe needs translation keys though

bytheway875:
Added translation keys for blank values!

spec/factories/intakes.rb
bytheway875: There are definitely tests that rely on values for each enumerable being created (the csv specs in particular), so maintaining it and making it work for situations that rely on it and those that need to be more specific seemed like the lowest-impact way forward right now.

PR #540, opened 2020-10-15
Touch updated_at on client for significant relationships and add updated at to client view

bytheway875:
PT Ticket: https://www.pivotaltracker.com/story/show/175214024

### DIFF
--- a/app/helpers/time_helper.rb
+++ b/app/helpers/time_helper.rb
@@ -11,4 +11,8 @@ module TimeHelper
   def timezone_select_options
     ActiveSupport::TimeZone.us_zones.map { |tz| [tz.name, tz.tzinfo.name] }
   end
+
+  def formatted_datetime(datetime)
+    datetime.strftime("%b %-d, %Y %l:%M %p %Z").strip
+  end
 end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -32,7 +32,7 @@ class Document < ApplicationRecord
   scope :of_type, ->(type) { where(document_type: type) }
 
   belongs_to :intake, optional: true
-  belongs_to :client, optional: true
+  belongs_to :client, optional: true, touch: true
   belongs_to :documents_request, optional: true
   has_one_attached :upload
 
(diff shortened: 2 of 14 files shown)


app/helpers/time_helper.rb
```diff
@@ -11,4 +11,8 @@ def formatted_time(datetime)
   def timezone_select_options
     ActiveSupport::TimeZone.us_zones.map { |tz| [tz.name, tz.tzinfo.name] }
   end
+
+  def formatted_datetime(datetime)
+    datetime.strftime("%b %-d, %Y %l:%M %p %Z").strip
+  end
```
bengolder, line 17: Nice! It's good to see this in a helper!

app/models/document.rb
```diff
@@ -32,7 +32,7 @@ class Document < ApplicationRecord
   scope :of_type, ->(type) { where(document_type: type) }
 
   belongs_to :intake, optional: true
-  belongs_to :client, optional: true
+  belongs_to :client, optional: true, touch: true
```
bengolder, line 35: Cool! I didn't know you could tell an object to touch an associated object automatically!

spec/helpers/time_helper_spec.rb
```diff
@@ -6,4 +6,14 @@
       expect(helper.timezone_select_options).to include(["Pacific Time (US & Canada)", "America/Los_Angeles"])
     end
   end
+
+  describe "#formatted_datetime" do
+    it "returns a string formatted to show the date and the time with zone" do
+      Time.use_zone("America/Los_Angeles") do
+        # in_time_zone coerces from -0800 -> PST
+        test_date = Time.zone.local(2004, 11, 24, 01, 04, 44).in_time_zone("America/Los_Angeles")
```
bengolder, line 14: I'm confused, what does the comment mean? Isn't -0800 the same as PST? It looks like it's converting from PST to PST. Was it returning a string with `-0800` before using `.in_time_zone`?
I don't know if a DateTime would behave any differently than Time, but it might be helpful to use a DateTime for specs since that is mostly the type it will be receiving from an ActiveRecord field.

bengolder (approved):
Great!

spec/helpers/time_helper_spec.rb
bytheway875: Yes, its converting the string of -0800 instead of PST and thats what the comment indicates. It's not an ideal way to create the date, but I didn't want to noodle on it too much for purposes getting a test going and trying to create a time value that mimicked what we're seeing.

PR #541, opened 2020-10-16
Creating a 1:1 relationship between Intake and Client

bengolder:

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -2,52 +2,24 @@
 #
 # Table name: clients
 #
-#  id               :bigint           not null, primary key
-#  email_address    :string
-#  phone_number     :string
-#  preferred_name   :string
-#  sms_phone_number :string
-#  created_at       :datetime         not null
-#  updated_at       :datetime         not null
-#  vita_partner_id  :bigint
-#
-# Indexes
-#
-#  index_clients_on_vita_partner_id  (vita_partner_id)
-#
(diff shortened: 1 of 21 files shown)


app/models/client.rb
```diff
@@ -2,52 +2,24 @@
...
 #
 class Client < ApplicationRecord
   belongs_to :vita_partner, optional: true
-  has_many :intakes
+  has_one :intake
   has_many :outgoing_text_messages
   has_many :outgoing_emails
   has_many :incoming_text_messages
   has_many :incoming_emails
-  has_many :documents
```
bytheway875, line 29: @bengolder @sniemeyer13 
Since documents can be added in multiple different contexts (during the original intake process, created from an attachment on an email or text, or added as a requested-later request from the VITA volunteer, maybe we should consider keeping documents’ direct relationship to client. (client has_many documents, document belongs to client.)

++: Per ticket https://www.pivotaltracker.com/story/show/175214024, we’ll need to touch client.updated_at any time an associated document is created or updated. We could do this by touching intake.updated_at on doc create/update, then let the touch on intake cascade to the client, but cutting out the middleman would save us a db update on the intake :)

—: Historical clients have documents that relate to their intake, so creating a client from an intake on demo would require associating the documents to the client. Since we still have to handle creating a client for any existing intake, it’s probably not too much extra effort to associate the documents with the new client on creation.

PR #542, opened 2020-10-16
Simplify date format per product guidance

bytheway875:
Remove timezone and year from date formatter per Kelly.

### DIFF
--- a/app/helpers/time_helper.rb
+++ b/app/helpers/time_helper.rb
@@ -13,6 +13,6 @@ module TimeHelper
   end
 
   def formatted_datetime(datetime)
-    datetime.strftime("%b %-d, %Y %l:%M %p %Z").strip
+    datetime.strftime("%b %d %-l:%M %p")
   end
 end
--- a/spec/helpers/time_helper_spec.rb
+++ b/spec/helpers/time_helper_spec.rb
@@ -10,9 +10,8 @@ describe TimeHelper do
   describe "#formatted_datetime" do
     it "returns a string formatted to show the date and the time with zone" do
       Time.use_zone("America/Los_Angeles") do
-        # in_time_zone coerces from -0800 -> PST
-        test_date = Time.zone.local(2004, 11, 24, 01, 04, 44).in_time_zone("America/Los_Angeles")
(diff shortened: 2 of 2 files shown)


app/helpers/time_helper.rb
```diff
@@ -13,6 +13,6 @@ def timezone_select_options
   end
 
   def formatted_datetime(datetime)
-    datetime.strftime("%b %-d, %Y %l:%M %p %Z").strip
+    datetime.strftime("%b %d %l:%M %p")
```
bytheway875, line 16: strip was really extraneous here... there's nothing it should ever have to strip based on this time format string.

spec/helpers/time_helper_spec.rb
```diff
@@ -10,9 +10,10 @@
   describe "#formatted_datetime" do
     it "returns a string formatted to show the date and the time with zone" do
       Time.use_zone("America/Los_Angeles") do
-        # in_time_zone coerces from -0800 -> PST
-        test_date = Time.zone.local(2004, 11, 24, 01, 04, 44).in_time_zone("America/Los_Angeles")
-        expect(helper.formatted_datetime(test_date)).to include("Nov 24, 2004  1:04 AM PST")
+        test_date = DateTime.new(2004, 11, 24, 01, 04, 44)
+        expect(helper.formatted_datetime(test_date)).to include("Nov 24  1:04 AM")
```
bytheway875, line 14: The behavior here is kind of weird... there is an extra space before one digit integers in order to make sure the strings are formatted consistently. I found this weird, and usually you can add an underscore between the % and the letter to say remove the extra spacing. However, after some research it looks like that behavior is not implemented on DateTime striptime for whatever odd reason. Anyways, I added a test for both 2 and one digit hours to try to document that both are what we expect.

sartan: I don't mind this, however if you use `%-l` you could remove the extra space if you wanted to.

`datetime.strftime("%b %d %-l:%M %p")`

sartan (approved):
Looks great!

spec/helpers/time_helper_spec.rb
bytheway875: @sartan GAH thank you!! I kept trying to do %_l which is "use spaces for padding" instead of "don't pad"

PR #543, opened 2020-10-16
Add last_response_at field to client and touch when client-provided data is updated

bytheway875:


app/models/client.rb
```diff
@@ -33,6 +34,8 @@ def intake
     intakes && intakes.first
   end
 
+  before_create { self.last_response_at = created_at }
```
bytheway875, line 37: The migration sets the default value to a "dynamic value" of CURRENT_TIMESTAMP, and per the rails docs dynamic values only get set on _existing_ columns. :( So we need to manually set it on creation for new clients. To me, it made sense to just set all three (created_at, updated_at, last_response_at) to the created_at time. 

I also considered that we could leave this value null until the client completes the intake, but dealing with null values seemed more dicey/difficult to manage. However, willing to consider a different value here.

app/models/document.rb
```diff
@@ -37,6 +37,7 @@ class Document < ApplicationRecord
   has_one_attached :upload
 
   before_save :set_display_name
+  before_create { client&.touch(:last_response_at) }
```
bytheway875, line 40: Only on create because we are allowing volunteers to update attributes on the document. Currently, volunteers can't add documents but if we change to allowing them to add docs then we'll need to revise the logic.

app/models/intake.rb
```diff
@@ -200,6 +200,9 @@ class Intake < ApplicationRecord
   belongs_to :vita_partner, optional: true
   belongs_to :triage_source, optional: true, polymorphic: true
 
+  attribute :updater
```
bytheway875, line 203: There is a ticket in this sprint to allow volunteers to update the intake. When we do that, we can add an attribute of :updater in the controller with their user object, and when that is passed it will indicate to us that it is a _volunteer_ updating the intake instead of the client.

Another potential approach was to, again, to only update the last_responded_at once when the client completed the intake, and then never update it after with the assumption that all of those updates would be from volunteers. Open to refactoring.

db/migrate/20201016195233_add_last_responded_at_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastRespondedAtToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :clients, :last_response_at, :datetime, null: false, default: -> { 'CURRENT_TIMESTAMP' }
```
bytheway875, line 3: When you pass a proc as the default, it sets a default for all _existing_ rows. This isn't exactly _the most accurate_ thing, but since we only have a handful of entries in demo and none in prod it seemed simpler and just as (in)accurate as any other value!

app/models/client.rb
jenny-heath: hm, just to make sure i understand: when you say "we need to manually set it" you're referring to this line?

bytheway875: Yep, thats what's happening here!

app/models/intake.rb
jenny-heath: interesting! also relevant: this seems like it overlaps with the audit trail work, since we'll want to keep a record of who updates client information when. if we don't refactor now, we could do it later when that work happens (in the event that the way we implement the auditing conflicts with or is redundant with this).

this way is fine with me for now, but i also don't have the energy to think through a decision like this at the end of friday 🤣

app/models/document.rb
sartan: That makes sense.

app/models/incoming_email.rb
```diff
@@ -31,6 +31,8 @@ class IncomingEmail < ApplicationRecord
 
   belongs_to :client, touch: true
 
+  before_create { client.touch(:last_response_at) }
```
sartan, line 34: This makes sense since after creating these incoming messages they are not updated (i.e. with status).

sartan (approved):
Looks great! Specs clearly describe the `last_response_at` behavior.

PR #546, opened 2020-10-20
Add icon for clients who need response from volunteer

sartan:

image

image

### DIFF
--- a/app/models/incoming_text_message.rb
+++ b/app/models/incoming_text_message.rb
@@ -21,11 +21,11 @@
 class IncomingTextMessage < ApplicationRecord
   include ContactRecord
 
-  belongs_to :client, touch: true
+  belongs_to :client
   validates_presence_of :body
   validates_presence_of :received_at
 
-  before_create { client.touch(:last_response_at) }
+  before_create { client.touch(:updated_at, :last_response_at) }
 
   def datetime
     received_at
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -23,6 +23,71 @@
 require "rails_helper"
 
 describe Client do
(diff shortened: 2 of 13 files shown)


spec/models/client_spec.rb
```diff
@@ -23,6 +23,71 @@
 require "rails_helper"
 
 describe Client do
+  describe "#needs_attention" do
+
+    context "when last_response_at is nil" do
+      let!(:client) { create :client }
+
+      it "doesn't not need attention" do
```
alaroia-cfa, line 31: > doesn't not

should be

> doesn't

app/models/incoming_text_message.rb
```diff
@@ -21,11 +21,11 @@
 class IncomingTextMessage < ApplicationRecord
   include ContactRecord
 
-  belongs_to :client, touch: true
+  belongs_to :client
   validates_presence_of :body
   validates_presence_of :received_at
 
-  before_create { client.touch(:last_response_at) }
+  before_create { client.touch(:updated_at, :last_response_at) }
```
alaroia-cfa, line 28: I personally found it confusing to explicitly update `updated_at` here, and would prefer just `client.touch(:last_response_at)`, but this is non-blocking and if you like it more this way, then I can 100% live with it this way!

spec/models/client_spec.rb
alaroia-cfa: Resolved by adding a commit on GitHub

alaroia-cfa (approved):
Great! Makes sense, and great to have this in to make today's work on removing the response-needed UI easier!

alaroia-cfa:
I'm going to merge this, in the interest of latency compensation!

PR #549, opened 2020-10-22
Temporarily load documents through intake to make documents request uploads reflect in ui

bytheway875:
This is a quick and dirty approach to making documents accessible through client but still stored on the intake to prevent a huge refactor but open up acceptance on new document upload tickets.

The client association was creating ONE copy of documents at the time the client was created, but we need to see documents uploaded by the user later reflected in the UI. 

If we want the relationship to be from client -> documents, we should eventually (hopefully in the shor term) work on this more to make it "right", but wanted to get a stop gap in so that the testers can see new uploads reflected in their documents list.

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -24,19 +24,17 @@
 #
 class Client < ApplicationRecord
   belongs_to :vita_partner, optional: true
-  has_many :intakes
+  has_one :intake
+  # has_many :documents
   has_many :outgoing_text_messages
   has_many :outgoing_emails
   has_many :incoming_text_messages
   has_many :incoming_emails
-  has_many :documents
   has_many :notes
   has_many :tax_returns
 
-  def intake
-    intakes && intakes.first
-  end
-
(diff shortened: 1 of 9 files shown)


app/models/client.rb
```diff
@@ -24,19 +24,17 @@
 #
 class Client < ApplicationRecord
   belongs_to :vita_partner, optional: true
-  has_many :intakes
+  has_one :intake
+  # has_many :documents
```
bengolder, line 28: why leave this commented out? is it to indicate that it exists

app/models/client.rb
```diff
@@ -24,19 +24,17 @@
...
   has_many :incoming_emails
-  has_many :documents
   has_many :notes
   has_many :tax_returns
 
-  def intake
-    intakes && intakes.first
-  end
-
+  delegate :documents, to: :intake
```
bengolder, line 36: smart!

bengolder (approved):
Looks good!

PR #551, opened 2020-10-22
Only allow access to clients from same VITA partner as the user

sartan:


sartan:
The more eyes on this one, the better. If nothing else, I would at least revisiting the test and refactoring the `vita_partner: vita_partner` all over the place. It was late and I was on autopilot.

PR #552, opened 2020-10-23
GYR admin should have access to all clients in GYR

sartan:


app/controllers/users_controller.rb
```diff
@@ -25,6 +25,6 @@ def update
   private
 
   def user_params
-    params.require(:user).permit(:is_beta_tester, :vita_partner_id, :timezone)
+    params.require(:user).permit(*(:is_super_admin if current_user.is_super_admin?), :is_beta_tester, :vita_partner_id, :timezone)
```
sartan, line 28: `(*(:is_super_admin if current_user.is_super_admin?)...`

Needlessly clever? Probably.

jenny-heath: never seen the syntax but assuming that it doesn't pass in any argument if `current_user.is_super_admin?` is false?

bytheway875:
in tax/returns edit, we should make sure to get the collection of assignable users from the intake's vita_partner instead of the admin's vita partner so that a superadmin can still assign!!

app/controllers/users_controller.rb
sartan:
```
puts(*(:is_super_admin if true), :is_beta_tester, :vita_partner_id, :timezone)
irb(main):001:0> puts(*(:is_super_admin if true), :is_beta_tester, :vita_partner_id, :timezone)
is_super_admin
is_beta_tester
vita_partner_id
timezone
=> nil


irb(main):003:0> puts(*(:is_super_admin if false), :is_beta_tester, :vita_partner_id, :timezone)
is_beta_tester
vita_partner_id
timezone
=> nil


irb(main):004:0> puts(*(:is_super_admin if nil), :is_beta_tester, :vita_partner_id, :timezone)
is_beta_tester
vita_partner_id
timezone
=> nil
```

PR #555, opened 2020-10-26
Remove duplicate PII from Client table

bengolder:


app/controllers/case_management/clients_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
       # Don't create additional clients if we already have one
       return redirect_to case_management_client_path(id: intake.client_id) if intake.client_id.present?
 
-      client = Client.create_from_intake(intake)
+      client = Client.create!
```
bytheway875, line 21: could also do Client.create!(intake: intake) which will associate the intake object to the client on creation, and remove line 22.

bytheway875: Also, if we are leaving vita_partner_id on client, we def should create the client with the vita_partner_id of the intake!

```Client.create!(intake: intake, vita_partner_id: intake.vita_partner_id)```

bytheway875 (changes requested):
We need to set vita_partner_id onto the client, I think!

app/controllers/case_management/clients_controller.rb
sniemeyer13: Thanks! we did exactly as you suggested, good catch

sniemeyer13:
Incorporated comments and added a commit directly to master: https://github.com/codeforamerica/vita-min/commit/b063602f6dc897973390012f965b60b21245dac5

PR #556, opened 2020-10-27
Extract assignee logic to partial and style

bytheway875:
Extracted assigning list into a partial and  use it in
Screen Shot 2020-10-26 at 5 04 54 PM
Screen Shot 2020-10-26 at 5 05 05 PM

 the client header.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -55,7 +55,6 @@ en:
         client_id_heading: Client ID
         filing_year: Filing Year
         title: All clients
-        assign: Assign
         needs_response_heading: Needs Response?
       navigation:
         client_documents: Documents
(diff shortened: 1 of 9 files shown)


alaroia-cfa (approved):
Great work! Very easy to review w/ the screenshots and the very straightforward code.

config/locales/en.yml
```diff
@@ -190,6 +189,7 @@ en:
     address: Address
     admin: Admin
     affirmative: 'Yes'
+    assign: Assign
```
alaroia-cfa, line 192: A non-blocking question that just came to me: is it possible to store this in I18n key in a place where `<%= t('.assign') %>` would have worked?

I have no real idea if it is possible, hence the question.

Cheers!

bytheway875: _maybe_ if we had a section for partials? im not sure if partials become "part of the page" in terms of the path for template strings or reconcile through the partial path.

PR #557, opened 2020-10-27
Only install default and Rails.env groups of gems 

bytheway875:
I _think_ we can avoid some of the issues where Listen gem and Spring is trying to load in demo rails console and causing errors by scoping the bundle installed gems to the :default group and the Rails.env group.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,7 +17,8 @@ require "rails/test_unit/railtie"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
-Bundler.require(*Rails.groups)
+# Bundler.require(*Rails.groups)
+Bundler.require(:default, Rails.env)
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
To test if this is going to really change things, you could `git push` it to the `aptible-staging` remote, and validate manually that `staging` doesn't have the development stuff. @bytheway875 Does that sound good to you? I'm happy to help with that if desired.

Then we can be super confident it does something before merging.

bytheway875:
@alaroia-cfa I closed this PR because I'm pretty sure the installation of Gems is happening in the bundle install step in the docker file, so I was going to try updating _that_ on the vita-min-staging site... it's not super high priority on my list for today since it's not user-facing and the demo rails c isn't blocking us as consistently post-Tuesday, but I'm hoping to get to it this week.

PR #559, opened 2020-10-28
Don't install test or development gems in docker instances

bytheway875:
In order to prevent test and development gems from being installed in our docker containers, we can add the --without flag to our bundle install. If these gems are not installed, they won't affect our rails console setup, meaning that the FATAL error from the Listen gem (used by Spring preloader) trying to index all of the files should stop affecting us.

See screenshots that show that show that Spring is no longer loading after the change.
Screen Shot 2020-10-28 at 8 44 07 AM
Screen Shot 2020-10-28 at 10 32 17 AM
Screen Shot 2020-10-28 at 10 32 58 AM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN curl -fsSLO "$SUPERCRONIC_URL" \
 ADD Gemfile /app/
 ADD Gemfile.lock /app/
 WORKDIR /app
-RUN bundle install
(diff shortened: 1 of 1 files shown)


bengolder (approved):
Thanks for testing this out on staging!
We still wonder why bundle install installs all the gems instead of just the environment and default. But this seems helpful.
-- Ben & Jenny

PR #562, opened 2020-10-28
Add tab for viewing your assigned clients

bytheway875:
This is now the "root" page when users log in, as discussed at Sprint Planning.

Screen Shot 2020-10-28 at 3 50 48 PM

Screen Shot 2020-10-28 at 3 50 55 PM

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -236,7 +236,7 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_in_path_for(_user)
-    user_profile_path
+    case_management_root_path
   end
 
   def set_time_zone
--- a/app/controllers/case_management/assigned_clients_controller.rb
+++ b/app/controllers/case_management/assigned_clients_controller.rb
@@ -1,0 +1,17 @@
+module CaseManagement
+  class AssignedClientsController < ApplicationController
+    include AccessControllable
+    include ClientSortable
+
+    before_action :require_sign_in
+    before_action :setup_sortable_client, only: [:index]
+    load_and_authorize_resource :client, parent: false
+    layout "admin"
+
+    def index
+      @page_title = I18n.t("case_management.assigned_clients.index.title")
+      @clients = @clients.assigned_to(current_user).delegated_order(@sort_column, @sort_order)
+      render "case_management/clients/index"
+    end
+  end
+end
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -34,6 +34,9 @@ class Client < ApplicationRecord
   end
 
   delegate *delegated_intake_attributes, to: :intake
+
+  scope :assigned_to, ->(user) { joins(:tax_returns).where({ tax_returns: { assigned_user_id: user } }) }
+
   scope :delegated_order, ->(column, direction) do
     raise ArgumentError, "column and direction are required" if !column || !direction
 
--- a/app/views/case_management/clients/index.html.erb
+++ b/app/views/case_management/clients/index.html.erb
@@ -1,9 +1,11 @@
-<% content_for :page_title, t(".title") %>
+<% content_for :page_title, @page_title %>
 
 <% content_for :card do %>
+  <%= render 'shared/user_header' %>
+  <%= render "shared/dashboard_navigation" %>
   <table class="index-table client-table">
-    <caption class="index-table__title">
-      <h1 class="h2"><%= t(".title") %></h1>
+    <caption class="index-table__title spacing-above-25">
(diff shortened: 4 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -247,6 +247,19 @@ def show_offseason_banner?
...
+    @sort_order = clients_sort_order
+  end
+
+  def clients_sort_order
+    %w[asc desc].include?(params[:order]) ? params[:order] : "desc"
+  end
+
+  def clients_sort_column
+    %w[preferred_name id updated_at locale].include?(params[:column]) ? params[:column] : "id"
+  end
```
bengolder, line 261: If this is used by a few Controllers, but not most, would it feel cleaner to put this in a `controllers/concerns/client_sortable.rb` module or something similar that can be mixed in to different controllers?

app/controllers/case_management/assigned_clients_controller.rb
```diff
@@ -0,0 +1,16 @@
+module CaseManagement
+  class AssignedClientsController < ApplicationController
+    include AccessControllable
+
+    before_action :require_sign_in
+    before_action :setup_sortable_client, only: [:index]
```
bengolder, line 6: I like that this got refactored out into shared code!

app/controllers/case_management/assigned_clients_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+    load_and_authorize_resource :client, parent: false
+    layout "admin"
+
+    def index
+      @page_title = I18n.t("case_management.assigned_clients.index.title")
```
bengolder, line 11: This breaks from the pattern that we've used in other views to set this variable in the template. Do you have a preference for setting it in the controller?

app/controllers/case_management/assigned_clients_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+      @clients = @clients.assigned_to(current_user).delegated_order(@sort_column, @sort_order)
```
bengolder, line 12: This came out really clean & concise from the new scope & `setup_sortable_client` method!

app/models/client.rb
```diff
@@ -35,6 +35,9 @@ def self.delegated_intake_attributes
 
   delegate :documents, to: :intake
   delegate *delegated_intake_attributes, to: :intake
+
+  scope :assigned_to, ->(user) { joins(:tax_returns).where({ tax_returns: { assigned_user_id: user } }) }
+
```
bengolder, line 40: This looks really useful. I'm also wondering if there's a way to set up a `has_many ... through: tax_returns` on the User model.

app/views/case_management/clients/index.html.erb
```diff
@@ -1,9 +1,11 @@
-<% content_for :page_title, t(".title") %>
+<% content_for :page_title, @page_title %>
```
bengolder, line 1: nice! now we can reuse with different titles! If we have a search or some other kind of client list, I could imagine us making use of this in those controllers too.

config/locales/en.yml
```diff
@@ -50,6 +50,10 @@ en:
         someone_invited_you: "%{their_name} (%{their_email}) has invited %{your_name} to create an account on GetYourRefund to help people with their taxes."
         subject: You've been invited to GetYourRefund
   case_management:
+
```
bengolder, line 53: a harmless extra newline. If you ever want to clean up the format of the translation files, you can run `i18n-tasks normalize` on the command line and it will alphabetize and organize all the keys.

config/routes.rb
```diff
@@ -128,6 +128,7 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
 
     # New Case Management Admin routes
     namespace :case_management do
+      root "assigned_clients#index"
```
bengolder, line 131: Nice! I like that this ended up as root!

spec/controllers/case_management/assigned_clients_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expect(assigns(:clients)).not_to include not_assigned_to_me
+    end
+
+    it "renders the clients table view" do
+      get :index
+
+      expect(response).to render_template "clients/index"
+    end
+  end
+end
```
bengolder, line 30: perfect tests

bengolder (approved):
Looks great!

app/controllers/application_controller.rb
bytheway875: good idea!

PR #566, opened 2020-10-29
Create CSV for IRS Impact Analysis 

bytheway875:
Uses fields outlined in PT ticket

### DIFF
--- a/app/helpers/eitc_zendesk_instance.rb
+++ b/app/helpers/eitc_zendesk_instance.rb
@@ -78,7 +78,7 @@ class EitcZendeskInstance
   EIP_STATUS_ID_UPLOAD = "reached_id_upload_page"
   EIP_STATUS_SUBMITTED = "submitted_eip_only_form"
   EIP_STATUS_READY_FOR_PHONE_CALL = "ready_for_phone_call"
-  EIP_STATUS_PREP_IN_PROGRESSS = "eip_prep_in_progress"
+  EIP_STATUS_PREP_IN_PROGRESS = "eip_prep_in_progress"
   EIP_STATUS_SIGNATURE = "waiting_on_signature"
   EIP_STATUS_COMPLETED = "completed_eip_return"
 
@@ -87,7 +87,7 @@ class EitcZendeskInstance
     EIP_STATUS_ID_UPLOAD => "Reached ID upload page",
     EIP_STATUS_SUBMITTED => "Submitted EIP only form",
     EIP_STATUS_READY_FOR_PHONE_CALL => "Ready for phone call",
-    EIP_STATUS_PREP_IN_PROGRESSS => "EIP prep in progress",
+    EIP_STATUS_PREP_IN_PROGRESS => "EIP prep in progress",
     EIP_STATUS_SIGNATURE => "Waiting on signature",
     EIP_STATUS_COMPLETED => "Completed EIP return",
   }
--- a/app/services/irs_impact_analysis_csv_service.rb
+++ b/app/services/irs_impact_analysis_csv_service.rb
@@ -17,7 +17,7 @@ class IrsImpactAnalysisCsvService
 
   # only grab intakes where the user consented.
   def intakes
-    Intake.where.not(primary_consented_to_service_at: nil).includes(:ticket_statuses)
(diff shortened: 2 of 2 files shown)


app/services/irs_impact_analysis_csv_service.rb
```diff
@@ -0,0 +1,103 @@
...
+    CSV.generate(headers: csv_headers, write_headers: true) do |csv|
+      intakes.find_each(batch_size: 100) do |intake|
+        csv << csv_row(decorated_intake(intake))
+      end
+    end
+  end
+
+  def send_csv
+    # TODO: Whats the best way to download this document?
+  end
```
bengolder, line 33: Now that we have Mailgun integrated, it might be easiest to send this as an email attachment.

app/helpers/eitc_zendesk_instance.rb
```diff
@@ -87,7 +87,7 @@ class EitcZendeskInstance
     EIP_STATUS_ID_UPLOAD => "Reached ID upload page",
     EIP_STATUS_SUBMITTED => "Submitted EIP only form",
     EIP_STATUS_READY_FOR_PHONE_CALL => "Ready for phone call",
-    EIP_STATUS_PREP_IN_PROGRESSS => "EIP prep in progress",
+    EIP_STATUS_PREP_IN_PROGRESS => "EIP prep in progress",
```
bengolder, line 90: Thanks for catching this! Do we use that misspelled constant anywhere?

app/services/irs_impact_analysis_csv_service.rb
```diff
@@ -0,0 +1,103 @@
...
+        key = EitcZendeskInstance::RETURN_STATUS_COMPLETED_RETURNS
+        verified_ticket_statuses.detect { |ts| ts.return_status == key }.present?
+      end
+    end
+
+    def return_status
+      # TODO: Can we get the return status? Kelly said there is a status for this on Zendesk called "Status After Filing",
+      # but I wasn't able to find it.
+      "??"
+    end
```
bengolder, line 93: I think we might have to look into this with Kelly

bengolder: I'm sure we could get it if we were willing to iterate over all the tickets and scrape data from Zendesk, but that could be tricky due to their API limits.

app/services/irs_impact_analysis_csv_service.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    def date_of_birth
+      primary_birth_date.strftime("%m/%d/%Y")
+    end
+
+    def reached_gathering_docs
+      key = EitcZendeskInstance::INTAKE_STATUS_GATHERING_DOCUMENTS
+      verified_ticket_statuses.detect { |ts| ts.intake_status == key }.present?
+    end
```
bengolder, line 69: I think this is the right approach. But reading it, I realize that we should give the caveat that this means the ticket _reached this status at some point_ and so theoretically could include tickets that were accidentally given this status and then reverted to a prior status.

bytheway875: does the verified: true cover the transition back and forth?

app/helpers/eitc_zendesk_instance.rb
bytheway875: No, these were the only direct usages in the application.

app/services/irs_impact_analysis_csv_service.rb
bytheway875: got the okay from Kelly to leave this off!

app/services/irs_impact_analysis_csv_service.rb
bengolder: no, verified true tells us if we know the _time_ that it reached that status. verified false means that we know it was in that status at some point, but we don't know when it _reached_ the status.

bengolder (approved):
Looks good!

app/services/irs_impact_analysis_csv_service.rb
bytheway875: okay, ill let kelly + Day know about the caveat!

PR #568, opened 2020-10-30
Move attachment concerns from controller into TwilioService

sartan:


bengolder (approved):
Definitely a helpful refactor, thank you!

PR #569, opened 2020-10-30
Create 2020-10-30-mailgun-web-hook-security.md

alaroia-cfa:
I'd like to merge this into the repo as a record of the research I did, so that people can know why we're taking the approach we will on a future story.

Seems like maybe a waste of space to have it in the repo, but I'm hopeful we can add more docs here and that we can feel good about doing so willy-nilly, so I'm hopeful someone will approve it (or suggest changes if they're essential)!


bengolder (approved):
Thanks for including this!

PR #571, opened 2020-11-06
Add statuses and form to edit them

jenny-heath:
Screen Shot 2020-11-09 at 2 50 41 PM
Screen Shot 2020-11-09 at 2 50 50 PM


app/helpers/vita_min_form_builder.rb
```diff
@@ -142,7 +142,7 @@ def vita_min_searchbar(method, label_text, label_icon: "", options: {}, classes:
     return <<~HTML.html_safe
       <div class="vita-min-searchbar form-group#{error_state(object, method)}" role="search">
         <div class="vita-min-searchbar__field">
-          <label class="vita-min-searchbar__label sr-only" for="#{object_name}_#{method}">#{label_text}</label>
+          <label class="vita-min-searchbar__label sr-only.scss" for="#{object_name}_#{method}">#{label_text}</label>
```
jenny-heath, line 145: is the `.scss` how this works? somehow i didn't see it until reviewing just now...

app/models/tax_return.rb
```diff
@@ -23,4 +24,94 @@
...
+
+  def self.stages
+    STAGES
+  end
+
+  def self.statuses_for(stage_key)
+    STATUSES[stage_key]
+  end
+
+  def self.grouped_status_options
```
jenny-heath, line 93: There seems to be an odd caching thing happening here - when I switch locales the options in the select stay the same but when I put a breakpoint in this method they changed. If no one has a suggestion I may try moving this into a view helper next.

config/locales/en.yml
```diff
@@ -1332,7 +1357,7 @@ en:
       spouse_identity:
         info_later: If not, we can send them a sign in link through text or e-mail.
         info_now: If your spouse is sitting with you, you can go ahead and sign in with ID.me
-        sign_in_html: Sign in spouse with <span class="sr-only">ID.me</span>
+        sign_in_html: Sign in spouse with <span class="sr-only.scss">ID.me</span>
```
jenny-heath, line 1360: same question

app/helpers/vita_min_form_builder.rb
bytheway875: oh, this should just be sr-only. I think this is a byproduct of using the refactor tool to rename the file that we forgot to add an extension to.

app/models/tax_return.rb
bytheway875: I think this makes more sense as a view helper anyway.

bytheway875: refactored into a helper.

app/models/tax_return.rb
```diff
@@ -23,4 +24,32 @@
...
+  INTAKE = "intake".freeze
+  PREP = "prep".freeze
+  REVIEW = "review".freeze
+  FINALIZE = "finalize".freeze
+  FILED = "filed".freeze
+  
+  STAGES = [INTAKE, PREP, REVIEW, FINALIZE, FILED].freeze
+
+  enum status: {
+    intake_in_progress: 101, intake_open: 102, intake_review: 103, intake_more_info: 104, intake_info_requested: 105, intake_needs_assignment: 106,
```
bytheway875, line 37: I considered using the consts in order to make it more clear that these mapped to a stage -- 
```"#{INTAKE}_in_progress" => 101```

but I wasn't sure if it was useful/better/could have side effects?

File: app/helpers/tax_return_status_helper.rb [hunk 5]
```diff
@@ -0,0 +1,28 @@
...
+      statuses.map do |status, _|
+        status_options.push([translated_status(status), status])
+      end
+      options << select
+    end
+    options
+  end
+
+  def translated_status(status)
+    I18n.t("case_management.tax_returns.status.#{status}")
```
bytheway875, line 18: Constructing the translation string like this ensures that all of the status translations are marked as used, and same with the stage helper. In order to better ensure that we really have translations, I added some tests for checking for translations in the tax_return model.

app/controllers/case_management/tax_returns_controller.rb
```diff
@@ -12,7 +12,7 @@ class TaxReturnsController < ApplicationController
     def edit; end
 
     def update
-      @tax_return.update!(tax_return_params)
+      @tax_return.update!(assign_params)
```
alaroia-cfa, line 15: If the method was called `tax_return_params` or `update_params` then CanCanCan should automatically have assigned the new attributes (for example calling `@tax_return.assign_attributes(update_params)`) In which case yo'd only have to call `@tax_return.save!`

bytheway875: 1) i didn't know that and thats really cool. 
2) this controller has TWO update methods -- one that only updates the status and one that only updates the assigned user. I kind of thing it makes sense to have them in separate controllers, so I could potentially refactor into two separate controllers.

app/views/shared/_tax_return_assignees.html.erb
```diff
@@ -1,16 +1,25 @@
+<% status_updateable ||= false %>
```
alaroia-cfa: This isn't a pattern I see a lot in our codebase. I wonder if we could make the custom elements for updateable status listing be a `yield`, or make this parameter required. Non-blocking.

spec/controllers/case_management/tax_returns_controller_spec.rb
```diff
@@ -99,5 +99,50 @@
...
+    let(:user) { create :user_with_org }
+    let(:tax_return) { create :tax_return, client: create(:client, vita_partner: user.vita_partner) }
+    let(:params) { { id: tax_return.id, client_id: tax_return.client } }
+
+    it_behaves_like :a_get_action_for_authenticated_users_only, action: :edit_status
+
+    context "as an authenticated user" do
+      before { sign_in user }
+
+      it "returns an ok response" do
```
alaroia-cfa, line 136: I'm a little confused about why this controller method supports both `post` and an `get`. Non-blocking.

spec/features/case_management/tax_return_status_change_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.feature "Change tax return status on a client" do
+  context "As a beta tester" do
+    let(:vita_partner) { create :vita_partner }
+    let(:logged_in_user) { create :user, vita_partner: vita_partner }
```
alaroia-cfa, line 6: Consider `let(:user)` rather than `let(:logged_in_user)` here, since they're not logged-in yet at the time of the `let`. Non-blocking.

spec/features/case_management/tax_return_status_change_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        click_on "Update"
+      end
+
+      expect(current_path).to eq(edit_status_case_management_client_tax_return_path(id: tax_return.id, client_id: tax_return.client.id))
+      expect(page).to have_select("tax_return_status", selected: "Open")
+
+      click_on "Save"
+      expect(current_path).to eq(case_management_client_messages_path(client_id: tax_return.client.id))
+      expect(page).to have_select("tax_return_status", selected: "Open")
```
alaroia-cfa, line 30: This appears to keep the same status over the course of editing it, namely `"Open"`. I'd like to see a test that validates that it saves properly.

app/models/tax_return.rb
alaroia-cfa: I like this the way it is pretty well, and I don't think `#{INTAKE}_` is clearer than `intake_`. I don't think there would be any side-effects from doing it the `#{INTAKE}_` way, though.

One alternative to consider is something like

```
STAGE_AND_STATUS = {
  intake: 
    {in_progress: 101, open: 102, review: 103, ...},
  prep:
    {ready_for_call: 201, more_info: 202, ...},
  ...
  }
```

Then you could compute something fancy with the `Hash[...]` constructor out of that.

Doesn't seem necessarily better, though.

alaroia-cfa (approved):
This is great work!

I think you're right to want a separate controller for these.

Here are some comments and questions. None are blocking.

app/assets/stylesheets/components/_tax-return-assignees.scss
```diff
@@ -1,5 +1,47 @@
 .tax-return-assignees {
+  margin: 0;
   li {
+    display: flex;
```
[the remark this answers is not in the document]
bytheway875: yes!

app/views/shared/_tax_return_assignees.html.erb
```diff
@@ -1,16 +1,25 @@
...
     <li id="tax-return-<%= tax_return.id %>">
-      <span><%= tax_return.year %></span>
-      <span>
-        <% if tax_return.assigned_user.present? %>
-          <%= tax_return.assigned_user&.name %>
-        <% else %>
-          <%= link_to t("general.assign"),
-                      edit_case_management_client_tax_return_path(client_id: client.id, id: tax_return.id),
-                      class: "button button--inline-action" %>
+      <div class="assignment">
```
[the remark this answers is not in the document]
bytheway875: yes, extra divs are required sometimes for styling purposes, especially when using flexbox.

spec/features/case_management/tax_return_status_change_spec.rb
bytheway875: you change the status on the first page, and then it takes you to this other page to persist the change, based on the pivotal tracker ticket. that's why its the same status when you get to the page after changing it on the previous page. but an additional test would make sense.

alaroia-cfa: I didn't see the status change on the earlier lines! Makes sense now.

spec/controllers/case_management/tax_returns_controller_spec.rb
bytheway875: edit status is a view and doesn't support a post.

spec/features/case_management/tax_return_status_change_spec.rb
bytheway875: changed

app/views/shared/_tax_return_assignees.html.erb
bytheway875: I've been trying to treat partials as similar to components in react (wrapped in a class I can style with its own partial _scss file, having required/optional params, etc).

spec/controllers/case_management/tax_returns_controller_spec.rb
alaroia-cfa: I swear I saw `post :edit_status` at some point in a test... ignore that then!

app/assets/stylesheets/components/_tax-return-assignees.scss
```diff
@@ -1,5 +1,47 @@
...
+    align-items: center;
     margin-bottom: 5px;
+    > div {
+      margin-right: 5px;
+    }
+  }
+
+  .assignment {
+    display: flex;
+    flex-grow: 1;
```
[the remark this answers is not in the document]
bytheway875: the LI is flexed, and there are two children: the assignment side and the status side. Setting this to flex-grow: 1 should indicate that the assignment side should fill the extra available space for the section because it is greater than the default on the other side which is flex-grow: 0

PR #572, opened 2020-11-06
Remove beta user

sartan:


sartan:
We're currently using `:user` factory for cases where vita partner is not necessary and `:user_with_org` for cases where it is. Do we want users to always be tied to an org? If so, these two factories will need to be combined.

alaroia-cfa:
> We're currently using `:user` factory for cases where vita partner is not necessary and `:user_with_org` for cases where it is. Do we want users to always be tied to an org? If so, these two factories will need to be combined.

I think yes - we want to always require a `.vita_partner` for our User model. IMHO, we should make `.vita_partner` mandatory, and we should create a Code for America "VitaPartner" to assign ourselves to. IMHO, let's do that in a separate pull request; it'll create another bit of complicated churn.

spec/lib/ability_spec.rb
```diff
@@ -102,25 +102,8 @@
     end
   end
 
-  context "as a non-beta tester" do
```
alaroia-cfa, line 105: So glad this test is going away!

spec/lib/ability_spec.rb
```diff
@@ -135,21 +118,4 @@
       expect(subject.can?(:manage, VitaPartner.new)).to eq true
     end
   end
-
-  context "as a non-beta tester admin" do
```
alaroia-cfa, line 139: Extra glad this test is going away!

alaroia-cfa (approved):
I have minor feedback which is non blocking. This is great work! I'll avoid merging since you've marked it as a draft.

app/views/layouts/admin.html.erb
```diff
@@ -1,16 +1,10 @@
...
-      <%= f.submit value: t("general.sign_out"), class: "button--link" %>
-    <% end %>
-  <% else %>
-    <div class="toolbar__item">
-      <%=t("views.layouts.admin.welcome_user_html", :name => current_user.name) %>
-    </div>
+  <%= form_with url: destroy_user_session_path, local: true, method: :delete, class: "toolbar__item" do |f| %>
+    <%= f.submit value: t("general.sign_out"), class: "button--link" %>
   <% end %>
 <% end if current_user %>
```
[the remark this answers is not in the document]
sartan: When pairing, we discovered that the `if current_user` applies to an entire block of ERB (postfix `if` operator in Ruby), which means "only do this ERB stuff if the user is logged in." We think but haven't checked that the admin template is used in a logged-out context. Seems good enough.

PR #574, opened 2020-11-06
Remove leftover beta tester references

sartan:


spec/channels/client_channel_spec.rb
```diff
@@ -5,8 +5,6 @@
   let(:user) { create :user_with_org }
   let(:params) { { id: client.id }}
 
-  it_behaves_like :a_channel_for_beta_testers, action: :subscribe
```
alaroia-cfa, line 8: should check that it's a logged in user, imho

sartan: Thank you! Adding another commit to handle this.

bytheway875 (approved):
👯  LGTM

PR #575, opened 2020-11-09
Re-enable skipped helper spec

alaroia-cfa:
Also:

- Do not add production VITA partners in CI test execution

- Use transaction in Rake task, not upsert_vita_partners method

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
       - run: dockerize -wait tcp://localhost:5432 -timeout 1m
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
-      - run: bundle exec rake db:insert_states db:upsert_vita_partners
+      - run: bundle exec rake db:insert_states
       - run: mkdir ~/test-results && mkdir ~/test-results/rspec && mkdir ~/test-results/jest
       - run: |
           bundle exec rspec \
--- a/app/lib/vita_partner_importer.rb
+++ b/app/lib/vita_partner_importer.rb
@@ -8,10 +8,8 @@ class VitaPartnerImporter < CliScriptBase
 
     yaml_partners = YAML.load_file(yml)['vita_partners']
     yaml_zendesk_group_ids = yaml_partners.map { |partner| partner["zendesk_group_id"] }
-    VitaPartner.transaction do
-      yaml_partners.each do |yaml_partner|
-        upsert_vita_partner(yaml_partner)
-      end
+    yaml_partners.each do |yaml_partner|
+      upsert_vita_partner(yaml_partner)
     end
     archive_absent_partners(VitaPartner.where.not(zendesk_group_id: yaml_zendesk_group_ids))
     report_progress "  => done"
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - run: dockerize -wait tcp://localhost:5432 -timeout 1m
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
-      - run: bundle exec rake db:insert_states db:upsert_vita_partners
```
alaroia-cfa, line 37: This is the only actually important part of this PR, for what it's worth.

app/lib/vita_partner_importer.rb
```diff
@@ -8,10 +8,8 @@ def self.upsert_vita_partners(yml = VITA_PARTNERS_YAML)
 
     yaml_partners = YAML.load_file(yml)['vita_partners']
     yaml_zendesk_group_ids = yaml_partners.map { |partner| partner["zendesk_group_id"] }
-    VitaPartner.transaction do
```
alaroia-cfa, line 11: I removed the `VitaPartner.transaction do` from this method because it's simpler conceptually to me if the Rake task is the one managing transactions.

I don't think this change is necessary, but I like it. Happy to undo it if desired.

lib/tasks/upsert_vita_partners.rake
```diff
@@ -3,6 +3,8 @@
 namespace :db do
   desc 'loads new vita partners, updates existing partners'
   task upsert_vita_partners: [:environment, :insert_states] do
-    VitaPartnerImporter.upsert_vita_partners
+    VitaPartner.transaction do
```
alaroia-cfa, line 6: This is where I moved the `VitaPartner.transaction do`.

spec/models/intake_spec.rb
```diff
@@ -743,23 +743,10 @@
     let(:source) { nil }
     let(:intake) { create :intake, state_of_residence: state, source: source }
 
-    before(:all) do
```
alaroia-cfa, line 746: It's simpler to not bother with `before(:all)`, so I went for `before do` instead. This doesn't seem to cause significant slowdowns.

.circleci/config.yml
bytheway875: ! oh, so we were adding them to the database for CI test runs manually. tricky!

app/lib/vita_partner_importer.rb
bytheway875: If we move the logic to be part of the rake task, when this runs in tests or someone were to run it in rails c, it wouldn't roll back if it fails somewhere. are we okay with that tradeoff?

bytheway875 (approved):
Approved with a couple of questions.

app/lib/vita_partner_importer.rb
alaroia-cfa: Given that each test runs in a transaction per https://relishapp.com/rspec/rspec-rails/docs/transactions (we have `config.use_transactional_fixtures = true`), it will automatically roll back if it fails in a test.

If someone runs it in local development `rails c`, then they could have a problem. I'm not very sympathetic to that.

If someone runs it in prod `rails c`, then we could have a problem. Good eye. However, we have a safe way to run it (`rake db:upsert_vita_partners`), and that's how Aptible runs it (per our `.aptible.yml`). Given that, I'm OK with this change.

Let me know if you think a follow-up commit is important! I'll roll forward with this for now.

PR #576, opened 2020-11-09
Add tax return status to client index views

bytheway875:
- Adds status to client view
- To facilitate, had to make the view container wider by adding "wide" class to grid and creating new styles for it (confirmed with Anu that widening was copacetic)


sartan (approved):
Looks great!

Maybe later on we can make status a required field of a TaxReturn? Not a blocker for this PR.

Cheers!

PR #577, opened 2020-11-10
Update Mailgun Webhook Signing Key

bengolder:
This updates the mailgun webhook signing key to the current correct value.


alaroia-cfa:
@bengolder I'm happy to merge this now, and fix demo. Thanks for writing it, and I think I may have been the one who mistakenly rotated this key (not 100% sure, but I remember being in the Mailgun console last week, and don't honestly remember if I did/didn't).

PR #578, opened 2020-11-11
add canonical and hreflang link tags for SEO improvements

loumoore:


app/views/layouts/application.html.erb
```diff
@@ -11,16 +11,21 @@
...
-    <meta property="twitter:url" content="<%= url_for(only_path: false) %>">
+    <meta property="twitter:url" content="<%= canonical_url %>">
     <meta property="twitter:title" content="<%= content_for(:page_title) -%> | GetYourRefund.org">
     <meta property="twitter:description" content="<%=t("views.layouts.application.meta.description") %>">
     <meta property="twitter:image" content="<%= image_url("social_share_banner.png") %>">
 
+    <link rel="canonical" href="<%= canonical_url %>" />
+    <link rel="alternate" href="<%= canonical_url(:en) %>" hreflang="x-default" />
+    <link rel="alternate" href="<%= canonical_url(:en) %>" hreflang="en" />
+    <link rel="alternate" href="<%= canonical_url(:es) %>" hreflang="es" />
```
loumoore, line 27: the canonical URL should be for the current language
the x-default and en alternates should always be english 
the es alternate should be spanish

i elected not to mess with trying to use the mireembolso domain for spanish. i think we want to focus all of our SEO on the getyourrefund domain. with this change mireembolso.org will still be indexed separately so it will show up in google search results, but we won't be telling Google to _always_ consider mireembolso.org the canonical spanish domain, so getyourrefund.org should rank higher given it has a lot more traffic. not sure if this is ideal, but it's a fine starting point.

config/routes.rb
```diff
@@ -23,7 +23,6 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
   # All routes in this scope will be prefixed with /locale if an available locale is set. See default_url_options in
   # application_controller.rb and http://guides.rubyonrails.org/i18n.html for more info on this approach.
   scope '(:locale)', locale: /#{I18n.available_locales.join('|')}/ do
-    get '/:locale' => 'public_pages#home'
```
loumoore, line 26: so this seems like it's been a bug for a long time, probably since i did the first pass i18n routes. this resulted in e.g. url_for /en outputting "/en/en". you can see this in the open graph and twitter meta tags in production currently. removing this line fixes that and i don't think it was doing anything good that we'll miss.

app/views/layouts/application.html.erb
alaroia-cfa: TIL `hreflang` - cool!

spec/features/visit_home_page_spec.rb
```diff
@@ -10,6 +10,43 @@
     expect(page).to have_text "sign up here"
   end
 
+  context "for the www.getyourrefund.org domain" do
+    before do
+      Capybara.default_host = "https://www.getyourrefund.org"
```
alaroia-cfa, line 15: I'm worried this will leak into other tests. What if we omit this entirely, and stick to the test host?

> The main reason is that changing the default_host or app_host option will change this value globally for each spec that follows. The author of Capybara, Jonas Nicklas, states "Never change Capybara.default_host, just don't." and recommends that subdomains are visited explicitly."

https://minimul.com/capybara-and-subdomains.html

alaroia-cfa (changes requested):
Hi @loumoore ! This is a great PR. I'm nervous about the `.default_host = ...` line, so I'm not ready to approve yet, but I'm open to a discussion about that. Hopefully it's a quick fix to change the assertions.

If you won't have time to address that, I'd be OK if you want me to see if I can address that feedback rather than you.

spec/features/visit_home_page_spec.rb
loumoore: yikes good catch!

alaroia-cfa (approved):
Thanks! This is slightly simpler than before and is totally great.

PR #579, opened 2020-11-11
Invite users only to orgs you can manage

alaroia-cfa:
This does 3 things.

- Limit the orgs you can see in the dropdown to those you can edit (there are more orgs than the ones listed below, but the list now only shows the ones that this user can access)

Screen Shot 2020-11-11 at 1 06 10 PM

- Add backend validation that you can only change a user's org to ones you can `:manage` when editing a user.

- Add backend validation that you can only set a new user's org to ones you can `:manage` when inviting a user.

Feedback request specifically on the repeated use of `Ability.new(current_user)` -- is there a pattern that would avoid creating this object? Should I not worry about it? It seems ugly to me to imperatively create it, but I'm OK with it, but I wanted to ask @bytheway875 and/or @bengolder since you have more cancancan experience than Jenny & I do.

https://www.pivotaltracker.com/story/show/175618733

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -20,7 +20,7 @@ class Users::InvitationsController < Devise::InvitationsController
   def create
     super do |invited_user|
       # set default values
-      invited_user.update(role: invited_user.role || "agent" )
+      invited_user.update(role: invited_user.role || "agent")
     end
   end
 
@@ -28,7 +28,11 @@ class Users::InvitationsController < Devise::InvitationsController
 
   # Override superclass method for default params for newly created invites, allowing us to add attributes
   def invite_params
-    params.require(:user).permit(:name, :email, :vita_partner_id)
+    vita_partner_id = params.require(:user).require(:vita_partner_id)
+    vita_partner = VitaPartner.find(vita_partner_id)
+    if Ability.new(current_user).can?(:manage, vita_partner)
+      params.require(:user).permit(:name, :email, :vita_partner_id)
+    end
   end
 
   # Override superclass method for accepted invite params, allowing us to add attributes
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -15,7 +15,16 @@ class Users::InvitationsController < Devise::InvitationsController
   end
 
   authorize_resource :user, only: [:new, :create]
+  # This line does not do what i would like it to do :(
+  # was hoping to get a @vita_partners that matched the current user's organization (to be used in vita_partner_helper)
+  # load_and_authorize_resource :vita_partner, through: :user
   before_action :require_valid_invitation_token, only: [:edit, :update]
+
+  # maybe this is the way instead?
+  # def new
+  #   @vita_partners = current_user.accessible_organizations
+  #   super
+  # end
 
   def create
     super do |invited_user|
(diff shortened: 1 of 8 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -20,15 +20,19 @@ class Users::InvitationsController < Devise::InvitationsController
...
   private
 
   # Override superclass method for default params for newly created invites, allowing us to add attributes
   def invite_params
-    params.require(:user).permit(:name, :email, :vita_partner_id)
+    vita_partner_id = params.require(:user).require(:vita_partner_id)
+    vita_partner = VitaPartner.find(vita_partner_id)
+    if Ability.new(current_user).can?(:manage, vita_partner)
+      params.require(:user).permit(:name, :email, :vita_partner_id)
+    end
```
bengolder, line 35: I'm surprised to see queries and ability checks in the `_params` method. Ordinarily these are handled elsewhere.

In general we should never be using Ability.new directly unless we are creating separate Ability classes. `Ability.new( current_user)` is always available as `#current_ability` in any controller that inherits from `ApplicationController`

bengolder: This impliese that we would permit different params depending on a user's permissions

app/helpers/vita_partner_helper.rb
```diff
@@ -1,6 +1,6 @@
 module VitaPartnerHelper
   def grouped_organization_options
-    VitaPartner.top_level.collect do |partner|
+    VitaPartner.accessible_by(Ability.new(current_user)).top_level.collect do |partner|
       [partner.name, [[partner.name, partner.id], *partner.sub_organizations.collect { |v| [v.name, v.id] }]]
     end
   end
```
bengolder, line 6: In a controller, CanCanCan can preload the correct organizations that a user is allowed to access and make it available in the controller as `@organizations` or `@vita_partners`

app/controllers/users_controller.rb
```diff
@@ -25,11 +25,15 @@ def update
...
+    vita_partner_id = params.require(:user).require(:vita_partner_id)
+    vita_partner = VitaPartner.find(vita_partner_id)
+    if Ability.new(current_user).can?(:manage, vita_partner)
+      params.require(:user).permit(
+        :vita_partner_id,
+        *(:is_admin if current_user.is_admin?),
+        :timezone,
+        current_user.is_admin ? { supported_organization_ids: [] } : {}
+      )
+    end
```
bengolder, line 37: this params method and the `require`, `permit` pattern should only be used to determine which _keys_ we allow in the params. It should not be checking values or doing validation. That should happen elsewhere in the controller.

bengolder:
This doesn't match how I would expect us to use CanCanCan. Here are some examples that might help to show a different way that I think is more consistent with conventions for using CanCanCan to control permissions and access.
If you want any refreshers, feel free to review details in [the CanCanCan Spike writeup](https://hackmd.io/jjGatRUBReu-roKyuCnDMg?view#Super-flexible-and-facilitates-more-complex-permissions-logic) 

I would recommend reading more about `load_and_authorize_resource`, `load_resource`, and `authorize_resource` in CanCanCan because they demonstrate a good pattern for loading resources and checking permission to access them

```ruby
class Users::InvitationsController < Devise::InvitationsController
  def new
    # this would be the most expected way to restrict which orgs you can invite to. the drop down should use this query.
    @vita_partners = current_user.accessible_organizations
  end
end
```

alaroia-cfa:
Superceded by https://github.com/codeforamerica/vita-min/pull/581

PR #581, opened 2020-11-11
Limits visible vita partners to ones users can access

alaroia-cfa:
🔥  Requesting review from @bengolder especially! 🔥 

This now limits what is visible **and** what the backend accepts, resolving [Adjust invitations so users can only invite people to orgs they can manage](https://www.pivotaltracker.com/story/show/175618733)

This screenshot shows that the set of vita partners is smallish. What isn't evident in the screenshot is that there are other VITA partners now removed from the set that is listed!

image

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -13,12 +13,13 @@ class Users::InvitationsController < Devise::InvitationsController
     # If an anonymous user tries to send an invitation, send them to the invitation page after sign-in.
     require_sign_in(redirect_after_login: new_user_invitation_path)
   end
-  before_action :load_vita_partners, only: [:new]
+  before_action :load_vita_partners, only: [:new, :create]
 
   authorize_resource :user, only: [:new, :create]
   before_action :require_valid_invitation_token, only: [:edit, :update]
 
   def create
+    authorize!(:manage, @vita_partners.find(invite_params[:vita_partner_id]))
     super do |invited_user|
       # set default values
       invited_user.update(role: invited_user.role || "agent")
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -13,6 +13,7 @@ class Users::InvitationsController < Devise::InvitationsController
     # If an anonymous user tries to send an invitation, send them to the invitation page after sign-in.
     require_sign_in(redirect_after_login: new_user_invitation_path)
   end
+  before_action :load_vita_partners, only: [:new]
 
   authorize_resource :user, only: [:new, :create]
   before_action :require_valid_invitation_token, only: [:edit, :update]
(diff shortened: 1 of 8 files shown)


alaroia-cfa:
This **replaces** the old #579 PR! It doesn't do everything yet, but that's OK!

app/controllers/users/invitations_controller.rb
```diff
@@ -13,19 +13,24 @@ class Users::InvitationsController < Devise::InvitationsController
     # If an anonymous user tries to send an invitation, send them to the invitation page after sign-in.
     require_sign_in(redirect_after_login: new_user_invitation_path)
   end
+  before_action :load_vita_partners, only: [:new]
```
bengolder, line 16: you might be able to replace this and the private method with
```ruby
load_resource :vita_partner, collection: [:new], parent: false, only: [:new]
```

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -3,13 +3,22 @@
...
     it_behaves_like :a_get_action_for_authenticated_users_only, action: :new
+
+    let!(:inaccessible_vita_partner) { create(:vita_partner) }
+
+    it "sets @vita_partners so the template can render a list of partners the user has access to" do
+      sign_in user
+      get :new
+      expect(assigns(:vita_partners)).to include(vita_partner)
+      expect(assigns(:vita_partners)).not_to include(inaccessible_vita_partner)
+    end
```
bengolder, line 21: for the sake of a thorough test, it might be helpful to include a vita partner besides user.vita_partner, that the user does have access to. For example, a suborganization for the users org, or an org that the user is assisting with.

spec/features/case_management/invitations_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.feature "Sending and accepting invitations" do
   context "As an authenticated user" do
-    let(:user) { create :user_with_org, role: "agent" }
```
bengolder, line 5: I think users should have orgs by default, so we should maybe just remove this subfactory and give orgs to all user factories.

bengolder (approved):
Looks good! Left some comments for possible improvements, but nothing blocking. Good work!

app/controllers/users/invitations_controller.rb
embarnard: @bengolder we tried this method but got 403 errors

spec/controllers/users/invitations_controller_spec.rb
embarnard: Good idea! We pushed a new commit to this PR that adds that.

spec/features/case_management/invitations_spec.rb
embarnard: Yeah good idea! We should that in another PR

app/controllers/users/invitations_controller.rb
```diff
@@ -13,19 +13,25 @@ class Users::InvitationsController < Devise::InvitationsController
...
       # set default values
-      invited_user.update(role: invited_user.role || "agent" )
+      invited_user.update(role: invited_user.role || "agent")
     end
   end
 
   private
 
+  def load_vita_partners
+    @vita_partners = VitaPartner.accessible_by(Ability.new(current_user))
```
bengolder, line 32: `current_ability`

bengolder (approved):
Looks good! Nice work!
💖😜

PR #586, opened 2020-11-13
Remove 3 deprecated document types: W-2, 1099-MISC, 1099-K

bengolder:
This removes the 3 employment document types that were deprecated a while back in favor of the newer "Employment" document type: W-2, 1099-MISC, and 1099-K

- documents stored with old types will still have strings in the columns in the database, and will be readable.
- We won't be able to save or update old docs with deprecated types unless we skip validation or convert them to a newer type (Employment)


alaroia-cfa (approved):
This is good!

I think that if we'll be removing last-year data from the production database, then there's no harm in not being able to save/update old documents. I'm a little nervous about it, but I guess it is OK. Looking forward to getting that data out of there.

PR #588, opened 2020-11-13
Delete all Zendesk admin pages

bengolder:
This deletes the Zendesk OAuth, along with all the pages that rely on it, including pages for viewing zendesk tickets or related uploaded documents.

This should remove the Dependabot warning about omniauth, because it will remove omniauth from our dependencies.

Other notes:
- I moved the AnonymizedCsvExtractsController into the case management system, because I didn't want to assume that we were ready to remove it.
- I'll squash all these commits into one before merging.


app/controllers/case_management/anonymized_intake_csv_extracts_controller.rb
```diff
@@ -0,0 +1,18 @@
+module CaseManagement
+  class AnonymizedIntakeCsvExtractsController < ApplicationController
+    layout "admin"
+
+    load_and_authorize_resource
```
alaroia-cfa, line 5: It's cool that these lines:

```
    if user.is_admin
      can :manage, :all
    end
```

mean that `load_and_authorize_resource` works magically.

app/models/user.rb
```diff
@@ -54,8 +54,7 @@
 #  fk_rails_...  (vita_partner_id => vita_partners.id)
 #
 class User < ApplicationRecord
-  devise :database_authenticatable, :lockable, :validatable, :timeoutable, :trackable, :invitable, :recoverable,
-         :omniauthable, omniauth_providers: [:zendesk]
+  devise :database_authenticatable, :lockable, :validatable, :timeoutable, :trackable, :invitable, :recoverable
```
alaroia-cfa, line 57: Splendid

config/locales/en.yml
```diff
@@ -1524,30 +1530,4 @@ en:
...
-          chart:
-            bank_info: Bank Info
-            consent: Consent
-            doc_type: Doc Type
-            drop_off_document_bundle: Document Bundle
-            header: The following chart does not include documents uploaded by VITA volunteers or client support
-            link: Link
-            notes: Notes
-            upload_date: Upload Date
+        search_radius: Within %{distance} miles of %{zip} (%{zip_name})
```
alaroia-cfa, line 1533: Very minor, non-blocking, FYI: If the PR retained the end-of-file newline, then this line would have not been shown in the diff. I configure RubyMine to add end-of-file newlines to avoid diff noise like this.

alaroia-cfa (approved):
This is great work!

app/controllers/case_management/anonymized_intake_csv_extracts_controller.rb
bengolder: I totally agree, I was happy about that! Less work for me. And makes me feel confident that we can add pages for admins and not worry about the wrong people see it because we'd need to add it explicitly to their list of permissions

config/locales/en.yml
bengolder: Sorry about that. I'm happy to add that same configuration. Could you show me sometime?

PR #590, opened 2020-11-13
Stop scanning for docs to link to ZD

bengolder:
This deletes a cron task that scans every 30 minutes for new document uploads, and then adds comments to Zendesk tickets to note that they've been uploaded.


alaroia-cfa (approved):
This is outstanding. Bye bye this silly code. Great work @bengolder .

PR #591, opened 2020-11-13
Remove all ZD webhooks and related jobs

bengolder:
I've already deactivated the triggers in Zendesk that would send POST requests to our webhook endpoint.

This includes deleting the EmailController, which was a email-based webhook for Zendesk that we deprecated in Spring of 2020, but never removed.

Removes:
- SMS Zendesk ticket linking
- Zendesk ticket return status updates


alaroia-cfa (approved):
I approve. This is great!

PR #592, opened 2020-11-14
Remove zendesk jobs and services

bengolder:
This deletes many files. If it feels difficult to review and you'd prefer it to be broken up into smaller chunks, please let me know.

This PR removes all the different jobs that were creating and updating Zendesk tickets. It also removes the services that those jobs were relying on to talk to Zendesk over the Zendesk API.

It also removes some custom Zendesk CLI tools that partner support made for managing Zendesk users.

### No longer relying on Zendesk Ticket ID
We were using the existing of a Zendesk Ticket ID as an import marker of progress for an intake. Without a Zendesk ticket, many downstream transactions in the services would fail.
In many cases, I just removed any expectation of a Zendesk Ticket ID. But there are two cases where I had to change the logic:
1. During partner assignment, we were preventing users from going back and changing their partner assignment if they've already consented and we created a zendesk ticket. Instead, I changed the assignment logic to prevent accidental reassignment if the client has already reached the "In Progress" status (after consent) when they could start showing up in a client index page.
2. In some places, we show Zendesk Ticket ID as a confirmation number. I started using `intake.client_id` instead.

### Other notes
- I couldn't yet remove `EitcZendeskInstance` because the CSV extractors rely on having all the status keys. I kept only that portion of the class, and moved the file from `app/helpers` to `app/lib`
- We have not integrated EIP, DIY, or Drop offs into the Hub, so these don't currently go anywhere.

### Things we'll want to confirm before launch
These are some things that the Zendesk jobs and services were doing that we will likely want to confirm are working to our satisfaction in the newer version of the intake & hub.
- Making sure we have sufficient monitoring throughout the intake process, particularly any transactions with 3rd party services.
- If we want any further change to status beyond marking tax returns as "Open" at additional docs.
- If we want any other updates to the 13614-c, or adding any notes about intake progress to a client's hub page. Previously we noted when someone reached the additional info page. We are not longer doing anything on that page.
- We had some miscellaneous tags for specific service offerings or intake methods (`triaged from stimulus`, `saw at capacity page`, `already filed for latest year`, `211intake`), we'll probably want to be able to easily create similar markers. Perhaps we should set up a way to tag Clients?
- When statuses were updated in Zendesk, we were sending automatic notifications to clients, as well as sending followup messages if we didn't hear from them. We'll want to make sure we get similar message in place.


app/controllers/questions/personal_info_controller.rb
```diff
@@ -5,5 +5,10 @@ def illustration_path; end
     def tracking_data
       {}
     end
+
+    def after_update_success
+      is_in_progress = current_intake.client.tax_returns.where("status >= ?", TaxReturn.statuses["intake_in_progress"]).exists?
+      current_intake.assign_vita_partner! unless is_in_progress
+    end
```
bengolder, line 12: This is a replacement for logic that I removed from `Intake#assign_partner!`

app/forms/personal_info_form.rb
```diff
@@ -7,6 +7,5 @@ class PersonalInfoForm < QuestionsForm
   def save
     state = ZipCodes.details(zip_code)[:state]
     @intake.update(attributes_for(:intake).merge(state_of_residence: state))
-    @intake.assign_vita_partner!
```
bengolder, line 10: this line was moved into `PersonalInfoController#after_update_success`

app/views/case_management/clients/show.html.erb
```diff
@@ -9,8 +9,8 @@
     <div><%= t(".preferred_name") %>: <%= @client.preferred_name %></div>
     <br/>
 
-    <div><%= t(".intake_language") %>: <%= client.intake&.locale ? t("general.language_options.#{client.intake&.locale}") : "N/A" %></div>
-    <div><%= t(".phone_interview_language") %>: <%= @client.intake&.preferred_interview_language ? t("general.language_options.#{client.intake&.preferred_interview_language}") : "N/A" %></div>
+    <div><%= t(".intake_language") %>: <%= @client.intake&.locale ? t("general.language_options.#{@client.intake&.locale}") : "N/A" %></div>
+    <div><%= t(".phone_interview_language") %>: <%= @client.intake&.preferred_interview_language ? t("general.language_options.#{@client.intake&.preferred_interview_language}") : "N/A" %></div>
```
bengolder, line 13: Some of my changes to tests exposed a few broken references to `client` instead of `@client`

spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -27,6 +25,9 @@
     check "2019"
     click_on "Continue"
 
+    intake = Intake.last
+    expect(intake.client.tax_returns.pluck(:year).sort).to eq [2017, 2019]
```
bytheway875, line 29: IMHO this is best on a controller spec, not this feature spec. Not a blocker.

spec/models/intake_spec.rb
```diff
@@ -767,9 +717,8 @@
         context "source matches an organization" do
           let(:source) { src }
 
-          it "assigns to the correct group and the correct instance" do
+          it "assigns to the correct group" do
```
bytheway875, line 720: Thrilling to see `.zendesk_instance` not being written to.

spec/models/intake_spec.rb
```diff
@@ -779,9 +728,8 @@
             # latergram: this is kinda awesome
           end
 
-          it "assigns to the correct group and the correct instance" do
+          it "assigns to the correct group" do
```
bytheway875, line 731: We could skip writing `.vita_partner_group_id` and then delete this test too, I think.

bytheway875 (approved):
This is great! We (@alaroia-cfa and I) have some non-blocking ideas and suggestions.

Code deleted is code tested! Great work.

spec/factories/intakes.rb
```diff
@@ -260,7 +260,7 @@
       city { ZipCodes::ZIP_CODES[zip_code][:name].split(", ").first }
       state { ZipCodes::ZIP_CODES[zip_code][:name].split(", ").last }
       state_of_residence { state }
-      zendesk_instance_domain { EitcZendeskInstance::DOMAIN }
+      zendesk_instance_domain { "eitc" }
```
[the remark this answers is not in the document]
bengolder: We actually still need this for the tests that use it, specifically the tests for extracting CSVs of existing intake data. So I thought I'd just leave it with a hardcoded value for now. We can delete a lot more once we delete the data in our database.

app/controllers/questions/personal_info_controller.rb
bengolder: I tried using the scope, but because it's a scope that returns a subset of clients, it doesn't work on the singular client instance returned by `current_intake.client` So I changed it to 
```ruby
    def after_update_success
      current_intake.assign_vita_partner! unless Client.after_consent.where(intake: current_intake).exists?
    end
```

bengolder: I think there are good reasons to create Client & TaxReturn records before consent, which I wouldn't want to lose track of in a future PR. Happy to talk more in-depth 1:1 if that would be interesting.

spec/features/new_in_person_intake_spec.rb
```diff
@@ -1,11 +1,6 @@
 require "rails_helper"
 
 RSpec.feature "Add a new intake case from an in-person drop-off site" do
-  before do
```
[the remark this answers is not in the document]
bengolder: 😂

spec/features/web_intake/new_single_filer_spec.rb
bengolder: I was thinking of moving the feature specs in a direction that includes some extra checks, and this was a step in that direction. Do you think thats a good direction? If so, do you think this check should still be moved out of the feature check?

spec/models/intake_spec.rb
bengolder: I'm not sure how to remove these group IDs in this scenario, because we are using them as unique IDs for the VitaPatners, and I'd have to refactor the tests more thoroughly to remove those unique IDs.

For these lines, I'd have to figure out a different way to reliably uniquely identify the partner:
```ruby
      it_behaves_like :source_group_matching, "uwkc", "360009173713"
      it_behaves_like :source_group_matching, "uwvp", "360009267673"
      it_behaves_like :source_group_matching, "RefundDay-B", "360009704234"
      it_behaves_like :source_group_matching, "branchesfl", "360009704234"
      it_behaves_like :source_group_matching, "uwfm", "360009415834"
      it_behaves_like :source_group_matching, "RefundDay-C", "360009704354"
      it_behaves_like :source_group_matching, "catalyst", "360009704354"
```

bengolder: It doesn't look like we have a uniqueness constraint on `:name` or any another field, and imagine primary key IDs wouldn't be reliable across environments.

PR #593, opened 2020-11-16
Filter by status and stage

bytheway875:
Screen Shot 2020-11-16 at 2 51 08 PM
Screen Shot 2020-11-16 at 2 56 08 PM
Screen Shot 2020-11-16 at 2 56 25 PM

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -24,13 +24,11 @@ module ClientSortable
   end
 
   def filtered_and_sorted_clients(assigned_to: nil)
-    scopes = [
-      :after_consent,
-      [:delegated_order, @sort_column, @sort_order],
-      ([:assigned_to, assigned_to.id] if assigned_to.present?),
-      ([:where, { tax_returns: { status: TaxReturnStatus::STATUSES_BY_STAGE[@filters[:stage]] } }] if @filters[:stage]),
-      ([:where, { tax_returns: { status: @filters[:status] } }] if @filters[:status])
-    ].compact
-    scopes.inject(@clients) { |collection, args| collection.send(*args) }
+    clients = @clients.after_consent
+    clients = clients.delegated_order(@sort_column, @sort_order)
+    clients = clients.assigned_to(assigned_to.id) if assigned_to.present?
+    clients = clients.where(tax_returns: { status: TaxReturnStatus::STATUSES_BY_STAGE[@filters[:stage]] }) if @filters[:stage].present?
+    clients = clients.where(tax_returns: { status: @filters[:status] }) if @filters[:status].present?
+    clients
   end
 end
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -1,7 +1,10 @@
 module ClientSortable
+
   def setup_sortable_client
+    params[:status] = nil if params[:clear]
     @sort_column = clients_sort_column
     @sort_order = clients_sort_order
+    @filters = { status: clients_tax_return_status_filter, stage: clients_tax_return_stage_filter }
   end
 
   def clients_sort_order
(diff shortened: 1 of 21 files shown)


app/lib/tax_return_status.rb
```diff
@@ -0,0 +1,31 @@
+class TaxReturnStatus
```
bytheway875, line 1: @alaroia-cfa @sartan There are a couple of reasons that I resurrected the TaxReturnStatus class, albeit in a more simplified form.

First, I'm always trying to keep the single responsibility principle + Sandi Metz' "rules" in mind (https://thoughtbot.com/blog/sandi-metz-rules-for-developers) when writing Ruby classes, and one of the big ones is to keep classes less than 100 lines. The TaxReturn model was pushing 100 lines, and most of it had to do with methods dealing only with setting up statuses and stages. 

Second, because of the way I'm trying to calculate the value of a const outside of the context of a method using a method, the method needed to be defined BEFORE it was used in the class, and it was really strange to do that in TaxReturn class itself. 

Third, since we're using the stages in backend code to create scopes and facilitate filtering, it made sense to me to have the calculation of the stages outside of the helper which are more for presentational purposes. The determine_statuses_by_stage method is really similar to what was in the helper, but now it returns the raw stage name, which then gets translated in the helper.

app/models/client.rb
```diff
@@ -35,16 +35,17 @@ def self.delegated_intake_attributes
...
-  scope :assigned_to, ->(user) { joins(:tax_returns).where({ tax_returns: { assigned_user_id: user } }) }
+  scope :after_consent, -> { distinct.joins(:tax_returns).merge(TaxReturn.where("status > 100")) }
+  scope :assigned_to, ->(user) { joins(:tax_returns).where({ tax_returns: { assigned_user_id: user } }).distinct }
 
   scope :delegated_order, ->(column, direction) do
     raise ArgumentError, "column and direction are required" if !column || !direction
 
     if delegated_intake_attributes.include? column.to_sym
-      joins(:intake).merge(Intake.order(Hash[column, direction]))
+      select = delegated_intake_attributes.inject(["clients.*"]) { |arr, attr| arr << ("intakes." + attr.to_s) }
```
bytheway875, line 45: In order to use "distinct", we need to include the intake properties in delegated_intake_attributes in our select. We _could_ limit it to just the column we're passing in, but I was worried it might trigger a Hakiri violation because we'd be creating a query from user-provided data.

app/views/shared/_column_sort_link.erb
```diff
@@ -1,3 +1,3 @@
 <% new_order = @sort_column == column_name && @sort_order == "asc" ? "desc" : "asc" %>
 <% arrow = if @sort_column == column_name then @sort_order == "desc" ? "↑" : "↓" end %>
-<%= link_to("#{title} #{arrow}", column: column_name, order: new_order) %>
+<%= link_to "#{title} #{arrow}", { column: column_name, order: new_order, status: @filters && (@filters[:stage] || @filters[:status]) }, id: "sort-#{column_name}"  %>
```
bytheway875, line 3: Allows us to maintain the existing filters when we sort by a column.

app/views/shared/_client_filters.erb
```diff
@@ -0,0 +1,26 @@
...
+        <option selected value></option>
+        <% TaxReturn.grouped_statuses.each do |stage, statuses| %>
+          <option value="<%= stage %>" <%= params[:status] == stage && "selected" %>><%= TaxReturnStatusHelper.stage_translation(stage) %></option>
+          <% statuses.each do |status| %>
+            <option value="<%= status %>" <%= params[:status] == status && "selected" %>>&emsp;<%= status_translation(status) %></option>
+          <% end %>
+        <% end %>
+      </select>
+    </div>
+    <%= f.hidden_field :order, value: @sort_order %>
```
bytheway875, line 17: Passing in the sort order and column as hidden fields allows us to maintain the current order

app/controllers/concerns/client_sortable.rb
```diff
@@ -11,4 +14,23 @@ def clients_sort_order
...
+  def clients_tax_return_status_filter
+    TaxReturnStatus::STATUSES.keys.find { |status| status == params[:status]&.to_sym }
+  end
+
+  def clients_tax_return_stage_filter
+    TaxReturnStatus::STAGES.find { |stage| stage == params[:status] }
+  end
+
+  def filtered_and_sorted_clients(assigned_to: nil)
+    scopes = [
```
bytheway875, line 27: I'm hoping this creates a pattern we can use to add scopes to the query as we add new filters in future tickets.

app/lib/tax_return_status.rb
bytheway875: I'm also willing to move this back into TaxReturn -- I actually haven't moved the tests because I'm waiting for feedback before making a final decision.

sartan:
We're seeing the dropdown cleared after applying a filter.

image

alaroia-cfa (approved):
This does the job and is great! We left a comment with one bug. Happy to have you make minor changes; feel free to request review or just merge with minor changes if the tests still pass.

alaroia-cfa:
to do list if Alex & Asheesh grab this

- use explicit method call instead of `.send`
- fix `selected` issue & write feature test for it
- use `params[:status]` instead of `@filters` in `link_to`
- see if we can stop `render_views: false`
- use `TaxReturnStatus` constants directly instead of `TaxReturn` method, maybe requiring some tests to move
- Use `[] + something.map` for delegated order

sartan:
Needs further investigation, but possibly a bug relating to assignment/tax year.

tax_year_filter_bug

alaroia-cfa:
It's green, and I'm gonna merge it!

PR #594, opened 2020-11-17
Hreflang link fix

loumoore:


app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="<%= I18n.locale %>">
```
loumoore, line 2: noticed this issue in the same report that identified the hreflang issues. this mismatch can also cause problems.

sartan (approved):
@alaroia-cfa and I think this is splendid! We also read through https://developers.google.com/search/docs/advanced/crawling/localized-versions?hl=en&visit_id=637412254789206291-2835717631&rd=1 to learn more about this. hreflang is cool.

PR #595, opened 2020-11-17
Take action form with prefilled messages

embarnard:


app/controllers/case_management/tax_returns_controller.rb
```diff
@@ -23,17 +23,93 @@ def update
       redirect_to case_management_clients_path, notice: success_message
     end
 
-    def edit_status; end
+    def edit_status
+      # prefill status if we received a param
+      preselected_status = params[:tax_return][:status] if params[:tax_return] && params[:tax_return][:status]
```
alaroia-cfa, line 28: Consider `.dig` here:

```
      preselected_status = param.dig(:tax_return, :status)
```

Given that simplicity, consider inlining it below.

Not a blocker.

spec/controllers/case_management/tax_returns_controller_spec.rb
```diff
@@ -100,54 +104,243 @@
...
     context "as an authenticated user" do
       before { sign_in user }
 
-      it "creates a system note" do
-        expect(SystemNote).to receive(:create_status_change_note).with(user, tax_return)
-        post :update_status, params: params
+      it "returns an ok response" do
+        post :edit_status, params: params
+
+        expect(response).to be_ok
```
alaroia-cfa, line 122: Is this a POST -> GET redirect? I'm confused -- I expect a POST to either redirect to a GET (in which case, I'd expect `expect(response).to redirect_to(xyz)`), or I expect to see some purpose to the response, like `expect(assigns(:form)).to be_instance_of(...)` or something like that. Confused about this test.

spec/controllers/case_management/tax_returns_controller_spec.rb
```diff
@@ -100,54 +104,243 @@
...
+      context "when the message body is present" do
+        let(:message_body) { "There's money in the banana stand" }
+
+        context "and the contact method is email" do
+          let(:contact_method) { "email" }
+          let(:mock_mailer) { double }
+          let(:example_now_time) { DateTime.new(2020, 11, 20) }
+
+          before do
+            allow(mock_mailer).to receive(:deliver_later)
```
alaroia-cfa, line 278: You can use Rails's mailer tests rather than mocking the mailer if you want. See e.g. https://relishapp.com/rspec/rspec-rails/docs/mailer-specs

Not essential, just a FYI.

spec/features/case_management/tax_return_status_change_spec.rb
```diff
@@ -5,32 +5,36 @@
...
-    let!(:intake) { create :intake, client: client }
+    let!(:intake) { create :intake, client: client, locale: "en" }
     let!(:tax_return) { create :tax_return, year: 2019, client: client, status: "intake_in_progress" }
 
     before do
       login_as user
     end
 
-    scenario "logged in user can change a status on a tax return" do
+    xscenario "logged in user can change a status on a tax return" do
```
alaroia-cfa, line 15: Is this skipped because the PR is not ready yet? Just curious to understand.

alaroia-cfa (approved):
This is good work! I suppose the `xscenario` will become `scenario` soon?

I'm happy to approve here, and let you make further small changes. I left some very minor ideas for making the code more like the rest of the codebase just as a FYI.

app/controllers/case_management/tax_returns_controller.rb
bengolder: Thanks! For some reason I forgot about dig

spec/controllers/case_management/tax_returns_controller_spec.rb
bengolder: This was a typo that apparently still passed tests. Thanks for catching!

spec/controllers/case_management/tax_returns_controller_spec.rb
bengolder: In refactor soon, I wanted to out all the message sending logic in one place so that the controllers wouldn't have to check so many things.

bengolder:
This includes a refactor to increase consistency and reduce duplication when sending messages.

I added a `MessageSending` concern module that can be included in controllers in order to be able to send emails & text messages.

It is responsible for setting all the correct defaults on outgoing messages and attributing the message to the current user.

For now, I've kept thorough unit tests for sending messages on `OutgoingEmailsController`, `OutgoingTextMessagesController`, and on `TaxReturnsController#update_status`, but if everyone feels good about this code, it might be helpful to unit test all the sending on a new spec for the `MessageSending` module, and then remove the detailed expectations from these controller tests.

spec/controllers/case_management/tax_returns_controller_spec.rb
bengolder: I've now included that refactor, but for now I've left the unit tests as they are. If you feel good about it, I'd like to move the unit tests onto the shared controller concern, and then just check that the controllers call the right method to send out emails.

spec/features/case_management/tax_return_status_change_spec.rb
bengolder: I stopped skipping it.

app/controllers/case_management/outgoing_emails_controller.rb
```diff
@@ -1,30 +1,23 @@
...
+        send_email(@client, body: outgoing_email_params[:body], attachment: outgoing_email_params[:attachment])
       end
-      ClientChannel.broadcast_contact_record(@outgoing_email)
       redirect_to case_management_client_messages_path(client_id: @outgoing_email.client_id)
     end
 
     private
 
     def outgoing_email_params
-      # Use client locale someday
```
bengolder, line 21: We are now using client locale!

app/forms/case_management/take_action_form.rb
```diff
@@ -0,0 +1,37 @@
+module CaseManagement
+  class TakeActionForm < Form
+    attr_accessor :status, :locale, :message_body, :contact_method, :internal_note
+
+    def initialize(client, *args, **attributes)
+      @client = client
+      super(*args, **attributes)
+    end
```
bengolder, line 8: I'm not sure if there's a better way to do this in Ruby. I wanted to add a single positional argument without affecting the rest of initialization. This is borrowed from Python's `*args, **kwargs` convention and seems to work.

spec/controllers/case_management/outgoing_emails_controller_spec.rb
```diff
@@ -57,7 +57,7 @@
           { client_id: client.id, outgoing_email: { body: " " } }
         end
 
-        it "sends no email & redirects to client show" do
+        it "sends no email & redirects to client messages" do
           expect do
             post :create, params: params
           end.not_to change(OutgoingEmail, :count)
```
bengolder, line 63: just correcting the expectation to what it actually does

app/controllers/case_management/tax_returns_controller.rb
```diff
@@ -23,17 +24,84 @@ def update
...
+      default = "email"
+      prefers_sms_only = @client.intake.sms_notification_opt_in_yes? && @client.intake.email_notification_opt_in_no?
+      prefers_sms_only ? "text_message" : default
+    end
+
+    def status_macro(status)
+      case status
+      when "intake_more_info", "prep_more_info", "review_more_info"
+        document_list = @client.intake.relevant_document_types.map do |doc_type|
+          "  - " + doc_type.translated_label(@client.intake.locale)
```
alaroia-cfa, line 88: Change to form locale

app/controllers/case_management/tax_returns_controller.rb
```diff
@@ -23,17 +24,84 @@ def update
...
+        end.join("\n")
+        I18n.t(
+          "case_management.tax_returns.edit_status.status_macros.needs_more_information",
+          required_documents: document_list,
+          document_upload_link: @client.intake.requested_docs_token_link,
+          locale: @client.intake.locale
+        )
+      when "prep_ready_for_review"
+        I18n.t("case_management.tax_returns.edit_status.status_macros.ready_for_qr", locale: @client.intake.locale)
```
alaroia-cfa, line 97: Change to form locale

app/controllers/case_management/tax_returns_controller.rb
```diff
@@ -23,17 +24,84 @@ def update
...
+      when "filed_accepted"
+        I18n.t("case_management.tax_returns.edit_status.status_macros.accepted", locale: @client.intake.locale)
```
alaroia-cfa, line 99: Change to form locale

app/forms/case_management/take_action_form.rb
alaroia-cfa: Could be `.from_client` fwiw

app/forms/case_management/take_action_form.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def language_difference_help_text
```
alaroia-cfa, line 10: Could this be a helper? Who knows.

app/views/case_management/tax_returns/edit_status.html.erb
```diff
@@ -1,22 +1,53 @@
...
-             t("general.status"),
-             grouped_status_options_for_select,
-             include_blank: true,
-             selected: params["tax_return"]["status"]
-          ) %>
+    <%= f.cfa_select(
+      :status,
+      t("general.status"),
+      grouped_status_options_for_select,
+      include_blank: true,
```
alaroia-cfa, line 14: Maybe you shouldn't be able set the status to blank

app/views/case_management/tax_returns/edit_status.html.erb
```diff
@@ -1,22 +1,53 @@
...
+            options: {rows: 8}
+        )%>
+
+    <%= f.cfa_textarea(
+            :note_body,
+            t(".note_body_label"),
+            options: {rows: 8}
+        )%>
+
+
```
alaroia-cfa, line 46: so spacious

spec/controllers/case_management/tax_returns_controller_spec.rb
```diff
@@ -100,55 +100,260 @@
...
+          expect(tax_return.reload.status).to eq("intake_needs_assignment")
+          expect(flash[:notice]).to match "Updated status"
+        end
+      end
+
+      context "when the status is the same as the current status" do
+        let(:status) { "intake_in_progress" }
+
+        it "does not create a system status change note" do
+          expect(SystemNote).not_to receive(:create_status_change_note).with(user, tax_return)
```
alaroia-cfa, line 237: I'd rather see a test that the model's count is unchanged, but this is legit fine.

spec/controllers/case_management/tax_returns_controller_spec.rb
alaroia-cfa: +1 to move the tests onto the concern & validate here we do/don't call the method on the concern

spec/helpers/tax_return_status_helper_spec.rb
```diff
@@ -49,6 +49,23 @@
     end
   end
 
+  describe "#language_options" do
+    context "with a locale set" do
+      around do |example|
+        existing_locale = I18n.locale
+        I18n.locale = :es
+        example.run
+        I18n.locale = existing_locale
```
alaroia-cfa, line 58: So long as, if the test raises e.g. RuntimeError, this last line runs and resets the locale, this is great.

alaroia-cfa: I'd like to know if a `allow(I18n).to receive(:locale).and_return(:es)` would do the job in a `before`. If so, I prefer that. Not a blocker.

alaroia-cfa (approved):
No blockers! Some ideas. A minor bug about client locale; info in the comments. Whee!

app/controllers/case_management/tax_returns_controller.rb
bengolder: this turned out to be trickier than I expected. I'm going to defer until later

spec/controllers/concerns/message_sending_spec.rb
```diff
@@ -1,48 +1,117 @@
...
-    sign_in user
+    allow(DateTime).to receive(:now).and_return(expected_time)
+    allow(ClientChannel).to receive(:broadcast_contact_record)
+  end
+
+  describe "#send_email", active_job: true do
+    context "without an authenticated user" do
+      it "raises an error" do
+        expect do
+          subject.send_email(client, body: "hello")
```
bengolder, line 22: One day maybe we can rely on `@client`

bengolder (approved):
this PR is still the best

PR #596, opened 2020-11-18
Remove deprecated backfill, add PR suggestions

bengolder:
I wasn't able to use `after_consent` on a single client instance.
```ruby
  current_intake.client.after_consent.exists?
```
But I liked the idea of having a consistent check, so I instead used 
```ruby
Client.after_consent.where(intake: current_intake).exists?
```

`TicketClientDocumentLink` is currently preventing deploys on master


app/controllers/questions/personal_info_controller.rb
```diff
@@ -7,8 +7,7 @@ def tracking_data
     end
 
     def after_update_success
-      is_in_progress = current_intake.client.tax_returns.where("status >= ?", TaxReturn.statuses["intake_in_progress"]).exists?
-      current_intake.assign_vita_partner! unless is_in_progress
+      current_intake.assign_vita_partner! unless Client.after_consent.where(intake: current_intake).exists?
```
alaroia-cfa, line 10: 👍

alaroia-cfa (approved):
Looks grand!

PR #600, opened 2020-11-20
Replace 'CaseManagement' with 'Hub'

bengolder:
Process I used:
1. use RubyMine to do a rename refactor from `CaseManagement` module to `Hub`
2. search and replace `case_management` with `hub` everywhere in the repo
3. go through and manually rename any `case_management` folders that weren't renamed from above steps
4. search for case insensitive `case management`, `case_management`, `casemanagement` and `management` to find any other usages or comments and manually fix them.


config/routes.rb
```diff
@@ -145,7 +145,7 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
     get "/users/invitations" => "invitations#index", as: :invitations
```
bytheway875, line 145: Can we take this opportunity to namespace these user routes under hub too?

bengolder: We can probably move most if not all of them. It might be helpful to separate that step out, because for the devise routes it would take a bit more work.

bengolder: Sure, I'll start on that. I do think it would be good to merge this and do a separate PR unless you think there's something in the current commit that would cause a problem.

bytheway875 (approved):

alaroia-cfa (approved):
super rad (subject to shannon's earlier comments)

PR #602, opened 2020-11-20
Create memberships

bytheway875:
Creates memberships model to keep track of a users role in relation to an org.

- Accessible organization ids now looks for relationship through memberships
- There is a data migration that will run on the "up" when we release to demo
- We can now limit certain functionality to people with lead role for the client's organization

OUT OF SCOPE:
- For this PR, we decided to shim the existing functionality on the update user page. "supported organizations" are now those for which you have lead roles, and most users have one primary relationship to an org as a member through the drop down. I have some ideas on how we can update this page to add new organizations with specific roles if thats something the user has permission to do, but it seemed out of scope for this PR AND we wanted to show that we could model the data the same on the front end even though there was a fairly significant change on the backend.

### DIFF
--- a/app/controllers/hub/clients/organizations_controller.rb
+++ b/app/controllers/hub/clients/organizations_controller.rb
@@ -11,6 +11,7 @@ module Hub
       def edit;end
 
       def update
+        authorize!(:edit_organization, @client)
         @client.update(client_params)
         redirect_to hub_client_path(id: @client.id)
       end
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -21,8 +21,7 @@ class Users::InvitationsController < Devise::InvitationsController
   def create
     authorize!(:manage, @vita_partners.find(invite_params[:vita_partner_id]))
     super do |invited_user|
-      # set default values
-      invited_user.update(role: invited_user.role || "agent")
+      invited_user.memberships.create(vita_partner_id: invite_params[:vita_partner_id])
     end
   end
 
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -5,12 +5,20 @@ class Ability
     return unless user.present?
 
     accessible_organizations = user.accessible_organizations
+    alias_action :response_needed, :update_take_action, :edit_take_action, to: :update
+    alias_action :read, :create, :update, :destroy, to: :crud
 
     if user.is_admin?
       can :manage, :all
     else
       can :manage, [VitaPartner], id: accessible_organizations.pluck(:id)
-      can :manage, [Client, User], vita_partner: accessible_organizations
+      can :manage, User, memberships: { vita_partner: accessible_organizations }
+
+      can :crud, Client, vita_partner: accessible_organizations
+      can :edit_organization, Client do |client|
+        user.can_lead?(client.vita_partner)
+      end
+
       can :manage, [
         IncomingTextMessage,
         OutgoingTextMessage,
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -79,11 +79,23 @@ class User < ApplicationRecord
     }).exists?
   end
 
-  # returns organizations a user is a member of, and their children (if any).
+  # Returns a role specific to the vita_partner provided, so that we can scope behaviors and hide/show things on
+  # a specific client page (based on their organization) to users
+  # @param [VitaPartner || Int] vita_partner
+  def role_for(vita_partner)
+    memberships.find_by(vita_partner: vita_partner)&.specific_role
+  end
+
+  # returns organizations a user is a member of, and their children if any.
+  # Also includes coalition member organizations if a user has a lead role in a coalition.
   def accessible_organizations
     return VitaPartner.all if is_admin?
 
-    VitaPartner.where("id IN (?) OR parent_organization_id IN (?)", vita_partner_ids, vita_partner_ids)
+    ids = vita_partner_ids
+    # coalition_leads = memberships.joins(:vita_partner).where(role: "lead", "vita_partners.coalition_lead": true)
+    # coalition_leads.each { |org| ids << org.coalition.member_organization_ids }
+
+    VitaPartner.where("id IN (?) OR parent_organization_id IN (?)", ids, ids)
   end
 end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -56,21 +56,41 @@
 class User < ApplicationRecord
   devise :database_authenticatable, :lockable, :validatable, :timeoutable, :trackable, :invitable, :recoverable
 
-  belongs_to :vita_partner, optional: true
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
+
+  #
+  belongs_to :vita_partner, optional: true
   has_and_belongs_to_many :supported_organizations,
            join_table: "users_vita_partners",
            class_name: "VitaPartner"
+  #
+
+  has_many :memberships
+  accepts_nested_attributes_for :memberships
 
   attr_encrypted :access_token, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
 
   validates_presence_of :name
   validates_inclusion_of :timezone, in: ActiveSupport::TimeZone.country_zones("us").map { |tz| tz.tzinfo.name }
 
+  # returns true if the user has role "lead" for organization or its parents.
+  # @param keyword [Boolean] exclude_ancestors
+  def can_lead?(vita_partner, exclude_ancestors: false)
+    memberships.where({
+        role: "lead",
+        vita_partner_id: [
+          vita_partner.id,
+          *(vita_partner.ancestors unless exclude_ancestors)
+        ].compact
+    }).exists?
+  end
+
+  # returns organizations a user is a member of, and their children (if any).
   def accessible_organizations
-    accessible_organization_ids = [vita_partner_id] + supported_organizations.pluck(:id)
-    VitaPartner.where(id: accessible_organization_ids).or(
-      VitaPartner.where(parent_organization_id: accessible_organization_ids)
+    member_org_ids = memberships.pluck(:vita_partner_id)
+    VitaPartner.where(id: member_org_ids).or(
+      VitaPartner.where(parent_organization_id: member_org_ids)
     )
   end
 end
+
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -58,15 +58,9 @@ class User < ApplicationRecord
 
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
 
-  #
-  belongs_to :vita_partner, optional: true
-  has_and_belongs_to_many :supported_organizations,
-           join_table: "users_vita_partners",
-           class_name: "VitaPartner"
-  #
-
   has_many :memberships
-  accepts_nested_attributes_for :memberships
+  accepts_nested_attributes_for :memberships, allow_destroy: true
+  has_many :vita_partners, through: :memberships
 
   attr_encrypted :access_token, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
 
@@ -87,10 +81,9 @@ class User < ApplicationRecord
 
   # returns organizations a user is a member of, and their children (if any).
   def accessible_organizations
-    member_org_ids = memberships.pluck(:vita_partner_id)
-    VitaPartner.where(id: member_org_ids).or(
-      VitaPartner.where(parent_organization_id: member_org_ids)
-    )
+    return VitaPartner.all if is_admin?
+
+    VitaPartner.where("id IN (?) OR parent_organization_id IN (?)", vita_partner_ids, vita_partner_ids)
   end
 end
 
--- a/app/models/users_vita_partner.rb
+++ b/app/models/users_vita_partner.rb
@@ -1,0 +1,16 @@
+# == Schema Information
+#
+# Table name: users_vita_partners
+#
+#  user_id         :bigint           not null
+#  vita_partner_id :bigint           not null
+#
+# Indexes
+#
+#  index_users_vita_partners_on_user_id          (user_id)
+#  index_users_vita_partners_on_vita_partner_id  (vita_partner_id)
+#
+class UsersVitaPartner < ApplicationRecord
+  belongs_to :user
+  belongs_to :vita_partner
+end
--- a/app/models/vita_partner.rb
+++ b/app/models/vita_partner.rb
@@ -33,6 +33,7 @@ class VitaPartner < ApplicationRecord
   accepts_nested_attributes_for :memberships
 
   belongs_to :parent_organization, class_name: "VitaPartner", optional: true
+  belongs_to :coalition, optional: true
   has_many :sub_organizations, -> { order(:id) }, class_name: "VitaPartner", foreign_key: "parent_organization_id"
   validate :one_level_of_depth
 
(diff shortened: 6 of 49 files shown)


app/models/vita_partner.rb
```diff
@@ -27,15 +27,25 @@ class VitaPartner < ApplicationRecord
...
+  # attribute :user_vita_partner
 
   scope :top_level, -> { where(parent_organization: nil).order(:display_name).order(:name) }
 
   after_initialize :defaults
 
+  # organization + coalition
+  def ancestors
+    [parent_organization, parent_organization&.parent_organization].compact
+  end
```
bengolder, line 47: This looks like it's set up to handle possible "grandparent" relationships (parent of a parent). Is that something that that exists and we need to support?

app/models/user.rb
```diff
@@ -56,21 +56,41 @@
...
+  # @param keyword [Boolean] exclude_ancestors
+  def can_lead?(vita_partner, exclude_ancestors: false)
+    memberships.where({
+        role: "lead",
+        vita_partner_id: [
+          vita_partner.id,
+          *(vita_partner.ancestors unless exclude_ancestors)
+        ].compact
+    }).exists?
+  end
```
bengolder, line 86: Why couldn't this be
```ruby
def can_lead(vita_partner)
  memberships.where(vita_partner: vita_partner, role: "lead").exists?
end
```

I couldn't find a unit test for this method. But I also couldn't find a situation in the code where we would call
```
user.can_lead?(vita_partner)
```
and expect it to return true if the user were a member on vita_partner but a lead on the vita_partner's parent organization.

app/controllers/hub/clients/organizations_controller.rb
```diff
@@ -11,6 +11,7 @@ class OrganizationsController < ApplicationController
       def edit;end
 
       def update
+        authorize!(:edit_organization, @client)
```
bengolder, line 14: If we already have permissions set up in `Ability` for both `Client` & `VitaPartner, then I believe we can do this check without having to create a custom action.

```ruby
      def update
        @client.update(vita_partner: @vita_partners.find(client_params[:vita_partner_id]))
        redirect_to hub_client_path(id: @client.id)
      end
```

app/controllers/users/invitations_controller.rb
```diff
@@ -21,8 +21,7 @@ class Users::InvitationsController < Devise::InvitationsController
   def create
     authorize!(:manage, @vita_partners.find(invite_params[:vita_partner_id]))
     super do |invited_user|
-      # set default values
-      invited_user.update(role: invited_user.role || "agent")
+      invited_user.memberships.create(vita_partner_id: invite_params[:vita_partner_id])
```
bengolder, line 24: I would add a param for "role" to this, and allow certain users to be able to set a role for new users they invite.

app/controllers/users_controller.rb
```diff
@@ -4,34 +4,54 @@ class UsersController < ApplicationController
...
+    supported_org_ids = (params[:user][:supported_organization_ids] || []).reject(&:empty?)
+    memberships = supported_org_ids.map { |so| { user_id: @user.id, vita_partner_id: so, role: "lead"} }
+    memberships.push({user_id: @user.id, vita_partner_id: params[:user][:vita_partner_id], role: "member"}) if params[:user][:vita_partner_id].present?
+    memberships
+  end
+
+  def authorize_vita_partner_updates
+    changing_vita_partner_ids = [params[:user][:vita_partner_id], params[:user][:supported_organization_ids]].flatten.select(&:present?)
+    changing_vita_partner_ids.each { |id| authorize!(:manage, @vita_partners.find(id)) }
+  end
```
bengolder, line 49: If we are having to write custom logic for handling params and the object creation process for multiple models, I'd suggest using a `Form` abstraction. Perhaps a `UserUpdateForm` that you can add a corresponding unit test file for. That would allow us to separate out the permissions and authorization responsibilities of the controller from the param restructuring and model updating that the form can handle itself. `TakeActionForm` also handles a bunch of different models, so it might be a good comparison to start.

app/controllers/users_controller.rb
```diff
@@ -4,34 +4,54 @@ class UsersController < ApplicationController
...
+
+  rescue ActiveRecord::RecordInvalid
+    logger.info("User update unexpectedly failed. Rolling back transaction.")
+  end
+
+  def memberships
+    supported_org_ids = (params[:user][:supported_organization_ids] || []).reject(&:empty?)
+    memberships = supported_org_ids.map { |so| { user_id: @user.id, vita_partner_id: so, role: "lead"} }
+    memberships.push({user_id: @user.id, vita_partner_id: params[:user][:vita_partner_id], role: "member"}) if params[:user][:vita_partner_id].present?
+    memberships
```
bengolder, line 43: I don't think this `supported_organizations` approach to creating many to many relationships to explicitly manage each user's access to each organization is sustainable. We would need a symmetrical process in the controllers for creating child organizations. Furthermore, we do not yet have greeters (which supported organizations was originally designed for).
If we only have clear hierarchical relationships between partners and the orgs that users should be able to edit (I can manage any child of my org) then we should rely on the relationships between orgs to manage access to sub organizations and their clients.

I'd recommend deleting the portion of this form that handles supported organizations and to remove supported organizations as part of this refactor.

app/controllers/users_controller.rb
```diff
@@ -4,34 +4,54 @@ class UsersController < ApplicationController
...
+
+  def authorize_vita_partner_updates
+    changing_vita_partner_ids = [params[:user][:vita_partner_id], params[:user][:supported_organization_ids]].flatten.select(&:present?)
+    changing_vita_partner_ids.each { |id| authorize!(:manage, @vita_partners.find(id)) }
+  end
 
   def user_params
     params.require(:user).permit(
       *(:is_admin if current_user.is_admin?),
-      :vita_partner_id,
```
bengolder, line 32: If we are migrating the primary vita partner relationship into a membership model, I would suggest removing the vita_partner_id column.

app/lib/ability.rb
```diff
@@ -5,12 +5,20 @@ def initialize(user)
     return unless user.present?
 
     accessible_organizations = user.accessible_organizations
+    alias_action :response_needed, to: :update
+    alias_action :read, :create, :update, :destroy, to: :crud
```
bengolder, line 9: this looks like it is defining a custom action that is only used in one place. Why is this needed?

app/lib/ability.rb
```diff
@@ -5,12 +5,20 @@ def initialize(user)
...
       can :manage, :all
     else
       can :manage, [VitaPartner], id: accessible_organizations.pluck(:id)
-      can :manage, [Client, User], vita_partner: accessible_organizations
+      can :manage, User, memberships: { vita_partner: accessible_organizations }
+
+      can :crud, Client, vita_partner: accessible_organizations
+      can :edit_organization, Client do |client|
+        user.can_lead?(client.vita_partner)
+      end
```
bengolder, line 20: I believe we can remove this block and just rely on the existing VitaPartner and Client permissions.

app/controllers/users_controller.rb
```diff
@@ -4,34 +4,54 @@ class UsersController < ApplicationController
...
     redirect_to edit_user_path(id: @user), notice: I18n.t("general.changes_saved")
   end
 
   private
+  # To leave the view as-is but allow the ability to save as memberships, we're doing the simplest thing possible:
+  # blowing away and recreating memberships on every update.
+  def update_user
+    ApplicationRecord.transaction do
+      @user.memberships.destroy_all && @user.update!(user_params.merge(memberships_attributes: memberships))
+    end
```
bengolder, line 33: When doing something like destroying and updating related objects, it was really smart to wrap it in a transaction! Good thinking

app/controllers/users_controller.rb
```diff
@@ -4,34 +4,54 @@ class UsersController < ApplicationController
   before_action :require_sign_in
   load_and_authorize_resource
   load_and_authorize_resource :vita_partner, collection: [:edit, :update], parent: false
-
+  before_action :authorize_vita_partner_updates, only: :update
```
bengolder, line 7: `authorize_vita_partner_updates` is called twice: as a before action and inside of update

app/models/user.rb
```diff
@@ -56,21 +56,41 @@
...
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
+
+  #
+  belongs_to :vita_partner, optional: true
   has_and_belongs_to_many :supported_organizations,
            join_table: "users_vita_partners",
            class_name: "VitaPartner"
+  #
+
+  has_many :memberships
```
bengolder, line 68: I think it would be really helpful to have a many many relation defined that allows us to easily query vita_partners through the memberships table.

```ruby
has_many :vita_partners, through: :memberships
```
Should a user truly have multiple memberships though? 

If this replaces `User#vita_partner`, then how will we know which membership is the "primary" one, the org this user belongs to?

- If we are replacing vita_partner_id with a :member role, should we enforce that a user has one and only one "role: member" membership?
- Should we also enforce that the orgs they are "leads" for are not the same as their "primary" organization that they are member for?
  - Or can you be both a lead and member of the same org?
  - it seems implicit that the org you belong to is the one you are a "member" for. Should we separate that more explicitly from other memberships?

Should this have uniqueness constraints or can a user have two memberships with the same role for the same organization?

app/models/user.rb
```diff
@@ -56,21 +56,41 @@
...
 
-  belongs_to :vita_partner, optional: true
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
+
+  #
+  belongs_to :vita_partner, optional: true
   has_and_belongs_to_many :supported_organizations,
            join_table: "users_vita_partners",
            class_name: "VitaPartner"
+  #
```
bengolder, line 66: I'm concerned about having vita_partner, supported_organizations, _and_ memberships all at the same time. For a PR that introduces memberships, I'd like to see it completely replace the need for `vita_partner` & `supported_organizations` (and remove them from the codebase).
I would struggle with having all these implementations that solve the same issue side by side.

app/models/user.rb
```diff
@@ -56,21 +56,41 @@
...
+
+  has_many :memberships
+  accepts_nested_attributes_for :memberships
```
bengolder, line 69: nice!

app/models/user.rb
```diff
@@ -56,21 +56,41 @@
...
+  # returns organizations a user is a member of, and their children (if any).
   def accessible_organizations
-    accessible_organization_ids = [vita_partner_id] + supported_organizations.pluck(:id)
-    VitaPartner.where(id: accessible_organization_ids).or(
-      VitaPartner.where(parent_organization_id: accessible_organization_ids)
+    member_org_ids = memberships.pluck(:vita_partner_id)
+    VitaPartner.where(id: member_org_ids).or(
+      VitaPartner.where(parent_organization_id: member_org_ids)
     )
   end
```
bengolder, line 94: This method is complicated by keeping vita_partner and supported_organizations in the model. 

Additionally`pluck` is a method on a query that returns an array.
This would be more performant for any downstream actions if it returned an active record query of VitaPartner.

```ruby
# up above
has_many :vita_partners, through: :memberships
# without supported_organizations or vita_partner, and with
def accessible_organizations
  vita_partners.or(VitaPartner.where(parent_organization: vita_partners)
end
```

app/models/users_vita_partner.rb
```diff
@@ -0,0 +1,16 @@
...
+#
+# Indexes
+#
+#  index_users_vita_partners_on_user_id          (user_id)
+#  index_users_vita_partners_on_vita_partner_id  (vita_partner_id)
+#
+class UsersVitaPartner < ApplicationRecord
+  belongs_to :user
+  belongs_to :vita_partner
+end
```
bengolder, line 16: This looks like it's only used by `supported_organizations` so I would hope that this PR removes this model.

app/models/vita_partner.rb
```diff
@@ -27,15 +27,25 @@ class VitaPartner < ApplicationRecord
   has_many :intakes
   has_and_belongs_to_many :states, association_foreign_key: :state_abbreviation
   has_many :source_parameters
-  has_many :users
+
+  has_many :memberships
+  has_many :users, through: :memberships
+  accepts_nested_attributes_for :memberships
```
bengolder, line 33: This is great! Would love to see the symmetrical `has_many :vita_partners, through: :memberships`

app/models/vita_partner.rb
bengolder: Discussed. I think we should explicitly _not_ handle grandparents until we have consensus on the team that we need to support it.

bengolder (changes requested):
I think this is a great start for making memberships, especially given the starting context of dealing with supported_organizations

However, I don't think memberships are compatible side by side with supported_organizations and User#vita_partner.

I would approve this if the PR fully replaced those other two relationships that seek to model the same information.

app/models/user.rb
bytheway875: We're planning on removing that column, but I generally remove references in one PR, and then drop columns in followup PRs/tasks once we've ensured on production that the data migration worked and that everything is working okay, rather than dealing with reverting to backups etc.

app/controllers/hub/clients/organizations_controller.rb
bytheway875: We're using a custom action to create a helper in the view that displays the button or hides it based on your role permissions.

app/controllers/users/invitations_controller.rb
bytheway875: Allowing users to invite people to specific roles is a ticket in next week's sprint, so we were planning on deferring that update to that ticket.

app/controllers/users_controller.rb
bytheway875: This is the shim that we plan on removing and replacing with a simplified structure once we decide on what the updated UI should look like, so a refactor into a form object for "bridge" code doesn't seem like the write abstraction.

app/controllers/users_controller.rb
bytheway875: This is not sustainable, and this is not how we were planning on managing relationships in the long term. It's basically just maintaining the current implementation. The plan is to replace with the ability to add a new organization with new role and remove memberships from this page. Check out the description for why Alex and I decided that the UI refactor was out of scope for this PR! We can remove the supported organizations altogether from the UI if you'd prefer, but Asheesh had implemented something that used supported_organizations to create "leads" of multiple organizations and we wanted to maintain that behavior while adding explicit role.

app/controllers/users_controller.rb
bytheway875: See above -- the plan was to get this in and then do an additional migration to drop vita_partner_id and user_vita_partners table.

app/models/user.rb
bytheway875: There's no existing ticket or use case that requires a designation of "primary" vita partner, but should a story require that in the future we can add a is_primary field to the membership and ensure that only one membership can be marked as primary.

app/models/user.rb
bytheway875: We can remove supported_organizations and vita_partner as soon as we remove the shim -- these relationships are literally only here to support the ability to use honeycrisp styled fields (that require that the field responds to the method "errors" and be a method on the parent object). We should have done a better job commenting that these relationships are deprecated.

app/models/user.rb
bytheway875: One caveat: supported_organizations and vita_partner are not part of this query. We removed that and we're only getting the vita_partner_ids from the memberships. Supported organizations and vita_partner should no longer be referenced directly in the tests, abilities, or the majority of views we found (a more thorough search would happen when we remove those associations completely in a followup PR). The optimization is definitely a good idea -- Alex and I talked about adding a has_many :vita_partners, through: :memberships, but because of the parent/child structure of grabbing the organizations, we didn't want to confuse people with the meaning of "vita_partners" vs "accessible_organizations". However, I'm willing to add it because it is a ncie optimization.

bytheway875:
Hi Ben, 

Thanks for the thorough review! Wanted to clarify that there is no intention of continuing to use vita_partner and supported_organizations directly within the code -- they're part of the "Shim" strategy we talked about which is pretty much to create a bridge between the new behavior and the old behavior until we finalized a UI update for the user#edit page that maintains these relationships. I think I can definitely refactor to remove the relationships from the model in this PR, though, and there will be a followup task to drop the tables and columns after we have a successful prod data migration.

PR #603, opened 2020-11-20
Namespace user routes

bytheway875:
Namespaces user and devise routes into hub for case management related tasks.


bengolder (approved):
This is amazing and excellent work! I'm really happy to see this passing all the tests. Definitely approve! Thank you Shannon

One thing I'm curious about: I noticed the users/invitations_controller and users/sessions were put under a hub route but weren't moved into the hub folder with the other users-related controller. Was that because it broke something with the 3rd party libraries or was that for some other reason?

bytheway875:
Because of the way that rails resolves routes by default, I needed to move the user controller into hub module namespace in order to have them resolve correctly. This wasn't necessary to get the devise routes working, so I didn't even attempt it. I'll take a look to see if i can nest them in ~20 mins, but if it takes more that that will leave it as is.

bytheway875:
@bengolder namespacing the devise routes under hub instead of just using the pathname will change the resolved devise mapping from user to hub_user or something like that. I'm going to merge as is because i think would be a longer refactor than the added value we'd achieve by moving it right now.

PR #604, opened 2020-11-21
Generate fake clients with tax returns for a user on demo only

bengolder:
Usage so far:

```ruby
# get a user
user = User.last

BetaTestDataGenerator.make_clients_for_user(user, all_possible_assignees: User.all)
```
This will make a bunch of clients, based on the criteria set out in the [pivotal tracker story](https://www.pivotaltracker.com/story/show/175487284).

This code is only intended to be run manually and rarely on demo. I've run it locally and reviewed results. But didn't think it made sense to write specs for it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,10 @@ gem 'devise_invitable', '~> 2.0.0'
 gem 'cancancan'
 gem 'webpacker'
 
+group :demo, :development, :test do
+  gem 'factory_bot_rails' # added to demo for creating fake data
+end
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
@@ -65,7 +69,6 @@ group :development, :test do
   # Adds support for Capybara system testing and selenium driver
   gem 'capybara', '>= 2.15'
   gem 'webdrivers'
-  gem 'factory_bot_rails'
   gem 'rspec-rails'
   gem 'rails-controller-testing'
   gem 'pry-byebug'
--- a/app/lib/beta_test_data_generator.rb
+++ b/app/lib/beta_test_data_generator.rb
@@ -1,0 +1,98 @@
+# This is used by the beta_test_client factory to create fake clients on demo for beta testing
+
+class BetaTestDataGenerator
+  FIRST_NAMES = [
+    "Logan", "Justin", "Gabriel", "Jose", "Austin", "Kevin", "Elijah", "Caleb", "Robert", "Thomas", "Jordan",
+    "Cameron", "Jack", "Hunter", "Jackson", "Angel", "Isaiah", "Evan", "Isaac", "Luke", "Mason", "Jayden", "Jason", "Gavin",
+    "Aaron", "Connor", "Aiden", "Aidan", "Kyle", "Juan", "Charles", "Luis", "Adam", "Lucas", "Brian", "Eric", "Adrian",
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -58,14 +58,17 @@ gem 'devise_invitable', '~> 2.0.0'
 gem 'cancancan'
 gem 'webpacker'
 
+group :demo, :development, :test do
+  gem 'factory_bot_rails'
+end
```
bengolder, line 63: FactoryBot is really helpful for this, and we would need it to be available on demo

app/lib/beta_test_data_generator.rb
```diff
@@ -0,0 +1,96 @@
...
+    "Tangelo", "Tangerine", "Tomatillo", "Tomato", "Turnip", "Thyme", "Tea", "Tamarind",
+    "Ube", "Uva",
+    "Vidalia", "Verdolaga", "Vanilla",
+    "Water Chestnut", "Watercress", "Watermelon", "Waxed Bean", "Wasabi", "Walnut",
+    "Xocota", "Ximenia",
+    "Yam", "Yucca",
+    "Zucchini",
+  ]
+
+  STATUSES_AFTER_OPEN = TaxReturnStatus::STATUSES.select { |key, value| value >= 102 }.keys.freeze
```
alaroia-cfa, line 68: Could use `TaxReturnStatus::STATUSES[:intake_open]` instead of `102`. Not a blocker. Might improve readability.

app/lib/beta_test_data_generator.rb
```diff
@@ -0,0 +1,96 @@
...
+    first_name = FIRST_NAMES.sample
+    last_name = LAST_NAMES.select { |last| last[0] == first_name[0] }.sample
+    return [first_name, last_name]
+  end
+
+  def self.get_status_open_or_later
+
+  end
+
+  def self.make_clients_for_user(user, all_possible_assignees:)
```
alaroia-cfa, line 80: It looks like `all_possible_assignees` is a mandatory parameter, so consider removing the `:` in `all_possible_assignees:`. Not a blocker.

spec/factories/beta_test_clients.rb
```diff
@@ -0,0 +1,39 @@
...
+  factory :beta_test_client, class: "Client" do
+    vita_partner { VitaPartner.where(name: "[United Way California] Online Intake").first }
+    after(:create) do |client|
+      first_name, last_name = BetaTestDataGenerator.get_name
+      is_married = ["yes", "no"].sample
+      intake = create(
+        :intake,
+        client: client,
+        email_address: "gyr-test-client+#{client.id}@codeforamerica.org",
+        phone_number: "+15005550006",
```
alaroia-cfa, line 11: Might be nice to leave a comment about what this number is. Not a blocker.

app/lib/beta_test_data_generator.rb
```diff
@@ -0,0 +1,96 @@
...
+    clients = FactoryBot.create_list :beta_test_client, rand(3..12), vita_partner: user.vita_partner
+    clients.each do |client|
+      client.tax_returns.each do |tax_return|
+        tax_return.status = STATUSES_AFTER_OPEN.sample
+        die_roll = rand
```
alaroia-cfa, line 85: You could avoid assigning `rand` to a variable by doing a `case` instead.

```
tax_return.assigned_user = 
  case rand(10)
  when 0..4
    user
  when 5
    all_possible_assignees.order(Arel.sql("RANDOM()")).first
  when 6..9
    nil
  end
```

That's just me having fun with Ruby syntax. Not a blocker at all.

alaroia-cfa (approved):
Looks great! I left some comments for the fun of it.

PR #605, opened 2020-11-23
Show and edit timezone and interview timing preference

bytheway875:


app/views/hub/clients/show.html.erb
```diff
@@ -11,12 +11,15 @@
 
     <div><%= t(".intake_language") %>: <%= @client.intake&.locale ? t("general.language_options.#{@client.intake&.locale}") : "N/A" %></div>
     <div><%= t(".phone_interview_language") %>: <%= @client.intake&.preferred_interview_language ? t("general.language_options.#{@client.intake&.preferred_interview_language}") : "N/A" %></div>
+    <div><%= "Preferred Interview Time" %>: <%= @client.intake&.interview_timing_preference %></div>
+    <div><%= t("general.timezone") %>: <%= displayed_timezone(@client.intake&.timezone) || "N/A" %></div>
```
sniemeyer13, line 15: I18n as well?

app/views/hub/clients/edit.html.erb
```diff
@@ -12,6 +12,8 @@
       <%= f.cfa_input_field(:primary_first_name, t(".legal_first_name"), classes: ["form-width--long"]) %>
       <%= f.cfa_input_field(:primary_last_name, t(".legal_last_name"), classes: ["form-width--long"]) %>
       <%= f.cfa_input_field(:preferred_name, t(".preferred_name"), classes: ["form-width--long"]) %>
+      <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options) %>
+      <%= f.cfa_input_field(:interview_timing_preference, "Interview Timing Preference") %>
```
sniemeyer13, line 16: Can we I18n this?

sniemeyer13 (approved):
Looks good! Alex and I added two small comments about I18n.

bytheway875:
@sartan @sniemeyer13 Good catches!! I updated both parts ("N/A" seems to be N/A in both languages, but using a translation node is better) -- I actually am realizing idk if we have any translations for the timezones themselves, but I'm going to deal with that as part of another ticket.

PR #609, opened 2020-11-24
Ensure clients are linked to VITA partners

bengolder:
I noticed that during the intake process, when we assign intakes to VITA partners, we don't yet link the client to the vita partner. All the code in Ability and elsewhere expects clients to be directly associated to a vita partner. This ensures we do it  during intake. 

I didn't want to interfere with the existing `assign_vita_partner!` method on intake because I think that method is overloaded and in need of refactor. I also didn't want to force that method to rely on the existence of a client record. So instead I kept the logic in the controller.

### DIFF
--- a/app/controllers/questions/personal_info_controller.rb
+++ b/app/controllers/questions/personal_info_controller.rb
@@ -7,7 +7,10 @@ module Questions
     end
 
     def after_update_success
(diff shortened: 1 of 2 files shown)


alaroia-cfa (approved):
Great to have this work in the interim!

PR #610, opened 2020-11-24
Remove document_bundle from drop off model

alaroia-cfa:
While working on [Manually add client](https://www.pivotaltracker.com/story/show/175466221), @jenny-heath and I noticed that we don't need `document_bundle` on drop-offs anymore, since we'll be creating `Document` records. This removes the property and related validation & tests.

### DIFF
--- a/app/models/intake_site_drop_off.rb
+++ b/app/models/intake_site_drop_off.rb
@@ -108,7 +108,6 @@ class IntakeSiteDropOff < ApplicationRecord
   validates :phone_number, allow_blank: true, phone: true
   validate :has_valid_pickup_date?
 
-  has_one_attached :document_bundle
   belongs_to :prior_drop_off, class_name: self.name, optional: true
 
   def phone_number=(value)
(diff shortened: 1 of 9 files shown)


alaroia-cfa:
Note that I did check with Kelly M. to validate that we don't need to access this data in the future.

The story has some copy-pasta to delete these records from demo (and production). It struck me as not essential to do that in a data migration.

app/models/intake_site_drop_off.rb
```diff
@@ -106,18 +106,11 @@ class IntakeSiteDropOff < ApplicationRecord
   validates :certification_level, allow_blank: true, inclusion: { in: CERTIFICATION_LEVELS }
   validates :email, allow_blank: true, 'valid_email_2/email': true
   validates :phone_number, allow_blank: true, phone: true
-  validate :has_document_bundle?
   validate :has_valid_pickup_date?
 
   has_one_attached :document_bundle
```
bengolder, line 111: it looks like we should also remove this line in order to remove the property.

- Ben & Sarah

alaroia-cfa: D'oh, yes, I meant to do that. Doing so right now.

sniemeyer13: We've got it covered!

PR #611, opened 2020-11-25
Add "service type" (drop off/online intake) to tax return

bengolder:
commits:
- basic new & create controller actions


embarnard (approved):
Reviewed w/ @sartan -- looks good to us! 

Alex expressed concerns about the fact that currently there is no user flow that will allow different tax years to have a distinct service type; since all tax returns are created as a result of a single intake, they will all receive the same service type. ([tracker comment](https://www.pivotaltracker.com/story/show/175466221/comments/219679588) for context)

PR #621, opened 2020-12-02
Add rack middleware to validate params before hitting the controller

embarnard:


bytheway875 (commented):
This is super awesome work, @embarnard! I'm tagging @ddefigueiredo to see if he has any feedback on the specific html response codes we're using or any additional feedback 👯‍♂️

bytheway875:
@embarnard I noticed there is still an issue with the fake controller thing :( -- are the specs passing locally?

embarnard:
> @embarnard I noticed there is still an issue with the fake controller thing :( -- are the specs passing locally?

Yay thanks for tagging dimirtri! And yes they are passing locally :( unsure why they are failing here -- do you know where that example of the other fake controller in the app is?

ddefigueiredo:
The 400 status code is exactly right here. I also like how you cleanly put the check inside `string_contains_invalid_character?` making it easy to modify this later.

PR #624, opened 2020-12-03
Small changes to client create for acceptance

bytheway875:
1. Show selected signature method on client profile
2. Add some missing translation strings
3. Force users to explicitly select a language preference instead of defaulting to German
4. Force users to select at least one contact preference
5. Make email address a required field.

### DIFF
--- a/app/helpers/error_message_helper.rb
+++ b/app/helpers/error_message_helper.rb
@@ -1,0 +1,10 @@
+module ErrorMessageHelper
+  def error_message(object, error_attribute)
+    errors = object.errors.messages[error_attribute.to_sym]
+    return unless errors.present?
+
+    content_tag :div, class: "text--error" do
+      content_tag(:i, "", class: "icon-warning") + content_tag(:span, errors.join(", "))
+    end
+  end
+end
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -451,12 +451,7 @@ class Intake < ApplicationRecord
   end
 
   def filing_years
-    [
-      ("2019" if needs_help_2019_yes?),
-      ("2018" if needs_help_2018_yes?),
-      ("2017" if needs_help_2017_yes?),
-      ("2016" if needs_help_2016_yes?),
-    ].compact
+    TaxReturn.filing_years.map { |year| year if send("needs_help_#{year}_yes?") }.compact.map(&:to_s)
   end
 
   def most_recent_filing_year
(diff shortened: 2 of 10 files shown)


app/models/intake.rb
```diff
@@ -451,12 +451,7 @@ def requested_docs_token_link
   end
 
   def filing_years
-    [
-      ("2019" if needs_help_2019_yes?),
-      ("2018" if needs_help_2018_yes?),
-      ("2017" if needs_help_2017_yes?),
-      ("2016" if needs_help_2016_yes?),
-    ].compact
+    TaxReturn.filing_years.map { |year| year if send("needs_help_#{year}_yes?") }.compact.map(&:to_s)
```
sartan, line 454: I like how there is now an authoritative, centralized list of tax years. While the previous code was in my opinion a little more readable, I think the new version is more future-proof and maintainable.

app/views/hub/clients/new.html.erb
```diff
@@ -6,19 +6,14 @@
       <h1 class="form-card__title">
         <%= @title %>
       </h1>
-      <%= f.cfa_select(:service_type, "Service Type", [["Drop off", "drop_off"]], options: { selected: "drop_off" } ) %>
-      <%= f.cfa_select(:signature_method, "Signature Method", signature_methods_for_select, options: { selected: "online" }) %>
+      <%= f.cfa_select(:service_type, t("hub.clients.fields.service_type"), [["Drop off", "drop_off"]], options: { selected: "drop_off" } ) %>
```
sartan, line 9: Should the select options also be translated?

app/helpers/error_message_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ErrorMessageHelper
```
sartan, line 1: This is great!

config/locales/es.yml
```diff
@@ -32,6 +32,11 @@ es:
         primary_contact_info: Información de contacto
         spouse_contact_info: Información de contacto del cónyuge
         state_of_residence: Estado de residencia
+        signature_method: Método para recoger
+        service_type: Service type
```
sartan, line 36: Looks like we might need to translate this into Spanish.

sartan (approved):
Looks good!

app/views/hub/clients/_primary_info_fields.html.erb
```diff
@@ -6,15 +6,17 @@
     <%= f.cfa_input_field(:primary_last_name, t("hub.clients.fields.legal_last_name"), classes: ["form-width--long"]) %>
   </div>
   <%= f.cfa_input_field(:preferred_name, t("hub.clients.fields.preferred_name"), classes: ["form-width--long"]) %>
-  <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options) %>
+  <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options, selected: current_user.timezone) %>
```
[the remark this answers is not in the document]
bytheway875: yep!

PR #625, opened 2020-12-03
 Add search param to client filtering

bengolder:


bytheway875:
Merged after local review with @bytheway875 and @bengolder

PR #626, opened 2020-12-03
Remove IntakeSiteDropOff controller, model, views, routes and translations

sniemeyer13:


bengolder (approved):
Looks good! Thank you!
I'm noticing that this does not change the schema, so we will continue to see the table represented in `schema.rb`
I'm also noticing that this (appropriately) does not change any prior migrations.
Good work!

PR #630, opened 2020-12-07
Add state to the client list table [#175706866]

sartan:
image

### DIFF
--- a/app/views/hub/clients/index.html.erb
+++ b/app/views/hub/clients/index.html.erb
@@ -29,6 +29,9 @@
         <th scope="col" class="index-table__header">
           <%= render "shared/column_sort_link", title: t("general.updated_at"), column_name: "updated_at" %>
         </th>
+        <th scope="col" class="index-table__header">
+          <%= render "shared/column_sort_link", title: t("general.state"), column_name: "state" %>
+        </th>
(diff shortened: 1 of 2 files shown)


sniemeyer13:
In doing this story, we noticed that the "State" is tied to the intake, rather than the tax return. This could be slightly misleading because tax returns in different years could be tied to different States. (For example, if someone worked/lived in Texas in 2017 and then moved to Alabama in 2019, we currently only tie them to one of those States). 

In the UI for the client list page, we decided to tie the State column to the client, rather than to each tax return because this is the most accurate representation of the data. 

If we wanted to tie each tax year to a different State, we would have to change the intake flow to account for this information.

bytheway875:
@sartan @sniemeyer13 The other thing here is that there are two fields on the intake: state_of_residence and then "state" which I think is the state filled out as part of the _mailing_ address. So state_of_residence is probably more accurate to the tax return _but_ im not sure if we require that people fill out state of residence on the intake... I looked on prod and we only maintain state of residence info for ~20 of the last 100 intakes

sartan:
@bytheway875 Good call! We'll switch it from `state` to `state_of_residence`.

Also, we discussed having state tied to a given tax return year with Yvonne and she will create a separate story for that work.

embarnard (approved):
Great work! Nice to see the discussion with product in the story.

PR #631, opened 2020-12-07
Client edit form validations #175982462

bytheway875:


app/controllers/hub/clients_controller.rb
```diff
@@ -31,11 +31,11 @@ def create
     def show; end
 
     def edit
-      @form = ClientIntakeForm.from_intake(@client.intake)
+      @form = UpdateClientForm.from_client(@client.intake)
```
bytheway875, line 34: The name of the forms (ClientIntakeForm & CreateClientForm) caused a lot of confusion to Alex and I when working on it, so updating to be a lot more clear on what the purpose of each form was made some sense to us. Also, since they are now inheriting from the same "ClientForm" class, it also made more sense from a naming and consistency standpoint pass the client to the form instead of the intake.

app/forms/hub/client_form.rb
```diff
@@ -0,0 +1,46 @@
+module Hub
```
bytheway875, line 1: Contains all of the shared validations for the create and update client forms.

app/models/client.rb
```diff
@@ -29,6 +29,8 @@ class Client < ApplicationRecord
   has_many :notes
   has_many :system_notes
   has_many :tax_returns
+  accepts_nested_attributes_for :tax_returns
```
bytheway875, line 32: Facilitates updating the tax returns and intake through the client on the client forms, avoiding multiple query updates and transactions in the form objects.

app/controllers/hub/clients_controller.rb
bengolder: I love this rename!

app/models/client.rb
bengolder: oh nice! we can pass it to the parent model!

app/forms/hub/create_client_form.rb
```diff
@@ -69,17 +53,12 @@ def initialize(attributes = {})
...
-          intake.client.tax_returns.create(tax_return_defaults.merge(v)) if create_tax_return_for_year?(v[:year])
-        end
-      end
-      @client = @intake.client
+
+      @client = Client.create(
+        vita_partner_id: attributes_for(:intake)[:vita_partner_id],
+        intake_attributes: attributes_for(:intake),
+        tax_returns_attributes: @tax_returns_attributes.map { |_, v| create_tax_return_for_year?(v[:year]) ? tax_return_defaults.merge(v) : nil }.compact
+      )
```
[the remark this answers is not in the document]
bengolder: It would be good to include some unit tests to document this behavior (only creates models if _all_ inputs are valid) and to document the basic outputs of the `#save` method (false if not valid, nothing saved, returns truthy client if valid) to document & prevent regression.

spec/forms/hub/client_form_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+          end
+        end
+      end
+
+      describe "#preferred_interview_language" do
+        context "when nil" do
+          before do
+            form_attributes[:preferred_interview_language] = nil
+            form.valid?
+          end
```
bengolder, line 146: If this set up feels repetitive, a common way we will structure these is to pull the value in question into a `let`.

for example:

```ruby
describe "#preferred_interview_language" do
  before do
    form_attributes[:preferred_interview_language] = language
    form.valid?
  end
  
  context "when nil" do
    let(:language) { nil }
    
    it "adds an error to the field" do
      expect(form.errors[:preferred_interview_language]).to eq ["Can't be blank."]
    end
  end

  context "when blank" do
    let(:language) { "" }

    it "adds an error to the field" do
      expect(form.errors[:preferred_interview_language]).to eq ["Can't be blank."]
    end
  end
end
```

spec/forms/hub/update_client_form_spec.rb
```diff
@@ -0,0 +1,256 @@
+require "rails_helper"
+
+RSpec.describe Hub::UpdateClientForm do
```
bengolder, line 3: if everything in here is testing an instantiated form, you could do this at the top of the spec:
```ruby
subject do
  described_class.new(params)
end
```

and then just call `.save` or `.valid?` in the tests, instead of having to also instantiate

bengolder (approved):
This overall feels like a great improvement, so I approve!

There are some potential areas for further improving this:
The form unit specs could cover more of the basics (outputs of save method) a few more cases (such as the behavior the transaction block was previously implementing), and could be structured a little more like model specs than controller specs when testing validations. I would also hope to see more deletion or refactoring of the controller specs now that logic was moved into these forms instead of the controller. Perhaps on the controller specs we could stub out the form.

Some other small things:
- this made me realize we might have different understandings of the preferred_name field across the team. Might be good to reaffirm a shared understanding about what it's for at some point.
- I think we should expect a state of residence in the update. If we don't have it in existing data, that's a bug we should fix.

There's a ton of great stuff in here so I wouldn't block on any of my comments!

Great work!

app/forms/hub/create_client_form.rb
```diff
@@ -37,30 +37,14 @@ class CreateClientForm < Form
...
     set_attributes_for :tax_return, :service_type
-    before_validation :normalize_phone_numbers
-
     attr_accessor :tax_returns, :tax_returns_attributes, :client, :intake
-    validates :primary_first_name, presence: true, allow_blank: false
-    validates :primary_last_name, presence: true, allow_blank: false
+
+    # Additional validations inherited from parent ClientForm
     validates :vita_partner_id, presence: true, allow_blank: false
-    validates :phone_number, allow_blank: true, phone: true
```
sniemeyer13, line 46: This looks great!

app/forms/hub/client_form.rb
```diff
@@ -0,0 +1,46 @@
...
+        :primary_last_name,
+        :email_address,
+        :preferred_interview_language,
+        :sms_notification_opt_in,
+        :email_notification_opt_in
+
+    before_validation do
+      self.sms_phone_number = PhoneParser.normalize(sms_phone_number)
+      self.phone_number = PhoneParser.normalize(phone_number)
+      self.preferred_name = preferred_name.presence || "#{primary_first_name} #{primary_last_name}"
```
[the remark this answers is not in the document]
bytheway875: When you go through the client-side flow, the first questions we ask are for "name" and "zip code" -- even though we don't specify, I think the UI is intended to imply that you enter your full name there since you're starting an official form to get your taxes done.

Because we're using the preferred name field as the display name on the clients index _and_ its currently the only clickable field to the show page, it seems like we are expecting it to be a first and last name -- it may be the case that in fact we want to be using Legal name on the client index and we should update that.

Alex and I actually did the update to preferred "full" name because we felt like it was _less_ clear on the manual create client form that the preferred name should be the users entire preferred name (first and last, or something more than a nickname).


I'm going to share this conversation with product so that we can figure out how we want to treat names moving forward.

app/forms/hub/update_client_form.rb
```diff
@@ -29,58 +29,58 @@ class ClientIntakeForm < Form
...
-      self.phone_number = PhoneParser.normalize(phone_number)
-    end
-    validates :primary_first_name, presence: true, allow_blank: false
-    validates :primary_last_name, presence: true, allow_blank: false
+                       :state_of_residence
+    attr_accessor :dependents_attributes
+
+    # Historical intakes don't always have state_of_residence set, which indicates we may not ask about it
+    #   in traditional flow. Holding off on requiring during update until we explicitly always require it.
+    validates :state_of_residence, allow_blank: true, inclusion: { in: States.keys }
```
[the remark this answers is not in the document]
bytheway875: Ah, that makes sense. I took a look at the last 100 intakes on prod to see if they had that value filled and I'd seen that most of them did not, so we were erring on the side of caution -- making a field required on the admin side when it is not required during intake flow could cause issues where we _cant_ update certain fields on a client easily without additional info from the client. However, once I refined the query to only include users that reached consent stage, they all had them like you said.

PR #632, opened 2020-12-07
Stub pdf creation for intakes except when explicitly testing pdf creation [#175675390]

bytheway875:
clients_controller_spec before: 
Screen Shot 2020-12-07 at 1 11 26 PM

clients_controller_spec after:
Screen Shot 2020-12-07 at 1 11 49 PM

### DIFF
--- a/spec/models/intake_spec.rb
+++ b/spec/models/intake_spec.rb
@@ -1320,6 +1320,8 @@ describe Intake do
     before do
       example_pdf = Tempfile.new("example.pdf")
       example_pdf.write("example pdf contents")
+      allow(intake).to receive(:create_original_13614c_document).and_call_original
+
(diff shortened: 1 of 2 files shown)


embarnard (approved):
@bytheway875 reviewing with @alaroia-cfa, looks good to us! Great to see it speed up 50%!

We discussed that it might be annoying if lots of methods get stubbed out in rails helper over time, but we concluded we can cross that bridge later.

🚢

bytheway875:
@embarnard yes! I considered that it might be more confusing to stub this in the rails_helper and that it might make more sense just to stub it at the top of clients_controller, especially since the marginal gain from putting it in rails_helper wasnt that big ... but in this case, it really did feel like something that was worth stubbing in all cases because of what a hefty transaction the pdf creation is.

PR #633, opened 2020-12-07
Rename 'Needs response' to 'Needs attention' [#175608321]

sniemeyer13:


sniemeyer13:
Note that this refactor requires a database migration that is a rename of the column. Though renames are not backwards-compatible, we went ahead with it anyway because we do not have real data for this yet.

PR #634, opened 2020-12-08
Require upload on all Documents [#175366339]

alaroia-cfa:
Shannon and I implemented this as part of https://www.pivotaltracker.com/story/show/175366339 because in that story, we're going to be showing the document display_name in one new place in the hub, and we think now is a good time to make sure documents always have an appropriate `.display_name`.

This fixes a bug where all `Document` records had a `.display_name` of `Untitled`. The cause of the bug was that we generally add the `.upload` field _after the first time a `Document` is saved._ This meant that the `before_save` hook wasn't able to find the document, so it wrote `Untitled` into the `.display_name`.

This pull request makes `Document` require a `upload` -- i.e., all documents in the database expect to to have a file as part of them. This ensures we have a `upload` at the time the `before_save` hook is called. It also removes the `:with_upload` factory, migrating that behavior into the default Document factory.

(No `Co-authored-by` b/c Shannon and I are on different schedules today, so when I finished it, I was solo-ing.)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -42,6 +42,7 @@ class Document < ApplicationRecord
   belongs_to :documents_request, optional: true
   belongs_to :contact_record, polymorphic: true, optional: true
   has_one_attached :upload
+  validates :upload, presence: true
 
   before_save :set_display_name
 
@@ -56,10 +57,6 @@ class Document < ApplicationRecord
   def set_display_name
     return if display_name
 
-    if upload.present?
-      self.display_name = upload.attachment.filename
-    else
-      self.display_name = "Untitled"
-    end
+    self.display_name = upload.attachment.filename
   end
 end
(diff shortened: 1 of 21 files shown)


app/models/document.rb
```diff
@@ -56,10 +57,6 @@ def document_type_label
   def set_display_name
     return if display_name
 
-    if upload.present?
-      self.display_name = upload.attachment.filename
-    else
-      self.display_name = "Untitled"
-    end
+    self.display_name = upload.attachment.filename
```
bytheway875, line 60: woohoo!

bytheway875 (approved):
This is such an awesome, thorough change that not only fixes the bug but also enforces that we create documents in a consistent manner! _And_ removes outdated stuff from the test setup. 👏 👏 👏

PR #637, opened 2020-12-08
Allow client support role to see all clients [#175600422]

embarnard:


spec/lib/ability_spec.rb
```diff
@@ -143,4 +143,15 @@
       expect(subject.can?(:manage, SystemNote.new)).to eq true
     end
   end
+
+  context "as a client support user" do
+    let!(:user) { create :user, is_client_support: true }
+    let(:client_1) { create(:client, vita_partner: create(:vita_partner)) }
+    let(:client_2) { create(:client, vita_partner: create(:vita_partner)) }
+
+    it "can see all the clients from all the organizations" do
```
alaroia-cfa, line 152: I would like to see a negative test here validating that they can't `:manage` clients. It doesn't have to block this pull request.

I would also like to avoid the use of `:all` in the Ability.rb -- I think it could confuse us into creating excessively-broad permissions.

app/lib/ability.rb
```diff
@@ -8,6 +8,8 @@ def initialize(user)
 
     if user.is_admin?
       can :manage, :all
+    elsif user.is_client_support?
```
alaroia-cfa, line 11: If the `Ability` class gets much longer, consider separating it out into different Ability classes per role. Since a user can only have one role, this might make it easier to unit-test each role's abilities (one unit test per ability file) and then express more specific permissions.

Not a blocker for this pull request.

Here's a long discussion explaining how to do that: https://github.com/CanCanCommunity/cancancan/issues/352

Here's some sample code copy-pasted from there to make the idea hopefully clearer. First, modify ApplicationController so the `current_ability` method returns a different class based on the user's role. Copy-pasting from that discussion:

```
class ApplicationController < ActionController::Base
   def current_company
         ...
   end 
 private
    def current_ability
      @current_ability ||=
        if current_user
          Ability.new(current_user, current_company)
        else current_super_admin
          SuperAdminAbility.new
        end
    end
end
```

Then we'd implement different classes -- not `Ability` and `SuperAdminAbility` like in the copy-pasted code, but maybe `AdminAbility` and `ClientSupportAbility`, since those are two roles we'd have.

Not a blocker for this PR, but something to consider.

alaroia-cfa (approved):
Great work moving forward on the client support role!

As far as I can tell, this doesn't add an interface in the app for marking a user as client support. I think those will be essential as part of the full functionality of client support. This to me would be a blocker before declaring the story finished; otherwise, product would be unable to do acceptance testing. I would also like to know that this role is visible in the user list, to improve the workflow (also not a blocker).

I'd like to see us avoid `:all` in the Ability.rb. Not a blocker.

I know there'll be more work coming w/r/t roles and organizations. Some of that might change how some of this is structured. I think it's OK to have some amount of continuous reworking. I hope we get moving on those big stories soon.

None of those are blockers for this PR. Happy to see it merged!

spec/lib/ability_spec.rb
embarnard: @alaroia-cfa good points! I added a test to check that client support users could not manage clients and changed `:all` in `Ability.rb` to `[Clients]`

sniemeyer13:
@alaroia-cfa The user interface for marking someone as client support is interesting.. we have an upcoming story to define roles for invited users, which is the only way that all roles will be able to be defined - except for admin: right now we have a checkbox for "is admin". 

We could add another checkbox for "is client support", but we have no point in time to add checkboxes for the other roles. I guess what I'm trying to say is: should we create a story for checkboxes for each of the roles?

alaroia-cfa:
Personally, I believe that any time we add a new role, we need to add an interface to create a user in that role. Otherwise, acceptance testing isn’t possible as far as I know. Given that, I’d vote that it be part of this story (could be a follow-up pull request).

Changing which role a user has (e.g., our checkboxes) could be a follow-up story to me, since it can be acceptance-tested separately. Up to you if you’re working on this story and/or up to product.

embarnard:
@alaroia-cfa could we just pair with product on review and change the user to is_client_support: true in the console for testing?

alaroia-cfa:
@embarnard Yes :) This is intended to be advice and my personal opinion, not a blocker. To me, it's not really acceptance testing if we're in the loop, but if product is fine with that, that's definitely up to them rather than me!

sniemeyer13:
@alaroia-cfa Thank you! We ended up adding the check box. :)

PR #643, opened 2020-12-10
Redirect to Rails storage URLs [#176083281]

alaroia-cfa:
This pull request relies on Rails/ActiveStorage's `#rails_blob_url` method, which allows us to redirect to signed URLs in Amazon S3 in dev, staging, and production. This addresses the cross-site scripting issue that arises from our previous use of `#download`.

It configures a 20 second timeout for those URLs per conversation with Dimitri; see e.g. https://www.pivotaltracker.com/story/show/176083281

It also removes a now-unused "helper" that wasn't a template helper anyway. 👊

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -1,6 +1,5 @@
 class DocumentsController < ApplicationController
   include AccessControllable
-  include FileResponseControllerHelper
   before_action :require_intake
 
   def destroy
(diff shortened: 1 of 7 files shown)


alaroia-cfa:
Not that anyone really needs to know, but FYI, I first wrote a version of this which was about 50 painstaking lines of code. I decreased it to 30. Then I figured out how to really lean on the framework and make it net negative in code!!! :)

alaroia-cfa:
Closing this -- I realize there's one way it misses the mark. I'll resubmit.

PR #644, opened 2020-12-10
Redirect to transient storage URLs [#176083281]

alaroia-cfa:
This pull request relies on Rails/ActiveStorage's `#service_url` method, which allows us to redirect to signed URLs in Amazon S3 in dev, staging, and production. This addresses the cross-site scripting issue that arises from our previous use of `#download`.

It configures a 20 second timeout for those URLs per conversation with Dimitri; see e.g. https://www.pivotaltracker.com/story/show/176083281

It also removes a now-unused "helper" that wasn't a template helper anyway. 👊

### DIFF
--- a/app/controllers/hub/documents_controller.rb
+++ b/app/controllers/hub/documents_controller.rb
@@ -1,7 +1,7 @@
 module Hub
   class DocumentsController < ApplicationController
     include AccessControllable
-    include FileResponseControllerHelper
+    include FilesConcern
 
     before_action :require_sign_in
     load_and_authorize_resource :client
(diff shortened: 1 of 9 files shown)


app/controllers/hub/documents_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
     end
 
     def show
-      render_active_storage_attachment @document.upload
+      redirect_to transient_storage_url(@document.upload.blob, disposition: :inline)
```
bytheway875, line 20: If inline is the fallback default, you could get away with passing it in here!

bytheway875 (approved):
LGTM! 👯

app/controllers/hub/documents_controller.rb
alaroia-cfa: Good call! Removed.

alaroia-cfa:
Thanks @bytheway875! I'll rebase & squash-merge this once my new commit is green.

PR #645, opened 2020-12-10
SQL script to delete 2020 data [#176061608]

sartan:


alaroia-cfa (approved):
Discussed with @bytheway875 and @sartan and I approve!

PR #646, opened 2020-12-10
Remove idme references

bytheway875:
- Drops table 
- Removes spouse id.me code
- Removes idme logo
- Removes idme model


alaroia-cfa (approved):
Code deleted is code tested. Great work!

PR #649, opened 2020-12-11
WIP non-urgent refactor: Add ClientFilterForm [#175207595]

alaroia-cfa:
This adds the start of ClientFilterForm, a `Form` that can centralize our client searching, filtering, and ordering logic.

The reason I began writing it was to work toward being able to TDD adding a new column to order by. I found that tests for the client filtering & sorting & searching logic was spread between the ClientsController and the ClientSortable, so I wanted to unify them somewhere. I found it easier to do this in a new Form than by modifying ClientSortable, based on advice from @bengolder that a form might be a good idea, and based on experience trying to add more tests to ClientSortable with @sniemeyer13 .

This is work in progress toward https://www.pivotaltracker.com/story/show/175207595

### DIFF
--- a/app/forms/hub/client_filter_form.rb
+++ b/app/forms/hub/client_filter_form.rb
@@ -1,0 +1,16 @@
+module Hub
+  class ClientFilterForm < Form
+    attr_accessor :sort_column, :sort_order, :_filtered_and_sorted_clients
+
+    def filtered_and_sorted_clients(clients)
+      clients = clients.after_consent
+      clients.delegated_order(sort_column, sort_order)
+    end
+
+    def initialize(*args, **attributes)
+      super(*args, **attributes)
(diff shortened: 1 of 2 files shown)


app/forms/hub/client_filter_form.rb
```diff
@@ -0,0 +1,16 @@
+module Hub
+  class ClientFilterForm < Form
+    attr_accessor :sort_column, :sort_order, :_filtered_and_sorted_clients
```
bengolder, line 3: I love how this very clearly shows which methods are exposed as a public API!

app/forms/hub/client_filter_form.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    def filtered_and_sorted_clients(clients)
+      clients = clients.after_consent
```
bengolder, line 6: I wonder if this should be moved into the Controllers, because it feels like a high level scope rather then a filter based on inputs

PR #650, opened 2020-12-11
seo site audit fixes:

loumoore:
- change approach to hreflang
- change locale switcher to not use query params
- redirect vanity urls and / to correct home url with locale

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,12 +13,6 @@ class ApplicationController < ActionController::Base
 
   def canonical_url(locale=I18n.locale)
     # Leave the locale out of canonical URLs in the default locale (works ok either way but needs to be consistent)
-    url_for(only_path: false, locale: locale == I18n.default_locale ? nil : locale)
-  end
-
-  def hreflang_url(locale=nil)
-    # The default hreflang URL shouldn't include locale but all other hreflang links should
-    # So unlike canonical_url we will not replace the default locale with nil
     url_for(only_path: false, locale: locale)
   end
 
@@ -164,8 +158,7 @@ class ApplicationController < ActionController::Base
   end
 
   def switch_locale(&action)
-    locale = available_locale(params[:new_locale]) ||
-      available_locale(params[:locale]) ||
+    locale = available_locale(params[:locale]) ||
       available_locale_from_domain ||
       http_accept_language.compatible_language_from(I18n.available_locales) ||
       I18n.default_locale
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -4,6 +4,10 @@ class PublicPagesController < ApplicationController
     true
   end
 
+  def redirect_locale_home
+    redirect_to root_path, { locale: I18n.locale }
+  end
+
   def home; end
 
   def diy_home; end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,7 +16,7 @@ module ApplicationHelper
 
   def link_to_locale(locale, label, additional_attributes = {})
     link_to(label,
-            "?" + request.query_parameters.merge("new_locale" => locale).to_query,
+            { :locale=>locale, :params=>request.query_parameters },
             lang: locale,
             "data-track-click": "intake-language-changed",
             "data-track-attributes-to_locale": locale,
--- a/app/views/public_pages/faq.html.erb
+++ b/app/views/public_pages/faq.html.erb
@@ -37,7 +37,7 @@
             <h3><%=t("views.public_pages.faq.getting_started.missing_docs.q") %></h3>
           </a>
           <div class="accordion__content" id="a3">
-            <p><%=t("views.public_pages.faq.getting_started.missing_docs.a_html", :transcript_link => link_to("https://www.irs.gov/individuals/get-transcript", target: "_blank", rel: "noopener")) %></p>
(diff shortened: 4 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -13,12 +13,6 @@ def self.default_url_options
 
   def canonical_url(locale=I18n.locale)
     # Leave the locale out of canonical URLs in the default locale (works ok either way but needs to be consistent)
-    url_for(only_path: false, locale: locale == I18n.default_locale ? nil : locale)
-  end
-
-  def hreflang_url(locale=nil)
-    # The default hreflang URL shouldn't include locale but all other hreflang links should
-    # So unlike canonical_url we will not replace the default locale with nil
```
loumoore, line 21: going back to a simpler approach where both canonical and hreflang URLs will always have the locale included so no need to have separate methods

app/controllers/application_controller.rb
```diff
@@ -164,8 +158,7 @@ def set_sentry_context
   end
 
   def switch_locale(&action)
-    locale = available_locale(params[:new_locale]) ||
-      available_locale(params[:locale]) ||
+    locale = available_locale(params[:locale]) ||
```
loumoore, line 161: replacing the new_locale param approach for switching locale with just setting the locale param in the path so this is no longer needed

app/controllers/public_pages_controller.rb
```diff
@@ -4,6 +4,10 @@ def include_analytics?
     true
   end
 
+  def redirect_locale_home
+    redirect_to root_path, { locale: I18n.locale }
+  end
+
```
loumoore, line 10: this will redirect to the homepage after locale has been set based on the logic in application_controller which importantly will inspect the HTTP_ACCEPT_LANGUAGE header and check for the mireembolso domain

app/helpers/application_helper.rb
```diff
@@ -16,7 +16,7 @@ def round_meters_up_to_5_mi(meters)
 
   def link_to_locale(locale, label, additional_attributes = {})
     link_to(label,
-            "?" + request.query_parameters.merge("new_locale" => locale).to_query,
+            { :locale=>locale, :params=>request.query_parameters },
```
loumoore, line 19: sets the locale as part of the path instead of as a separate query param, retaining other query params

app/views/public_pages/faq.html.erb
```diff
@@ -37,7 +37,7 @@
             <h3><%=t("views.public_pages.faq.getting_started.missing_docs.q") %></h3>
           </a>
           <div class="accordion__content" id="a3">
-            <p><%=t("views.public_pages.faq.getting_started.missing_docs.a_html", :transcript_link => link_to("https://www.irs.gov/individuals/get-transcript", target: "_blank", rel: "noopener")) %></p>
+            <p><%=t("views.public_pages.faq.getting_started.missing_docs.a_html", :transcript_link => link_to("https://www.irs.gov/individuals/get-transcript", "https://www.irs.gov/individuals/get-transcript", target: "_blank", rel: "noopener", "data-track-click": "faq-irs-transcript", "data-track-attribute-position": "faq3")) %></p>
```
loumoore, line 40: this was a bug, was just linking to itself with the IRS url as text. since it had funky params it showed up in the site audit

loumoore: and it didn't have link tracking so i added it

config/routes.rb
```diff
@@ -20,6 +20,12 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
   end
 
   mount Cfa::Styleguide::Engine => "/cfa"
+
+  # In order to disambiguate versions of english pages with and without locales, we redirect to URLs including the locale
+  # If we redirect here in the route declarations, we can't inspect accept headers to determine the proper default locale,
+  # hence the redirect actions in public_pages.
+  root "public_pages#redirect_locale_home", as: :redirected_root
```
loumoore, line 27: putting this before the locale scope allows us to keep the locale as optional within the locale scope, which is a much less intrusive change than having to redirect all urls without locales to the version with locales.

config/routes.rb
```diff
@@ -145,7 +151,7 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
     ### END Hub Admin routes (Case Management)
 
     # Any other top level slash just goes to home as a source parameter
-    get "/:source" => "public_pages#home", constraints: { source: /[0-9a-zA-Z_-]{1,100}/ }
+    get "/:source" => "public_pages#redirect_locale_home", constraints: { source: /[0-9a-zA-Z_-]{1,100}/ }
```
loumoore, line 154: this accomplishes 2 things -- it ensures we are consistent in including locales, and the redirect also fixes duplicate content issues that this approach causes e.g. /211 and / and /en and /whatevs all have the same content and really should be treated as a single canonical page for search. and now they are, because they all redirect to /en

loumoore:
Since my first couple attempts failed to get the canonical/hreflang tag dance right, I figured out how I could run my SEO audit tool locally (I just used ngrok), and actually verified these fixes not only do what I was trying to do but actually pass the audit this time. We had a audit score of 26 out of 100 before 😬 and with this change it's 💯

loumoore:
I also did some more research and verified that there's no penalty for redirecting your homepage like we will now be doing from / to /en or /es. Most of the issues with my previous attempts were related to trying to preserve the locale-free URLs in English in the canonical and hreflang tags and it turns out it just doesn't matter.

bytheway875:
LGTM! 👍

PR #653, opened 2020-12-11
Remove state-based routing

alaroia-cfa:
This makes it easier to rename the VitaPartners model to VitaGroup.

It also paves the way toward dropping reliance on the vita_partners.yml file. We're going to have to stop relying on that file when admins edit organizations in the interface.

Note the reliance on `accepts_overflow: true` to ensure routing continues to operate.

On demo, it would be sensible to mark **California Online Intake Group** as `accepts_overflow: true`.

Please don't merge without asking me; I am hoping to sequence this with other work. Review is welcome.


alaroia-cfa:
Ben and I figured out how to not need this, so I'm closing it.

PR #655, opened 2020-12-14
Make the hub full width by default [#176065075]

bytheway875:
Screen Shot 2020-12-14 at 10 39 58 AM
Screen Shot 2020-12-14 at 10 39 47 AM

### DIFF
--- a/app/assets/stylesheets/molecules/_breadcrumb.scss
+++ b/app/assets/stylesheets/molecules/_breadcrumb.scss
@@ -1,8 +1,8 @@
 .breadcrumb {
-  background-color: #F7F5F4;
+  background-color: $color-grey-light;
   height: 3.5rem;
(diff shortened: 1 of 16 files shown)


bytheway875:
I opted to go with making the container full width and then adding a wrapper class to put around each section instead of adding a default margin and then using negative margin for the tables... I think this makes things more explicit, composable, and extensible, and makes new styling less tricky to implement later on.

app/assets/stylesheets/molecules/_breadcrumb.scss
```diff
@@ -1,8 +1,8 @@
 .breadcrumb {
-  background-color: #F7F5F4;
+  background-color: $color-grey-light;
```
embarnard, line 2: nice! I like the addition of SASS variables

PR #662, opened 2020-12-15
Update twilio credentials in all environments

jenny-heath:
and update texting code to reflect new subaccounts and messaging services


bytheway875:
I was a little confused in the code to see client.messages because I  thought we were creating a message object for one of our internal clients. I guess twilio_client as a variable name would make that less confusing, but in the context of the entire file the confusion isn't as apparent!

bytheway875 (approved):

jenny-heath:
@bytheway875 oh i didn't even think about changing that! makes a lot more sense to call it twilio_client

PR #664, opened 2020-12-15
Show and hide bank account information for client [#175406521]

bytheway875:
I made a good-faith attempt to get this working using server-side only code because of @bengolder's note on the ticket about accessibility concerns with a js implementation. Here is the draft PR of that approach for review: https://github.com/codeforamerica/vita-min/pull/665

However, I ran into a bunch of potential security issues with the server-side approach due to the fact that users could use the browser forward/back buttons and see the bank account information again. I wrote some JS to try to mitigate that issue, but the bank account info still flashed on the page AND it muddied the browser history a lot.

In an effort to make this section accessible, I did some research and it seemed like adding role="alert" to the section would make it start reading that section as soon as its added to the page, which seems to be what we want since it will have been added to the page as a result of clicking the "View" link. I also added text that will read "Hidden" instead of displaying the dots when on a screen reader.


config/routes.rb
```diff
@@ -120,6 +120,8 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
       resources :clients do
         get "/organization", to: "clients/organizations#edit", on: :member, as: :edit_organization
         patch "/organization", to: "clients/organizations#update", on: :member, as: :organization
+        get "/bai", to: "clients/bank_accounts#show", on: :member, as: :show_bank_account
```
sartan, line 123: Non-blocker: wondering if "bai" acronym is enough or if it would be better to just spell out `bank_account_info`.

spec/controllers/hub/clients/bank_accounts_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
+
```
sartan, line 2: Excellent! Nice and concise.

spec/features/hub/toggle_client_bank_account_info_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "rails_helper"
```
sartan, line 1: Wondering if it would be enough to test this as part of an existing flow in the `show_client_spec` (i.e. click "View" as we're validating other client fields").

sartan (approved):
Looks great! No blockers.

spec/features/hub/toggle_client_bank_account_info_spec.rb
bytheway875: The reason for breaking it out is that i need to require js for the test in order for it to work, and i wasnt sure if its just the SETUP for enabling js that takes a while or if it actually slows down the rest of the scenario. Since I couldn't tell, I tried to isolate it... not sure if that makes sense or makes it worse.

PR #665, opened 2020-12-15
Server Side Bank Account Information Approach

bytheway875:
I wanted to let people take a look at the spike on showing the bank account info via server side request, which I ultimately dropped and replaced with a javascript implementation due to security and UX concerns.

### DIFF
--- a/app/views/hub/clients/show.html.erb
+++ b/app/views/hub/clients/show.html.erb
@@ -114,10 +114,77 @@
       </div>
     </div>
 
+    <div id="bai">
+      <div class="spacing-below-35">
+        <h2 class="text--small text--bold"><%= "Bank Account Info" %>
+          <span>
+            <% if @client.bank_account_info? %>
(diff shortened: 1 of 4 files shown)


app/views/hub/clients/show.html.erb
```diff
@@ -114,10 +114,77 @@
...
   </div>
 
+
   <%= render "client_take_action_footer" %>
+
 <% end %>
+
+<% content_for :script do %>
+  <script>
+      if (window.location.search.includes("bai")) {
```
sartan, line 178: Not really important since we're not using this branch, but using `URLSearchParams` would help here, so we don't accidentally match all words that contain  "bai" (e.g. "bail" or "bailey").

https://developer.mozilla.org/en-US/docs/Web/API/Location/search

bytheway875: I have another PR coming up soon for some search params stuff where it can be super simplified using URLSearchParams! Thanks for the tip :D

PR #666, opened 2020-12-15
Only update one tax return at a time on a single client macro page [#176016581]

bytheway875:

### DIFF
--- a/app/forms/hub/take_action_form.rb
+++ b/app/forms/hub/take_action_form.rb
@@ -1,10 +1,28 @@
 module Hub
   class TakeActionForm < Form
-    attr_accessor :tax_returns, :locale, :message_body, :contact_method, :internal_note_body
+    include MessageSending
+    attr_accessor :tax_return,
+                  :tax_return_id, 
+                  :tax_returns, 
+                  :status, 
+                  :locale, 
+                  :message_body, 
+                  :contact_method, 
+                  :internal_note_body, 
+                  :action_list
+    validate :belongs_to_client
+    validate :status_has_changed
 
-    def initialize(client, *args, **attributes)
+    def initialize(client, current_user, *args, **attributes)
       @client = client
+      @current_user = current_user
+      @tax_returns = client.tax_returns.order(year: :asc)
+      @action_list = []
       super(*args, **attributes)
+
+      set_default_locale if @locale.blank?
+      set_default_message_body if @message_body.nil?
+      set_default_contact_method if @contact_method.nil?
     end
 
     def language_difference_help_text
@@ -40,14 +58,94 @@ module Hub
       end
     end
 
+    def take_action
+      return false unless valid?
+
+      tax_return.status = status
+      if tax_return.save
+        SystemNote.create_status_change_note(@current_user, tax_return)
+        @action_list << I18n.t("hub.clients.update_take_action.flash_message.status")
+        send_message if message_body.present?
+        create_note if internal_note_body.present?
+      end
+      true
+    end
+
     def self.permitted_params
-      [:tax_returns, :locale, :message_body, :contact_method, :internal_note_body, { tax_returns: {} }]
+      [:tax_return_id, :status, :locale, :message_body, :contact_method, :internal_note_body]
+    end
+
+    def tax_return
+      @tax_return ||= @client.tax_returns.find_by(id: tax_return_id)
     end
 
     private
 
+    def send_message
+      case contact_method
+      when "email"
+        send_email(message_body, subject_locale: locale)
+        @action_list << I18n.t("hub.clients.update_take_action.flash_message.email")
+      when "text_message"
+        send_text_message(message_body)
+        @action_list << I18n.t("hub.clients.update_take_action.flash_message.text_message")
+      end
+    end
+
+    def create_note
+      Note.create!(
+          body: internal_note_body,
+          client: @client,
+          user: @current_user
+      )
+      @action_list << I18n.t("hub.clients.update_take_action.flash_message.internal_note")
+    end
+
+    def current_user
+      @current_user
+    end
+
     def language_label(key)
       I18n.t("general.language_options.#{key}")
+    end
+
+    def set_default_message_body
+      @message_body = case status
+                      when "intake_more_info", "prep_more_info", "review_more_info"
+                        document_list = @client.intake.relevant_document_types.map do |doc_type|
+                          "  - " + doc_type.translated_label(@client.intake.locale)
+                        end.join("\n")
+                        I18n.t(
+                            "hub.status_macros.needs_more_information",
+                            required_documents: document_list,
+                            document_upload_link: @client.intake.requested_docs_token_link,
+                            locale: @client.intake.locale
+                        )
+                      when "prep_ready_for_review"
+                        I18n.t("hub.status_macros.ready_for_qr", locale: @client.intake.locale)
+                      when "filed_accepted"
+                        I18n.t("hub.status_macros.accepted", locale: @client.intake.locale)
+                      else
+                        ""
+                      end
+    end
+
+    def set_default_contact_method
+      default = "email"
+      prefers_sms_only = @client.intake.sms_notification_opt_in_yes? && @client.intake.email_notification_opt_in_no?
+      @contact_method = prefers_sms_only ? "text_message" : default
+    end
+
+    def set_default_locale
+      @locale = @client.intake.locale
+    end
+
+    def status_has_changed
+      errors.add(:status, I18n.t("forms.errors.status_must_change")) if status == tax_return&.status
+    end
+
+    def belongs_to_client
+      errors.add(:tax_return_id, I18n.t("forms.errors.tax_return_belongs_to_client")) unless tax_return.present?
     end
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -56,13 +56,12 @@ en:
         language_mismatch: This client requested %{interview_language} for their interview
         prefers_one_contact_method: This client prefers %{preferred} instead of %{other_method}
         send_clarifier: By clicking send, you will also update status, send a team note, and update followers.
-        status: Status of %{tax_year} tax return
       update_take_action:
         flash_message:
           email: sent email
-          internal_note: added internal note.
+          internal_note: added internal note
           status: updated status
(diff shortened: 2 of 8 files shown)


bytheway875:
Refactored to move a bunch of logic from the controller into the take action form. Also, per conversation with product we're requiring a changed status to use this form -- if a user wants to simply write a message or a note, you can do that on the notes or messages page directly.

spec/forms/hub/take_action_form_spec.rb
```diff
@@ -93,4 +197,162 @@
       end
     end
   end
+  describe "#take_action!" do
+    context "when tax_return_id is not part of clients tax returns" do
+      let(:intake) { create :intake }
+      let(:form) { Hub::TakeActionForm.new(client, current_user, { tax_return_id: create(:tax_return).id }) }
+
+      it "throws an error" do
+        expect { form.take_action! }.to raise_error TakeActionFormError
```
alaroia-cfa, line 206: I saw this custom exception, and I realized we make custom exceptions rarely, and I think I figured out a pattern by which we can avoid creating the custom exception. Non-blocking suggestion:

One pattern I've seen is to use ActiveModel validators within forms. This way, the controller can use a pattern like

```
if @take_action_form.save
  redirect_to it_worked_path
else
  render :edit
end
```

Then the TakeActionForm's `save` method might look something like:

```
def save
  return false unless valid?
  
   doActualStuff()
end
```

It would require renaming `take_action!` to `save`, so it would be a somewhat sizable change. Curious for your thoughts about that approach. (Non-blocking.)

spec/forms/hub/take_action_form_spec.rb
```diff
@@ -93,4 +197,162 @@
...
+      let(:tax_return) { create :tax_return, client: client, year: 2019 }
+
+      context "only status change" do
+        let(:form) { Hub::TakeActionForm.new(client, current_user, { tax_return_id: tax_return.id, status: "intake_more_info", message_body: "" }) }
+
+        it "changes the status of the client" do
+          expect {
+            form.take_action!
+            tax_return.reload
+          }.to change(tax_return, :status).to "intake_more_info"
```
alaroia-cfa, line 231: Putting `.reload` in the block passed to `expect { ... }` is an nice approach!

spec/forms/hub/take_action_form_spec.rb
```diff
@@ -93,4 +197,162 @@
...
+
+        it "does not create a note" do
+          expect {
+            form.take_action!
+          }.not_to change(Note, :count)
+        end
+
+        it "creates an action list that only includes changing status" do
+          form.take_action!
+          expect(form.action_list).to eq ["updated status", "sent email"]
```
alaroia-cfa, line 308: The test says "only includes changing status" (L306) but the assertion includes "sent email" (L308). Is that an inconsistency?

spec/forms/hub/take_action_form_spec.rb
```diff
@@ -93,4 +197,162 @@
...
+        end
+      end
+
+      context "status change with note" do
+        let(:internal_note_body) { "hi" }
+        let(:form) { Hub::TakeActionForm.new(client, current_user, { tax_return_id: tax_return.id, internal_note_body: internal_note_body, message_body: "", status: "intake_more_info" }) }
+        it "creates an action list that only includes changing status" do
```
alaroia-cfa, line 315: Same here - the `it...` maybe is inconsistent.

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -804,91 +811,27 @@
...
       end
 
       context "when status is changed, message body is present, and internal note is present" do
         let(:new_status_2019) { "review_in_review" }
         let(:message_body) { "hi" }
         let(:internal_note_body) { "wyd" }
-        before { allow(subject).to receive(:send_email) }
+
+        before do
+          allow_any_instance_of(Hub::TakeActionForm).to receive(:send_email).and_return true
```
alaroia-cfa, line 833: Can you get away without `.and_return true` here? Just curious.

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -804,91 +811,27 @@
...
-
-      context "when the message body is blank" do
-        let(:message_body) { " \n" }
-        let(:contact_method) { "email" }
-
-        it "does not send a message using the chosen contact method" do
-          expect do
-            post :update_take_action, params: params
-          end.not_to change(OutgoingEmail, :count)
+          expect(flash[:notice]).to include "added internal note"
```
alaroia-cfa, line 823: Consider mocking the `TakeActionForm` object, or some state on it, in these controller specs. As I understand it, "added internal note" is text that comes from `TakeActionForm`, so it might be advantageous to avoid testing it here.

config/locales/en.yml
```diff
@@ -56,13 +56,12 @@ en:
...
         send_clarifier: By clicking send, you will also update status, send a team note, and update followers.
-        status: Status of %{tax_year} tax return
       update_take_action:
         flash_message:
           email: sent email
-          internal_note: added internal note.
+          internal_note: added internal note
           status: updated status
-          success: 'Success: Action taken! %{action_list}'
+          success: 'Success: Action taken! %{action_list}.'
```
alaroia-cfa, line 64: Nice job taking care of punctuation!

app/forms/hub/take_action_form.rb
```diff
@@ -40,14 +48,86 @@ def contact_method_help_text
...
+    def create_note
+      Note.create!(
+          body: internal_note_body,
+          client: @client,
+          user: @current_user
+      )
+      @action_list << I18n.t("hub.clients.update_take_action.flash_message.internal_note")
+    end
+
+    def current_user
```
alaroia-cfa, line 90: This makes the `Model` into sort-of a controller. I wonder if this is evidence that MessageSending is the wrong abstraction, and perhaps `send_email` needs to live on `OutgoingEmail`, or perhaps somewhere else. Non-blocking, but curious for your take.

app/forms/hub/take_action_form.rb
```diff
@@ -40,14 +48,86 @@ def contact_method_help_text
...
+      @contact_method = prefers_sms_only ? "text_message" : default
+    end
+
+    def set_default_locale
+      @locale = @client.intake.locale
+    end
   end
 end
+
+class TakeActionFormError < StandardError
```
alaroia-cfa, line 131: See notes later-on, in the test suite, for how we can avoid needing this exception if the form uses `.save`.

app/forms/hub/take_action_form.rb
```diff
@@ -40,14 +48,86 @@ def contact_method_help_text
...
+        @action_list << I18n.t("hub.clients.update_take_action.flash_message.status")
+        SystemNote.create_status_change_note(@current_user, tax_return)
+        send_message if message_body.present?
+        create_note if internal_note_body.present?
+      end
+    end
+
     def self.permitted_params
-      [:tax_returns, :locale, :message_body, :contact_method, :internal_note_body, { tax_returns: {} }]
+      [:tax_return_id, :status, :locale, :message_body, :contact_method, :internal_note_body]
```
alaroia-cfa, line 65: Nice!

alaroia-cfa (approved):
This is great work! I left some non-blocking suggestions and questions.

spec/controllers/hub/clients_controller_spec.rb
bytheway875: yes i can!

app/forms/hub/take_action_form.rb
bytheway875: agreed! I think we should have those be class methods on the model, or some kind of service object instead of a controller mixin -- I think we're going to be wanting to send emails /messages from a variety of places, not just controllers

app/forms/hub/take_action_form.rb
```diff
@@ -1,10 +1,28 @@
 module Hub
   class TakeActionForm < Form
-    attr_accessor :tax_returns, :locale, :message_body, :contact_method, :internal_note_body
+    include MessageSending
+    attr_accessor :tax_return,
+                  :tax_return_id, 
+                  :tax_returns,
```
alaroia-cfa, line 6: I wonder if you need all 3 - tax_return, tax_returns (plural), tax_return_id (id). Given that you have your own attribute accessor for tax_return `def tax_return` below, you may be able to remove all three.

Non-blocking.

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -785,117 +779,34 @@
...
           post :update_take_action, params: params
-
+          expect(response).to redirect_to hub_client_path(id: client.id)
           expect(flash[:notice]).to eq "Success: Action taken! Updated status, sent email, added internal note."
         end
       end
+
+
     end
   end
```
alaroia-cfa, line 811: This is nice and clear IMHO!

spec/forms/hub/take_action_form_spec.rb
```diff
@@ -15,7 +119,7 @@
 
     context "when the locale and preferred interview language match" do
       let(:intake) { create :intake, preferred_interview_language: "es" }
-      let(:form) { Hub::TakeActionForm.new(client, locale: "es") }
+      let(:form) { Hub::TakeActionForm.new(client, current_user, locale: "es") }
 
       it "returns nil" do
         expect(form.language_difference_help_text).to be_nil
```
alaroia-cfa, line 125: Seeing this change makes me wonder if `TakeActionForm#take_action` should be the one to take the user parameter, since I think it's not used in any other methods. 100% non-blocking, but wanted to mention that.

alaroia-cfa (approved):
Looks even crisper than before! I have some comments which are 100% non-blocking.

spec/forms/hub/take_action_form_spec.rb
bytheway875: oh interesting! i hadn't thought of that but that's clutch and makes a ton of sense

app/forms/hub/take_action_form.rb
bytheway875: i think i still need tax_return_id like that in order to grab tax_return_id the way I do in the tax_return method, and then tax_returns is used to set up the list of tax returns in the dropdowns, so unfortunately i think i do need all three!

PR #642, opened 2020-12-09
Coalitions orgs sites #175978608

bengolder:

PR #669, opened 2020-12-16
Coalitions, orgs, and sites

alaroia-cfa:
This is a bit of a large pull request.

It takes the work done in https://github.com/codeforamerica/vita-min/pull/642/commits/8b51d5177baabcbd45be06b83ab91c93befe30b0 and migrates it to a model where a `VitaPartner` object can be either an Organization or a Site. The `VitaPartner.organizations` and `VitaPartner.sites` scopes allow picking just the right ones.

This completes the story in theory, but I will probably take some time to delete `.display_name`, `.zendesk_group_id`, and `.zendesk_instance_domain` from `VitaPartner` before picking up the next story.

After merging this PR, I will remove the "VITA Partners" button from the profile page, too.

### DIFF
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -1,0 +1,34 @@
+module Hub
+  class OrganizationsController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    load_and_authorize_resource :vita_partner, parent: false
+
+    layout "admin"
+
+    def new
+      @coalitions = Coalition.all
+    end
+
+    def create
+      render :new unless @vita_partner.save!
+
+      redirect_to hub_organizations_path
+    end
+
+    def index
+      @coalitions = Coalition.includes(:organizations)
+      @organizations = @vita_partners.organizations
+      @independent_organizations = @vita_partners.organizations.where(coalition: nil)
+    end
+
+    private
+
+    def vita_partner_params
+      params.require(:vita_partner).permit(:name, :coalition_id).merge(
+        zendesk_group_id: "unused",
+        zendesk_instance_domain: "unused"
+      )
+    end
+  end
+end
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -11,15 +11,29 @@ module Hub
     end
 
     def create
-      render :new unless @vita_partner.save!
-
-      redirect_to hub_organizations_path
+      @coalitions = Coalition.all
+      if @vita_partner.save
+        redirect_to hub_organizations_path
+      else
+        render :new
+      end
     end
 
     def index
       @coalitions = Coalition.includes(:organizations)
       @organizations = @vita_partners.organizations
       @independent_organizations = @vita_partners.organizations.where(coalition: nil)
+    end
+
+    def edit
+      @coalitions = Coalition.all
+      @organization = @vita_partner
+    end
+
+    def update
+      render :edit unless @vita_partner.update(vita_partner_params)
+
+      redirect_to hub_organizations_path
     end
 
     private
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -8,13 +8,14 @@ module Hub
 
     def new
       @coalitions = Coalition.all
+      @organization = @vita_partner
     end
 
     def create
-      @coalitions = Coalition.all
       if @vita_partner.save
         redirect_to hub_organizations_path
       else
+        @coalitions = Coalition.all
         render :new
       end
     end
--- a/app/controllers/hub/sites_controller.rb
+++ b/app/controllers/hub/sites_controller.rb
@@ -1,0 +1,47 @@
+module Hub
+  class SitesController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    before_action :load_organizations, only: [:new, :edit]
+
+    load_and_authorize_resource :vita_partner, parent: false
+
+    layout "admin"
+
+    def new
+      @site = VitaPartner.new(parent_organization_id: params[:parent_organization_id])
+    end
+
+    def create
+      @site = VitaPartner.new(vita_partner_params)
+      if @site.save
+        redirect_to edit_hub_organization_path(id: @site.parent_organization)
+      else
+        render :new
+      end
+    end
+
+    def edit
+      @site = @vita_partner
+    end
+
+    def update
+      render :edit unless @vita_partner.update(vita_partner_params)
+
+      redirect_to edit_hub_organization_path(id: @vita_partner.parent_organization_id)
+    end
+
+    private
+
+    def load_organizations
+      @organizations = VitaPartner.accessible_by(current_ability).organizations
+    end
+
+    def vita_partner_params
+      params.require(:vita_partner).permit(:name, :parent_organization_id).merge(
+        zendesk_instance_domain: "unused",
+        zendesk_group_id: "unused",
+      )
+    end
+  end
+end
--- a/app/models/coalition.rb
+++ b/app/models/coalition.rb
@@ -13,6 +13,6 @@
 #
 class Coalition < ApplicationRecord
   validates :name, uniqueness: true
-  # A coalition has many Organizations. We use the VitaPartner.organizations scope to ensure we only find Organizations, not Sites.
+  # Use the VitaPartner.organizations scope to ensure we only find Organizations, not Sites.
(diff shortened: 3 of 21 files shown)


app/controllers/hub/organizations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    load_and_authorize_resource :vita_partner, parent: false
+
+    layout "admin"
+
+    def new
+      @coalitions = Coalition.all
+    end
+
+    def create
+      @coalitions = Coalition.all
```
bytheway875, line 14: Is @coalitions only used in the case where the organization.save fails and rerenders new? In that case, might be better to move it into the block that renders :new vs always calling it!

app/controllers/hub/organizations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      render :edit unless @vita_partner.update(vita_partner_params)
+
+      redirect_to hub_organizations_path
+    end
+
+    private
+
+    def vita_partner_params
+      params.require(:vita_partner).permit(:name, :coalition_id).merge(
+        zendesk_group_id: "unused",
```
bytheway875, line 43: gross -- hoping we can delete that dead code soon!

app/controllers/hub/sites_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    include AccessControllable
+    before_action :require_sign_in
+    before_action :load_organizations, only: [:new, :edit]
+
+    load_and_authorize_resource :vita_partner, parent: false
+
+    layout "admin"
+
+    def new
+      @site = VitaPartner.new(parent_organization_id: params[:parent_organization_id])
```
bytheway875, line 12: I think that CanCanCan might be initializing a new object called @vita_partner secretly... might make more sense to just authorize_resource vita_partner for new and create since we don't need the objects they instantiate on load!

app/models/coalition.rb
```diff
@@ -13,4 +13,6 @@
 #
 class Coalition < ApplicationRecord
   validates :name, uniqueness: true
+  # A coalition has many Organizations. We use the VitaPartner.organizations scope to ensure we only find Organizations, not Sites.
+  has_many :organizations, -> { organizations }, class_name: "VitaPartner", foreign_key: "coalition_id"
```
bytheway875, line 17: If I'm understanding this right, these are organizations because only organizations store the coalition_id, so something having a coalition_id means that it is an organization.

app/models/vita_partner.rb
```diff
@@ -45,6 +45,9 @@ class VitaPartner < ApplicationRecord
   scope :top_level, -> { where(parent_organization: nil).order(:display_name).order(:name) }
   scope :organizations, -> { where(parent_organization: nil) }
   scope :sites, -> { where.not(parent_organization: nil) }
+  has_many :child_sites, -> { order(:id) }, class_name: "VitaPartner", foreign_key: "parent_organization_id"
```
bytheway875, line 48: is this called child_sites because we already have a scope called sites? I don't know if it would make it super confusing, but since the scope is a class method and the relationship is an instance method (I think), I'm pretty sure we could call both sites.

app/models/vita_partner.rb
```diff
@@ -45,6 +45,9 @@ class VitaPartner < ApplicationRecord
...
+
+  default_scope { includes(:child_sites) }
```
bytheway875, line 50: This feels like an appropriate use of a default_scope!

app/controllers/hub/organizations_controller.rb
alaroia-cfa: I have a local commit ready for that :D as soon as this gets merged :D

app/views/hub/organizations/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<% @title = t(".title") %>
+<% content_for :page_title, @title %>
+<% content_for :card do %>
+  <div class="hub-form">
+    <%= form_with model: @vita_partner, url: hub_organizations_path, method: :post, local: true, builder: VitaMinFormBuilder do |f| %>
```
bytheway875, line 5: These forms always confuse me ... I was wondering how the model: @vita_partner worked (normally im needing to do [:hub, @vita_partner]), but then i realized that you've got an explicit url + method which overrides it. makes sense

app/models/vita_partner.rb
alaroia-cfa: That's right - called child_sites b/c the scope is called sites.

I thought about relying on the instance method vs. class method distinction. I think it's less confusing if they have separate names.

app/views/hub/organizations/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<% content_for :card do %>
+  <div class="hub-form">
+    <%= link_to t("general.all_organizations"), hub_organizations_path, class: "button button--small" %>
+
+    <%= form_with model: @organization, url: hub_organizations_path, method: :post, local: true, builder: VitaMinFormBuilder do |f| %>
+      <h1 class="form-card__title">
+        <%= @title %>
+      </h1>
+
+      <%= f.cfa_input_field :name, t("general.name") %>
```
bytheway875, line 12: This section looks repeated between new and edit -- it might be worth pulling this out into a partial so that we make sure we're setting up the coalition select the same way, but its not blocking.

app/views/hub/organizations/new.html.erb
alaroia-cfa: Oh interesting! Is it clearer if we remove the explicit `url:` here? Earnestly curious.

config/locales/es.yml
```diff
@@ -195,6 +195,22 @@ es:
       add_site: Agregar sitio
       index:
         title: Socios de VITA
+    organizations:
```
bytheway875, line 198: Nitpicky nitpick, but I think most of these don't need quote marks around it.

app/views/hub/organizations/edit.html.erb
alaroia-cfa: Good eye. I lean toward duplicating it if it's just these two places, since it's two lines, but I can't say I really thought about it before you mentioned it.

config/routes.rb
```diff
@@ -135,6 +135,8 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
       resources :tax_returns, only: [] do
         patch "update_certification", to: "tax_returns/certifications#update", on: :member
       end
+      resources :organizations, only: [:index, :create, :new, :edit, :update]
```
bytheway875, line 138: "except" also exists here, so you could change this to except: :destroy if you felt compelled

config/locales/es.yml
alaroia-cfa: :) True - I think RubyMine added them for me, and I didn't stop it.

It might be a good time to run i18n-tasks normalize anyway immediately after this gets merged. I've validated locally (in the past 30 seconds) that it will remove the unnecessary quotation marks.

bytheway875 (approved):

app/views/hub/organizations/edit.html.erb
```diff
@@ -0,0 +1,31 @@
+<% @title = @organization.name %>
+<% content_for :page_title, @title %>
+<% content_for :card do %>
+  <div class="hub-form">
+    <%= link_to t("general.all_organizations"), hub_organizations_path, class: "button button--small" %>
+
+    <%= form_with model: @organization, url: hub_organizations_path, method: :post, local: true, builder: VitaMinFormBuilder do |f| %>
```
bytheway875, line 7: Try `model: [:hub, @organization]` and removing `url: ...`

app/views/hub/organizations/new.html.erb
bytheway875: Consider calling this `@organization`

spec/features/hub/create_organization_hierarchy_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      expect(page).to have_selector("h2", text: "Koala Koalition")
+      expect(page).to have_selector("h3", text: "Orangutan Organization")
+
+      # create a new organization
+      click_on "New Organization"
+      fill_in "Name", with: "Origami Organization"
+      select "Koala Koalition", from: "Coalition"
+      click_on "Save"
+
+      # update the organization
```
bytheway875, line 25: duplicated with L26

app/controllers/hub/sites_controller.rb
alaroia-cfa: I couldn't get this proposed change to work, and I am going to punt on it.

I do have a remaining question, though: `vita_partner_params` does a `require(:vita_partner)` - so... how can it possibly apply to the `new` method?

PR #674, opened 2020-12-16
Change message when status changes

bytheway875:


bytheway875:
Reviewed over tuple with @sniemeyer13 !!

PR #677, opened 2020-12-17
Use text message id instead of object when sending texts to job

bytheway875:


bytheway875:
Reviewed / Paired with @sniemeyer13

PR #685, opened 2020-12-22
Synthetic Notes for Outbound Calls + Webhook on dial event

bytheway875:
Screen Shot 2020-12-22 at 1 42 34 PM

### DIFF
--- a/app/controllers/hub/outbound_calls_controller.rb
+++ b/app/controllers/hub/outbound_calls_controller.rb
@@ -19,6 +19,7 @@ module Hub
       @outbound_call = OutboundCall.find(params[:id])
     end
 
+    # The form that posts to update is on the show page, and only includes a textarea to update the note.
     def update
       @outbound_call = OutboundCall.find(params[:id])
       @outbound_call.update(params.require(:outbound_call).permit(:note))
@@ -28,6 +29,8 @@ module Hub
     def new
       @form = OutboundCallForm.new(client: @client, user: current_user)
     end
+
+    private
 
     def permitted_params
       params.require(OutboundCallForm.form_param).permit(:user_phone_number, :client_phone_number)
--- a/app/controllers/hub/outbound_calls_controller.rb
+++ b/app/controllers/hub/outbound_calls_controller.rb
@@ -19,6 +19,12 @@ module Hub
       @outbound_call = OutboundCall.find(params[:id])
     end
 
+    def update
+      @outbound_call = OutboundCall.find(params[:id])
+      @outbound_call.update(params.require(:outbound_call).permit(:note))
+      redirect_to hub_client_notes_path(client_id: @client.id)
+    end
+
     def new
       @form = OutboundCallForm.new(client: @client, user: current_user)
     end
--- a/app/controllers/twilio_webhooks_controller.rb
+++ b/app/controllers/twilio_webhooks_controller.rb
@@ -12,26 +12,8 @@ class TwilioWebhooksController < ActionController::Base
     return unless call.present?
 
     update_params = { twilio_status: params["CallStatus"] }
-    update_params[:call_duration] = params["CallDuration"] if params["CallDuration"].present?
+    update_params[:twilio_call_duration] = params["CallDuration"] if params["CallDuration"].present?
     call.update(update_params)
-  end
-
-  # This url is used to provide instructions to Twilio for how to handle calls from the user to the client from the
-  # OutboundCallForm. Once the user picks up and connects, we dial the clients phone number.
-  def dial_client
-    @outbound_call = OutboundCall.find(params[:id])
-    twiml = Twilio::TwiML::VoiceResponse.new
-    twiml.say(message: 'Please wait while we connect your call.')
-    # The status callback for the call is attached to the dial event to the client.
-    # This means that the length of the call will be based on how long the user was connected to the client,
-    # And the status will be based on whether the client picked up the call.
-    twiml.dial do |dial|
-      dial.number(@outbound_call.to_phone_number,
-                  status_callback_event: 'answered completed',
-                  status_callback: _outbound_call_webhook_url(@outbound_call),
-                  status_callback_method: 'POST')
-    end
-    render xml: twiml.to_xml
   end
 
   def create_incoming_text_message
@@ -76,16 +58,5 @@ class TwilioWebhooksController < ActionController::Base
 
   def validate_twilio_request
     return head 403 unless TwilioService.valid_request?(request)
-  end
-
-  def _outbound_call_webhook_url(outbound_call)
-    params = { id: outbound_call.id, locale: nil }
-
-    if Rails.env.development?
-      raise NgrokNeededError unless Rails.configuration.try(:ngrok_url).present?
-
-      return Rails.configuration.ngrok_url + outbound_calls_webhook_path(params)
-    end
-    outbound_calls_webhook_url(params)
   end
 end
--- a/app/controllers/twilio_webhooks_controller.rb
+++ b/app/controllers/twilio_webhooks_controller.rb
@@ -15,8 +15,10 @@ class TwilioWebhooksController < ActionController::Base
     update_params[:call_duration] = params["CallDuration"] if params["CallDuration"].present?
     call.update(update_params)
   end
-  
-  def dial
+
+  # This url is used to provide instructions to Twilio for how to handle calls from the user to the client from the
+  # OutboundCallForm. Once the user picks up and connects, we dial the clients phone number.
+  def dial_client
     @outbound_call = OutboundCall.find(params[:id])
     twiml = Twilio::TwiML::VoiceResponse.new
     twiml.say(message: 'Please wait while we connect your call.')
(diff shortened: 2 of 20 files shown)


app/controllers/hub/outbound_calls_controller.rb
```diff
@@ -19,6 +19,12 @@ def show
       @outbound_call = OutboundCall.find(params[:id])
     end
 
+    def update
+      @outbound_call = OutboundCall.find(params[:id])
+      @outbound_call.update(params.require(:outbound_call).permit(:note))
```
jenny-heath, line 24: curious about a couple of things:
- why are the permitted params here not separated into a separate method like the ones in `create`?
- why is the existing `permitted_params` method not private?

also, did github change something about where you can comment? i wanted to leave a comment on `create` above just to say "i don't understand it but i assume it has already been through review?" but it wouldn't let me :/

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -15,13 +15,20 @@ def update_outbound_call
     update_params[:call_duration] = params["CallDuration"] if params["CallDuration"].present?
     call.update(update_params)
   end
-
+  
   def dial
```
jenny-heath, line 19: possibly silly question: why is it that `OutboundCallForm` has a `dial` method and so does `TwilioWebhooksController` and so does `Twilio::TwiML::VoiceResponse`? seems a bit confusing

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -15,13 +15,20 @@ def update_outbound_call
...
     twiml = Twilio::TwiML::VoiceResponse.new
     twiml.say(message: 'Please wait while we connect your call.')
-    twiml.dial(number: @outbound_call.to_phone_number)
-
+    # The status callback for the call is attached to the dial event to the client.
+    # This means that the length of the call will be based on how long the user was connected to the client,
+    # And the status will be based on whether the client picked up the call.
+    twiml.dial do |dial|
+      dial.number(@outbound_call.to_phone_number,
+                  status_callback_event: 'answered completed',
```
jenny-heath, line 28: what does answer completed mean? is the status event always the same when we get into this block? should we save things like this status callback event as constants instead of just having the string here?

i realize a lot of these questions could probably be answered by the API documentation so maybe they are unnecessary

app/presenters/notes_presenter.rb
```diff
@@ -4,6 +4,7 @@ def self.grouped_notes(client)
     # Generate notes from documents on-the-fly, so the summary can change as clients upload more documents.
     notes = (client.notes.includes(:user) + SystemNote.where(client: client))
     synthetic_notes = synthetic_notes_from_documents(client.documents.order(created_at: :asc))
+    synthetic_notes += synthetic_notes_from_outbound_calls(client.outbound_calls.order(created_at: :asc))
```
jenny-heath, line 7: why do we call them synthetic notes? 🤔

app/presenters/notes_presenter.rb
```diff
@@ -20,6 +21,25 @@ def self.synthetic_notes_from_documents(documents)
...
 
+  def self.synthetic_notes_from_outbound_calls(outbound_calls)
+    outbound_calls.map do |outbound_call|
+      next if outbound_call.twilio_status == "queued"
+      duration = ""
+      if outbound_call.call_duration
+        duration_minutes = outbound_call.call_duration&.to_i / 60
+        duration_seconds = outbound_call.call_duration&.to_i % 60
+        duration += "#{duration_minutes}m" if duration_minutes > 0
+        duration += "#{duration_seconds}s"
```
jenny-heath, line 32: i'm curious whether there's some ruby function that would allow you to just get the formatted string from the number...

jenny-heath:
I realized in reviewing this that I really should understand more about how Twilio calls work. I left comments on what I could but it seems like it would be most efficient if someone could just review this with you on Tuple

app/controllers/hub/outbound_calls_controller.rb
bytheway875: I should leave a note on this method -- in create, we're creating the object and dialing the number to start the call. That then directs you to the show page that includes some info about the call and a "notes" textarea. If the user enters notes and then submits, the object will be updated with the note.

app/controllers/twilio_webhooks_controller.rb
bytheway875: OutboundCallForm has a dial method that dials the call from our twilio number to the VITA User. When we instantiate the call object, we pass it the `dial` method from the TwilioWebhooksController (which I renamed to dial_client and added a comment to hopefully help with the confusion), which is a method that Twilio will post to when the call with the User connects so that we can then connect them to the client. The `Twilio::TwiML::VoiceResponse` dial builds a dial node into the XML response so that twilio knows to call the user.

app/controllers/twilio_webhooks_controller.rb
bytheway875: I don't think they need to be constants because what we want to track could change per call we make depending on what/why we're calling and what we want to track about it.

These are a list of statuses that we're telling Twilio we want to be notified about. By default, the value for this is just "completed", but I added answered here as well so that if a call is in progress to the client, we have a note on the notes page that indicates that a call was initialized and it is in progress.

app/presenters/notes_presenter.rb
bytheway875: Notes for documents and outbound calls are not saved into the database -- they're created as we load the notes information onto the page for display purposes only. This allows us to group together document information to show all docs uploaded on the same day as one message, and show the latest statuses of the phone call. Basically, we're creating the bodies for the notes on the fly

app/controllers/hub/outbound_calls_controller.rb
```diff
@@ -19,6 +19,12 @@ def show
       @outbound_call = OutboundCall.find(params[:id])
     end
 
+    def update
```
bytheway875, line 22: abilities cancancan

app/presenters/notes_presenter.rb
```diff
@@ -20,6 +21,25 @@ def self.synthetic_notes_from_documents(documents)
...
+      end
+      note = I18n.t("hub.notes.index.outbound_call_note_html", user_name: outbound_call.user.name, client_name: outbound_call.client.preferred_name, status: outbound_call.twilio_status, duration: duration)
```
bytheway875, line 34: Consider renaming `note = ...` to `body` or `message`

app/models/outbound_call.rb
```diff
@@ -4,8 +4,8 @@
 #
 #  id                :bigint           not null, primary key
 #  call_duration     :string
-#  completed_at      :datetime
 #  from_phone_number :string           not null
+#  note              :text
```
bytheway875, line 8: I'd love to see a comment in this file (not in the autogenerated part) indicating the lifecycle of a call, so that when we're debugging the `twilio_status` later, we'll know `queued` could mean waiting for user (not client) to pick up.

config/environments/development.rb
```diff
@@ -73,4 +73,5 @@
   # Allow incoming connections over ngrok
   config.hosts << /[a-z0-9]+\.ngrok\.io/
   config.offseason = false
+  config.ngrok_url = "http://d0d376ec6d64.ngrok.io"
```
bytheway875, line 76: noisy :)

config/locales/en.yml
```diff
@@ -129,12 +129,20 @@ en:
         document_note:
           one: Client added 1 document.
           other: Client added %{count} documents.
+        outbound_call_note_html: |
```
bytheway875, line 132: Consider `outbound_call_synthentic_note_html`

app/models/outbound_call.rb
```diff
@@ -4,8 +4,8 @@
 #
 #  id                :bigint           not null, primary key
 #  call_duration     :string
```
bytheway875, line 6: Consider turning this into an integer in the database. Not a blocker.

spec/presenters/notes_presenter_spec.rb
```diff
@@ -61,5 +61,19 @@
...
+      let(:day1) { DateTime.new(2019, 10, 5, 8, 1).utc }
+      let(:day2) { DateTime.new(2020, 10, 5, 5, 1).utc }
+      let!(:outbound_call_completed) { create :outbound_call, twilio_status: "completed", created_at: day1, call_duration: "75", client: client, user: user, note: "I talked to them!" }
+      let!(:outbound_call_queued) { create :outbound_call, twilio_status: "queued", created_at: day2, client: client, user: user }
+      it "does not include outbound_calls in status queued" do
+        result = NotesPresenter.grouped_notes(client)
+        notes = result.values.flatten
+        expect(notes.length).to eq 1
+        expect(notes[0].body).to include "Called by Gary Gnome. Call was completed and lasted 1m15s."
+        expect(notes[0].body).to include "I talked to them!"
```
bytheway875, line 75: Consider keeping the `to eq 1` here, but moving the text assertions to another test.

bytheway875:
Reviewed by Asheesh over tuple, pushed a commit to address the majority of his feedback and Jennys!

PR #686, opened 2020-12-22
Create polymorphic User.role property [#175978650]

alaroia-cfa:
This change gives every User a `.role` that points at the actual role they will have. Each role is in a different table, due its different data storage needs, so this is polymorphic.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,7 @@
 #  provider                  :string
 #  reset_password_sent_at    :datetime
 #  reset_password_token      :string
+#  role_type                 :string
 #  sign_in_count             :integer          default(0), not null
 #  suspended                 :boolean
 #  ticket_restriction        :string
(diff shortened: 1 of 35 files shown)


app/models/user.rb
```diff
@@ -61,16 +64,18 @@ class User < ApplicationRecord
...
+  belongs_to :organization_lead_role, -> { where(users: { role_type: 'OrganizationLeadRole' }) }, foreign_key: 'role_id', optional: true
 
   attr_encrypted :access_token, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
 
   validates_presence_of :name
   validates_inclusion_of :timezone, in: ActiveSupport::TimeZone.country_zones("us").map { |tz| tz.tzinfo.name }
 
   def accessible_organizations
-    organization_lead_role = OrganizationLeadRole.find_by_user_id(id)
+    organization_lead_role = role.class.name == "OrganizationLeadRole"
```
embarnard, line 76: Could be `role_type`, which is always on `User`

embarnard (approved):
Looks good to me! A few non-blocking comments

PR #689, opened 2021-01-04
Update statuses and tests that used old statues and write script for migrating existing data

bytheway875:
We can delete the script and tests once we run it, but it was helpful to use TDD for making sure that each of the transitions was happening.

### DIFF
--- a/.aptible.yml
+++ b/.aptible.yml
@@ -1,3 +1,4 @@
 before_release:
     - bundle exec rake db:migrate
     - bundle exec rake db:insert_states
+    - bundle exec rake db:update_tax_return_statuses
--- a/.aptible.yml
+++ b/.aptible.yml
@@ -1,4 +1,3 @@
 before_release:
     - bundle exec rake db:migrate
     - bundle exec rake db:insert_states
-    - bundle exec rake db:update_tax_return_statuses
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -161,29 +161,28 @@ en:
         intake: Intake
         prep: Tax prep
         review: Quality review
-        finalize: Final steps
-        filed: Filing completed
+        file: Final steps
       status:
         intake_before_consent: Before consent
-        intake_in_progress: In progress
-        intake_open: Open
-        intake_review: In review
-        intake_more_info: Needs more information
+        intake_in_progress: Not ready
+        intake_ready: Ready for review
+        intake_reviewing: Reviewing
+        intake_ready_for_call: Ready for call
         intake_info_requested: Info requested
-        intake_needs_assignment: Needs assignment
-        prep_ready_for_call: Ready for call
-        prep_more_info: Needs more information
-        prep_preparing: Entering in TaxSlayer
-        prep_ready_for_review: Ready for QR
-        review_in_review: In review
-        review_more_info: Needs more information
-        review_complete_signature_requested: Complete/Signature requested
-        finalize_closed: Closed
-        finalize_signed: Return signed
-        filed_e_file: Return e-filed
-        filed_mail_file: Return filed by mail
-        filed_rejected: Rejected
-        filed_accepted: Accepted
+        prep_ready_for_prep: Ready for prep
+        prep_preparing: Preparing
+        prep_info_requested: Info requested
+        review_ready_for_qr: Ready for QR
+        review_reviewing: Reviewing
+        review_ready_for_call: Ready for call
+        review_signature_requested: Signature requested
+        review_info_requested: Info requested
+        file_ready_to_file: Ready to file
+        file_efiled: E-filed
+        file_mailed: Filed by mail
+        file_rejected: Rejected
+        file_accepted: Accepted
+        file_not_filing: Not filing
     status_macros:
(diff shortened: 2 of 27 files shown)


config/locales/en.yml
```diff
@@ -160,29 +160,28 @@ en:
...
-        finalize: Final steps
-        filed: Filing completed
+        file: Final steps
       status:
         intake_before_consent: Before consent
-        intake_in_progress: In progress
-        intake_open: Open
-        intake_review: In review
-        intake_more_info: Needs more information
+        intake_in_progress: Not ready
```
bytheway875, line 166: I updated a lot of the enum names of the statuses to reflect the new statuses, but in the case of "in progress" I decided to leave it as "in_progress" for developer facing stuff -- it seemed really weird to have a negative in the name of the status, especially because its not a "bad" thing

config/locales/en.yml
```diff
@@ -160,29 +160,28 @@ en:
...
-        filed_accepted: Accepted
+        prep_ready_for_prep: Ready for prep
+        prep_preparing: Preparing
+        prep_info_requested: Info requested
+        review_ready_for_qr: Ready for QR
+        review_reviewing: Reviewing
+        review_ready_for_call: Ready for call
+        review_signature_requested: Signature requested
+        review_info_requested: Info requested
+        file_ready_to_file: Ready to file
```
bytheway875, line 179: Changed filed to present tense "file" because this now includes some states surrounding the filing (ready) and the opt-out state (not-filing)

.aptible.yml
```diff
@@ -1,3 +1,4 @@
 before_release:
     - bundle exec rake db:migrate
-    - bundle exec rake db:insert_states
\ No newline at end of file
+    - bundle exec rake db:insert_states
+    - bundle exec rake db:update_tax_return_statuses
```
alaroia-cfa, line 4: I'd like to see the migration be done as a SQL data migration, or even a Ruby script that lives as a migration. I know we talked about some data migrations being slow, but I think this is a good time to use data migrations because of the the following reasons:

* The app isn't really in a good state if we upgrade the code without migrating the data.
* We have no prod users, so we can't really cause an outage.

I'm open to other ideas, including the approach you have here; this is not a blocker.

If it's helpful, I can try writing the SQL to do the migration, either as a comment on GitHub, or while pairing with you.

config/locales/en.yml
alaroia-cfa: Good idea!

config/locales/en.yml
alaroia-cfa: Seems OK to me!

spec/models/tax_return_spec.rb
```diff
@@ -92,29 +93,24 @@
...
 
     it "includes all intake statuses except before consent" do
-      expect(result["intake"].length).to eq 6
+      expect(result["intake"].length).to eq 5
       expect(result["intake"]).not_to include "intake_before_consent"
     end
 
     it "includes all prep statuses" do
-      expect(result["prep"].length).to eq 4
+      expect(result["prep"].length).to eq 3
```
alaroia-cfa, line 105: These tests are sort of hard to read. We have some repetition (all prep statuses, all review statuses, etc.). I wonder if we can merge them into one test and/or remove some of them. Not a blocker at all.

spec/services/migrate_statuses_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    it 'changes to 202 / prep_preparing' do
+      MigrateStatuses.migrate_all
+      tax_return.reload
+      expect(tax_return.status).to eq "prep_preparing"
+    end
+  end
+
+  context '204 (was prep_ready_for_qr)' do
+    let(:status) { 101 }
+    # 106 no longer exists -- force it by updating the column directly
```
alaroia-cfa, line 59: Stray comment?

spec/services/migrate_statuses_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    it 'changes to 401 / file_ready_to_file' do
+      MigrateStatuses.migrate_all
+      tax_return.reload
+      expect(tax_return.status).to eq "file_ready_to_file"
+    end
+  end
+
+  context '502 (was return signed)' do
+    let(:status) { 101 }
+    # 106 no longer exists -- force it by updating the column directly
```
alaroia-cfa, line 110: Stray comment?

spec/services/migrate_statuses_spec.rb
```diff
@@ -0,0 +1,163 @@
+require 'rails_helper'
```
alaroia-cfa, line 1: If we do switch the migration to a big SQL statement, we have the upside (?) that we can delete these tests. :)

alaroia-cfa (approved):
This is a great change. One big-ish suggestion - adjust the migration to be a SQL or otherwise data migration. Not a blocker. Happy to chat or provide input otherwise.

.aptible.yml
bytheway875: Awesome! We moved this into a migration file, but still using AR for now.

PR #690, opened 2021-01-04
Link to list view from all tabs on the client profile [#176211801]

alaroia-cfa:
## Note for reviewers

Is it OK I used `hub.clients.navigation.return_to_clients` as the i18n key? It seemed like 50/50 a good idea/too clever.

If you approve, please consider **merging directly**.

### Client profile (unchanged before & after)

image

### Messages list

image

### Document list

image

### Notes list

image

### Client profile

image

### Messages list

image

### Document list

image

### Notes list

image

### DIFF
--- a/app/views/hub/clients/show.html.erb
+++ b/app/views/hub/clients/show.html.erb
@@ -1,5 +1,4 @@
-<%= content_for :user_return_link_name, t(".return_to_clients") %>
-<%= content_for :user_return_link_path, hub_clients_path %>
+<%= content_for :back_to, "all_clients" %>
 
 <% content_for :page_title, "#{@client.preferred_name} ##{@client.id}" %>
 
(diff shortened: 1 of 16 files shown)


app/views/hub/clients/show.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= content_for :user_return_link_name, t(".return_to_clients") %>
+<%= content_for :user_return_link_name, t("hub.clients.navigation.return_to_clients") %>
 <%= content_for :user_return_link_path, hub_clients_path %>
```
bengolder, line 2: The repetition of these two lines across the different views seems like a clear cut case for including this in either a shared layout or in a shared partial. If refactored into a shared template, I would update the i18n key to use ".return_to_clients" and put the key whereever it needed to be based on the location of the layout/partial.

bengolder (approved):
Looks like it would work great! 
Could be refactored into a layout or partial to reduce repetition.

PR #692, opened 2021-01-05
Fix outbound calls

bengolder:

### DIFF
--- a/app/controllers/twilio_webhooks_controller.rb
+++ b/app/controllers/twilio_webhooks_controller.rb
@@ -54,7 +54,33 @@ class TwilioWebhooksController < ActionController::Base
     head :ok
   end
 
+  def outbound_call_connect
+    @outbound_call = OutboundCall.find(params[:id])
+    twiml = Twilio::TwiML::VoiceResponse.new
+    # The status callback for the call is attached to the dial event to the client.
+    # This means that the length of the call will be based on how long the user was connected to the client,
+    # And the status will be based on whether the client picked up the call.
+    twiml.dial do |dial|
+      dial.number(@outbound_call.to_phone_number,
+                  status_callback_event: 'answered completed',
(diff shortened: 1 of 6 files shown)


app/controllers/twilio_webhooks_controller.rb
```diff
@@ -54,8 +54,34 @@ def create_incoming_text_message
...
+                  status_callback_event: 'answered completed',
+                  status_callback: webhook_url,
+                  status_callback_method: 'POST')
+    end
+    render xml: twiml.to_xml
+  end
+
   private
 
+  def webhook_url
```
bytheway875, line 74: Webhook_url is too general now that its in the webhooks controller! update name.

spec/forms/hub/outbound_call_form_spec.rb
```diff
@@ -72,12 +72,11 @@
   end
 
   context "twiml" do
```
bytheway875, line 74: Move tests for this into the testing for #dial, expecting the twilio_client object to receive this parameter.

app/forms/hub/outbound_call_form.rb
```diff
@@ -46,35 +46,30 @@ def dial
...
+      # this code only executes if the user does not enter a digit within the timeout period (5 secs)
+      twiml.say(message: "We didn't hear from you so we're hanging up!")
+      twiml.hangup
       twiml.to_xml
     end
 
     private
 
-    def webhook_url
+    def connect_to_client_url
```
bytheway875, line 65: consider making this some sort of shareable method since we're using the pattern in multiple places...

bytheway875: test the behavior once there instead of multiple tests in different files for the same behavior.

bytheway875:
merged locally.

PR #693, opened 2021-01-05
Format all tables for accessibility, add scopes for table headers

embarnard:


app/views/hub/anonymized_intake_csv_extracts/index.html.erb
```diff
@@ -10,21 +10,21 @@
   <table class="data-table document-table">
     <thead>
     <tr>
-      <th class="extract-column__filename">
+      <th scope="col" class="extract-column__filename">
         <%= t(".filename") %>
       </th>
-      <th class="extract-column__record-count">
+      <th scope="col"  class="extract-column__record-count">
```
alaroia-cfa, line 16: There are two spaces after `"col". ` (could be `"col" `). Not a blocker at all.

alaroia-cfa (approved):
@embarnard This is cool! I'm curious if it has an impact on what screen readers do, or what WebAIM shows!

I didn't see a request for review on Slack, so I'm just showing up and commenting, so feel free to ignore this random comment.

app/views/hub/anonymized_intake_csv_extracts/index.html.erb
embarnard: good catch!

embarnard:
I misread the ticket as format all tables for accessibility and not just the all clients table but I assume there is not harm in making these additional changes. 

The all clients table includes "scopes" which should be enough to read for column and row headers. I am a little apprehensive about the Intakes section though. The screen reader would read the header of this section as the full string "Filing Year, Assignee, Certification, Status" and then all the associated items as one list. It might be more clear to organize "Filing Year, Assignee, Certification, Status"  in separate <th scope="col"> elements under an intake <th scope="colgroup"> item but this would change the look of the table.  

Screen Shot 2021-01-07 at 12 03 25 PM

There are some recommendations to add <caption> elements to tables to provide a description to what the table is but since we don't seem to have something analogous for users that don't use screen readers this would not be aligned to the design.

PR #694, opened 2021-01-06
Add coalition lead and site coordinator roles

jenny-heath:


app/models/admin_role.rb
```diff
@@ -7,4 +7,6 @@
 #  updated_at :datetime         not null
 #
 class AdminRole < ApplicationRecord
+  TYPE = "AdminRole"
+  DISPLAY_NAME = I18n.t("general.admin")
```
alaroia-cfa, line 11: I believe this implementation approach computes the translation string at module load time, which means it'll always be English. I could be wrong, but consider checking that.

alaroia-cfa:
I'm excited to see more roles 🥖

app/models/admin_role.rb
jenny-heath: ah i checked and this is true - i will merge this as is but will leave a note in the story

alaroia-cfa:
Works for me!

PR #698, opened 2021-01-07
Add uploaded_by to documents for record keeping and interaction tracking + shared examples for interaction tracking + adds documentation file to record current interaction tracking behavior.

bytheway875:


app/models/concerns/interaction_tracking.rb
```diff
@@ -3,16 +3,24 @@ module InteractionTracking
...
+      attention_needed_since: nil
+    )
   end
 
-  # When we take an "internal" action on a client (i.e. write a note, we want to update our last touch but NOT alter the attention flag)
+  # When we take an "internal" action on a client we only want to touch updated_at on the associated client without
+  # recording it as an "interaction" or clearing the needs_attention flag
   def record_internal_interaction
-    client&.touch(:last_interaction_at)
+    client&.touch
```
sniemeyer13, line 24: Does this no longer count as a last interaction?

app/controllers/hub/documents_controller.rb
```diff
@@ -25,6 +25,7 @@ def edit; end
     def create
       document_params[:upload].each do |file_upload|
         Document.create!(
+          uploaded_by: current_user,
```
bengolder, line 28: glad we are now recording this!

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -511,11 +511,13 @@
...
-        patch :attention_needed, params: params
-        client.reload
-        expect(client.attention_needed_since).to be_present
+        expect {
+          patch :attention_needed, params: params
+          client.reload
+        }.to change(client, :attention_needed_since)
+         .and change(client, :last_incoming_interaction_at)
       end
     end
```
bengolder, line 522: I think it would make sense to have two contexts for setting needs attention: the context where a client does not need attention (as already described here) and another where the client already needs attention (to make sure we don't inadvertently reduce the duration of attention_needed_since)

spec/support/shared_examples/interaction_tracking.rb
```diff
@@ -0,0 +1,34 @@
...
+
+shared_examples_for "an incoming interaction" do
+  it "updates the associated client" do
+    expect { subject.save }
+      .to change(subject.client, :attention_needed_since)
+      .and change(subject.client, :last_incoming_interaction_at)
+      .and change(subject.client, :last_interaction_at)
+      .and change(subject.client, :updated_at)
+  end
+end
```
bengolder, line 11: This doesn't seem to cover the case when a client already needs attention. It looks like if a client kept messaging us, each time it would reset attention_needed_since and last_incoming_interaction_at.

spec/models/client_spec.rb
```diff
@@ -195,6 +191,16 @@
           expect { create :document, intake: client.intake }.to change(client, :attention_needed_since)
         end
       end
+
+      context "when a user is uploading the document" do
+        it "does not update client last_interaction_at" do
```
sniemeyer13: should this update the last interaction at?

spec/support/shared_examples/interaction_tracking.rb
```diff
@@ -0,0 +1,34 @@
+RSpec::Matchers.define_negated_matcher :not_change, :change
+
```
sniemeyer13, line 2: This is super awesome!

spec/support/shared_examples/interaction_tracking.rb
```diff
@@ -0,0 +1,34 @@
...
+shared_examples_for "an outgoing interaction" do
+  # ensure attention_needed_since is set on the client so that we can test that it was cleared properly.
+  before { subject.client.attention_needed_since = Time.now }
+
+  it "updates the associated client" do
+    expect { subject.save }
+      .to change(subject.client, :attention_needed_since).to(nil)
+      .and change(subject.client, :last_interaction_at)
+      .and change(subject.client, :created_at)
```
sniemeyer13, line 46: Should this be updated at?

bengolder (approved):
Approved! Thanks for the changes!

PR #699, opened 2021-01-07
Client Authentication Spike

bengolder:


app/forms/portal/client_login_form.rb
```diff
@@ -0,0 +1,5 @@
+module Portal
+  class ClientLoginForm < Form
+    attr_accessor :last_four_ssn, :confirmation_number, :token
```
bengolder, line 3: curious if this will be blacklisted successfully by our  sensitive fields blacklist

app/forms/portal/request_client_login_form.rb
```diff
@@ -0,0 +1,15 @@
+module Portal
+  class RequestClientLoginForm < Form
+    attr_accessor :phone_number, :email_address
+
+    validates :phone_number, phone: true, allow_blank: true, format: { with: /\A\+1[0-9]{10}\z/ }
```
bengolder, line 5: arguably good to do confirm here? since typos are legion

app/controllers/portal/client_logins_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+    end
+
+    private
+
+    def ensure_valid_token
+      @raw_token = params[:token]
+      if @raw_token.present?
+        encrypted_token = Devise.token_generator.digest(Client, :login_token, @raw_token)
+        @clients = Client.where(login_token: encrypted_token)
```
bengolder, line 63: Would be nice to have a constant time version of this check

app/controllers/portal/client_logins_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+        render :edit
+      end
+    end
+
+    private
+
+    def ensure_valid_token
+      @raw_token = params[:token]
+      if @raw_token.present?
+        encrypted_token = Devise.token_generator.digest(Client, :login_token, @raw_token)
```
bengolder, line 62: Will rotating the Rails app secret invalidate all login links? It's probably fine if they do, so it's not a huge problem.

PR #700, opened 2021-01-08
Remove supported orgs and is_client_support

jenny-heath:


app/models/user.rb
```diff
@@ -59,9 +58,6 @@ class User < ApplicationRecord
   before_validation :format_phone_number
   validates :phone_number, phone: true, allow_blank: true, format: { with: /\A\+1[0-9]{10}\z/ }
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
-  has_and_belongs_to_many :supported_organizations,
```
alaroia-cfa, line 62: @jenny-heath This is great! I didn't see a migration deleting this join table (UsersVitaPartner). Does it make sense to add a migration to delete that join table?

spec/features/hub/show_client_spec.rb
```diff
@@ -72,9 +72,11 @@
   end
 
   skip "user without admin access, but is coalition lead for client organization" do
-    let(:user) { create :user, supported_organizations: [client.vita_partner, other_vita_partner] }
+    # this test may not be entirely correct
```
alaroia-cfa, line 75: It seems correct to me! Why the comment? :D

spec/lib/ability_spec.rb
```diff
@@ -104,10 +104,11 @@
     end
   end
 
-  context "a coalition lead" do
-    let(:coalition_member_organization) { create(:vita_partner) }
+  xcontext "a coalition lead" do
```
alaroia-cfa, line 107: Why skip this?

alaroia-cfa (approved):
I left some comments; others may want to review too; I'd be happy to see this land!

app/models/user.rb
jenny-heath: 🤦🏻‍♀️  definitely missed that...

spec/lib/ability_spec.rb
jenny-heath: because i didn't want to go and make the ability file work for all the new roles (this test applied to our previous "implementation" of coalition leads) - i guess i could also just delete the test instead - thoughts?

spec/features/hub/show_client_spec.rb
jenny-heath: well i don't entirely know why it was skipped in the first place and it's not like i made it pass or anything so i wanted the next person to know that this test may never have been correct to begin with?

bengolder: Makes sense!

@alaroia-cfa and I took a look and clarified the spec in this follow-up commit, fyi: c8eccadb

spec/lib/ability_spec.rb
bengolder: Seems peaceful to leave a test with some hints for what coalition lead will look like!

PR #702, opened 2021-01-08
Allow user to add/edit tax return on document, and to edit document type

sniemeyer13:


app/controllers/hub/documents_controller.rb
```diff
@@ -20,35 +20,36 @@ def show
       redirect_to transient_storage_url(@document.upload.blob)
     end
 
+    def new; end
+
     def edit; end
 
     def create
       document_params[:upload].each do |file_upload|
         Document.create!(
```
bytheway875, line 29: Does it make sense to allow for uploading multiple files with the same display name? Maybe we should limit this upload to uploading one file at a time!

app/models/document.rb
```diff
@@ -55,8 +60,12 @@ def document_type_label
...
 
   def set_display_name
-    return if display_name
+    return if display_name.present?
 
     self.display_name = upload.attachment.filename
   end
+
+  def tax_return_belongs_to_client
+    errors.add(:tax_return) unless tax_return.blank? || tax_return.client == client
```
bytheway875, line 69: Is it possible to add a specific error message so that it's clear what the problem is?

app/views/hub/documents/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<section class="slab" xmlns="http://www.w3.org/1999/html">
+  <div class="grid">
+    <%= form_with model: @document, url: url, local: true, method: method, builder: VitaMinFormBuilder do |f| %>
+
+      <h1 class="h2"><%= @title %></h1>
+      <div style="padding-bottom:2rem">
+        <%= f.label(:upload, t('general.select_files')) %>
+        <%= f.file_field :upload, multiple: "multiple", class: "attachment-upload file-input" %>
```
bytheway875, line 8: This partial is used for edit and new, but the file upload field seems like it should only exist on new. Maybe we can use a variable on the partial to make sure that the file upload field only shows when rendering new and not edit!

app/controllers/hub/documents_controller.rb
```diff
@@ -20,35 +20,36 @@ def show
...
         )
       end
       redirect_to(hub_client_documents_path(client_id: @client))
     end
 
     def update
-      @form = Hub::DocumentForm.new(@document, document_params)
-      if @form.valid?
-        @form.save
+      if @document.update(document_params)
```
bytheway875, line 42: This will pass the upload param (which is a multi select) when updating the document. Does that cause an issue? Does it clear out the existing attachment and add another? Asking because in the create method, we're having to loop through the upload value to create individual documents and here we're passing the array in as a value to upload. Seems like it could cause unexpected behavior, but maybe not!

bytheway875 (changes requested):
Unless it was specifically requested that we be able to do multiple file upload at once, I think we could simplify this code and make the UX more straightforward if we changed the multi upload to single file upload. 

I also left some comments surrounding what happens when you edit and update a document using multi-select, and re: whether we want to allow for updating the file at all.

app/controllers/hub/documents_controller.rb
sniemeyer13: Yes, we are in talks with Yvonne about that now. It's not perfect, but figured a good interim step would be to have all files named the same (unless no name given) rather than take away the multi upload functionality.

Waiting on design for how to handle this.

sniemeyer13:
@bytheway875 I spoke with Yvonne about the multi upload flow. We are going to keep the multi upload functionality, and have the interim solution of simply renaming files/editing attributes after upload. They have mocks and stories in the pipeline to clean up this functionality.

We are also going to keep the functionality to update the file in the edit view, however we will change the verbiage to 'Update file' on the edit page to make things more clear. I'll go through and make this change, as well as the others you suggested. Thanks!

app/controllers/hub/documents_controller.rb
sniemeyer13: You're right, it does fail! Removing upload functionality from the edit page.

app/views/hub/documents/_form.html.erb
sniemeyer13: Got it!

app/models/document.rb
sniemeyer13: Will do

app/controllers/hub/documents_controller.rb
sniemeyer13: After talking with Yvonne, we are not going to remove multi upload functionality - we do have a story in the works to make this process smoother though. In the mean time if you want to multi upload, you'll have to go in and edit the features later.

sniemeyer13:
@bytheway875 Made a new comit with changes - ready for review!

app/views/hub/documents/_form.html.erb
bengolder: I considered that and I think there is a good use case for allowing users to update files for specific documents: downloading/rotating/cropping is a common use case.

bytheway875 (approved):
👍

PR #703, opened 2021-01-08
Client Authentication

bengolder:
- Add devise migrations to Client

### DIFF
--- a/app/controllers/portal/client_logins_controller.rb
+++ b/app/controllers/portal/client_logins_controller.rb
@@ -1,0 +1,76 @@
+module Portal
+  class ClientLoginsController < ApplicationController
+    before_action :redirect_to_portal_if_client_authenticated
+    before_action :validate_token, only: [:edit, :update]
+    before_action :redirect_locked_clients, only: [:edit, :update]
+    layout "question"
+
+    helper_method :illustration_path
+
+    def illustration_path; end
+
+    def new
+      @form = RequestClientLoginForm.new
+    end
+
+    def link_sent; end
+
+    def create
+      @form = RequestClientLoginForm.new(request_client_login_params)
+      if @form.valid?
+        ClientLoginRequestJob.perform_later(
+          email_address: @form.email_address, phone_number: @form.phone_number
+        )
+        redirect_to login_link_sent_portal_client_logins_path
+      else
+        render :new
+      end
+    end
+
+    def invalid_token; end
+
+    def account_locked; end
+
+    def edit
+      @form = ClientLoginForm.new(possible_clients: @clients)
+    end
+
+    def update
+      @form = ClientLoginForm.new(client_login_params)
+      if @form.valid?
+        sign_in @form.client
+        redirect_to portal_root_path
+      else
+        @clients.each(&:increment_failed_attempts)
+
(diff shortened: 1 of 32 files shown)


app/controllers/portal/client_logins_controller.rb
```diff
@@ -0,0 +1,33 @@
+module Portal
+  class ClientLoginsController < ApplicationController
+    before_action :skip_login_if_authenticated
```
alaroia-cfa, line 3: Non-blocker: I suggest renaming this to `redirect_to_portal_if_client_authenticated` or something. I found it a little confusing, but anyway I get it now.

app/jobs/client_login_request_job.rb
```diff
@@ -0,0 +1,34 @@
+class ClientLoginRequestJob < ApplicationJob
+  def perform(email_address: nil, phone_number: nil)
+    raise StandardError("No contact info provided") unless email_address.present? || phone_number.present?
+
+
```
alaroia-cfa, line 5: I remember we talked about two different models we could use as a key for rate-limit data: The `Client`, or the contact point (phone number/email address). I recall that you were in favor of using the `Client` as the rate limiting key, since most of the time, it'd be 1 client, and the worst you could do is lock out another client who shares your contact info, which would be annoying, but we'd get past it. That could be fine.

As I understand it, the rate limit here would be primarily about not annoying our users, secondarily about not getting us blocked as spam from contacting specific users, and not fundamentally about authentication security.

For that situation, you can imagine a max of (say) 12 login attempts per day, past which we refuse to take action. If that's adequate, then you could imagine a table like `ClientLoginRequest` which has:

- client_id
- number
- period_start

`period_start` would be the current time rounded to the start of the day, Calling it `period_start` implies we might change the time period, but we could call it `day_start` and that would be fine, too.

You could `find_or_create` the `client_id, period_start` pair, and then increment the number, and if it's too high, refuse to take action.

This would create some database garbage at a rate of 1 record per client per day. We would probably be fine leaving this database garbage in place. If we want to clear the garbage, an opportune time to clear it is after reading each client.

alaroia-cfa: In the case that the rate limit is exceeded, what exactly do we do? I guess we can send a final message to the client saying that we rate limited them but please try back tomorrow, or email our support?

alaroia-cfa, line 5: Another approach is to skip this rate-limit now, which is fine with me, and hopefully we can come back to it.

app/models/client.rb
```diff
@@ -26,6 +26,8 @@
 #  fk_rails_...  (vita_partner_id => vita_partners.id)
 #
 class Client < ApplicationRecord
+  devise :lockable, :timeoutable, :trackable
```
alaroia-cfa, line 29: On the secret number page, let's lock after 5 attempts, with no unlock policy (since Devise an auto-unlock by time or by sending an email confirmation link; let's do neither).

I think there a method to `unlock` a `lockable`, but I don't _really_ know.

If they're locked, we should a message saying they should contact support. Support should file a chore for eng to contact the client, find out if it's really them, and then ask for eng to unlock the account.

PR #704, opened 2021-01-11
Primary Taxpayer can sign an 8879 for a tax return [#176046722]

bytheway875:
NOTE: This does not include the client-side interface for signing the doc OR the logic for saving the signature doc to AWS -- those are additional tasks.

* The 8879 forms we get when uploading documents from TaxSlayer (and the ones directly from the IRS) do NOT make the signature lines fillable fields. This means that we can't use the same PDF filling logic/gem that we use for fillable fields.
* The combine_pdf gem allows us to write text into the pdf at a specific location, so I've created some wrappers around combine_pdf that allow us to write onto the 8879 signature lines.

Screen Shot 2021-01-11 at 10 25 30 AM
Screen Shot 2021-01-11 at 10 26 11 AM


app/lib/document_types/completed_form8879.rb
```diff
@@ -0,0 +1,17 @@
+module DocumentTypes
+  class CompletedForm8879 < DocumentType
```
bytheway875, line 2: Confirming w/ Yvonne whether this is the appropriate doc type

PR #705, opened 2021-01-12
Fix nil document_type error, refactor Hub::DocumentsController#create, and remove intake setting on hub uploaded documents

bytheway875:


app/controllers/hub/documents_controller.rb
```diff
@@ -25,17 +25,8 @@ def new; end
     def edit; end
 
     def create
-      document_params[:upload].each do |file_upload|
-        Document.create!(
-          uploaded_by: current_user,
-          client: @client,
-          intake: @client.intake,
-          document_type: document_params[:document_type],
```
bytheway875, line 33: When we set the params directly like this, when document_params[:document_type] is nil, it explicitly sets document_type to nil. When a nil value is provided, the default value from the database cannot be applied and the transaction fails the field required validation. By refactoring to apply only the params that are passed, we can ensure that a nil value does not get set.

sniemeyer13 (approved):
👍🏻

bengolder (commented):
Looks great! Thanks for the fix!

bengolder:
Oops, too late on the review

PR #706, opened 2021-01-12
Fix layout of help text during account creation [#175978650]

alaroia-cfa:
Before:

image

After:

image


alaroia-cfa:
Reviewed with @bengolder

PR #708, opened 2021-01-13
Primary taxpayer can sign an 8879 form for a tax return.

bytheway875:
Screen Shot 2021-01-13 at 11 13 42 AM
Screen Shot 2021-01-13 at 11 14 02 AM


app/views/tax_returns/success.html.erb
```diff
@@ -0,0 +1 @@
+<h1>Thank you for filing your taxes with us!</h1>
```
bytheway875, line 1: This is MVP for testing purposes only, so we didn't translate it or style it.

app/forms/portal/sign_form8879.rb
```diff
@@ -0,0 +1,59 @@
...
+  class SignForm8879 < Form
+    attr_accessor :primary_accepts_terms, :primary_confirms_identity, :ip, :transaction_failed
+
+    def initialize(tax_return, params = {})
+      @tax_return = tax_return
+      super(params)
+    end
+
+    def sign
+      errors.add(:primary_accepts_terms, :blank) && return unless primary_accepts_terms == "yes"
```
bytheway875, line 11: form.valid? pattern instead.

app/forms/portal/sign_form8879.rb
```diff
@@ -0,0 +1,59 @@
...
+      errors.add(:primary_confirms_identity, :blank) && return unless primary_confirms_identity == "yes"
+      return true if ActiveRecord::Base.transaction do
+        record_primary_signer_information_on_tax_return
+        change_tax_return_status
+        @tax_return.save!
+        create_signed_8879
+        true
+      end
+
+      errors.add(:transaction_failed)
```
bytheway875, line 21: how do we get away with not returning a falsey value from this method and having it work?

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
+class TaxReturnsController < ApplicationController
+  # Because these aren't CRUD verb methods, they need to be explicitly listed on the before actions
```
jenny-heath, line 2: nit: i think we can remove this comment

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  before_action :block_access_on_production, only: [:authorize_signature, :sign, :success]
+  before_action :load_tax_return, except: [:success]
+  before_action :check_for_forms, only: [:sign, :authorize_signature]
+
+  def authorize_signature
+    tax_return = TaxReturn.find(params[:tax_return_id])
```
jenny-heath, line 8: would it be a good idea to do something if we can't find the tax return? genuinely not sure 🤔

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    @form = Portal::SignForm8879.new(tax_return)
+  end
+
+  def sign
+    @form = Portal::SignForm8879.new(@tax_return, permitted_params)
+    if @form.valid? && @form.sign
+      redirect_to tax_return_success_path(params[:tax_return_id])
+    else
+      flash.now[:alert] = I18n.t("controllers.tax_returns_controller.errors.#{@form.errors.keys.first}")
```
jenny-heath, line 17: should this be the first error message or all? i think there's a `.full_messages` or some such but idk whether it will string them together in a nice way

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      render :authorize_signature
+    end
+  end
+
+  def success; end
+
+  private
+
+  def load_tax_return
+    @tax_return = TaxReturn.find(params[:tax_return_id])
```
jenny-heath, line 27: oh lol i think this is redundant with line 8

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def permitted_params
```
jenny-heath, line 30: i think the naming convention we currently have is `"#{model name}_params"` but it doesn't super matter

app/controllers/tax_returns_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    params
+      .require(Portal::SignForm8879.form_param)
```
jenny-heath, line 32: neat! i didn't know about this

app/forms/portal/sign_form8879.rb
```diff
@@ -0,0 +1,69 @@
...
+    validate :terms_accepted
+    validate :identity_confirmed
+
+    def initialize(tax_return, params = {})
+      @tax_return = tax_return
+      super(params)
+    end
+
+    def sign
+      return false unless valid?
```
jenny-heath, line 13: just to confirm, we are checking that it is valid once in the controller and once here? i think we can probably skip this check unless it's the case in our other forms that we check validity at the beginning of what is usually the `save` method

app/controllers/tax_returns_controller.rb
bytheway875: 👍

app/controllers/tax_returns_controller.rb
bytheway875: I think it's fine to just show one flash error for both checkboxes at the top.

The way this is set up, it will put error messages on both check box fields when it validates, but it will add a more detailed error as a flash alert.

app/controllers/tax_returns_controller.rb
bytheway875: This is a weird case because they're not actually tax_return_params -- they're virtual params for the form object, so tax_return_params would be a misnomer. For that reason, I fell back on using the rails convention of calling the method permitted_params

app/controllers/tax_returns_controller.rb
bytheway875: Yes! I added this a while back to make working with form objects a bit easier.

PR #709, opened 2021-01-13
Add 2020 backtax option

embarnard:


app/models/intake.rb
```diff
@@ -453,6 +453,7 @@ def filing_years
   end
 
   def most_recent_filing_year
+    # TODO: should the default be changed to 2020?
```
embarnard, line 456: Question for reviewer

bytheway875: yes! I think this should be updated to 2020!

app/views/questions/life_situations/edit.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :page_title, t('views.questions.life_situations.title') %>
+<% content_for :page_title, t('views.questions.life_situations.title', year: current_intake.most_recent_filing_year ) %>
```
bytheway875, line 1: Awesome! I'm glad you were able to incorporate this change.

spec/services/mixpanel_service_spec.rb
```diff
@@ -280,14 +283,14 @@
 
           it "includes stimulus triage data" do
             expect(data_from_intake).to include(
```
bytheway875, line 285: yay! I enjoy the re-indentation.

bytheway875 (approved):
Looks good to me! The only outstanding question I have is -- are we continuing to do 2017 this year? Or do we need to remove 2017 once we add 2020 because VITA only preps x years each year?

embarnard:
> Looks good to me! The only outstanding question I have is -- are we continuing to do 2017 this year? Or do we need to remove 2017 once we add 2020 because VITA only preps x years each year?

@bytheway875 That's a good question! I checked with Yvonne and it looks like 2017 should be an option until April 15th. Thanks for the review!

PR #711, opened 2021-01-14
Allow coalition leads to see clients in their coalition and edit their data

jenny-heath:


app/models/vita_partner.rb
```diff
@@ -33,6 +33,7 @@ class VitaPartner < ApplicationRecord
   has_many :source_parameters
   has_many :users
   belongs_to :parent_organization, class_name: "VitaPartner", optional: true
+  # what is the difference between this and child_sites?
```
jenny-heath, line 36: doesn't really belong in this PR but i was curious

spec/lib/ability_spec.rb
```diff
@@ -85,6 +85,32 @@
     end
   end
 
+  context "a site coordinator" do
```
jenny-heath, line 88: this just got moved, it's the same

spec/lib/ability_spec.rb
```diff
@@ -85,6 +85,32 @@
...
+    end
+
+    it "can manage data from clients in their site" do
+      expect(subject.can?(:manage, accessible_client)).to eq true
+    end
+
+    it "cannot manage data from clients in other sites" do
+      expect(subject.can?(:manage, other_vita_partner_client)).to eq false
+    end
+  end
```
jenny-heath, line 112: this just got moved, it's the same

spec/models/user_spec.rb
```diff
@@ -111,28 +111,44 @@
...
+      let!(:user) { create :site_coordinator_user }
+      let!(:unaccessible_site) { create :site }
+
+      it "should return the user's site" do
+        accessible_group_ids = user.accessible_groups.pluck(:id)
+        expect(accessible_group_ids).to include(user.role.site.id)
+        expect(accessible_group_ids).not_to include(unaccessible_site.id)
+      end
+    end
+
```
jenny-heath, line 124: also moved

app/models/user.rb
```diff
@@ -76,6 +76,13 @@ def accessible_groups
       )
     elsif role_type == TeamMemberRole::TYPE || role_type == SiteCoordinatorRole::TYPE
       VitaPartner.sites.where(id: role.site.id)
+    elsif role_type == CoalitionLeadRole::TYPE
+      coalition = Coalition.find(role.coalition.id)
+      organizations = coalition.organizations
+      sites = organizations.reduce([]) do |sites, organization|
+        [*sites, *organization.child_sites]
+      end
+      [coalition, *organizations, *sites]
```
jenny-heath, line 85: it seems to me there are approx a million ways to collect these things - is this the best? i don't know. i tried to think of a way that involved fewer queries but i could not

jenny-heath: which is not to say that i think this is a lot of queries, only that if there is a more optimal path i could not think of it

embarnard: this wouldn't affect the time to query but could also use a case statement

embarnard (approved):
LGTM! No blocking feedback ✅

PR #712, opened 2021-01-14
Upgrade config to use defaults from Rails 6.0

alaroia-cfa:


bengolder (approved):
looks good

PR #713, opened 2021-01-14
Fix bin/setup

alaroia-cfa:
* Depend on postgis directly, to avoid postgresql version skew issues

* Fix buggy invocation of `tr`

* Move Aptible install into Brewfile

* Skip rbenv if already installed

* Upgrade Bundler to avoid a warning


bengolder (approved):
great fixes!

PR #717, opened 2021-01-14
Rename file for new autoloader

alaroia-cfa:
Just a file rename since the class name doesn't match the filename.


alaroia-cfa:
I'm going to self-merge this since it's trivial. cc: @bengolder - this is one we missed. If you can leave a comment saying you approve, I'd appreciate it.

PR #718, opened 2021-01-14
Run tests in parallel

alaroia-cfa:
## Before

Tests took about 3.5 minutes to run

## After

Tests take about 1-2 minutes to run. (I've seen them **as fast as 1 minute 6 seconds** and as slow as 1 minute 50 seconds).

## How to use this

Once this is merged, run `bin/setup` to create your many test databases (or run `bin/rails parallel:create` then `bin/rails parallel:setup`). 

If you have lots of annoying warnings from the macOS firewall, you can enable "Block all incoming connections" in the macOS firewall:

image

## Additional changes

* `bin/setup`: Add parallel test setup.
* `bin/test`: When Ruby tests fail, do not run JS tests; this makes the output cleaner.
* Adjust some classes to avoid autoloader warnings
* Before the test suite, use `database_cleaner` to empty the database
* Enable `config.cache_classes` for full test runs, i.e., `bin/test` and Circle-CI. This is a performance trade-off similar to eagerly loading the app.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
       - run: bundle exec rake db:schema:load
       - run: mkdir ~/test-results && mkdir ~/test-results/rspec && mkdir ~/test-results/jest
       - run: |
-          bundle exec rspec \
+          RAILS_CACHE_CLASSES=1 bundle exec rspec \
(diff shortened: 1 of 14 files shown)


bengolder:
block incoming connections

PR #719, opened 2021-01-15
Put i18n specs in their own files

alaroia-cfa:
This allows more parallelism when running the tests, since
the tests are parallelized by file.

Also upgrade a caching tool we use, bootsnap, and warm its cache
before starting the tests.

This brings test duration to < 1 min on my laptop.


bengolder (approved):
Great work! Thank you!

PR #723, opened 2021-01-19
Add final tax documents and links #176063241

bytheway875:
Screen Shot 2021-01-18 at 8 27 30 PM


app/controllers/portal/documents_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Portal
```
sniemeyer13, line 1: Liking the namespace

app/controllers/portal/tax_returns_controller.rb
```diff
@@ -42,9 +42,8 @@ def success; end
     private
 
     def load_client_tax_return
-      @tax_return = current_client.tax_returns.includes(client: [:intake]).find(params[:tax_return_id])
-    rescue ActiveRecord::RecordNotFound
-      redirect_to :root
+      @tax_return = current_client.tax_returns.includes(client: [:intake]).find_by(id: params[:tax_return_id])
```
bytheway875, line 45: This was an update requested by @alaroia-cfa on the previous PR. Instead of capturing the error raised by a #find, we use find_by to return nil, and then redirect if the value is nil.

app/controllers/portal/tax_returns_controller.rb
```diff
@@ -57,7 +56,7 @@ def permitted_params(form_class)
     def redirect_unless_signature_required(signature_type)
```
bytheway875, line 56: In cases where the client is authenticated but they aren't able to take the action, we want to redirect to their portal dashboard instead of the homepage -- this was a non-blocking request by @alaroia-cfa on prev PR

app/models/tax_return.rb
```diff
@@ -71,15 +71,23 @@ def filing_joint?
   def ready_for_signature?(signature_type)
     return false if signature_type == TaxReturn::PRIMARY_SIGNATURE && primary_has_signed?
     return false if signature_type == TaxReturn::SPOUSE_SIGNATURE && (spouse_has_signed? || !filing_joint?)
-    return false if documents.find_by(document_type: DocumentTypes::CompletedForm8879.key).present?
+    return false if signed_8879.present?
```
sniemeyer13, line 74: I like this signed/unsinged document name - much clearer

sniemeyer13 (approved):
This looks great!

bytheway875:
@sniemeyer13 Thanks, Sarah!

PR #727, opened 2021-01-20
bin/setup: Exit with error msg if parallel test databases do not exist

alaroia-cfa:


alaroia-cfa:
@jenny-heath this should provide a useful warning message for the environment problem you had!

PR #728, opened 2021-01-20
updated the support email address on the privacy policy and footer [#…

benjaminleesmith:
…176507252]


bytheway875:
I think that there are a bunch of other references to hello@getyourrefund.org as a contact email in the application, but I'm pretty sure that the ticket wasn't comprehensive enough because we've got contact us links smattered throughout the app that also use hello@ 

For some recent context i got from Kelly when working on some other email stuff, we are now going to be SENDING emails from within the app from hello@getyourrefund.org for dealing with clients that we're already working with, but we'll be forwarding all emails to support@getyourrefund.org into Intercom to be triaged by our service staff.

I'll comment same in Pivotal, and tag Kelly for her to weight in.

PR #732, opened 2021-01-21
Route by referring org source param

bytheway875:


app/controllers/questions/personal_info_controller.rb
```diff
@@ -8,7 +8,14 @@ def tracking_data
 
     def after_update_success
       unless Client.after_consent.where(intake: current_intake).exists?
-        ClientRouter.route(current_intake.client)
+        routing_service = OrganizationRoutingService.new(
```
bytheway875, line 11: Change to RoutingService

bytheway875: Done!

PR #733, opened 2021-01-21
[WIP] Route client by zip code

embarnard:


app/lib/vita_partner_zip_code_routing_importer.rb
```diff
@@ -0,0 +1,30 @@
...
+      if vita_partner.nil?
+        puts "Unable to find VitaPartner with name '#{organization_name}'"
+        next
+      end
+
+      zip_codes.to_s.split(",").map do |zip_code|
+        begin
+          VitaPartnerZipCode.create!(zip_code: zip_code, vita_partner: vita_partner)
+          puts "Created VitaPartnerZipCode with #{zip_code} for #{vita_partner&.name}"
+        rescue => e
```
bytheway875, line 24: Yay! Nice use of error handling.

app/models/vita_partner.rb
```diff
@@ -31,6 +31,7 @@ class VitaPartner < ApplicationRecord
   has_many :intakes
   has_many :source_parameters
   has_many :users
+  has_many :vita_partner_zip_codes
```
bytheway875, line 34: i wonder if we should change this to has_many :serviced_zip_codes, class_name: "VitaPartnerZipCode" so that it's a little more readable.

app/models/vita_partner_zip_code.rb
```diff
@@ -0,0 +1,30 @@
...
+#  index_vita_partner_zip_codes_on_zip_code_and_vita_partner_id  (zip_code,vita_partner_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (vita_partner_id => vita_partners.id)
+#
+class VitaPartnerZipCode < ApplicationRecord
+  belongs_to :vita_partner
+  validate :record_of_zip_code
+  validates :zip_code, uniqueness: { scope: :vita_partner }
```
bytheway875, line 23: Awesome, so this means that multiple vita partners can service the same zip code but that each VP can only have one entry for each zip code, right? That's awesome.

app/services/partner_routing_service.rb
```diff
@@ -28,8 +29,15 @@ def route_from_source_param
     end
   end
 
-  def vita_partner_from_zip_code
-    # @routing_method = :zip_code
+  def route_from_zip_code
+    return false unless @zip_code.present?
+
+    vita_partner = VitaPartnerZipCode.where(zip_code: @zip_code).first&.vita_partner
```
bytheway875, line 35: Right now, this will always assign to the _first_ vita partner based on ID if there are multiple VPs with a particular zip code. Seems like we'll need to make this more sophisticated over time, but it might be okay to start with. 

IF we're already dealing with overlapping zip codes, we def should handle this before merging. We could do so easily by doing the where, and then randomly selecting one of the VPs from the list and assigning to them. Then, once we've got some capacity numbers we can start to assign based on relative capacity, too.

spec/features/web_intake/routing_spec.rb
```diff
@@ -39,7 +41,41 @@
     # ---
```
bytheway875, line 41: Let's refactor this spec so that we're making sure we're actually routing based on param, and then add another scenario that ensures we're routing based on zip code! Right now, the zip code + source param apply to the same vita_partner, so we can't tell whether our logic is "falling through" in the right way.

app/models/vita_partner.rb
embarnard: Great idea!

app/models/vita_partner_zip_code.rb
embarnard: Yes exactly! I added another test in the scenario where there is a dupe zip code record with different VP to make this clearer

embarnard: Note after talking Kelly & Jenny: should not be more than zip code for different VP

PR #734, opened 2021-01-22
Rename classes to avoid conflict

alaroia-cfa:
`CreateVitaPartnerZipCodes` used to be both a regular class and a migration.
Rename one to `VitaPartnerZipCodesImporter`.

I believe this only showed up when doing a deploy because we "eagerly load" all the code when doing a deployment.

(Unrelated: I've been interested in making the test suite eagerly load the code. That would have caught this issue. I may do that sometime.)


bengolder (approved):
Looks good!

PR #735, opened 2021-01-22
Source param routing updates #176262947

bytheway875:
Two requested updates to the client param routing stuff:

1) Allow any version of /Hub /HUB /hub /hUb to match the partner's source param. We downcase the code before adding it as a source parameter, and then downcase what we save as the source param when we search for it in the Partner Router. (I didn't transform it when putting it into the db, because that source param seems to be used for a couple of different values...)

2) Make the section that displays this info more readable/understandable to users.

Screen Shot 2021-01-22 at 9 42 35 AM

### DIFF
--- a/app/models/source_parameter.rb
+++ b/app/models/source_parameter.rb
@@ -23,4 +23,12 @@ class SourceParameter < ApplicationRecord
   validates_presence_of :vita_partner_id
   validates_presence_of :code
   validates_uniqueness_of :code
+
+  before_validation :downcase_code
+
+  private
+
+  def downcase_code
+    self.code = code.downcase if code_changed?
+  end
 end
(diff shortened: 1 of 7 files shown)


app/models/source_parameter.rb
```diff
@@ -23,4 +23,12 @@ class SourceParameter < ApplicationRecord
   validates_presence_of :vita_partner_id
   validates_presence_of :code
   validates_uniqueness_of :code
+
+  before_validation :downcase_code
+
+  private
+
+  def downcase_code
+    self.code = code.downcase if code_changed?
```
bengolder, line 32: why is this conditional on a change? Are there scenarios when you don't want it to downcase?

bengolder (approved):
Looks good! One minor question

app/models/source_parameter.rb
bytheway875: code_changed? is part of active model dirty -- it will only reassign the downcased code programmatically if the code is one of the values that changed. This prevents us from rewriting the downcased version of code on every write (for example, say we wanted to increment a "count" attribute every time we see a source param used -- we don't want to re-downcase the code every time we do that)

PR #741, opened 2021-01-25
Fix eip intake creation

sniemeyer13:
In the course of requiring visitor_ids onto intakes to facilitate mixpanel updates, we realized that this form/controller was requiring us to explicitly set visitor_id in _two_ places. We realized that the controller after_update_success method was actually creating a _second_ new intake! I think it was expecting to update the current_intake from the application_controller (which would have been updated when setting the intake_id into session), but because this controller has it's own current_intake method that instantiates a new Intake, we were creating a second new intake. For every eip intake we created, we were then creating a second blank one that only stored triage source!

### DIFF
--- a/app/controllers/questions/eip_overview_controller.rb
+++ b/app/controllers/questions/eip_overview_controller.rb
@@ -21,7 +21,7 @@ module Questions
       session[:intake_id] = @form.intake.id
       stimulus_triage_id = session.delete(:stimulus_triage_id)
       if stimulus_triage_id.present?
-        current_intake.update(triage_source: StimulusTriage.find(stimulus_triage_id), visitor_id: visitor_id)
+        @form.intake.update(triage_source: StimulusTriage.find(stimulus_triage_id))
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/questions/eip_overview_controller.rb
```diff
@@ -21,7 +21,7 @@ def after_update_success
       session[:intake_id] = @form.intake.id
       stimulus_triage_id = session.delete(:stimulus_triage_id)
       if stimulus_triage_id.present?
-        current_intake.update(triage_source: StimulusTriage.find(stimulus_triage_id), visitor_id: visitor_id)
+        @form.intake.update(triage_source: StimulusTriage.find(stimulus_triage_id))
```
jenny-heath, line 24: another thing that could be done to fix this is the `current_intake` method on this controller could return `super || Intake.new(eip_only: true)` which would also take care of people who arrived on this page with an intake already in their session? unclear whether this is what we want for people arriving on the page or not (unclear duplicate behavior etc etc)

bytheway875: It seems like the logic was intended to always create a new intake for EIP only, even if they already had another intake. So super || would fix the after_update issue, but would cause a bug on the before update. I think this fix is the "least tricky" way to ensure that we're actually doing what we want, which is updating the immediately-previously-created intake with this new data.

PR #742, opened 2021-01-25
change intake guard to look at consented at vs completed

benjaminleesmith:
Signed-off-by: Asheesh Laroia <alaroia@codeforamerica.org>

### DIFF
--- a/app/helpers/duplicate_intake_guard.rb
+++ b/app/helpers/duplicate_intake_guard.rb
@@ -2,7 +2,7 @@ class DuplicateIntakeGuard < SimpleDelegator
   def has_duplicate?
     permitted_eip_only_values = eip_only ? true : [nil, false]
     Intake
-      .where.not(completed_at: nil)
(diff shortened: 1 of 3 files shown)


app/helpers/duplicate_intake_guard.rb
```diff
@@ -2,7 +2,7 @@ class DuplicateIntakeGuard < SimpleDelegator
   def has_duplicate?
     permitted_eip_only_values = eip_only ? true : [nil, false]
     Intake
-      .where.not(completed_at: nil)
```
bytheway875, line 5: It seems so odd to be looking this up by a "string" value, but I guess it maps to an enum and thus an int in the db... It's definitely easier to read than where.not(primary_consented_to_service_at: nil)

bytheway875 (approved):
Based on how we have the data set up, this looks as accurate as possible to me.

alaroia-cfa:
Thanks @bytheway875 !

PR #743, opened 2021-01-25
Touch client on tax return updates [#176600406]

bytheway875:
When you add "touch" to a belongs_to association, updated_at on the associated record will be updated anytime you update the child. Here, any time we update the tax return we will update updated_at on the client. 

This covers the two cases outlined in the ticket (Assigning a user to a tax return and changing the status), as well as the case when you update the certification status.


alaroia-cfa (approved):
Splendid! Excited to see this get merged.

PR #748, opened 2021-01-26
Eager load some associations for client index pages

bengolder:
For client index pages, this eager loads the intakes, vita partners, tax returns, and users assigned to tax returns, to reduce the proliferation of unecessary queries

### DIFF
--- a/app/controllers/hub/assigned_clients_controller.rb
+++ b/app/controllers/hub/assigned_clients_controller.rb
@@ -10,7 +10,7 @@ module Hub
 
     def index
       @page_title = I18n.t("hub.assigned_clients.index.title")
-      @clients = filtered_and_sorted_clients
+      @clients = filtered_and_sorted_clients.with_eager_loaded_associations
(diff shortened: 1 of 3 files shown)


app/controllers/hub/assigned_clients_controller.rb
```diff
@@ -10,7 +10,7 @@ class AssignedClientsController < ApplicationController
 
     def index
       @page_title = I18n.t("hub.assigned_clients.index.title")
-      @clients = filtered_and_sorted_clients
+      @clients = filtered_and_sorted_clients.with_eager_loaded_associations
```
alaroia-cfa, line 13: Does it make sense to call `.with_eager_loaded_associations` in `filtered_and_sorted_clients`?

alaroia-cfa (approved):
This looks peaceful and perhaps excellent.

I'd love some kind of stats on this, perhaps before/after page load time or # of queries. Not a blocker!

PR #749, opened 2021-01-27
Current intake for authenticated clients should be their client intake

bytheway875:


spec/controllers/portal/portal_controller_spec.rb
```diff
@@ -1,6 +1,28 @@
...
+      before do
+        sign_in client, scope: :client
+      end
+      it "sets the current intake to that clients intake" do
+        expect(subject.current_intake).to eq client.intake
+      end
+    end
+    context "when the client is not authenticated" do
+      it "uses the intake_id from the session to determine the current_intake (like in unauthenticated intake flow" do
+        expect(subject.current_intake).to eq session_intake
```
alaroia-cfa, line 22: It'd be fine with me if `current_intake` were defined as nil in this controller until you're logged in, since it's not really relevant. An option for aggressive cleanup. Shrug; not a blocker.

app/controllers/application_controller.rb
```diff
@@ -59,6 +59,8 @@ def set_visitor_id
       visitor_id = SecureRandom.hex(26)
       cookies.permanent[:visitor_id] = { value: visitor_id, httponly: true }
     end
+    # If we run into cases where the intake does not have an associated visitor_id persisted onto it,
```
alaroia-cfa, line 62: Consider deleting the `visitor_record` method and calling `current_intake` here instead, since `visitor_record` is just a wrapper for `current_intake`.

If the record is persisted, then it'll always have a `visitor_id`, so consider removing this case.

If the record is not persisted yet, then it's good to tag the `current_intake` with the `visitor_id` from the cookie. I presume that's handled elsewhere.

None of the above is a blocker.

alaroia-cfa (approved):
I approve! I propose some minor simplifications.

app/controllers/portal/portal_controller.rb
```diff
@@ -11,6 +11,10 @@ def home
       @tax_returns = current_client.tax_returns.order(year: :desc)
     end
 
+    def current_intake
```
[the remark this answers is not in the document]
bytheway875: Hmm... yeah, I'd explictly set up the test to have it fall through so ended up having to call to super to make that work, but you're right -- if you are on these pages somehow, it should never use the session[:intake_id] to get intake and it should be nil. Will update.

app/controllers/application_controller.rb
bytheway875: yeah, i think that might clear some things up. visitor_record gets called from some other methods, but if its easy to transition into current_intake i will do that.

spec/controllers/portal/portal_controller_spec.rb
bytheway875: Updated this.

app/controllers/application_controller.rb
bytheway875: unfortunately, I just checked on this and visitor_record is overridden with some other "intake types" in other controllers -- for example, current_diy_intake, which I guess is stored distinctly from current_intake. So I think for now, leaving visitor_record is a necessity.

PR #753, opened 2021-01-27
Show unlinked clients for admins [#176039635]

alaroia-cfa:
## New interface

See **Unlinked clients** tab

image

Click it and you get

image

Click the phone number and you get

image


app/controllers/hub/unlinked_clients_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      @page_title = I18n.t("hub.clients.unlinked_clients.title")
+      @sort_column = "updated_at"
+      @sort_order = params[:order] == "desc" ? "desc" : "asc"
+      @clients = Client.where(vita_partner: @vita_partner).order(created_at: @sort_order).with_eager_loaded_associations
+    end
+
+    private
+
+    def load_and_authorize_national_org
+      @vita_partner = VitaPartner.find_or_create_national_org
```
bytheway875, line 20: This view shows all of the clients that are assigned to the national organization, but those national organizations can also be assigned "valid" clients that they're needing to reassign to things. Seems like it would make more sense to _not_ assign those incoming orphans to any vita_partner and then have this view only display truly "orphaned" clients that have no vita_partner assigned at all. This might have been considered!

bytheway875: I'm wondering if there are some different approaches we can take to the orphans.

- Do not assign these orphan convo clients to any vita partner
- Create a client, but do not create an intake for these new clients. Then we can target orphans by people who have no intake. (This will solve this problem, but possibly create problems down the line since we rely hardcore on every client having an intake.) 
- Do not create client/intake objects for these incoming messages at all. Instead, create a "Conversation" model that can have incoming/outgoing messages attached to non-clients. Client support can go back and forth in these interactions without a client object. Over time, we can build out logic that allows us to "convert" these conversations into client objects that save the conversation history.

alaroia-cfa:
I appreciate the thoughtful review! I'll reply in bulk rather than inline.

## Discussion

I think removing `find_or_create_` from the method name is a good idea!

You mentioned the possibility of not linking these Clients to _any_ VitaPartner. It's an interesting idea. I mentioned this to Yvonne, and she wondered if the absence of a link to a VitaPartner would make these records get confused with other records. I'm worried about this risk of confusion, so I like the idea of linking them to a VitaPartner.

You mentioned the possibility of not creating Clients at all, which is also an interesting idea. I checked with Yvonne, and she said that they really want the ability to create internal notes linked to these messages, so I think the path of least resistance is to make a Client. Yvonne's primary goal is to eventually be able to merge these records with other real Client records. She's OK with whatever so long as she can leave internal notes and do that kind of merging. The simplest implementation approach from my perspective is to make a Client, but I agree that it's fundamentally inaccurate. My preference is to land this and not worry about it too much, but I'm open to rethinking it if it's important.

You mentioned the possibility of not creating Intakes, which has its appeal, but that's where we put the phone number and email address, so if we're creating Clients, then I want to keep creating Intakes for now.

You also mentioned a concern that national_overflow_locations can be assigned valid clients. I think that concern arose because we use `national` to refer to this "national org" (unlinked clients) and also to refer to national overflow locations. That is very confusing! I think this is addressed if I can rename the `find_or_create_national_org` method to something else.

## Plan

I plan to rename `VitaPartner.find_or_create_national_org` to `VitaPartner.unlinked_clients_org`, which isn't one of the options you suggested, but maybe embodies the real purpose of this org. (Yvonne says we can call this org `customer_support`, or `national`, or whatever, so we can take whatever liberties make sense to us with the name.)

Happy to discuss further!

bytheway875:
👍 That all makes sense to me -- I hadn't considered the issues with not making an intake because we store the information on the intake. We'd talked about using this tax season to guide us on what should be stored on intake and what should be stored on client, and this seems like particularly strong evidence than contact information makes sense on a client object. I have a _slight_ preference for calling the org `client_support_org` because "unlinked" implies they aren't linked to anything but they are 😂

alaroia-cfa:
@bytheway875 Good thinking. I'll go with `client_support_org`. I'll merge this once it's green. Thanks!

PR #754, opened 2021-01-28
Add resend invite link to the user edit page

sniemeyer13:


spec/features/hub/edit_user_spec.rb
```diff
@@ -31,6 +31,19 @@
...
+        visit edit_hub_user_path(id: user_to_edit.id)
+
+        click_on "Resend invitation email"
+
+        within(".flash--notice") do
+          expect(page).to have_text "Invitation re-sent to #{user_to_edit.email}"
+        end
+
+        invited_user = User.where(invited_by: current_user).last
+        expect(invited_user.invitation_token).to be_present
```
bengolder, line 45: Would it make sense to check here that the invitation email went out, like we do in the invitations feature spec?

app/controllers/hub/users_controller.rb
```diff
@@ -28,6 +28,14 @@ def update
       end
     end
 
+    def resend_invitation
+      user = User.find_by(id: params[:user_id])
```
bengolder, line 32: I'd recommend narrowing this down to a subset of users that should be accessible to the current user, and using `find(params[:id])` to raise a not found error if the user id is not within the set of accessible users.

bengolder: What are the access expectations for this action? I'm not sure who can currently see the user edit page, aside from admins.

bytheway875: There is also currently a resend user invitation endpoint in users/invitations controller that does very similar logic -- I think we should consolidate to using one or the other!

spec/features/hub/edit_user_spec.rb
sniemeyer13: Yes, great idea!

app/controllers/hub/users_controller.rb
sniemeyer13: Let me double check.. If that is the case, I'm guessing you're correct in that it's only admins.

app/controllers/hub/users_controller.rb
```diff
@@ -28,6 +28,18 @@ def update
...
+      ability = Ability.new(current_user)
+
+      if ability.can?(:manage, user)
```
bengolder, line 35: `current_ability` is a helper that is added to ApplicationController, and is the same as `Ability.new(current_user)`

bengolder (approved):
Approved with a request for one minor change: use `current_ability` in place of `Ability.new(current_user)`

app/controllers/hub/users_controller.rb
sniemeyer13: Neat!

PR #757, opened 2021-01-28
Remove diy from offboarding pages #176469036

sniemeyer13:


bengolder (approved):
Has some code that does not appear related, and also a merge conflict (possible due to extra code). Assuming those two things will be dealt with, approved!

sniemeyer13:
I've just cherry-picked the commit onto main so that we aren't in this funky state.

PR #759, opened 2021-01-28
Create client support org explicitly [#176039635]

alaroia-cfa:
This removes the call to `.upsert()` in `VitaPartner`. I discovered the `upsert` was always creating a new client support org every time it was called. This avoids the problems by creating the client support organization explicitly in:

* `rails_helper.rb` for the tests
* `seed.rb` for local dev

For demo & prod, I'll manually ensure we have exactly 1 matching org.


alaroia-cfa:
@bytheway875 says this is good!

PR #760, opened 2021-01-28
For autocreated clients, opt them in to comms [#176039635]

alaroia-cfa:


alaroia-cfa:
@bytheway875 approved!

PR #763, opened 2021-01-29
Show sign-in on production

alaroia-cfa:
**Before** this change, the "Volunteer sign in" link was missing on production.

**After** this change, we show it.


sniemeyer13 (approved):
Wahoo!

PR #765, opened 2021-01-30
add SSN/ITIN to client edit and show pages

jenny-heath:


app/views/hub/clients/_primary_info_fields.html.erb
```diff
@@ -27,7 +27,13 @@
     <%= f.cfa_checkbox(:sms_notification_opt_in, t("general.opt_in_sms"), options: { checked_value: "yes", unchecked_value: "no" }) %>
   </div>
   <%= f.cfa_input_field(:phone_number, t("general.phone_number"), classes: ["form-width--long"]) %>
-
+  <%= f.cfa_input_field(
+          :primary_last_four_ssn,
+          t("general.last_four_ssn"),
+          prefix: "XXX-XX-",
+          type: :tel,
```
bytheway875, line 34: oh! clever use of tel to create an all number input.

bytheway875 (approved):
woo hoo! seems like _maybe_ could use a feature spec, but I like how this builds off of the existing pattern and extends the AccessLog class in a simple way.

jenny-heath:
@bytheway875 i added to the `edit_client_spec.rb` feature spec for the viewing the hidden field part but there's no feature spec for editing because i figured that was covered by the controller spec. but maybe i misunderstood what kind of feature spec you're referring to? thanks for reviewing!

PR #767, opened 2021-01-30
Don't show org edit links to coalition leads

bengolder:
This fixes a bug that I forgot to close in a recent feature. 

We shouldn't show org edit links to coalition leads on the org index because they are not allowed to access the org edit page. Instead, they should be able to click into org show pages.

Feel free to merge or edit while I'm out on vacation!


alaroia-cfa (approved):
Splendid! Happy vacation to you!

PR #769, opened 2021-01-31
Edit site unique URLs [#176419201]

alaroia-cfa:
Allow editing a site's unique URLs.

image

I made a few unusual choices worth explaining:

* We don't show the unique URLs for a site when _creating_ it, but we do when _editing_ it. This is consistent with what we do for orgs as of #764.
* I changed the site edit page so it redirects back to itself on success, rather than to the org edit page on success. This is consistent with what we do for orgs in #764.
* I bundled #766 into this PR because I saw it there.

I'm hoping to use this on prod on Monday morning to roll out some org/site unique URL changes.

### DIFF
--- a/app/views/hub/sites/_form.html.erb
+++ b/app/views/hub/sites/_form.html.erb
@@ -1,4 +1,3 @@
-<div class="hub-form">
   <%= form_with model: @site, url: url, method: http_method, local: true, builder: VitaMinFormBuilder do |f| %>
     <h1 class="form-card__title">
       <%= @title %>
@@ -6,6 +5,21 @@
 
     <%= f.cfa_input_field :name, t("general.name") %>
     <%= f.cfa_select :parent_organization_id, t("general.organization"), @organizations.map { |c| [c.name, c.id] } %>
+
+    <% if @show_unique_links %>
+      <% @site.source_parameters.each do |source_parameter| %>
+        <%= f.fields_for :source_parameters, source_parameter, builder: VitaMinFormBuilder do |source_parameter_form| %>
+          <%= source_parameter_form.cfa_input_field :code, t("general.unique_link"), classes: ["form-width--short"] %>
+          <%= source_parameter_form.cfa_checkbox(:_destroy, t("general.remove")) %>
+        <% end %>
+      <% end %>
+
+      <h3><%= f.label t("general.new") %></h3>
+
+      <%= f.fields_for :source_parameters, @site.source_parameters.build do |source_parameter_form| %>
+        <%= source_parameter_form.cfa_input_field :code, t("general.unique_link"), classes: ["form-width--short"] %>
+      <% end %>
+    <% end %>
 
     <button class="button button--primary button--wide" type="submit">
       <%= t("general.save") %>
--- a/spec/controllers/hub/sites_controller_spec.rb
+++ b/spec/controllers/hub/sites_controller_spec.rb
@@ -74,10 +74,27 @@ RSpec.describe Hub::SitesController, type: :controller do
         expect(assigns(:organizations)).to include other_org
         expect(response).to be_ok
       end
+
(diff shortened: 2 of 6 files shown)


app/views/hub/sites/_form.html.erb
```diff
@@ -7,6 +6,21 @@
     <%= f.cfa_input_field :name, t("general.name") %>
     <%= f.cfa_select :parent_organization_id, t("general.organization"), @organizations.map { |c| [c.name, c.id] } %>
 
+    <% if @show_unique_links %>
+      <% @site.source_parameters.each do |source_parameter| %>
+        <%= f.fields_for :source_parameters, source_parameter, builder: VitaMinFormBuilder do |source_parameter_form| %>
+          <%= source_parameter_form.cfa_input_field :code, t("general.unique_link"), classes: ["form-width--short"] %>
+          <%= source_parameter_form.cfa_checkbox(:_destroy, t("general.remove")) %>
```
benjaminleesmith, line 13: Consider making this checkbox be in the same "row" visually as the form input it relates to.

benjaminleesmith: If there's a way to save the deletion on click of the 'Remove', that might be a more common pattern.

app/views/hub/sites/_form.html.erb
```diff
@@ -7,6 +6,21 @@
...
+        <% end %>
+      <% end %>
+
+      <h3><%= f.label t("general.new") %></h3>
```
benjaminleesmith, line 17: This sections feel like it should be a "New" button, that clicking that opens a new text field to add N more Unique Urls. If this is only used internally it's probably fine however.

spec/controllers/hub/sites_controller_spec.rb
```diff
@@ -101,7 +128,8 @@
         site.reload
         expect(site.name).to eq "Silly Site"
         expect(site.parent_organization).to eq other_organization
-        expect(response).to redirect_to edit_hub_organization_path(id: other_organization)
+        expect(site.source_parameters.pluck(:code)).to eq(["newshortlink"])
```
benjaminleesmith, line 131: Consider adding an assertion that the old record was deleted (for example) to gain code coverage on the `:id` part of the change to the controller.

spec/features/hub/create_organization_hierarchy_spec.rb
```diff
@@ -39,12 +39,14 @@
...
-      # Update a site
+      # Update the site
       click_on "Llama Library"
       fill_in "Name", with: "Lima Bean Library"
       click_on "Save"
+      expect(page).to have_selector("h1", text: "Lima Bean Library")
 
-      expect(page).to have_selector("h1", text: "Oregano Org")
+      # Navigate to the org
+      click_on "Oregano Org"
```
benjaminleesmith, line 49: I would expect the feature specs to test: 1) that the Unique Urls are not shown on the new page, 2) that a list of all the Unique Urls are shown on the edit page, 3) that you can edit an existing Unique Url, 4) that you can add a new Unique Url, 5) you can remove a Unique Url.

spec/controllers/hub/sites_controller_spec.rb
alaroia-cfa: Done; good call

spec/features/hub/create_organization_hierarchy_spec.rb
alaroia-cfa: This is a good idea. I'll punt on it for now.

app/views/hub/sites/_form.html.erb
alaroia-cfa: Agree on both points.

alaroia-cfa:
Green locally, so merging.

PR #771, opened 2021-02-01
Show 14446 consent form in documents tab

embarnard:


app/controllers/questions/consent_controller.rb
```diff
@@ -9,6 +9,7 @@ def form_params
     end
 
     def after_update_success
+      ConsentPdf.new(current_intake).create_consent_document
```
bytheway875, line 12: If we create the consent document here, we will not be creating a consent document for drop off clients. If we move this to a hook on the model that runs when a user consents, that will cover it getting generated in both workflows.

app/controllers/hub/documents_controller.rb
```diff
@@ -10,6 +10,8 @@ class DocumentsController < ApplicationController
     layout "admin"
 
     def index
+      ConsentPdf.new(@client.intake).create_consent_document if Document.where(document_type: "Consent Form 14446", intake: @client.intake).count.zero?
```
embarnard, line 13: This is a database write on a HTTP GET. Can we move the DB write somewhere else perhaps?

bytheway875: I'd prefer that we run a one-time script to generate these for clients that have not had them generated yet this season rather than generating them when hitting the documents index view.

app/controllers/questions/consent_controller.rb
bytheway875: UTA: Learned from product that we don't need to create any intake or consent documents for drop off clients because they should be signed on site and uploaded.

embarnard: @bytheway875 oh great! This was moved to the after_save method in the model anyways but I think it still wouldn't create a consent document if the primary_consented_to_service_at isn't filled for the drop off client

PR #772, opened 2021-02-01
Update tax year 2020 13614c [#176404841]

bytheway875:
Screen Shot 2021-02-01 at 3 22 18 PM
Screen Shot 2021-02-01 at 3 22 28 PM
Screen Shot 2021-02-01 at 3 22 36 PM
Screen Shot 2021-02-01 at 3 22 44 PM

The PDF mapping was ONLY accepting 0 as the "filled" value for a checkbox, and I have no idea why. But that's what filled the checkbox. 

This year's PDF was significantly different than last years and included Y/N/Unsure radio buttons instead of Y/N boolean checkboxes. The new format resulted in a significant re-mapping of values


app/lib/intake_pdf.rb
```diff
@@ -136,45 +141,45 @@ def dependents_info
...
+      demographic_spouse_race_black_african_american: bool_checkbox_0(@intake.demographic_spouse_black_african_american),
+      demographic_spouse_race_native_hawaiian_pacific_islander: bool_checkbox_0(@intake.demographic_spouse_native_hawaiian_pacific_islander),
+      demographic_spouse_race_white: bool_checkbox_0(@intake.demographic_spouse_white),
+      demographic_spouse_race_prefer_not_to_answer_race: bool_checkbox_0(@intake.demographic_spouse_prefer_not_to_answer_race),
+      demographic_primary_ethnicity: @intake.demographic_primary_ethnicity,
+      demographic_spouse_ethnicity: @intake.demographic_spouse_ethnicity,
+    }
+  end
+
+  def signature_info
```
bytheway875, line 165: @benjaminleesmith See this section -- this is what will add the signature onto this years PDF set for the carryover consent.

app/jobs/intake_pdf_job.rb
```diff
@@ -0,0 +1,6 @@
+class IntakePdfJob < ApplicationJob
+  def perform(intake_id, filename)
```
bytheway875, line 2: When Alex and I were doing some DB cleaning last tax season, we noticed that jobs store the params in the db so that they can be run. When they fail, the entire intake would get saved as an attribute on the job. This is why it is usually more standard to pass IDs into jobs instead of the full object.

app/lib/document_types/F13614CF150802020.rb
```diff
@@ -0,0 +1,19 @@
+module DocumentTypes
+  class F13614CF150802020 < DocumentType
```
bytheway875, line 2: This is a super funky class name, but there is some precedent to using the numbered file name as the document type... Seemed like it would be possibly easier to debug than using a key that did not reflect the class name.

app/models/intake.rb
```diff
@@ -572,10 +576,11 @@ def formatted_mailing_address
     ADDRESS
   end
 
-  def create_13614c_document(filename)
+  def create_intake_document(filename)
+    filename ||= "2020 13614-C with 15080.pdf"
```
bytheway875, line 580: Ruby only seems to use the default value _if no value is passed in at all_ for that parameter. That means that passing in an explicit "nil" value into this method would cause the default filename to get overwritten if i did 

`def create_intake_document(filename='2020...')`

spec/rails_helper.rb
```diff
@@ -96,10 +96,7 @@
     # Stub required credentials to prevent need for RAILS_MASTER_KEY in test
     allow(EnvironmentCredentials).to receive(:dig).and_call_original
     allow(EnvironmentCredentials).to receive(:dig).with(:db_encryption_key).and_return('any-32-character-string-here!!!!')
-
-    allow_any_instance_of(Intake).to receive(:create_13614c_document)
```
bytheway875, line 100: Our default setup probably already includes not running jobs in test. We're no longer calling this method explicity in the after update.

app/lib/document_types/F13614CF150802020.rb
jenny-heath: suggest `Form13614CForm15080`

app/lib/intake_pdf.rb
```diff
@@ -2,7 +2,7 @@ class IntakePdf
   include PdfHelper
 
   def source_pdf_name
-    "f13614c"
+    "f13614c-TY2020"
```
jenny-heath, line 5: get rid of old PDF

app/helpers/pdf_helper.rb
```diff
@@ -1,25 +1,33 @@
...
   def yes_no_unfilled_to_checkbox(value)
     value == "yes" ? "Yes" : nil
   end
 
-  def yes_no_unfilled_to_radio(value)
-    {
-      "yes" => "Yes",
-      "no" => "No",
-    }[value]
+  # Oddly, 0 is checked and 1 is unchecked in the 2020 f13614-c.
```
jenny-heath, line 7: refactor names of shared methods

app/lib/intake_pdf.rb
```diff
@@ -122,10 +126,11 @@ def dependents_info
         relationship: dependent.relationship,
         months_in_home: dependent.months_in_home.to_s,
         disabled: yes_no_unfilled_to_YN(dependent.disabled),
-        north_american_resident: yes_no_unfilled_to_YN(dependent.north_american_resident),
-        on_visa: yes_no_unfilled_to_YN(dependent.on_visa),
+        resident: yes_no_unfilled_to_YN(dependent.north_american_resident),
+        # on_visa: yes_no_unfilled_to_YN(dependent.on_visa),
```
jenny-heath, line 130: delete this line?

app/lib/intake_pdf.rb
```diff
@@ -122,10 +126,11 @@ def dependents_info
...
         student: yes_no_unfilled_to_YN(dependent.was_student),
         marital_status: married_to_SM(dependent.was_married),
+        citizen: nil
```
jenny-heath, line 133: clarify with Rae what to do with this question

PR #773, opened 2021-02-01
Generate 14446 13614 c #176404841

benjaminleesmith:


bytheway875:
I do not think that we should continue regenerating different versions of the 13614 forms with the same name like this -- it will make it really difficult for preparers to understand which one is actually important for them to pay attention to.

app/lib/document_types/consent_form.rb
```diff
@@ -0,0 +1,19 @@
...
+  class ConsentForm < DocumentType
+    class << self
+      def relevant_to?(intake)
+        # This is not used in a flow or to determine relevant types for an intake
+        # and will likely be deprecated after we implement client authentication
+        false
+      end
+
+      def key
+        "Consent Form"
```
bytheway875, line 11: If this is the 14446 Form, might make sense to make the key or display name make that really clear by including "f14446" in the key and/or the creation name.

bytheway875:
By doing all of these form creation tasks inside of the controllers for manual onboarding, I think we're missing creating the consent form for drop-off clients who I'm thinking also need to "sign" these forms.

app/controllers/questions/additional_info_controller.rb
```diff
@@ -8,7 +8,7 @@ def tracking_data
 
     def after_update_success
       current_intake.update(completed_yes_no_questions_at: DateTime.current) if current_intake.completed_yes_no_questions_at.nil?
-      IntakePdfJob.perform_later(current_intake.id, "Preliminary 13614-C.pdf")
+      IntakePdfJob.perform_later(current_intake.id, "Preliminary 13614-C with 15080.pdf")
```
bytheway875, line 11: Thanks for updating these!

app/lib/question_navigation.rb
```diff
@@ -26,6 +26,8 @@ class QuestionNavigation
 
     # Consent
     Questions::ConsentController, # Advances statuses to "In Progress"
```
bytheway875, line 28: These comments make what's going on in each question action / after update easier to understand. Thanks!

app/models/intake.rb
```diff
@@ -576,15 +575,36 @@ def formatted_mailing_address
     ADDRESS
   end
 
-  def create_intake_document(filename)
-    filename ||= "2020 13614-C with 15080.pdf"
+  def update_or_create_13614c_document(filename)
     pdf_tempfile = pdf
```
bytheway875, line 579: I know you didn't implement this, but the pattern of having this other method that creates the Tempfile and then it gets turned into something useable/persist-able here seems strange. I wonder how this could be refactored (not now, but something I'd love to hear about an alternate approach to)

app/models/intake.rb
```diff
@@ -576,15 +575,36 @@ def formatted_mailing_address
...
-    client.documents.create!(document_type: DocumentTypes::Form13614CForm15080.key, intake: self, upload: {
-      io: pdf_tempfile,
-      filename: filename,
-      content_type: "application/pdf",
-      identify: false
-    })
+    document = client.documents.find_or_initialize_by(document_type: DocumentTypes::Form13614CForm15080.key)
+    document.update!(
+      document_type: DocumentTypes::Form13614CForm15080.key,
+      intake: self,
```
bytheway875, line 584: I actually think it is safe to remove the reference to intake on these documents -- it's kind of a vestige of when we were connecting documents to the intake instead of the client last season. Not necessary to do now, but could be a good clean up task.

config/locales/en.yml
```diff
@@ -1752,6 +1752,9 @@ en:
...
           review_return: I understand I need to review my completed tax return to ensure the names, social security numbers, address, banking information, income, and expenses are correct to the best of my knowledge.
           sign_below: By signing below, you agree to participate in the process and consent to allow GetYourRefund assist you with your taxes. You agree to the terms of the GetYourRefund privacy policy at www.GetYourRefund.org/privacy.
+          efin: I understand that the VITA site I am connected with may have a Relational EFIN number (meaning their site reports to a parent VITA organization), and I understand the VITA site has the ability to see reporting data for all EFINs under the relational (primary) EFIN. This consent form replaces the Relational EFIN consent form from the tax preparation software.
+          personal_tax_info: "Consent to Disclose and Use Personal Tax Return Information: I agree to allow the VITA preparation site to use aggregate return data for statistical purposes. This consent form replaces the Use and Disclose Consent form."
```
bytheway875, line 1756: Ah, guessing the quotes are necessary because of the colons. 👍

bytheway875 (approved):
Couple of comments for "nice to dos", but not blocking.

PR #776, opened 2021-02-02
Calculate capacity

alaroia-cfa:

### DIFF
--- a/app/lib/tax_return_status.rb
+++ b/app/lib/tax_return_status.rb
@@ -45,4 +45,8 @@ class TaxReturnStatus
   def self.message_template_for(status, locale = "en")
     message_templates[status.to_sym] ? I18n.t(message_templates[status.to_sym], locale: locale) : ""
   end
+
+  def self.statuses_that_count_towards_capacity
+    STATUSES.keys - [:intake_before_consent, :intake_in_progress, :file_accepted, :file_not_filing]
+  end
 end
--- a/app/lib/tax_return_status.rb
+++ b/app/lib/tax_return_status.rb
@@ -39,14 +39,11 @@ class TaxReturnStatus
       file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406
   }.freeze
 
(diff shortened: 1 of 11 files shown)


alaroia-cfa:
This is just the work @jenny-heath already did on this branch, packaged up as a pull request.

alaroia-cfa:
@embarnard If you're comfortable taking this forward tomorrow, I'd be grateful!

app/lib/tax_return_status.rb
```diff
@@ -45,4 +45,8 @@ def message_templates
   def self.message_template_for(status, locale = "en")
     message_templates[status.to_sym] ? I18n.t(message_templates[status.to_sym], locale: locale) : ""
   end
+
+  def self.statuses_that_count_towards_capacity
```
bytheway875, line 49: In a quick skim, my eye goes straight towards the excepted statuses. Maybe we can set up a const called `STATUSES_EXCLUDED_FROM_CAPACITY = [:intake_before_consent, :intake_in_progress, :file_accepted, :file_not_filing].freeze`, then update this method to use that const. Might be a bit more readable for the people who come after us!

app/models/vita_partner.rb
```diff
@@ -79,28 +76,15 @@ def no_coalitions_for_sites
...
+      errors.add(:capacity_limit, "Sites cannot be assigned a capacity")
+    end
+  end
+
   def one_level_of_depth
     if parent_organization&.parent_organization.present?
       errors.add(:parent_organization, "Only one level of sub-organization depth allowed.")
     end
   end
-
```
bytheway875, line 87: Woo hoo! Yay for removing old code.

app/models/intake.rb
```diff
@@ -607,4 +603,8 @@ def update_or_create_14446_document(filename)
     )
     document
   end
+
```
bytheway875, line 606: I'm guessing this is what gets used in the views to display the delay message? Based on the standup convo yesterday, it seems like this works out of the box for us, with no updates needed, right?

bytheway875 (approved):
👍  One requested change, but not blocking.

PR #778, opened 2021-02-02
Allow percent signs in emails without failure

bytheway875:
When fixing the Delayed::Jobs issues on prod, I looked into some of the specific error messages and realized that replacement parameter replacement failed when the body included a % sign that was not a replacement field (%{key}).

This fixes that by escaping the pre-existing % signs in the body.

### DIFF
--- a/app/services/replacement_parameters_service.rb
+++ b/app/services/replacement_parameters_service.rb
@@ -20,22 +20,21 @@ class ReplacementParametersService
 
   def process_replacements_hash(replacements_hash)
     # escape existing percent signs
-    body.gsub!(/%/, "%%")
+    body.gsub!(/%(?!{\S*})/, "%%")
 
     # replace valid <<key>> with %{key}
     replacements_hash.each_key { |key| body.gsub!(/<<\s*#{key}\s*>>/i, "%{#{key}}") }
 
     body % replacements_hash
   end
-
-
+  
   def replacements
     {
-        "Client.PreferredName" => client&.preferred_name,
-        "Preparer.FirstName" => preparer_first_name,
-        "Documents.List" => documents_list,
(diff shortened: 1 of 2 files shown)


app/services/replacement_parameters_service.rb
```diff
@@ -19,14 +19,15 @@ def process_sensitive_data
...
-    body % replacements_hash
-  end
+    # escape existing percent signs
+    body.gsub!(/%(?!{\S*})/, "%%")
 
-  def convert_template_keys_to_string_format_syntax(keys)
-    keys.each{ |key| body.gsub!(/<<\s*#{key}\s*>>/i, "%{#{key}}") }
-  end
+    # replace valid <<key>> with %{key}
+    replacements_hash.each_key { |key| body.gsub!(/<<\s*#{key}\s*>>/i, "%{#{key}}") }
```
alaroia-cfa, line 26: This makes sense! The computer scientist in me wishes we had a more conceptually holistic way to handle all these cases.

For example, I wonder can we set`.default` on the replacements_hash?

Anyway what you've done here is great. I'm happy to see it merged.

alaroia-cfa (approved):
Splendid. Great work catching these and writing a clear fix.

app/services/replacement_parameters_service.rb
bytheway875: Oh! This is something to think about... So this change covers use cases where percentage signs are used that dont look like %{...}. So if there is a `50%` or `your random string is %{45#e`.  What it DOESN'T cover is situations in which there are string in the format of `%{...}` that don't match a replacement parameter. For example `your random string is %{45#e}`. That's still a risk, and I'm not sure how we can mitigate it. Def a downside of doing `body % replacement_params` replacement on a string we don't have control over.

PR #779, opened 2021-02-03
Add ssn to permitted params for create page, add N/A state, and add feature tests [#176629554]

bytheway875:


app/views/shared/_tax_return_list.html.erb
```diff
@@ -4,9 +4,11 @@
   <% client.tax_returns.order(year: :asc).each do |tax_return| %>
     <li id="tax-return-<%= tax_return.id %>">
       <div class="tax-return-list__assignment">
-        <span class="tax-return-list__year"><%=
-          tax_return.year
-        %></span><span class="tax-return-list__service-type"><%
+        <span class="tax-return-list__year">
```
bytheway875, line 7: I have NO IDEA why i need to add this space here in order for the feature test to run clean when using JS true, but without this space, it cant find that content. 

This little section is funky in general -- I'm not sure why the other spans are relying on the template tags being on separate lines. I remember someone telling me it was for styling reasons, which seems like maybe we should style it differently instead of having such delicate html.😂 

BUT that is a problem for another day!

jenny-heath (approved):
looks great, thank you!

PR #780, opened 2021-02-03
Add script to generate consent form for old intakes

embarnard:

### DIFF
--- a/app/lib/generate_consent_documents.rb
+++ b/app/lib/generate_consent_documents.rb
@@ -1,0 +1,37 @@
+class GenerateConsentDocuments
+  def create_consent_docs
+    # GenerateConsentDocuments.new.create_consent_docs
+    intakes_without_consent_doc = []
+    successes = []
+    problems = []
+
+    Intake.all.each do |intake|
+      consent_form_exists = false
+
+      intake.documents.each do |doc|
+        if doc.document_type == DocumentTypes::Form14446.key
+          consent_form_exists = true
+        end
+      end
+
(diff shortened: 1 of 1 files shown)


app/lib/generate_consent_documents.rb
```diff
@@ -0,0 +1,37 @@
+class GenerateConsentDocuments
+  def create_consent_docs
+    # GenerateConsentDocuments.new.create_consent_docs
+    intakes_without_consent_doc = []
+    successes = []
+    problems = []
+
+    Intake.all.each do |intake|
```
alaroia-cfa, line 8: It would be substantially more efficient to do this with some kind of `JOIN` query, but I suppose it doesn't matter much if we run this in the background and/or on a temporary `aptible ssh` worker.

app/lib/generate_consent_documents.rb
```diff
@@ -0,0 +1,37 @@
...
+      consent_form_exists = false
+
+      intake.documents.each do |doc|
+        if doc.document_type == DocumentTypes::Form14446.key
+          consent_form_exists = true
+        end
+      end
+
+      intakes_without_consent_doc << intake.id unless consent_form_exists || intake.primary_consented_to_service_at.nil?
```
alaroia-cfa, line 17: If you can use the `.primary_consented_to_service_at` column to decrease the number of Intake records you loop across, that would also provide a substantial speed-up. Not a blocker.

alaroia-cfa (approved):
Great work dealing with the old data! Looks great; feel free to merge.

I have some suggestions for speedups, which seem like they won't matter much given the way we'll run this.

PR #782, opened 2021-02-03
Only regenerate login token when sending login link

embarnard:

### DIFF
--- a/app/services/replacement_parameters_service.rb
+++ b/app/services/replacement_parameters_service.rb
@@ -45,9 +45,13 @@ class ReplacementParametersService
   # this should only be called when we aren't saving the output to our database
   # for example, when adding a link to send to the client only
   def sensitive_replacements
-    {
-      "Link.E-signature": client.login_link
-    }
+    if body.match(/<<\s*Link\.E-signature\s*>>/i)
+      {
+        "Link.E-signature": client.login_link
+      }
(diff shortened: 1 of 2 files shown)


app/services/replacement_parameters_service.rb
```diff
@@ -45,9 +45,13 @@ def you_or_your
   # this should only be called when we aren't saving the output to our database
   # for example, when adding a link to send to the client only
   def sensitive_replacements
-    {
-      "Link.E-signature": client.login_link
-    }
+    if body.match(/<<\s*Link\.E-signature\s*>>/i)
```
alaroia-cfa, line 48: I'm surprised you didn't need to do `\-` instead of `-` but if it works, fine with me!

alaroia-cfa (approved):
Great work!

app/services/replacement_parameters_service.rb
```diff
@@ -45,9 +45,13 @@ def you_or_your
...
+      {
+        "Link.E-signature": client.login_link
```
bytheway875, line 50: This is kind of a "dangerous" method in that invoking it actually updates stuff on the client, but it kind of just looks like an attribute. We might consider changing this method name to "generate_login_link" to make it clear that its doing something other than just _providing_ the login link.

PR #784, opened 2021-02-04
Circle CI: Only launch database for tasks that need it

alaroia-cfa:
This can cause a (approx) 10 sec build speedup.


bytheway875 (approved):
This is cool!

alaroia-cfa:
@bytheway875 Thanks for saying it's cool! :D I'm really hitting a bit of a wall with build speedups. The only good next step I can find is to save approx 45 seconds on a deploy, if we move our assets to S3/CDN - this saves time b/c we can do the asset build outside the Docker container build.

PR #787, opened 2021-02-04
Add spouse SSN and fix bug where editing clears ssn info

bytheway875:
There was a bug that was causing an existing SSN to get cleared when the user edited it. That is also fixed here.

Probably could have refactored to reuse some templates between spouse and primary, but I didn't feel that it greatly added to complexity to keep them seperate.

### DIFF
--- a/app/forms/hub/update_client_form.rb
+++ b/app/forms/hub/update_client_form.rb
@@ -22,6 +22,7 @@ module Hub
                        :state,
                        :zip_code,
                        :primary_last_four_ssn,
+                       :spouse_last_four_ssn,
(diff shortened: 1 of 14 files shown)


alaroia-cfa (approved):
Clear code and solid tests. Approved!

app/forms/hub/update_client_form.rb
```diff
@@ -53,6 +54,11 @@ def dependents
       @client.intake.dependents
     end
 
+    def self.existing_attributes(intake)
```
alaroia-cfa, line 57: I suppose the absence of this was the core of the bug?

PR #788, opened 2021-02-04
Allow slow test to be slow

alaroia-cfa:
This was being flaky here: https://app.circleci.com/pipelines/github/codeforamerica/vita-min/4139/workflows/78a830c8-bc18-432e-a281-e82bfb383834/jobs/11196/parallel-runs/2

Maybe if we wait longer, it'll just be fine.


alaroia-cfa:
A slippery slope! But better than not having tests.

PR #789, opened 2021-02-05
Use faster JS compressor called Terser

alaroia-cfa:
This improves the speed of our code deploys by improving the speed of `rails assets:precompile`. In my local testing, it improves `rake assets:precompile` by approx 25%, which could speed up a deploy by approx 10-15 seconds.

GitLab also uses terser, suggesting that it's a pretty OK choice: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/44034


alaroia-cfa:
👍

PR #790, opened 2021-02-05
Remove offseason code

alaroia-cfa:
@bytheway875 says it's better if we do this differently next time we have an offseason, and it's not the offseason, and the tests for offseason are slow, so let's delete this!

```
❯ grep spec/routing/offseason_routing_spec.rb tmp/parallel_runtime_rspec.log
spec/routing/offseason_routing_spec.rb:23.584138000151142
```


alaroia-cfa:
We can't remove all offseason code b/c we're not fully publicly launched.

We can remove the one slow routing spec if we like.

PR #791, opened 2021-02-05
Speed up i18n-tasks rspec checks

alaroia-cfa:
Before:

```
$ grep i18n tmp/parallel_runtime_rspec.log
spec/i18n/inconsistent_interpolations_spec.rb:0.37661600000046747
spec/i18n/missing_keys_spec.rb:32.69252199999937
spec/i18n/unused_keys_spec.rb:49.159835999999814
```

After:

```
$ grep i18n tmp/parallel_runtime_rspec.log
spec/i18n_spec.rb:6.919614000000365
```

This does the job of all 3 tests in one big assertion, which is convenient because they share work, so this decreases total test execution time.

It also ignores the zip codes file when validating i18n keys. This was crucial.

I figured this out by running `rb-spy` when this test was running, noticing that most of the time was being spent in `/Users/asheesh/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/i18n-tasks-0.9.31/lib/i18n/tasks/scanners/ruby_ast_scanner.rb`, specifically `scan_file`, so I modified it to print out which file it was processing and how long it was taking. It was clear the ZIP codes file was taking a huge amount of time. Removing it from the config makes the task >50% faster.


alaroia-cfa:
Woo hoo!

PR #792, opened 2021-02-05
Fix timezone bug on user edit page

bytheway875:
From investigation, found that the timezone field was originally only being set on the valet drop off page, so we set the default timezone to the current users' timezone as a convenience. However, when we started using a shared partial between the edit client and create client pages, we never updated the logic. This meant that when users from another timezone edited a client, their timezone box value would get changed to the current users timezone in error.

The current fix is to NOT prefill the value with the current users in the valet drop off, and ensure that the current value gets used on the clients page.


alaroia-cfa (approved):
> The current fix is to NOT prefill the value with the current users in the valet drop off

I think this is removing a feature that product might want. If they're OK with it, I'm OK with it! Curious if you have any insight into that. Happy to Tuple if that's easier.

Approved! Great work tracking this down.

PR #796, opened 2021-02-05
Redirect DIY flow to homepage with diy_off config flag

embarnard:


embarnard (commented):
There is an argument to just remove the logic deciding whether or not to redirect the DIY flow but since we are hoping to turn the flow back on soon I thought this would be easier to keep mostly the tests in tact

alaroia-cfa (approved):
Simple and good plan. 🚢 

I have found in the past that tests with `reload_routes!` are slow to execute, but when I run one of your files here, it only takes a few seconds, so I'll reserve judgement on that question. (And hopefully we'll remove the option and restore DIY soon anyway.)

PR #797, opened 2021-02-08
Load ZIP code & state data lazily

alaroia-cfa:
Before, it takes 20 seconds to warm the bootsnap cache:

```
$ rm -rf tmp/cache/bootsnap-compile-cache && time bundle exec bootsnap precompile --gemfile app/ lib/
bundle exec bootsnap precompile --gemfile app/ lib/  15.16s user 5.30s system 98% cpu 20.712 total
```

After, it takes 14 seconds to warm the bootsnap cache:

```
$ rm -rf tmp/cache/bootsnap-compile-cache && time bundle exec bootsnap precompile --gemfile app/ lib/
bundle exec bootsnap precompile --gemfile app/ lib/  8.16s user 5.16s system 96% cpu 13.741 total
```

I verified with `Hashdiff.diff` that the old hash & new hash are the same.


bytheway875:
LGTM, but I'm interested to understand more about why these changes have this effect!

bytheway875 (approved):

PR #801, opened 2021-02-08
Convert heic #175489605

benjaminleesmith:


app/jobs/heic_to_jpg_job.rb
```diff
@@ -0,0 +1,6 @@
+class HeicToJpgJob < ApplicationJob
```
alaroia-cfa, line 1: Simple and nice!

lib/tasks/convert_heic_documents_to_jpg.rake
```diff
@@ -0,0 +1,17 @@
+namespace :documents do
+  desc "Converts existing documents with heic attachments to jpg"
+  task convert_heic_to_jpg: [:environment] do
+    Document.find_each do |document|
+      if document.upload.filename.extension_without_delimiter.downcase == "heic"
+        puts "Converting document #{document.id}"
+        begin
+          document.convert_heic_upload_to_jpg!
+        rescue Expection => e
```
alaroia-cfa, line 9: I think `Expection` here is a typo for `Exception`. I'd like to see that fixed in the PR.

Non-blocking: Since it's a rake task and going to be executed manually, I prefer you remove the begin/rescue and allow the rake task to crash. I think we don't need to proceed through and process the other HEIC files if one raises an error; we can fix it or remove the image, then re-run the Rake task. I'm open to the existing behavior (work through all HEIC files that don't crash, printing out a message for those that crash) if you prefer.

spec/jobs/heic_to_jpg_job_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+describe HeicToJpgJob, type: :job do
+  describe "#perform" do
+    let(:document) { create(:document) }
+
+    before do
+      allow(Document).to receive(:find).and_return(document)
```
alaroia-cfa, line 8: This is a nice tidy way of avoiding `alllow_any_instance_of`!

spec/models/document_spec.rb
```diff
@@ -131,4 +155,14 @@
...
     end
   end
+
+  describe "#convert_heic_upload_to_jpg!" do
+    it "converts a heic attachment to jpg" do
+      document = create :document, upload_path: Rails.root.join("spec", "fixtures", "attachments", "IMG_4851.HEIC")
+
+      expect {
+        document.convert_heic_upload_to_jpg!
+      }.to change { document.upload.attachment.filename.extension }.from("HEIC").to("jpg")
```
alaroia-cfa, line 165: This test could be more exhaustive - perhaps we can ask the `mimemagic` library that we have in the project what the file's detected MIME type is. Not a blocker.

alaroia-cfa (changes requested):
Just one requested change, plus some suggestions.

Great job!

PR #803, opened 2021-02-09
Add basic monitors for Twilio & Mailgun

bengolder:


app/controllers/twilio_webhooks_controller.rb
```diff
@@ -3,7 +3,9 @@ class TwilioWebhooksController < ActionController::Base
   before_action :validate_twilio_request
 
   def update_outgoing_text_message
-    OutgoingTextMessage.find(params[:id]).update(twilio_status: params["MessageStatus"])
+    status = params["MessageStatus"]
+    DatadogApi.increment("twilio.outgoing_text_messages.updated.status.#{status}")
```
alaroia-cfa, line 7: dumb question - are these always nice lowercase under_scored_words strings? If they're not, might be nice to transform them.

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -12,7 +14,13 @@ def update_outbound_call
     return unless call.present?
 
     update_params = { twilio_status: params["CallStatus"] }
-    update_params[:twilio_call_duration] = params["CallDuration"] if params["CallDuration"].present?
+    DatadogApi.increment("twilio.outbound_calls.updated.status.#{params["CallStatus"]}")
```
alaroia-cfa, line 17: Same question here

app/controllers/twilio_webhooks_controller.rb
```diff
@@ -12,7 +14,13 @@ def update_outbound_call
...
+
+    if params["CallDuration"].present?
+      update_params[:twilio_call_duration] = params["CallDuration"]
+      DatadogApi.gauge("twilio.outbound_calls.updated.duration", params["CallDuration"].to_i)
```
alaroia-cfa, line 21: I _think_ that using a gauge would imply the current duration of outbound calls at this moment _across the whole system_ is this _one_ call's duration, based on reading https://prometheus.io/docs/concepts/metric_types/#:~:text=A%20gauge%20is%20a%20metric,the%20number%20of%20concurrent%20requests. . Let me know if you agree/disagree.

alaroia-cfa: I think we'd need to separate these, for example by call ID, like how one container's current CPU usage is separated from a different container's CPU usage. I don't personally know how to do that off the top of my head, but I'm willing to look into it with you to find out.

alaroia-cfa (approved):
This code is great, and covers the basics. I approve. Great tests & great prep work toward having graphs in Datadog.

I think we need to rework the gauge metric you created, but unfortunately I don't know how off the top of my head. I'm fine with this going in as-written if you like.

Having these alerts in Datadog sets us up well to be able to add them to our primary dashboard, which is going to be hugely helpful in seeing problems.

app/controllers/twilio_webhooks_controller.rb
bengolder: As far as I can tell, yes, they're always lower case single words.
https://support.twilio.com/hc/en-us/articles/223134347-What-are-the-Possible-SMS-and-MMS-Message-Statuses-and-What-do-They-Mean-

app/controllers/twilio_webhooks_controller.rb
bengolder: You might be right that I'm not using gauge appropriately here. I don't think it would cause any problems and we should still be able to view stats like max, min, avg over a given time period. Based on [this documentation of metric types](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#metric-types), it looks like the `distribution` type is closest to what I'm looking for. However, I can't tell if it's something we can report through the HTTP API.

All calls to datadog are scoped to an environment by default. But other than that, I can't think of some other scope we'd like to use for filtering. A metric per phone call wouldn't make sense to me, because that would just be logging (which isn't a bad idea, just not what Datadog metrics are for).

I also don't think it's critical to include duration reporting. I thought it might be helpful because a consistent short durations could indicate calls that are successfully connected but are failing in some way that prevents a proper voice interaction.

I'll merge for now, but I think I'll try adjusting the duration to be a distribution type instead of a gauge as a follow up to see if that's possible.

PR #804, opened 2021-02-10
Client login from home page

bengolder:
- change logic/models so that a new login link doesn't destroy old ones.
  - new model
    - email
    - phone number
    - hashed token
    - expires after a certain time (optional)
    - (possible) many2many association with client
  - matching contact info to client will work the same
  - locking clients due to failed attempts would be the same
  - if the email or text message fails, we should add a note to the client's account
- review/refine designs
- "send another link" functionality (should just link to the "new" page with prefill contact info)
- carry locale through to email & text message and add translations
- on bad token page, use hello@ instead of support@
- redirect to sign in page if unauthenticated (retain path they were trying to visit) for any portal page
- redirect to path after sign in if they tried to visit something before login


app/services/login_link_insertion_service.rb
```diff
@@ -0,0 +1,20 @@
+class LoginLinkInsertionService
+  class << self
+    include Rails.application.routes.url_helpers
+
+    def insert_links(contact_record)
+      e_signature_link_pattern = /<<\s*link\.e-signature\s*>>/i
+
+      return contact_record.body unless e_signature_link_pattern.match?(contact_record.body)
+
+      raw_token = if contact_record.is_a?(OutgoingTextMessage)
```
alaroia-cfa, line 8: Could be a `case` statement on the `contact_record.class`

alaroia-cfa (approved):
This is all good. Let's merge it and keep working.

PR #807, opened 2021-02-10
Breach dashboard for needs attention breaches

bytheway875:
Screen Shot 2021-02-10 at 2 42 30 PM

### DIFF
--- a/app/controllers/hub/metrics_controller.rb
+++ b/app/controllers/hub/metrics_controller.rb
@@ -1,0 +1,22 @@
+module Hub
+  class MetricsController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    layout "admin"
+    load_and_authorize_resource :vita_partner, parent: false
+
+    def index
+      # Run SLA queries and cache the result for 10 minutes.
+      # The view will only display values for vita partners the user has access to view.
+      data = Rails.cache.fetch("metrics/sla_breaches/attention_needed", expires_in: 10.minutes) do
+        sla_service = SLABreachService.new
+        { breach_threshold: sla_service.breach_threshold,
+          current_as_of: sla_service.report_generated_at,
+          breach_counts: sla_service.attention_needed_breach
+        }
+      end
+      data[:total_breaches] = data[:breach_counts].slice(*@vita_partners.map(&:id)).values.inject(:+)
+      @attention_needed = OpenStruct.new(data)
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


app/controllers/hub/metrics_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  class MetricsController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    layout "admin"
+    load_and_authorize_resource :vita_partner, parent: false
+
+    def index
+      # Run SLA queries and cache the result for 10 minutes.
+      # The view will only display values for vita partners the user has access to view.
+      data = Rails.cache.fetch("metrics/sla_breaches/attention_needed", expires_in: 10.minutes) do
```
bytheway875, line 11: 10 minute server cache on the SLA breach data -- enough to ensure we're not re-running this query on every page load or for every User who accesses the metrics, but often enough to make sure the data is useful to those who access it.

config/business_time.yml
```diff
@@ -0,0 +1,10 @@
+business_time:
+  beginning_of_workday: 12:00am
```
bytheway875, line 2: Per product, we mostly want to exclude the weekends from our calculations. By extending the workday from 12:00am to 11:59pm.

app/controllers/hub/metrics_controller.rb
sniemeyer13: Neat

app/services/sla_breach_service.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def breach_threshold
+    3.business_days.before(@report_generated_at)
+  end
+
+  # attention_needed_since is present and is before breach threshold
+  # returns data in the format { vita_partner_id: breach_count }
+  def attention_needed_breach
+    Client.sla_tracked.group(:vita_partner_id).where(Client.arel_table[:attention_needed_since].lteq(breach_threshold)).count(:id)
```
sniemeyer13, line 14: This took me a minute to wrap my brain around.. may want to refactor to make it more clear - not urgent or blocking though

sniemeyer13 (approved):
Cool! Left one suggestion but nothing blocking.

app/services/sla_breach_service.rb
bytheway875: @sniemeyer13 do you think it would make it more readable if i broke it into separate lines so I could comment what each part is doing?

PR #809, opened 2021-02-10
Add cancel button to take action page

sniemeyer13:


bytheway875 (approved):
seems straightforward!! 👍

bytheway875:
Also, I think there _might_ be some built-in rails magic you can use to have the cancel button work like a "back" button so that it will drop you back on whatever page you clicked it from:

https://stackoverflow.com/questions/620962/back-browser-action-in-ruby-on-rails

PR #811, opened 2021-02-11
Use /tmp for storage when running tests

alaroia-cfa:
On my laptop, `~/projects/vita-min/tmp/storage` was taking up 22 GB of space.

This adjusts ActiveStorage to store its data somewhere that macOS automatically cleans. (See e.g. [superuser.com](https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted) or [developer.apple.com](https://developer.apple.com/forums/thread/71382) for info about /tmp autocleaning.)

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -7,7 +7,7 @@ deploy_default: &deploy_default
 test:
   service: Disk
   # Use /tmp which is autocleared periodically on macOS
-  root: "/tmp/via-min-test-storage"
+  root: "/tmp/vita-min-test-storage"
 
 local:
   service: Disk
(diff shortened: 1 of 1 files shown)


bytheway875:
@alaroia-cfa does this only affect us locally, or does it have implications in other environments as well?

alaroia-cfa:
> @alaroia-cfa does this only affect us locally, or does it have implications in other environments as well?

It's only in the `test` environment for ActiveStorage, so I believe it will affect CI as well as us when running the tests manually. CI has a /tmp and it's cleared at the end of every CI job, so I think it's peaceful there.

bytheway875 (approved):

PR #812, opened 2021-02-11
Add basic, unstyled pagination for users & clients

bengolder:
- Includes basic 25-per-page pagination for both client & user indices.
- Removes the stop-gap "client search" page for admins (I'm assuming that the regular search is sufficient)
- Works fine with filters and search (out of the box!)
- Does not include translations for pagination widgets (just shows up in English on Spanish hub pages)
- Does not include final styling for pagination widgets
- I only added one unit test because it seemed like this PR was a small amount of boilerplate to add a well-tested library.
pagination


app/models/client.rb
```diff
@@ -32,6 +32,8 @@
 class Client < ApplicationRecord
   devise :lockable, :timeoutable, :trackable
 
+  self.per_page = 25
```
bytheway875, line 35: This doesn't particularly feel like a Model concern to me, but it seems like this is how will_paginate gem directs us to set these default values, so 👍

app/views/hub/clients/index.html.erb
```diff
@@ -2,13 +2,23 @@
...
     <%= render "shared/client_filters" %>
 
     <div class="actions-wrapper">
       <%= link_to t(".add_client"), new_hub_client_path, class: "button" %>
     </div>
+
+    <%= page_entries_info @clients, model: "entry" %>
+    <%= will_paginate(
+          @clients,
+          previous_label: "<i class=\"icon icon-keyboard_arrow_left\"></i><span class=\"hide-on-mobile\"> Previous</span>",
```
bytheway875, line 16: seems a little verbose that we're having to pass in the exact icons as a string. would love to see a partial or helper for the icon/label so we can easily reuse, or better yet a wrapper around the will_paginate helper that creates a consistent styling that we can add to other pages. But as we're only using in 2 places, doesn't violate the rule of threes to leave this refactor for another time!

bytheway875 (approved):
👍  A couple comments with a suggested refactor to ensure visual consistency of pagination within the application, but def not blocking especially as this is fairly urgent!!

app/views/hub/clients/index.html.erb
bengolder: This is a great point! I plan to follow up soon with a styling commit, which would be a good opportunity to address this.

PR #813, opened 2021-02-11
First Unresponded To Message logic for clients

bytheway875:
Business stakeholders want to know when the last unresponded to client message was. After considering some ways to query for this, it seemed like a better idea to use a cachetribute to track this. We set / clear this attribute in interaction tracking similarly to the needs_attention notifier. However, it cannot be cleared by anything other than a direct message to the client -- interactions with the profile and manually resolving the indicator won't wipe it out.

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -2,24 +2,24 @@
 #
 # Table name: clients
 #
-#  id                                          :bigint           not null, primary key
-#  attention_needed_since                      :datetime
-#  current_sign_in_at                          :datetime
-#  current_sign_in_ip                          :inet
-#  failed_attempts                             :integer          default(0), not null
-#  first_unanswered_incoming_correspondence_at :datetime
-#  last_incoming_interaction_at                :datetime
-#  last_interaction_at                         :datetime
-#  last_sign_in_at                             :datetime
-#  last_sign_in_ip                             :inet
-#  locked_at                                   :datetime
-#  login_requested_at                          :datetime
-#  login_token                                 :string
-#  routing_method                              :integer
-#  sign_in_count                               :integer          default(0), not null
-#  created_at                                  :datetime         not null
-#  updated_at                                  :datetime         not null
-#  vita_partner_id                             :bigint
+#  id                                       :bigint           not null, primary key
+#  attention_needed_since                   :datetime
+#  current_sign_in_at                       :datetime
+#  current_sign_in_ip                       :inet
+#  failed_attempts                          :integer          default(0), not null
+#  first_unanswered_incoming_interaction_at :datetime
+#  last_incoming_interaction_at             :datetime
+#  last_interaction_at                      :datetime
+#  last_sign_in_at                          :datetime
+#  last_sign_in_ip                          :inet
+#  locked_at                                :datetime
+#  login_requested_at                       :datetime
+#  login_token                              :string
+#  routing_method                           :integer
+#  sign_in_count                            :integer          default(0), not null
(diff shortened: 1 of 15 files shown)


app/models/client.rb
```diff
@@ -2,23 +2,24 @@
...
-#  sign_in_count                :integer          default(0), not null
-#  created_at                   :datetime         not null
-#  updated_at                   :datetime         not null
-#  vita_partner_id              :bigint
+#  id                                          :bigint           not null, primary key
+#  attention_needed_since                      :datetime
+#  current_sign_in_at                          :datetime
+#  current_sign_in_ip                          :inet
+#  failed_attempts                             :integer          default(0), not null
+#  first_unanswered_incoming_correspondence_at :datetime
```
bytheway875, line 10: i used _correspondence_ here because originally product had asked only to track based on incoming messages (not documents) but since the intention is to compare as close to apples-to-apples with needs attention (other than using the manual resolve button) im wondering if "first_unanswered_incoming_interaction_at" makes more sense?

app/services/sla_breach_service.rb
```diff
@@ -10,7 +10,12 @@ def breach_threshold
 
   # attention_needed_since is present and is before breach threshold
   # returns data in the format { vita_partner_id: breach_count }
-  def attention_needed_breach
+  def attention_needed_breaches
     Client.sla_tracked.group(:vita_partner_id).where(Client.arel_table[:attention_needed_since].lteq(breach_threshold)).count(:id)
   end
+
+  # clients who messaged us and have not been responded to _with a message, call or email_ within the breach threshold
+  def outgoing_communication_breaches
```
bytheway875, line 18: This PR adds the query, but a subsequent one will add this data to the metrics view -- I want to make sure we have the seeded data in prod before we add the view for it.

app/models/client.rb
sniemeyer13: True. What about something like "sla_response_needed_since"?

app/services/sla_breach_service.rb
```diff
@@ -10,7 +10,12 @@ def breach_threshold
...
+    Client.sla_tracked.group(:vita_partner_id).where(Client.arel_table[:first_unanswered_incoming_correspondence_at].lteq(breach_threshold)).count(:id)
```
sniemeyer13, line 19: It took me a second to understand how being less than a threshold would lead to a breach. Then I realized that we were comparing dates.. maybe calling it breach_threshold_date would be clearer?

app/services/sla_breach_service.rb
sniemeyer13: Awesome!

spec/services/sla_breach_service_spec.rb
```diff
@@ -107,23 +109,109 @@
...
               vita_partner_2.id => 2
+            }
+          )
+        end
+      end
+
+      context "on Monday @ 11:25am UTC" do
+        it "trips the 10:55am client into SLA breach" do
+          t = Time.utc(2021, 2, 6, 0, 0, 0) # 2/6/21
+          Timecop.freeze(t.next_occurring(:monday) + 11.hours + 25.minutes) do
```
sniemeyer13, line 127: Just learned that we've had some issues with Timecop in the past - might be worth a discussion

sniemeyer13 (approved):
We might want to add a chore to delete the backfill job after it's run. Some small renaming suggestions, but otherwise great work! Approved.

PR #814, opened 2021-02-11
Restore hard deletion of unlinked users

bengolder:
- This restores the ability to "hard delete" users who aren't linked to notes, tax returns, or messages
- It includes the proper DB transaction scoping to ensure that we don't delete the role when the user can't be deleted.


bengolder:
Reviewed and approved by @sniemeyer13

PR #816, opened 2021-02-11
(For Lou) Remove unused diy_home template

alaroia-cfa:
Hey @loumoore - it looks to us that this template cannot be routed to, and that it's unused. I'm hoping to delete it, but wanted to check with you just in case I'm missing something. I saw it was related to ad campaigns.


alaroia-cfa:
@loumoore We no longer send links for DIY, so the text is inaccurate, so I'm going to remove this template, but please do let me know if we need to do something!

loumoore:
@alaroia-cfa tl;dr it's fine that you deleted it :)

much longer answer: it sounds like we would need to update that template anyway in order for it to make sense, so I think it's fine that you removed it. on the other hand, it is not unlikely that we'll end up adding a DIY landing page in again, because targeted landing pages for the various service options worked a lot better than sending everyone to the homepage last year. on the other other hand, i'm not totally sure how that will play out this year yet. it's possible we won't want to drive a lot of traffic to DIY if our full service flow performs much better and we have enough VITA capacity, in which case we can treat DIY more like an emergency outlet as opposed to a primary channel and wouldn't run as much or any ads.

in practice, after public launch, when we're ready to turn on ads, we'll start experimenting and we might just need to create a new landing page at that point, which seems fine!

PR #822, opened 2021-02-12
Sla breach last interaction value update

bytheway875:
last_interaction_at was previously recording incoming as well as outgoing interactions, but we need to isolate our internal/outgoing interactions in a field. We could have _added_ a field called last_outgoing_interaction_at, but a) we don't use the last_interaction_at value for anything currently and b) updated_at on client records a very similar value and it's what we're displaying in the hub. (additionally captures saves/updates to client object, but if we want last_interaction_at to capture that as well, we can update it to do that).


lib/tasks/last_interaction.rake
```diff
@@ -0,0 +1,14 @@
+namespace :last_interaction do
+  desc "seeds more appropriate last_interaction value onto all existing clients."
+  task "backfill" => :environment do
+    # break job into separate processable jobs in batches of 1000.
+    (0..Client.all.size).step(1000) do |i|
```
bengolder, line 5: I knew about `.count` vs `.length` but this is the first I learned of `.size`! interesting!

app/jobs/backfill_appropriate_last_interaction_value.rb
```diff
@@ -0,0 +1,36 @@
...
+        next
+      end
+
+      # Looking for the most recent of each interactionable item.
+      itm = client.outgoing_text_messages.last
+      ie = client.outgoing_emails.last
+      doc = client.documents.reverse.find { |doc| doc.uploaded_by && doc.uploaded_by != client }
+      call = client.outbound_calls.last
+      note = client.notes.last
+      tax_return = client.tax_returns.max_by { |tr| tr&.created_at }
```
bengolder, line 19: consider changing to `updated_at`

app/jobs/backfill_appropriate_last_interaction_value.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      last_interaction_item = [itm, ie, doc, call, note, tax_return].compact.max_by { |msg| msg&.created_at }
```
bengolder, line 21: needs some different max comparison because of tax return

bengolder (approved):
Thank you!

PR #824, opened 2021-02-12
Remove Zendesk user columns [#176896651]

alaroia-cfa:
These were only used in a factory.

Thanks to @bengolder for noticing it's a good time to remove them.


alaroia-cfa:
approved by @bengolder

PR #829, opened 2021-02-16
Persist reports

bytheway875:
Reports were ephemeral and (attempting to) cache. Now, saving them into a reports model in a JSON column so that we can see progression over time.

Currently running the report MAX once every 10 minutes in the foreground -- the way its set up, new reports will only be run when someone requests the metrics dashboard. Probably makes more sense to run this in a cron job, but baby stepping on this!

NOTE TO SELF: For comparison and review purposes, I have the comparison brach as the UI branch that this builds off of. Must retarget to main once that branch is merged!!

### DIFF
--- a/app/controllers/hub/metrics_controller.rb
+++ b/app/controllers/hub/metrics_controller.rb
@@ -6,37 +6,16 @@ module Hub
     load_and_authorize_resource :vita_partner, parent: false
 
     def index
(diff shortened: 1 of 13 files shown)


bytheway875 (commented):
Stellar ⭐

bytheway875:
@alaroia-cfa proposed some non-blocking suggestions; feel free to use or not

PR #831, opened 2021-02-17
Add back button to questions pages #176807862

bytheway875:
Screen Shot 2021-02-17 at 9 22 58 AM

Back button on questions and documents pages.


app/lib/concerns/controller_navigation.rb
```diff
@@ -24,16 +24,22 @@ def next
...
+
+    controllers_to_beginning = controllers[0..index - 1].reverse
+    seek(controllers_to_beginning)
+  end
+
   private
 
   def index
-    controllers.index(@current_controller.class)
+    controllers.index(current_controller.class)
```
bytheway875, line 37: The way that Rspec implements mock controllers means that current_controller.class on a mocked controller returns an instance of a class that looks like this `<Class#1239123>` instead of the class name as a const. As such get the tests for prev_path working in the controller tests, I needed to stub current_controller, and I cant stub an instance variable, so I turned it into an accessor method.

bytheway875: The alternatives were to:
- not test the method
- use eval(current_controller.class.name) to turn the stringified class name back into a const
- refactor a lot of this controller navigation code that I don't fully understand the implications of.

So this seemed like the best option.

app/lib/document_navigation.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DocumentNavigation
```
bytheway875, line 3: Looks like we pulled controllerNavigation into a concern when we started having multiple "flows", but we never imported the concern into here and had most of the methods duplicated. Where possible, I removed the duplicated methods.

spec/support/shared_examples/controller_navigation_flow.rb
```diff
@@ -58,4 +58,22 @@ class ThirdController < BaseController; end
...
+    it "returns numeric index for next non-skipped controller in main flow" do
+      navigation = navigation_class.new(ThirdController.new)
+      expect(navigation.prev).to eq(FirstController)
+    end
+  end
+
+  context "when current controller is the first" do
+    it "returns nil" do
+      navigation = navigation_class.new(FirstController.new)
+      expect(navigation.prev).to be_nil
```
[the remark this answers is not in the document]
alaroia-cfa: Answer: Documents flow has no test coverage :D

alaroia-cfa (approved):
It's great! Merge or address non-blocking comments.

app/views/layouts/document_upload.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :page_title, (yield :form_question) %>
 <% content_for :main do %>
-  <section class="slab slab--white">
+  <section class="slab slab--white slab--not-padded question-layout">
```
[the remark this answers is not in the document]
bytheway875: Putting this off because document_upload is its own template, and nesting templates in templates is kind of complex and outside scope of this task.

PR #832, opened 2021-02-18
Email users when they get assigned to a client

bytheway875:
Screen Shot 2021-02-17 at 5 05 52 PM

### DIFF
--- a/app/controllers/hub/tax_returns_controller.rb
+++ b/app/controllers/hub/tax_returns_controller.rb
@@ -14,8 +14,7 @@ module Hub
     def show; end
 
     def update
-      @tax_return.update!(assign_params)
-      SystemNote.create_assignment_change_note(current_user, @tax_return)
+      @tax_return.assign!(assigned_user_id: assign_params[:assigned_user_id], assigned_by: current_user)
       flash.now[:notice] = I18n.t("hub.tax_returns.update.flash_success",
                                   client_name: @tax_return.client.preferred_name,
                                   tax_year: @tax_return.year,
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -1,0 +1,18 @@
+class UserMailer < ApplicationMailer
+  default from: Rails.configuration.address_for_transactional_authentication_emails
+  helper :time
+  def assignment_email(
+    assigned_user:,
+    assigning_user:,
+    tax_return:,
+    assigned_at:
+  )
+    @assigned_user = assigned_user
+    @assigning_user = assigning_user
+    @assigned_at = assigned_at.in_time_zone(@assigned_user.timezone)
+    @client = tax_return.client
+    subject = "GetYourRefund Client ##{@client.id} Assigned to You"
+
+    mail(to: @assigned_user.email, subject: subject)
+  end
+end
--- a/app/models/tax_return.rb
+++ b/app/models/tax_return.rb
@@ -155,6 +155,21 @@ class TaxReturn < ApplicationRecord
     true
   end
 
+  def assign!(assigned_user_id: nil, assigned_by: nil)
+    if update!(assigned_user_id: assigned_user_id)
+      SystemNote.create_assignment_change_note(assigned_by, self)
+      if assigned_user_id.present?
+        assigned_user = User.find(assigned_user_id)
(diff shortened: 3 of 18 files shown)


app/controllers/hub/tax_returns_controller.rb
```diff
@@ -14,8 +14,7 @@ def edit; end
     def show; end
 
     def update
-      @tax_return.update!(assign_params)
-      SystemNote.create_assignment_change_note(current_user, @tax_return)
+      @tax_return.assign!(assigned_user_id: assign_params[:assigned_user_id], assigned_by: current_user)
```
alaroia-cfa, line 17: I find this confusing, but I see why it's there. Maybe it's peaceful to do `.merge(assigned_by: current_user)` at the end of `def assign_params`?

Not a blocker.

app/mailers/user_mailer.rb
```diff
@@ -0,0 +1,18 @@
+class UserMailer < ApplicationMailer
+  default from: Rails.configuration.address_for_transactional_authentication_emails
```
alaroia-cfa, line 2: I was hoping to only use this email address for Devise stuff, but maybe it makes sense here? Might be good to 🪑  . Not strictly a blocker.

app/mailers/user_mailer.rb
```diff
@@ -0,0 +1,18 @@
...
+  helper :time
```
alaroia-cfa, line 3: I don't quite understand the `helper :time` line (probably due my Rails naievete).

app/models/tax_return.rb
```diff
@@ -155,6 +155,21 @@ def sign_spouse!(ip)
     true
   end
 
+  def assign!(assigned_user_id: nil, assigned_by: nil)
+    if update!(assigned_user_id: assigned_user_id)
```
alaroia-cfa, line 159: I think `update!` either updates or raises an exception; perhaps you want `if update`?

Given that we always expect it to succeed, I recommend removing the `if`, keeping the `update!`, and letting it crash if it needs to.

app/views/user_mailer/assignment_email.html.erb
```diff
@@ -0,0 +1,15 @@
+<p>
```
alaroia-cfa, line 1: Great to see the emails have HTML & text parts!

spec/mailers/previews/user_mailer_preview.rb
```diff
@@ -0,0 +1,12 @@
+# Preview all emails at http://localhost:3000/rails/mailers/outbound_email_mailer
```
alaroia-cfa, line 1: I think this URL is specific to `OutboundEmailMailer` - if so, perhaps we need to change the URL here?

alaroia-cfa (approved):
Approved with comments! I'd love to 🪑 and learn more about the email from address choice and the `helper :time` invocation, but not blockers.

app/controllers/hub/tax_returns_controller.rb
bytheway875: I started with that approach, but it ended up looking a lot uglier and trickier looking in my opinion. I think its weird to pass raw params or a permitted params object into an instance method in the controller, so i was having to do **assign_params.to_h.symbolize_keys! in order to get it passed into the method correctly.

app/mailers/user_mailer.rb
bytheway875: I need things to come from a no-reply email, and thats what address_for_transactional_authentication_emails does in demo + prod. Seems like we should have an address_for_transactional_emails and not worry about having a separate email address for "authentication" emails -- maybe I'm missing some context on why we'd want something seperate just for devise emails, it doesn't seem necessary when we're ultimately sending from no-reply in demo and prod.

app/mailers/user_mailer.rb
bytheway875: Mailers don't include helpers by default, but you can include the template helpers you want to include in this way. This will load in the TimeHelper class so that I can use those helpers to format the date properly in the view.

spec/mailers/previews/user_mailer_preview.rb
bytheway875: Ah, good catch.

PR #830, opened 2021-02-17
Add Metabase docs & initial implementation

alaroia-cfa:
There are lots of docs in here and not much code. :)

@bytheway875 maybe tomorrow morning, we can work through the docs and have you create a new Metabase installation on staging, and ask any questions you have.

PR #834, opened 2021-02-18
Allow Metabase to access some data in every table

alaroia-cfa:
This will be a bit of a monster PR to review, and might merit Tuple review.

Based on #830 

## How I tested

Locally I ran:

```
rake analytics:create_views
```

Then I ran:

```
psql -d vita-min_development
```

and I ran this and got this output:

```
vita-min_development=# select * from analytics.outgoing_emails;
 id | client_id |         created_at         |          sent_at           |         updated_at         | user_id
----+-----------+----------------------------+----------------------------+----------------------------+---------
  1 |         1 | 2021-02-12 16:23:33.641902 | 2021-02-12 16:23:33.640883 | 2021-02-12 16:23:33.641902 |       4
(1 row)
```

You can do that too. It's kinda fun.

### DIFF
--- a/db/create_analytics_views.sql
+++ b/db/create_analytics_views.sql
@@ -15,51 +15,192 @@
 CREATE SCHEMA IF NOT EXISTS analytics;
 
 CREATE OR REPLACE VIEW analytics.access_logs AS
-    SELECT id, client_id, created_at, event_type, updated_at, user_id, user_agent FROM public.access_logs;
+    SELECT id, client_id, created_at, event_type, updated_at, user_id, user_agent
+    FROM public.access_logs;
 
 CREATE OR REPLACE VIEW analytics.active_storage_attachments AS
-    SELECT id, blob_id, created_at, record_id, record_type FROM public.active_storage_attachments;
+    SELECT id, blob_id, created_at, record_id, record_type
+    FROM public.active_storage_attachments;
 
 CREATE OR REPLACE VIEW analytics.active_storage_blobs AS
-    SELECT id, byte_size, checksum, content_type, created_at, key FROM public.active_storage_blobs;
+    SELECT id, byte_size, checksum, content_type, created_at, key
+    FROM public.active_storage_blobs;
 
 CREATE OR REPLACE VIEW analytics.admin_roles AS
-    SELECT id, created_at, updated_at FROM public.admin_roles;
+    SELECT id, created_at, updated_at
+    FROM public.admin_roles;
 
--- TODO: anonymized_diy_intake_csv_extracts
--- TODO: anonymized_intake_csv_extracts
--- TODO: client_success_roles
--- TODO: clients
--- TODO: coalition_lead_roles
--- TODO: coalitions
--- TODO: delayed_jobs
--- TODO: dependents
--- TODO: diy_intakes
--- TODO: documents
--- TODO: documents_requests
--- TODO: greeter_coalition_join_records
--- TODO: greeter_organization_join_records
--- TODO: greeter_roles
--- TODO: incoming_emails
--- TODO: incoming_text_messages
--- TODO: intakes
-CREATE OR REPLACE VIEW analytics.intakes as
-    SELECT id from intakes;
--- TODO: notes
--- TODO: organization_lead_roles
--- TODO: outbound_calls
--- TODO: outgoing_emails
--- TODO: outgoing_text_messages
--- TODO: provider_scrapes
--- TODO: signups
--- TODO: site_coordinator_roles
--- TODO: source_parameters
--- TODO: stimulus_triages
--- TODO: system_notes
--- TODO: tax_returns
--- TODO: team_member_roles
--- TODO: users
--- TODO: vita_partner_states
--- TODO: vita_partner_zip_codes
--- TODO: vita_partners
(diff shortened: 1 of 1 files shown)


bytheway875 (commented):
You know what would be totally rad was if there was some way to like... patch the migrations behavior so that we could automatically add a field to this file when we migrate it. But that sounds like a lot of work and i dont even know if its feasbile.

That being said, we should at least have a convo in CGC or maybe a doc that says "its a shared understanding that we add something to this file when we add a new column OR we ask product when we're adding another column whether it should be exposed to metabase.

Also, I think it might be super useful for you to do a quick overview of what SQL views are at CGC, or at least point the team in the direction of what they are. 

Also also, do you think we should be building custom views in this file, or should we just leave them as anonymized replicas of the parent table, and then do more complex queries/reports in metabase? Not having an opinion on that at this time is also fine.

db/create_analytics_views.sql
```diff
@@ -15,51 +15,192 @@
...
+
+CREATE OR REPLACE VIEW analytics.client_success_roles AS
+    SELECT id, created_at, updated_at
+    FROM public.client_success_roles;
+
+CREATE OR REPLACE VIEW analytics.clients AS
+    SELECT id, attention_needed_since, created_at, current_sign_in_at, failed_attempts, first_unanswered_incoming_interaction_at, last_incoming_interaction_at, last_internal_or_outgoing_interaction_at, last_sign_in_at, locked_at, login_requested_at, routing_method, sign_in_count, updated_at, vita_partner_id
+    FROM public.clients;
+
+CREATE OR REPLACE VIEW analytics.coalition_lead_roles AS
```
bytheway875, line 49: wondering if we can do a select.* here, or if you made a conscious choice to never do a select.* with the thinking that at some point we will have sensitive data we dont want to expose.

ALSO, is there a way we can say that we DO send all user data to metabase (non anonymized) and only anonymize client data? I think it would help to expose things like "who are the admins" without us doing a query by ssh-ing into aptible.

db/create_analytics_views.sql
```diff
@@ -15,51 +15,192 @@
...
+
+CREATE OR REPLACE VIEW analytics.documents AS
+    SELECT id, client_id, contact_record_id, contact_record_type, created_at, document_type, documents_request_id, intake_id, tax_return_id, updated_at, uploaded_by_id, uploaded_by_type
+    FROM public.documents;
+
+CREATE OR REPLACE VIEW analytics.documents_requests AS
+    SELECT id, created_at, intake_id, updated_at
+    FROM public.documents_requests;
+
+CREATE OR REPLACE VIEW analytics.greeter_coalition_join_records AS
```
bytheway875, line 77: oh wait... we actually have a greeter role? I didnt even know that. Rad.

bytheway875 (approved):

db/create_analytics_views.sql
alaroia-cfa: I did make a conscious decision to never `SELECT *` in the spirit of treating this as a whitelist.

> ALSO, is there a way we can say that we DO send all user data to metabase (non anonymized) and only anonymize client data? I think it would help to expose things like "who are the admins" without us doing a query by ssh-ing into aptible.

I think this is a good idea. It would mean we have to trust everyone with Metabase access to view that information. I propose we revisit this question in about 7 days, just to give product people used to using Metabase and have more qualitative sense of how we want to use it.

db/create_analytics_views.sql
alaroia-cfa: Yeah uh same

alaroia-cfa:
> You know what would be totally rad was if there was some way to like... patch the migrations behavior so that we could automatically add a field to this file when we migrate it. But that sounds like a lot of work and i dont even know if its feasbile.

I was thinking we could make a model method like

```
class Intake < ApplicationRecord::Base
  # normal intake stuff
  expose_to_analytics :id, :adopted_child, :already_applied_for_stimulus
end
```

We could adjust the `rake analytics:create_views` method to introspect our models and look for that kind of declaration.

I hadn't thought about adding something to migrations, but it sounds like a similar idea. I personally want the columns exposed to Metabase to be a very intentionally-maintained whitelist and never add something to this automatically.

> That being said, we should at least have a convo in CGC or maybe a doc that says "its a shared understanding that we add something to this file when we add a new column OR we ask product when we're adding another column whether it should be exposed to metabase.
>
> Also, I think it might be super useful for you to do a quick overview of what SQL views are at CGC, or at least point the team in the direction of what they are.

Agreed! I already have something on the agenda.

> Also also, do you think we should be building custom views in this file, or should we just leave them as anonymized replicas of the parent table, and then do more complex queries/reports in metabase? Not having an opinion on that at this time is also fine.

I think that custom views in this file (or similar) are a IMHO really good idea, and yet I don't want the engineering team to be spending time making them if we can avoid it.

One way to accomplish this is to suggest that our data science partner add SQL views of his liking as pull requests and perhaps introduce CI around the SQL views.

FWIW, this approach of using SQL VIEWs defined in the app repo is not a pattern I've seen in use elsewhere. I want to try it, but I have to warn you that this does make us early adopters, which means that best practices may be hard for us to find.

PR #835, opened 2021-02-18
Remove unused Zendesk data & code

alaroia-cfa:
Remove unused Zendesk data & code

There are other deletions worth noting:

* Remove the intake.anonymous field. I don't quite know what this used to do, but it is unused now.
* Remove the unused `trace` method of `ConsolidatedTraceHelper`. I removed it b/c I noticed it was unused and I had to make other changes to the file. I'm open to keeping it, though I hope we delete it since it's unused. We can always get it back later.
* Remove `spec/support/shared_examples/exceptions_with_raven_context.rb`. This file created some unused shared examples.


app/models/intake.rb
```diff
@@ -170,9 +166,6 @@
 #  created_at                                           :datetime
 #  updated_at                                           :datetime
 #  client_id                                            :bigint
-#  intake_ticket_id                                     :bigint
-#  intake_ticket_requester_id                           :bigint
-#  primary_intake_id                                    :integer
```
bengolder, line 175: I'm assuming you've checked that `primary_intake_id` is not in use anywhere.

bengolder (approved):
It's cathartic to see this cleaned up. Thanks!

app/models/intake.rb
alaroia-cfa:
```

Comment (alaroia-cfa) [reply app/models/intake.rb:175]:
```
$ git grep -l primary_intake_id
db/migrate/20200612170554_add_primary_intake_id_to_intake.rb
db/migrate/20210218014835_delete_more_zendesk_columns.rb
$
```

once added, then once
tossed away. our friend, column:
primary intake

PR #837, opened 2021-02-18
Update client filtering styles

bytheway875:
Screen Shot 2021-02-18 at 12 55 55 PM

### DIFF
--- a/app/assets/stylesheets/components/_client-filters.scss
+++ b/app/assets/stylesheets/components/_client-filters.scss
@@ -4,6 +4,22 @@
     margin-bottom: 0;
   }
 
+  /* Overrides per Anu for an even more compact honeycrisp-compact style for filter inputs */
+  .form-question {
+    margin-bottom: .5rem !important;
+  }
+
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/components/_client-filters.scss
```diff
@@ -4,6 +4,22 @@
     margin-bottom: 0;
   }
 
+  /* Overrides per Anu for an even more compact honeycrisp-compact style for filter inputs */
```
jenny-heath, line 7: i don't think we need to leave a comment about it since there are so many special styles already haha

jenny-heath (approved):

app/assets/stylesheets/components/_client-filters.scss
bytheway875: This makes some significantly custom changes, like changing the height of the selects on just this one page. I felt like that's weird and unexpected enough to document why this was added so that the person who comes after knows why we did this.

PR #840, opened 2021-02-19
SLA Breach Percentages -- Quick + Dirty implementation

bytheway875:
The new additions aren't tested, and there's lots of repeated code that could be cleaned up.

### DIFF
--- a/app/javascript/lib/metrics_table_sort.js
+++ b/app/javascript/lib/metrics_table_sort.js
@@ -3,9 +3,19 @@ import { initSortableColumn } from './table_sort';
 // Sum site-level counts into org-level values for org row.
 function setupOrgLevelCounts() {
     $(".org-metrics").each(function() {
+        let denominator = 0;
         let attention_count = 0;
         let communication_count = 0;
         let interaction_count = 0;
+
+        if ($(this).find('tr.site').length == 0) {
+            denominator += parseInt($(this).find('tr.org').first().attr('data-js-count'));
+        }
+
+        $(this).find('tr.site').each(function() {
+            denominator += parseInt($(this).attr('data-js-count'));
+        });
+
         $(this).find('tr.site td.attention-needed-breach').each(function() {
             attention_count += parseInt($(this).attr('data-js-count'));
         });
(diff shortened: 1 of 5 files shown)


app/javascript/lib/metrics_table_sort.js
```diff
@@ -3,9 +3,19 @@ import { initSortableColumn } from './table_sort';
 // Sum site-level counts into org-level values for org row.
 function setupOrgLevelCounts() {
     $(".org-metrics").each(function() {
+        let denominator = 0;
         let attention_count = 0;
         let communication_count = 0;
         let interaction_count = 0;
+
+        if ($(this).find('tr.site').length == 0) {
```
bytheway875, line 11: If there is ONLY an org and no sites, we need to use the data-js-count value from the org. Otherwise, count it from the sites.

app/javascript/lib/metrics_table_sort.js
```diff
@@ -19,16 +29,51 @@ function setupOrgLevelCounts() {
...
-        if (attention_count > 0) {
-            $(this).find('.attention-needed-breach').first().text(attention_count).attr('data-js-count', attention_count);
-        }
-        if (communication_count > 0) {
-            $(this).find('.communication-breach').first().text(communication_count).attr('data-js-count', communication_count);
-        }
-        if (interaction_count > 0) {
-            $(this).find('.interaction-breach').first().text(interaction_count).attr('data-js-count', interaction_count);
-        }
+        $(this).find('.attention-needed-breach').first().text(attention_count).attr('data-js-count', attention_count);
```
bytheway875, line 32: This is crazy repetitive and can be cleaned up, but just wanted to get it working.

app/views/hub/metrics/index.html.erb
```diff
@@ -21,79 +21,172 @@
     </div>
   </div>
 
-  <div>
+  <div class="scrollable-table-wrapper">
```
bytheway875, line 24: I feel like there is an org or users table somewhere that's not being wrapped with this class, and we should probably do that.

app/views/hub/metrics/index.html.erb
```diff
@@ -21,79 +21,172 @@
...
     <table class="index-table org-metrics-table">
       <thead class="index-table__head">
         <tr class="index-table__row">
           <th scope="col" class="index-table__header sortable" id="organization-name" width="40%">
             Organization Name
           </th>
           <th scope="col" class="index-table__header sortable" id="needs-attention-breaches">
-            Needs attention breaches<sup>1</sup>
+            Attention breaches<sup>1</sup>
```
bytheway875, line 32: I shortened the names on these to try to mitigate the need to horizontally scroll.

app/views/hub/metrics/index.html.erb
```diff
@@ -21,79 +21,172 @@
...
           </th>
+          <th scope="col" class="index-table__header" id="needs-attention-percentage">
```
bytheway875, line 34: These columns aren't sortable currently, but since I set the data-js value in the js I can probably set it up to do so. I just dont want to add too much untested functionality all at once.

app/views/hub/metrics/index.html.erb
```diff
@@ -21,79 +21,172 @@
...
+              <td class="index-table__cell interaction-breach-percentage">
+                <%= @report.active_sla_clients[site.id] %>
+
+              </td>
             </tr>
           <% end %>
         <% end %>
         </tbody>
       <% end %>
+      <% if @report.active_sla_clients[0] && current_user.admin? %>
```
bytheway875, line 139: This is funky, but essentially when there are Unassigned clients, they come back in the grouping query as id 0. Since the admin vita partners are unscoped, the unassigned clients come back and screw up the total counts if we don't include the unassigned people. I want to update this, but I thought it was more important to get the Unassigned clients represented somehow instead of leaving the math to not add up.

ALSO, It's become really clear that I can use a partial for the rendering of the columns, but that's for a later PR.

app/views/hub/metrics/index.html.erb
```diff
@@ -21,79 +21,172 @@
...
             </td>
+            <td class="index-table__cell communication-breach-percentage">
+              <%= @report.active_sla_clients_count[vita_partner.id] %>
+
+            </td>
             <td class="index-table__cell interaction-breach" data-js-count=<%= @report.interaction_breaches[vita_partner.id] %>>
               <%= @report.interaction_breaches[vita_partner.id] %>
             </td>
+            <td class="index-table__cell interaction-breach-percentage">
+              <%= @report.active_sla_clients_count[vita_partner.id] %>
```
bengolder, line 104: I see, we start out with the denominator as the text in these percentage cells

bytheway875: Yeah, we don't really need to... I'm not sure if it makes more sense to just not show anything in the case you don't have JS turned on or leave it blank.

app/javascript/lib/metrics_table_sort.js
```diff
@@ -19,16 +29,51 @@ function setupOrgLevelCounts() {
...
+    let comm_percentage, comm_percentage_text;
+    [comm_percentage, comm_percentage_text] = determineBreaches(parseInt($('.metrics-totals').find('.communication-breach').attr('data-js-count')), totalSLATracked);
+    $('.metrics-totals').find('.communication-breach-percentage').first().text(comm_percentage_text).attr('data-js-percentage', comm_percentage);
+    let interaction_percentage, interaction_percentage_text;
+    [interaction_percentage, interaction_percentage_text] = determineBreaches(parseInt($('.metrics-totals').find('.interaction-breach').attr('data-js-count')), totalSLATracked);
+    $('.metrics-totals').find('.interaction-breach-percentage').first().text(interaction_percentage_text).attr('data-js-percentage', interaction_percentage);
+}
+
+function determineBreaches(breachCount, totalCount) {
+    let percentage = isNaN(breachCount/totalCount) ? 0 : (breachCount/totalCount * 100).toFixed(2).replace(/[.,]00$/, "");
```
bengolder, line 74: this looks like it will handle a total count of zero well!

app/javascript/lib/metrics_table_sort.js
bengolder: ah, makes sense

app/views/hub/metrics/index.html.erb
bengolder: it's the users table

app/views/hub/metrics/index.html.erb
bengolder: good call

app/views/hub/metrics/index.html.erb
bengolder: oh wow, that would not have occurred to me. Good catch!

bengolder (approved):
Thank you!

PR #841, opened 2021-02-20
No hub errors for clients without email

alaroia-cfa:
This fixes a pretty common crash, where partners create drop-off clients who have no email address, but who are opted-in to receiving emails, and then they set the tax return to signature requested status, and our system tries to email the client and crashes.

See [Investigate crashes when Hub users update status of clients w/o email addresses](https://www.pivotaltracker.com/story/show/177026702)

I think that this is an adequate patch that fixes the problem, but that a better solution would be to make an object like a Service object that takes an Intake and then has two properties, `.usable_email_address` and `.usable_sms_phone_number`. We could start to use that everywhere. However, I have lingering questions about exactly how to do that, so I think the best thing is to merge this to stop the unhandled server errors.


spec/forms/hub/take_action_form_spec.rb
```diff
@@ -232,28 +240,34 @@
...
+      let(:intake) { create :intake, email_notification_opt_in: "yes", email_address: "example@example.com" }
 
       it "shows only email as a contact option" do
         expect(form.contact_method_options).to eq([{value: "email", label: "Email message"}])
       end
     end
 
     context "with a client opted-in to both email and text message" do
-      let(:intake) { create :intake, email_notification_opt_in: "yes", sms_notification_opt_in: "yes" }
+      let(:intake) { create :intake, email_notification_opt_in: "yes", email_address: "example@example.com", sms_notification_opt_in: "yes" }
```
bytheway875, line 251: Some time in the future, we should do something for text messages that is similar!

bytheway875 (approved):

PR #842, opened 2021-02-20
No crash on long email messages

alaroia-cfa:
To test this, you can run:

```
$ rails c
> message = IncomingEmail.create(client: Client.find(1), body_html: "HI " * 8000, body_plain: "HI " * 8000, from: "example@example.com", to: "example2@example.com", received_at: DateTime.now, recipient: "example3@example.com", sender: "example4@example.com")
> ClientChannel.broadcast_contact_record(message)
```

Before this change, you'll see an exception in `rails c`.

After it, you can visit http://localhost:3000/en/clients/1 and if you re-run the `rails c` code, you'll see a helpful message appear on the screen in real time.

### DIFF
--- a/app/channels/client_channel.rb
+++ b/app/channels/client_channel.rb
@@ -14,6 +14,11 @@ class ClientChannel < ApplicationCable::Channel
   end
 
   def self.broadcast_contact_record(contact_record)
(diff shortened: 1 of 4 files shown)


bytheway875 (approved):
I think we could think about a different implementation at some point (esp if we start doing more pub sub) to listen for IDs and fetch the new stuff on the subscription side.

PR #844, opened 2021-02-21
Don't count automated system messages as outgoing interactions

bengolder:
Automated messages are clearing out needs attention and needs response. Consequently, clients who complete the intake aren't being marked as needs attention, nor as needs response. At the end of intake, we set needs attention and then immediately clear it by sending an automated confirmation message.

For example, in the `FinalInfoController`
```ruby
 def after_update_success
      current_intake.update(completed_at: Time.now) # records an incoming interaction through the intake
      IntakePdfJob.perform_later(current_intake.id, "Original 13614-C with 15080.pdf")

      MixpanelService.send_event(
        event_id: current_intake.visitor_id,
        event_name: "intake_finished",
        data: MixpanelService.data_from([current_intake.client, current_intake])
      )
      send_confirmation_message # records an outgoing interaction through the outgoing email or text message
    end
```

This fixes the above example , so that  outgoing text messages and outgoing emails only record outgoing interactions if they are attributed to a user.

### DIFF
--- a/app/models/outgoing_email.rb
+++ b/app/models/outgoing_email.rb
@@ -34,7 +34,8 @@ class OutgoingEmail < ApplicationRecord
   validates_presence_of :sent_at
   has_one_attached :attachment
 
-  after_create :record_outgoing_interaction, :deliver, :broadcast
+  after_create :deliver, :broadcast
+  after_create :record_outgoing_interaction, if: ->(email) { email.user.present? }
 
   def datetime
     sent_at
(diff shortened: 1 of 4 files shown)


app/models/outgoing_email.rb
```diff
@@ -34,7 +34,8 @@ class OutgoingEmail < ApplicationRecord
   validates_presence_of :sent_at
   has_one_attached :attachment
 
-  after_create :record_outgoing_interaction, :deliver, :broadcast
+  after_create :deliver, :broadcast
+  after_create :record_outgoing_interaction, if: ->(email) { email.user.present? }
```
alaroia-cfa, line 38: I noticed it's a little bit complicated copy-pasta to remember to `if: ->(email) { email.user.present? }` here as well as similarly in text messages.

We discussed two different options while reviewing:

* Make an InteractionTrackingService so the complication is out of the model
* Make a SystemEmail model (but BG showed me that this doesn't solve all problems that a service does, so service is probably just better)

alaroia-cfa (approved):
Splendid. Thanks for catching.

PR #846, opened 2021-02-22
Sort by SLA Breach Percentage

bytheway875:
- Update JS so that you can sort by breach percentage
- Refactors and tests JS for breach percentages and sorting by numeric column
- Swaps default order so that numeric columns will show highest # breaches first on first click
- Adds "sticky" header and footer for table as recommended by Anu
- Refactors CSS to accomplish this.

### DIFF
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -209,25 +209,7 @@
   }
 
   .client-navigation {
-    margin: {
-      left: 0;
-      right: 0;
-    }
-    padding: {
-      left: 0;
-      right: 0;
-    }
-
-    @include media($tablet-up) {
-      margin: {
-        left: 0;
-        right: 0;
-      }
-      padding: {
-        left: 0;
-        right: 0;
-      }
-    }
+    @include no-bleed();
   }
 }
 
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -60,17 +60,6 @@
     left: 0;
     right: 0;
   }
-
-  @include media($tablet-up) {
-    margin: {
-      left: 0;
-      right: 0;
-    }
-    padding: {
-      left: 0;
-      right: 0;
-    }
-  }
 }
 
 // GYR customizations
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -51,18 +51,53 @@
 .background--white {
   background-color: white;
 }
+@mixin no-bleed() {
+  margin: {
+    left: 0;
+    right: 0;
+  }
+  padding: {
+    left: 0;
+    right: 0;
+  }
+
+  @include media($tablet-up) {
+    margin: {
+      left: 0;
+      right: 0;
+    }
+    padding: {
+      left: 0;
+      right: 0;
+    }
+  }
+}
 
 // GYR customizations
 .honeycrisp-compact {
+  .main-header {
+    @include no-bleed();
+    background: black;
+    height: $s60;
+    line-height: $s60;
+  }
+
   .grid {
     margin-left: 0;
     margin-right: 0;
   }
-  .slab {
-    margin-left: -3rem;
-    margin-right: -3rem;
-  }
 
+  // Override the negative margin and padding on the full-bleed classes that messes with default 100% width in browser.
+  .slab {
+    margin: {
+      left: 0;
+      right: 0;
+    }
+    padding: {
+      left: $s25;
+      right: $s25;
+    }
+  }
   .slab--padded {
     padding-top: 1.5rem;
     padding-bottom: 2rem;
@@ -181,6 +216,28 @@
     button:last-of-type {
       border-top-left-radius: 0;
       border-bottom-left-radius: 0;
+    }
+  }
+
+  .client-navigation {
+    margin: {
+      left: 0;
+      right: 0;
+    }
+    padding: {
+      left: 0;
+      right: 0;
+    }
+
+    @include media($tablet-up) {
+      margin: {
+        left: 0;
+        right: 0;
+      }
+      padding: {
+        left: 0;
+        right: 0;
+      }
     }
   }
 }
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -51,18 +51,42 @@
 .background--white {
   background-color: white;
 }
+@mixin no-bleed() {
```
bytheway875, line 54: This can be used in honeycrisp classes that normally use full-bleed() to create equal negative margin and positive padding. Neg margin, positive padding and expecting it to be 0 meant that overriding margin or padding on containers would cause the container to be wider than 100% of the parent, which caused us to put an overflow-auto on the page wrapper to "force" the page to only display at 100%. However, some css attributes - including sticky - don't work if a parent has an "overflow" attribute set.

I'm not sure why the original negative margin, + equal padding was introduced into honeycrisp, but in the hub at least I'm not seeing a negative effect from removing it.

Another note is that the honeycrisp docs insinuate that the "slab" class is primarily designed for use to create full-width sections on _marketing pages_. I wonder if it's worth reconsidering whether we should be wrapping everything in the hub with slab and overriding like this or instead create a unique container pattern for honeycrisp-compacy. But that's a bigger question for later.

app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -183,5 +207,9 @@
       border-bottom-left-radius: 0;
     }
   }
+
+  .client-navigation {
```
bytheway875, line 211: The client tabs were the component that was causing the over 100% problem because it had its _own_ negative margin, positive padding inside of the slab that was doing that, and it got wonky.

app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -51,18 +51,42 @@
...
+
+  // Override the negative margin and padding on the full-bleed classes that messes with default 100% width in browser.
   .slab {
-    margin-left: -3rem;
-    margin-right: -3rem;
+    margin: {
+      left: 0;
+      right: 0;
+    }
+    padding: {
```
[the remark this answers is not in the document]
bytheway875: use no bleed then override.

app/assets/stylesheets/_honeycrisp-compact.scss
bytheway875: check to see if this class override should be on tab-bar

app/assets/stylesheets/templates/_admin.scss
```diff
@@ -1,16 +1,14 @@
...
     margin-top: 3rem;
   }
-
   .hub-section {
     margin: 3.5rem;
   }
-
   .page-wrapper {
     padding: 0;
+    overflow: unset;
```
bytheway875, line 11: maybe make a note of why this is here or use mixin to set this so we can find to remove it later.

app/assets/stylesheets/templates/_hub-metrics.scss
```diff
@@ -13,31 +15,74 @@
...
+  white-space: nowrap;
+  border-spacing: 0 1px;
+  border-top: 1px solid black;
+  background: darken($color-grey-light, 20%);
+  max-height: 500px;
+  margin-bottom: 0;
   thead {
+    th:first-child {
+      left: 0;
+      z-index: 3;
```
bytheway875, line 29: remove

app/assets/stylesheets/templates/_hub-metrics.scss
```diff
@@ -1,5 +1,7 @@
 .hub-metrics {
   .header {
+    position: sticky;
```
bytheway875, line 3: remove

app/assets/stylesheets/templates/_hub-metrics.scss
```diff
@@ -13,31 +15,74 @@
...
+    background: white;
+  }
+
+  tr > :first-child {
+    position: sticky;
+    left: 0;
+  }
+
+  tfoot th {
+    position: sticky;
```
bytheway875, line 57: remove -- gets this from the th

app/assets/stylesheets/templates/_hub-metrics.scss
```diff
@@ -13,31 +15,74 @@
...
+      }
     }
   }
+
+  tr {
+    background: white;
+  }
+
+  tr > :first-child {
+    position: sticky;
```
bytheway875, line 52: this doesnt do anything.

bytheway875: this was when we were making the column sticky.

app/assets/stylesheets/templates/_hub-metrics.scss
```diff
@@ -13,31 +15,74 @@
...
-      }
-    }
+    background: white;
+    border-bottom: 1px solid black;
+    position:sticky;
+    top:0;
+    left: 0;
+  }
+
+  tbody {
```
bytheway875, line 75: probs dont need this either?

spec/javascript/lib/metrics_table_sort.spec.js
```diff
@@ -241,13 +242,69 @@ test("collapse and expand sites and zeros simultaneously", () => {
...
+    expect($('.org-metrics').first().attr('data-js-vita-partner-name')).toEqual("Apple Org");
+    initMetricsTableSortAndFilter();
     $('th#organization-name').click();
     expect($('.org-metrics').first().attr('data-js-vita-partner-name')).toEqual("United Way of Greater Richmond and Petersburg");
+    $('th#organization-name').click();
+    expect($('.org-metrics').first().attr('data-js-vita-partner-name')).toEqual("Apple Org");
 });
 
+test('sorting by attention breach count', () => {
+    expect($('.org-metrics').first().attr('data-js-vita-partner-name')).toEqual("Apple Org");
```
bytheway875, line 263: Consider `before(:each)` style thing for initial 2 lines

bytheway875: Jest `describe` https://jestjs.io/docs/en/setup-teardown

bytheway875 (commented):
@alaroia-cfa left comments & approves

PR #847, opened 2021-02-22
Configure JS to use webpack entirely instead of sprockets

bytheway875:
Maintaining JS coming from Sprockets and Webpack has a lot of overhead and seems potentially dangerous. For example, we run the risk of using one version of jquery from sprockets and a different version from within webpack. Currently, we're getting around that by only loading jquery into the testing environment in webpack and in production having our webpacked assets use the global jquery window object. This is _fine_, but I wanted to take a look at what it would look like to bypass sprockets for JS altogether, especially because I wanted to import JQuery-UI into the webpack setup to do some autocomplete work.

I have a PR out in Honeycrisp for the honeycrisp.js javascript.

Overall this is just a START in a direction I think GYR can move to take the lead on how we use webpacker to manage javascript in CFA applications.


Gemfile.lock
```diff
@@ -1,13 +1,13 @@
 GIT
   remote: https://github.com/codeforamerica/honeycrisp-gem
-  revision: 3f63cc61fef1ac15947b6019eb7abe61f6a5ec8a
+  revision: 320bff4e9359bf2ba9db7236c85499589b632315
```
bytheway875, line 3: Bumped the version of honeycrisp in the app now. there was a breaking change in this version that changes the default font size in honeycrisp-compact from 19 to 18, so you'll see some of those updates made in the code.

app/javascript/lib/document_submitting_indicator.js
```diff
@@ -0,0 +1,14 @@
+export var documentSubmittingIndicator = (function() {
```
bytheway875, line 1: There was some code here that was duplicating the implementation of the file uploader code that comes from honeycrisp, so I pulled out the stuff that was duplicated and let this function do what was _unique_ and not supported by the honeycrisp.js implemention, which was adding the submitting indicator properly.

app/javascript/lib/honeycrisp.js
```diff
@@ -0,0 +1,323 @@
+var incrementer = (function() {
```
bytheway875, line 1: This is a total copy paste port of the JS from honeycrisp. I've made a PR to make this an importable file from honeycrisp, but I don't know how quickly that will move: https://github.com/codeforamerica/honeycrisp-gem/pull/241

app/javascript/listeners/index.js
```diff
@@ -1,25 +1,34 @@
...
-    document.addEventListener("DOMContentLoaded", function() {
-        if (window.appData.controller_action == "Hub::Users::InvitationsController#edit") {
-            helpers.setDefaultTimezone();
-        }
-        if (window.appData.controller_action == "Hub::MessagesController#index") {
-            consumer.subscriptions.create(getChannelName(window.location.href), callback);
-        }
+const Listeners =  (function(){
+    return {
+        init: function () {
```
bytheway875, line 11: I refactored this to use the IFFE pattern with an init closure that we use a lot in other places. No real reason other than to follow a pattern that there seems to be consensus on and give some symmetry to our implementations.

Gemfile.lock
alaroia-cfa: How confident are we in this change not making the app ugly in mysterious ways? I haven't gone and validated it - would it be smart to do screenshot diffing?

alaroia-cfa: I see - it's just Honeycrisp Compact that had the breaking changes. If it doesn't affect clients, I am quite OK with breaking changes.

bytheway875: @alaroia-cfa This actually solves a mystery that Anu and I were looking into last week where like ... Honeycrisp compact was significantly more compact in the docs than on our app, and it turns out its because of these changes! So we do want them and I did spot check - albeit briefly and incompletely - that there are no breaking changes.

app/javascript/lib/ajax_mixpanel_events.js
```diff
@@ -0,0 +1,38 @@
+import "../vendor/navigator.sendbeacon.min.js";
+
+const AjaxMixpanelEvents = (function () {
```
alaroia-cfa, line 3: There are a few things that bother me aesthetically about this code.

- It's using an immediately-invoked function expression in order to create an object with an `init` function, but it would be simpler in my opinion to have `AjaxMixpanelEvents` be a function (or perhaps `initAjaxMixpanelEvents` be a function).
- It calls `$` by assuming it's available as `window.$` but I'd prefer to see it do `import $ from jquery`.
- I'd prefer to see this code _not_ hook `$(document).ready`. In that case, whoever uses `AjaxMixpanelEvents` to be responsible for calling `$(document).ready(AjaxMixpanelEvents.init)`. That will make it easier to test this code if we ever want to add tests.

I know you inherited all of that before this pull request. :) None of these are blockers. Curious what you think.

alaroia-cfa: (This comment applies to most of the JS in this pull request.)

app/javascript/lib/document_submitting_indicator.js
alaroia-cfa: Neat! Perhaps we can upstream it someday. :)

app/javascript/lib/honeycrisp.js
```diff
@@ -0,0 +1,323 @@
...
+                });
+            });
+        }
+    }
+    return {
+        init: i.init
+    }
+})();
+
+var radioSelector = (function() {
```
alaroia-cfa, line 49: This whole function is very funny.

It looks like we are mutating DOM class lists to avoid having to write CSS?? Maybe we can delete it someday by using SASS mixins more aggressively.

I know that's a honeycrisp issue, not a vita-min issue, so... whatever. :)

app/javascript/lib/ajax_mixpanel_events.js
alaroia-cfa: Seems like Ben Sheldon gives a similar comment in https://github.com/codeforamerica/honeycrisp-gem/pull/241

> One file (or more) that has the implementation code and is not self-initializing and that could be used with Webpacker. And another file for backwards compatibility/sprockets that does the self-initializing.

Not a blocker, but curious for your take.

app/javascript/lib/honeycrisp.js
alaroia-cfa: ho ho ho, it's merged :) merry christmas!

app/javascript/listeners/index.js
alaroia-cfa: Makes sense, but as I wrote earlier, I don't actually love the pattern. :) Not a blocker; symmetry is fine.

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+import RailsUJS from "@rails/ujs";
+import * as ActiveStorage from "@rails/activestorage";
+
+import "jquery-ui";
+import Listeners from "../listeners";
+import AjaxMixpanelEvents from "../lib/ajax_mixpanel_events";
+
+// Setting globally isn't very webpack-y,
+// but we have things (views, test runner) in the app that require jquery on the window object.
+window.jQuery = $;
```
alaroia-cfa, line 12: I'd love to see this be an import `import $ from jQuery`.

app/views/layouts/application.html.erb
```diff
@@ -29,8 +29,7 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= stylesheet_link_tag    'application', media: 'all' %>
-    <%= javascript_include_tag 'application' %>
```
alaroia-cfa, line 33: Removed in favor of `javascript_pack_tag` - cool

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+window.$ = $;
+
+RailsUJS.start();
+ActiveStorage.start();
+
+// *
+// IMPLEMENTATION NOTE 2/22/21
```
alaroia-cfa, line 19: Hopefully this comment can be deleted now I think?

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+import "jquery-ui";
+import Listeners from "../listeners";
+import AjaxMixpanelEvents from "../lib/ajax_mixpanel_events";
+
+// Setting globally isn't very webpack-y,
+// but we have things (views, test runner) in the app that require jquery on the window object.
+window.jQuery = $;
+window.$ = $;
+
+RailsUJS.start();
```
alaroia-cfa, line 15: In https://github.com/rails/rails/tree/main/actionview/app/assets/javascripts they seem to recommend

```
import Rails from "@rails/ujs"
Rails.start()
```

Any reason to do Rails vs. RailsUJS?

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+ActiveStorage.start();
+
+// *
+// IMPLEMENTATION NOTE 2/22/21
+// I want us to be able to import honeycrisp JS directly from honeycrisp, but the main JS we need is all included
+// right inside of the cfa_styleguide_main sprockets JS file. I'm planning on making a PR to honeycrisp to change this imminently. - SB
+//
+//import "<%#= File.join(Gem.loaded_specs['cfa-styleguide'].full_gem_path, 'app', 'assets', 'javascripts', 'honeycrisp') %>";
```
alaroia-cfa, line 23: Given that the comment is now gone, I wonder if this can now no longer be a `.js.erb` and can be a `.js`. Not a blocker.

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+
+import "../lib/honeycrisp";
+
+// *
+
+AjaxMixpanelEvents.init();
+Listeners.init();
```
alaroia-cfa, line 30: I recommend inlining `Listeners` into `application.js`.

To me, `Listeners` does the job of deciding which JS to execute given what page the user is visiting. However, `AjaxMixpanelEvents` duplicates that behavior by _always_ calling some JS initialization. Furthermore, `application.js` is now doing the job of calling both of those.

I think we can simplify.

config/webpack/environment.js
```diff
@@ -1,3 +1,40 @@
 const { environment } = require('@rails/webpacker')
+const erb = require('./loaders/erb')
+const webpack = require('webpack');
+
+environment.loaders.prepend('erb', erb)
+
+environment.plugins.prepend(
+    'Provide',
+    new webpack.ProvidePlugin({
+        $: 'jquery',
```
alaroia-cfa, line 10: I'm hopeful we can do this with an honest-to-goodness `import` instead!

config/webpack/environment.js
```diff
@@ -1,3 +1,40 @@
...
+        jQuery: 'jquery',
+        Rails: '@rails/ujs'
```
alaroia-cfa, line 12: Since we have the `import` of `Rails`, do we need this?

config/webpack/environment.js
```diff
@@ -1,3 +1,40 @@
 const { environment } = require('@rails/webpacker')
+const erb = require('./loaders/erb')
+const webpack = require('webpack');
+
+environment.loaders.prepend('erb', erb)
```
alaroia-cfa, line 5: I'm hopeful we can remove the `erb` loading.

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
+import "core-js/stable"
```
alaroia-cfa, line 1: Out of curiosity, what causes us to need this? I'm hopeful we can remove the import. Not a blocker.

app/javascript/packs/application.js.erb
```diff
@@ -0,0 +1,31 @@
...
+import "regenerator-runtime/runtime"
```
alaroia-cfa, line 2: What causes us to need regenerator? Is it due to supporting pre-Promise browsers, or something like that? If so, I think I'd like to see it loaded as a Babel transform e.g. https://babeljs.io/docs/en/babel-plugin-transform-regenerator instead. I'd love to know more. 

I'd also love to commit firmly to which browser versions we support, and then perhaps test the app with them, and certainly if we're doing something for IE11, I'd like clarity on that so I don't accidentally remove that down the road.

yarn.lock
```diff
@@ -1069,6 +1069,18 @@
   resolved "https://registry.yarnpkg.com/@rails/actioncable/-/actioncable-6.0.3-4.tgz#f17c0ef2235d1624d6052b749005608260bd55ef"
   integrity sha512-uie5McTpl69vHnXCVzKj7iYpVKFfLuE7aqul/I7MX5zVd+yVyyC/g75T1Q10GofyqHYNYWgPiIxgQIoYPyyeKw==
 
+"@rails/activestorage@^6.1.3":
```
alaroia-cfa, line 1072: We're using Rails 6.0.x - should I worry about the 6.1.3 version number here?

yarn.lock
```diff
@@ -1069,6 +1069,18 @@
...
+  version "6.1.3"
+  resolved "https://registry.yarnpkg.com/@rails/activestorage/-/activestorage-6.1.3.tgz#d76ce0fe59b5778e05b967c22c61b2964fde112a"
+  integrity sha512-9it2rc+79E+GP4GWX/P2M9/INcp1SicJaiF5yqj+k40x/JY+/eVdSAGfRCgq2G/FwkUdGlaKnG2OoDcXchczlw==
+  dependencies:
+    spark-md5 "^3.0.0"
+
+"@rails/ujs@^6.1.3":
```
alaroia-cfa, line 1079: Same question about Rails version numbers

alaroia-cfa (approved):
I think this is good! Happy to see it merged if everything is still working, which the feature tests validate already, so great.

I have some questions and suggestions for simplification. Let me know what you think.

app/javascript/packs/application.js.erb
bytheway875: honestly, it was in a tutorial of how to move from using sprockets to webpacker and I didn't think too much about it

app/javascript/lib/ajax_mixpanel_events.js
bytheway875: i agree on the whole that I'd like to shake up how we're doing a lot of this stuff. I don't want to change too much right now because my goal is to have all of the JS work exactly like it did before the change to using webpacker so that we can move forward with some new feature work we'd like to do (ex. a tested autocomplete dropdown for filtering) without having to refactor all of the existing JS to do it. 

I'm hoping that we can kind of find time to refactor as we go along. For example, now that this can be testable code, it might make sense to test this code and refactor it to make it more testable as an eng task.

PR #848, opened 2021-02-22
Remove non-utf-8 characters from filenames

alaroia-cfa:
This fixes https://sentry.io/organizations/codeforamerica/issues/2211796010/?referrer=pivotal_plugin

Only 1 occurrence, but I think it's easy enough to fix, so would love to see it merged.


bytheway875 (approved):
AWESOME!

PR #849, opened 2021-02-22
Remove ValidateRequestParams middleware

alaroia-cfa:
This fixes strange EOFErrors when clients upload docs - https://www.pivotaltracker.com/story/show/177026957

It also will fix inbound email attachments, based on my manual testing on staging.

I looked into the null-byte issue that this originally aimed to address, and I believe Rails already does a good job filtering null bytes so they don't reach Postgres, and that our code doesn't need any code beyond Rails's. Rationale: https://hackerone.com/reports/394253


bytheway875:
okay! Just make sure to review this decision with @ddefigueiredo too -- he was the one that wanted to make sure we were returning at 4** error instead of a 500 error when we had null byte errors.

alaroia-cfa:
@ddefigueiredo I've looked into this more, and concluded Rails is already checking for null bytes and ensuring we don't send bad data to Postgres. Given that, I think the 500 vs. 4xx issue is not essential. Furthermore, this filtering as implemented is corrupting `multipart/form-data` HTTP POST file uploads to the app.

I think it's best to roll forward without this middleware. If we continue to get annoying 500s, we can patch them closer to where they occur.

Let me know if you want to block, to discuss, or anything else! :)

alaroia-cfa:
@ddefigueiredo said he's OK with this, and I will roll forward with it.

PR #850, opened 2021-02-22
Validate files must be nonzero length [#176927311]

alaroia-cfa:
There are approx 192 zero-byte uploads, which surely are user error. This PR prevents them from being attached to `Document`s.

This accomplishes 3 useful things:

* Tell clients during intake their uploaded docs are invalid if they're 0-bytes long

* Tell Hub users during doc upload their uploaded docs are invalid if they're 0-bytes long

* For existing files, show Hub users that a doc _is_ 0-bytes long so they know it's a broken file

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -51,6 +51,13 @@ class Document < ApplicationRecord
   belongs_to :uploaded_by, polymorphic: true, optional: true
   has_one_attached :upload
   validates :upload, presence: true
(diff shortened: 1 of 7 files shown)


alaroia-cfa (commented):
@bytheway875 approves! Thanks for doing this.

alaroia-cfa:
Make sure to tell product about `(empty file)` and maybe tell them to newsletter it up

PR #853, opened 2021-02-22
Validate that user emails are on domains that accept email [#176721446]

alaroia-cfa:
This adds stronger validation to _User_ emails, not _Intake_ (Client) emails.

This should help us avoid situations where Mailgun rejects user invitations b/c Mailgun believes the email addresses are invalid.

We don't do this kind of strong validation on Intake/Client emails yet, but maybe we should. I propose we not worry about that, and then we start worrying about if we find it's a problem.


alaroia-cfa:
Not ready yet

PR #854, opened 2021-02-22
Signing is an incoming interaction

bengolder:


app/jobs/correct_response_needed_values_job.rb
```diff
@@ -5,7 +5,8 @@ def perform
       incoming_sms_times = client.incoming_text_messages.pluck(:received_at)
       incoming_email_times = client.incoming_emails.pluck(:received_at)
       incoming_doc_times = Document.where(uploaded_by: client).pluck(:created_at)
-      incoming_times = (incoming_sms_times + incoming_email_times + incoming_doc_times).compact
+      incoming_signature_doc_times = Document.where(document_type: DocumentTypes::CompletedForm8879.key).pluck(:created_at)
+      incoming_times = (incoming_sms_times + incoming_email_times + incoming_doc_times + incoming_signature_doc_times).compact
```
bengolder, line 9: This script backfills the "needs response" times based on historical interactions, in order to clear out false positives and more correctly represent previous client interactions. Since we will begin counting Signed 8879 forms as an incoming interaction, I'll need to include it in the set of client-initiated events.

spec/services/sign8879_service_spec.rb
```diff
@@ -37,13 +37,14 @@
...
 
-    it "creates a signed document for the tax return" do
+    it "creates a signed document for the tax return, uploaded by the client" do
       expect {
         Sign8879Service.create(tax_return)
       }.to change(tax_return.documents, :count).by 1
       new_doc = Document.last
       expect(new_doc.document_type).to eq "Form 8879 (Signed)"
       expect(new_doc.display_name).to eq "Taxpayer Signed 2019 8879"
+      expect(new_doc.uploaded_by).to eq client
```
bengolder, line 47: I assumed this was the simplest way to ensure that this document is counted as an "incoming interaction", due to this code inside of `Document.rb`

```ruby

  after_create do
    uploaded_by.is_a?(Client) ? record_incoming_interaction : record_internal_interaction

    if upload.filename.extension_without_delimiter.downcase == "heic"
      HeicToJpgJob.perform_later(id)
    end
  end
```

app/jobs/correct_response_needed_values_job.rb
```diff
@@ -5,7 +5,8 @@ def perform
       incoming_sms_times = client.incoming_text_messages.pluck(:received_at)
       incoming_email_times = client.incoming_emails.pluck(:received_at)
       incoming_doc_times = Document.where(uploaded_by: client).pluck(:created_at)
-      incoming_times = (incoming_sms_times + incoming_email_times + incoming_doc_times).compact
+      incoming_signature_doc_times = Document.where(client: client, document_type: DocumentTypes::CompletedForm8879.key).pluck(:created_at)
```
alaroia-cfa: Wouldn't be needed if we backfill first

alaroia-cfa (approved):
rad

PR #856, opened 2021-02-22
Validate that user emails are on domains that accept email [#176721446]

alaroia-cfa:
This adds stronger validation to User emails, not Intake (Client) emails.

This should help us avoid situations where Mailgun rejects user invitations b/c Mailgun believes the email addresses are invalid.

We don't do this kind of strong validation on Intake/Client emails yet, but maybe we should. I propose we not worry about that, and then we start worrying about if we find it's a problem.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -58,7 +58,7 @@ class User < ApplicationRecord
 
   before_validation :format_phone_number
   validates :phone_number, phone: true, allow_blank: true, format: { with: /\A\+1[0-9]{10}\z/ }
-  validates :email, 'valid_email_2/email': true
+  validates :email, 'valid_email_2/email': { mx: true }
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -58,7 +58,7 @@ class User < ApplicationRecord
 
   before_validation :format_phone_number
   validates :phone_number, phone: true, allow_blank: true, format: { with: /\A\+1[0-9]{10}\z/ }
-  validates :email, 'valid_email_2/email': true
+  validates :email, 'valid_email_2/email': { mx: true }
```
bytheway875, line 61: Will this break the seeds? (yes i think)

bytheway875: maybe not!

bytheway875: example.com is real

bytheway875 (approved):

bytheway875:
We should probably check to see if there are any users with invalid emails actively using the site, just in case they take some action to update themselves or something and break things.

PR #857, opened 2021-02-22
Change default ordering in client index view

bengolder:
- makes the "Waiting on Response" column visible
- defaults to ordering on that column

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -1,6 +1,6 @@
 module ClientSortable
   def filtered_and_sorted_clients(default_order: nil)
-    @default_order = default_order || { "id" => "desc" }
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/client_sortable.rb
```diff
@@ -1,6 +1,6 @@
 module ClientSortable
   def filtered_and_sorted_clients(default_order: nil)
-    @default_order = default_order || { "id" => "desc" }
+    @default_order = default_order || { "first_unanswered_incoming_interaction_at" => "asc" }
```
bytheway875, line 3: cool, i actually like this because it was kind of bothering me how the default order was literally just what the db does by default, and thus seemed a little unneccessary. But now that we are using it for something other than the default, I'm glad its there so it was straightforward to customize!

bytheway875 (approved):
LGTM! 👍

PR #859, opened 2021-02-23
Don't crash on inbound emails w/ zero byte files

alaroia-cfa:


app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -48,9 +48,10 @@ def create_incoming_email
 
         attachment = params[key]
         attachment.tempfile.seek(0) # just in case we read the same file for multiple clients
+        size = attachment.tempfile.size
```
bytheway875, line 52: This would be great in a Service Object instead of controller!

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -48,9 +48,10 @@ def create_incoming_email
...
 
         processed_attachments <<
-          if FileTypeAllowedValidator::VALID_MIME_TYPES.include? attachment.content_type
+          if (FileTypeAllowedValidator::VALID_MIME_TYPES.include? attachment.content_type) && (size > 0)
             {
                 io: attachment,
                 filename: attachment.original_filename,
```
bytheway875, line 57: Consider dealing with utf8 characters here, too.

bytheway875: Perhaps a `FileCleanerService`

bytheway875 (approved):

PR #860, opened 2021-02-23
Fix Sentry crash on VitaProviders search page

alaroia-cfa:
This uses Rails model (form) validation to fix a Sentry crash on the VitaProviders search page.

https://sentry.io/organizations/codeforamerica/issues/2206272501/?project=1880327&query=vitaproviders&sort=freq&statsPeriod=14d


alaroia-cfa:
I did this sorta largely as an exercise in getting more practice with validation. I also like the 0-exceptions Sentry status quo, I admit.

bytheway875 (approved):

PR #862, opened 2021-02-23
Ensure clients w/o names are clickable in index

bengolder:
Some of the [clients who have been waiting for weeks for a response](https://www.getyourrefund.org/en/hub/clients/sla-breaches) do not appear to have any recorded preferred name in our database. As a result, there is no way to click on them in the index  page in order to view their case. This just adds a default placeholder if the name is missing so there's something to click.


alaroia-cfa (approved):
Splendid. I would personally have made the ID clickable, but I think your solution is probably actually better.

PR #864, opened 2021-02-23
Swap out intake svgs to new versions from Anu

bytheway875:
Anu made AWESOME new SVGs, and we want to use them in the intake flow.


bytheway875:
Reviewed on Tuple w/ @embarnard

PR #865, opened 2021-02-23
Add unsure button #176807904

bytheway875:
DRAFT-- I need time to make sure im doing similar pattern everywhere


app/assets/stylesheets/components/_progress-indicator.scss
```diff
@@ -1,7 +1,6 @@
-
 .progress-indicator {
   width: 345px;
-  margin: -25px 0 25px 0;
+  margin: 0 0 25px 0;
```
bytheway875, line 3: We're making up for the negative margin by adding the "back" button above it and then giving that button a bottom margin.

app/controllers/questions/asset_sale_income_controller.rb
```diff
@@ -5,5 +5,11 @@ class AssetSaleIncomeController < QuestionsController
     def self.show?(intake)
       intake.sold_assets_yes?
     end
+
```
bytheway875, line 8: A quick overview of all of these changes is that previously we were setting the "method_name" in the view as a content_for. However, I want to use this method_name to determine whether there is an unsure method when the method is an enu and we're loading into the yes/no question layout. Moving it into the controller as a helper method allows us to continue using it in the view but also have it available at the controller level.

PR #866, opened 2021-02-23
Send signup emails

embarnard:


app/mailers/signup_followup_mailer.rb
```diff
@@ -1,5 +1,8 @@
 class SignupFollowupMailer < ApplicationMailer
-  def followup(email_address)
+  default from: Rails.configuration.address_for_transactional_authentication_emails
```
bytheway875, line 2: @alaroia-cfa How do you feel about us updating this "transactional authentication" emails configuration to just "transactional" emails? I think now we're using it in multiple other places where we need a no-reply email, and im not sure that there's a distinct reason to make the designation

app/models/signup.rb
```diff
@@ -17,15 +18,20 @@ class Signup < ApplicationRecord
   validates :phone_number, phone: true, allow_blank: true, format: { with: /\A\+1[0-9]{10}\z/ }
   validates :email_address, 'valid_email_2/email': true
 
-  def self.valid_emails
-    distinct(:email_address).pluck(:email_address).filter do |email|
-      ValidEmail2::Address.new(email).valid?
+  def self.valid_emails_with_unsent_followups
+    distinct(:email_address).where(sent_followup: false).pluck(:email_address, :name).filter do |signup_info|
```
bytheway875, line 22: I find it a bit unexpected that the return value of this query is a multi-dimensional array instead of a signup object. I wonder if we could remove the pluck and have it return signup objects. Then we can call for signup_info.email and signup_info.name in both this method and the send_followup_emails method!

app/models/signup.rb
```diff
@@ -17,15 +18,20 @@ class Signup < ApplicationRecord
...
   end
 
-  def self.send_followup_emails
-    valid_emails.each_with_index do |email, index|
-      SignupFollowupMailer.followup(email).deliver_later(wait: index * 2)
+  def self.send_followup_emails(batch_size = nil)
+    emails_to_send = batch_size.present? ? valid_emails_with_unsent_followups.slice(0, batch_size) : valid_emails_with_unsent_followups
+
+    emails_to_send.each_with_index do |signup_info, index|
+      SignupFollowupMailer.followup(signup_info[0], signup_info[1]).deliver_later(wait: index * 2)
```
bytheway875, line 31: deliver_later should enqueue these as jobs, so i dont _think_ we need to explicitly tell it to wait longer to process. But I also didn't know you could pass a "wait" and I'm interested to learn from you what it does!

db/migrate/20210222175117_add_sent_followup_to_signup.rb
```diff
@@ -0,0 +1,5 @@
+class AddSentFollowupToSignup < ActiveRecord::Migration[6.0]
+  def change
+    add_column :signups, :sent_followup, :boolean, default: false
```
bytheway875, line 3: I think its nice to create things like this as a timestamp because it does that same indication of boolean value for us but also gives us some more interesting data that we can use to track like ... how long it took someone to go from being sent the followup to signing up, or how long it took us to send all the followups out, etc.

app/mailers/signup_followup_mailer.rb
alaroia-cfa: Big +1 - let's rename it to `transactional` something or `noreply` something in the config

app/models/signup.rb
alaroia-cfa: I worked on this a little - the idea is to spread these out over time so we don't overload the queue with these low-priority tasks.

alaroia-cfa:
I left some comments since I was pinged, but I will now wander off unless pinged again. :)

PR #867, opened 2021-02-23
Only show CLIENT uploaded messages on the message upload note

bytheway875:
The implementation of the synthetic notes on the notes presenter was around since before we started keeping track of WHO uploaded a document. This only displays client actions.


bytheway875:
REviewed/approved over tuple with @embarnard !

PR #870, opened 2021-02-24
Deprecate "needs attention" replace with "needs response"

bengolder:


alaroia-cfa:
FYI @bengolder I squashed this since there was a WIP commit, plus if we need to revert it, we need to revert it as a unit.

Yippee!

PR #871, opened 2021-02-24
Use sti for system note

bytheway875:
It seems like we're needing to use system notes of different types in different places. In order to manage this, we can start by making different types of system notes set a "type" column. Using STI here made sense to me because the classes can share validations, but we can move the generator methods into their own classes. I also think the method names become more readable.


embarnard (approved):
LGTM!

PR #877, opened 2021-02-26
Fix stimulus flow

bytheway875:
We're not currently using the stimulus flow and there aren't any feature tests for the functionality so it wasn't caught that some of the changes to the questions flow (adding idk) affected the stimulus flow. Namely, the stimulus flow shares layouts with questions in the intake flow but wasn't sharing a parent (QuestionsController) in order to get the methods it needed to support an "I don't know" answer.


embarnard (approved):
LGTM!

PR #878, opened 2021-02-26
Add contact preferences to messages page [#177124054]

bytheway875:
Add a list of contact preferences to the messages page so that greeters and vita users dont need to tab back and forth to see this critical messaging-related information.

### DIFF
--- a/app/views/hub/messages/index.html.erb
+++ b/app/views/hub/messages/index.html.erb
@@ -33,7 +33,7 @@
   <% end %>
   <div class="hub-section">
     <div class="communication-preferences">
-      <% if !@client&.intake.email_address.present? && !@client&.intake.sms_phone_number.present? %>
+      <% unless @client&.intake.email_address.present? || @client&.intake.sms_phone_number.present? %>
         <%= t(".no_way_to_communicate") %>
       <% else %>
         <%= t(".preference_intro") %>
--- a/app/views/hub/messages/index.html.erb
+++ b/app/views/hub/messages/index.html.erb
@@ -32,10 +32,31 @@
     <ul class="message-list" data-js="messages-pub-sub" data-client-id="<%=@client.id%>"></ul>
   <% end %>
   <div class="hub-section">
-    <%= form_with model: @outgoing_text_message, url: hub_client_outgoing_text_messages_path(client: @client), local: true, builder: VitaMinFormBuilder, method: "post", html: {class: 'text-message-form'} do |f| %>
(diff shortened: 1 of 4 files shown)


app/views/hub/messages/index.html.erb
```diff
@@ -32,10 +32,31 @@
     <ul class="message-list" data-js="messages-pub-sub" data-client-id="<%=@client.id%>"></ul>
   <% end %>
   <div class="hub-section">
-    <%= form_with model: @outgoing_text_message, url: hub_client_outgoing_text_messages_path(client: @client), local: true, builder: VitaMinFormBuilder, method: "post", html: {class: 'text-message-form'} do |f| %>
-      <%= f.cfa_input_field :body, t(".text_message_form.label") %>
-      <%= f.hidden_field :client_id %>
-      <%= f.submit t(".text_message_form.submit"), class: "button #{"button--disabled" if @client.sms_phone_number.blank?}", disabled: @client.sms_phone_number.blank? %>
+    <div class="communication-preferences">
+      <% if !@client&.intake.email_address.present? && !@client&.intake.sms_phone_number.present? %>
```
embarnard, line 36: I think `unless @client&.intake.email_address.present? && @client&.intake.sms_phone_number.present?` would be a little more clear

spec/controllers/hub/messages_controller_spec.rb
```diff
@@ -28,6 +28,60 @@
         end
       end
 
+      context "viewing communication preferences" do
+        render_views
+        context "without way to communicate" do
+          before do
+            intake.update(sms_notification_opt_in: "no", email_notification_opt_in: "no", email_address: "", sms_phone_number: nil)
```
embarnard, line 35: why do we save blank email addresses as empty strings and blank sms phone numbers as nil?

embarnard (approved):
Looks good! Left a few comments but nothing blocking :)

spec/controllers/hub/messages_controller_spec.rb
bytheway875: It's a validation problem -- the create client form is allowing people to send in empty emails but click that they are opted into email. That's funky, but i wanted to replicate that funky behavior in the tests to make sure this catches that case.

app/views/hub/messages/index.html.erb
bytheway875: good call! thank you.

bytheway875: BUT, just a clarification that the opposite of `if !x && !y` would need to be `unless x || y`

PR #879, opened 2021-03-01
Add greeter requested info status [#176971703]

bytheway875:
This is a much-simplified version of the greeter requested info status. To avoid doing a production data migration (which might include downtime), we're just going to add it to the end of the intake statuses. This will be fine, because "Info Requested" is already at the end of the intake statuses and that's the state that Greeters are currently toggling between, so the "advancement-only" concern does not apply -- we are and need to be moving statuses forward and backward here.


alaroia-cfa (approved):
Simple and nice!

PR #880, opened 2021-03-01
Add macro templates for additional statuses

bytheway875:


app/models/intake.rb
```diff
@@ -526,7 +526,7 @@ def document_types_definitely_needed
     end
   end
 
-  def document_types_possibly_needed
+    def document_types_possibly_needed
```
alaroia-cfa, line 529: added whitespace; meh

alaroia-cfa (approved):
+1 :ship:

PR #881, opened 2021-03-02
Style consistency tweaks

bytheway875:
Based on design pairing with Anu, a PR for consistency across the app AND to stabilize some honeycrisp-compact inputs for use in a more "compact" way. 

**The change from $font-size-19 to $font-size-18 is required from the latest upgrade to honeycrisp compact, where the default font size was changed from 16px to 15px**

Significant changes include:

- Consistency in button font size + font weight: The larger font size and font-weight made buttons that sit right next to each other appear unbalanced. Made all buttons honeycrisp compact default font size with 500 font weight.
- Consistent label margin-bottom: Anu prefers a smaller gap between the label and input for a more "compact" look than honeycrisp provides
- New error style on inputs: Remove the left-hand side red bar, decrease size of error message, and add border to fields with errors. This was important to keep the elements "compact" when there are errors added, and especially important when we have multiple form elements in a row.
- Add a "hub_checkbox" form-builder element using the simplified checkboxes we use on filtering for a more compact look/feel on large forms
- Tweak form-group a bit: Putting selects inline with text areas looked funky because the margins on selects were a bit different than the margins on a text field. Updates somewhat simplify which parts of a form-group add margin to the page and ensure consistency across text boxes and selects.
Screen Shot 2021-03-01 at 5 11 40 PM
Screen Shot 2021-03-01 at 5 20 51 PM
Screen Shot 2021-03-01 at 5 21 01 PM
Screen Shot 2021-03-01 at 5 33 38 PM


embarnard (approved):
Looks good!

PR #883, opened 2021-03-02
Upgrade to Rails 6.0.3.5

alaroia-cfa:
This addresses two security vulnerabilities in Rails which do not affect us, but it's nice to resolve the warning anyway.

* https://github.com/advisories/GHSA-8hc4-xxm3-5ppp if we used an ActiveRecord feature that we do not use, then people could make our website slow
* https://github.com/advisories/GHSA-8877-prq4-9xfw if our allowed_hosts config option started with a `.` sign, which it does not, then under some circumstances, with specially crafted URLs or headers, people can make a page on our site redirect to a page of their choosing


bytheway875:
woo!

bytheway875 (approved):

bytheway875:
Are there any changes to be made in package.json to the yarn dependencies we have as a result of this?

alaroia-cfa:
@bytheway875 Good call about package.json! I updated that (and yarn.lock) and once this goes green, I'll merge. Thanks!

PR #886, opened 2021-03-03
Change ready for review macro to reviewing macro

bytheway875:
Rae requested we change the macro for "ready for review" to "reviewing" to make it more clear when the action is actually being taken.


embarnard (approved):
Looks great! 🎉

PR #888, opened 2021-03-03
Style the client login flow per Anu

alaroia-cfa:
Also:

* Use "Cell phone number" everywhere, not "Phone for texting", per Anu
* In the Client ID/SSN/ITIN last 4 phase of login, use 1 input not 2


app/controllers/portal/client_logins_controller.rb
```diff
@@ -12,25 +12,34 @@ def new
...
-          locale: I18n.locale,
-          visitor_id: visitor_id
-        ) if @form.sms_phone_number.present?
+        if @form.email_address.present?
+          ClientEmailLoginRequestJob.perform_later(
+            email_address: @form.email_address,
+            locale: I18n.locale,
+            visitor_id: visitor_id
+          )
+          session[:email_address] = @form.email_address
```
alaroia-cfa, line 21: We add this data to the session so we can show the client's email address/phone number on the page where we tell them to check their email/phone for the next step.

Note that we remove it immediately upon showing it.

app/views/portal/portal/home.html.erb
```diff
@@ -1,7 +1,7 @@
-<% content_for :title, t(".title", name: current_client.intake.primary_full_name) %>
+<% content_for :title, t(".title", name: current_client.intake.preferred_name) %>
```
alaroia-cfa, line 1: This is an unrelated change to the purpose of the pull request, but we were in the neighborhood and wanted to do it.

app/views/questions/notification_preference/edit.html.erb
```diff
@@ -24,7 +24,7 @@
         </div>
         <div class="text-input--small-label question-with-follow-up__follow-up" id="sms-opt-in">
           <div class="form-group">
-            <%= f.cfa_input_field(:sms_phone_number, t("views.questions.notification_preference.options.sms_phone_number"), type: "tel", classes: ["form-width--long"]) %>
+            <%= f.cfa_input_field(:sms_phone_number, t("general.sms_phone_number"), type: "tel", classes: ["form-width--long"]) %>
```
alaroia-cfa, line 27: This is unrelated to the purpose of the pull request, but we talked to Anu and it seemed like a good change.

app/controllers/portal/client_logins_controller.rb
```diff
@@ -12,25 +12,34 @@ def new
...
+        end
 
         redirect_to login_link_sent_portal_client_logins_path
       else
         render :new
       end
     end
 
-    def link_sent; end
+    def link_sent
```
bytheway875, line 39: This feels like a weird name for a method that's actually executing a side effect. Maybe call it "after_link_sent" to make it clear it gets called as a result of something happening?

app/forms/portal/client_login_form.rb
```diff
@@ -14,22 +14,20 @@ def client
     private
 
     def matches_client
-      if last_four.present?
+      if number.present?
```
bytheway875, line 17: Does this mean that product doesnt want to require both last 4 ssn AND confirmation number? I thought it was a both thing, not an either or thing!

app/views/portal/client_logins/link_sent.html.erb
```diff
@@ -1,9 +1,22 @@
 <% content_for :page_title, t(".title") %>
 
 <% content_for :card do %>
+  <div class="question__illustration spacing-above-0">
```
bytheway875, line 4: question__illustration class should already have top margin of 0, so i dont know if the spacing-above-0 is necessary here.

app/views/portal/client_logins/link_sent.html.erb
```diff
@@ -1,9 +1,22 @@
...
   <h1 class="h2"><%= t(".title") %></h1>
 
-  <p><%= t(".link_sent") %></p>
+  <% if @email_address.present? || @sms_phone_number.present? %>
+    <p class="spacing-below-15"><%= t(".link_sent_at") %></p>
+  <% else %>
+    <p class="spacing-below-15"><%= t(".link_sent") %></p>
+  <% end %>
+
+  <p class="spacing-below-15 text--bold"><%= @email_address %></p>
```
bytheway875, line 16: when i see the same spacing being added to a bunch of elements, to me it would make more sense to add a wrapper around this and style the p tags in css rather than with these custom classes

app/forms/portal/client_login_form.rb
bytheway875: okay i checked with them and i guess it is either or. This means we need to update the macro text i was provided that says you need both 😂

app/forms/portal/client_login_form.rb
```diff
@@ -14,22 +14,20 @@ def client
...
         @client = possible_clients.find do |client|
-          primary_match = ActiveSupport::SecurityUtils.secure_compare(client.intake&.primary_last_four_ssn.to_s, last_four)
-          spouse_match = ActiveSupport::SecurityUtils.secure_compare(client.intake&.spouse_last_four_ssn.to_s, last_four)
-          primary_match || spouse_match
+          primary_match = ActiveSupport::SecurityUtils.secure_compare(client.intake&.primary_last_four_ssn.to_s, number)
```
bytheway875, line 19: instead of checking all three, could we return as soon as we find a match?

config/locales/en.yml
```diff
@@ -771,19 +772,19 @@ en:
     client_logins:
       form:
         errors:
-          at_least_one: "Please fill out one of the two fields."
-          bad_confirmation_number: "Incorrect confirmation number. After 5 failed attempts, accounts are locked."
-          bad_last_four: "Incorrect last four. After 5 failed attempts, accounts are locked."
+          bad_input: "Incorrect client ID or last 4 of SSN/ITIN. After 5 failed attempts, accounts are locked."
       new:
         title: "To view your progress, we’ll send you a secure link."
+        one_form_of_contact: "Please enter 1 form of contact below:"
```
bytheway875, line 778: I think we generally try to go quote-less in translation strings unless we need them because of the contents of the string. So like, one form of contact: would need quotes bc of the colon but a lot of the other ones wouldn't.

bytheway875 (approved):
only nits

alaroia-cfa:
Thanks for the thorough review!

app/controllers/portal/client_logins_controller.rb
alaroia-cfa: Since it's a controller action, I kinda think the name is OK! I'd be 100% up to spend a few min brainstorming ideas for this with you.

alaroia-cfa: Personally, I don't like `after_link_sent` since it sounds like an ActiveRecord hook.

app/forms/portal/client_login_form.rb
alaroia-cfa: Great :) I assume you'll take care of that. :)

app/views/portal/client_logins/link_sent.html.erb
alaroia-cfa: Good call. I tested this, and it's not required, so I removed it in a push straight to main: https://github.com/codeforamerica/vita-min/commit/f52efddaebc8387aeec383ed341ae367c59a467a

app/views/portal/client_logins/link_sent.html.erb
alaroia-cfa: I see what you mean. I'm going to leave this as-is for now, but I may do that in the future.

app/forms/portal/client_login_form.rb
alaroia-cfa: Yes - we could return as soon as we find a match. Good call.

Originally I was thinking that it'd be nice to always check all 3 is to avoid leaking any information about _which_ one matched, if you get it right. I've been taught to be scared of [timing attacks on the web.](http://crypto.stanford.edu/~dabo/papers/webtiming.pdf)

In this case, I guess you've submitted the correct value, so now you're logged in, so in the long run, we're going to permit you to view the client's data, and therefore you're authorized to click around in the app and find out which one you matched, so it's OK to exit early.

alaroia-cfa: Pushed in https://github.com/codeforamerica/vita-min/commit/ca4aa2077462d202b36f4879f10f0509d93ed9e5

config/locales/en.yml
alaroia-cfa: It's true that that's our style. I have been ignoring this aspect of our style, hoping someone would run `i18n-tasks normalize` and fix all of them at once. :)

PR #890, opened 2021-03-04
Hide last_four_or_client_id from logs and Sentry

alaroia-cfa:
Minor code quality PR; if this is approved, please merge if possible.

### DIFF
--- a/app/forms/portal/client_login_form.rb
+++ b/app/forms/portal/client_login_form.rb
@@ -1,8 +1,8 @@
 module Portal
   class ClientLoginForm < Form
-    attr_accessor :number, :possible_clients
+    attr_accessor :last_four_or_client_id, :possible_clients
     before_validation :possible_clients_present
-    validate :number_present
+    validate :last_four_or_client_id_present
     validate :matches_client
 
     def client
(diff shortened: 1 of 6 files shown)


app/forms/portal/client_login_form.rb
```diff
@@ -1,8 +1,8 @@
 module Portal
   class ClientLoginForm < Form
-    attr_accessor :number, :possible_clients
+    attr_accessor :last_four_or_client_id, :possible_clients
```
bytheway875, line 3: woah, this is a big shift from super general to super specific. Can we split the middle and call it :access_code ?

(Also, if we want to be really specific, I think we should change this to `last_four_or_confirmation_number` so that, as previously discussed, we have less work to do all over the place when we move from using client_id to a generated confirmation number soon.

bytheway875 (changes requested):
Take this request for changes with a grain of salt, but I do feel somewhat strongly about it!

app/forms/portal/client_login_form.rb
alaroia-cfa: > woah, this is a big shift from super general to super specific. Can we split the middle and call it :access_code ?

Yes! The reason I wanted something really specific to add to `Rails.application.config.filter_parameters` so that when someone reads the list, it's easy to find the exact one thing in the codebase it refers to, plus so that if you look at the name, you think, yup, that's bad to log.

> I think we should change this to last_four_or_confirmation_number so that, as previously discussed, we have less work to do all over the place when we move from using client_id to a generated confirmation number soon.

I like the idea of saving work! When I looked at the Pivotal Tracker story about the [random 6-digit code](https://www.pivotaltracker.com/story/show/176618828), it seems to say that we'll call it _client ID_ everywhere. If not for that, I would be +1 to switching to `last_four_or_confirmation_number`.

I know you said you felt strongly about it! Let me know if the link that Tracker story changes your mind. I'm still willing to change this in the future.

alaroia-cfa:
🧂 merged for now 🧂

PR #892, opened 2021-03-04
Expire access tokens after two days

alaroia-cfa:
@bengolder and I discussed that access tokens could expire after two days. The important thing is that valid tokens not live _forever_.

The implementation here is a little messy, but I think it's messy in a good way.

I created an untested model scope called `lookup` which looks up tokens by content and also checks them for being too old. It's a weird name - `lookup` - but it also centralizes the logic, which is good in my opinion.

The service spec is where I test this model scope, in effect, which is good because I don't just want to test that the model scope _exists_; I want to test that we _use_ it; and it seemed like a lot of paperwork to unit-test it in the models, then go mock it out in the service spec.

I'm OK reworking it if a reviewer thinks it's important.

### DIFF
--- a/app/models/email_access_token.rb
+++ b/app/models/email_access_token.rb
@@ -5,6 +5,8 @@
 #  id            :bigint           not null, primary key
 #  email_address :citext           not null
 #  token         :string           not null
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
 #
 # Indexes
 #
@@ -15,8 +17,8 @@ class EmailAccessToken < ApplicationRecord
   validates_presence_of :email_address
   validate :one_or_more_valid_email_addresses
 
-  scope :by_raw_token, ->(raw_token) do
-    where(token: Devise.token_generator.digest(EmailAccessToken, :token, raw_token))
+  scope :lookup, ->(raw_token) do
+    where(token: Devise.token_generator.digest(EmailAccessToken, :token, raw_token)).where("created_at > ?", Time.current - 2.days)
   end
 
   private
--- a/app/models/text_message_access_token.rb
+++ b/app/models/text_message_access_token.rb
@@ -5,6 +5,8 @@
 #  id               :bigint           not null, primary key
 #  sms_phone_number :string           not null
 #  token            :string           not null
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
 #
 # Indexes
 #
@@ -14,7 +16,7 @@ class TextMessageAccessToken < ApplicationRecord
   validates_presence_of :token
   validates :sms_phone_number, phone: true, format: { with: /\A\+1[0-9]{10}\z/ }
 
-  scope :by_raw_token, ->(raw_token) do
-    where(token: Devise.token_generator.digest(TextMessageAccessToken, :token, raw_token))
+  scope :lookup, ->(raw_token) do
+    where(token: Devise.token_generator.digest(TextMessageAccessToken, :token, raw_token)).where("created_at > ?", Time.current - 2.days)
   end
 end
--- a/app/services/client_logins_service.rb
+++ b/app/services/client_logins_service.rb
@@ -61,11 +61,11 @@ class ClientLoginsService
 
     def clients_for_token(raw_token)
       # these might have multiple email addresses
-      to_addresses = EmailAccessToken.by_raw_token(raw_token).pluck(:email_address)
+      to_addresses = EmailAccessToken.lookup(raw_token).pluck(:email_address)
       emails = to_addresses.map { |to| to.split(",") }.flatten(1)
       email_intake_matches = Intake.where(email_address: emails)
       spouse_email_intake_matches = Intake.where(spouse_email_address: emails)
-      phone_numbers = TextMessageAccessToken.by_raw_token(raw_token).pluck(:sms_phone_number)
+      phone_numbers = TextMessageAccessToken.lookup(raw_token).pluck(:sms_phone_number)
       phone_intake_matches = Intake.where(sms_phone_number: phone_numbers)
       intake_matches = email_intake_matches.or(spouse_email_intake_matches).or(phone_intake_matches)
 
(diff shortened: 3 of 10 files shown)


bytheway875:
Have we discussed this update with product + CS? From a security standpoint its necessary, but we might want to get some input on this so that we can mitigate client confusion. (Update the macro for ready_for_signature to let the user know that access token will expire after 48 hours, etc.)

app/models/email_access_token.rb
```diff
@@ -5,6 +5,8 @@
 #  id            :bigint           not null, primary key
 #  email_address :citext           not null
 #  token         :string           not null
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
```
bytheway875, line 9: ! Thanks for adding the timestamps. We should make sure that as a rule all of our models are initialized with t.timestamps. IIRC, old versions of rails used to do this by default. Or maybe the migration gets initialized with t.timestamps already there if you name the migration in the "create_table" syntax? I can't remember.

app/models/email_access_token.rb
```diff
@@ -15,8 +17,8 @@ class EmailAccessToken < ApplicationRecord
   validates_presence_of :email_address
   validate :one_or_more_valid_email_addresses
 
-  scope :by_raw_token, ->(raw_token) do
-    where(token: Devise.token_generator.digest(EmailAccessToken, :token, raw_token))
+  scope :lookup, ->(raw_token) do
```
bytheway875, line 20: I thought that there was a "devise way" to create auto-expiring tokens, but this is also super valid and fine!

app/models/text_message_access_token.rb
```diff
@@ -14,7 +16,7 @@ class TextMessageAccessToken < ApplicationRecord
   validates_presence_of :token
   validates :sms_phone_number, phone: true, format: { with: /\A\+1[0-9]{10}\z/ }
```
bytheway875, line 17: we've got a scope in both "access_token" classes that's v similar. Wondering whether it makes sense for these classes to inherit from a parent class called AccessToken so that we can ensure similar behavior when things change?

bytheway875: Obviously, that's out of scope for this PR but just spitballing. I've not had a chance to review this area of the code yet.

app/services/client_logins_service.rb
```diff
@@ -61,11 +61,11 @@ def create_text_message_login(sms_phone_number:, visitor_id:, locale:)
 
     def clients_for_token(raw_token)
       # these might have multiple email addresses
-      to_addresses = EmailAccessToken.by_raw_token(raw_token).pluck(:email_address)
+      to_addresses = EmailAccessToken.lookup(raw_token).pluck(:email_address)
```
bytheway875, line 64: Now I'm kind of thinking would be nice to have AccessToken be an STI table so that we could do one query to look for matches on either phone numbers or emails instead of querying across multiple tables.

db/migrate/20210304015405_add_timestamps_to_access_tokens.rb
```diff
@@ -0,0 +1,10 @@
+class AddTimestampsToAccessTokens < ActiveRecord::Migration[6.0]
+  def change
+    add_timestamps :email_access_tokens, default: Time.now
+    add_timestamps :text_message_access_tokens, default: Time.now
+    change_column_default :text_message_access_tokens, :created_at, to: nil, from: Time.now
```
bytheway875, line 5: will this invalidate all current access tokens? Or not affect "existing" access tokens at all. I'm not sure how SQL treats nil values when doing comparsions like "created_at > ?", Time.current - 2 days...

Maybe we're not using these tables yet so thats why it doesn't affect them? 

Also, why do we want these default values to be nil?

app/models/email_access_token.rb
alaroia-cfa: Yes! BG and I realized this was missing and knew I'd follow up with this PR. :)

app/models/email_access_token.rb
alaroia-cfa: I was curious how Devise/devise-invitations does this, so I just checked. `devise-invitations` exposes a config option called `invite_for` which powers a method in that Gem called `valid_invitation?` (which we call via the Gem's `find_by_invitation_token` method from our InvitationsController's `require_valid_invitation_token`).

app/models/text_message_access_token.rb
alaroia-cfa: I agree it'd be good! I didn't want to add too many refactors here.

app/services/client_logins_service.rb
alaroia-cfa: That's interesting! Not a bad idea.

I also looked at [Delegated Types](https://guides.rubyonrails.org/6_1_release_notes.html#delegated-types) recently and thought they were cool.

alaroia-cfa: If they were STI, then I think we couldn't make `email_address` `NOT NULL` in the DB, and I love my database constraints.

db/migrate/20210304015405_add_timestamps_to_access_tokens.rb
alaroia-cfa: Here's the sequence of operations:

* `add_timestamps` creates columns (which are `NOT NULL` in the DB) - this _would_ fail for existing rows, except that I also set a database-level default, so existing rows get timestamps (of the time this migration runs)
* `change_column_default` removes the DB-level default

This means all existing tokens are valid for 2 days from when the migration runs.

The default value isn't really nil; it's really no default at all; that's the common pattern for these timestamp columns (check out schema.rb for other examples).

So... this was a bit of juggling to jam into one migration, but it seems to work.

alaroia-cfa:
@bytheway875 I'm merging since I didn't see you indicate your comments were blockers, but let me know if you think otherwise! Thanks for the detailed review.

alaroia-cfa:
> Have we discussed this update with product + CS? From a security standpoint its necessary, but we might want to get some input on this so that we can mitigate client confusion. (Update the macro for ready_for_signature to let the user know that access token will expire after 48 hours, etc.)

Yes! I emailed Anu and she'll take care of ensuring copy makes sense.

PR #893, opened 2021-03-04
Fix green links

bytheway875:
Use hub table styles on document page. This is what is in Anu's mocks for the docs page.


embarnard (approved):
Screen Shot 2021-03-04 at 5 06 48 PM
Nice! Looks good 👍

PR #889, opened 2021-03-03
Consent updates part 2

bytheway875:
Consent updates part 2 - create an additional consent pdf, a consents table, some factories, and consolidate logic for pdf creation through intake.

PR #884, opened 2021-03-03
Consent updates part 1

bytheway875:

PR #894, opened 2021-03-05
Consent updates part 3

jenny-heath:
_Note: somehow rebasing the base branch `consent-updates-part-1` and this branch `consent-updates-part-3` resulted in the base branch's commits becoming un-squashed? not sure what happened, but before the rebase there were only 3 commits in this PR and now there are 8_

Context: 
This is based on branch `consent-updates-part-1`, everything on which has been reviewed by me and Shannon. The reason we did not merge the first branch into main is that the individual parts of this consent mega-story cannot go into prod on their own. The first two (reviewed) PR's are #884 and #889 in case they provide helpful context.

Summary:
Add optional consent controller/form/view
- rename additional consent -> optional consent
- add optional consent page to question navigation
- add links for each type of consent that currently go to root path in absence of the actual pages
- style optional consent elements

image
image


app/forms/optional_consent_form.rb
```diff
@@ -0,0 +1,28 @@
...
+    current_time = DateTime.now
+    consent_attrs = {
+      disclose_consented_at: attributes[:disclose_consented] == "true" ? current_time : nil,
+      global_carryforward_consented_at: attributes[:global_carryforward_consented] == "true" ? current_time : nil,
+      relational_efin_consented_at: attributes[:relational_efin_consented] == "true" ? current_time : nil,
+      use_consented_at: attributes[:use_consented] == "true" ? current_time : nil,
+      user_agent: attributes[:user_agent],
+      ip: attributes[:ip]
+    }
+    intake.client.consent.present? ?
```
bytheway875, line 24: ohh smart.

bytheway875: You could also use find_or_initialize by here which _might_ save one RT to the db? Because I think you'll have to do a query for the consent to see if its present, then another to update or create. But I guess with find or initialize by, you'd just be grabbing the record and then updating or using the same record so ... actually this is probably simpler!

app/views/questions/optional_consent/edit.html.erb
```diff
@@ -0,0 +1,20 @@
+<% @main_heading = t("views.questions.optional_consent.title") %>
+
+<% content_for :form_question, @main_heading %>
+<% content_for :body_class, "consent-form" %>
+
+<% content_for :form_card do %>
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: {class: "form-card form-card--long"} do |f| %>
+    <h1 class="h2"><%= @main_heading %></h1>
+
+    <%= f.cfa_checkbox(:use_consented, t("views.questions.optional_consent.edit.consent_to_use", consent_to_use_url: root_path), options: { checked_value: true, unchecked_value: false, checked: true }) %>
```
bytheway875, line 10: Do you want this WIP root_path in here knowing you'll change this later?

spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -87,6 +87,15 @@
     select "1971", from: "Year"
     click_on "I agree"
 
+    # Optional consent form
+    expect(page).to have_selector("h1", text: "A few more things...")
```
bytheway875, line 91: Do these have_checked_fields work with just a portion of the text matching?

app/assets/stylesheets/atoms/_form-elements.scss
```diff
@@ -14,6 +14,34 @@
   }
 }
 
+.checkbox--simplified {
```
jenny-heath, line 17: this whole thing feels like it looks questionable to me but between:
a) using both the old class `.checkbox` and a new class `.checkbox--simplified` and overwriting old styles with the new class (current implementation)
and 
b) "starting from scratch" using just `.checkbox--simplified` and trying to copy the needed styles over from `.checkbox`
after playing around a bit, (a) seemed much simpler

app/assets/stylesheets/components/_reveal.scss
```diff
@@ -0,0 +1,29 @@
+.reveal {
+  color: #034E46;
+  border: 2px solid #034E46;
+  background-color: #EBFFFA;
+  border-radius: 10px;
+  padding: .2em .5em .3em .7em;
```
jenny-heath, line 6: these are copied from `.reveal__link` in honeycrisp - the design had the green box around the whole thing so i sort of just moved the styles from the child element to the parent

app/helpers/vita_min_form_builder.rb
```diff
@@ -75,6 +75,29 @@ def vita_min_select(
     html_output.html_safe
   end
 
+  def simplified_cfa_checkbox(method, label_text, options: {})
```
jenny-heath, line 78: this is identical to `CfaFormBuilder`'s `cfa_checkbox` except with the new class for styles, because you can't pass a class to `cfa_checkbox` to apply to the label. a classic example of honeycrisp not being very extendable :(

config/locales/en.yml
```diff
@@ -1552,7 +1552,7 @@ en:
         global_carryforward_html: <a href="%{global_carryforward_url}" target="_blank">Global Carryforward:</a> You allow us to make your tax return information available to other VITA programs you may visit.
         legal_details_title: The legal details
         legal_details_html: |
-          <p>Federal law requires these consent forms be provided to you. Unless authorized by law, we cannot disclose your tax return information to third parties for purposes other than the preparation and filing of your tax return without your consent. If you consent to the disclosure of your tax return information, Federal law may not protect your tax return information from further use or distribution.</p>
+          <p class="spacing-below-25">Federal law requires these consent forms be provided to you. Unless authorized by law, we cannot disclose your tax return information to third parties for purposes other than the preparation and filing of your tax return without your consent. If you consent to the disclosure of your tax return information, Federal law may not protect your tax return information from further use or distribution.</p>
```
jenny-heath, line 1555: this class looks weird i know but i couldn't figure out how else to get rid of the bottom margin on just one of the p tags within the reveal (the only one i wanted to remove the margin from wraps the reveal link, but we don't have access to it)

app/assets/stylesheets/atoms/_form-elements.scss
bytheway875: did you try using the hub_checkbox form helper we talked about the other day? I think we were going to try renaming it something like simple_checkbox or something.

app/helpers/vita_min_form_builder.rb
bytheway875: I think now we have 2 simplified checkbox versions -- maybe it makes sense to create ONE that allows us to take modifier classes.

jenny-heath:
this branch got into a pretty messy state so i made a new one and merged that instead

PR #895, opened 2021-03-05
Send surveys to clients who have completed 

bengolder:

### DIFF
--- a/app/models/tax_return.rb
+++ b/app/models/tax_return.rb
@@ -46,7 +46,7 @@ class TaxReturn < ApplicationRecord
   validates :year, presence: true
 
   attr_accessor :status_last_changed_by
-  after_update :send_mixpanel_status_change_event, :record_internal_interaction
+  after_update_commit :send_mixpanel_status_change_event, :record_internal_interaction, :send_client_completion_survey
 
   before_save do
     if status == "prep_ready_for_prep" && status_changed?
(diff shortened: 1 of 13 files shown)


app/models/tax_return.rb
```diff
@@ -189,6 +189,12 @@ def system_change_status(new_status)
     SystemNote::StatusChange.generate!(tax_return: self, old_status: self.status, new_status: new_status)
     self.status = new_status
   end
+
+  def send_client_completion_survey
+    if saved_change_to_status? && %w[file_accepted file_rejected file_mailed].include?(status)
```
alaroia-cfa, line 194: Move these into a constant called `TERMINAL_STATUSES` in TaxReturnStatus or something like that to indicate these statuses mean that the tax return is done.

spec/jobs/send_client_completion_survey_job_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+          it "sends it by text" do
+            described_class.perform_now(client)
+
+            expect(ClientMessagingService).to have_received(:send_system_text_message).with(
+              client,
+              a_string_including("qualtrics.com"),
+            )
+            expect(ClientMessagingService).not_to have_received(:send_system_email)
+            expect(client.reload.completion_survey_sent_at).to be_present
+          end
```
alaroia-cfa, line 48: Would be nice to test we only send it to one contact method if the client has both contact methods available

spec/jobs/send_client_completion_survey_job_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+          allow(ClientMessagingService).to receive(:contact_methods).and_return({email: "example@example.com"})
+          client.update(completion_survey_sent_at: DateTime.new(2021, 1, 1))
+        end
+
+        it "does not send it" do
+          expect {
+            described_class.perform_now(client)
+          }.not_to change { client.reload.completion_survey_sent_at }
+
+          expect(ClientMessagingService).not_to have_received(:send_system_email)
```
alaroia-cfa, line 65: Might be nice to assert no send text message here

alaroia-cfa (approved):
Approved!

PR #901, opened 2021-03-06
Create an access log when user views document in the hub

jenny-heath:
This is for a small fry ticket to create an access log when users view a document. We created this after doing a threat modeling exercise and realizing that documents have the most potential to contain the most sensitive information clients give to us.

Part of this story included making the client association on `access_log` polymorphic to include documents as an associated record.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_05_233848) do
+ActiveRecord::Schema.define(version: 2021_03_06_011032) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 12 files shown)


db/schema.rb
```diff
@@ -111,6 +112,19 @@
     t.index ["name"], name: "index_coalitions_on_name", unique: true
   end
 
+  create_table "consents", force: :cascade do |t|
```
bytheway875, line 115: I think you've got a dirty DB from the other ticket you were working on! Might need to drop your db, or can just manaully remove this from the schema.

bytheway875 (changes requested):
Dirty schema!

bytheway875:
Two notes: 

- This won't capture views from the edit page using the Greeter bookmarklet, so we won't really be getting full coverage access logs. 😢 
- If/when we eventually add viewing the document from the edit page, we'll need to add an access log there as well I think? Might be worth updating that ticket to reflect that.

db/schema.rb
jenny-heath: good catch!!

bytheway875:
👍

PR #902, opened 2021-03-06
Generate app walkthroughs and send to GitHub pages

alaroia-cfa:


bengolder (approved):
Looks good!

alaroia-cfa:
* If the walkthrough name has an apostrophe in it, we don't render it right.

* Also for URLs that navigate to other sites, we don't get the URL right.

* It'd be nice to have section heading anchor fragment links within each walkthrough.

PR #903, opened 2021-03-08
Add uploaded by column to client hub document tab

embarnard:


app/views/hub/documents/index.html.erb
```diff
@@ -37,6 +40,10 @@
           <td class="index-table__cell">
             <%= document.tax_return&.year %>
           </td>
+          <td class="index-table__cell">
+            <% uploaded_by_name = document.uploaded_by_type == 'User' ? document.uploaded_by&.name : document.uploaded_by&.legal_name || document.uploaded_by&.preferred_name %>
```
bytheway875, line 44: This would be really nice to put in a helper so that we can keep logic in the view pretty simple. Also, instead of putting the legal name or preferred name for the client, we might be able to get away with just putting "Client".

bytheway875: Instead of looking directly at uploaded_by_type, you could type check: `document.uploaded_by.is_a? User`. I think that would be slightly more semantically "rubyish" :D

bytheway875: Instead of looking directly at uploaded_by_type, you could type check: `document.uploaded_by.is_a? User`. I think that would be slightly more semantically "rubyish" :D

app/models/document.rb
```diff
@@ -87,6 +87,17 @@ def tax_return_belongs_to_client
     errors.add(:tax_return, I18n.t("forms.errors.tax_return_belongs_to_client")) unless tax_return.blank? || tax_return.client == client
   end
 
+  def uploaded_by_name_label
+    name_label = if uploaded_by.nil?
```
bytheway875, line 91: could also invert this and have this be the "else" statement (without even having to explicitly do the nil check -- just set it to auto-generated if it wasn't uploaded by a user or a client. But that's 6 of one half dozen of the other.

The indenting looks a little strange on this block to me... does rubymine highlight this to fix?It could also just be this view, but it seems like the elsif statements should be indented into the same place as the original if

bytheway875 (approved):
a couple tiny stylistic questions, but awesome!! The helper is great 🗡️

PR #904, opened 2021-03-08
Persist filters as cookie

bytheway875:
Refactors client filters logic to store filters in a cookie. "Clear" now clears the cookie and reloads the page.

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -1,5 +1,4 @@
 module ClientSortable
-  COOKIE_NAME = "persisted_filters"
   def filtered_and_sorted_clients(default_order: nil)
     @default_order = default_order || { "response_needed_since" => "asc" }
     setup_sortable_client unless @filters.present?
@@ -33,12 +32,20 @@ module ClientSortable
   private
 
   def setup_sortable_client
-    cookies.delete(COOKIE_NAME) if params[:clear]
+    delete_cookie if params[:clear]
     @sort_column = clients_sort_column
     @sort_order = clients_sort_order
     filter_source = has_search_and_sort_params? ? params : cookie_filters
     @filters = filters_from(filter_source)
-    cookies[COOKIE_NAME] = JSON.generate(@filters.select { |_, v| v.present? })
+    set_cookie
+  end
+
+  def delete_cookie
+    cookies.delete(filter_cookie_name) if filter_cookie_name.present?
+  end
+
+  def set_cookie
+    cookies[filter_cookie_name] = JSON.generate(@filters.select { |_, v| v.present? }) if filter_cookie_name.present?
   end
 
   def filters_from(source)
@@ -60,7 +67,9 @@ module ClientSortable
   end
 
   def cookie_filters
-    cookies[COOKIE_NAME] ? HashWithIndifferentAccess.new(JSON.parse(cookies[COOKIE_NAME])) : {}
+    return {} unless filter_cookie_name.present?
+
+    cookies[filter_cookie_name] ? HashWithIndifferentAccess.new(JSON.parse(cookies[filter_cookie_name])) : {}
   end
 
   def clients_sort_order
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -1,4 +1,5 @@
 module ClientSortable
+  COOKIE_NAME = "persisted_filters"
   def filtered_and_sorted_clients(default_order: nil)
     @default_order = default_order || { "response_needed_since" => "asc" }
     setup_sortable_client unless @filters.present?
@@ -15,7 +16,7 @@ module ClientSortable
     clients = clients.where(tax_returns: { year: @filters[:year] }) if @filters[:year].present?
     clients = clients.where(tax_returns: { status: @filters[:status] }) if @filters[:status].present?
     if @filters[:vita_partner_id].present?
-      id = @filters[:vita_partner_id]
+      id = @filters[:vita_partner_id].to_i
       clients = clients.where('vita_partners.id = ? OR vita_partners.parent_organization_id = ?', id, id)
     end
     clients = clients.where(intake: Intake.search(@filters[:search])) if @filters[:search].present?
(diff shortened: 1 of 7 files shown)


spec/controllers/concerns/client_sortable_spec.rb
```diff
@@ -123,20 +134,18 @@
...
+      before do
+        allow(cookies).to receive(:delete)
+        allow(cookies).to receive(:[]).with(anything)
+      end
+
+      it "removes the filter cookie" do
         subject.filtered_and_sorted_clients
-        expect(assigns(:filters).values.compact).to be_empty
+        expect(cookies).to have_received(:delete).with("some_filter_cookie_name")
       end
```
bengolder, line 149: Since the `it "removes..."` test validates that we remove the cookie, I'd expect to see a test in this file that verifies we set the cookie, too.

app/views/shared/_client_filters.html.erb
```diff
@@ -10,9 +10,9 @@
         <select name="status" class="select__element" id="status-filter">
           <option value></option>
           <% TaxReturnStatus.available_statuses_for(current_user).each do |stage, statuses| %>
-            <option value="<%= stage %>" <%= params[:status] == stage && "selected" %>><%= TaxReturnStatusHelper.stage_translation(stage) %></option>
+            <option value="<%= stage %>" <%= @filters[:stage] == stage && "selected" %>><%= TaxReturnStatusHelper.stage_translation(stage) %></option>
             <% statuses.each do |status| %>
-              <option value="<%= status %>" <%= params[:status] == status.to_s && "selected" %>>&emsp;<%= status_translation(status) %></option>
+              <option value="<%= status %>" <%= [status, status.to_s].include?(@filters[:status]) && "selected" %>>&emsp;<%= status_translation(status) %></option>
```
bengolder, line 15: I see we're checking for either a value (int? symbol?) or its stringified version here.

I'd rather see `@filters[:status]` always have either the value or the stringified value because I like knowing exactly what data to look for.

Not a blocker.

app/controllers/hub/unattended_clients_controller.rb
```diff
@@ -1,5 +1,6 @@
 module Hub
   class UnattendedClientsController < ApplicationController
+    FILTER_COOKIE_NAME = "sla_violations_filters".freeze
```
bengolder, line 3: One option is to compute this from the class or action name, like 
```ruby
def filter_cookie_name
  self.class.name + "_filters"
end
```

I like the possibility of computing the cookie name without adding a new public attribute to the class itself (`FILTER_COOKIE_NAME`).

If we take this approach, we can define `filter_cookie_name` in `ClientSortable`, which is nice because that's where it's consumed.

I have a general rule that I like code where the number of lines of code between a variable's definition and its use are as low as possible. The best way to win at that metric is to not to define variables, which sometimes is crazy, but in this case, seems good.

Not a blocker.

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,38 +24,52 @@ def filtered_and_sorted_clients(default_order: nil)
 
   # see if there are any overlapping keys in the provided params and search/sort set
   def has_search_and_sort_params?
-    (params.keys.map(&:to_sym) & search_and_sort_params).any?
+    overlapping_keys = (params.keys.map(&:to_sym) & search_and_sort_params)
+    hash_params = params.try(:to_unsafe_h) || params
```
bengolder, line 28: I'd like to see a test covering these changes. From looking at it, I don't immediately understand the purpose of `params.try(:to_unsafe_h)`. Also, that sounds scary!

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,38 +24,52 @@ def filtered_and_sorted_clients(default_order: nil)
...
+    filter_source = has_search_and_sort_params? ? params : cookie_filters
+    @filters = filters_from(filter_source)
+    set_cookie
+  end
+
+  def delete_cookie
+    cookies.delete(filter_cookie_name) if filter_cookie_name.present?
+  end
 
+  def set_cookie
```
bengolder, line 47: If possible, I'd like to see a unit spec that covers this since it's somewhat complicated. I see it's also covered in the feature spec.

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,38 +24,52 @@ def filtered_and_sorted_clients(default_order: nil)
...
+    overlapping_keys = (params.keys.map(&:to_sym) & search_and_sort_params)
+    hash_params = params.try(:to_unsafe_h) || params
+    overlapping_keys.any? && hash_params.slice(*overlapping_keys).any? { |_, v| v.present? }
   end
 
   private
 
   def setup_sortable_client
-    reset_filter_params if params[:clear]
+    delete_cookie if params[:clear]
```
bengolder, line 35: I'd love to see a unit spec for these cookie-related changes , perhaps in a `"with filter cookies"` context, calling  `fitered_and_sorted_clients`.

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,38 +24,52 @@ def filtered_and_sorted_clients(default_order: nil)
...
-      stage: stage_filter,
-      assigned_to_me: params[:assigned_to_me],
-      unassigned: params[:unassigned],
-      needs_response: params[:needs_response],
-      year: params[:year],
-      vita_partner_id: params[:vita_partner_id].present? ? params[:vita_partner_id].to_i : nil,
-      assigned_user_id: params[:assigned_user_id].present? ? params[:assigned_user_id].to_i : nil,
+    filter_source = has_search_and_sort_params? ? params : cookie_filters
+    @filters = filters_from(filter_source)
+    set_cookie
```
bengolder, line 40: This is implemented using cookies, but I'd prefer to see it implemented using our existing session. This has a few advantages in my opinion:

* The session is cleared when users log out. This means that if two users use the same device after each other, their filters won't persist across users. Mostly this is important to me aesthetically! It does have a minor practical implication - for example, if the first user searches for a name/email and then the second user visits the same area in the site, the second user will see a name/email they weren't supposed to know was a client.

* The session uses Rails for checking authenticity (it's stored with signed & encrypted cookies). There's nothing I see here that really requires that level of authenticity checking, but it's nice to lean on when it's there imho.

I know that leaning on the session means that the data will get lost if the Hub user is inactive for 30 minutes. I think that's OK for now - I _think_ that most of the need here is when people click a client from a filtered Clients view, then see the same filter a few minutes later when navigating around the site.

Related - if that is the only need, do you think that adjusting the `back_to` behavior would meet the user need, too? That is, when you click on a Client, there's a "<< back to All Clients" link at the top of the Client page; that could be adjusted to include the query params of the filter that let the user navigate to this client. It would be more work in some ways, but I do personally being explicit about state management. It would also ensure one browser tab can't interfere with another; if you're trying to do two filters in two different browser tabs, the session/cookie approach will result in them clobbering each other, which I personally don't like. (Not a blocker.)

Another option is to consider storing these filters in the database. If our users really want this to persist, then maybe that's better!

If we want the transient approach, I do have a strong preference for session over cookie here. Let me know what you think.

app/controllers/concerns/client_sortable.rb
bengolder: I think this would be easier to read with a explicit if.

```
if overlapping_keys.any?
  params.require(*overlapping_keys)
end
```

Which I guess is the same as

`params.require(*overlapping_keys) if overlapping_keys.any?`

I'm not 100% sure that my proposed use of `require(*...)` is the same as your use of `.slice`. I am fearful when I see `:to_unsafe_h` so I'm hopeful there's a way to avoid it.

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,38 +24,52 @@ def filtered_and_sorted_clients(default_order: nil)
 
   # see if there are any overlapping keys in the provided params and search/sort set
   def has_search_and_sort_params?
-    (params.keys.map(&:to_sym) & search_and_sort_params).any?
+    overlapping_keys = (params.keys.map(&:to_sym) & search_and_sort_params)
```
bengolder, line 27: 100% not a blocker: I hope when we upgrade to Ruby 2.7, we can switch this to `.intersection` so I can forget what `&` means in this context. :)

bengolder, line 27: You could drop the `(` and `)` on this line I suppose.

bengolder (approved):
I'd love to see unit test coverage of _writing_ the cookie, and personally I'd like to see this data storage move from out of a cookie and into our normal session. See also the query params ideas & database ideas in a comment.

Let me know what you think. No blockers.

Glad to see us responding to the user needs of keeping their filters! Happy to re-review quickly if you make big changes and want a fresh review.

(Paired with @alaroia-cfa on this review.)

app/controllers/concerns/client_sortable.rb
bytheway875: The session cookie has a limited length, so you're at real risk of cramming too much into the session if we were to save a hash of filters for each dashboard into the filters -- I believe based on code I looked at that this might have happened last season in the session cookie on the intake side of things? 

For this reason, I think its important to be super discerning of what's tied to the session and what can be managed using a regular cookie. I think the risks here off overloading the session cookie outweigh some of the funkiness with storing in a cookie. 

I do think that a better way to do this would be to use localstorage and do this in JS, but based on the current implementation of the filters a cookie was more straightforward -- we can set cookies on the server, but can only set/retrieve localstorage in js. sometimes you have to choose the most optimal solution based on the time constraints and limitations of the original implementation, so using a cookie felt more straightforward than figuring out the best way/place to save into localStorage.

I think this filtering and sorting logic is a really high-touch area of the code and will likely continue to be refactored based on changing needs, so my thought is to get something out the door and then we revise based on ongoing feedback.

bytheway875:
im going to merge this so i can let product start acceptance, but will come back and add a test for setting of the cookie.

PR #905, opened 2021-03-08
Fix bug that sets message_body to take action form when there are no contact options

bytheway875:
message_body was getting _set_ on the form even if weren't sending a message because there is no contact info. 

Adds logic that does not set the message body from status macro if there is no way to contact the user, which ensures that validations on that "hidden" message don't run.

### DIFF
--- a/app/forms/hub/take_action_form.rb
+++ b/app/forms/hub/take_action_form.rb
@@ -22,7 +22,6 @@ module Hub
       @tax_returns = client.tax_returns.order(year: :asc)
       @action_list = []
       super(*args, **attributes)
-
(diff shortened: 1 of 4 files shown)


app/forms/hub/take_action_form.rb
```diff
@@ -73,7 +72,7 @@ def language_label(key)
     end
 
     def set_default_message_body
-      @message_body = "" and return unless status.present?
+      @message_body = "" and return unless status.present? && contact_method_options.present?
```
embarnard, line 75: Good catch!

embarnard (approved):
Looks good! Approved

PR #907, opened 2021-03-08
Open all MFT links in new tab

jenny-heath:


bytheway875 (approved):
obvi you do not need to do this as part of this, but since we're doing this same thing in a lot of places we could create like a link_to_target_blank helper that handles putting the link in the "right" state

PR #908, opened 2021-03-09
Sign ALL Unsigned 8879s when you sign for a given tax year

bytheway875:
We used to only allow you to have one 8879 and sign one. But some places have state and federal 8879s, some sites are having to reupload 8879s and get the taxpayer to re-sign after adjustments are made, etc.

Also, it was requested that instead of keeping the unsigned version of the 8879, we convert it into a Signed 8879.

### DIFF
--- a/app/views/portal/portal/home.html.erb
+++ b/app/views/portal/portal/home.html.erb
@@ -25,14 +25,32 @@
           <% end %>
         <% end %>
 
-        <% if tax_return.signed_8879.present? %>
-          <div class="tax-return-status__solo">
-            <%= link_to t(".tax_returns.view_signed_8879"), portal_document_path(id: tax_return.signed_8879.id), target: "_blank" %>
-          </div>
-        <% elsif tax_return.unsigned_8879.present? %>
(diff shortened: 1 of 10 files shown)


app/views/portal/portal/home.html.erb
```diff
@@ -25,14 +25,32 @@
...
-        <% if tax_return.signed_8879.present? %>
-          <div class="tax-return-status__solo">
-            <%= link_to t(".tax_returns.view_signed_8879"), portal_document_path(id: tax_return.signed_8879.id), target: "_blank" %>
-          </div>
-        <% elsif tax_return.unsigned_8879.present? %>
-          <div class="tax-return-status__row tax-return-status__solo">
-            <%= link_to t(".tax_returns.view_unsigned_8879"), portal_document_path(id: tax_return.unsigned_8879.id), target: "_blank" %>
-          </div>
+        <% if tax_return.signed_8879s.present? %>
+          <% if tax_return.signed_8879s.count == 1 %>
```
jenny-heath, line 29: can use i18n pluralization

jenny-heath (approved):

PR #909, opened 2021-03-09
Always scroll to last note or message

bengolder:
From design pairing, this changes the navigation for messages and notes index pages to always scroll to the last note or message in the index. It adds an id to the last note or message, and then uses a URI anchor (`/messages#last-item`) to navigate to that element.

[Pivotal Tracker Story](https://www.pivotaltracker.com/story/show/177149167)

I paired on this with Anu

### DIFF
--- a/app/views/hub/messages/index.html.erb
+++ b/app/views/hub/messages/index.html.erb
@@ -11,12 +11,15 @@
   <% if @messages_by_day.present? %>
     <ul class="message-list" aria-live="polite" data-js="messages-pub-sub" data-client-id="<%=@client.id%>">
 
-      <% @messages_by_day.each do |datetime, contact_records| %>
-
+      <% @messages_by_day.each_with_index do |(datetime, contact_records), day_index| %>
         <li class="message__day-heading"><%= date_heading(datetime) %></li>
 
-        <% contact_records.each do |contact_record| %>
-          <%= render "shared/message_list_contact_record", contact_record: contact_record %>
+        <% contact_records.each_with_index do |contact_record, contact_record_index| %>
+          <% if contact_record_index == (contact_records.size - 1) && day_index == (@messages_by_day.size - 1) %>
+            <%= render "shared/message_list_contact_record", contact_record: contact_record, is_last_item: true %>
+          <% else %>
+            <%= render "shared/message_list_contact_record", contact_record: contact_record %>
+          <% end %>
         <% end %>
 
       <% end %> <%# end of day %>
@@ -31,7 +34,8 @@
     %>
     <ul class="message-list" data-js="messages-pub-sub" data-client-id="<%=@client.id%>"></ul>
   <% end %>
-  <div class="slab">
+
+  <div class="slab" id="composer">
     <div class="communication-preferences">
       <% unless @client&.intake.email_address.present? || @client&.intake.sms_phone_number.present? %>
         <%= t(".no_way_to_communicate") %>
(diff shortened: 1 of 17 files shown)


spec/controllers/hub/outbound_calls_controller_spec.rb
```diff
@@ -58,10 +58,11 @@
     context "with a logged in user" do
       before { sign_in user }
 
-      it "updates the outbound call with the note body" do
+      it "updates the outbound call with the note body and redirects back to messages" do
         put :update, params: params
 
         expect(outbound_call.reload.note).to eq "I talked to them!"
+        expect(response).to redirect_to(hub_client_messages_path(client_id: client.id, anchor: "last-item"))
```
bengolder, line 65: we didn't previously have an expectation about where the user would be redirected after submitting an outbound phone call form, but we were redirecting to the notes tab. This adds the expectation and corrects it to redirect to messages instead, where we now display outbound calls.

app/views/hub/messages/index.html.erb
```diff
@@ -11,12 +11,15 @@
   <% if @messages_by_day.present? %>
     <ul class="message-list" aria-live="polite" data-js="messages-pub-sub" data-client-id="<%=@client.id%>">
 
-      <% @messages_by_day.each do |datetime, contact_records| %>
-
+      <% @messages_by_day.each_with_index do |(datetime, contact_records), day_index| %>
```
bengolder, line 14: I'd never used this syntax before, but encountered it here and thought it worked okay: https://stackoverflow.com/a/2083597/399726

app/views/hub/notes/_note.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="note message message--<%= note.user == current_user ? "outgoing" : "incoming"%>">
+<li class="note message message--<%= note.user == current_user ? "outgoing" : "incoming"%>"<%= " id=\"last-item\"".html_safe if local_assigns[:is_last_item] %>>
```
alaroia-cfa, line 1: Consider `id="<%= "last-item" if ... %>"`

alaroia-cfa (approved):
Splendid! One suggestion for how to remove `.html_safe` :D

app/views/hub/notes/_note.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="note message message--<%= note.user == current_user ? "outgoing" : "incoming"%>">
+<li class="note message message--<%= note.user == current_user ? "outgoing" : "incoming"%>" id="<%= "last-item" if local_assigns[:is_last_item] %>">
```
bytheway875, line 1: is it necessary to use local_assigns in the partial? can we check based on `is_last_item`'s presence directly. Maybe local_assigns is necessary since is_last_item is an optional parameter and may be undefined?

bytheway875:
This is more product feedback than development feedback, but as a feature this kind of concerns me. Will this be clear enough to the user that we're scrolling to the _last_ message, or is there potential that they assume they're being scrolled to the _most recent message they should care about_? Will the user miss potentially necessary context because of the fact that they're potentially not scrolling up to see additional questions the client might have sent?

app/views/hub/notes/_note.html.erb
bengolder: > Maybe local_assigns is necessary since is_last_item is an optional parameter and may be undefined?

I think so. This was my conclusion because it would be undefined unless passed. But honestly I don't like it that much. I just couldn't come up with a better solution. I looked for other options and this was the best I could find in a short amount of time.

bengolder:
@bytheway875 Thanks for voicing that. I think that's a valid concern. I agree there is a risk of not seeing everything worth reading on page load. That being said, I'm hopeful this is still an incremental improvement over the older situation of staying at the top of the page without recent messages visible.
I'll go ahead and relay that feedback to Yvonne & Anu, unless you'd rather do that yourself.

app/views/hub/notes/_note.html.erb
bytheway875: @bengolder I think the pattern I've used before for optional params is to set a "default value" in the partial using erb, where it would be `<%= some_variable ||= false %>`. I don't know that that's any better, but I also don't know that I knew you could retrieve from local_assigns.

PR #910, opened 2021-03-09
Don't list independent orgs in every coalition

bengolder:
Independent orgs are orgs that don't belong to any coalition.

Right now, we are listing them repeatedly after every coalition on the org index page:
https://www.getyourrefund.org/hub/organizations


bytheway875:
👍

bytheway875 (approved):

PR #914, opened 2021-03-10
Record when 'needs response' toggle is clicked

bengolder:


app/controllers/hub/clients_controller.rb
```diff
@@ -56,8 +56,15 @@ def update
     end
 
     def response_needed
-      @client.clear_response_needed if params.fetch(:client, {})[:action] == "clear"
-      @client.set_response_needed! if params.fetch(:client, {})[:action] == "set"
+      case params.fetch(:client, {})[:action]
```
jenny-heath, line 59: whoa what is `params.fetch(:client, {})`? specifically why the empty hash what does it do

bengolder: Good question! that was existing code that I did not reexamine.  It appears to try to grab client params, and defaults to an empty hash if not available. However, this should probably be utilizing the standard secure parameters pattern, where we have a `response_needed` method that calls `params.require(:client).permit(:action)`

jenny-heath (approved):
discussed over tuple and decided this can go out after changing 2 things:
- fetching params in the usual secure rails way
- adding type to the system note

bengolder:
@jenny-heath it turns out that the type column is automagically populated by rails using the class name of the record.

PR #915, opened 2021-03-10
Tighten styles for pagination and client filters for consistency

bytheway875:
BEFORE
Screen Shot 2021-03-10 at 9 06 06 AM

AFTER
Screen Shot 2021-03-10 at 9 07 57 AM

### DIFF
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -96,7 +96,7 @@
   }
   .slab--padded {
     padding-top: 1.5rem;
-    padding-bottom: 2rem;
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -96,7 +96,7 @@
   }
   .slab--padded {
```
alaroia-cfa, line 97: I think this is actually what comes from honeycrisp compact right now.

alaroia-cfa (approved):
Looks great! You had some ideas for improvements -- fine with me! Consider also the rename suggestion.

PR #916, opened 2021-03-10
Tweak styling on tax returns list to fix some visual bugginess

bytheway875:
revert some styling to use flex box conventions, and deal (as elegantly as possible) with forcing assignment dropdowns to be a useable size. (Previously on prod, I was seeing some assignment components that didn't have enough width to show up correctly. 

Also removes HSA from certification display per product / user research teams.

### DIFF
--- a/app/assets/stylesheets/components/_tax-return-list.scss
+++ b/app/assets/stylesheets/components/_tax-return-list.scss
@@ -1,13 +1,12 @@
 .tax-return-list {
-  font-size: 1.6rem;
+  font-size: $font-size-18;
   margin: 0;
-
   a {
     color: black;
   }
 
   .certification-label {
-    width: 80px;
+    width: 60px;
     text-align: center;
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/components/_tax-return-list.scss
```diff
@@ -1,13 +1,12 @@
 .tax-return-list {
-  font-size: 1.6rem;
+  font-size: $font-size-18;
```
alaroia-cfa, line 2: test for removal?

alaroia-cfa:
Before

I see there's the negative top placement; maybe there's a way to simplify and remove that.

image

alaroia-cfa (changes requested):
It'd be nice to fix the new weird alignment per the screenshots. Overall these are good changes, and I like that we can rely on flexbox more and more.

PR #918, opened 2021-03-10
Scenic views -- organization capacity

bytheway875:


app/models/organization_capacity.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# Table name: organization_capacities
+#
+#  active_client_count :bigint
+#  capacity_limit      :integer
+#  name                :string
+#  vita_partner_id     :bigint
+#
+class OrganizationCapacity < ApplicationRecord
+  self.primary_key = :vita_partner_id # vita partner id
```
alaroia-cfa, line 11: Consider renaming the column in the VIEW so it's `id` and then we can ditch this line. (Then again, if you add the `belongs_to` below, it's actually nice that this is `vita_partner_id`.)

app/models/organization_capacity.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  # Prevents us from calling #save on a view,
```
alaroia-cfa, line 13: Completely trivial, but consider making this two-line comment into one line. :)

db/schema.rb
```diff
@@ -728,4 +728,34 @@
   add_foreign_key "vita_partner_zip_codes", "vita_partners"
   add_foreign_key "vita_partners", "coalitions"
   add_foreign_key "vita_providers", "provider_scrapes", column: "last_scrape_id"
+
+  create_view "organization_capacities", sql_definition: <<-SQL
+      WITH organization_id_by_vita_partner_id AS (
+           SELECT vita_partners_1.id,
```
alaroia-cfa, line 734: Do you need to rename the table as `vita_partners_1` rather than `vita_partners`? If not, it'd be nice to use the table name.

db/migrate/20210310205504_create_organization_capacities.rb
```diff
@@ -0,0 +1,5 @@
+class CreateOrganizationCapacities < ActiveRecord::Migration[6.0]
+  def change
```
alaroia-cfa, line 2: This is nice syntax, and honestly maybe the `rake analytics:create_views` system should use this! Not a blocker at all.

db/schema.rb
```diff
@@ -728,4 +728,34 @@
...
+                  CASE
+                      WHEN (vita_partners_1.parent_organization_id IS NULL) THEN vita_partners_1.id
+                      ELSE vita_partners_1.parent_organization_id
+                  END AS organization_id
+             FROM vita_partners vita_partners_1
+          ), client_ids AS (
+           SELECT DISTINCT tax_returns.client_id
+             FROM tax_returns
+            WHERE ((tax_returns.status > 100) AND (tax_returns.status < 405))
```
alaroia-cfa, line 743: If you're sure this is correct, then great. This will include greeter_info_requested, which surprises me.

db/schema.rb
```diff
@@ -728,4 +728,34 @@
...
+
+  create_view "organization_capacities", sql_definition: <<-SQL
+      WITH organization_id_by_vita_partner_id AS (
+           SELECT vita_partners_1.id,
+                  CASE
+                      WHEN (vita_partners_1.parent_organization_id IS NULL) THEN vita_partners_1.id
+                      ELSE vita_partners_1.parent_organization_id
+                  END AS organization_id
+             FROM vita_partners vita_partners_1
+          ), client_ids AS (
```
alaroia-cfa, line 740: Trivial, but I think it's more readable if the `)` and `client_ids` are on separate lines.

db/views/organization_capacities_v01.sql
```diff
@@ -0,0 +1,20 @@
+WITH
```
alaroia-cfa, line 1: How come you need both the `schema.rb` change and this file? Is this basically the meat of the migration?

app/models/organization_capacity.rb
alaroia-cfa: Consider adding Rails `belongs_to :vita_partner` so you can do `OrganizationCapacity.where(vita_partner: { name: "Oregon Helpers" })` or whatever.

db/schema.rb
```diff
@@ -728,4 +728,34 @@
...
+           SELECT DISTINCT tax_returns.client_id
+             FROM tax_returns
+            WHERE ((tax_returns.status > 100) AND (tax_returns.status < 405))
+          ), partner_and_client_count AS (
```
alaroia-cfa, line 744: Trivial, but now that I think about it, in Rails, table names are usually plural, so consider `partner_and_client_counts`. Also consider `client_count_by_organization_id`. I love using `x_by_y` in names.

db/views/organization_capacities_v01.sql
alaroia-cfa: I'm curious - does `rails db:rollback` work? If so, that's pretty cool.

alaroia-cfa: How does Scenic know it's v01 that you need to migrate to? Does it add a database table for tracking migrations on views? Asking earnestly.

db/schema.rb
```diff
@@ -728,4 +728,34 @@
   add_foreign_key "vita_partner_zip_codes", "vita_partners"
   add_foreign_key "vita_partners", "coalitions"
   add_foreign_key "vita_providers", "provider_scrapes", column: "last_scrape_id"
+
+  create_view "organization_capacities", sql_definition: <<-SQL
```
alaroia-cfa, line 732: Because we're using Scenic, we have SQL in a string in a Ruby file. By contrast, if we were using the [SQL format for the Rails schema,](https://blog.appsignal.com/2020/01/15/the-pros-and-cons-of-using-structure-sql-in-your-ruby-on-rails-application.html), then we'd have SQL in a file. My instinct is to prefer SQL in a file over SQL in a string in Ruby.

Having said that, I think my instinct doesn't apply here. Thanks for implementing this with Scenic instead. It seems to me that schema.rb has better support for handling/preventing merge conflicts. (Now, if you _do_ find a way to make merge conflicts easy to avoid/prevent/handle with `structure.sql`, then my opinion might change!)

alaroia-cfa (approved):
Stellar. I have a few non-blocking questions and comments; would be happy to learn answers or see relevant trivial revisions.

app/models/organization_capacity.rb
alaroia-cfa: Perhaps the inverse, `has_one :organization_capacity` might be cool on `class VitaPartner`.

db/views/organization_capacities_v01.sql
bytheway875: when you call rake scenic:migration something something, it creates this file and the migration file. Then we write the sql in here, and it does some magic to add it to the schema (i think that's where vita_partners_1 et al come from). YES! I did rollbacks and fowards several times! 

You can check out the gem documentation for all the cool stuff it can do -- you can create a new version of an existing view or update the same version directly. It also has support for materialized views.

db/schema.rb
bytheway875: i believe that is accurate right now in the same way "not ready" is included in the counts. I'll double check with product and partner though and reflect capacity queries here and the const in tax_return_status if necessary.

app/models/organization_capacity.rb
bytheway875: done!

db/schema.rb
bytheway875: in our raw .sql file that's how i have it formatted but i think that scenic turns that into this and i dont have control over it -- i think its better not to touch scenic's generated sql and/or edit the schema file directly!

app/models/vita_partner.rb
```diff
@@ -50,12 +51,9 @@ class VitaPartner < ApplicationRecord
...
     return parent_organization.at_capacity? if site?
 
-    return false if capacity_limit.blank?
+    return false if capacity_limit.blank? || organization_capacity.active_client_count.nil?
 
-    Client
-      .where(vita_partner_id: [id, *child_site_ids])
-      .joins(:tax_returns).where(tax_returns: { status: TaxReturnStatus::STATUS_KEYS_INCLUDED_IN_CAPACITY })
-      .count >= capacity_limit
+    return organization_capacity.active_client_count >= capacity_limit
```
bytheway875, line 56: Replacing this query with the result of our view! Had a discussion with Ben Sheldon about how active record is able to "drop down" the query for this one vita partner into the view code so that it only queries for the vita partner we want. super cool!

app/lib/tax_return_status.rb
```diff
@@ -46,11 +46,15 @@ def message_templates
...
 
-  STATUS_KEYS_INCLUDED_IN_CAPACITY = STATUSES.keys - [:intake_before_consent, :intake_in_progress, :file_accepted, :file_not_filing].freeze
+
+  ONBOARDING_STATUSES = [:intake_before_consent, :intake_in_progress, :intake_greeter_info_requested]
   STATUS_KEYS_INCLUDED_IN_SLA = STATUSES.keys - [:intake_before_consent, :file_accepted, :file_not_filing].freeze
   STATUSES_BY_STAGE = determine_statuses_by_stage.freeze
   STAGES = STATUSES_BY_STAGE.keys.freeze
   TERMINAL_STATUSES = [:file_accepted, :file_rejected, :file_mailed].freeze
+  # If you change the statuses included in capacity, please also update the organization capacities sql view
+  # status > 101 AND status < 403 AND status != 106 AND status != 404
```
alaroia-cfa, line 56: > AND status != 404

This made me smile!

app/models/organization_capacity.rb
```diff
@@ -0,0 +1,18 @@
...
+# Table name: organization_capacities
+#
+#  active_client_count :bigint
+#  capacity_limit      :integer
+#  name                :string
+#  vita_partner_id     :bigint           primary key
+#
+class OrganizationCapacity < ApplicationRecord
+  belongs_to :vita_partner
+  self.primary_key = :vita_partner_id # vita partner id
```
alaroia-cfa, line 12: I think the comment isn't necessary, but also that's fine to have!

app/models/vita_partner.rb
```diff
@@ -50,12 +51,9 @@ class VitaPartner < ApplicationRecord
   def at_capacity?
     return parent_organization.at_capacity? if site?
 
-    return false if capacity_limit.blank?
+    return false if capacity_limit.blank? || organization_capacity.active_client_count.nil?
```
alaroia-cfa, line 54: Since you call `organization_capacity.active_client_count` twice in this method, you might be doing the query twice. Not a blocker, but fyi.

alaroia-cfa (commented):
Splendid!

PR #919, opened 2021-03-11
Alphabetize assignable users

bengolder:


alaroia-cfa (approved):
I approve! I would _love_ to know the impact of this on SQL queries. I'd love to see a screenshot of `rack-mini-profiler` on the web app showing the SQL queries before & after. Not a blocker.

alaroia-cfa:
@bengolder dumb question - is product OK with this? Just want to check, but I presume they would be A-OK with it.

bengolder:
> I would love to know the impact of this on SQL queries

Yeah, I was thinking about this. Some considerations
- I wasn't sure how to always put current user and the existing assigned user first without just pulling them in separately (I looked up how to do a SQL UNION in rails and felt like it was overkill). Making an array felt like the simplest way to enforce this ordering
- I felt that terminating the query into an array wouldn't significantly impact the number of queries, because it's only run once for edit & update, and is stored as an instance variable.

I'll check out the profiler

PR #920, opened 2021-03-11
State routing to available partners only #176790537

bytheway875:
Only do state routing to _available_ partners, otherwise fall through to national partners.

- Adds with_capacity scope to vita_partners, which is somewhat more complex than I'd like due to the fact that we're returning nil instead of 0 for orgs with no active clients.


app/models/vita_partner.rb
```diff
@@ -45,15 +45,22 @@ class VitaPartner < ApplicationRecord
...
+  }
   accepts_nested_attributes_for :source_parameters, allow_destroy: true, reject_if: lambda { |attributes| attributes['code'].blank? }
 
   def at_capacity?
     return parent_organization.at_capacity? if site?
 
     return false if capacity_limit.blank? || organization_capacity.active_client_count.nil?
 
-    return organization_capacity.active_client_count >= capacity_limit
+    organization_capacity.active_client_count >= capacity_limit
```
bytheway875, line 63: redundant return because ruby.

app/services/partner_routing_service.rb
```diff
@@ -9,11 +9,14 @@ def initialize(source_param: nil, zip_code: nil)
 
   # @return VitaPartner the object of the vita_partner we recommend routing to.
   def determine_partner
-    return vita_partner_from_source_param if @source_param.present? && vita_partner_from_source_param.present?
+    from_source_param = vita_partner_from_source_param if @source_param.present?
```
bytheway875, line 12: I discovered while testing this that these methods were actually getting called 2x each, so I switched up the formatting. I think I could also memoize the results of each method in order to prevent this.

app/services/weighted_routing_service.rb
```diff
@@ -0,0 +1,32 @@
+class WeightedRoutingService
+  def initialize(collection)
```
bytheway875, line 2: I moved this to its own file since now we're not always acting on _all_ VitaPartnerState objects from a particular state, nor are they always particularly "real" VitaPartnerState objects.

spec/models/vita_partner_spec.rb
```diff
@@ -32,81 +32,88 @@
     let(:in_range_statuses) { TaxReturnStatus::STATUS_KEYS_INCLUDED_IN_CAPACITY }
 
     context "an organization" do
-      let(:organization) { create :organization, capacity_limit: 10 }
 
       context "at capacity" do
+        let(:at_capacity_org) { create :organization, capacity_limit: 10 }
```
bytheway875, line 37: When running these tests in parallel, I was running into some issues with sharing the same organization in the top level context -- it seemed like we were getting flaky fails because of it, like maybe the teardown of the other tests weren't finished before we started running the next set of tests.

spec/models/organization_capacity_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+# Table name: organization_capacities
+#
+#  active_client_count :bigint
+#  capacity_limit      :integer
+#  name                :string
+#  vita_partner_id     :bigint           primary key
+#
+require 'rails_helper'
+
+describe OrganizationCapacity do
```
bytheway875, line 12: I know we generally try to avoid too many looped tests like this, but this felt like a particularly good use case for it given that the intention of the test is to make sure that our view always appropriately tracks with the statuses we consider to be part of capacity.

spec/services/partner_routing_service_spec.rb
```diff
@@ -61,12 +61,72 @@
...
+              (vita_partner.capacity_limit + 1).times do
+                create :client_with_status, vita_partner: vita_partner, status: "intake_ready"
+              end
+            }
+            before do
+              weighted_service_double = double(WeightedRoutingService)
+              allow(WeightedRoutingService).to receive(:new).and_return weighted_service_double
+              allow(weighted_service_double).to receive(:weighted_routing_ranges).and_return(
+                [
+                  { id: 1, low: 0.0, high: 0.2 },
```
embarnard, line 96: We tested with an empty array returned for weighted_routing_ranges and it still worked--consider using the empty array?

app/services/partner_routing_service.rb
```diff
@@ -9,11 +9,14 @@ def initialize(source_param: nil, zip_code: nil)
...
+    return from_source_param if from_source_param.present?
 
-    return vita_partner_from_zip_code if @zip_code.present? && vita_partner_from_zip_code.present?
+    from_zip_code = vita_partner_from_zip_code if @zip_code.present?
+    return from_zip_code if  from_zip_code.present?
```
embarnard, line 16: extra space after `if`

spec/services/partner_routing_service_spec.rb
```diff
@@ -61,12 +61,72 @@
...
+                  { id: org_with_capacity_and_routing_1.id, low: 0.2, high: 0.6 },
+                  { id: 3, low: 0.6, high: 1.0 }
+                ]
+              )
+            end
+
+            it "only considers vita partners with capacity" do
```
embarnard, line 103: This test could be simplified and made more of a unit test if it mocked-out `VitaPartner.with_capacity` to return just a set of partners who have capacity (and validate that `determine_partner` calls `VitaPartner.with_capacity`). We suggest doing that, but it's not a blocker.

Since `with_capacity` is not currently covered with a test directly on it, it would be a good idea to add one there if you make the above change.

app/models/vita_partner.rb
```diff
@@ -45,15 +45,21 @@ class VitaPartner < ApplicationRecord
   has_many :child_sites, -> { order(:id) }, class_name: "VitaPartner", foreign_key: "parent_organization_id"
 
   default_scope { includes(:child_sites).order(name: :asc) }
-
+  scope :with_capacity, lambda {
+    capacities = OrganizationCapacity.arel_table
+    joins(:organization_capacity).where(
+      capacities[:active_client_count].lteq(arel_table[:capacity_limit])
+          .or(arel_table[:capacity_limit].eq(nil)) # if capacity_limit is not set, assume org capacity
+    ).where.not(capacity_limit: 0)
```
embarnard, line 53: What does it mean for an organization when they have a capacity limit of nil vs 0?

app/models/vita_partner.rb
```diff
@@ -45,15 +45,21 @@ class VitaPartner < ApplicationRecord
...
+  }
   accepts_nested_attributes_for :source_parameters, allow_destroy: true, reject_if: lambda { |attributes| attributes['code'].blank? }
 
   def at_capacity?
```
embarnard, line 57: We are in favor of just using either the with_capacity scope or the at_capacity? method and getting rid of the other. Or rewriting `at_capacity?` to be a thin wrapper over the scope -- no more `if`s in `at_capacity?`. I personally think the at_capacity method is easier to read but understand that the scope might have computational benefits

embarnard (approved):
Reviewed by Em and Asheesh -- Approved, looks great! :D We left a few non-blocking comments

PR #921, opened 2021-03-12
Log document access

bengolder:
This is an updated version of https://github.com/codeforamerica/vita-min/pull/901

- it makes the migration more reliable
- it adds document access logging to the edit view

### DIFF
--- a/app/controllers/hub/documents_controller.rb
+++ b/app/controllers/hub/documents_controller.rb
@@ -17,6 +17,7 @@ module Hub
     end
 
     def show
+      log_document_access!
       redirect_to transient_storage_url(@document.upload.blob)
     end
 
@@ -47,6 +48,17 @@ module Hub
     end
 
     private
+
+    def log_document_access!
+      AccessLog.create!(
+        user: current_user,
+        record: @document,
+        created_at: DateTime.now,
+        event_type: "viewed_document",
+        ip_address: request.remote_ip,
+        user_agent: request.user_agent,
+      )
+    end
 
     def document_params
       params.require(:document)
--- a/db/migrate/20210312004640_add_record_to_access_log.rb
+++ b/db/migrate/20210312004640_add_record_to_access_log.rb
@@ -1,0 +1,17 @@
+class AddRecordToAccessLog < ActiveRecord::Migration[6.0]
+  def up
+    # add the new columns, we will need to make these "null: false" in a later deploy
+    add_reference :access_logs, :record, polymorphic: true
+    # set the new columns
+    update "UPDATE access_logs SET record_type = 'Client', record_id = client_id"
+    # make the old column nullable
+    change_column_null :access_logs, :client_id, true
+  end
+
+  def down
+    # remove the old columns
+    remove_reference :access_logs, :record, polymorphic: true
+    # we can't reset client_id to null false, it's not safe
+  end
+end
+
--- a/app/controllers/hub/clients/bank_accounts_controller.rb
+++ b/app/controllers/hub/clients/bank_accounts_controller.rb
@@ -8,7 +8,7 @@ module Hub
       def show
         AccessLog.create(
           user: current_user,
-          client: @client,
+          record: @client,
           created_at: DateTime.now,
           event_type: "read_bank_account_info",
           ip_address: request.remote_ip,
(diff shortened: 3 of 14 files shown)


app/controllers/hub/documents_controller.rb
```diff
@@ -17,12 +17,15 @@ def index
...
+      log_document_access!
       redirect_to transient_storage_url(@document.upload.blob)
     end
 
     def new; end
 
-    def edit; end
+    def edit
+      log_document_access!
+    end
```
bengolder, line 28: I added this in response to https://github.com/codeforamerica/vita-min/pull/901#issuecomment-792976981

db/migrate/20210312004640_add_record_to_access_log.rb
```diff
@@ -0,0 +1,20 @@
...
+    change_column_null :access_logs, :client_id, true
+  end
+
+  def down
+    # remove the old columns
+    remove_reference :access_logs, :record, polymorphic: true
+    # we can't reset client_id to null false, it's not safe
+  end
+end
+
```
bengolder, line 20: The previous migration:
https://github.com/codeforamerica/vita-min/pull/901/files#diff-43b5d75f9dc345eb9e4109db7ef08d244489f81ed86238e6dedadd30dd6f3014R1-R15

Once this is deployed, we can follow up with a migration that removes the client_id column

app/controllers/hub/documents_controller.rb
```diff
@@ -17,12 +17,15 @@ def index
     end
 
     def show
+      log_document_access!
```
alaroia-cfa, line 20: Consider adjusting we get the URL so it always logs document by default.

app/controllers/hub/documents_controller.rb
alaroia-cfa: This seems unnecessary; is peaceful I suppose to keep! The document URL doesn't actually get sent to any web browsers through this method.

app/controllers/hub/documents_controller.rb
```diff
@@ -48,6 +51,17 @@ def update
 
     private
 
+    def log_document_access!
+      AccessLog.create!(
+        user: current_user,
+        record: @document,
+        created_at: DateTime.now,
```
alaroia-cfa, line 58: flawless

db/migrate/20210312004640_add_record_to_access_log.rb
```diff
@@ -0,0 +1,20 @@
+class AddRecordToAccessLog < ActiveRecord::Migration[6.0]
+  def up
+    # add the new columns (does this make indices?)
```
alaroia-cfa, line 3: Can toss `(...)` in comment

db/migrate/20210312004640_add_record_to_access_log.rb
```diff
@@ -0,0 +1,20 @@
...
+    add_reference :access_logs, :record, polymorphic: true
+    # set the new columns
+    update "UPDATE access_logs SET record_type = 'Client', record_id = client_id"
+    # make the new columns non-nullable
+    change_column_null :access_logs, :record_id, false
```
alaroia-cfa, line 8: Making this non-nullable is unsafe for the same reason as deleting a column.

db/migrate/20210312004640_add_record_to_access_log.rb
```diff
@@ -0,0 +1,20 @@
...
+    change_column_null :access_logs, :record_type, false
+    # make the old column nullable
+    change_column_null :access_logs, :client_id, true
+  end
+
+  def down
+    # remove the old columns
+    remove_reference :access_logs, :record, polymorphic: true
```
alaroia-cfa, line 16: One option if you want is to delete the AccessLogs that don't foreign-key to a Client. That's sorta weird so let's that not write that code and hope we never need to do anything like that.

alaroia-cfa (changes requested):
Looks great overall! Please do change the migration so that we set `record_id` and `record_type` to non-null in the _next_ migration that removes `client_id`. Other than that, 🚢

bengolder:
Requested changes have been added! Merging now.

PR #922, opened 2021-03-12
Send in-progress survey from job [#177066379]

alaroia-cfa:


crontab
```diff
@@ -1 +1,2 @@
-0 0 * * 0 bundle exec rake vita_providers:scrape_vita_providers
\ No newline at end of file
+0 0 * * 0 bundle exec rake vita_providers:scrape_vita_providers
+@hourly bundle exec rake client_surveys:send_client_in_progress_surveys
```
embarnard, line 2: Do we want to execute this at night? Can we limit it to business hours?

embarnard (approved):
Looks good! There is a failing test on ```VitaPartner#at_capacity? an organization over capacity sum of clients assigned to sites within organization exceed capacity limit returns true``` But it seems to be failing on other branches as well...I will do some more investigation before merging into main

PR #925, opened 2021-03-12
Add hold status under final steps

embarnard:


app/lib/tax_return_status.rb
```diff
@@ -43,18 +43,18 @@ def message_templates
...
       review_ready_for_qr: 301, review_reviewing: 302, review_ready_for_call: 303, review_signature_requested: 304, review_info_requested: 305,
 
-      file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406
+      file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406, file_hold: 407
   }.freeze
 
 
   ONBOARDING_STATUSES = [:intake_before_consent, :intake_in_progress, :intake_greeter_info_requested]
-  STATUS_KEYS_INCLUDED_IN_SLA = STATUSES.keys - [:intake_before_consent, :file_accepted, :file_not_filing].freeze
+  STATUS_KEYS_INCLUDED_IN_SLA = STATUSES.keys - [:intake_before_consent, :file_accepted, :file_not_filing, :file_hold].freeze
```
bengolder, line 51: I loved seeing file_hold added to `EXCLUDED_FROM_CAPACITY` and I wonder if it would also be helpful to create an `EXCLUDED_FROM_SLA`. Definitely not a blocker, just an observation.

app/lib/tax_return_status.rb
```diff
@@ -43,18 +43,18 @@ def message_templates
 
       review_ready_for_qr: 301, review_reviewing: 302, review_ready_for_call: 303, review_signature_requested: 304, review_info_requested: 305,
 
-      file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406
+      file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406, file_hold: 407
```
bengolder, line 46: Not a blocker, but when we are adding statuses we could choose to allow for more values in between. For example, this could be `470` instead of `407`. That way we could add more statuses before or after the new status easily in the future.

bengolder (approved):
Overall looks excellent! I added a few observations that are definitely non-blocking and non-urgent.

PR #929, opened 2021-03-13
Remove client reference from access logs

bengolder:
This is a followup to https://github.com/codeforamerica/vita-min/pull/921

It's not safe to make columns non-nullable or to remove them in a migration until _after_ we deploy code that stops referencing the column or ensures the columns have values. Now that the prior PR is deployed, we can remove the foreign key to client from access_logs, and we can make the new polymorphic association non-nullable.

### DIFF
--- a/db/migrate/20210313061212_remove_client_from_access_log.rb
+++ b/db/migrate/20210313061212_remove_client_from_access_log.rb
@@ -1,0 +1,20 @@
+class RemoveClientFromAccessLog < ActiveRecord::Migration[6.0]
+  # Follow up migration for AddRecordToAccessLog
+  def up
+    update <<-SQL.squish
+      UPDATE access_logs
+      SET record_type = 'Client', record_id = client_id
+      WHERE record_id = NULL
+    SQL
+
+    remove_reference :access_logs, :client
+    change_column_null :access_logs, :record_id, false
+    change_column_null :access_logs, :record_type, false
+  end
+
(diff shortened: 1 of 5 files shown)


db/migrate/20210313061212_remove_client_from_access_log.rb
```diff
@@ -0,0 +1,14 @@
+class RemoveClientFromAccessLog < ActiveRecord::Migration[6.0]
+  # Follow up migration for AddRecordToAccessLog
+  def up
+    remove_reference :access_logs, :client
+    change_column_null :access_logs, :record_id, false
```
alaroia-cfa, line 5: This line might fail if any AccessLog records were creating when the old code was running. Just a FYI; not a blocker; if this is a problem, then the migration will happily undo it self due to transactions. :)

alaroia-cfa (approved):
Looks good to me! One non-blocking comment. Thanks for doing this!

alaroia-cfa:
I encourage you to merge & deploy this on its own if possible!

db/migrate/20210313061212_remove_client_from_access_log.rb
bengolder: Do you think I should do the same update to these columns that I did in the earlier migration? I imagine that would make it more likely to succeed.
(set record_type to "Client", set record_id to "client_id")

PR #930, opened 2021-03-15
Add language and service type filters

bytheway875:
Screen Shot 2021-03-15 at 8 51 24 AM
Screen Shot 2021-03-15 at 8 51 33 AM

Bc there are so many filters, I'm needing to limit the sizes of the dropdowns so that they all fit on one line. Also, tweaked the styling on the dropdowns so that there is less extreme padding on the right size of the dropdown, which probably should eventually expanded to all of the "compact" selects. (Fixes the issue where the width of the 80px year dropdown had only a 20px "displayed" area -- the rest was padding.)


app/controllers/concerns/client_sortable.rb
```diff
@@ -14,10 +14,13 @@ def filtered_and_sorted_clients(default_order: nil)
...
     clients = clients.where(tax_returns: { year: @filters[:year] }) if @filters[:year].present?
     clients = clients.where(tax_returns: { status: @filters[:status] }) if @filters[:status].present?
+    clients = clients.where("intakes.locale = :language OR intakes.preferred_interview_language = :language", language: @filters[:language]) if @filters[:language].present?
+    clients = clients.where(tax_returns: { service_type: "online_intake" }) if @filters[:exclude_dropoffs].present?
     clients = clients.where(intake: Intake.where(had_unemployment_income: "yes")) if @filters[:unemployment_income].present?
+
     if @filters[:vita_partner_id].present?
       id = @filters[:vita_partner_id].to_i
-      clients = clients.where("vita_partners.id = ? OR vita_partners.parent_organization_id = ?", id, id)
+      clients = clients.where('vita_partners.id = :id OR vita_partners.parent_organization_id = :id', id: id)
```
coltborg, line 23: What is this `where` change suppose to accomplish? Was the previous syntax with the three arguments incorrect?

app/helpers/tax_return_status_helper.rb
```diff
@@ -27,9 +27,12 @@ def self.stage_and_status_translation(status)
     "#{stage_translation_from_status(status)}/#{status_translation(status)}"
   end
 
-  def language_options
+  def language_options(only_locales: true)
     all_interview_languages = I18n.backend.translations.dig(I18n.locale, :general, :language_options)
-    all_interview_languages.select { |key, _| I18n.locale_available?(key) }.invert # only show available locales
+    if only_locales
+      return all_interview_languages.select { |key, _| I18n.locale_available?(key) }.invert
```
coltborg, line 33: When `only_locales` is `true`, the hash (?) is inverted but not sorted, like if `only_locales` is `false`. Does its also need `.sort`?

app/models/client.rb
```diff
@@ -193,6 +193,12 @@ def clients_with_dupe_contact_info
     Client.after_consent.where(intake: matching_intakes).pluck(:id)
   end
 
+  def preferred_language
+    return intake.preferred_interview_language if intake.preferred_interview_language && intake.preferred_interview_language != "en"
```
coltborg, line 197: When I set a client language to "Spanish" I see Spanish in the `Language` column. But when I set a client to "English", it shows as "N/A" in the `Language` column. Is that expected?

set-client-to-english

language-col-set-to-n:a

app/controllers/concerns/client_sortable.rb
bytheway875: No, it wasn't broken, but I've since learned about a modified updated rails/active record syntax that we're leveraging in another clause. This is now using "named" parameters using symbols for the id instead of using the ? replacement. It's slightly more "Railsy", and continues to pass the tests.

app/helpers/tax_return_status_helper.rb
bytheway875: I'm attempting to use the same helper to accomplish a slightly different task, without breaking or altering original functionality. If you take a look at the keys in the en.yml file that is the source here, the keys for the languages are in ABC order. This means that de: German is at the top, which is kind of odd when you see them in a drop down menu, so I'm sorting them. en: English and es: Spanish are already in alphabetical order and those are the only existing locales, so I didn't need to apply a sort to them in order to maintain the current functionality. There's no harm to not sorting them, but also no real harm to calling sort on an array of length two. Good catch on the asymmetry.

bytheway875:
@coltborg Just to clarify, are you checked out onto this `more-filters` branch when you did the testing for the displayed language? On main, the "language" column only displays the set locale on the client intake. This is set to nil for clients who go through drop off. On this branch, the logic should be updated to show EITHER the preferred interview language OR the locale (whichever is more specific), and would only fall through to N/A if neither is selected. If this was found on this branch, I'll investigate as a bug in this code.

app/models/client.rb
bytheway875: @coltborg Ah, this was a bug where if preferred interview language was english, we wanted to check if the locale was spanish and use that as language, but then weren't falling back on locale if preferred language was nil (like for valet clients). I updated this and added a test. Thanks!

PR #931, opened 2021-03-15
Add current_step to intake, and add appropriate link to portal home dashboard

bytheway875:
Records "current_step" field onto intake, and updates portal dashboard to display a link back to the appropriate location when  onboarding is still in progress.

Some info:

- I decided to store the path instead of the controller name because it simplifies the implementation overall — we can just link directly to the appropriate page without finding the corresponding location and then pathname in the navigation flow. 
- I also decided to set the current path in a before filter on the controller for simplicity’s sake, but that might be worth revisiting! We could save the second post action on each page by instead merging`current_step: next_path` into the form params on the update action of the questions controller which would mean we don’t have to make two post actions on each page. I think that could be done by updating QuestionsForm somehow.
- For returning clients who went through the flow before we started storing current_step on the intake, I decided to simplify the documents logic a bit, sending them to the documents/overview page when they still have documents they need to upload.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,7 +23,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_intake
-    Intake.find_by_id(session[:intake_id])
+    current_client&.intake || Intake.find_by_id(session[:intake_id])
   end
 
   def intake_from_completed_session
@@ -252,6 +252,13 @@ class ApplicationController < ActionController::Base
 
   def load_users
     @users = User.accessible_by(current_ability).order(name: :asc)
+  end
+
+  def set_current_step
+    return unless current_intake.present?
+    return unless request.method_symbol == :get # skip uploads
+
+    current_intake.update(current_step: current_path) unless current_intake.current_step == current_path
   end
 
   rescue_from CanCan::AccessDenied do |exception|
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -256,7 +256,7 @@ class ApplicationController < ActionController::Base
 
   def set_current_step
     return unless current_intake.present?
-    return unless request.method_symbol == :get # skip uploads
+    return unless request.get? # skip uploads
 
     current_intake.update(current_step: current_path) unless current_intake.current_step == current_path
   end
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -1,7 +1,7 @@
 class DocumentsController < ApplicationController
   include AccessControllable
-  before_action :require_intake
-  
+  before_action :require_intake, :set_current_step
+
   def illustration_path; end
 
   def destroy
@@ -14,5 +14,9 @@ class DocumentsController < ApplicationController
     else
       redirect_to overview_documents_path
     end
+  end
+
+  def current_path(params = {})
+    documents_path(self.class.to_param, params)
   end
 end
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -15,8 +15,4 @@ class DocumentsController < ApplicationController
       redirect_to overview_documents_path
     end
   end
-
-  def current_path(params = {})
-    documents_path(self.class.to_param, params)
-  end
 end
--- a/app/controllers/portal/portal_controller.rb
+++ b/app/controllers/portal/portal_controller.rb
@@ -4,7 +4,14 @@ module Portal
     layout "portal"
 
     def home
-      @tax_returns = current_client.tax_returns.order(year: :desc)
+      @current_step = nil
+      @tax_returns = []
+
+      if current_client.completed_intake_process?
+        @tax_returns = current_client.tax_returns.order(year: :desc)
+      else
+        @current_step = current_client.intake.determine_current_step
+      end
     end
 
     def current_intake
--- a/app/controllers/portal/portal_controller.rb
+++ b/app/controllers/portal/portal_controller.rb
@@ -7,7 +7,7 @@ module Portal
       @current_step = nil
       @tax_returns = []
 
-      if current_client.completed_intake_process?
+      if completed_onboarding_process?
         @tax_returns = current_client.tax_returns.order(year: :desc)
       else
         @current_step = current_client.intake.determine_current_step
@@ -22,6 +22,18 @@ module Portal
 
     def require_client_sign_in
       redirect_to root_path unless current_client.present?
+    end
+
+    # We'll consider a client to have completed onboarding process if they've
+    # a) completed_at the intake
+    # b) once any of their tax returns are passed the intake stage
+    # The reason we CANNOT simply rely on completed_at? is because
+    # 1) many times clients "fall off" the intake flow but we complete their taxes anyway.
+    # 2) don't currently (3/22/21) set completed_at on drop-off clients.
+    # Once we've started preparing their taxes, we don't want to prompt them through the intake flow, but instead
+    # show their tax return status information.
+    def completed_onboarding_process?
+      current_client.intake.completed_at? || current_client.tax_returns.map(&:status_before_type_cast).any? { |status| status >= 200 }
     end
   end
 end
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -1,6 +1,6 @@
 module Questions
   class QuestionsController < ApplicationController
-    before_action :require_intake
+    before_action :require_intake, :set_current_step
 
     delegate :form_name, to: :class
     delegate :form_class, to: :class
--- a/app/lib/question_navigation.rb
+++ b/app/lib/question_navigation.rb
@@ -164,4 +164,30 @@ class QuestionNavigation
     Questions::SuccessfullySubmittedController,
     Questions::FeedbackController,
   ].freeze
+
+  def self.determine_current_question(intake)
+    return nil if intake.completed_at?
+    return Questions::ConsentController.to_path_helper unless intake.primary_consented_to_service_at?
+
+    if intake.completed_yes_no_questions_at? && intake.document_types_definitely_needed.present?
+      return Documents::OverviewController.to_path_helper
+    end
+
+    # If yes/no questions completed + docs uploaded, start at InterviewSscheduling. Else, start at consent
+    i = intake.completed_yes_no_questions_at? ? FLOW.index(Questions::OverviewDocumentsController) : FLOW.index(Questions::ConsentController)
+    found_path = nil
+    while found_path.nil?
+      i += 1
+      next if i == FLOW.index(Questions::OverviewDocumentsController) # we handle documents seperately above
+
+      question = FLOW[i]
+      next unless question.show?(intake)
+
+      answer = intake.send(question.form_class.attribute_names.first)
+      next unless ["unfilled", nil].include? answer
+
+      found_path = question.to_path_helper # otherwise, this is the found_path
+    end
+    found_path.to_s
+  end
 end
--- a/app/lib/question_navigation.rb
+++ b/app/lib/question_navigation.rb
@@ -165,10 +165,14 @@ class QuestionNavigation
     Questions::FeedbackController,
   ].freeze
 
-  def self.determine_current_question(intake)
+  # Provides a backfill to determine the current_step value for clients who started intake previous to the addition of
+  # determining current_step during the intake flow
+  def self.determine_current_step(intake)
     return nil if intake.completed_at?
     return Questions::ConsentController.to_path_helper unless intake.primary_consented_to_service_at?
 
+    # If yes/no questions have been completed and we definitely still need certain documents, send
+    # them to the upload docs page.
     if intake.completed_yes_no_questions_at? && intake.document_types_definitely_needed.present?
       return Documents::OverviewController.to_path_helper
     end
--- a/app/lib/question_navigation.rb
+++ b/app/lib/question_navigation.rb
@@ -167,6 +167,7 @@ class QuestionNavigation
 
   # Provides a backfill to determine the current_step value for clients who started intake previous to the addition of
   # determining current_step during the intake flow
+  # TODO: Remove after 2021 tax season closes.
   def self.determine_current_step(intake)
     return nil if intake.completed_at?
     return Questions::ConsentController.to_path_helper unless intake.primary_consented_to_service_at?
@@ -177,8 +178,8 @@ class QuestionNavigation
       return Documents::OverviewController.to_path_helper
     end
 
-    # If yes/no questions completed + docs uploaded, start at InterviewSscheduling. Else, start at consent
-    i = intake.completed_yes_no_questions_at? ? FLOW.index(Questions::OverviewDocumentsController) : FLOW.index(Questions::ConsentController)
+    # If yes/no questions completed + docs uploaded, start at InterviewSscheduling. Else, start after OptionalConsent
+    i = intake.completed_yes_no_questions_at? ? FLOW.index(Questions::OverviewDocumentsController) : FLOW.index(Questions::OptionalConsentController)
     found_path = nil
     while found_path.nil?
       i += 1
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -527,7 +527,7 @@ class Intake < ApplicationRecord
     end
   end
 
-    def document_types_possibly_needed
+  def document_types_possibly_needed
     relevant_document_types.reject(&:needed_if_relevant?).reject do |document_type|
       document_type == DocumentTypes::Other
     end.reject do |document_type|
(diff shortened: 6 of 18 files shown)


spec/controllers/portal/portal_controller_spec.rb
```diff
@@ -33,24 +34,49 @@
...
+
+        it "is ok" do
+          get :home
+
+          expect(response).to be_ok
+        end
+
+        it "loads the client tax returns in desc order" do
+          get :home
+          expect(assigns(:tax_returns).map(&:year)).to eq [2020, 2018, 2017]
```
jenny-heath, line 78: should it check that current_step is nil?

app/controllers/application_controller.rb
```diff
@@ -23,7 +23,7 @@ def hub?
   end
 
   def current_intake
-    Intake.find_by_id(session[:intake_id])
+    current_client&.intake || Intake.find_by_id(session[:intake_id])
```
jenny-heath, line 26: not yet understanding how this relates to current_step - in other words, why couldn't it have been changed separately from this work?

app/controllers/documents_controller.rb
```diff
@@ -15,4 +15,8 @@ def destroy
       redirect_to overview_documents_path
     end
   end
+
+  def current_path(params = {})
```
jenny-heath, line 19: why is this being overwritten? i thought current_path was a rails thing

app/lib/question_navigation.rb
```diff
@@ -163,4 +163,30 @@ class QuestionNavigation
...
+
+  def self.determine_current_question(intake)
+    return nil if intake.completed_at?
+    return Questions::ConsentController.to_path_helper unless intake.primary_consented_to_service_at?
+
+    if intake.completed_yes_no_questions_at? && intake.document_types_definitely_needed.present?
+      return Documents::OverviewController.to_path_helper
+    end
+
+    # If yes/no questions completed + docs uploaded, start at InterviewSscheduling. Else, start at consent
```
jenny-heath, line 175: above steps could also use a comment maybe

app/lib/question_navigation.rb
```diff
@@ -163,4 +163,30 @@ class QuestionNavigation
...
+    i = intake.completed_yes_no_questions_at? ? FLOW.index(Questions::OverviewDocumentsController) : FLOW.index(Questions::ConsentController)
+    found_path = nil
+    while found_path.nil?
+      i += 1
+      next if i == FLOW.index(Questions::OverviewDocumentsController) # we handle documents seperately above
```
jenny-heath, line 180: "we handle documents seperately above" doesn't make much sense to me as an explanation

app/lib/question_navigation.rb
```diff
@@ -163,4 +163,30 @@ class QuestionNavigation
...
+
+      question = FLOW[i]
+      next unless question.show?(intake)
+
+      answer = intake.send(question.form_class.attribute_names.first)
+      next unless ["unfilled", nil].include? answer
+
+      found_path = question.to_path_helper # otherwise, this is the found_path
+    end
+    found_path.to_s
```
jenny-heath, line 190: i actually find this whole function rather difficult to read and still don't understand what it's doing in all of the various cases

app/models/client.rb
```diff
@@ -192,6 +192,12 @@ def clients_with_dupe_contact_info
     Client.after_consent.where(intake: matching_intakes).pluck(:id)
   end
 
+  # We'll consider a client to have completed intake process once any of their
+  # tax returns has passed the intake stage
+  def completed_intake_process?
```
jenny-heath, line 197: now it seems to me like we are using 3 different things to determine where the client is:
- the last page they were on (`current_step`)
- their question answers (`determine_current_question`)
- the status of their tax returns (this method)

bytheway875: current_step is the current step when they go through the process. Determine_current_question provides the backfill to determine the current_step for clients who went through the flow _before_ we started updating current_step on every page. This method is what determines whether we care about that _at all_ because once your tax returns are beyond a certain status we don't send you back into the "onboarding flow"

app/lib/question_navigation.rb
bytheway875: I'm happy to take specific or vague  suggestions of how to improve its readability. It's a utility function to determine the question the client hasn't yet answered. What it's doing is looping through every question form to see which question the client hasn't yet answered and set that as their current question.

app/controllers/documents_controller.rb
bytheway875: I don't know. Questions controller has an explicit method called current_path, so I added one here. I know that questions controller is set up to use the question path as a param, though, so I should check and make sure this isn't extraneous.

app/controllers/application_controller.rb
```diff
@@ -254,6 +254,13 @@ def load_users
     @users = User.accessible_by(current_ability).order(name: :asc)
   end
 
+  def set_current_step
+    return unless current_intake.present?
```
bytheway875, line 258: This is why we need to know current_intake, and why current intake needs to be the intake of the current_client.

app/controllers/application_controller.rb
bytheway875: We're dumping existing clients with existing intakes back into the intake flow for the first time. Controllers like DocumentsController and QuestionsController inherit directly from ApplicationController and need access to the current_clients intake, or else the "intake" that we redirect them to is a new anonymous intake. I guess this could be done separately as part of another task, but my guess is there wasn't a use case where this needed to be updated and now there is.

app/controllers/application_controller.rb
```diff
@@ -254,6 +254,13 @@ def load_users
...
+    return unless request.method_symbol == :get # skip uploads
```
bengolder, line 259: I think there might be a slightly more railsy way to check this

app/lib/question_navigation.rb
```diff
@@ -164,4 +164,34 @@ class QuestionNavigation
     Questions::SuccessfullySubmittedController,
     Questions::FeedbackController,
   ].freeze
+
+  # Provides a backfill to determine the current_step value for clients who started intake previous to the addition of
+  # determining current_step during the intake flow
+  def self.determine_current_step(intake)
+    return nil if intake.completed_at?
+    return Questions::ConsentController.to_path_helper unless intake.primary_consented_to_service_at?
```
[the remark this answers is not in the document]
alaroia-cfa: This unless statement felt much clearer than later portions of this method.
if we just have a series of milestone unless statements there would be less risk of unknown failures (perhaps 6-8 specific answer checks?). dropping them a little earlier with their answers pre-filled is low-risk. They can just click "continue" to move forward.

app/lib/question_navigation.rb
alaroia-cfa: This loop examines the flow & answers in a way that should handle most of the situations I can think of, but I'm not sure if there are edge cases we should be considering. I don't feel confident in my ability to assess it reliably. I'd prefer something simpler if possible.

alaroia-cfa: Would also still prefer most of this to live in a backfill rake task. It feels like permanent complexity for a one-time need.

app/models/client.rb
```diff
@@ -193,6 +193,12 @@ def clients_with_dupe_contact_info
...
+    tax_returns.map(&:status_before_type_cast).any? { |status| status >= 200 }
```
alaroia-cfa, line 199: alternatively 
```ruby
tax_returns.where("status >= ?", 200).exists?
```

alaroia-cfa: See comments on portal controller for more concerns about this method of checking doneness

app/models/intake.rb
```diff
@@ -580,4 +580,13 @@ def update_or_create_additional_consent_pdf
...
   end
+
+  # Backfills current_step for clients who started intake before we tracked current_step
+  def determine_current_step
+    return current_step if current_step.present?
+
+    step = QuestionNavigation.determine_current_step(self)
+    update(current_step: step)
+    step
+  end
```
alaroia-cfa, line 591: maybe we can move this into a rake task

spec/controllers/portal/portal_controller_spec.rb
```diff
@@ -33,24 +34,49 @@
...
+          expect(assigns(:current_step)).to eq "/en/questions/additional-info"
+          expect(assigns(:tax_returns)).to eq []
+        end
       end
 
-      it "loads the client tax returns in desc order" do
-        get :home
-        expect(assigns(:tax_returns).map(&:year)).to eq [2020, 2018, 2017]
+      context "post-onboarding" do
+        let(:client) { create :client, intake: (create :intake) }
```
alaroia-cfa, line 62: would be good to see the current_step set in both cases, and ideally to show a client who has _just_ finished intake (is in ready for review)

spec/features/portal/client_portal_dashboard_spec.rb
```diff
@@ -1,44 +1,69 @@
...
+    let(:client) { create :client, intake: (create :intake, preferred_name: "Randall", current_step: "/en/documents/overview") }
+    before do
+      login_as client, scope: :client
+    end
+    scenario "linking to next step" do
+      visit portal_root_path
+      expect(page).to have_text "Welcome back Randall!"
+      expect(page).to have_link("Submit remaining tax documents", href: "/en/documents/overview" )
+    end
+  end
```
alaroia-cfa, line 26: I'd recommend moving these feature specs into the portal controller spec with a render_views, as they are mainly testing the template conditional statement.

I think there is a bigger, more important feature spec to add later, in which a client partially fills out intake, drops off, returns and signs into the portal, then sees one of these links and follows it to complete their intake, but that can be added with a later story.

spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -130,6 +134,7 @@
     click_on "No"
 
     # Income from working
+    expect(intake.reload.current_step).to eq("/en/questions/job-count")
```
alaroia-cfa, line 137: I don't feel like we need so many of these same kinds of assertions. One or a few seems good.

spec/lib/question_navigation_spec.rb
```diff
@@ -0,0 +1,397 @@
...
+                 had_wages: "yes",
+                 had_self_employment_income: "no",
+                 had_disability_income: "no",
+                 had_interest_income: "unsure",
+                 sold_assets: "yes",
+                 had_asset_sale_income: "unfilled"
+        }
+        it "has next step as other-states" do
+          expect(described_class.determine_current_step(intake)).to eq "/en/questions/asset-sale-income"
+        end
```
alaroia-cfa, line 120: looks like other-states is a typo here

alaroia-cfa (changes requested):
(Asheesh & Ben pairing on this review)
This overall looks good! Thanks for pushing the returning clients work forward sooner!

I think there are a few things that would be helpful to change
- Moving the backfill into a rake task that we run once
- having a consistent way to check if an intake is complete. This introduces a new check based on tax return status (and it's possible for a client to finish intake but still return false through this method). I don't think it will be clear to future engineers which check should be used.

Another thought: How should drop off clients be handled in the portal? Should they see these links?
- drop off is automatically set as ready for prep
- drop off does not get completed_at

Happy to talk more via Tuple or re-review promptly!

spec/features/web_intake/new_single_filer_spec.rb
bytheway875: I was using these tests to ensure that the code actually worked in most cases that branch off in different ways to ensure that the backfill works in all cases.

app/models/intake.rb
bytheway875: My take was that there are so many unused/unusable intakes, so we'd be looping through a bunch of extraneous data in order to accomplish the backfill. Here, we only need to run the backfill for clients who actually decide to use the returning client feature. This is intended to be removed at the end of the tax season, because moving forward we'll already keep track of it throughout the flow. My intention here, like indicated in the ticket, was to implement something fairly efficiently without having to spend additional time creating a query and running it on prod. I'm willing to be convinced otherwise, but it just seems like running this on every intake would mean having to split things into groups, consider which intakes to run it on etc etc.

app/controllers/questions/questions_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Questions
   class QuestionsController < ApplicationController
-    before_action :require_intake
+    before_action :require_intake, :set_current_step
```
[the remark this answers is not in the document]
bytheway875: Thats a good idea, but I kind of think it's outside the scope of this particular PR. I know it needs to be done, but this is already kind of a big PR and I don't want to increase the imprint.

app/controllers/portal/portal_controller.rb
```diff
@@ -4,7 +4,14 @@ class PortalController < ApplicationController
...
 
     def home
-      @tax_returns = current_client.tax_returns.order(year: :desc)
+      @current_step = nil
+      @tax_returns = []
+
+      if current_client.completed_intake_process?
+        @tax_returns = current_client.tax_returns.order(year: :desc)
+      else
+        @current_step = current_client.intake.determine_current_step
```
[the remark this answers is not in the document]
bytheway875: I originally had it as current_client.intake.current_step || current_client.intake.determine_current_step, but that seemed kind of redundant since determine_current_step was already returning current_step if it wasn't already determined.

Also, the idea of showing this until they've started prep is by design and was dictated by product/design/customer research. The idea is that we're going to be showing this list of steps up until we've begun preparing their taxes. I spoke to them about it at length and discussed some tradeoffs, but we went this direction at their request.

app/controllers/application_controller.rb
```diff
@@ -254,6 +254,13 @@ def load_users
...
+
+    current_intake.update(current_step: current_path) unless current_intake.current_step == current_path
+  end
```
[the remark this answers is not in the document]
bytheway875: I attempted to test these in the documents controller spec + questions controller spec, but take a look at those specs -- they aren't even actual controller specs, and when I attempted to convert them into controller specs I ran into issues with them because the routes are set up in some funky ways.

app/lib/question_navigation.rb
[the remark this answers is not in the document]
bytheway875: The original questions logic seems like it was set up to be super parallelized and similar... almost like it was made to be able to do something like this to it. If setting them back at the exact point they dropped off is possible/already implemented _and_ is what was requested by stakeholders, that seems like an appropriate place to prioritize user experience over a bit of additional (and temporary -- we can remove this code after this season) code complexity.

app/controllers/portal/portal_controller.rb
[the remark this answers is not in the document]
bytheway875: completed_at cannot be used for this value because so many of our existing users _never_ have their intake set as completed -- they're signing their tax returns and their intake was never "officially" marked as completed_at because so many drop off at the document upload step. So we needed to choose some other way to make the cut off between when we stop showing the intake part and when we show the tax return part. Keep in mind that this is an "in between" state to bridge the existing implementation of e-signatures and the introduction of returning client and I don't particuarly expect that this is the way that we will determine this state forever, but its reliable and works for the cases that product/design wants to address in the short term. Upcoming stories will ensure that there are certain features available to users in the "in between" stages (message your tax team, upload a document), but this prevents us from showing "finish your intake" steps to users who we're actually successfully preparing taxes for with the information they've already provided.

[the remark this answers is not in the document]
bytheway875: This also appropriately covers the drop-off scenario because we _don't_ mark those drop-offs as completed_at, but we put them right into the prep state. In the drop-off case, individuals who log in would see the tax return steps with the "there are no documents for you to review yet" message, which seems like what we'd want/expect.

spec/features/portal/client_portal_dashboard_spec.rb
bytheway875: I wrote these as separate feature specs because to my mind they are two distinct client flows that will continue to diverge more and more as we build out these features. Also, I think render_views in controller tests should be used sparingly, as it's kind of seen as a "last resort" feature in rails -- thats why Rails makes you load it in as an option instead of being on by default.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
bytheway875: changed.

app/lib/question_navigation.rb
```diff
@@ -163,4 +163,30 @@ class QuestionNavigation
     Questions::SuccessfullySubmittedController,
     Questions::FeedbackController,
   ].freeze
+
+  def self.determine_current_question(intake)
```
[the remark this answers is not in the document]
bytheway875: I refactored this into a background job, but the background jobs fails because we're grabbing data from controllers that aren't available/set up correctly when running outside of the application flow. There's a larger refactor that can be done to deal with that issue and not rely on the controllers at all, but all of the current "Flow" logic is based on controllers. This doesn't seem like the right time to do that refactor, and I don't want to spend too much more time on the code path for backfilling -- this should work fine, and I've added TODOs to remove at the end of the season.

app/lib/question_navigation.rb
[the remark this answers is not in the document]
alaroia-cfa: If we can block clients from logging in if they haven't consented yet, then this problem goes away. I'll plan on us doing that. cc: @bengolder

alaroia-cfa (approved):
Whee! Merging.

PR #937, opened 2021-03-17
Create My Updates tab

embarnard:


bytheway875:
A couple of things:

- Can we check with product on whether we want to create a notification _when you assign something to yourself_? We currently send emails when you assign to yourself, but product actually wants us to change this behavior because it creates a lot of noise. I think we'll have the same problem here -- if you're assigning a tax return to yourself, I think we should assume you'll know you did that and that it will appear on your assigned clients list, but not create a notification OR email you about this. We can fix this by ensuring that the assigned_user != assigning_user in the logic block.
- There is a "timestamp" helper in the messages branch that provides a version of the displayed time without the "at" which is more succinct and I think matches the mock.
- My thought is that we create the "tax return assignment" in the assign block, and then allow the assignment to create the notification, mailer, and system note objects.
- Moving to have "tax return assignments" is a great change, and I think we should eventually move towards using it to store the current tax_return assignee. This seems like a bigger refactor that I'm not sure we have time to do now, but I'd love to look at this with you eventually.

PR #939, opened 2021-03-18
Expose type column on system note for analytics

bytheway875:
@alaroia-cfa Katie Eggers would like to be able to do some metabase magic using a particular type of system note, so we need to expose the type column. (Is there anything special that needs to be done to propagate these changes once we release)

### DIFF
--- a/db/create_analytics_views.sql
+++ b/db/create_analytics_views.sql
@@ -166,7 +166,7 @@ CREATE OR REPLACE VIEW analytics.stimulus_triages AS
     FROM public.stimulus_triages;
 
 CREATE OR REPLACE VIEW analytics.system_notes AS
-    SELECT id, client_id, created_at, updated_at, user_id
+    SELECT id, client_id, created_at, type, updated_at, user_id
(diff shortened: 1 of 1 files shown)


db/create_analytics_views.sql
```diff
@@ -166,7 +166,7 @@ CREATE OR REPLACE VIEW analytics.stimulus_triages AS
     FROM public.stimulus_triages;
 
 CREATE OR REPLACE VIEW analytics.system_notes AS
-    SELECT id, client_id, created_at, updated_at, user_id
+    SELECT id, client_id, created_at, type, updated_at, user_id
```
bengolder, line 169: I just noticed that should probably change client_id too, since it no longer exists (replaced by record_id & record_type)! If you have any ideas for how to keep these analytics better in sync with migrations, let me know.

bengolder (approved):
Thanks!

db/create_analytics_views.sql
bytheway875: That's the access_logs table!

PR #940, opened 2021-03-18
State routing read only

bytheway875:
Screen Shot 2021-03-18 at 2 54 06 PM

### DIFF
--- a/app/views/hub/state_routings/index.html.erb
+++ b/app/views/hub/state_routings/index.html.erb
@@ -1,0 +1,35 @@
+<% @page_title = t("general.state_routing") %>
+<% content_for :page_title, @page_title %>
+
+<% content_for :card do %>
+  <div class="slab slab--half-padded">
+    <div class="grid">
+      <h1><%= @page_title %></h1>
+      <div class="grid__item width-one-half">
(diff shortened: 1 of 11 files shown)


app/views/hub/state_routings/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+                  <% state_routing[1].each do |entry| %>
+                    <li>
+                      <div>
+                        <%= entry.vita_partner.name  %>
+                      </div>
+                      <div>
+                        <%= routing_fraction_to_percentage entry.routing_fraction %>
+                      </div>
+                    </li>
```
bengolder, line 24: It looks like there are a couple minor things missing here that are in the story:
- each of these rows should be a link to the organization's edit page
- if the percentage is 0%, then we should show the org as greyed out

bengolder (dismissed):
It looks like there a couple minor details missing based on the story (linking, gray out 0%), but otherwise this is perfect!

bytheway875:
Screen Shot 2021-03-19 at 6 32 15 PM
Screen Shot 2021-03-19 at 6 32 21 PM

PR #941, opened 2021-03-19
Fix Spanish translation in SMS verification codes flow

alaroia-cfa:


coltborg (approved):
@alaroia-cfa and I walked through this PR and confirmed the language is updated.

Screen Shot 2021-03-19 at 1 24 52 PM

PR #942, opened 2021-03-19
Update valet icon & add tooltip

bengolder:
Screen Shot 2021-03-19 at 11 57 30 AM


coltborg (changes requested):
The code and tests look good.

One thing I think should be address is the vertical alignment of the icon and the text surrounding it. The icon is a little high between the year text before and the button after.

I was able to solve this by adding `vertical-align: middle;` to both `.tooltip` and `.tooltip__trigger`. This could be best scoped under the `.tax-return-list__service-type` to only add this vertical alignment in this situation.

center-with-vertical-align


To show the difference, here is it without `vertical-align`:

before

Here it is with vertical align:

after

coltborg (approved):
Reviewed the change with Ben, my previous comments are addressed.

PR #947, opened 2021-03-24
Clear out document request after a document has been added and a client clicks "continue"

bytheway875:
Based on feedback from Anu/Yvonne:

- They didn't want us to keep a document request history for the whole "session", so when someone hits continue or go back after they've uploaded a document, we'll clear the documents request and set it as "completed" so that the old documents don't show in the view. (I did explain to Yvonne that the documents get uploaded as soon as someone hits "upload", so it might be someone unintuitive to clients that "go back" doesn't cancel your document upload. She seemed fine with that)
- They wanted the document type to be "Other" instead of requested later.


app/controllers/portal/upload_documents_controller.rb
```diff
@@ -52,11 +52,15 @@ def form_class
     end
 
     def form_params
-      params.fetch(form_class.form_param, {}).permit(form_class.attribute_names)
+      params.fetch(form_class.form_param, {}).permit(form_class.attribute_names).merge(document_type: DocumentTypes::Other.key)
```
jenny-heath, line 55: why do we need this now? did the fact that the type changed affect the params?

app/forms/requested_document_upload_form.rb
```diff
@@ -9,10 +9,11 @@ def initialize(documents_request, *args, **kwargs)
 
   def save
     document_file_upload = attributes_for(:documents_request)[:document]
+    document_type = attributes_for(:documents_request)[:document_type]
     if document_file_upload.present?
       @documents_request.documents.create(
         uploaded_by: @documents_request.intake.client,
-        document_type: DocumentTypes::RequestedLater.key,
+        document_type: document_type || DocumentTypes::RequestedLater.key,
```
jenny-heath, line 16: so was it just not using `UploadDocumentsController#document_type` before? would love to go over this

app/models/intake.rb
```diff
@@ -16,6 +16,7 @@
 #  completed_at                                         :datetime
 #  completed_yes_no_questions_at                        :datetime
 #  continued_at_capacity                                :boolean          default(FALSE)
+#  current_step                                         :string
```
jenny-heath, line 19: maybe left over from another branch?

config/routes.rb
```diff
@@ -118,7 +118,10 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
         get '/success', to: 'tax_returns#success', as: :success
       end
       resources :documents, only: [:show]
-      resources :upload_documents, only: [:new, :create, :destroy]
+      resources :upload_documents, only: [:destroy]
+      match 'upload-documents', to: 'upload_documents#edit', via: :get
+      match 'upload-documents', to: 'upload_documents#update', via: :put
+      match 'complete-documents-request', to: 'upload_documents#complete_documents_request', via: :get
```
jenny-heath, line 124: is this how the fancy routes for intake quesitons / documents work?

app/controllers/portal/upload_documents_controller.rb
jenny-heath: ammendment: switch to `document_type.key`

PR #948, opened 2021-03-24
Use a relationship to notifications instead of access control for notifications page

bytheway875:
Using access control for this page seemed weird to me because a) everyone can access it and b) everyone's set of data is the same regardless of role. Restricting access at the ability level is great _when access is dependent on role_. However, when it's based explicitly on ownership/relationship to the data object, as in this case, grabbing data through an active record relationship is probably simpler and more efficient.

### DIFF
--- a/app/controllers/hub/user_notifications_controller.rb
+++ b/app/controllers/hub/user_notifications_controller.rb
@@ -1,18 +1,16 @@
 module Hub
   class UserNotificationsController < ApplicationController
     include AccessControllable
-
     before_action :require_sign_in
-    load_and_authorize_resource only: [:index]
     layout "admin"
 
     def index
-      @page_title = "Notifications"
-      @user_notifications = @user_notifications.order(created_at: :desc).page(params[:page])
+      @page_title = I18n.t("hub.clients.navigation.notifications")
+      @user_notifications = current_user.notifications.order(created_at: :desc).page(params[:page])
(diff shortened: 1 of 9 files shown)


app/controllers/hub/user_notifications_controller.rb
```diff
@@ -1,18 +1,16 @@
...
     include AccessControllable
-
     before_action :require_sign_in
-    load_and_authorize_resource only: [:index]
     layout "admin"
 
     def index
-      @page_title = "Notifications"
-      @user_notifications = @user_notifications.order(created_at: :desc).page(params[:page])
+      @page_title = I18n.t("hub.clients.navigation.notifications")
```
jenny-heath, line 8: oh good catch!

app/models/user.rb
```diff
@@ -61,6 +61,7 @@ class User < ApplicationRecord
   validates :email, 'valid_email_2/email': { mx: true }
   has_many :assigned_tax_returns, class_name: "TaxReturn", foreign_key: :assigned_user_id
   has_many :access_logs
+  has_many :notifications, class_name: "UserNotification"
```
jenny-heath, line 64: oh wow didn't expect this to be missing

jenny-heath (approved):
looks great! let's test it manually and revisit the unit tests i added before merging

PR #949, opened 2021-03-24
Stop generating walkthroughs automatically in CircleCI

alaroia-cfa:
Walkthrough generation is currently broken.

I have a guess as to why -- maybe the CircleCI ramdisk runs out of ramdisk disk space, so we should move our CircleCI config to stop using a ramdisk.

I haven't tested if that fixes it b/c I have other fish to fry at the moment!

Note that with this change, you can still push to a branch called `circleci-update-walkthroughs` to trigger walkthrough generation. Therefore it should be reasonably easy to test a fix for this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ workflows:
           requires: [run_js_tests, run_ruby_tests]
           filters:
             branches:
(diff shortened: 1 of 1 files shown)


embarnard (approved):
Approved! Testing by merging to main

PR #951, opened 2021-03-24
Allow clients to log out

embarnard:


app/views/shared/_header.html.erb
```diff
@@ -14,6 +14,11 @@
       <% if I18n.locale != :es %>
         <%= link_to_spanish(class: "toolbar__item text--small") %>
       <% end %>
+      <% if current_client.present? %>
```
bytheway875, line 17: i think rails actually allows you to do a link_to with a method delete, but this is probably semantically correct in terms of the division between buttons and links.

PR #953, opened 2021-03-24
Require client sign in after consent

bengolder:
I created two separate controllers:
- AnonymousIntakeController
- AuthenticatedIntakeController

Both inherit from QuestionsController, but handle the session differently (authenticated vs not authenticated)
All the intake controllers now inherit from one or the other of these two base controllers.

- The ConsentController now signs in the client after a successful POST (note that this seems to clear the session automatically)
- The SuccessfullySubmittedController, which previously cleared the session, now logs out the client after they finish.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,6 +35,7 @@ class ApplicationController < ActionController::Base
   end
 
   def clear_intake_session
+    sign_out current_client
     session.delete("intake_id")
   end
 
--- a/app/controllers/questions/anonymous_intake_controller.rb
+++ b/app/controllers/questions/anonymous_intake_controller.rb
@@ -1,0 +1,5 @@
+module Questions
+  class AnonymousIntakeController < QuestionsController
+    before_action :require_intake
+  end
+end
--- a/app/controllers/questions/authenticated_intake_controller.rb
+++ b/app/controllers/questions/authenticated_intake_controller.rb
@@ -1,0 +1,11 @@
+module Questions
+  class AuthenticatedIntakeController < QuestionsController
+    include ClientAccessControlConcern
+
+    before_action :require_client_login
+
+    def current_intake
+      current_client&.intake
+    end
+  end
+end
--- a/app/controllers/dependents_controller.rb
+++ b/app/controllers/dependents_controller.rb
@@ -1,5 +1,7 @@
 class DependentsController < ApplicationController
-  before_action :require_intake
+  include ClientAccessControlConcern
+
+  before_action :require_client_login
   helper_method :next_path
 
   def index
--- a/app/controllers/documents/document_upload_question_controller.rb
+++ b/app/controllers/documents/document_upload_question_controller.rb
@@ -1,5 +1,5 @@
 module Documents
-  class DocumentUploadQuestionController < Questions::QuestionsController
+  class DocumentUploadQuestionController < Questions::AuthenticatedIntakeController
     layout "document_upload"
 
     delegate :document_type_key, to: :class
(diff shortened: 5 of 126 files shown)


app/controllers/application_controller.rb
```diff
@@ -35,6 +35,7 @@ def current_stimulus_triage
   end
 
   def clear_intake_session
+    sign_out current_client
     session.delete("intake_id")
```
bengolder, line 39: I thought that it made sense to automatically sign out the client at the point that we would normally clear their session.

bytheway875: I think this implementation fits the scope of this ticket! Do you know if there is a follow up ticket for what to do with authenticated clients once they've finished intake? If not, it might be worth putting on product's radar!

app/controllers/questions/anonymous_intake_controller.rb
```diff
@@ -0,0 +1,5 @@
+module Questions
+  class AnonymousIntakeController < QuestionsController
+    before_action :require_intake
```
bytheway875, line 3: I took a look at the source code because I was wondering if we could move require_intake definition directly into this file. It looks like `DependentsController` still inherits directly from ApplicationController and uses the `before_action :require_intake`. I know that controller acts a bit differently than others in the flow, but I'm wondering if it should/can inherit from the AuthenticatedIntakeController instead?

(Because the definition of require_intake calls code that looks for the intake either from the current_client or the session intake, it seems like the code path would be "safe" even without the refactor of DependentsController!)

app/controllers/questions/authenticated_intake_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Questions
+  class AuthenticatedIntakeController < QuestionsController
```
bytheway875, line 2: The setup in this controller looks hella similar to what's happening in the Portal::PortalController, which we're inheriting from in some of the sub-portal controllers. I'm wondering if we could consolidate. It seems like right now, that's not a particularly straightforward task (because this controller actually uses QuestionsController as its parent).  

I think this would require a significant refactor to do it in a way that would diminish complexity instead of increase it, but perhaps we can get around to it some day.

bytheway875 (commented):
I'd like to see us  refactor DocumentsController to inherit from the AuthenticatedIntakeController, and then move the require_intake logic to the AnonymizedIntakeController so that we can simplify some of the current_intake logic. I'd suggest either doing it within the scope of this ticket, or we can file a follow-up ticket for the refactor! Otherwise, LGTM!

app/controllers/application_controller.rb
bengolder: There is not currently any follow up ticket, so yeah, that's a great suggestion! I'll bring it up with them.

app/controllers/questions/anonymous_intake_controller.rb
bengolder: You are right to call out the `:require_intake` in dependentsController! It should have the same client login behavior as the AuthenticatedIntakeController. However, I didn't want it to inherit all the QuestionsController stuff, so instead I treated it similar to a portal controller, and used the client access control concern

```ruby
class DependentsController < ApplicationController
  include ClientAccessControlConcern

  before_action :require_client_login
```

app/controllers/questions/authenticated_intake_controller.rb
bengolder: I 100% agree! I considered putting the whole authenticated intake portion into the portal, but felt that in the short term that it made sense to distinguish between a protected portion of the intake and the portal. But I definitely feel like there is potential for good refactors in the future.

bengolder:
Thank you Shannon! I appreciate and agree with your review!

I had completely missed the DependentsController, so I'm super grateful that you caught that! Since the DependentsController is significantly different from the other linear question controllers, I decided to include the ClientAccessControlConcern and require client login instead of inheriting from AuthenticatedIntakeController, which would have introduced all the QuestionsController functionality.

PR #954, opened 2021-03-25
Revert 5ab879fa to allow drop-off clients to e-sign

alaroia-cfa:


alaroia-cfa:
I am going to self-merge this to get prod workable for drop-off clients.

PR #955, opened 2021-03-25
Fix optional content error

bytheway875:
I wasn't quite sure how to test this in a unit test (Maybe could create an optional content pdf fixture, but I actually wasn't sure how to do that), but manual testing was successful.


bengolder (approved):
The comment is super helpful, thanks for including that!

PR #962, opened 2021-03-26
Change consented at column to display created at date on all clients index page

embarnard:


bytheway875:
@embarnard I think there might be some tests that need to be updated as a result of this change? The ruby tests step seems to be failing.

PR #966, opened 2021-03-27
Allow archiving docs in hub

bengolder:
Completes _[Allow users to archive documents](https://www.pivotaltracker.com/story/show/177273449)_

archiving-files


app/assets/images/icons/folder.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="15" height="13" viewBox="0 0 15 13" fill="none" xmlns="http://www.w3.org/2000/svg">
```
bytheway875, line 1: This looks to be the folder icon from Material Icons -- can we use that instead?

app/assets/stylesheets/templates/_hub-documents-index.scss
```diff
@@ -0,0 +1,13 @@
+.document-index-actions a {
+  color: inherit;
+}
+
+.index-table__cell:first-child {
```
bytheway875, line 5: This isn't scoped to the documents-index as the filename suggest, so its going to bold the first child any time index-table__cell is applied in any table -- are we confident that's what we want?

app/views/hub/documents/_form.html.erb
```diff
@@ -17,6 +17,7 @@
       <%= f.vita_min_text_field :display_name, t('hub.documents.display_name'), classes: ["form-width--long"] %>
       <%= f.cfa_select :document_type, t("general.document_type"), DocumentTypes::ALL_TYPES.map { |doc_type| [doc_type.label, doc_type.key] }, include_blank: true %>
       <%= f.cfa_select :tax_return_id, t("general.tax_return"), @client.tax_returns.map { |tax_return| [tax_return.year, tax_return.id] }, include_blank: true %>
+      <%= f.cfa_checkbox :archived, t(".archived_label") %>
```
bytheway875, line 20: Can you swap this out with f.hub_checkbox so that it matches the checkbox style in the mock?

app/controllers/hub/documents_controller.rb
```diff
@@ -10,10 +10,13 @@ class DocumentsController < ApplicationController
     layout "admin"
 
     def index
-      @sort_order = sort_order
-      @sort_column = sort_column
-      @documents = @documents.except(:order).order({ @sort_column => @sort_order })
-      @document = Document.new # used for form to upload documents
+      @documents = sorted_documents.active
```
bytheway875, line 13: This is more of a design/UX question, but do we have any sense of whether a user would find it helpful to see the archived documents in the same list as the regular documents (for comparison purposes, locating a similar/previous version of a document, etc?) It seems like it would be just as simple to click on the docs folder and have that reload the page with the archived documents expanded below the active docs.

bytheway875 (changes requested):
I'd particularly like to see the hub_checkbox used on the documents/form, and I think its worth attempting to use the Material Icon instead of adding a new svg if possible.

PR #967, opened 2021-03-29
Add documents to edit view so that people can view the image while they're editing information about it.

bytheway875:
Screen Shot 2021-03-29 at 9 47 04 AM
Screen Shot 2021-03-29 at 9 47 17 AM


alaroia-cfa (approved):
A+!

`transient_storage_url` is safe to use here the way you've used it b/c the `@document` has already been access-checked.

So exciting. :ship:

alaroia-cfa:
@bytheway875 consider deploying this is so good 🥂

PR #969, opened 2021-03-30
Update returning-clients page help text & link & send events to mixpanel [#177501480]

alaroia-cfa:
After this change:

image

image


bengolder (approved):
Looks great!

PR #972, opened 2021-03-30
Add spacing between filter and clear buttons on filter client header

embarnard:


app/assets/stylesheets/components/_actions-wrapper.scss
```diff
@@ -4,4 +4,8 @@
   .button {
     margin-bottom: 0;
   }
+
+  div input.button{
+    margin-right: 15px !important;
```
bytheway875, line 9: im guessing this wasnt working without the important?

bytheway875: looks like you do need the important. grr. I wonder what styling is preventing that override.. im guessing specificity conflicts from honeycrisp but idk. Can you swap out the pixel values to rem, too?However, I think you can remove this custom rule and target the existing .button class rules directly above:
`margin-right: 1.5rem`

embarnard: Nice! Made those changes. Yeah I think its overwritten by compact Honeycrisp elements

PR #977, opened 2021-03-31
Send confirmation message with client ID to drop off valet clients

embarnard:


config/locales/en.yml
```diff
@@ -2351,4 +2351,22 @@ en:
       title: Access not allowed
       body_html: Oops! Looks like you're trying to access a page or client you do not have access to. Please contact Partner Support at <a href="mailto:partnersupport@getyourrefund.org">partnersupport@getyourrefund.org</a> if you need to merge a duplicate client you do not have access to.
     home:
-      already_started: "Already started? Sign in"
\ No newline at end of file
+      already_started: "Already started? Sign in"
```
embarnard, line 2354: The text for these messages is very similar to the successful submission message text from `en.messages.successful_submission.email_body`

app/forms/hub/create_client_form.rb
```diff
@@ -61,6 +61,27 @@ def save(current_user)
         tax_returns_attributes: @tax_returns_attributes.map { |_, v| create_tax_return_for_year?(v[:year]) ? tax_return_defaults.merge(v) : nil }.compact
       )
 
+      locale = @client.intake.preferred_interview_language == "es" ? "es" : "en"
```
embarnard, line 64: Wasn't sure the best way to set the locale on these messages since they depend on the preferred_interview_language

bytheway875: that makes sense to me.

app/forms/hub/create_client_form.rb
```diff
@@ -61,6 +61,27 @@ def save(current_user)
...
+      if @client.intake.sms_notification_opt_in == "yes"
```
bytheway875, line 65: Eventually we should consolidate this logic so that we don't have to do so much setup. something like @client.send_message(), where the method takes care of determining which form to send it in. There's a lot of duplicated logic across the app we'd be able to clean up! No action necessary now, just something to think about.

config/locales/en.yml
```diff
@@ -2357,3 +2357,21 @@ en:
...
+    email:
+      subject: Thank you for submitting your tax information
+      body: |
+        Hello %{preferred_name},
+
+        Your tax information has been successfully submitted to your tax preparer at %{org_name}! Your Client ID is %{confirmation_number}.
+
+        Your tax specialist will review your information and will reach out to you in 3-5 business days to review your situation before preparing your taxes.
+
+        You can securely upload any additional tax documents here: %{doc_upload_link}
```
bytheway875, line 2370: I spoke with @yvonnefong, and she was saying we could probably swap this out with the portal login link: "You can check the status of your tax returns and upload additional documents by logging in at https://getyourrefund.org/portal/login". Can you replace the doc upload link with the login link?

bytheway875 (changes requested):
Can you swap out the documents requested later link with copy that directs the client to use the login functionality to upload docs?

config/locales/en.yml
embarnard: @bytheway875 definitely! Thanks for the review

PR #978, opened 2021-03-31
Update macro and add useful descriptions for requested documents

bytheway875:


config/locales/es.yml
```diff
@@ -232,6 +232,21 @@ es:
         file_not_filing: No archivar
         file_hold: En pausa
     status_macros:
+      intake_greeter_info_requested: |
+        ¡Hola <<Client.PreferredName>>!
+
+        Para continuar presentando sus impuestos, iniciar sesión en https://www.getyourrefund.org/es/portal/login.
```
alaroia-cfa, line 238: Consider adjusting the text so `.` isn't next to the URL; otherwise, some email clients may linkify `/es/portal/login.` (with the . as part of the URL).

One way to do that is to say: `iniciar sesión en: https://www.getyourrefund.org/es/portal/login`

This also might be a non-issue. Feel free to ignore.

alaroia-cfa (approved):
Splendid! Approved. One minor piece of advice that I don't know if it's actually relevant.

config/locales/es.yml
bytheway875: good call. I think most clients deal with this okay in the year of our lord 2021, but you never know so I removed the periods because not fixing this later > grammar.

PR #979, opened 2021-04-05
Redirect invalid  login links to portal/login

embarnard:


app/controllers/portal/client_logins_controller.rb
```diff
@@ -63,7 +63,9 @@ def check_verification_code
       render :enter_verification_code
     end
 
-    def invalid_token; end
+    def invalid_token
+      redirect_to portal_client_logins_path
+    end
```
bengolder, line 68: It looks like we could probably delete this action in a followup deploy, since we'll no longer be redirecting to it.

bengolder (approved):
Overall looks good! It seems like we'll probably want a follow up to delete the invalid token page, since we won't be using it anymore. In this commit, we could delete the invalid_token template, since we are no longer showing it.

Thanks!

app/controllers/portal/client_logins_controller.rb
bengolder: I think it's good to deploy a deletion separately, in the small chance that someone has been redirected to this page right before this deploy.

PR #984, opened 2021-04-05
Taggable users method + controller tests for @taggable_users

bytheway875:

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -154,6 +154,22 @@ class User < ApplicationRecord
     end
   end
 
+  def self.taggable_for(client)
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -154,6 +154,22 @@ def accessible_users
     end
   end
 
+  def self.taggable_for(client)
+    users = User.where(role_type: [AdminRole::TYPE, ClientSuccessRole::TYPE, GreeterRole::TYPE])
+    coalition = client.vita_partner&.coalition || client.vita_partner&.parent_organization&.coalition
```
jenny-heath, line 159: nit: i think this would read better if it were based on `client.vita_partner.organization?` / `client.vita_partner.site?` (same thing, just expressed differently)

jenny-heath (approved):
looks good!

PR #985, opened 2021-04-05
Create notification when tagging a user in a note

bytheway875:
Screen Shot 2021-04-05 at 5 04 15 PM

### DIFF
--- a/app/controllers/hub/notes_controller.rb
+++ b/app/controllers/hub/notes_controller.rb
@@ -16,6 +16,7 @@ module Hub
     def create
       return render :index unless @note.save
 
+      create_notifications if params.dig(:note, :mentioned_ids).present?
       redirect_to hub_client_notes_path(client_id: params[:client_id], anchor: "last-item")
     end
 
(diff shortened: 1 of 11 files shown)


app/controllers/hub/notes_controller.rb
```diff
@@ -24,5 +25,10 @@ def create
     def note_params
       params.require(:note).permit(:body).merge(user: current_user, client: @client)
     end
+
+    def create_notifications
+      mentioned_user_ids = params.dig(:note, :mentioned_ids).split(",")
+      mentioned_user_ids.each { |id| UserNotification.create(notifiable: @note, user_id: id) }
```
embarnard, line 31: I noticed in the Notes model it doesn't have the same association that tax_return_assignment has to notifiable with `has_one :user_notification, as: :notifiable`

app/views/hub/user_notifications/_tax_return_assignment.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="notification">
   <div class="notification__heading"><%= t(".title") %></div>
-  <div class="notification__date"><%= long_formatted_datetime(notification.created_at, use_day: false) %></div>
+  <div class="notification__date"><%= timestamp(notification.created_at) %></div>
```
embarnard, line 3: nice catch

embarnard (approved):
Looks good!

app/controllers/hub/notes_controller.rb
bytheway875: @embarnard Are we using that association anywhere? I think here it might have to be a has_many (since many users can be tagged in the same note). I'm going to try to get away without it for now and then if we end up needing it, I'll add it in.

PR #986, opened 2021-04-05
Give all greeters access to orgs that allow greeters

embarnard:


app/controllers/users/invitations_controller.rb
```diff
@@ -53,12 +53,8 @@ def load_and_authorize_role
       when ClientSuccessRole::TYPE
         ClientSuccessRole.new
       when GreeterRole::TYPE
-        greeter_params = params.require(:greeter_organization_join_record).permit(organization_ids: []).merge(
-          params.require(:greeter_coalition_join_record).permit(coalition_ids: [])
-        )
         GreeterRole.new(
-          coalitions: @coalitions.where(id: greeter_params[:coalition_ids]),
-          organizations: @vita_partners.organizations.where(id: greeter_params[:organization_ids]),
+          organizations: @vita_partners.organizations.where(allows_greeters: true),
```
bytheway875, line 57: This is going to assign the greeter role to all users who have allows_greeters set at that moment in time. But when this list changes, the list of organizations this greeter has access to _wont_ change. Instead, I think we want the greeter role to act like the admin role -- we won't save any relationships to it at all right now. Instead, we should restrict the clients the user has access to through the `accessible_organizations` method on User model.

caitbot: That is a really great point! Will update to use `accessible_organizations` 👍

embarnard:
Still have to backfill allows_greeters on [VitaPartner for the correct orgs](https://docs.google.com/spreadsheets/d/1PhN7PC05bkaqmLN6y8yXUeGzr8URLlWfFbMWvG1hMiY/edit#gid=0) on prod/demo and drop the GreeterOrg/CoalitionJoins models but will do this in a separate release

db/migrate/20210407003612_change_default_valuefor_allows_greeters.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDefaultValueforAllowsGreeters < ActiveRecord::Migration[6.0]
+  def change
+    change_column_default(:vita_partners, :allows_greeters, from: nil, to: true)
```
alaroia-cfa, line 3: I'm surprised you want the default to be `true`. Is that OK with product?

db/schema.rb
```diff
@@ -70,12 +70,6 @@
     t.datetime "updated_at", precision: 6, null: false
   end
 
-  create_table "bulk_edits", force: :cascade do |t|
```
alaroia-cfa, line 73: I'm surprised you're removing `bulk_edits` from the schema. Is that peaceful?

spec/controllers/hub/organizations_controller_spec.rb
```diff
@@ -181,6 +181,7 @@
           name: "Oregano Organization",
           timezone: "America/Chicago",
           capacity_limit: "200",
+          allows_greeters: "true",
```
alaroia-cfa, line 184: This test sends `allows_greeters: "true"` in the POST, but since the factory will create the organization with `allows_greeters: "true"`, it seems to have no effect. Does it make sense to adjust the test to set `allows_greeters` to `false`?

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -219,22 +219,16 @@
       end
 
       context "inviting a greeter" do
-        let(:coalition) { create(:coalition) }
-        let(:coalition_2) { create(:coalition) }
-        let(:organization) { create(:organization) }
+        let!(:org_1) { create(:organization, allows_greeters: true) }
+        let!(:org_2) { create(:organization, allows_greeters: true) }
```
alaroia-cfa, line 223: What's the significance of creating these organizations? I think this test would work fine without any orgs, so consider removing them.

spec/features/hub/edit_organization_spec.rb
```diff
@@ -22,13 +22,15 @@
 
         select "Central Time (US & Canada)", from: "Timezone"
         fill_in "Capacity limit", with: "200"
+        check "Allows Greeters"
```
alaroia-cfa, line 25: I think this will have no effect (it'll already be checked) since the default is true.

spec/helpers/vita_partner_helper_spec.rb
```diff
@@ -115,26 +115,26 @@
...
     context "when the user's role is Greeter" do
-      let!(:site) { create :site, parent_organization: first_org }
       let!(:greeter) { create :greeter_user }
-      let(:first_org) { create :organization }
-      before do
-        greeter.role.update(organizations: [first_org, (create :organization)] )
-      end
+      let(:first_org) { create :organization, allows_greeters: true }
+      let!(:site) { create :site, parent_organization: first_org }
+      let(:second_org) { create :organization, allows_greeters: true }
```
alaroia-cfa, line 121: Consider adding a `allows_greeters: false` organization too, to test the full behavior of `.accessible_vita_partners` in this case.

db/migrate/20210407003612_change_default_valuefor_allows_greeters.rb
embarnard: Yeah I asked Kelly about this and she said it should default to true. It will still be a checkbox option on the create org page but will be selected upon the page load. Since greeters are currently greeting for almost every organization, except for a few that have requested no greeters I think it makes sense to set true as the default.

spec/lib/ability_spec.rb
```diff
@@ -149,6 +149,15 @@
...
           it_behaves_like :cannot_manage_any_sites_or_orgs
         end
+
+        context "a greeter" do
+          let(:user) { create :greeter_user }
+
+          it_behaves_like :can_manage_accessible_client
+          it_behaves_like :cannot_manage_inaccessible_client
+          it_behaves_like :can_only_read_accessible_org_or_site
+          it_behaves_like :cannot_manage_any_sites_or_orgs
```
alaroia-cfa, line 159: Stellar!

spec/models/greeter_role_spec.rb
```diff
@@ -10,33 +10,10 @@
 
 RSpec.describe GreeterRole, type: :model do
   describe "required fields" do
```
alaroia-cfa, line 12: I think there are now no fields at all, so consider deleting this `describe` block.

spec/features/hub/edit_organization_spec.rb
embarnard: Oh good catch! Yeah my bad I made these changes before the default was set to true

alaroia-cfa (changes requested):
Great work here! I have some questions here, most of which are not blockers. The seemingly spurious change to schema.rb does seem worth blocking on. Happy to chat if that's helpful, or happy to have you revise and push without further input (so long as you clean up the spurious schema changes.)

db/schema.rb
embarnard: Yeah I'm not sure where this came from... maybe from one of my rebases. I will remove it

alaroia-cfa:
@embarnard One more question - can a _site_ opt out of greeting, or just an _organization_ ? I haven't honestly dug into that question w/r/t this PR, but I want to bring it up, since the site vs. organization distinction is a bit challenging in our codebase, and I want to make sure whatever we do with greeters and that is something product is happy with.

embarnard:
> @embarnard One more question - can a _site_ opt out of greeting, or just an _organization_ ? I haven't honestly dug into that question w/r/t this PR, but I want to bring it up, since the site vs. organization distinction is a bit challenging in our codebase, and I want to make sure whatever we do with greeters and that is something product is happy with.

@alaroia-cfa Yes! I asked product about the coaltion/org/site specific associations for greeters and they said for now the allows_greeters access should only be determined on the org level. To Kelly's knowledge, there is not a case where some sites within an organization want greeters and some do not.

alaroia-cfa:
> I asked product about the coaltion/org/site specific associations for greeters and they said for now the allows_greeters access should only be determined on the org level. To Kelly's knowledge, there is not a case where some sites within an organization want greeters and some do not.

In that case, consider a model validation on `VitaPartner` that it's invalid if it's a site **and** `allows_greeters` is False.

db/schema.rb
embarnard: @alaroia-cfa do I have to revert the rebase or can I just copy-pasta it back in?

alaroia-cfa: Copy-pasta is fine. If you are going to be doing a squash merge in GitHub, then the only important thing is the diff.

embarnard:
> > I asked product about the coaltion/org/site specific associations for greeters and they said for now the allows_greeters access should only be determined on the org level. To Kelly's knowledge, there is not a case where some sites within an organization want greeters and some do not.
> 
> In that case, consider a model validation on `VitaPartner` that it's invalid if it's a site **and** `allows_greeters` is False.

@alaroia-cfa Oh right that's a good idea! There shouldn't be anything that checks for sites' `allows_greeters` field but it makes sense to fill it for consistency

alaroia-cfa:
> There shouldn't be anything that checks for sites' allows_greeters field but it makes sense to fill it for consistency

Yeah - whether or not you intend it, since the default is `TRUE` for `vita_partners.allows_greeters`, every site with have that field as true, so it might be good to insist no one sets it false.

PR #987, opened 2021-04-06
Add stimulus payment question

bytheway875:
- Add question to QuestionNavigationFlow
- Add Question to 13614C
- Set up question in layout
- Add attribute to intake
- Create form for question

Screen Shot 2021-04-06 at 10 48 12 AM


caitbot (commented):
Waiting for response from product since there are UX implications for merging this.

bytheway875:
Product wants us to skip sending clients back to stimulus-payment if they're returning -- they can get the answer on the intake call.

PR #999, opened 2021-04-10
Only send messages to accessible clients

bengolder:
bugfix for https://www.pivotaltracker.com/story/show/177351982
- also handles the case when all clients have a `nil` locale.


embarnard (approved):
Looks good to me!

PR #1000, opened 2021-04-12
Create confirmation page for Client documents that will be displayed to the client

bytheway875:
This removes the multi-select for uploads and limits to uploading one doc at a time.

### DIFF
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -299,10 +299,22 @@
     flex-shrink: 0;
   }
   .item {
-    margin-right: 60px;
+    margin-right: 6rem;
   }
   .item:last-of-type {
     margin-right: 0;
+  }
+
+  .item-15r {
+    margin-right: 1.5rem;
+  }
+
+  &.full-height, .full-height {
+    height: 100vh;
+  }
+
+  &.column-column {
+    flex-direction: column;
   }
 }
 
--- a/app/views/hub/documents/confirm.html.erb
+++ b/app/views/hub/documents/confirm.html.erb
@@ -21,7 +21,7 @@
           </div>
           <div class="item">
             <%= form_with  %>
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -299,10 +299,22 @@
     flex-shrink: 0;
   }
   .item {
-    margin-right: 60px;
+    margin-right: 6rem;
```
alaroia-cfa, line 302: Nice

app/views/hub/documents/confirm.html.erb
```diff
@@ -0,0 +1,39 @@
...
+          <%= t("hub.documents.confirm.text",
+                document_type: @document.document_type,
+                client_name: @document.client.legal_name,
+                filing_year: @document.tax_return.year
+                )
+          %>
+        </p>
+        <div class="grid-flex">
+          <div class="item-15r">
+            <%= link_to hub_client_documents_path(client_id: @document.client.id), class: "button button--icon button--icon--centered", "data-track-click": "document-confirmed-yes" do %>
```
alaroia-cfa, line 18: Nice event here. It'll be interesting to see the yes vs. no data in Mixpanel.

app/views/hub/documents/confirm.html.erb
```diff
@@ -0,0 +1,39 @@
...
+              <%= image_tag("checkmark.svg", alt: "") %><%= t("general.affirmative") %>
+            <% end %>
+          </div>
+          <div class="item">
+            <%= link_to edit_hub_client_document_path(id: @document.id), class: "button button--icon button--icon--centered", "data-track-click": "document-confirmed-no" do %>
```
alaroia-cfa, line 23: The doc will already be saved at this point. Will Hub users know what to do to archive the doc? Is it the case that clients will never see archived docs?

alaroia-cfa (approved):
This code is good! I am not sure if the feature will make sense to Hub users - since the doc remains saved (as far as I can tell) even if the Hub user clicks "No," I think I'd like to see a different flow.

I'm OK with the code, so if you're OK with the flow as-is, 🚢

PR #1001, opened 2021-04-12
Add method to send system message for all opted in contact methods in client messaging service

embarnard:
Co-authored-by: Ben Golder <bgolder@codeforamerica.org>


bengolder (approved):
Approved! This is a great way to get the system messages to better respect contact methods in the short term, and we discussed that there is potential for further refactoring here.

PR #1002, opened 2021-04-12
Remove stray } when displaying call notes

alindeman:
Hi folks!

I'm a VITA volunteer and noticed that there seemed to be a trailing `}` when displaying call notes. I think I tracked it down to this section of code which looks like it was just a inadvertent extra `}` in the variable interpolation string.

Screenshot 2021-04-12 14 58 45

I debated whether modifying the test was even worthwhile. Feedback welcomed.


bytheway875 (approved):
👍  Thank you!

PR #1005, opened 2021-04-13
Create triage flow.

bytheway875:
The changes look more extreme than they actually are. There are a lot of SVG additions and deletions, and some "changes" to SVGs because I ran them through svg-omg to compress them for web-readiness (removing extraneous markup).

Some notes: 

- Answers to questions in the triage flow are reported to mixpanel but they are NOT saved in the database.
- The triage controllers inherit from the TriageController which inherits from the AnonymousIntakeController. This allows us to set up shared controllers only once and deal with sending data to mixpanel in a clean and consistent manner.


app/assets/stylesheets/templates/_question.scss
```diff
@@ -24,4 +24,7 @@
       }
     }
   }
+  .question-wrapper {
```
bytheway875, line 27: Anu wants to limit all intake pages to a max width of 570px instead of a % of the width of the container.

coltborg (approved):
Reviewed this PR with @bytheway875, looks good to go! :shipit:

app/assets/stylesheets/templates/_question.scss
coltborg: We moved this to a variable: `$intake-max-width`.

PR #1007, opened 2021-04-13
Refactor portal home template and portal controller

embarnard:

### DIFF
--- a/app/views/portal/portal/home.html.erb
+++ b/app/views/portal/portal/home.html.erb
@@ -2,89 +2,52 @@
 
 <% content_for :card do %>
   <h1 class="h2"><%= t(".title", name: current_client.intake.preferred_name) %></h1>
-
-  <% if @current_step.present? %>
+  <% if @heres_what_we_need %>
     <p><%= t(".heres_what_we_need") %></p>
+  <% end %>
+
+  <% if @submit_additional_documents %>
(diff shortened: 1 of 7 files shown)


app/views/portal/portal/home.html.erb
```diff
@@ -3,97 +3,105 @@
...
-              <% tax_return.final_tax_documents.each do |tax_doc| %>
-                <div class="tax-return-status__row tax-return-status__solo">
-                  <%= image_tag("icons/check.svg", alt: "") %>
-                  <%= link_to t(".tax_returns.view_document", display_name: tax_doc.display_name), portal_document_path(id: tax_doc.id), target: "_blank" %>
-                </div>
-              <% end %>
-            <% end %>
-          <% end %>
+          <% messages.each do |message| %>
+            <% tax_return = current_client.tax_returns.where(year: year).first %>
```
embarnard, line 30: I think this would be easier to read if you start by looping across the tax_returns - that way you can avoid the `<% tax_return = current_client.tax_returns.where(year: year).first %>`.

In that case, you might want to do something like:

`<% if @tax_return_messages[tax_return.year].include?(:final_tax_document) %>`

to get the same effect as the loop over messages

alaroia-cfa (approved):
Super rad!

PR #1008, opened 2021-04-13
Application layout for GYR emails

bytheway875:
Screen Shot 2021-04-13 at 12 08 48 PM


app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,41 @@
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <style>
-      /* Email styles need to be inline */
+      * {
```
alaroia-cfa, line 6: Consider using the `body` selector here instead of `*`.

`*` won't work in as many email clients, according to this page: https://templates.mailchimp.com/resources/email-client-css-support/

app/views/layouts/mailer.html.erb
```diff
@@ -3,11 +3,41 @@
...
+        margin-bottom: 2.5rem;
+      }
     </style>
   </head>
 
   <body>
-    <%= yield %>
+    <div class="wrapper">
+      <div class="logo">
+        <%= image_tag "GYR_logo.png", style: "width:180px;height:auto" %>
```
alaroia-cfa, line 35: Would it be OK to pick one of adding inline style here _or_ adding the image style to the CSS? If the key is to always style this image, even in the face of abysmal email clients which don't support `<style>`, then I'd say pick only the inline style. Not a blocker.

config/environments/shared_deployment_config.rb
```diff
@@ -31,7 +31,7 @@
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
   config.assets.compile = false
-
+
```
alaroia-cfa, line 34: Irrelevant whitespace change; truly not a blocker, but I saw it, so there you go.

app/views/layouts/mailer.html.erb
alaroia-cfa, line 35: Does this image tag involve a ``? If so, then email users with email image privacy features enabled won't see the image. That's OK, but honestly I love email image privacy features, and personally want everyone to enable them, and have them enabled myself, and would like to see GYR emails w/ the logo.

To get the logo into the message itself, you can use a reference like this: http://www.lindsaar.net/2010/6/7/inline-attachments-for-actionmailer

Not a blocker, but I love it, so there you go.

alaroia-cfa: Blocker: When I use this code on my laptop, I see the IMG tag has a `src="/assets/...` path, which is not an absolute URL. This will mean it will look like it works in the preview you saw, but actually never show an image in a client's email client. If you stick to using the URL to our site (which is OK), make sure you use an absolute URL not a path.

alaroia-cfa (changes requested):
1 blocker: make sure the IMG URL is either a full URL or a `cid:...` inline attachment URL. That's the only blocker; feel free to fix that and push. (Happy to re-review if you prefer.)

Also additional comments. Paired w/ @coltborg on this!

app/views/layouts/mailer.html.erb
bytheway875: switched to the inlined attachment, but also set it up so that we're using asset_host for emails in config for demo, staging, and prod.

config/environments/shared_deployment_config.rb
bytheway875: i edited this line and now my editor is being stubborn about adding this tab on the empty line... tried for 10 seconds to remove it and it didnt work so im giving up :D

app/views/layouts/mailer.html.erb
bytheway875: done! left inline to ensure it doesnt end up huge on a wonky client.

config/environments/production.rb
```diff
@@ -6,6 +6,8 @@
   config.action_mailer.default_options = { from: "hello@getyourrefund.org" }
   config.address_for_transactional_authentication_emails = 'no-reply@getyourrefund.org'
   config.action_mailer.default_url_options = { host: 'www.getyourrefund.org' }
+  config.action_mailer.asset_host = "https://getyourrefund.org"
```
alaroia-cfa, line 9: I think `https://getyourrefund.org/*` redirects to `https://www.getyourrefund.org/*`, based on visiting this URL https://getyourrefund.org/assets/checkbox-logo--white-252f6970859e7b1ba37367e7820af3472b45415066bb501fae76813011fde48e.png and seeing it redirect. It's nicer if we don't redirect people, so I'll probably submit a trivial PR to make this say `www.`

It's also not a blocker.

spec/features/hub/authenticate_spec.rb
```diff
@@ -62,7 +62,7 @@
     # Expect subject to
     expect(email.subject).to eq("Reset password instructions")
 
-    html_body = email.body.parts[1].decoded
+    html_body = email.body.encoded
```
alaroia-cfa, line 65: I'm surprised you need to use the encoded version of the email, rather than the decoded version here. I would expect the decoded version to have a valid link and the encoded one maybe not to.

Not a blocker, but curious what motivated the change.

spec/features/portal/client_login_spec.rb
```diff
@@ -32,8 +32,7 @@
           perform_enqueued_jobs
 
           mail = ActionMailer::Base.deliveries.last
-          text_body = mail.body.parts[0].decoded
-          expect(text_body).to include("Your 6-digit GetYourRefund verification code is: 000004. This code will expire after two days.")
+          expect(mail.html_part.body.to_s).to include("Your 6-digit GetYourRefund verification code is: 000004. This code will expire after two days.")
```
alaroia-cfa, line 35: I like this `mail.html_part.body.to_s` formulation!

alaroia-cfa (commented):
This is a great change!

I wonder if the `.asset_host` is bad not good -- if we want to continue down the path of always including all assets within the email, then why have an asset_host at all?

On the assumption we'll keep it, I'll submit a trivial PR to make it be `www.` in production.

PR #1009, opened 2021-04-13
Create client portal messages

bytheway875:
Screen Shot 2021-04-13 at 2 00 52 PM
Screen Shot 2021-04-12 at 5 47 14 PM


app/helpers/contact_record_helper.rb
```diff
@@ -16,6 +16,8 @@ def message_heading(contact_record)
       "#{I18n.t("hub.messages.from")} #{contact_record.from}"
     when :outgoing_email
       "#{I18n.t("hub.messages.to")} #{contact_record.to}"
+    when :incoming_portal_message
```
embarnard, line 19: is it obvious that this is from the client if it just says from the client portal? Maybe redundant but would also mimic the structure of the other messages if included their name

embarnard (approved):
Looks good! No blocking comment

app/views/portal/portal/home.html.erb
```diff
@@ -92,8 +92,12 @@
...
-    <div class="spacing-above-60">
+  <% end %>
+
+  <div class="spacing-above-60">
+    <% if @tax_returns.present? %>
       <%= link_to t(".submit_documents"), portal_upload_documents_path, class: "button button--wide" %>
-    </div>
-  <%end %>
+    <% end %>
+    <%= link_to t(".send_message"), new_portal_message_path, class: "button button--wide button--primary" %>
```
embarnard, line 101: the placement of the button might change a little with the changes from this refactoring branch that was merged https://github.com/codeforamerica/vita-min/pull/1007

app/helpers/contact_record_helper.rb
bytheway875: It will include their name! See screenshot. This is just what gets appended after the name.

PR #1014, opened 2021-04-15
Greeter flow statuses

bytheway875:
Screen Shot 2021-04-15 at 10 54 45 AM

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -4,7 +4,7 @@ module ClientSortable
     setup_sortable_client unless @filters.present?
     clients = if current_user&.greeter?
               # Greeters should only have "search" access to clients in intake stage AND clients assigned to them.
-                @clients.in_intake.or(Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct)
+                @clients.greetable.or(Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct)
               else
                 @clients.after_consent
               end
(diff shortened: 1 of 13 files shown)


alaroia-cfa:
BG & I reviewed & refactored this. I plan to merge on Monday after talking to product,
to confirm they're OK with us removing the blank status option. If users choose a blank
status, they get error messages, so I hope product is OK with that!

While reviewing, we came up with some future refactor ideas that we wanted to share.
- We think the only reason we use `distinct.` is to `.or.` it with another query that is `distinct`. If that's right, one option is to remove the `.distinct` call from `.greetable`. In that case, we could call `.distinct` after the `.or.` in `filtered_and_sorted_clients`.
- Another option is to replace `.greetable` with a `.greetable_by(user)` scope, to encapsulate the `.or` that happens in `filtered_and_sorted_clients`.
- It struck us that `browseable` is a possibly better name than `greetable`, since access is still allowed to other clients.
- If we lean on the `browseable` idea, we could have `Client.browseable_by(user)` scope that *further* encapsulates what `filtered_and_sorted_clients` is doing. This could work for greeters and non-greeters!
- We noticed the `.values.flatten` is a little complex, and wondered if a GREETABLE_STATUSES constant used for the querying could simplify things.
- We noticed the `grouped_status_options_for_select` tests could validate that the method translates status labels properly (which it already does!). One easy way to do that is to change the locale to `es` in the spec.
- We could make a helper that creates the select options so that `_client_filters_and_add_client.html.erb` can use f.select.

Great work! This is going to make greeters' lives easier! Well, it's really going to make greeting _possible!_

PR #1016, opened 2021-04-15
Add hero image to full-service public page #177564220

coltborg:
[Add hero image to full-service public page #177564220](https://www.pivotaltracker.com/story/show/177564220)

The image was very similar to the image in `questions/welcome.svg`, but not exactly the same. So I added this new image as `full-service-hero.svg`. Following a similar pattern for the home page hero, although this is just on SVG image for both mobile and desktop.

Screen Shot 2021-04-15 at 4 29 01 PM

Screen Shot 2021-04-15 at 4 29 10 PM


alaroia-cfa (approved):
Easy and great

PR #1032, opened 2021-04-21
Restrict zip code routing to organizations with capacity for new clients

bytheway875:
Rae requested that we take capacity into account when routing by zip code. The query is the same as how we restrict by capacity for state routing. Expand the diff to compare.


caitbot (approved):
👍

PR #1033, opened 2021-04-22
send metric to datadog for outbound call queue times

jenny-heath:
i would specifically like feedback on the datadog attribute name which i have currently called `twilio.outbound_calls.queue_time_ms`

note: just want to call out that it _is_ redundant to be saving this in the db as well as sending it to datadog - happy to hear opinions on this, otherwise will assume this redundancy doesn't hurt

### DIFF
--- a/app/forms/hub/outbound_call_form.rb
+++ b/app/forms/hub/outbound_call_form.rb
@@ -42,6 +42,7 @@ module Hub
         )
 
         DatadogApi.increment("twilio.outbound_calls.initiated")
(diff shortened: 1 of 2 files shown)


bytheway875:
I don't mind the redundancy... Right now it seems like Datadog information is somewhat more meaningful/useful, but its also potentially helpful to be able to dig down on a per-record basis.

bytheway875 (approved):

bytheway875 (approved):

PR #1034, opened 2021-04-23
Zip code routing

bytheway875:
Some notes:

- This PR also refactors the Unique Links to use the same pattern. This is both for UX consistency but also for the practical reason that something about the existing fields_for implementation was causing the feature tests to fail when I was running using `js: true`. 🔍 mysterious!
- For both source params and zip codes, if the param or zip code is in use by another partner, the error message links to the partners page in case a change needs to be made on that other partner.
- Uses js to do the adding / removing because it requires less scrolling AND because this is used for both organizations and sites and handling redirecting back to different locations based on whether the partner is an org or site added complexity to the controllers. I added an `aria-live="assertive"` tag and an sr-only success message to alert screen readers of a newly added object when the form re-renders using javascript.

### DIFF
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -94,6 +94,8 @@
   .grid {
     margin-left: 0;
     margin-right: 0;
+    max-width: unset;
+    width: 100%;
   }
 
   .slab {
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -21,7 +21,9 @@ module Hub
     end
 
     def show
-      @organization = VitaPartner.organizations.find(params[:id])
+      @organization = @vita_partner
+      raise ActionController::RoutingError.new('Not Found') unless @vita_partner.organization?
+
       @sites = @organization.child_sites
     end
 
@@ -33,6 +35,8 @@ module Hub
 
     def edit
       @coalitions = Coalition.all
+      @routing_form = ZipCodeRoutingForm.new(@vita_partner)
+      @source_params_form = SourceParamsForm.new(@vita_partner)
       @organization = @vita_partner
     end
 
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -21,9 +21,9 @@ module Hub
     end
 
     def show
-      @organization = @vita_partner
       raise ActionController::RoutingError.new('Not Found') unless @vita_partner.organization?
 
+      @organization = @vita_partner
       @sites = @organization.child_sites
     end
 
--- a/app/controllers/hub/source_params_controller.rb
+++ b/app/controllers/hub/source_params_controller.rb
@@ -8,8 +8,8 @@ module Hub
       vita_partner = VitaPartner.find(params[:vita_partner_id])
       @form = form_class.new(vita_partner, permitted_params)
       if @form.valid?
-        source_param = @form.save
-        @success_message = I18n.t("hub.source_params.success", code: source_param.code, name: vita_partner.name)
+        @form.save!
+        @success_message = I18n.t("hub.source_params.success", code: @form.code, name: vita_partner.name)
       else
         flash.now[:alert] = @form.error_summary
       end
--- a/app/controllers/hub/source_params_controller.rb
+++ b/app/controllers/hub/source_params_controller.rb
@@ -1,0 +1,42 @@
+module Hub
+  class SourceParamsController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    authorize_resource :vita_partner, parent: false, only: :create
+
+    def create
+      vita_partner = VitaPartner.find(params[:vita_partner_id])
+      @form = form_class.new(vita_partner, permitted_params)
+      if @form.valid?
+        source_param = @form.save
+        @success_message = I18n.t("hub.source_params.success", code: source_param.code, name: vita_partner.name)
+      else
+        flash.now[:alert] = @form.error_summary
+      end
+      respond_to do |format|
+        format.js
+      end
+    end
+
+    def destroy
+      @source_param = SourceParameter.find_by(id: params[:id])
+
+      unless @source_param.present?
+        flash[:error] = I18n.t("hub.source_params.not_found")
+        redirect_to request.referrer and return
+      end
+      @source_param.destroy!
+      respond_to do |format|
+        format.js
+      end
+    end
+
+    def form_class
+      SourceParamsForm
+    end
+
+    def permitted_params
+      params.require(form_class.form_param).permit(:code)
+    end
+  end
+end
--- a/app/controllers/hub/source_params_controller.rb
+++ b/app/controllers/hub/source_params_controller.rb
@@ -19,12 +19,8 @@ module Hub
     end
 
     def destroy
-      @source_param = SourceParameter.find_by(id: params[:id])
+      @source_param = SourceParameter.find(params[:id])
 
-      unless @source_param.present?
-        flash[:error] = I18n.t("hub.source_params.not_found")
-        redirect_to request.referrer and return
-      end
       @source_param.destroy!
       respond_to do |format|
         format.js
--- a/app/views/hub/zip_codes/_form.html.erb
+++ b/app/views/hub/zip_codes/_form.html.erb
@@ -1,6 +1,6 @@
 <h2>Zip Code Routing</h2>
 <% if @success_message %>
-  <p aria-live="assertive" class="sr-only"><%= @success %></p>
+  <p aria-live="assertive" class="sr-only"><%= @success_message %></p>
 <% end %>
 
 <ul class="organization-list">
--- a/app/views/hub/zip_codes/_form.html.erb
+++ b/app/views/hub/zip_codes/_form.html.erb
@@ -1,0 +1,23 @@
+<h2>Zip Code Routing</h2>
+<% if @success_message %>
+  <p aria-live="assertive" class="sr-only"><%= @success %></p>
+<% end %>
+
+<ul class="organization-list">
+  <% form.vita_partner.serviced_zip_codes.each do |szc| %>
+  <% next unless szc.id.present? %>
+    <li id="<%= "zip-code-routing-rule-#{szc.id}" %>">
+      <%= szc.zip_code %> <%= szc.city_state %>
+      <%= link_to t("general.delete"), hub_zip_code_path(id: szc.id), method: :delete, remote: true, "data-confirm": I18n.t("hub.zip_codes.confirm", code: szc.zip_code, name: form.vita_partner.name), class: "button button--small spacing-below-0" %>
+    </li>
+  <% end %>
+</ul>
+
+<%= form_with model: form, url: hub_zip_codes_path(vita_partner_id: form.vita_partner_id), builder: VitaMinFormBuilder do |f| %>
+  <div>
+    <%= f.cfa_input_field :zip_code, "Zip code", classes: ["form-width--short"] %>
+    <div>
+      <%= f.submit t("general.save"), class: "button button--cta button--wide" %>
+    </div>
+  </div>
+<% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -17,6 +17,7 @@ en:
       already_taken: "%{zip_code} is already routed to <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
       already_applied: "%{zip_code} is already routed to this partner."
       not_found: Zip code not found. Try again.
+      success: Added %{code} to %{name}.
       confirm: Are you sure you want to delete %{code} from %{name}?
     source_params:
       already_taken: "%{code} is already in use by <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -13,6 +13,17 @@ en:
       return_to_profile: Return to Profile
       my_clients: My Clients
       my_profile: My Profile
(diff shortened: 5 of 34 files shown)


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -94,6 +94,8 @@
   .grid {
     margin-left: 0;
     margin-right: 0;
```
bytheway875, line 96: The grid is restricted to 960px in honeycrisp, but my expectation when there are css classes like "width-one-half" is that the grid would be full width of its parent and then you can restrict that width if necessary by adding a wrapper class to the parent with a max width. For the purposes of our app where the default is full-width pages, unsetting the inherited max-width makes our width-one-half and width-two-thirds classes truly 1/2 and 2/3 of the page respectively.

app/controllers/hub/organizations_controller.rb
```diff
@@ -21,7 +21,9 @@ def create
     end
 
     def show
-      @organization = VitaPartner.organizations.find(params[:id])
+      @organization = @vita_partner
+      raise ActionController::RoutingError.new('Not Found') unless @vita_partner.organization?
```
bytheway875, line 25: Still raises an error when you can't find the organization _without_ doing a second database lookup for the organization.

app/views/hub/zip_codes/_form.html.erb
```diff
@@ -0,0 +1,23 @@
+<h2>Zip Code Routing</h2>
+<% if @success_message %>
+  <p aria-live="assertive" class="sr-only"><%= @success_message %></p>
+<% end %>
+
+<ul class="organization-list">
+  <% form.vita_partner.serviced_zip_codes.each do |szc| %>
+  <% next unless szc.id.present? %>
```
alaroia-cfa, line 8: Would this case ever occur? I can't see a reason it would, so I encourage you to delete this line.

config/locales/en.yml
```diff
@@ -13,6 +13,18 @@ en:
       return_to_profile: Return to Profile
       my_clients: My Clients
       my_profile: My Profile
+    zip_codes:
+      already_taken: "%{zip_code} is already routed to <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
```
alaroia-cfa, line 17: Consider quotes around the parameter to href, e.g.

```
      already_taken: '%{zip_code} is already routed to <a href="%{vita_partner_path}">%{vita_partner_name}</a>.'
```

config/locales/en.yml
```diff
@@ -13,6 +13,18 @@ en:
...
+      already_applied: "%{zip_code} is already routed to this partner."
+      not_found: Zip code not found. Try again.
+      success: Added %{code} to %{name}.
+      confirm: Are you sure you want to delete %{code} from %{name}?
+    source_params:
+      already_taken: "%{code} is already in use by <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
```
alaroia-cfa, line 23: Same issue as above about quotes for the parameter to HREF

config/locales/es.yml
```diff
@@ -13,6 +13,18 @@ es:
       return_to_profile: Volver al Perfil
       my_clients: Mis Clientes
       my_profile: Mi Perfil
+    zip_codes:
+      already_taken: "%{zip_code} ya está enrutado a <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
```
alaroia-cfa, line 17: Same issue about quotes when passing parameters to HREF

config/locales/es.yml
```diff
@@ -13,6 +13,18 @@ es:
...
+      already_applied: "%{zip_code} ya está enrutado a este socio."
+      not_found: Código postal no encontrado. Intentar otra vez.
+      success: Se agregó %{code} a %{name}.
+      confirm: ¿Estás segura de que quieres eliminar? %{code} de %{name}?
+    source_params:
+      already_taken: "%{code} ya está siendo utilizado por <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
```
alaroia-cfa, line 23: Same issue about quotes when passing parameters to HREF

alaroia-cfa: This makes me think one day we might want to add https://github.com/fnando/i18n-js to our stack. No rush; just mentioning it in case you have ideas about it.

spec/controllers/hub/organizations_controller_spec.rb
```diff
@@ -75,7 +75,7 @@
         it "is not found" do
           expect do
             get :show, params: params
-          end.to raise_error(ActiveRecord::RecordNotFound) # in deployment configs, this would be a 404
+          end.to raise_error(ActionController::RoutingError) # in deployment configs, this would be a 404
```
alaroia-cfa, line 78: I'd prefer this be RecordNotFound since I think that's more accurate. Alternatively, your custom code could respond with a real 404 status code, and we could adjust this test accordingly.

Not a blocker.

spec/controllers/hub/organizations_controller_spec.rb
```diff
@@ -166,6 +167,15 @@
           expect(response.body).to include("shortlink2")
         end
       end
+
+      context "zip code routings" do
+        it "displays existing routings and prepares the form" do
```
alaroia-cfa, line 172: 💪

spec/controllers/hub/source_params_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      before do
+        sign_in admin_user
+      end
+
+      it "responds with js" do
+        post :create, params: params, format: :js, xhr: true
+        expect(response.media_type).to eq "text/javascript"
+      end
+
+      it "does not respond with html" do
```
alaroia-cfa, line 30: Awesome

This might make a good shared example someday.

spec/controllers/hub/source_params_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    it "does not respond with html" do
+      expect { delete :destroy, params: params }.to raise_error ActionController::UnknownFormat
+    end
+
+    context "when the source parameter does not exist" do
+      let(:id) { "fake" }
+
+      before do
+        request.env['HTTP_REFERER'] = "https://www.getyourrefund.org/welcome"
```
alaroia-cfa, line 82: It might be nice to have a test case for a situation where the referrer header is missing. I suspect writing that test will be annoying and boring. Given that only admins will use this, not a blocker.

spec/controllers/hub/source_params_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+
+      it "flashes an error and redirects" do
```
alaroia-cfa, line 85: Dumb question - given that the response type is JS, will the browser actually do the redirect? My guess would be no, but maybe I don't know how Rails & JS work together.

Another approach if you want to retain the `:js` response type is to set up the flash message on the server, then send some JS that triggers a redirect, presumably with `window.location.href = ...`.

spec/controllers/hub/source_params_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        delete :destroy, params: params, format: :js, xhr: true
+        expect(response).to redirect_to "https://www.getyourrefund.org/welcome"
```
alaroia-cfa, line 87: I'd also like to see the flash message tested in this test case.

spec/controllers/hub/zip_codes_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+        it "does not make a new source parameter" do
+          expect {
+            post :create, params: params, format: :js, xhr: true
+          }.not_to change(VitaPartnerZipCode, :count)
+        end
+      end
+    end
+
+    context "other user types" do
```
alaroia-cfa, line 55: Consider using the `a_post_action_for_admins_only` shared example in this and the other controller action test below.

spec/features/hub/create_organization_hierarchy_spec.rb
```diff
@@ -28,26 +28,63 @@
...
+        select "Coati Coalition", from: "Coalition"
+        click_on "Save"
+      end
       expect(page).to have_selector("h1", text: "Oregano Org")
 
+      # adding / removing zip codes
+      within "#zip-code-routing-form" do
+        fill_in "Zip code", with: "94606"
+        click_on "Save"
+        Capybara.using_wait_time(10) do
```
alaroia-cfa, line 42: Did you find that you needed the increased wait time? Just curious - if so, we could consider raising it everywhere. :/

spec/features/hub/edit_organization_spec.rb
```diff
@@ -23,31 +23,41 @@
...
         expect(page).to have_select("Timezone", selected: "Central Time (US & Canada)")
         expect(find_field('Capacity limit').value).to eq "200"
         expect(find_field('Allows Greeters').value).to eq "true"
 
-        # Now do the same for the child site
+        within "#zip-code-routing-form" do
+          expect(page).to have_field("Zip code")
+          fill_in "Zip code", with: "94606"
+          click_on "Save"
+          # Now do the same for the child site
```
alaroia-cfa, line 40: IMHO move this comment out of the `within` and above the `click_on` on L44

spec/forms/hub/source_params_form_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      context "when it belongs to a different vita partner" do
+        let(:other_vita_partner) { create :vita_partner, name: "Oregano Org" }
+        before do
+          create :source_parameter, vita_partner: other_vita_partner, code: code
+        end
+
+        it "is invalid with appropriate message" do
+          expect(subject.valid?).to eq false
+          expect(subject.errors[:code]).to include "my_code is already in use by <a href=/en/hub/organizations/#{other_vita_partner.id}/edit#source-params-form>Oregano Org</a>."
```
alaroia-cfa, line 34: I'd like to see the expected HTML include a `"` around the URL in the A HREF.

spec/forms/hub/source_params_form_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    context "when the source parameter is unique and valid" do
+      it "is valid" do
+        expect(subject.valid?).to eq true
+      end
+    end
+  end
+
+  describe "#save!" do
+    it "creates a new source parameter for the vita partner" do
```
alaroia-cfa, line 47: I think it would be good to check the contents of the new source parameter in this test.

spec/forms/hub/zip_code_routing_form_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    context "when the zip code is unique and valid" do
+      it "is valid" do
+        expect(subject.valid?).to eq true
+      end
+    end
+  end
+
+  describe "#save!" do
```
alaroia-cfa, line 55: I think it would be good to validate _which_ ZIP code got stored in this test.

alaroia-cfa (commented):
Lots of great new code!

We're really leaning on Rails's `respond_to` `:js`, and I need to take some time to understand that better. That doesn't have to be a blocker.

I left some other comments as well. No other blockers, but I'd appreciate any changes you feel like making in response to those comments.

I need to validate that the `delete` stuff doesn't cause a cross-site request forgery issue before I click approve. I plan to do that approx immediately after lunch.

There's an error case where we redirect to the referrer & set a flash message. I don't see how that would work, so I'll test it approx immediately after lunch as well.

app/controllers/hub/organizations_controller.rb
bytheway875: Hmm... I actually disagree for the reason that we did already _find_ the record -- we just aren't allowing you to route to it because it's not an organization and thus does not fulfill the routing requirements for this controller method. I'm open to changing the order of the variable definitions, and I actually think it's not necessary at all to use @organization instead of @vita_partner in the template. To me, the main goal was to stop doing another find for a record we already had available to us.

app/controllers/hub/source_params_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    def create
+      vita_partner = VitaPartner.find(params[:vita_partner_id])
+      @form = form_class.new(vita_partner, permitted_params)
+      if @form.valid?
+        @form.save!
+        @success_message = I18n.t("hub.source_params.success", code: @form.code, name: vita_partner.name)
+      else
+        flash.now[:alert] = @form.error_summary
+      end
+      respond_to do |format|
```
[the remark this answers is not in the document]
bytheway875: We could. I think that's what's happening in the bank accounts controller. Tbh, it doesn't really make any difference other than removing a few lines of code.

app/controllers/hub/source_params_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        format.js
+      end
+    end
+
+    def destroy
+      @source_param = SourceParameter.find_by(id: params[:id])
+
+      unless @source_param.present?
```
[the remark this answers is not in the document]
bytheway875: In the case that there is not a valid id present, that means that you probably weren't using our interface in order to make the request -- you instead were attempting to go to the destroy url directly or some other way of making a delete request to that url. In that case, ejecting with an html response and redirect over js is because we can't do anything legitimate with the data you've provided and it's likely you aren't even looking at our interface.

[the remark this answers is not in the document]
bytheway875: It would be just as valid to raise an error or something i guess

spec/features/hub/create_organization_hierarchy_spec.rb
bytheway875: yes i did, intermittantly. We could raise it everywhere. It only has an impact if it actually needs to wait longer.

PR #1035, opened 2021-04-23
Fix run on sentence typo in notice

coltborg:
# What does this PR do?

- Fix run on sentence in en for home page notice
- Fix run on sentence in en for file with help notice


caitbot (approved):
Beaaautiful

PR #1036, opened 2021-04-23
Scope tax return documents to only un-archived docs

bytheway875:


spec/factories/documents.rb
```diff
@@ -45,6 +45,11 @@
       upload_path { Rails.root.join("spec", "fixtures", "attachments", "picture_id.jpg") }
     end
 
+    trait :pdf do
```
jenny-heath, line 48: (very) nit: `:pdf_attachment` could be a little more clear

jenny-heath (approved):

PR #1037, opened 2021-04-23
Remove db constraint for to on incoming emails

bytheway875:


jenny-heath (approved):
seems good! my only question is: we don't use this field anywhere for anything, right? i can't see why we would but that would be the only reason not to make this optional so just wanted to check

bytheway875:
@jenny-heath based on my searches, no!

PR #1038, opened 2021-04-23
Fix heading hierarchy on homepage

jenny-heath:
not 100% certain that turning each slab heading into an h2 was the right move but:
- my WAVE plugin no longer has the alert for the skipped heading
- the homepage looks exactly the way it did before


bytheway875 (approved):
🥳

PR #1039, opened 2021-04-23
Show client and spouse last four of SSN/ITIN on outbound call page

jenny-heath:
[story](https://www.pivotaltracker.com/story/show/177729431) had minimal instruction so this is what it looks like:

image

apologies for the quality of my spanish translations :(

### DIFF
--- a/app/controllers/hub/outbound_calls_controller.rb
+++ b/app/controllers/hub/outbound_calls_controller.rb
@@ -17,6 +17,14 @@ module Hub
 
     def show
       @outbound_call = OutboundCall.find(params[:id])
+      AccessLog.create(
+        user: current_user,
+        record: @client,
+        event_type: "viewed_call_page_ssn_itin",
+        created_at: DateTime.now,
+        ip_address: request.remote_ip,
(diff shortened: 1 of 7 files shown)


coltborg (approved):
LGTM! :shipit:

alaroia-cfa:
@jenny-heath If we're showing the client SSN/ITIN last 4, I'd like to see us record an AccessLog record; see the `AccessLog` model for details. Not strictly a blocker I suppose, but I would really like to see it.

coltborg (approved):
Reviewed the access log code. Looks good to me, thanks for adding the test as well!

app/controllers/hub/outbound_calls_controller.rb
```diff
@@ -17,6 +17,14 @@ def create
 
     def show
       @outbound_call = OutboundCall.find(params[:id])
+      AccessLog.create(
```
alaroia-cfa, line 20: Consider `.create!`. Not a blocker.

alaroia-cfa (approved):
Agree with @coltborg !

PR #1041, opened 2021-04-26
Use create! instead of create when creating access logs

jenny-heath:


alaroia-cfa (approved):

alaroia-cfa:
@jenny-heath I took the liberty of merging this!

PR #1043, opened 2021-04-27
Preserve visibility of bulk actions "Take Action" button even when user clicks "back"

alaroia-cfa:
See the follow **before** and **after** videos

https://user-images.githubusercontent.com/67708639/116308625-9430fb80-a75c-11eb-9c5f-40c52d474516.mov

https://user-images.githubusercontent.com/67708639/116308636-9bf0a000-a75c-11eb-91fe-d91d3074c6cb.mov

Implementation notes:

There seem to be two key changes. The first is hooking the `window.load` event rather than `DOMContentLoaded`. I believe this is because the HTML contains `display: none` and that's being applied _after_ JS executes. The second is calling `handleCheckboxChange` in `initBulkAction` at the start of the function. This allows any checkbox state on the page to be respected.

I also did other trivial refactoring.

### DIFF
--- a/app/javascript/lib/bulk_action.js
+++ b/app/javascript/lib/bulk_action.js
@@ -10,6 +10,15 @@ function changeElDisplay(el, property) {
     el.style.display = property;
 }
 
+function handleCheckboxChange(formEl, takeActionCountEl, takeActionFooterEl) {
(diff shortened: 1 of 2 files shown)


caitbot (approved):
I dig the refactor, that made things a lot nicer, this looks goooood

PR #1044, opened 2021-04-28
Freeze name column on client/index table #177845027

coltborg:
[Freeze name column client tables #177845027](https://www.pivotaltracker.com/story/show/177845027)

# What this PR does

- Make the "name" column on client index table be sticky when you horizontally scroll
- Clean up whitespace in `app/assets/stylesheets/components/_index-table.scss`
- Need to manually add borders around the name header column with `:before` and `:after` pseudo elements because `border-collapse: collapse` on the `table` doesn't interact will with `sticky` cells.

### DIFF
--- a/app/assets/stylesheets/components/_index-table.scss
+++ b/app/assets/stylesheets/components/_index-table.scss
@@ -39,6 +39,31 @@ It expects to be the primary content of the given page (full width).
     font-weight: $font-weight-normal;
     white-space: nowrap;
 
+    &--sticky {
+      position: sticky;
+      left: 0;
+      background-color: inherit;
+      z-index: 1;
+    }
+
+    &--sticky:after,
+    &--sticky:before {
+      content: '';
+      position: absolute;
+      left: 0;
+      width: 100%;
+    }
+
+    &--sticky:before {
+      top: -1px;
+      border-top: 1px solid black;
+    }
+
+    &--sticky:after {
+      bottom: -1px;
+      border-bottom: 1px solid black;
+    }
+
     .sortable-column-link {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_index-table.scss
```diff
@@ -51,36 +53,84 @@ It expects to be the primary content of the given page (full width).
...
+    position: sticky;
+    left: 0;
+    z-index: 1;
+    background-color: inherit;
     white-space: nowrap;
-    padding: 0.8rem $s25/2;
+    padding: 0.8rem $s25 / 2;
+  }
+
+  &__name-column {
```
bytheway875, line 71: I'm wondering whether it makes sense to call this &__name-column, or if we could call it something a bit more generic like &__sticky-column so that using this class makes sense regardless of what type of index table you're using it on. I guess the only other example where we're doing something similar is the metrics dashboard table and that's also a name, albeit an organization name..

bytheway875 (approved):
This is awesome! No blocking comments. I'm like wavering on the comment on made because it gets in to whether you should provide class names based on what they are or what they do... doesn't matter either way! As I mentioned in that comment, I'm wondering if we can eventually refactor out a lot of the customizations in _hub-metrics.scss now that index table supports a sticky first column :D

coltborg:
@bytheway875, I like your idea and changed the classes to be "modifiers". I think that approach makes a lot more sense and name what the code is trying to accomplish, now what data in the column that it is acting upon.

bytheway875 (approved):
sweet! thanks :D

PR #1046, opened 2021-04-28
Update flash alert color based on design updates

bytheway875:
The design mocks have an updated color for all alert messages. Instead of throwing this into a PR I'm working on, it seemed better to isolate it as its own PR.

Screen Shot 2021-04-28 at 3 11 28 PM


alaroia-cfa (approved):
🔴 🟢 🔵 

💵  

✅

PR #1047, opened 2021-04-29
Fix page overflow bug in Chrome #177953214

coltborg:
[Clients table causes bad horizontal scroll on Chrome based browsers #177953214](https://www.pivotaltracker.com/story/show/177953214)

# What does this PR do?

- Adds `position: relative` to the parent of the `.sr-only` classes for the label of the individual bulk action checkbox.

This fix is for Chome/Safari based browsers. Firefox was unaffected and is still unaffected.

### DIFF
--- a/app/assets/stylesheets/templates/_bulk-action-form.scss
+++ b/app/assets/stylesheets/templates/_bulk-action-form.scss
@@ -12,6 +12,7 @@
 
 .bulk-action-checkbox {
   margin-left: 1rem;
+  position: relative;
 }
 
 html[lang="en"] {


app/assets/stylesheets/templates/_bulk-action-form.scss
```diff
@@ -12,6 +12,7 @@
 
 .bulk-action-checkbox {
   margin-left: 1rem;
+  position: relative;
```
caitbot, line 15: this looks good to me (admittedly a css newbie), but for my learning -- why does this work?

caitbot (approved):

app/assets/stylesheets/templates/_bulk-action-form.scss
coltborg: The [`.sr-only` class used to hide the checkbox label](https://github.com/codeforamerica/vita-min/blob/e3b5fab2889a0048af132d360bd0f80f2ff9981b/app/assets/stylesheets/mixins/sr-only.scss#L2) to only be read by screen readers is moving the text way off screen. For some reason, that is causing the size of whole table column to then be included for the page size. `Relative` here is like an anchor for the `absolute` to be positioned against, (which is in the scrollable container) and that solves the problem. I don't have 100% certainty as to why.

PR #1052, opened 2021-04-30
Add noopener to any _blank link that doesn't have it #174205239

coltborg:
[Make sure target="_blank" links use noopener, nofollow #174205239](https://www.pivotaltracker.com/story/show/174205239)

# What does this PR do?

- Adds `rel="noopener"` to any `_blank` link in the app.


bytheway875 (approved):
🥳

PR #1054, opened 2021-05-04
Add marked_as_flagged to client

caitbot:
Adds marked_as_flagged boolean to client as well as a backfill based on whether the client is currently marked in order to preserve the red dots!

### DIFF
--- a/app/jobs/backfill_marked_as_flagged.rb
+++ b/app/jobs/backfill_marked_as_flagged.rb
@@ -1,0 +1,23 @@
+class BackfillMarkedAsFlagged < ApplicationJob
+  def perform(start: 0, finish: nil)
+    successes = []
+    errors = []
+    Client.find_each(start: start, finish: finish) do |client|
+      if client.marked_as_flagged.present?
+        puts "SKIPPING #{client.id}: marked_as_flagged is already set."
+        next
+      end
+
+      response_needed = client.response_needed_since
+
+      marked_as_flagged_value = response_needed.present?
+      if client.update(marked_as_flagged: marked_as_flagged_value)
+        successes << client.id
+      else
+        errors << client.id
+      end
+    end
+    puts "Successfully updated the following clients: #{successes}"
+    puts "Errors updating the following clients: #{errors}"
+  end
+end
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -18,6 +18,7 @@
 #  locked_at                                :datetime
 #  login_requested_at                       :datetime
 #  login_token                              :string
+#  marked_as_flagged                        :boolean
 #  response_needed_since                    :datetime
 #  routing_method                           :integer
 #  sign_in_count                            :integer          default(0), not null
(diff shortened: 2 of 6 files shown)


app/jobs/backfill_marked_as_flagged.rb
```diff
@@ -0,0 +1,23 @@
...
+    Client.find_each(start: start, finish: finish) do |client|
+      if client.marked_as_flagged.present?
+        puts "SKIPPING #{client.id}: marked_as_flagged is already set."
+        next
+      end
+
+      response_needed = client.response_needed_since
+
+      marked_as_flagged_value = response_needed.present?
+      if client.update(marked_as_flagged: marked_as_flagged_value)
```
caitbot, line 14: change to `update_all` and do it in batches to avoid updating `updated_at` on Clients since it is used to manage workflow

app/jobs/backfill_marked_as_flagged.rb
```diff
@@ -0,0 +1,23 @@
...
+    errors = []
+    Client.find_each(start: start, finish: finish) do |client|
+      if client.marked_as_flagged.present?
+        puts "SKIPPING #{client.id}: marked_as_flagged is already set."
+        next
+      end
+
+      response_needed = client.response_needed_since
+
+      marked_as_flagged_value = response_needed.present?
```
alaroia-cfa, line 13: Consider `marked_as_flagged_at` - it's almost as cheap to store a boolean as a timestamp, and sometimes it's helpful to have the timestamp data.

app/jobs/backfill_marked_as_flagged.rb
alaroia-cfa: You could also use `touch: false` per [the Rails docs.](https://api.rubyonrails.org/v6.0.1/classes/ActiveRecord/Persistence.html#method-i-save-21)

> By default, save! also sets the updated_at/updated_on attributes to the current time. However, if you supply touch: false, these timestamps will not be updated.

Feel free to use `.update_all` or to write raw SQL! Just wanted to mention that additional option.

app/jobs/backfill_marked_as_flagged.rb
```diff
@@ -0,0 +1,23 @@
+class BackfillMarkedAsFlagged < ApplicationJob
+  def perform(start: 0, finish: nil)
```
alaroia-cfa, line 2: Is the intent that you'll run this job manually in `rails c`? If so, this makes sense. Wasn't entirely sure what the usage pattern was going to be, so figured I'd ask.

alaroia-cfa: If that's how you'll do it, then I guess you'll let it run and watch the logs for the output at the end? That makes sense to me.

alaroia-cfa (changes requested):
Gotta stop updating `updated_at` I suppose; other than that, looks good!

app/jobs/backfill_marked_as_flagged.rb
caitbot: yepp, that was the intended usage pattern!

app/jobs/backfill_marked_as_flagged.rb
caitbot: huh, good point, I don't feel strongly, so we can update it to that. It is more similar to the prior implementation.

PR #1058, opened 2021-05-05
Redirect outstanding document upload links to portal login

bytheway875:


alaroia-cfa (approved):
This is good! Approved.

Is now an OK time to delete the requested documents later code? If not, I propose you file a chore for it, so we don't forget. IMHO it could be "small fry".

alaroia-cfa (commented):
Oh, one more thing - can you remove the routes to other methods on this controller?

bytheway875:
@alaroia-cfa I was going to delete code in a later PR, and can't remove these routes until the tests for those routes are removed!

PR #1063, opened 2021-05-05
Requested docs code removal 177447861

bytheway875:
- Drops columns on intake
- Removes routes
- Removes controllers
- Removes methods on intake
- Removes references to files/methods from other files and tests


db/migrate/20210505195336_remove_requested_documents_later_fields_from_intake.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveRequestedDocumentsLaterFieldsFromIntake < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :intakes, :requested_docs_token, :string
```
alaroia-cfa, line 3: If possible, it's best to do this in two steps.

1. Fully deploy the new code that doesn't read/write this column
2. Deploy a migration that deletes the column

Doesn't have to be a blocker, but FYI.

config/routes.rb
```diff
@@ -61,9 +61,6 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
     end
 
     namespace :documents do
-      delete "/requested-documents-later/remove", to: "requested_documents_later#destroy", as: :remove_requested_document
```
alaroia-cfa, line 64: Code deleted is code tested!

db/migrate/20210505195336_remove_requested_documents_later_fields_from_intake.rb
alaroia-cfa: When deploying, you're going to run into an issue that the `analytics.intakes` table has this column.

The easiest fix IMHO is to adjust `.aptible.yml` so that instead of doing:

```
before_release:
    - bundle exec rake db:migrate
```

it does

```
before_release:
    - bundle exec rake analytics:drop_views db:migrate analytics:create_views
```

I know that [this ticket](https://www.pivotaltracker.com/story/show/177352119) discusses moving the views to Scenic; I don't think that's strictly required, but _something_ is going to have to be done. :)

alaroia-cfa: One related idea is you can mark the column as [ignored](https://bigbinary.com/blog/rails-5-adds-active-record-ignored-columns) so that you can confidently delete all the code that references the column.

spec/models/system_note/document_help_spec.rb
```diff
@@ -1,3 +1,26 @@
+# == Schema Information
```
alaroia-cfa, line 1: I'm surprised this comment got autogenerated with this commit here when it wasn't autogenerated with the previous commit! But that's OK.

alaroia-cfa (approved):
I approve! I also recommend moving the removal of the column into a separately-deployable commit.

I also wonder if there are `*.erb` files that can be deleted now, too.

Code deleted is code tested.

PR #1064, opened 2021-05-05
Replace esignature link for the hub parameter replacement service with generic portal login links

embarnard:


bytheway875 (commented):
I believe the only reason to maintain the LoginLinkInsertionService separately from the ReplacementParamsService is because the LoginLink included sensitive token data that we didn't want to persist into the message body. I think this is fine to merge as is, but I also think it would create less maintained code if moved the E-signatureLink back into the ReplacementParamatersService.

PR #1065, opened 2021-05-05
Show document model validation errors to clients during intake [#177322945]

alaroia-cfa:
## Screenshot

image

## Implementation approach

On this page, we now surface any model-level validation errors to clients during intake.

### DIFF
--- a/app/forms/document_type_upload_form.rb
+++ b/app/forms/document_type_upload_form.rb
@@ -1,6 +1,7 @@
 class DocumentTypeUploadForm < QuestionsForm
   set_attributes_for :intake, :document
   validates :document, file_type_allowed: true
+  validate :load_and_validate_document
 
   def initialize(document_type, *args, **kwargs)
     @document_type = document_type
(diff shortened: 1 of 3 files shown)


coltborg (approved):
I tried uploading empty files to doc uploads in the questions section and they appropriatly displayed a validation error on the client side! 🎉 

---

### Non-blocking

Non-blocking question, should we surface the same empty validation error on the `portal/upload-documents` ( `upload_documents_controller.rb`)? I can upload empty files to this page and it says successful, while the invalid doc does not appear attached to the client in the Hub, assumingly because it isn't valid but it looks success for a client perspective.

**Gif showing me uploading to `/portal/upload-documents`**

additonal-docs-submit-empty-success

**Image of the hub not showing the invalid document attached to the client**

Screen Shot 2021-05-06 at 11 01 40 AM

app/forms/document_type_upload_form.rb
```diff
@@ -1,29 +1,45 @@
...
 
   def save
+    return false unless valid?
+
+    @doc.save!
+  end
+
+  private
+
+  def load_and_validate_document
```
alaroia-cfa, line 19: rename to `instantiate_document`

alaroia-cfa:
consider also AdditionalDocumentsForm

PR #1066, opened 2021-05-06
Add user notification partial for SystemNote::DocumentHelp

bytheway875:


app/lib/synthetic_note.rb
```diff
@@ -20,7 +20,7 @@ def self.from_client_documents(client)
       SyntheticNote.new(
         created_at: values[-1].created_at,
         body: I18n.t("hub.notes.index.document_note", count: values.length),
-        contact_record_type: "system_note"
+        contact_record_type: "_system_note"
```
embarnard: Why is it important to add the underscore before the string?

embarnard (approved):
Looks good to me! :D

PR #1069, opened 2021-05-07
Fix inconsistent locale_counts calculation for sending bulk messages

caitbot:
I paired with Asheesh about a week ago and we updated how the template/controller get `locale_counts` when sending bulk client messages. Unfortunately, the way `send_bulk_message` calculated locale counts is different (namely, it wasn't normalizing nil to "en") and so we were needlessly throwing an `ArgumentError`.

### DIFF
--- a/app/services/client_messaging_service.rb
+++ b/app/services/client_messaging_service.rb
@@ -79,17 +79,20 @@ class ClientMessagingService
     end
 
     def send_bulk_message(tax_return_selection, sender, **message_bodies_by_locale)
-      locales = Intake.where(client: tax_return_selection.clients).pluck(:locale).uniq
+      locale_counts = tax_return_selection.clients.locale_counts
+      client_locales = locale_counts.keys.filter { |key| locale_counts[key].nonzero? }.sort
 
-      sorted_locales_without_nil = locales.compact.blank? ? ["en"] : locales.compact.sort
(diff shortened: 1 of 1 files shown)


app/services/client_messaging_service.rb
```diff
@@ -79,17 +79,20 @@ def contact_methods(client)
     end
 
     def send_bulk_message(tax_return_selection, sender, **message_bodies_by_locale)
-      locales = Intake.where(client: tax_return_selection.clients).pluck(:locale).uniq
+      locale_counts = tax_return_selection.clients.locale_counts
+      client_locales = locale_counts.keys.filter { |key| locale_counts[key].nonzero? }.sort
 
-      sorted_locales_without_nil = locales.compact.blank? ? ["en"] : locales.compact.sort
```
caitbot, line 84: this line combined with the line above was causing the issue, on clients we normalize `nil` to `en`. but here, we take `nil` and throw it away UNLESS it is the only thing in the array, in which case we normalize it to `nil`. we always want to normalize `nil` to `en` when sending messages (a client with a nil intake locale will get an en message by default).

what this line resulted in was a `[nil, "es"]` locale array becoming `["es"]` and comparing to a `message_bodies_by_locale` with both `en` and `es` keys caused an ArgumentError. now `[nil, "es"]` will become `["en", "es"]`

bytheway875: nice!! thanks for distilling that.

bytheway875 (approved):

bytheway875:
@caitbot im going to merge this so that we can get the rejected part of our pr through acceptance 👯

PR #1072, opened 2021-05-07
Change status label colors in the Hub #177917593

coltborg:
[color code HUB statuses #177917593](https://www.pivotaltracker.com/story/show/177917593)

# What does this PR do?

- sets the status text in a `data-status` attribute so that it can be used in CSS
- adds CSS to style different labels based on logic described in the ticket

### DIFF
--- a/app/assets/stylesheets/components/_tax-return-list.scss
+++ b/app/assets/stylesheets/components/_tax-return-list.scss
@@ -190,7 +190,7 @@
     border-color: $color-stimulus-blue;
   }
 
-  &[data-status^="files"] {
+  &[data-status^="file"] {
     background-color: $color-grey-medium-light;
     border-color: $color-grey-medium-light;
   }
--- a/app/assets/stylesheets/components/_tax-return-list.scss
+++ b/app/assets/stylesheets/components/_tax-return-list.scss
@@ -171,5 +171,27 @@
   line-height: $font-size-18;
   padding: $s5 $s10;
   border: 1px solid $color-green-money;
-}
 
+  // More info about styling with data attributes can be found at
+  // https://css-tricks.com/a-complete-guide-to-data-attributes/
+  &[data-status="Intake/Not ready"],
+  &[data-status="Intake/Greeter - info requested"] {
+    background-color: $color-orange-medium;
+    border-color: $color-orange-medium;
+  }
+
+  &[data-status^="Tax prep"] {
+    background-color: $color-lavender;
+    border-color: $color-lavender;
+  }
+
+  &[data-status^="Quality review"] {
+    background-color: $color-stimulus-blue;
+    border-color: $color-stimulus-blue;
+  }
+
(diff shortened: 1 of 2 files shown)


bytheway875:
since these are translation strings, i wonder if you're going to run into a problem where the status labels won't change color when you're in spanish translation mode. For that reason, id recommend using the raw status (instead of the translated label)  as the data attribute and styling based on those.

bytheway875 (changes requested):
I think this will break in espanol 😢

coltborg:
Yep, address that now!

app/assets/stylesheets/components/_tax-return-list.scss
```diff
@@ -171,5 +171,32 @@
   line-height: $font-size-18;
   padding: $s5 $s10;
   border: 1px solid $color-green-money;
-}
 
+  // More info about styling with data attributes can be found at
+  // https://css-tricks.com/a-complete-guide-to-data-attributes/
+  &[data-status="Intake/Not ready"],
+  &[data-status="Admisión/No listo"],
```
coltborg, line 178: @bytheway875 added support for Spanish labels. I wonder now if this isn't the best approach, but I figured if the CSS follow the label content, it would be less logic to maintain. But now there's this CSS logic so 🤷🏻

bytheway875:
I would use the raw status instead of the translations -- it feels really odd to me to have translations in the css.

app/views/shared/_tax_return_list.html.erb
```diff
@@ -45,7 +45,8 @@
           <% end %>
         <% else %>
           <div>
-            <div class="tax-return-list__status label label--status"><%= stage_translation_from_status(tax_return.status) %>/<%= status_translation(tax_return.status) %></div>
+            <% status_text = "#{stage_translation_from_status(tax_return.status)}/#{status_translation(tax_return.status)}" %>
+            <div class="tax-return-list__status label label--status" data-status="<%= status_text %>"><%= status_text %></div>
```
bytheway875, line 49: set data-status to tax_return.status, and then you wont have to worry about spacing or translations in the css, which I think will make things less brittle. For example, say we decide to change how we display the label: put it on two lines, remove the spaces between the back slashes. We'd have to update the css or the label colors would break. If we use the raw status, we only have to update if we change the enumerable for the status (which is far less likely). the raw status includes both the stage and status

app/assets/stylesheets/components/_tax-return-list.scss
```diff
@@ -171,5 +171,27 @@
...
+    background-color: $color-lavender;
+    border-color: $color-lavender;
+  }
+
+  &[data-status^="review"] {
+    background-color: $color-stimulus-blue;
+    border-color: $color-stimulus-blue;
+  }
+
+  &[data-status^="files"] {
```
bytheway875, line 193: i think this should be file singular!

PR #1074, opened 2021-05-07
Add mixpanel click tracking to doc help page links

jenny-heath:


bytheway875:
Does it make sense to normalize how we're sending this data to mixpanel with how we refer to each of these links within the app? either by updating what we call them in the app or how we send to mixpanel? Seems like using the same "jargon" could help us communicate between product and dev better.

jenny-heath:
@bytheway875 Good question!  I took these names from the story and didn't really think about it too much. When searching for "data-track-click" in the codebase I noticed that there are a bunch of them and there doesn't seem to be a consistent naming convention...do you have an idea of what to do with these particular ones? I don't know how to make the naming more consistent between the dev and product

bytheway875:
@jenny-heath we called "finding" "can't locate" in the app and "cant get" "cant obtain". The string for each is located at the beginning of the link. We could reuse the same string in the mixpanel event name, or even normalize in variables on DocumentTypes module and reuse. Or we could merge as is and that would be fine too!

jenny-heath:
@bytheway875 ohhh i totally missed the `help_type`! weird that it's mismatched with the copy? (for example, the link with copy "I can't get this document" has `help_type` `cant_obtain` - like what's the difference between get and obtain?)

anyway, i think it makes to keep it consistent with the code - i talked to kelly and she's okay with it. so i'll update the `data-track-click` names to match the `help_type` but am trying to get this done in the next 5 min so i won't do the refactors but at least this change will make them possible.

PR #1081, opened 2021-05-10
Admins can delete clients

embarnard:


coltborg:
@embarnard / @alaroia-cfa I paired with Suran to adjust the top spacing of the delete button, we should be good to go now.

coltborg (approved):
Walked through the code with @alaroia-cfa!

PR #1082, opened 2021-05-10
Undo change to old last_outgoing_interaction_at migration so that it is backwards compatible

caitbot:
It looks like when we refactored `last_outgoing_interaction_at` to `last_outgoing_communication_at`, we changed the column name in the initial migration. The problem is that if you mess up your db (like I did) and have to drop and rerun all migrations, rails gets very confused because it cannot find `last_outgoing_interaction_at` and change it to `last_outgoing_communication_at` in the later migration. This fixes that!


caitbot (commented):
Reviewed with Shannon in pair

PR #1084, opened 2021-05-10
Greetable sites

bytheway875:
We added the greetable filter but didn't take into account sites (which inherit their greetability from their parent organization). This fixes that, and makes a scope for grabbing greetable organizations as a set.


bytheway875 (commented):
tuple reviewed with Em! thanks @embarnard

PR #1086, opened 2021-05-10
Metabase rake tasks to run metabase tasks on CI

bytheway875:


bytheway875 (commented):
Paired on this with @alaroia-cfa last week

PR #1087, opened 2021-05-10
Diy landing page 178079705

bytheway875:
Screen Shot 2021-05-10 at 5 49 21 PM


app/assets/stylesheets/templates/_home.scss
```diff
@@ -99,7 +99,7 @@
         }
       }
 
-      &.stimulus {
+      &.diy {
```
bytheway875, line 102: Stimulus page / class is no longer in use

spec/features/diy/new_diy_client_spec.rb
```diff
@@ -9,6 +9,10 @@
 
   scenario "a new client files through TaxSlayer" do
     allow(MixpanelService).to receive(:send_event)
+    visit "/diy"
+    expect(page).to have_selector("h1", text: "File taxes on your own for free!")
+    expect(page).to have_selector("p", text: "We provide free access to an online tax prep site called TaxSlayer. We'll send you a link to create your own TaxSlayer login, so you can file on your own.")
+    click_on "Continue"
```
alaroia-cfa, line 15: If you are doing `click_on` here, then I recommend not doing `visit` below. Personally I would not cover a landing page with a feature spec, but it's OK with me if we do!

app/controllers/public_pages_controller.rb
```diff
@@ -38,4 +38,6 @@ def tax_questions; end
   def stimulus_recommendation; end
 
   def sms_terms; end
+
+  def diy; end
```
alaroia-cfa, line 42: FYI - this `def diy; end` isn't strictly necessary (at least, from what I remember from looking into this a few months ago). No change required, just wanted to share that, and also if you have learned it _is_ required, then I want to know that.

app/assets/stylesheets/templates/_home.scss
alaroia-cfa: Stellar

app/controllers/public_pages_controller.rb
bytheway875: oh, i didn't know that! im going to leave it because seems like the pattern in this controller is to create an empty method to signify a custom controller action that renders a corresponding template, but i guess we could decide to remove all of those if we want to.

PR #1089, opened 2021-05-11
Routing: stop falling back to national site, instead redirect to at capacity page

jenny-heath:
_Note: this should not be merged until [this story](https://www.pivotaltracker.com/story/show/178103177) is merged. that, or the code for this needs to change (talk to me about how)._

Would love some feedback on this! Particularly:
1) replacing `PersonalInfoController#after_update_success` the way i did, and whether there's a better way to do the redirect
2) whether to add a new `routing_method` for those who are now routed to no one

### DIFF
--- a/app/controllers/questions/personal_info_controller.rb
+++ b/app/controllers/questions/personal_info_controller.rb
@@ -19,6 +19,7 @@ module Questions
           current_intake.update(vita_partner: vita_partner)
           super
         else
+          current_intake.client.update(routing_method: routing_service.routing_method)
           return redirect_to at_capacity_questions_path
         end
       end
--- a/app/controllers/questions/personal_info_controller.rb
+++ b/app/controllers/questions/personal_info_controller.rb
@@ -6,16 +6,21 @@ module Questions
       {}
     end
 
-    def after_update_success
+    def update
       unless Client.after_consent.where(intake: current_intake).exists?
         routing_service = PartnerRoutingService.new(
           source_param: current_intake.source,
-          zip_code: current_intake.zip_code,
+          zip_code: form_params[:zip_code],
         )
         vita_partner = routing_service.determine_partner
 
-        current_intake.client.update(vita_partner: vita_partner, routing_method: routing_service.routing_method)
-        current_intake.update(vita_partner: vita_partner)
+        if vita_partner.present?
+          current_intake.client.update(vita_partner: vita_partner, routing_method: routing_service.routing_method)
+          current_intake.update(vita_partner: vita_partner)
+          super
+        else
+          return redirect_to at_capacity_questions_path
+        end
       end
     end
   end
--- a/app/services/partner_routing_service.rb
+++ b/app/services/partner_routing_service.rb
@@ -18,7 +18,7 @@ class PartnerRoutingService
     from_state_routing = vita_partner_from_state if @zip_code.present?
     return from_state_routing if from_state_routing.present?
 
-    route_to_national_overflow_partner
+    nil
   end
 
   private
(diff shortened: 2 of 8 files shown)


app/controllers/questions/personal_info_controller.rb
```diff
@@ -6,16 +6,21 @@ def tracking_data
...
 
-        current_intake.client.update(vita_partner: vita_partner, routing_method: routing_service.routing_method)
-        current_intake.update(vita_partner: vita_partner)
+        if vita_partner.present?
+          current_intake.client.update(vita_partner: vita_partner, routing_method: routing_service.routing_method)
+          current_intake.update(vita_partner: vita_partner)
+          super
+        else
+          return redirect_to at_capacity_questions_path
+        end
```
jenny-heath, line 23: janky? not janky?

app/services/partner_routing_service.rb
```diff
@@ -18,7 +18,7 @@ def determine_partner
     from_state_routing = vita_partner_from_state if @zip_code.present?
     return from_state_routing if from_state_routing.present?
 
-    route_to_national_overflow_partner
+    nil
```
jenny-heath, line 21: `route_to_national_overflow_partner` sets a `routing_method` - i wonder if we should add one like `not_routed` or `none` to leave a paper trail so we know which clients were not assigned a partner due to capacity problems.

i should probably mention that this change will be coupled with removing the "continue anyway" link on the at capacity page, so no one will be getting past this and therefore should not get lost in the hub (i think? at this point their status should be something that makes them invisible in the hub?)

however, even assuming no one gets past the at capacity page, is it a bad idea to assign these clients to no one? my thinking right now is that it's okay because they won't have entered contact info and therefore won't be able to come back as a "returning client"

app/services/partner_routing_service.rb
```diff
@@ -69,7 +69,7 @@ def vita_partner_from_state
     end
   end
 
-  # TODO: What happens in the case that there are no national overflow partners? Is there a GYR organization we can route them to?
+  # TODO: keep this method until we bring it back or discard and search git history?
```
jenny-heath, line 72: this

app/services/partner_routing_service.rb
bytheway875: you could add a routing method called "over_capacity" or something like that

bytheway875: the default return of a method is nil, so you don't even really have to explicitly return nil like this. I'd probably go with commenting out national routing line and adding a new line that sets the routing method to "none" or "over_capacity"

app/controllers/questions/personal_info_controller.rb
bytheway875: what are the implications of changing this from after_update_success to update? I'm guessing we did this so that we can do the redirect and that's why we call super, and i actually think that's great and a better pattern than after_update_success, but does not having after_update_success defined mean that it uses the questions_controller implementation of that method and if so does it do anything?

app/services/partner_routing_service.rb
jenny-heath: ah good point about it already returning nil. funny thing is that if i add the line for `@routing_method = :at_capacity` it no longer returns nil 😆

app/controllers/questions/personal_info_controller.rb
bytheway875: It actually feels more in line with Products expectations for us to just toss that data, actually, since they haven't consented to anything, so im kind of cool with you doing that. It might be worth noting that it was an intentional decision as a comment, though.

bytheway875 (approved):
👍

PR #1090, opened 2021-05-11
move documents association before intake association to fix cascading deletion error

bytheway875:


app/models/client.rb
```diff
@@ -41,9 +41,9 @@ class Client < ApplicationRecord
   self.per_page = 25
 
   belongs_to :vita_partner, optional: true
```
bytheway875, line 43: intake cascades to deleting the documents requests which will still have documents pointing to the client unless we delete those first.

This problem could also be prevented by moving documents request association to the client instead of having it on the document_request, which we might consider doing in a future PR

embarnard (approved):
Looks good!

PR #1091, opened 2021-05-11
Refactor client messaging + replace params before sending messages

bytheway875:
This will now pass every email through replacement parameters service before sending, which means that we can use replacement parameters / status macros by default for bulk messaging, and users can use replacement parameters for custom messages on the messages page or take action page.

It also refactors client messaging to consolidate some logic while maintaining the same public API, method wise I also switched to using keyword arguments for all messaging methods since they're unordered.

### DIFF
--- a/app/controllers/questions/consent_controller.rb
+++ b/app/controllers/questions/consent_controller.rb
@@ -13,15 +13,12 @@ module Questions
     def after_update_success
       sign_in current_intake.client
       current_intake.advance_tax_return_statuses_to("intake_in_progress")
-      if current_intake.email_notification_opt_in_yes?
-        body = I18n.t("messages.getting_started.email_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id)
-        subject = I18n.t("messages.getting_started.email_subject", locale: current_intake.locale)
-        ClientMessagingService.send_system_email(current_intake.client, body, subject)
-      end
-      if current_intake.sms_notification_opt_in_yes?
-        body = I18n.t("messages.getting_started.sms_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id)
-        ClientMessagingService.send_system_text_message(current_intake.client, body)
-      end
+      ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
+        client: current_intake.client,
+        email_body: I18n.t("messages.getting_started.email_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id),
+        subject: I18n.t("messages.getting_started.email_subject", locale: current_intake.locale),
+        sms_body: I18n.t("messages.getting_started.sms_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id)
+      )
       Intake14446PdfJob.perform_later(current_intake, "Consent Form 14446.pdf")
       IntakePdfJob.perform_later(current_intake.id, "Preliminary 13614-C.pdf")
     end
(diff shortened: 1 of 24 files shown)


app/controllers/questions/consent_controller.rb
```diff
@@ -13,15 +13,12 @@ def form_params
...
-      if current_intake.email_notification_opt_in_yes?
-        body = I18n.t("messages.getting_started.email_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id)
-        subject = I18n.t("messages.getting_started.email_subject", locale: current_intake.locale)
-        ClientMessagingService.send_system_email(current_intake.client, body, subject)
-      end
-      if current_intake.sms_notification_opt_in_yes?
-        body = I18n.t("messages.getting_started.sms_body", preferred_name: current_intake.preferred_name, portal_login_url: new_portal_client_login_url(locale: current_intake.locale), locale: current_intake.locale, client_id: current_intake.client_id)
-        ClientMessagingService.send_system_text_message(current_intake.client, body)
-      end
+      ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
```
alaroia-cfa, line 16: _awesome_

app/services/client_messaging_service.rb
```diff
@@ -1,71 +1,86 @@
...
-    def send_system_text_message(client, body)
-      OutgoingTextMessage.create!(
-        client: client,
-        body: body,
-        to_phone_number: client.sms_phone_number,
-        sent_at: DateTime.now
-      )
+    def send_system_text_message(client:, body:, tax_return: nil, locale: nil)
+      args = { client: client, body: body, user: nil }
+      args[:tax_return] = tax_return if tax_return.present?
```
alaroia-cfa, line 46: I recommend not bothering to filter for `tax_return.present?`. Not a blocker.

app/models/outgoing_email.rb
```diff
@@ -38,6 +38,7 @@ class OutgoingEmail < ApplicationRecord
   validates_presence_of :body
   validates_presence_of :subject
   validates :mailgun_status, inclusion: { in: ALL_KNOWN_MAILGUN_STATUSES }
+  attr_accessor :locale, :tax_return
```
alaroia-cfa, line 41: I don't understand the purpose of this yet.

app/services/client_messaging_service.rb
```diff
@@ -1,71 +1,86 @@
...
+
       if client.intake.email_notification_opt_in_yes? && client.email_address.present? && email_body.present?
-        message_records[:outgoing_email] = send_system_email(client, email_body, subject)
+        message_records[:outgoing_email] = send_system_email(**args)
       end
       if client.intake.sms_notification_opt_in_yes? && client.sms_phone_number.present? && sms_body.present?
-        message_records[:outgoing_text_message] = send_system_text_message(client, sms_body)
+        args.delete(:subject)
+        args[:body] = sms_body
+        message_records[:outgoing_text_message] = send_system_text_message(**args)
```
alaroia-cfa, line 83: I usually have a personal preference for seeing all arguments passed directly, rather than `**args`'d. Definitely not a blocker.

spec/controllers/questions/consent_controller_spec.rb
```diff
@@ -239,8 +239,8 @@
           BODY
 
           expect(ClientMessagingService).to have_received(:send_system_text_message).with(
-            intake.client,
-            body.chomp,
+              client: intake.client,
+              body: body.chomp
```
alaroia-cfa, line 243: It's great that we have these unit tests; they effectively validate the use of `**args`.

spec/controllers/questions/notification_preference_controller_spec.rb
```diff
@@ -71,13 +71,7 @@
...
         expect(intake.email_notification_opt_in).to eq("yes")
         expect(intake.sms_phone_number).to eq("+15005550006")
-        expect(ClientMessagingService).to_not have_received(:send_system_text_message).with(
-          intake,
-          I18n.t(
-            "messages.sms_opt_in",
-            locale: intake.locale
-          )
-        )
+        expect(ClientMessagingService).to_not have_received(:send_system_text_message)
```
alaroia-cfa, line 74: Great change.

alaroia-cfa (approved):
This is great! I left some non-blocking comments. Feel free to ship!

app/models/outgoing_email.rb
bytheway875: oh good catch. it is unneccessary, a remnant of when i was going to process replacement params on the model instead of in the service.

PR #1093, opened 2021-05-11
Refactor system emails to use replacement params

bytheway875:


app/controllers/documents/documents_help_controller.rb
```diff
@@ -7,18 +7,11 @@ class DocumentsHelpController < DocumentUploadQuestionController
     def show; end
 
     def send_reminder
+      doc_type = params[:doc_type].constantize.key
```
alaroia-cfa, line 10: I'm personally nervous when I see `.constantize` because unless we're careful, it can lead to [code execution.](http://gavinmiller.io/2016/the-safesty-way-to-constantize/)

I think this is a use that is vulnerable to "Command Injection" documented in that doc.

What do you think about using something other than `.constantize`, or am I misunderstanding the code flow here?

app/models/outgoing_email.rb
```diff
@@ -38,6 +38,7 @@ class OutgoingEmail < ApplicationRecord
   validates_presence_of :body
   validates_presence_of :subject
   validates :mailgun_status, inclusion: { in: ALL_KNOWN_MAILGUN_STATUSES }
+  attr_accessor :locale, :tax_return
```
alaroia-cfa, line 41: I'm not sure I understand why this is needed, and would love to. Not a blocker.

config/locales/en.yml
```diff
@@ -941,34 +941,34 @@ en:
     successful_submission:
       subject: Your tax information has been successfully submitted!
       email_body: |
-        Hello %{preferred_name},
+        Hello <<Client.PreferredName>>,
```
alaroia-cfa, line 944: ✨

config/locales/es.yml
```diff
@@ -932,34 +932,34 @@ es:
...
         Si tiene alguna pregunta, por favor responda a este mensaje.
 
         ¡Estamos aquí para ayudar!
 
         Su equipo de impuestos en GetYourRefund.org
       sms_body: |
-        Hola %{preferred_name},
-        ¡Su información de impuestos ha sido presentada con éxito a su preparador de impuestos! Su número de confirmación es %{client_id}.
+        Hola <<Client.PreferredName>>,
+        ¡Su información de impuestos ha sido presentada con éxito a su preparador de impuestos! Su cliente id es <<Client.ClientId>>.
```
alaroia-cfa, line 950: ✨

spec/controllers/questions/consent_controller_spec.rb
```diff
@@ -159,89 +159,45 @@
...
-
-            ¡Estamos aquí para servirle!
-
-            Su Equipo de Impuestos de GetYourRefund.org
-          BODY
-
           expect(ClientMessagingService).to have_received(:send_system_email).with(
               client: intake.client,
-              body: email_body,
+              body: I18n.t("messages.getting_started.email_body", locale: "es"),
```
alaroia-cfa, line 189: Nice!

spec/services/replacement_parameters_service_spec.rb
```diff
@@ -72,27 +79,10 @@
     end
   end
 
-  context "<<Client.YouOrMaybeYourSpouse>>" do
```
alaroia-cfa, line 75: Why remove the `YouOrMaybeYourSpouse` test block?

alaroia-cfa (changes requested):
Looks like a great cleanup!

I am concerned about the call to `.constantize` in this situation and want to be sure I understand that before approving.

app/controllers/documents/documents_help_controller.rb
bytheway875: Right now, we're using the passed doc type to get the key. Maybe instead of the doc_type, we can use the translation key and look it up directly from the translation file. Since this is an existing implementation and will require a refactor of this area of the code in more depth, I think we should file a  chore to come up with a better way to do this vs blocking this PR.

app/models/outgoing_email.rb
bytheway875: i feel like i removed this from the other PR but had already branched off of it so it got added back. will remove.

spec/services/replacement_parameters_service_spec.rb
bytheway875: because i removed the code itself in the replacement parameter service as it isnt being used anywhere currently.

PR #1097, opened 2021-05-12
Use document type parameters instead of class name

bytheway875:
funky URLS like documents/doc-help/DocumentTypes::Identity now have nicer urls (documents/doc-help/identity)

### DIFF
--- a/app/controllers/documents/documents_help_controller.rb
+++ b/app/controllers/documents/documents_help_controller.rb
@@ -29,6 +29,13 @@ module Documents
 
     def document_type_from_param
       document_type = DocumentType.from_param(params[:doc_type])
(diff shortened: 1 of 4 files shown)


alaroia-cfa (approved):
Are you concerned about handling people in the middle of this flow (i.e., existing URLs)? I think it'd be nice to handle that somehow. Perhaps we can `redirect_back` in that case? I haven't fully thought that through.

alaroia-cfa:
Since this flow is so short-lived, we could roll out a change where we accept both, but only generate the new way, then 30 min later, deploy a change where we only accept the new way.

alaroia-cfa (approved):
new commit lgtm, yay

PR #1099, opened 2021-05-13
Update gated questions to no when someone answers the gated retirement questions to no

bytheway875:
This gating question already existed, so we just needed to update the gating questions to no as well.


spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -174,6 +174,11 @@
     expect(page).to have_selector("h1", text: "In 2020, did you have Social Security income, retirement income, or retirement contributions?")
     click_on "No"
 
+    # check for gating logic
```
bytheway875, line 177: I don't know if this makes sense in a feature spec, but we do some other "check our wizard logic" in the feature specs, so i went ahead and did it. Willing to remove.

caitbot (approved):
✨

PR #1102, opened 2021-05-14
Foreman Spike: Modified procfile

bytheway875:
We already had a procfile (news to me!), so I just modified it.

Foreman suggests that you do NOT add foreman to your Gemfile and instead people should install it on their computer outside of bundler:

`gem install foreman`

Make sure that all other processes are dead before trying to start with this, or else all processes will terminate.

Spike result:

Since we already have a procfile that I was able to just modify, I think we should just merge this after typical review. People can decide whether to use or not use foreman to start the app. I will show at next computer gardening.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
-cmd: bundle exec rails s -b 0.0.0.0 -p 3000
+server: bundle exec rails s -b 0.0.0.0 -p 3000
 worker: bundle exec rails jobs:work
-cron: exec supercronic /app/crontab
+webpack: exec bin/webpack-dev-server


alaroia-cfa (changes requested):
Since we use this Procfile for Aptible per https://deploy-docs.aptible.com/docs/procfiles , I'm concerned that if we change the name of the `cmd` key in the Procfile, we'll need to adjust config elsewhere that assumes `vita-min-prod-cmd` is the main web app (e.g., Datadog - maybe nowhere else? Not sure).

Given this, I'd prefer an approach where we don't rename `cmd` and don't remove `cron`. Then presumably we can scale the `webpack` service in Aptible to 0 containers, which would avoid it running there.

Given that you were surprised that the Procfile is used by Aptible, perhaps we should add a comment that that's the case? It's kind of funny to have this one file used by Foreman (one of the least important things) and Aptible (one of the most important things); on the upside, it's also nice that we don't have to duplicate the service list. :) I think it's fine to have one file shared by them both.

Requesting changes since if we merge this with `cmd` renamed, it creates more work than I think is a good idea, but I'm open to changing my mind.

bytheway875:
@alaroia-cfa Oh, i didnt realize that the procfile was used by anything else because superchronic was throwing an error when i tried to run using the procfile locally.

I can instead close this and create a Procfile.dev and then we can keep them seperate.

PR #1105, opened 2021-05-14
Update personal info controller to validate the form again and set up next_path to at_capacity if a vita partner is not returned

bytheway875:
Turns out the form is what validates the zip code, and if we don't validate the zip code the user doesnt see an error message when they enter a bad zip AND that bad zip gets through to our partner router.  😢  So reverting to use update and after_update_success and overriding next_path when there is no vita partner ensures we're validating and also sending people to the right place.


embarnard (approved):
LGTM!

PR #1107, opened 2021-05-17
Procfile take two

bytheway875:
Created a procfile.dev and set up foreman to reference it. Also added a comment in the deployment procfile.

I called the Rails server "server" because "cmd" is really short and throws of the visuals for me on the logging 😢 
Screen Shot 2021-05-17 at 11 10 53 AM


alaroia-cfa (approved):
Yippee! Looks great.

alaroia-cfa:
@bytheway875 Non-blocking feedback that I just realized: it's probably smart to add foreman to our `Brewfile` so that `bin/setup` will auto-install it. This is compatible with the desire not to add `foreman` to the _Gemfile_ (hopefully this is clear; lmk if not clear).

PR #1108, opened 2021-05-17
Add documentation of system_notes

jenny-heath:


coltborg (approved):
Thanks for doing this research! It'll come in handy later.

PR #1110, opened 2021-05-17
Add bulk messaging and message template to bulk status/assignee updates

caitbot:

### DIFF
--- a/app/javascript/lib/dynamic_take_action_changes.js
+++ b/app/javascript/lib/dynamic_take_action_changes.js
@@ -1,19 +1,22 @@
-export function initTakeActionOnChangeHandlers() {
-    $("#hub_take_action_form_status, #hub_take_action_form_locale").change(function() {
+export function initTakeActionOnChangeHandlers(formName) {
+
+    $(`#hub_${formName}_form_status, #hub_${formName}_form_locale`).change(function() {
         $(".hub-form :input").prop("disabled", true);
-        const pathname = window.location.pathname.replace("update_take_action", "edit_take_action");
-        window.location.replace(pathname + "?" + searchParamsString() + "#status");
+        const pathname = window.location.pathname.replace(`update`, `edit`);
+        window.location.replace(pathname + "?" + searchParamsString(formName) + "#status");
     });
 }
-
-function searchParamsString() {
+function searchParamsString(formName) {
     const searchParams = new URLSearchParams();
-    const taxReturnId = $("#hub_take_action_form_tax_return_id").val();
-    const selectedStatus = $("#hub_take_action_form_status").val();
-    const locale = $("#hub_take_action_form_locale").val();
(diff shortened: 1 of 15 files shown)


app/javascript/lib/dynamic_take_action_changes.js
```diff
@@ -1,19 +1,22 @@
-export function initTakeActionOnChangeHandlers() {
-    $("#hub_take_action_form_status, #hub_take_action_form_locale").change(function() {
+export function initTakeActionOnChangeHandlers(formName) {
+
+    $(`#hub_${formName}_form_status, #hub_${formName}_form_locale`).change(function() {
         $(".hub-form :input").prop("disabled", true);
-        const pathname = window.location.pathname.replace("update_take_action", "edit_take_action");
-        window.location.replace(pathname + "?" + searchParamsString() + "#status");
+        const pathname = window.location.pathname.replace(`update`, `edit`);
```
bytheway875, line 5: can change this to normal quotes if we want to now since no string interpolation is happening.

app/services/client_messaging_service.rb
```diff
@@ -94,10 +94,10 @@ def contact_methods(client)
 
     def send_bulk_message(tax_return_selection, sender, **message_bodies_by_locale)
       locale_counts = tax_return_selection.clients.locale_counts
```
bytheway875, line 96: Is this just a refactor or is there a behavior change too?

spec/features/hub/bulk_actions_spec.rb
```diff
@@ -136,13 +136,19 @@
     expect(page).to have_text "Keep current status"
 
     select new_user.name, from: "New Assignee"
-    select "Ready to file", from: "New Status"
+    select "Greeter - info requested", from: "New Status"
```
bytheway875, line 139: nice!!

should we also make sure that "New Assignee" is still selected when the page updates?

bytheway875 (approved):
Yay! I think there is one more tiny test we can add to flesh out the feature spec and make sure the assignee stays when the page reloads.

app/services/client_messaging_service.rb
caitbot: small behavior change, due to message templating and that issue we discussed where it fills both "es" and "en" message bodies even if both are not shown on the UI, the old check wasn't working (there would be an "es" message body when no clients had "es" locale). the real purpose of this check was to ensure no client locale is missing so that clients get messages in their preferred locale, and so I updated to check for inclusion rather than equality.

spec/features/hub/bulk_actions_spec.rb
caitbot: That is implicitly checked due to the notification check

`expect(page).to have_text "You successfully assigned 2 tax returns to Admin the Second."`

caitbot: But I can add it to be explicit

PR #1113, opened 2021-05-18
Add defer to script tag #178200050

coltborg:
[Add defer to <script> tags with src #178200050](https://www.pivotaltracker.com/story/show/178200050)

# What does this PR do?

- Adds the defer attribute to our application.js. This will allow it to be non-render blocking and improve performance.

## Performance locally

The performance locally won't be the same as in production, but I did still see improvements! (Local performance is worse because assets aren't minimized)

**Before**
before-defer

**After**
after-defer

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -44,7 +44,7 @@
     </script>
     <%= stylesheet_pack_tag 'application', media: 'all' %>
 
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
@coltborg I reviewed, and I read https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script , and I approve!

PR #1114, opened 2021-05-18
Suspend/unsuspend users

caitbot:

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -5,6 +5,7 @@ en:
     unlock_confirmation: "Unlock %{name}'s account? Have they confirmed their identity?"
     suspend_account: "Suspend account"
     suspend_confirmation: "Suspend %{name}'s account? This will remove them from all currently assigned tax returns and prevent them from logging in."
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -3,6 +3,9 @@ en:
   hub:
     unlock_account: "Unlock account"
     unlock_confirmation: "Unlock %{name}'s account? Have they confirmed their identity?"
+    suspend_account: "Suspend account"
+    suspend_confirmation: "Suspend %{name}'s account?"
```
bytheway875, line 7: "Suspend %{name}'s account? This will remove them from all currently assigned tax returns and prevent them from logging in." Might help 'self-document' what this feature does.

bytheway875 (approved):
Want to see what you think about making the alert a bit more descriptive, but it's kind of whatever since no one explicitly asked for an alert.

ALSO, should we check in with product and see if we should take suspension into account for "accessible users" (i.e remove suspended accounts from various assignment lists)

PR #1116, opened 2021-05-18
Allow site coordinators and org leads to unlock clients #178150745

coltborg:
[Allow org leads & site coordinators to unlock clients #178150745](https://www.pivotaltracker.com/story/show/178150745)

# What does this PR do?

- Allow site coordinators and org leads roles to `unlock` on clients
- Add `site_coordinator?` to user
- Add tests to `clients_controller_spec`
- Add tests for `site_coordinator?` and missing tests for `greeter?` and `org_lead?`

### DIFF
--- a/app/controllers/hub/clients_controller.rb
+++ b/app/controllers/hub/clients_controller.rb
@@ -97,7 +97,7 @@ module Hub
     end
 
     def unlock
-      raise CanCan::AccessDenied unless current_user.admin?
+      raise CanCan::AccessDenied unless current_user.admin? || current_user.org_lead? || current_user.site_coordinator?
 
       @client.unlock_access! if @client.access_locked?
       flash[:notice] = I18n.t("hub.clients.unlock.account_unlocked", name: @client.preferred_name)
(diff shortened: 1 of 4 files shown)


app/controllers/hub/clients_controller.rb
```diff
@@ -97,7 +97,7 @@ def update_take_action
     end
 
     def unlock
-      raise CanCan::AccessDenied unless current_user.admin?
```
coltborg, line 100: I could be convinced to break this into two lines or a way to make it read easier.

app/controllers/hub/clients_controller.rb
```diff
@@ -97,7 +97,7 @@ def update_take_action
...
+      raise CanCan::AccessDenied unless current_user.admin? || current_user.org_lead? || current_user.site_coordinator?
```
coltborg, line 100: I could be convinced to break this into two lines or a way to make it read easier.

app/models/user.rb
```diff
@@ -197,6 +197,10 @@ def org_lead?
     role_type == OrganizationLeadRole::TYPE
   end
 
+  def site_coordinator?
```
alaroia-cfa, line 200: Thanks for writing a test for this ✨

PR #1117, opened 2021-05-19
Use port 3000 in Foreman, too

alaroia-cfa:
This allows us to get the same URL with & without foreman, which I prefer.


caitbot (approved):
ty ty, this looks good

PR #1127, opened 2021-05-25
Organization index refactor #178205649

bytheway875:
Before: Queries for organization capacity were getting called on each organization individually, + queries for count of sites for each org, because the eagerloaded organization capacity query was on the wrong query due to a misunderstanding of how the instance variables were used
Screen Shot 2021-05-26 at 2 53 39 PM

After: We've decreased the number of queries and organization capacity is properly eager loaded, I think. All queries are executed in the action vs executing on render.
Screen Shot 2021-05-26 at 2 54 17 PM

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,7 +32,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_stimulus_triage
-    StimulusTriage.find_by_id(session[:stimulus_triage_id])
+    StimulusTriage.find_by_id(session[:stimulus_triage_id]) unless hub?
   end
 
   def clear_intake_session
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -32,14 +32,10 @@ module Hub
       organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
 
       @organizations_by_coalition = if can? :read, Coalition
-         organizations.group_by(&:coalition).sort do |coalition_a, coalition_b|
-          a_name = coalition_a[0]&.name
-          b_name = coalition_b[0]&.name
-          a_name && b_name ? (a_name <=> b_name) : 1 # put independent orgs at the end
-        end
-      else
-        { nil => organizations }
-      end
+                                      organizations.group_by(&:coalition).sort_by { |el| [el[0]&.name ? 0 : 1, el[0]&.name || 0] } # sort independent org (nil coalition) to end of list
+                                    else
+                                      { nil => organizations }
+                                    end
     end
 
     def edit
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -29,15 +29,16 @@ module Hub
 
     def index
       # Load organizations slowly first, to avoid lots of queries later
-      @organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
-      @organizations_by_coalition = if current_user.coalition_lead? || current_user.admin?
-        @organizations.group_by(&:coalition).sort do |coalition_a, coalition_b|
+      organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
+
+      @organizations_by_coalition = if can? :read, Coalition
+         organizations.group_by(&:coalition).sort do |coalition_a, coalition_b|
           a_name = coalition_a[0]&.name
           b_name = coalition_b[0]&.name
-          a_name && b_name ? (a_name <=> b_name) : 1 # put nil coalition at the end
+          a_name && b_name ? (a_name <=> b_name) : 1 # put independent orgs at the end
         end
       else
-        { nil => @organizations }
+        { nil => organizations }
       end
     end
 
--- a/app/controllers/hub/organizations_controller.rb
+++ b/app/controllers/hub/organizations_controller.rb
@@ -28,9 +28,17 @@ module Hub
     end
 
     def index
-      @coalitions = Coalition.accessible_by(current_ability).includes(:organizations)
-      @organizations = @vita_partners.organizations.includes(:organization_capacity)
-      @independent_organizations = @organizations.where(coalition: nil)
+      # Load organizations slowly first, to avoid lots of queries later
+      @organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
+      @organizations_by_coalition = if current_user.coalition_lead? || current_user.admin?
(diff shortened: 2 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -32,7 +32,7 @@ def intake_from_completed_session
   end
 
   def current_stimulus_triage
-    StimulusTriage.find_by_id(session[:stimulus_triage_id])
+    StimulusTriage.find_by_id(session[:stimulus_triage_id]) unless hub?
```
bytheway875, line 35: I noticed that this was getting called needlessly in the hub, and experimented with adding an "unless hub?" ... i could also do this with current_intake, or i could remove this one too.

app/controllers/hub/organizations_controller.rb
```diff
@@ -28,9 +28,18 @@ def show
...
 
     def index
-      @coalitions = Coalition.accessible_by(current_ability).includes(:organizations)
-      @organizations = @vita_partners.organizations.includes(:organization_capacity)
-      @independent_organizations = @organizations.where(coalition: nil)
+      # Load organizations slowly first, to avoid lots of queries later
+      organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
+
+      @organizations_by_coalition = if can? :read, Coalition
+         organizations.group_by(&:coalition).sort do |coalition_a, coalition_b|
```
bytheway875, line 35: This sort is a little bit messy, but we have to deal with the fact that there is a nil in the bunch.

bytheway875: For purposes of the view and looping through the data one time, we set up every role to loop through @organizations_by_coalition, but some roles don't have access to Coalitions so we force a nil coalition.

app/views/hub/organizations/index.html.erb
```diff
@@ -11,59 +11,25 @@
...
-          <p><%= t(".organizations_by_coalition") %></p>
-
-          <% @coalitions.each do |coalition| %>
-            <div class="grid-flex space-between center-aligned spacing-above-10">
-              <h2 class="spacing-below-0"><%= coalition.name %></h2>
-              <% unless coalition.organizations.count.zero? %>
-                <div class="help-text">Capacity</div>
-              <% end %>
-            </div>
+        <% @organizations_by_coalition.each do |coalition, organizations| %>
```
bytheway875, line 14: I think I added a bit of view complexity in terms of determining whether to display "Independent Organizations" or nothing for the title, but gained some data simplicity in that we aren't dealing with data from multiple queries with different eager loaded tables that act differently.

app/controllers/application_controller.rb
alaroia-cfa: I think this is OK to add to this PR.

I would like to see further refactors, but they're not essential. (For example, do we even really still need this?)

app/controllers/hub/organizations_controller.rb
alaroia-cfa: I encourage an approach where you separate the nil values from the non-nil values, so that you can treat them differently.

I also personally enjoy the 'guard clause' style, e.g.

```
return 1 if coalition_a[0].nil? # independent orgs at the end

coalition_a[0].name <=> coalition_b[0].name
```

The approach you've taken here is fine, too.

app/controllers/hub/organizations_controller.rb
```diff
@@ -28,9 +28,18 @@ def show
     end
 
     def index
-      @coalitions = Coalition.accessible_by(current_ability).includes(:organizations)
-      @organizations = @vita_partners.organizations.includes(:organization_capacity)
-      @independent_organizations = @organizations.where(coalition: nil)
+      # Load organizations slowly first, to avoid lots of queries later
+      organizations = @vita_partners.organizations.includes(:coalition, :child_sites, :organization_capacity).load
+
+      @organizations_by_coalition = if can? :read, Coalition
```
alaroia-cfa, line 34: ⭐

app/views/hub/organizations/index.html.erb
alaroia-cfa: Makes sense!

app/views/hub/organizations/index.html.erb
```diff
@@ -11,59 +11,25 @@
...
+          <div class="grid-flex space-between center-aligned spacing-above-10">
+            <h2 class="spacing-below-0"><%= coalition&.name || (t(".independent_organizations", count: @organizations_by_coalition.length)) %></h2>
```
alaroia-cfa, line 16: I think in this case, `<%= coalition.nil?  ? t(".independent_organizations", count: @organizations_by_coalition.length) : coalition.name %>` would be a nice way to use an explicit nil check.

What you have is fine, too.

app/views/hub/organizations/index.html.erb
```diff
@@ -11,59 +11,25 @@
 
     <div class="grid">
       <div class="grid__item width-one-half">
-        <% if @coalitions.count > 0 %>
-          <p><%= t(".organizations_by_coalition") %></p>
```
alaroia-cfa, line 15: I guess you removed this heading b/c it's clear what's happening without? OK!

config/locales/en.yml
```diff
@@ -425,9 +425,10 @@ en:
           one: "%{count} site"
           other: "%{count} sites"
       index:
-        independent_organizations: "Independent Organizations"
+        independent_organizations:
+          one: " "
```
alaroia-cfa, line 429: If `""` would work just as well as `" "`, then I encourage you to remove the space.

alaroia-cfa (approved):
This is a good simplifying change!

Since the purpose of the refactor was (as I understood it) a performance improvement, I'd like to see some quantitative statement. Here's one totally made-up example: "Before this PR, loading the org page made one query per org (8 on my dev environment) and it always does 5 queries. Here's a screenshot of that in rack mini profiler. This is important because in production, it does >100 queries; each query takes 40 milliseconds, which means the page takes 4 seconds to load. I've validated that the queries it now executes take approx 200ms in a prod `rails c`."

alaroia-cfa:
Net negative 3 lines - nice!

app/controllers/application_controller.rb
bytheway875: @alaroia-cfa I don't think so, but if we're going to remove it we also need to remove all of the accompanying stimulus triage flow code that relies on this which, unfortunately, didn't seem isolated enough to just take care of as a consequence of this code. I'll make a ticket for this code removal.

Ultimately, I'd like to do a larger refactor where we isolate controllers for the hub and intake a bit more so that we're not trying to load current intake in the hub, etc.

app/controllers/hub/organizations_controller.rb
bytheway875: Oh that's way better.

PR #1128, opened 2021-05-26
Still needs help flow: Block client login as needed

alaroia-cfa:
When clients are marked as `triggered_still_needs_help_at.present?`, they won't be able to use the portal. They'll have to answer questions about if they still need help.

The controllers for the still needs help flow won't use `require_client_login` so it's peaceful to adjust `require_client_login`.

## About this pull request

This is safe to deploy as-is. No clients have `triggered_still_needs_help_at.present?` in production.

This is part 1 of N to build the [Still needs help](https://www.pivotaltracker.com/n/projects/2409240/stories/178206756) flow.

It adds a feature spec with a `return` in the middle, to mark how much progress has been made.

I'm hoping to use about 5 pull requests to deliver this feature, to avoid creating a situation where merge conflicts create trouble, and to give me a sensation of hope and forward progress.

## Pivotal Tracker story link

[#178206756]

### DIFF
--- a/app/controllers/concerns/client_access_control_concern.rb
+++ b/app/controllers/concerns/client_access_control_concern.rb
@@ -4,9 +4,13 @@ module ClientAccessControlConcern
   private
 
   def require_client_login
-    unless current_client.present?
+    if current_client.blank?
       session[:after_client_login_path] = request.original_fullpath if request.get?
-      redirect_to new_portal_client_login_path
+      redirect_to new_portal_client_login_path and return
+    end
+
+    if current_client.triggered_still_needs_help_at.present?
+      redirect_to portal_still_needs_helps_path
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/client_access_control_concern.rb
```diff
@@ -4,9 +4,13 @@ module ClientAccessControlConcern
...
 
   def require_client_login
-    unless current_client.present?
+    if current_client.blank?
       session[:after_client_login_path] = request.original_fullpath if request.get?
-      redirect_to new_portal_client_login_path
+      redirect_to new_portal_client_login_path and return
+    end
+
+    if current_client.triggered_still_needs_help_at.present?
```
bytheway875, line 12: Consider extracting the still_needs_help logic into its own method due to single responsibility principle of methods -- method name is require_client_login, but this is now also doing a redirect to a still needs help flow for clients who are logged in.

jenny-heath: we just reviewed this - seems like a good idea and will refactor after merging this

PR #1132, opened 2021-05-27
Still needs help 178206756

alaroia-cfa:


app/controllers/concerns/still_needs_help_access_control_concern.rb
```diff
@@ -0,0 +1,14 @@
+module StillNeedsHelpAccessControlConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def require_still_needs_help_client_login
+    if current_client.nil?
```
bytheway875, line 7: Building off of suggestion from the previous PR, I suggest reverting require_client_login to only handle login logic. Then create a new method in this controller that redirects TO still needs help flow, and use the before filters in conjunction with each other where appropriate instead of duplicating the session[:after_client_login_path] logic and redirect here for unauthenticated users.

app/controllers/portal/still_needs_helps_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Portal
+  class StillNeedsHelpsController < ApplicationController
+    include StillNeedsHelpAccessControlConcern
+    before_action :require_still_needs_help_client_login
```
bytheway875, line 4: If you do the refactor suggested above, you can inherit from the PortalController which already requires logging in, and then just redirect away from here in the case that they should not be seeing the still needs help stuff.

app/controllers/concerns/client_access_control_concern.rb
```diff
@@ -4,9 +4,13 @@ module ClientAccessControlConcern
...
 
   def require_client_login
-    unless current_client.present?
+    if current_client.blank?
       session[:after_client_login_path] = request.original_fullpath if request.get?
-      redirect_to new_portal_client_login_path
+      redirect_to new_portal_client_login_path and return
+    end
+
+    if StillNeedsHelpService.must_show_still_needs_help_flow?(current_client)
```
bytheway875, line 12: See previous PR suggestion to break this latter part of the method into its own method. You could update portal controller to chain the before_actions :require_client_login, :redirect_if_still_needs_help

app/controllers/portal/still_needs_helps_controller.rb
bytheway875: with all suggested refactors, this would look like:

```
module Portal
  class StillNeedsHelpsController < PortalController
  skip_before_action :redirect_if_still_needs_help
  before_action :redirect_unless_still_needs_help
  ...
```

The suggested method naming is a bit confusing, but I think in context it might make sense -- you are in the still needs help area, so we want to do the opposite thing we do if you're in another area and may need help

app/controllers/concerns/still_needs_help_access_control_concern.rb
```diff
@@ -0,0 +1,14 @@
...
+      session[:after_client_login_path] = request.original_fullpath if request.get?
+      redirect_to new_portal_client_login_path and return
+    end
+
+    redirect_to portal_root_path unless StillNeedsHelpService.may_show_still_needs_help_flow?(current_client)
```
jenny-heath, line 12: maybe omit this check for now

PR #1133, opened 2021-05-27
CTC FAQ

jenny-heath:
there still isn't a link to this anywhere

Screen Shot 2021-05-27 at 9 18 55 AM


app/controllers/public_pages_controller.rb
```diff
@@ -40,4 +40,8 @@ def stimulus_recommendation; end
   def sms_terms; end
 
   def diy; end
+
+  def faq; end
```
jenny-heath, line 44: oh, not sure where this came from - must have already been there when i took over the story

coltborg: Ya @caitbot and I added that. It seemed like it didn't need to be defined (I assume rails implicitly creates it since there's a view) and it still worked. But we wanted to be explicit so that folks can navigate from the controller to the template.

bytheway875:
Super nitpicky, but I noticed that the "subtitles" in the Questions about your tax return section  of the screenshot have inconsistent punctuation. One has a period and the other doesn't!

app/assets/stylesheets/templates/_ctc-faq.scss
```diff
@@ -0,0 +1,12 @@
+.ctc-faq {
+  .slab--stimulus-blue {
```
bytheway875, line 2: This seems like a potentially odd way to target a specific slab. The modifier class of --stimulus-blue is primarily there to make it change color, but now we're also using it to override other things. In the hypothetical case where we add a second slab--stimulus-blue to this page, do we want it to act the same way? In the case that we want to change the color of the slab to purple or pink, do we want the div to lose these modifications?

I suggest potentially adding another modifier class to the slab you want to target in order to isolate the modifications OR take the entirely opposite approach and write a modifier class that affects grid directly to add flex/align center methods (which would likely be reuseable and may already exist?), and then use the spacing-bottom-0 modifier class on the h3 -- that will prevent about 10~ lines of new css generated in our resultant code.

config/routes.rb
```diff
@@ -91,6 +91,7 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
     get "/about-us", to: "public_pages#about_us"
     get "/tax-questions", to: "public_pages#tax_questions"
     get "/faq", to: "public_pages#faq"
+    get "/ctc_faq", to: "public_pages#ctc_faq"
```
bytheway875, line 94: maybe we should dasherize this url?

bytheway875 (approved):

PR #1137, opened 2021-05-27
Suspend/activate users on organizations

caitbot:

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -64,14 +64,13 @@ class User < ApplicationRecord
   has_many :notifications, class_name: "UserNotification"
   belongs_to :role, polymorphic: true
 
-  belongs_to :organization_lead_role, -> { where(users: { role_type: 'OrganizationLeadRole' }) }, foreign_key: 'role_id', optional: true
-
   attr_encrypted :access_token, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
 
   validates_presence_of :name
   validates_inclusion_of :timezone, in: ActiveSupport::TimeZone.country_zones("us").map { |tz| tz.tzinfo.name }
 
   scope :active, -> { where(suspended_at: nil) }
+  scope :suspended, -> { where.not(suspended_at: nil) }
 
   def accessible_coalitions
     case role_type
@@ -213,6 +212,10 @@ class User < ApplicationRecord
     suspended_at.present?
   end
 
+  def active?
+    !suspended?
+  end
+
   def active_for_authentication?
     # overrides
     super && !suspended?
@@ -221,5 +224,9 @@ class User < ApplicationRecord
   def suspend!
     assigned_tax_returns.update(assigned_user: nil)
     update!(suspended_at: DateTime.now)
+  end
+
+  def activate!
+    update!(suspended_at: nil)
   end
 end
--- a/app/models/vita_partner.rb
+++ b/app/models/vita_partner.rb
@@ -30,7 +30,6 @@ class VitaPartner < ApplicationRecord
   has_many :clients
   has_many :intakes
   has_many :source_parameters
-  has_many :users
   has_many :serviced_zip_codes, class_name: "VitaPartnerZipCode"
   has_many :serviced_states, class_name: "VitaPartnerState"
   belongs_to :parent_organization, class_name: "VitaPartner", optional: true
@@ -77,6 +76,18 @@ class VitaPartner < ApplicationRecord
   def self.client_support_org
     # When a person messages us, but their contact info does not match any Client, link them to this org.
     VitaPartner.find_by!(name: "GYR National Organization")
+  end
+
+  def organization_leads
+    User.where(role: OrganizationLeadRole.where(organization: self))
+  end
+
+  def site_coordinators
+    User.where(role: SiteCoordinatorRole.where(site: organization? ? child_sites : self))
+  end
+
+  def team_members
+    User.where(role: TeamMemberRole.where(site: organization? ? child_sites : self))
   end
 
   private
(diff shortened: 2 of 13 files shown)


app/models/user.rb
```diff
@@ -64,14 +64,13 @@ class User < ApplicationRecord
   has_many :notifications, class_name: "UserNotification"
   belongs_to :role, polymorphic: true
 
-  belongs_to :organization_lead_role, -> { where(users: { role_type: 'OrganizationLeadRole' }) }, foreign_key: 'role_id', optional: true
```
caitbot, line 67: @alaroia-cfa why does this line of code exist? Is it ok to remove? We could not find usage of it anywhere and didn't really understand what it is meant to do.

app/models/vita_partner.rb
```diff
@@ -30,7 +30,6 @@ class VitaPartner < ApplicationRecord
   has_many :clients
   has_many :intakes
   has_many :source_parameters
-  has_many :users
```
caitbot, line 33: We opted to remove this because `User` has no foreign key referencing `VitaPartner`.

Rather, `Role` has the foreign key for a `VitaPartner` and `User` has the foreign key for a `Role`.

Optionally, we could probably set up a through relationship on `User` -> `Role` -> `VitaPartner` but that wouldn't have helped us with this ticket since we start from a `VitaPartner`.

app/models/vita_partner.rb
```diff
@@ -79,6 +78,20 @@ def self.client_support_org
     VitaPartner.find_by!(name: "GYR National Organization")
   end
 
+  def organization_leads
+    User.where(role_type: OrganizationLeadRole::TYPE, role_id: OrganizationLeadRole.where(organization: self).pluck(:id))
```
caitbot, line 82: @alaroia-cfa any thoughts on how to simplify this? We really wanted to set up a through relationship on VitaPartner -> Role -> User but the foreign key setup is in the other direction (ie User -> Role -> VitaPartner) so this is where we landed.

app/models/user.rb
alaroia-cfa: I remember there being some query I wrote once that depended on this line.

I re-read the commit that introduced the line, and it seems to be:

```
      @assignable_users = User.preload(:role).joins(:organization_lead_role).where("organization_lead_roles.vita_partner_id = ?", @client.vita_partner_id)
```

in `hub/tax_returns_controller.rb`. I believe the `belongs_to` was needed for the `.joins(:organization_lead_role)` to work.

I see the code has changed to no longer do `.joins(:organization_lead_role)`. Therefore I believe that the `belongs_to` can be safely deleted.

app/models/vita_partner.rb
alaroia-cfa: Can you do this?

```
User.where(role: OrganizationLeadRole.where(organization: self))
```

When I test that out in `rails c`, it seems to do something reasonable.

```
irb(main):012:0> some_org = VitaPartner.organizations.third
  VitaPartner Load (1.6ms)  SELECT "vita_partners".* FROM "vita_partners" WHERE "vita_partners"."parent_organization_id" IS NULL ORDER BY "vita_partners"."name" ASC LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 2]]
  VitaPartner Load (0.5ms)  SELECT "vita_partners".* FROM "vita_partners" WHERE "vita_partners"."parent_organization_id" = $1 ORDER BY "vita_partners"."name" ASC, "vita_partners"."id" ASC  [["parent_organization_id", 2]]
  VitaPartner Load (0.4ms)  SELECT "vita_partners".* FROM "vita_partners" WHERE "vita_partners"."parent_organization_id" = $1 ORDER BY "vita_partners"."name" ASC, "vita_partners"."id" ASC  [["parent_organization_id", 4]]
=> #<VitaPartner id: 2, allows_greeters: nil, archived: false, capacity_limit: nil, coalition_id: 1, created_at: "2021-04-15 00:27:16", logo_path: nil, name: "Oregano Org", national_overflow_location: false, parent_organization_id: nil, timezone: "America/New_York", updated_at: "2021-04-15 00:27:16">

irb(main):013:0> User.where(role: OrganizationLeadRole.where(organization: some_org))
  User Load (3.0ms)  SELECT "users".* FROM "users" WHERE ("users"."role_type" = $1) AND ("users"."role_id" IN (SELECT "organization_lead_roles"."id" FROM "organization_lead_roles" WHERE "organization_lead_roles"."vita_partner_id" = $2)) LIMIT $3  [["role_type", "OrganizationLeadRole"], ["vita_partner_id", 2], ["LIMIT", 11]]
=> #<ActiveRecord::Relation [#<User id: 1, created_at: "2021-04-15 00:27:17", email: "skywalker@example.com", encrypted_access_token: nil, encrypted_access_token_iv: nil, name: "Luke Skywalker", phone_number: nil, role_id: 1, role_type: "OrganizationLeadRole", suspended_at: nil, timezone: "America/New_York", updated_at: "2021-04-15 00:27:17", access_token: nil>, #<User id: 5, created_at: "2021-04-15 00:27:18", email: "princess@example.com", encrypted_access_token: nil, encrypted_access_token_iv: nil, name: "Lea Amidala Organa", phone_number: nil, role_id: 2, role_type: "OrganizationLeadRole", suspended_at: nil, timezone: "America/New_York", updated_at: "2021-04-15 00:27:18", access_token: nil>]>
```

app/models/user.rb
```diff
@@ -213,6 +212,10 @@ def suspended?
     suspended_at.present?
   end
 
+  def active?
```
alaroia-cfa, line 215: Consider not defining a new method and relying on `active_for_authentication?` in the tests. That will prove that Devise won't log the user in, which seems like a great test.

alaroia-cfa (approved):
This looks great; a handful of non-blocking comments.

PR #1138, opened 2021-05-28
Add sla_breach_date filter to client sortable

embarnard:


alaroia-cfa (changes requested):
One blocker: I'd like to see a unit test in the client sortable spec covering this new functionality.

Two non-blocking suggestions:

* This change doesn't seem to need `search_and_sort_params` to be public, but it makes it public. Consider leaving it private.

* If you're making this change, I'd like to see the unattended clients controller be adjusted to use this new functionality.

alaroia-cfa:
> If we are using the search_and_sort_params as a helper_method in other controllers it would have to be public right?

I can't remember if the `helper_method :search_and_sort_params` declaration requires that `search_and_sort_params` be public (though when we were pairing before, I think I assumed it had to be.)

Also there's a chance that if you rely on `@filters` you won't need `search_and_sort_params`.

PR #1139, opened 2021-05-28
Add drop off survey link for drop off clients and add Organization name to completion email

bytheway875:
https://www.pivotaltracker.com/story/show/177386999 

The ticket also asked that the messages include the organization name that prepared the taxes.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1027,14 +1027,14 @@ en:
       completion:
         email:
           body: |
-            Thank you for filing your taxes with GetYourRefund! You can check the status of your refund at the IRS website https://www.irs.gov/refunds.
+            Thank you for filing your taxes with GetYourRefund and <<Client.AssignedOrganization>>! You can check the status of your refund at the IRS website https://www.irs.gov/refunds.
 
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -1027,14 +1027,14 @@ en:
       completion:
         email:
           body: |
-            Thank you for filing your taxes with GetYourRefund! You can check the status of your refund at the IRS website https://www.irs.gov/refunds.
+            Thank you for filing your taxes with GetYourRefund and <<Client.AssignedOrganization>>! You can check the status of your refund at the IRS website https://www.irs.gov/refunds.
```
caitbot, line 1030: wooah, I didn't know you could do this. where does `Client` come from?

caitbot (approved):
looks good, left a question because I am unfamiliar w/ the message body syntax (:

config/locales/en.yml
bytheway875: This is all set up in the ReplacementParametersService, which replaces everything, and then ClientMessagingService ensures that all email bodies are run through the ReplacementParametersService before they get sent.

PR #1144, opened 2021-05-28
Remove arp triage page

caitbot:


spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -29,9 +29,30 @@
     expect(page).to have_selector("h1", text: "Have you filed taxes for 2017, 2018, and 2019?")
     click_on "No"
 
-    expect(page).to have_selector("h1", text: "Your tax return may be delayed so that we can ensure you receive the highest refund!")
+    expect(page).to have_selector("p", text: "Do any of the situations below apply to 2020?")
```
caitbot, line 32: It is weird to me that before the ARP page would allow you to skip past all of these questions and now they are required. This seems to be due to falling back on `super` for `next_path`.

caitbot: Talked with @yvonnefong about this and we definitely DO want to skip these questions in order to get non-filers thru triage and into the full service flow ASAP. Going to add back `next_path` to do that.

app/controllers/questions/triage_backtaxes_controller.rb
```diff
@@ -9,7 +9,7 @@ def illustration_path
     end
 
     def next_path
-      @form.filed_previous_years? ? super : triage_arp_questions_path
+      @form.filed_previous_years? ? super : file_with_help_questions_path
```
caitbot, line 12: From my discussion with @yvonnefong, we want non-filers to be routed directly to file with help, in the hopes that it will get more of them to file. Prior to this change, the triage ARP question routed them there after letting them know there may be a delay.

spec/controllers/questions/triage_lookback_controller_spec.rb
```diff
@@ -31,9 +31,9 @@
 
     context "when at least one situation is yes" do
```
jenny-heath, line 32: looks like this now tests basically the same thing as the next context block? unless i'm missing a meaningful difference you can probably delete this one

spec/controllers/questions/triage_prepare_solo_controller_spec.rb
```diff
@@ -39,7 +39,7 @@
       let(:will_prepare) { "no" }
       it "redirects to the next triage question" do
```
jenny-heath, line 40: `it "redirects to the file with help page"`

spec/features/web_intake/new_joint_filers_spec.rb
```diff
@@ -30,40 +30,15 @@
...
     check "I purchased health insurance through the marketplace"
     click_on "Continue"
 
-    expect(page).to have_selector("h1", text: "Your tax return may be delayed so that we can ensure you receive the highest refund!")
-    click_on "Go back"
-
-    expect(page).to have_selector("p", text: "Do any of the situations below apply to 2020?")
-    check "None of the above"
-    click_on "Continue"
-
```
jenny-heath, line 66: wow this was a weird sequence to have in this feature spec...glad we're getting rid of it!

jenny-heath (approved):
just a couple of small suggestions, otherwise looks good!

PR #1147, opened 2021-06-01
Select clients beyond pagination for bulk edits

embarnard:
[Select clients beyond pagination for bulk edits #178205976](https://www.pivotaltracker.com/story/show/178205976)

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -35,11 +35,11 @@ module ClientSortable
     overlapping_keys.any? && hash_params.slice(*overlapping_keys).any? { |_, v| v.present? }
   end
 
+  private
+
   def search_and_sort_params
     [:search, :status, :unassigned, :assigned_to_me, :flagged, :unemployment_income, :year, :vita_partners, :assigned_user_id, :language, :service_type, :greetable, :sla_breach_date]
   end
-
-  private
 
   def setup_sortable_client
     delete_cookie if params[:clear]
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -18,6 +18,7 @@ module ClientSortable
     clients = clients.where(tax_returns: { service_type: @filters[:service_type] }) if @filters[:service_type].present?
     clients = clients.where(intake: Intake.where(had_unemployment_income: "yes")) if @filters[:unemployment_income].present?
     clients = clients.where(vita_partner: VitaPartner.allows_greeters) if @filters[:greetable].present?
+    clients = clients.first_unanswered_incoming_interaction_communication_breaches(@filters[:sla_breach_date]) if @filters[:sla_breach_date].present?
 
     if @filters[:vita_partners].present?
       ids = JSON.parse(@filters[:vita_partners]).map { |vita_partner| vita_partner["id"] }
@@ -35,7 +36,7 @@ module ClientSortable
   end
 
   def search_and_sort_params
-    [:search, :status, :unassigned, :assigned_to_me, :flagged, :unemployment_income, :year, :vita_partners, :assigned_user_id, :language, :service_type, :greetable]
+    [:search, :status, :unassigned, :assigned_to_me, :flagged, :unemployment_income, :year, :vita_partner_id, :assigned_user_id, :language, :service_type, :greetable, :sla_breach_date]
   end
 
   private
@@ -72,6 +73,7 @@ module ClientSortable
       language: source[:language],
       service_type: source[:service_type],
       greetable: source[:greetable],
+      sla_breach_date: source[:sla_breach_date],
     }
   end
 
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -4,7 +4,7 @@ module ClientSortable
     setup_sortable_client unless @filters.present?
     clients = if current_user&.greeter?
                 # Greeters should only have "search" access to clients in intake stage AND clients assigned to them.
-                @clients.greetable.or(Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct)
+                @clients.greetable || Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct
               else
                 @clients.after_consent
               end
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -36,7 +36,7 @@ module ClientSortable
   end
 
   def search_and_sort_params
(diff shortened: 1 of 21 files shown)


spec/controllers/concerns/client_sortable_spec.rb
```diff
@@ -64,6 +80,19 @@ def filter_cookie_name
...
+    #   before do
+    #     subject.instance_variable_set(@always_on_filters, { assigned_to_me: true })
+    #   end
+    #
+    #   it "merges the always-on filters with the filters" do
+    #     subject.filtered_and_sorted_clients
+    #     expect(assigns(:filters)[:assigned_to_me]).to be_present
+    #   end
+    # end
+
```
coltborg, line 95: @embarnard should this still be commented out? Can it be returned or deleted?

embarnard: yeah we should just remove this

app/javascript/lib/bulk_action.js
```diff
@@ -10,13 +10,19 @@ function changeElDisplay(el, property) {
...
-function handleCheckboxChange(formEl, takeActionCountEl, takeActionFooterEl) {
+function handleCheckboxChange(formEl, takeActionCountEl, takeActionFooterEl, totalTaxReturnsCount, takeActionAllEl) {
     changeCountDisplay(takeActionCountEl, getCheckedNum(formEl));
     if (getCheckedNum(formEl) >= 1) {
         changeElDisplay(takeActionFooterEl, "");
     } else if (getCheckedNum(formEl) === 0) {
         changeElDisplay(takeActionFooterEl, "none");
     }
+
+    if (getCheckedNum(formEl) == totalTaxReturnsCount) {
```
caitbot, line 21: For my understanding -- this is if user has checked all TRs (because they are all on one page?) then do NOT show the option to select all?

app/controllers/hub/tax_return_selections_controller.rb
```diff
@@ -30,9 +30,10 @@ def create
     end
 
     def new
-      @tr_ids = new_params[:tr_ids]
-      @client_count = Client.accessible_by(current_ability).distinct.joins(:tax_returns).where(tax_returns: { id: @tr_ids }).count
-      @tax_return_count = TaxReturn.accessible_by(current_ability).where(id: @tr_ids).count
+      @clients = Client.accessible_by(current_ability).with_eager_loaded_associations
+      @tr_ids = (new_params.dig(:create_tax_return_selection, :action_type) == "filtered-clients") ? TaxReturn.where(client: filtered_and_sorted_clients).pluck(:id) : new_params[:tr_ids]
```
caitbot, line 34: `== "filtered-clients"` is just checking that its coming from "Take Action on All" right? If that is the case, I would consider renaming it to reflect that. Maybe even `all-filtered-clients`. What my brain found confusing is that they don't necessarily have to be filtered by the user prior to a bulk-action being initiated on all of the tax returns. And so this took me a second to wrap my head around.

Is it possible that we could use `filtered_and_sorted_clients` to get `@clients`, it would save us from having to do the sql query for `@client_count` below?

app/models/client.rb
```diff
@@ -151,6 +151,10 @@ def self.sortable_intake_attributes
     accessible_by(Ability.new(user))
   end
 
+  scope :without_pagination, -> do
+    except(:limit, :offset)
```
caitbot, line 155: What does this do?

caitbot (approved):
This looks good and for the most part makes sense to me. 

I left some questions and with it some suggestions but none are blocking.

app/controllers/hub/tax_return_selections_controller.rb
coltborg: I think `all-filtered-clients` makes sense. I'll change that.

coltborg: After playing around with different options, we can't using `filtered_and_sorted_clients` to first define `@clients` because that method [itself needs a reference to `@clients` 😭](https://github.com/codeforamerica/vita-min/blob/df7a250bf82172d6617cae60d66d4e0c99af11b0/app/controllers/concerns/client_sortable.rb#L13). 

Looking at the mini-profiler, the callers are pretty small because their counts. I'm going to stick with this implementation for now.

app/javascript/lib/bulk_action.js
coltborg: Thanks for pointing this out! I also wasn't 100% on this. After talking with Suran, here's the [criteria on when the link should show up](https://www.pivotaltracker.com/story/show/178205976/comments/224692402).

coltborg: Here's a [link to a video](https://www.pivotaltracker.com/story/show/178205976/comments/224693261) showing the situations where is shows/doesn't show.

app/controllers/concerns/client_sortable.rb
```diff
@@ -1,14 +1,17 @@
...
+    @default_order = default_order
     setup_sortable_client unless @filters.present?
+    filtered_clients.delegated_order(@sort_column, @sort_order)
+  end
+
+  def filtered_clients
     clients = if current_user&.greeter?
                 # Greeters should only have "search" access to clients in intake stage AND clients assigned to them.
-                @clients.greetable.or(Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct)
+                @clients.greetable || Client.joins(:tax_returns).where(tax_returns: { assigned_user: current_user }).distinct
```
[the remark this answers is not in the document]
coltborg: This wasn't working the same for @embarnard and I. We were getting an error when trying to login as a greeter locally with the `.or()` approach.

app/models/client.rb
coltborg: This removes the pagination limit from the pagination gem, I added [comments to point to the documentation](https://github.com/kaminari/kaminari#unscoping).

app/javascript/lib/bulk_action.js
```diff
@@ -10,14 +10,16 @@ function changeElDisplay(el, property) {
...
+function handleCheckboxChange(formEl, takeActionCountEl, takeActionFooterEl) {
     changeCountDisplay(takeActionCountEl, getCheckedNum(formEl));
     if (getCheckedNum(formEl) >= 1) {
         changeElDisplay(takeActionFooterEl, "");
     } else if (getCheckedNum(formEl) === 0) {
         changeElDisplay(takeActionFooterEl, "none");
     }
+}
 
+function showTakeActionAll(formEl, totalTaxReturnsCount, takeActionAllEl) {
```
caitbot, line 22: oh I like this decision to separate it out, much easier to grok

caitbot (approved):
double approved~!

PR #1150, opened 2021-06-03
Only populate nil message bodies for present locales on bulk status updates

caitbot:
Fixes https://www.pivotaltracker.com/story/show/178397489 and https://www.pivotaltracker.com/story/show/178259385

### DIFF
--- a/app/forms/hub/bulk_action_form.rb
+++ b/app/forms/hub/bulk_action_form.rb
@@ -18,8 +18,9 @@ module Hub
     def initialize(tax_return_selection, *args, **attributes)
       @tax_return_selection = tax_return_selection
       super(*args, **attributes)
-      set_default_message_body("es") if @message_body_es.blank?
-      set_default_message_body("en") if @message_body_en.blank?
+      locale_counts = @tax_return_selection.clients.locale_counts
+      set_default_message_body("es") if @message_body_es.nil? && locale_counts["es"].nonzero?
+      set_default_message_body("en") if @message_body_en.nil? && locale_counts["en"].nonzero?
     end
 
     def assigned_user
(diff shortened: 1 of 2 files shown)


app/forms/hub/bulk_action_form.rb
```diff
@@ -18,8 +18,9 @@ class BulkActionForm < Form
     def initialize(tax_return_selection, *args, **attributes)
       @tax_return_selection = tax_return_selection
       super(*args, **attributes)
-      set_default_message_body("es") if @message_body_es.blank?
```
caitbot, line 21: Oops, `blank` means it will overwrite a blank string(`""`) which is set by the user.

When initialized the form param should be `nil` and so this is a better check and matches what we do on the individual `TakeActionForm`.

app/forms/hub/bulk_action_form.rb
```diff
@@ -18,8 +18,9 @@ class BulkActionForm < Form
...
-      set_default_message_body("en") if @message_body_en.blank?
+      locale_counts = @tax_return_selection.clients.locale_counts
```
caitbot, line 21: If we only change `blank` to `nil` there is an issue where a user set blank string will trigger a "missing locales error" and this is due to us ALWAYS setting both the `es` and `en` message bodies even if there are no clients which they are applicable to. This was sneaky and due to cause a bug but was "ok" for a while because we just "ignored" the populated message body (ie it didn't send to anyone). Turns out it was causing a [bug](https://www.pivotaltracker.com/story/show/178259385) because it was populating the message body with a message too long that no one could delete 🤦🏻 oops.

Adding the `locale_counts` check means that we only ever populate message bodies that a) have UI b) are needed.

coltborg:
https://www.pivotaltracker.com/story/show/178259385 is not fixed in this branch. This might not stop this PR from going out, it might need its own PR.


https://user-images.githubusercontent.com/9101728/120714419-c7b42380-c488-11eb-84c0-9556e9b91b69.mov

coltborg (approved):
Both bugs are resolved locally, 🚢

PR #1153, opened 2021-06-04
Still needs help form #178206756

jenny-heath:
image

image

image


app/views/portal/still_needs_helps/edit.html.erb
```diff
@@ -0,0 +1,28 @@
...
+        <main role="main">
+          <div class="question__illustration">
+            <%= image_tag("portal/still-need-help.svg", alt: "") %>
+          </div>
+          <h1 class="h2 spacing-above-0"><%= t(".title") %></h1>
+
+          <p class="help-text"><%= t(".help_text") %></p>
+
+          <%= form_with(url: portal_update_still_needs_help_path, local: true, method: :put, builder: VitaMinFormBuilder) do |f| %>
+            <%= f.hidden_field :still_needs_help, value: "yes" %>
```
alaroia-cfa, line 16: I learned something reading this pull request - comment below, not a blocker.

This `f.hidden_field` pattern is OK (and I'm the one who added it).

I looked at the satisfaction survey HTML, and it has a different pattern: `name: "#{f.object_name}[experience_survey]", value: "neutral"` within the `f.button :submit`. I like the satisfaction survey pattern more, and you could apply it here.

If you make that change, no need for re-review.

config/locales/es.yml
```diff
@@ -1141,6 +1141,21 @@ es:
         return: "Volver a la página de bienvenida"
         title: "¡Gracias por enviar su firma final del %{year}!"
         description: "¿Desea descargar sus formularios de impuestos finales?"
+    still_needs_helps:
+      edit:
+        title: "🦐 Are you still interested in filing your taxes with us?"
```
alaroia-cfa, line 1146: Needs translation

spec/controllers/portal/still_needs_helps_controller_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+  describe "#chat_later" do
+    it_behaves_like :a_get_action_for_authenticated_clients_only, action: :chat_later
+  end
+
+  describe "#no_longer_needs_help" do
+    it_behaves_like :a_get_action_for_authenticated_clients_only, action: :no_longer_needs_help
+  end
+
+  describe "#experience_survey" do
+    it_behaves_like :a_post_action_for_authenticated_clients_only, action: :experience_survey
```
alaroia-cfa, line 115: I think the way the shared example works, it's best if you provide meaningful params on this line. Not a blocker.

alaroia-cfa (approved):
Looks good to me.

Suggestions for follow-ups:

* Finish translation strings (not a blocker since there are no clients in the still needs help state right now)
* Other low-priority comments

PR #1155, opened 2021-06-07
Cleanup of multiple vita partner query

caitbot:


app/controllers/concerns/client_sortable.rb
```diff
@@ -25,7 +25,7 @@ def filtered_clients
 
     if @filters[:vita_partners].present?
       ids = JSON.parse(@filters[:vita_partners]).map { |vita_partner| vita_partner["id"] }
-      clients = clients.where('vita_partners.id IN (:id)', id: ids)
```
caitbot, line 28: Per a conversation with @alaroia-cfa 

> This line conceptually asks the database to do a JOIN before it can filter.
> 
> >       clients = clients.where('vita_partners.id IN (:id)', id: ids)
> 
> If you filter instead on a column on clients, you can filter before the JOIN, which may be faster. For example, this line does that.
> 
> >       clients = clients.where('vita_partner_id IN (:id), id: ids)
> 
> or this, which turns into the same SQL:
> 
> >       clients = clients.where(vita_partner_id: ids)
> 
> Doing the filter before the JOIN can possibly help a lot with performance. If you have 1000 clients and they have 10 bytes of data each (10KB), and the JOIN will 2x the amount of data the database needs to process, it needs to read 20KB of data before it can apply the filtering. By contrast, if the database can filter first, it can narrow to just the (say) 10 clients with matching vita_partner_id before doing the JOIN, so approx 100 bytes of client data read, then 200 bytes after the JOIN. The numbers I've picked are very conservative. We have >100K clients and >100 vita partners, so they could start to meaningfully add up, and our JOIN is across a lot of tables, so filtering first might be even more important.
> 
> Given that, the performance-minded way to write it would be as a filter before the JOIN - "vita_partner_id" not "vita_partners.id". However, in practice, Postgres's query optimizer might notice that "vita_partner_id" and "vita_partners.id" are equal, and it may do the filter before the JOIN.
> 
> I bring this all up because if you see me write code that uses the "*_id" field, it's because I'm trying to filter before JOIN. I bucket this under the trendy term "mechanical sympathy."

spec/controllers/concerns/client_sortable_spec.rb
```diff
@@ -117,7 +117,7 @@ def filter_cookie_name
 
       it "creates a query for the search and scopes to vita partner" do
         expect(subject.filtered_and_sorted_clients).to eq clients_query_double
-        expect(clients_query_double).to have_received(:where).with('vita_partners.id IN (:id)', id: [vita_partner.id])
+        expect(clients_query_double).to have_received(:where).with({:vita_partner_id=>[vita_partner.id]})
```
alaroia-cfa, line 120: Syntax suggestion: You can use `.with(vita_partner_id: [vita_partner.id])` I believe.

Not a blocker. If you make that change, feel free to push.

alaroia-cfa (approved):
Looks good! One non blocking suggestion. Cheers!

PR #1156, opened 2021-06-07
Spike - Add demo of different content on different domains

alaroia-cfa:
localhost:3000/ looks like

image

ctc.localhost:3000/ looks like

image

### DIFF
--- a/app/lib/routes/ctc_domain.rb
+++ b/app/lib/routes/ctc_domain.rb
@@ -1,0 +1,9 @@
+module Routes
+  class CtcDomain
+    def matches?(request)
+      app_base_domains = ["localhost", "getyourrefund.org", "staging.getyourrefund.org", "demo.getyourrefund.org"]
+      ctc_hosts = app_base_domains.map { |domain| ["ctc.#{domain}", "www.ctc.#{domain}"] }.flatten
+      ctc_hosts.include?(request.host)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


bytheway875:
@alaroia-cfa will you take me through what actually changed in the routes files?

app/lib/routes/ctc_domain.rb
```diff
@@ -0,0 +1,9 @@
+module Routes
+  class CtcDomain
+    def matches?(request)
+      app_base_domains = ["localhost", "getyourrefund.org", "staging.getyourrefund.org", "demo.getyourrefund.org"]
+      ctc_hosts = app_base_domains.map { |domain| ["ctc.#{domain}", "www.ctc.#{domain}"] }.flatten
```
bytheway875, line 5: Does this mean that the ctc domains should work fine out of the box in demo and staging? Should we attempt to push this to staging to validate that?

bytheway875:
Screen Shot 2021-06-07 at 5 01 39 PM
Screen Shot 2021-06-07 at 5 02 50 PM


I think there are some things we need to configure in order to get this working on staging. Before we commit to going with this approach, should we ensure we can get it working on the staging environment?

bytheway875:
Got this working on staging after doing the Route53/Aptible setup that was necessary!

Screen Shot 2021-06-07 at 5 42 40 PM

PR #1157, opened 2021-06-08
Upgrade Puma to 5.3.2

alaroia-cfa:
This upgrades us from Puma 5.3.1 to Puma 5.3.2.

We're getting a handful of errors from Puma that might relate to our recent Puma 5.3.1 upgrade.

Sentry link: https://sentry.io/organizations/codeforamerica/issues/2443686659/?project=1880327&referrer=slack

Slack thread: https://cfa.slack.com/archives/G013YJMPVAA/p1623153520000500

I believe this will get rid of the errors. I don't know what caused the errors to occur today vs. other days. The devices that are triggering the problem appear to usually be smartphones; I wonder if they use HTTP keep-alive more than other clients.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ ruby '2.6.5'
 
 gem 'rack', '>= 2.0.8'
 gem 'rails', '>= 6.0.3.7'
-gem 'puma', '>= 4.3.5'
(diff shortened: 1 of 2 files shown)


caitbot (approved):
Looks good, install went okay and I clicked around locally and nothing alarming happened!

PR #1158, opened 2021-06-08
Upgrade webpacker, to upgrade js dependencies

alaroia-cfa:
I read the webpacker diff https://github.com/rails/webpacker/compare/v5.2.0...v5.4.0 and I like it.


alaroia-cfa (commented):
:ship:

PR #1159, opened 2021-06-08
Rake task for sending still-needs-help emails

caitbot:
https://www.pivotaltracker.com/story/show/178418620

### DIFF
--- a/app/jobs/send_still_needs_help_emails_job.rb
+++ b/app/jobs/send_still_needs_help_emails_job.rb
@@ -1,0 +1,55 @@
+class SendStillNeedsHelpEmailsJob < ApplicationJob
+  def perform()
+    urban_upbound = VitaPartner.find_by_name("Urban Upbound")
+    clients = Client.where(vita_partner_id: urban_upbound.id).or(Client.where(vita_partner: VitaPartner.where(parent_organization_id: urban_upbound.id)))
+
+    clients = clients.joins(:tax_returns).where(tax_returns: { status: ["intake_ready", "intake_in_progress"] }).select do |client|
+      client.tax_returns.all? { |tax_return| ["intake_ready", "intake_in_progress", "file_not_filing"].include? tax_return.status }
+    end.uniq
+
+    tax_returns = clients.map(&:tax_returns).flatten
+    tr_successes = []
+    tr_fails = []
+    tax_returns.each do |tax_return|
+      if tax_return.update(status: "file_not_filing")
+        tr_successes << tax_return.id
+      else
+        tr_fails << tax_return.id
+      end
+    end
+    puts "Successfully updated tax returns: #{tr_successes}"
+    puts "Failed to update tax returns: #{tr_fails}"
+
+    client_successes = []
+    client_fails = []
+    link = "https://getyourrefund.org/portal/still-needs-help"
+    clients.each do |client|
+      if client.update(triggered_still_needs_help_at: Time.now)
+        client_successes << client.id
+      else
+        client_fails << client.id
+      end
+
+      puts "Sending messages to Client with id #{client.id}"
+      locale = client.intake.locale
+      ClientMessagingService.send_system_email(
+        client: client,
+        body: I18n.t("messages.still_needs_help.email.body", locale: locale, link: link),
+        subject: I18n.t("messages.still_needs_help.email.subject", locale: locale),
+        locale: locale
+      )
+
+      preferred_contact_method = ClientMessagingService.contact_methods(client).keys.first
+      if preferred_contact_method == :sms_phone_number
+        ClientMessagingService.send_system_text_message(
+          client: client,
+          body: I18n.t("messages.still_needs_help.sms.body", locale: locale, link: link),
+          locale: locale
+        )
+      end
+    end
+
+    puts "Successfully updated clients: #{client_successes}"
+    puts "Failed to update clients: #{client_fails}"
+  end
+end
--- a/app/jobs/send_still_needs_help_emails_job.rb
+++ b/app/jobs/send_still_needs_help_emails_job.rb
@@ -1,30 +1,17 @@
 class SendStillNeedsHelpEmailsJob < ApplicationJob
   def perform()
     urban_upbound = VitaPartner.find_by_name("Urban Upbound")
-    clients = Client.where(vita_partner_id: urban_upbound.id).or(Client.where(vita_partner: VitaPartner.where(parent_organization_id: urban_upbound.id)))
+    clients = Client.where(vita_partner: urban_upbound).or(Client.where(vita_partner: urban_upbound.child_sites))
 
     clients = clients.joins(:tax_returns).where(tax_returns: { status: ["intake_ready", "intake_in_progress"] }).select do |client|
       client.tax_returns.all? { |tax_return| ["intake_ready", "intake_in_progress", "file_not_filing"].include? tax_return.status }
     end.uniq
 
-    tax_returns = clients.map(&:tax_returns).flatten
-    tr_successes = []
-    tr_fails = []
-    tax_returns.each do |tax_return|
-      if tax_return.update(status: "file_not_filing")
-        tr_successes << tax_return.id
-      else
-        tr_fails << tax_return.id
-      end
-    end
-    puts "Successfully updated tax returns: #{tr_successes}"
-    puts "Failed to update tax returns: #{tr_fails}"
-
     client_successes = []
     client_fails = []
     link = "https://getyourrefund.org/portal/still-needs-help"
     clients.each do |client|
-      if client.update(triggered_still_needs_help_at: Time.now)
+      if StillNeedsHelpService.trigger_still_needs_help_flow(client)
         client_successes << client.id
       else
         client_fails << client.id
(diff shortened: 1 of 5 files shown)


app/jobs/send_still_needs_help_emails_job.rb
```diff
@@ -0,0 +1,55 @@
+class SendStillNeedsHelpEmailsJob < ApplicationJob
+  def perform()
+    urban_upbound = VitaPartner.find_by_name("Urban Upbound")
+    clients = Client.where(vita_partner_id: urban_upbound.id).or(Client.where(vita_partner: VitaPartner.where(parent_organization_id: urban_upbound.id)))
+
+    clients = clients.joins(:tax_returns).where(tax_returns: { status: ["intake_ready", "intake_in_progress"] }).select do |client|
```
alaroia-cfa, line 6: The `.select` I believe isn't doing anything. Do you mean `.load`?

I'm fearful this big SQL query, plus a Ruby loop, will be very slow,. If you've tried it in prod and it's acceptably fast, great. If not, then I encourage you to avoid the loop in Ruby. Can you move the `.all?` into an ActiveRecord query?

app/jobs/send_still_needs_help_emails_job.rb
```diff
@@ -0,0 +1,55 @@
...
+    clients.each do |client|
+      if client.update(triggered_still_needs_help_at: Time.now)
+        client_successes << client.id
+      else
+        client_fails << client.id
+      end
+
+      puts "Sending messages to Client with id #{client.id}"
+      locale = client.intake.locale
+      ClientMessagingService.send_system_email(
```
alaroia-cfa, line 35: Are L35-L47 the same as `ClientMessagingService.send_message_to_all_opted_in_contact_methods`? If so, I encourage you to use that instead.

Also, `.send_system_email` assumes that the client has an email address. I recommend only calling it if `.contact_methods` has `:email` in it.

app/jobs/send_still_needs_help_emails_job.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+    end
+    puts "Successfully updated tax returns: #{tr_successes}"
+    puts "Failed to update tax returns: #{tr_fails}"
+
+    client_successes = []
+    client_fails = []
+    link = "https://getyourrefund.org/portal/still-needs-help"
+    clients.each do |client|
+      if client.update(triggered_still_needs_help_at: Time.now)
```
alaroia-cfa, line 27: I think you can (and probably should) use `StillNeedsHelpService.trigger_still_needs_help_flow` so that the TaxReturn records are updated.

alaroia-cfa (commented):
I left a few comments; I strongly recommend examining if you can rely on `StillNeedsHelpService.trigger_still_needs_help_flow`. If you can't, that's OK; let me know and I'll re-review.

app/jobs/send_still_needs_help_emails_job.rb
```diff
@@ -0,0 +1,55 @@
+class SendStillNeedsHelpEmailsJob < ApplicationJob
+  def perform()
+    urban_upbound = VitaPartner.find_by_name("Urban Upbound")
+    clients = Client.where(vita_partner_id: urban_upbound.id).or(Client.where(vita_partner: VitaPartner.where(parent_organization_id: urban_upbound.id)))
+
+    clients = clients.joins(:tax_returns).where(tax_returns: { status: ["intake_ready", "intake_in_progress"] }).select do |client|
+      client.tax_returns.all? { |tax_return| ["intake_ready", "intake_in_progress", "file_not_filing"].include? tax_return.status }
+    end.uniq
+
+    tax_returns = clients.map(&:tax_returns).flatten
```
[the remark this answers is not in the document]
caitbot: do you mean `find_each` in place of `map`? and then mapping over the batches?

I replaced this with the `StillNeedsHelpService.trigger_still_needs_help_flow` so this `map` is irrelevant now but I'm still curious for learnings sake how to optimize this map because I was also concerned about slowness.

app/jobs/send_still_needs_help_emails_job.rb
caitbot: We want to always send to email if they have an email and only send to text if they _prefer_ text, per a convo w/ Yvonne. 

I suspected that this was a tiny bit different from `ClientMessagingService.send_message_to_all_opted_in_contact_methods` but looking at Clients now, I realize that you can only be `opted-in` and there is no measure of preference. Going to update to'ClientMessagingService.send_message_to_all_opted_in_contact_methods'.

app/jobs/send_still_needs_help_emails_job.rb
caitbot: Beautiful, did not realize this existed, ty ty.

app/jobs/send_still_needs_help_emails_job.rb
caitbot: Ooooh actually thinking about this a bit more, the intent was to only send a text message if text message is the ONLY means of contact. Which is in fact different from `ClientMessagingService.send_message_to_all_opted_in_contact_methods` and is the reason for `    preferred_contact_method = ClientMessagingService.contact_methods(client).keys.first`

caitbot: I'm going to clarify the requirement with Yvonne.

app/jobs/send_still_needs_help_emails_job.rb
caitbot: Agreeeed, it will be easier to talk about why I did it this way over Tuple. But to try and summarize -- we want to grab  (clients who have ALL of their returns in the statuses "intake_ready" or "intake_in_progress") OR (clients with tax returns with the statuses "intake_ready" or "intake_in_progress" and have some tax returns with the status "file_not_filing")

But it is only ok to have tax returns in "file_not_filing" if you also have tax returns in "intake_ready" or "intake_in_progress". 

Which is why I do "grab all clients with SOME tax returns in either of these statuses" then "filter to clients with ALL tax returns in the acceptable statuses."

There might be a clever active record way of doing this but I couldn't think of it off the top of my head or find anything on the internet.

app/jobs/send_still_needs_help_emails_job.rb
caitbot: Per convo with Yvonne, we will just send to all opted-in methods. We may need a distinct sms body to avoid char limits.

spec/jobs/send_still_needs_help_emails_job_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        context 'and they are opted-in to text messaging' do
+          before do
+            allow(ClientMessagingService).to receive(:contact_methods).and_return({ sms_phone_number: "+14155551212", email: "example@example.com" })
+          end
+
+          it 'sends them a text message' do
+            described_class.perform_now
+
+            expect(ClientMessagingService).to have_received(:send_system_email).twice
```
caitbot, line 51: [AL writing] Consider using `.with(array_including(uu_client))` and `.with(array_including(another_uu_client))` if you want to prove that it really passes in both clients.

spec/jobs/send_still_needs_help_emails_job_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+          end
+        end
+
+        context 'and they have some tax returns in other statuses' do
+          let!(:info_requested_tr) { create(:tax_return, client: another_uu_client, status: "review_info_requested", year: 2017) }
+
+          it 'does not include them and their tax returns in the updates' do
+            described_class.perform_now
+
+            expect(Client.where(triggered_still_needs_help_at: nil)).to include(another_uu_client)
```
caitbot, line 62: [AL writing] I'm surprised there isn't a scope for this. Not a blocker.

spec/jobs/send_still_needs_help_emails_job_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    let!(:inapplicable_client) { create(:client, vita_partner: create(:organization, name: "Definitely Not UU")) }
+    let!(:some_inapplicable_tr) { create(:tax_return, client: inapplicable_client) }
+    let!(:another_inapplicable_tr) { create(:tax_return, client: inapplicable_client, status: "intake_ready", year: 2018) }
+
+    context 'when there are clients with UU as their vita partner' do
+      let!(:urban_upbound) { create(:organization, name: "Urban Upbound") }
+      let!(:uu_client) { create(:client, intake: create(:intake), vita_partner: urban_upbound) }
+      let!(:another_uu_client) { create(:client, intake: create(:intake), vita_partner: urban_upbound) }
```
caitbot, line 18: Consider `uu_site_client` here instead of `another_uu_client`.

config/locales/en.yml
```diff
@@ -1033,6 +1033,34 @@ en:
...
+          If you still need help with your taxes and would like to be contacted in June, log into the GetYourRefund portal and let us know using this link: %{link}
+
+          If you are interested in filing sooner, you can use our free Do It Yourself (DIY) service here: https://www.getyourrefund.org/en/diy/file-yourself.
+
+          Thank you,
+
+          Your tax team at GetYourRefund
+        subject: Do you still need help with your taxes?
+      sms:
+        body: |
```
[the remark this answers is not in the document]
caitbot: Going to keep this body, per discussion with @yvonnefong

alaroia-cfa (approved):
Nice SQL! 🚢

PR #1162, opened 2021-06-10
Forward hub messages to intercom for clients with tax return statuses in file_accepted, file_mailed or file_not_filing

embarnard:


Gemfile
```diff
@@ -64,6 +64,7 @@ gem 'rails_autolink'
 gem 'ice_nine'
 gem 'business_time'
 gem 'scenic'
+gem 'intercom', '~> 4.1'
```
bytheway875, line 67: 👀  oh interesting! I didn't know there was a gem.

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -83,6 +83,8 @@ def create_incoming_email
 
       end
 
+      IntercomService.create_intercom_message_from_email(contact_record, fw_from_hub: true) if (client.tax_returns.pluck(:status).map(&:to_sym) & TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES).any?
```
bytheway875, line 86: We do this client.tax_returns.pluck(:status) thing to check for "if client has tax return in a particular status" in at least a handful of places if not more now.

I'm starting to think it makes sense to have a method thats like client.tax_return_in(status) where status can be one or an array of statuses.

Not necessary right now, but it would ultimately make our code drier and more readable to have a convenience method like that.

app/services/incoming_text_message_service.rb
```diff
@@ -51,6 +51,8 @@ def self.process(params)
         documents: documents
       )
 
+      IntercomService.create_intercom_message_from_sms(contact_record, fw_from_hub: true) if (client.tax_returns.pluck(:status).map(&:to_sym) & TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES).any?
```
bytheway875, line 54: or even `client.forward_messages_to_intercom?` convenience method just to ensure the two incoming messages implementations stay in sync.

app/services/intercom_service.rb
```diff
@@ -0,0 +1,113 @@
+require 'intercom'
+
+class IntercomService
+  def self.create_intercom_message_from_email(incoming_email, fw_from_hub: false)
```
bytheway875, line 4: It seems like there is a lot of "consistent state" for this object (like the email and client, etc) that get used amongst all of these class methods. To me, it seems like creating an instance of the intercom service in the controller and calling the send method on an instance would be prudent because it would set this class up in a way that could share instance variables without passing around parameters to methods. However, calling a class method off of a service is also an established pattern, so this also makes sense. Just something to think about in terms of class design I guess.

app/controllers/mailgun_webhooks_controller.rb
embarnard: Yeah definitely! I was searching through to code to see if we already had a method for this bc I feel like we do this so much

app/services/incoming_text_message_service.rb
embarnard: Oh yeah it makes sense to link them together

app/services/intercom_service.rb
```diff
@@ -0,0 +1,113 @@
...
+    email = incoming_email.client&.intake&.email_address || incoming_email.sender
```
bytheway875, line 5: since we have "incoming_email" which is an email object, I'd suggest changing variable "email" to email_address to make it more clear the distinction between the two variables.

Also, why attempt to look at incoming email.client.intake.email instead of just grabbing from sender? Will those emails ever be different? And if they are different, do we care?

app/services/intercom_service.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    if contact_id_from_email.present? && most_recent_conversation(contact_id).present?
+      reply_to_existing_intercom_thread(contact_id, body)
+    else
+      create_new_intercom_thread(contact_id, body)
+      send_handoff_email(incoming_email.client) if fw_from_hub
+    end
+  end
+
+  def self.create_intercom_message_from_sms(incoming_sms, fw_from_hub: false)
```
bytheway875, line 18: In what cases are we or would we pass fw_from_hub as false? Also, does fw_from_hub being true only affect the sending of the handoff email in the case that it is necessary? in that case, may make sense to change the option to something like `handoff` or `inform_of_handoff`

embarnard: Yeah I definitely struggled to name this one well but I like `inform_of_handoff`! I was looking at future scenarios where we might want to create intercom messages that don't come from the hub and thought that having this option would make that easier but maybe that was too pre-emptive of me

app/services/intercom_service.rb
embarnard: True! I think we should just grab it from the sender, not sure why we did it this way in the first place

app/services/intercom_service.rb
bytheway875: oh i see! I think if we're never executing the code with fw_from_hub as false in the codebase currently, we can probably remove keyword arg altogether for now and just allow it to inform of handoff any time the "else" scenario runs, and then when the case comes up that we sometimes _dont_ want to send the handoff, we can implement this switch.

PR #1168, opened 2021-06-11
Nil protector for empty notes

bytheway875:
The nil guard was removed on the notes page because NotesPresenter service object always responds with an object and should never be nil. 

HOWEVER, notes#create re-renders the index when there is a failure (because the note creation is on the index), but render doesn't load the template variables for the rendered template, so the page was erroring out.

I felt that the appropriate level of effort at this point in time was just to add back the nil check, and create a ticket for refactoring. Reasoning: The change exposed a bug we did not know we had, but happens rarely enough that it hadn't been reported to is. Thus, it is important to fix in a more thorough manner but not at this point in time.

### DIFF
--- a/app/views/hub/notes/index.html.erb
+++ b/app/views/hub/notes/index.html.erb
@@ -10,7 +10,7 @@
   <div class="slab slab--padded">
     <div class="client-container">
       <ul class="day-list">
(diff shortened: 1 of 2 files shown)


alaroia-cfa (approved):
Thanks for being so thoughtful about how to handle this in terms of fast rollout of a medium quality fix and slow rollout of a high quality fix. Approved!

PR #1169, opened 2021-06-14
Refactor client primary partials to allow for rearrangement, style sections based on latest mock

bytheway875:

### DIFF
--- a/app/views/hub/ctc_clients/new.html.erb
+++ b/app/views/hub/ctc_clients/new.html.erb
@@ -8,12 +8,15 @@
           <h1 class="form-card__title">
             <%= @title %>
           </h1>
-          <%= f.cfa_select(:service_type, t("hub.clients.fields.service_type"), [["Drop off", "drop_off"]], options: { selected: "drop_off" } ) %>
+          <%= f.hidden_field :service_type, value: "drop_off" %>
           <%= f.cfa_select(:signature_method, t("hub.clients.fields.signature_method"), signature_methods_for_select, options: { selected: "online" }) %>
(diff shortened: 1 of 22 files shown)


app/views/hub/ctc_clients/new.html.erb
```diff
@@ -8,20 +8,22 @@
           <h1 class="form-card__title">
             <%= @title %>
           </h1>
-          <%= f.cfa_select(:service_type, t("hub.clients.fields.service_type"), [["Drop off", "drop_off"]], options: { selected: "drop_off" } ) %>
+          <%= f.hidden_field :service_type, value: "drop_off" %>
```
bytheway875, line 11: Move service_type to a hidden field since the value is always constant -- no reason for a dropdown.

app/views/shared/_client_primary_info_fields.html.erb
```diff
@@ -1,84 +1,51 @@
...
-  </div>
-  <%= f.cfa_input_field(:preferred_name, t("hub.clients.fields.preferred_name")) %>
+    <%= f.cfa_input_field(:preferred_name, t("hub.clients.fields.preferred_name")) %>
 
-  <% if is_dropoff %>
-    <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options, selected: current_user.timezone) %>
-  <% else %>
-    <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options) %>
-  <% end %>
+    <% if is_ctc %>
```
bytheway875, line 10: If the form continues to diverge, I think we might want to think about a different pattern. However, I'm planning on asking design if we can swap out the regular drop_off creation with a hidden field too since there are very low chances that the person coming into a physical office will use a different time zone than the user creating their account.

PR #1170, opened 2021-06-14
Add dependents to CTC Valet 178452857

bytheway875:
Screen Shot 2021-06-14 at 10 19 52 AM
Screen Shot 2021-06-14 at 10 20 36 AM

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -45,7 +45,7 @@ class Client < ApplicationRecord
 
   belongs_to :vita_partner, optional: true
   has_many :documents, dependent: :destroy
-  has_one :intake, dependent: :destroy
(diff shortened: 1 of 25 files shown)


app/models/client.rb
```diff
@@ -45,7 +45,7 @@ class Client < ApplicationRecord
 
   belongs_to :vita_partner, optional: true
   has_many :documents, dependent: :destroy
-  has_one :intake, dependent: :destroy
+  has_one :intake, inverse_of: :client, dependent: :destroy
```
bytheway875, line 48: @alaroia-cfa I learned what inverse_of does: When you are creating an object _through_ another object (setting dependents_attributes on an intake while creating a client), if you use inverse_of on the relationship it will allow you to create the children objects through the parent. If not, you can't do it.

PR #1185, opened 2021-06-16
Remove unused backfills and rake tasks

caitbot:


bytheway875 (commented):
Woo hoo removed code!

PR #1201, opened 2021-06-28
Remove stimulus triage code

bytheway875:
- StimulusTriage was not used this season, and product says we will not need it moving forward. 
- Routes were not accessible this season (except by mistake) and we do not allow direct linking into the flow, so we should not need to maintain redirects for stimulus routes.
- Will create a followup PR to remove tables and columns on intake.


app/models/intake.rb
```diff
@@ -207,7 +207,6 @@ class Intake < ApplicationRecord
   belongs_to :client, inverse_of: :intake, optional: true
   has_many :tax_returns, through: :client
   belongs_to :vita_partner, optional: true
-  belongs_to :triage_source, optional: true, polymorphic: true
```
kayline, line 210: Just to quick check, Intakes can't have any other kind of triage source?

kayline (commented):
Looks clean! I just had one question about whether there might potentially be other types of triage that were not a Stimulus triage. If not, this is good to go.

app/models/intake.rb
bytheway875: There are no other triage sources -- seems like the original intention was that there would be since it was set up to be polymorphic, but we don't use triage_source in our current implementation of triaging, either.

PR #1205, opened 2021-06-28
Clean MIME type headers to avoid InvalidMimeType exceptions

alaroia-cfa:
This will avoid some meaningless annoying traces sent to Sentry.

It also will reduce the raw frequency of HTTP 500 responses, which is nice.

InvalidMimeType seems vaguely correlated with sad "machine stack overflow in critical region" messages.

I personally think that our "machine stack overflow in critical region" errors relate to some kind of Rails bug (or a bug in a gem? or in our code???) showing our HTTP 500 page, so decreasing the frequency of exceptions is a way to help.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,4 +1,5 @@
 require_relative 'boot'
+require_relative "../lib/middleware/cleanup_mime_type_headers"
 
 require "rails"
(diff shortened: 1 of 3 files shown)


jenny-heath (approved):
we pair reviewed

PR #1208, opened 2021-06-28
Add CTC & full capacity notice to home, question/welcome and full service pages

embarnard:


embarnard:
Screen Shot 2021-06-28 at 6 41 15 PM
Screen Shot 2021-06-28 at 6 40 33 PM
Screen Shot 2021-06-28 at 6 41 01 PM

config/locales/en.yml
```diff
@@ -1588,6 +1588,7 @@ en:
         security:
           body: In accordance with national IRS guidelines, all Personally Identifiable Information (PII) provided by taxpayers to volunteers must be protected. We keep your information safe and secure.
           header: We're committed to keeping your information secure.
+        ctc_full_capacity_html: "Our tax preparers are at capacity and unable to complete full returns right now, but our Child Tax Credit service and DIY options are still available. Click <strong><a rel='noopener' target='_blank' href='https://www.getctc.org/en#claim-credit'>here</a></strong> if you need help claiming the Child Tax Credit. Click <strong><a rel='noopener' target='_blank' href='https://www.getyourrefund.org/en/diy/file-yourself'>here</a></strong> if you want to file your own return using free software."
```
bytheway875, line 1591: I guess this is more of a product question, but is our CTC service really "available"?

bytheway875:
When showing this at-capacity message, we say that we are no longer to complete returns but we're leaving the continue button there. Will this allow people to continue filling out a full intake? Should we hide the continue button?

embarnard:
> When showing this at-capacity message, we say that we are no longer to complete returns but we're leaving the continue button there. Will this allow people to continue filling out a full intake? Should we hide the continue button?

@bytheway875 yeah that's a good point! I'll raise it to Yvonne

PR #1218, opened 2021-06-30
Xml builder: Build Manifest file to XSD Schema

bytheway875:


app/models/efile_submission.rb
```diff
@@ -13,13 +13,19 @@
...
   has_many :efile_submission_transitions, class_name: "EfileSubmissionTransition", autosave: false, dependent: :destroy
 
   include Statesman::Adapters::ActiveRecordQueries[
     transition_class: EfileSubmissionTransition,
     initial_state: EfileSubmissionStateMachine.initial_state,
   ]
 
+  def irs_submission_id
+    Rails.logger.warn "Submission id overflow warning: modify irs_submission_id logic to prevent possible non-unique values" if id.digits.length > 11
+    ("0%012d%7s" % [id, intake.primary_last_name.downcase.first(7)]).gsub(" ", "x")
```
alaroia-cfa, line 26: non-ascii might be a problem

> string.encode("UTF-8", :invalid => :replace, :undef => :replace, :replace => "")

alaroia-cfa: also consider `has_one :client` and `client.intake.primary_last_name`

app/lib/submission_builder/manifest.rb
```diff
@@ -0,0 +1,28 @@
...
+          xml.SubmissionId @submission.irs_submission_id
+          xml.EFIN EnvironmentCredentials.dig(:irs, :efin)
+          xml.GovernmentCd "IRS"
+          xml.FederalSubmissionTypeCd "1040"
+          xml.TIN @intake.primary_ssn
+        }
+      end
+
+      xsd = Nokogiri::XML::Schema(File.open(SCHEMA_FILE))
+      xml = Nokogiri::XML(document.to_xml)
```
alaroia-cfa, line 24: In the long run, this will be the same across all `Buildable`s. Opportunity for refactor?

alaroia-cfa (approved):
yay

PR #1225, opened 2021-07-01
Intake: make a searchable_data tsvector column instead of regenerating tsvector every search

tgrathwell:
Currently every time someone does a search of the intake records, pg_search
generates a big query that generates a ton of `tsvector` columns just-in-time
to evaluate them against a `tsquery`

This is tremendously wasteful and all our search queries take at least 4 seconds
as a result, when the DB is under no load. We currently have 300k intakes, this
will only increase as we get more.

This approach adds a `searchable_data` column to `intakes` that attempts to store
all the searchable things as a `tsvector`. To keep `searchable_data` up to date,
we periodically call `rake search:refresh` which regenerates the `searchable_data`
for every row that has been marked with `needs_to_flush_searchable_data_set_at`

There's a lot of alternative approaches, we could:
* Update the column every time a row is modified, bypassing the need for a timestamp / cron thing
* Update the column with a database trigger
* Upgrade to Postgres 12+ and keep the column up to date by making it 'GENERATED STORED'

but I think the advantage of this approach is it may make it a little easier to
regenerate the whole search index if we change what columns are searchable or how

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -101,6 +101,7 @@
 #  needs_help_2018                                      :integer          default(0), not null
 #  needs_help_2019                                      :integer          default(0), not null
 #  needs_help_2020                                      :integer          default(0), not null
+#  needs_to_flush_searchable_data_set_at                :datetime
 #  no_eligibility_checks_apply                          :integer          default(0), not null
 #  no_ssn                                               :integer          default(0), not null
 #  other_income_types                                   :string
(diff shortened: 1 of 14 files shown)


bytheway875:
Nice nice. This is cool. There's likely more we'll need to do to optimize other areas of ClientSortable, because I think we've just been adding to that indiscriminately and probably have a couple more complex/show things happening there. Also, would you be down to take the team through how you approached this at CGC next week/ sometime in the future?

PR #1226, opened 2021-07-01
Remove Rails 6.1 column and table for cleanliness & doc upload correctness

alaroia-cfa:


db/migrate/20210701150705_undo_rails_61_schema_changes.rb
```diff
@@ -0,0 +1,9 @@
+class UndoRails61SchemaChanges < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :active_storage_blobs, :service_name if column_exists?(:active_storage_blobs, :service_name)
+    drop_table :active_storage_variant_records if table_exists?(:active_storage_variant_records)
```
alaroia-cfa, line 4: The reason I bother with `if table_exists?` is that the commit _before_ this has `schema.rb` excluding those columns, so it could result in weird situations. Prod and demo will both have the column & table.

alaroia-cfa (commented):
@bytheway875 approves

PR #1227, opened 2021-07-01
Add CTC valet id types

kayline:
Add photo ID and taxpayer ID types to the CTC valet form. Display the selected types on the client profile page if the client's intake is CTC.
- At least one of each type must be selected.

 #178699474

### DIFF
--- a/app/forms/hub/create_ctc_client_form.rb
+++ b/app/forms/hub/create_ctc_client_form.rb
@@ -28,6 +28,10 @@ module Hub
                        :with_incarcerated_navigator,
                        :with_limited_english_navigator,
                        :with_unhoused_navigator,
+                       :with_drivers_license_photo_id,
+                       :with_passport_photo_id,
+                       :with_other_state_photo_id,
+                       :with_vita_approved_photo_id,
                        :bank_account_number,
                        :bank_routing_number,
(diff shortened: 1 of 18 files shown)


app/forms/hub/create_ctc_client_form.rb
```diff
@@ -148,5 +158,45 @@ def tax_return_attributes
...
+
+    def selected_social_security?
+      with_social_security_taxpayer_id == "1"
+    end
+
+    def selected_itin?
+      with_itin_taxpayer_id == "1"
+    end
+
+    def at_least_one_photo_id_type_selected
```
bytheway875, line 190: Can we leverage PHOTO_ID_TYPES and TAXPAYER_ID_TYPES to create this list of acceptable photo id types? Instead of different predicate methods for each id type, you could create one method that employs :send like your methods on Intake.

```
def selected_photo_id_type
  has_photo_id = false
  while has_photo_id == false
    Intake::PHOTO_ID_TYPES.each do |type|
       has_photo_id = self.send(type[:field_name] == 1
    end
  end
  errors.add(...) unless has_photo_id
end
```

kayline: Good call, will do!

PR #1233, opened 2021-07-02
Add ReturnHeader XML Builder class

bytheway875:


app/lib/submission_builder/formatting_methods.rb
```diff
@@ -0,0 +1,28 @@
+module SubmissionBuilder
```
bytheway875, line 1: TODO: Add tests for these formatters.

bytheway875 (commented):
Pair reviewed + paired on w/Catie.

PR #1234, opened 2021-07-02
CTC Valet form: collect SSN for dependents

tgrathwell:
Change validation messages on client dependents sub-form to show
alongside errored fields rather than all at the top

Change client show page to show dependent information a little
more like the future mockup shows it's gonna look like

Dependent SSNs will show up on the 'edit client' page, but
confirmation is only required if you change the SSN from
its persisted value

Genericize the Hub::Clients::SsnItinsController to become
Hub::Clients::SecretsController, which can more easily be
extended for more secrets (like IP PIN)


embarnard:
@tgrathwell Looks good! The only thing is that when we edit the client the “re-enter ssn/itin” field for the existing dependents is blank and they are not hiding everything bu the last four digits of the ssn like we do for primary
Screen Shot 2021-07-06 at 2 55 52 PM

embarnard:
which would mean that folks could see the ssn without the access log recording that they saw it if they decided to edit the client

tgrathwell:
That's a good point but i think it's also true right now for every kind of secret we have...

the only reason it shows last 4 SSN is because that's all we collect for primary/spouse on non-CTC clients. we don't have anything special that will try to hide things on edit, seems like it would take a bunch of code

embarnard:
> That's a good point but i think it's also true right now for every kind of secret we have...
> 
> the only reason it shows last 4 SSN is because that's all we collect for primary/spouse on non-CTC clients. we don't have anything special that will try to hide things on edit, seems like it would take a bunch of code

oh interesting... but that last 4 SSN is on the edit of the CTC client?

tgrathwell:
yeah when you add a full SSN for a primary/spouse right now it also populates the last4 so you can "log in" as them

so when you go to edit it's allowing you to edit the last4 because the form doesn't branch on edit yet to show you the primary/spouse full SSN, that will be in https://www.pivotaltracker.com/story/show/178338699

it's in kind of a weird state right now!

PR #1235, opened 2021-07-05
Flow Explorer for GYR

tgrathwell:


Gemfile.lock
```diff
@@ -592,6 +592,7 @@ DEPENDENCIES
   lograge
   mailgun-ruby
   memory_profiler
+  mini_magick
```
tgrathwell, line 595: https://github.com/janko/image_processing is better I think - AL

app/controllers/flows_controller.rb
```diff
@@ -0,0 +1,147 @@
...
+    ctc: { emoji: "👶", name: "CTC Flow" },
+  }
+
+  def index
+    @page_title = 'GetYourRefund Flows'
+    @flow_config = FLOW_CONFIG
+  end
+
+  def show
+    type = params[:id].to_sym
```
tgrathwell, line 16: You could `.map(&:to_s)` to validate before the `.to_sym`, but it's not the end of the world this way. - AL

app/controllers/flows_controller.rb
```diff
@@ -0,0 +1,147 @@
+class FlowsController < ApplicationController
+  protect_from_forgery except: :show
```
tgrathwell, line 2: You should be able to delete this line if you fetch the page with the authenticity token (aka CSRF token). The jQuery fetch might solve this for you.

PR #1236, opened 2021-07-06
Fix person.svg to not have cut off shoulders

caitbot:
Implemented as part of design-eng pairing


bytheway875:
Can we run this svg through svgomg to compress it? https://jakearchibald.github.io/svgomg/

PR #1239, opened 2021-07-06
Persist bank account fields to BankAccount object for CTC intakes

bytheway875:


caitbot (approved):
reviewed as pair, yessss

PR #1244, opened 2021-07-07
Add legal consent #178788519

kayline:
Add legal consent page to CTC intake flow


app/controllers/ctc/questions/consent_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      layout "intake"
+
+      def update
+        super
+      end
+
+      private
+
+      def prev_path
```
bytheway875, line 14: prev_path and next_path only need to be manually set if there is something "funky" with the flow from CtcQuestions. For example, the contact preferences questions "jump around" (you only go to one or the other depending on an option you choose). Since you will always come from the previously defined question to this page and then to the next question, the "super" value for these methods should already redirect appropriately.

app/lib/ctc_question_navigation.rb
```diff
@@ -7,6 +7,8 @@ class CtcQuestionNavigation
     Ctc::Questions::ContactPreferenceController,
     Ctc::Questions::CellPhoneNumberController,
     Ctc::Questions::EmailAddressController,
+    # VerifyIdentityController goes here
```
bytheway875, line 10: You can create a placeholder VerificationController here with the form_class defined as NullForm here, if you'd like. There won't be any naming interference because i haven't created the controller yet -- im still working on modifying the backend logic.

app/views/ctc/questions/consent/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <%= f.cfa_input_field(:primary_first_name, t("views.ctc.questions.legal_consent.first_name"), classes: ["form-width--long"]) %>
+      <%= f.cfa_input_field(:primary_middle_initial, t("views.ctc.questions.legal_consent.middle_initial"), classes: ["form-width--short"]) %>
+      <%= f.cfa_input_field(:primary_last_name, t("views.ctc.questions.legal_consent.last_name"), classes: ["form-width--long"]) %>
+      <%= f.vita_min_date_text_fields(
+            :primary_birth_date,
+            t("hub.clients.show.date_of_birth"),
+            ) %>
+      <%= f.cfa_input_field(:primary_ssn, t("views.ctc.questions.legal_consent.ssn"), classes: ["form-width--long"]) %>
+      <%= f.cfa_input_field(:primary_ssn_confirmation, t("views.ctc.questions.legal_consent.ssn_confirmation"), classes: ["form-width--long"]) %>
+      <%= f.cfa_input_field(:sms_phone_number, t("views.ctc.questions.legal_consent.sms_phone_number"), classes: ["form-width--long"]) %>
```
bytheway875, line 26: based on the form values, this should probably be phone_number instead of sms_phone_number. sms_phone_number is gathered in the CellPhoneNumber form.

app/lib/ctc_question_navigation.rb
caitbot: Discussed on slack and decided to leave this comment and let the contact pages route directly to the consent page. Shannon will add her Verify Identity work inbetween when it is ready

app/views/ctc/questions/consent/edit.html.erb
caitbot: goooood catch

PR #1245, opened 2021-07-07
Backend refactor for verification codes

bytheway875:
Refactor the services for verification codes to decouple from "logging in" (mostly by changing location of some methods + renaming some model files.

I did not change the table_names because that seemed like it could cause runtime errors on release -- I'll do a follow up so that we can release that alone during an "off hours" release potentially.


app/services/email_verification_code_service.rb
```diff
@@ -0,0 +1,69 @@
+class EmailVerificationCodeService
```
bytheway875, line 1: This is a port from ClientLoginsService to create distinct services for Email and TextMessage codes. It allows for passing a validation_type, which alters the message and the logic for how it checks to see whether you are "eligible" for a code.

There is a lot of commonality between the text and email logic, but i decided to leave the refactor in this state due to time constraints and to prevent divergence from the original implementation too much.

app/services/email_verification_code_service.rb
```diff
@@ -0,0 +1,69 @@
...
+  # sends an email synchronously using deliver_now so that we can store the mailgun ID,
+  # so best to send this from a background job.
+  def send_verification_code
+    verification_code, access_token = generate_verification_code
+    mailer_response = VerificationCodeMailer.with(
+      to: @email_address,
+      verification_code: verification_code,
+      locale: @locale,
+      verification_type: @verification_type
+    ).with_code.deliver_now
```
bytheway875, line 38: The biggest change between the previous implementation and this implementation is that I'm using deliver_now for the email so that I can get the message ID that we use for detecting mailgun webhooks and update the status of the emails. Also, Since we always call this service from a background job, using deliver_later for sending the messages was queuing this job, and then queuing another job to send the email. Seemed like we could make that more efficient by sending the email in the job itself.

app/services/email_verification_code_service.rb
```diff
@@ -0,0 +1,69 @@
+class EmailVerificationCodeService
+  VERIFICATION_TYPES = [:ctc_intake, :gyr_login]
+  def initialize(email_address: , locale: :en, visitor_id: , client_id: nil, verification_type: :gyr_login)
```
caitbot, line 3: remove default??

app/services/email_verification_code_service.rb
```diff
@@ -0,0 +1,69 @@
...
+    can_send_code? ? send_verification_code : send_no_match_message
+  end
+
+  private
+
+  def can_send_code?
+    case @verification_type
+    when :ctc_intake
+      true
+    when :gyr_login
```
caitbot, line 23: Move this into the controller for wherever this is handled. Shannon understands what this means.

app/services/text_message_verification_code_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    can_send_code? ? create_code : send_no_match_message
+  end
+
+  private
+
+  def can_send_code?
+    case @verification_type
+    when :ctc_intake
+      true
+    when :gyr_login
```
caitbot, line 23: same here

app/services/text_message_verification_code_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      twilio_sid: twilio_response&.sid
+    )
+  end
+
+  def generate_verification_code
+    existing_tokens = TextMessageAccessToken.where(sms_phone_number: @phone_number, token_type: "verification_code")
+    existing_tokens.order(created_at: :asc).limit(existing_tokens.count - 4).delete_all if existing_tokens.count > 4
+    DatadogApi.increment("client_logins.verification_codes.text_message.created")
+    raw_verification_code, hashed_verification_code = VerificationCodeService.generate(@phone_number)
+    [raw_verification_code, TextMessageAccessToken.create!(
```
caitbot, line 48: maybe move the logic to limit to 5 objects into the model itself? but not blocking.

spec/services/text_message_verification_code_service_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      it "creates a VerificationTextMessage, sends an email, and creates an EmailAccessToken object" do
+        described_class.request_code(**params)
+        text_body = "Your 6-digit GetCTC verification code is: 123456. This code will expire after two days."
+
+        expect(TwilioService).to have_received(:send_text_message).with(a_hash_including(
+                                                                        to: phone_number,
+                                                                        body: text_body
+                                                                    ))
+        # expect(email.body.encoded).to include "Your 6-digit GetCTC verification code is: 123456"
```
caitbot, line 51: remove reference to email stuff.

caitbot (changes requested):
Refactor the no match stuff into the client logins stuff.

app/services/client_login_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      phone_numbers = TextMessageAccessToken.lookup(raw_token).pluck(:sms_phone_number)
+      phone_intake_matches = accessible_intakes.where(sms_phone_number: phone_numbers)
+      intake_matches = email_intake_matches.or(spouse_email_intake_matches).or(phone_intake_matches)
+
+      # Client.by_raw_login_token supports login links generated from mid-Jan through early March 2021.
+      Client.where(intake: intake_matches).or(Client.by_raw_login_token(raw_token))
+    end
+
+    def handle_email_request(email_address:, visitor_id:, locale: :en)
+      if can_login_by_email_verification?(email_address)
```
alaroia-cfa, line 18: Can this be in a bg job, so that timing doesn't reveal if the email exists with us?

Then the `perform_later` here can be `.perform_now` too I guess :D

app/services/client_login_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      if can_login_by_sms_verification?(phone_number)
+        RequestVerificationCodeTextMessageJob.perform_later(
+          phone_number: phone_number,
+          locale: locale,
+          visitor_id: visitor_id,
+          service_type: :gyr
+        )
+      else
+        home_url = Rails.application.routes.url_helpers.root_url(locale: locale)
+        TwilioService.send_text_message(
```
alaroia-cfa, line 43: SB you pointed out wisely that this is synchronous in the web app now; could be job-ified

app/services/email_verification_code_service.rb
```diff
@@ -0,0 +1,52 @@
+class EmailVerificationCodeService
+  SERVICE_TYPES = [:ctc, :gyr]
+  def initialize(email_address: , locale: :en, visitor_id: , client_id: nil, service_type: :gyr)
```
alaroia-cfa, line 3: I personally dislike defaults :D

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -92,12 +92,15 @@ def create_incoming_email
   end
 
   def update_outgoing_email_status
-    email_to_update = OutgoingEmail.find_by(message_id: params.dig("event-data", "message", "headers", "message-id"))
+    message_id = params.dig("event-data", "message", "headers", "message-id")
+    email_to_update = OutgoingEmail.find_by(message_id: message_id)
```
alaroia-cfa, line 96: In the super cool future, we could encode _which_ model we expect to match on, before we communicate to Mailgun. That way this can be 100% deterministic. Completely not a big deal, just a dream.

alaroia-cfa: Alternatively, one day, we could make a table called `MailgunStatus` and store the message ID in it, and then only update that one. Then we'd need to JOIN onto it when we are showing an OutgoingEmail's status. Useful? Too useless to be worth the bother? I leave that question unanswered.

caitbot (approved):
reviewed as pair

PR #1254, opened 2021-07-10
Add routing number, account number, and bank confirmation page

bytheway875:
Screen Shot 2021-07-09 at 7 57 57 PM
Screen Shot 2021-07-09 at 7 58 11 PM
Screen Shot 2021-07-09 at 7 58 38 PM


app/assets/stylesheets/templates/_dependents-index.scss
```diff
@@ -1,25 +1,13 @@
 .dependents-index {
-  .dependents-list {
```
bytheway875, line 2: .dependents-list was repurposed as "review-box" and moved

app/assets/stylesheets/templates/_dependents-index.scss
```diff
@@ -1,25 +1,13 @@
...
-    padding: 2.5rem 2rem;
-    border: 2px solid black;
-    border-radius: 1rem;
-
-    @media screen and (min-width: $tablet-up) {
-      max-width: 57rem;
-    }
-  }
   .dependent {
+    display: flex;
```
bytheway875, line 3: there must have been a regression on this page at some point -- i think maybe the image svg became bigger when we did a style update some time back and the image icon was sitting on top of the person's name. This update prevents that from happening.

app/forms/ctc/direct_deposit_form.rb
```diff
@@ -9,5 +9,12 @@ class DirectDepositForm < QuestionsForm
     def save
       @intake.update(bank_account_attributes: attributes_for(:bank_account))
     end
+    
+    def self.existing_attributes(intake)
+      return super unless intake.bank_account.present?
+
+      # bank_name is encrypted, but we want it to be editable for clients
```
tgrathwell, line 16: I don't think this is because the attr is encrypted, but because it is on a sub-model; the form is dealing only with the Intake right now so it won't read the existing value of `bank_name` (or `account_type`) unless we copy it over onto the form explicitly.

There should be some more general way to do this (maybe `fields_for :bank_account` in the form) but I haven't quite got it to work in experimenting so maybe this is fine for now (with a more accurate comment)

app/views/ctc/questions/account_number/edit.html.erb
```diff
@@ -0,0 +1,19 @@
+<% @main_question = t("views.ctc.questions.account_number.title") %>
+
+<% content_for :page_title, @main_question %>
+
+<% content_for :form_card do %>
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long", autocomplete: "off" } do |f| %>
+    <h1 class="h2"><%= @main_question %></h1>
+
+    <div class="form-card__content">
+      <%= f.cfa_input_field(:account_number, t("views.ctc.questions.account_number.account_number"), type: :password) %>
```
tgrathwell, line 10: Was this asked to be a password field? I feel like most of the time I enter these details on websites it's in cleartext to start with; it's tough for me to enter long digits of numbers without seeing them as I type

app/forms/ctc/direct_deposit_form.rb
tgrathwell: if we keep the explicit merge of `bank_account` fields you may want to add `account_type` to it so it is still populated when someone goes back to edit

tgrathwell (approved):

app/forms/ctc/direct_deposit_form.rb
bytheway875: ah, so its a combo -- because of the way "existing attributes" works, it only slices off "real" attributes, so the encrypted ones always need to be explicitly set (wondering whether account_name even needs to be encrypted though if we want it to be editable), but also because they aren't on intake, all of the attrs need to be set. I had a prev implementation that overwrote from_intake that dealt with account_type properly, but yes -- i think this will not preserve account type properly.

tgrathwell: yah I didn't realize in the last comment that `intake.attributes` isn't gonna give any of the encrypted stuff :)

---

For controllers/forms that pull things off a submodel like `bank_account` I'm trying to imagine if there's a better implementation for `FormAttributes.existing_attributes` that handles submodel attributes by default

Like whenever we say a form has `attributes_for :bank_account, ...` the base `existing_attributes` method could see if `:bank_account` was a real association on the main model and try to read attributes from it onto the base form

then we could also replace calls like `model.attributes` with something like

`scoped_attributes[model.table_name].map { |attr| model.send(attr) }`

(I wish there was a more succinct way to write that) to handle anything with fancy getters such as encrypted fields

---

though having dived deeper into `FormAttributes` now I'm a little creeped out by how it attempts to flatten a bunch of things that potentially live on subobjects into one flat list of form attributes which then get unflattened on `save`. but maybe that's just something I need to adapt to

PR #1256, opened 2021-07-12
Return builder 178735377

bytheway875:
- Builds IRS1040 document to fit the needs of the AdvCtc revenue procedure.
 - Refactors the builder classes to inherit from a Base class instead of including a Buildable concern


app/lib/submission_builder/base.rb
```diff
@@ -0,0 +1,38 @@
+module SubmissionBuilder
```
bytheway875, line 1: Shared "Base" class for Submission XML builder classes. some of this logic used to be located in the "Buildable" concern, but as more parts got built out, it became apparent that the builder classes shared a lot of core logic.

app/lib/submission_builder/base.rb
```diff
@@ -0,0 +1,38 @@
...
+      { "xmlns:efil" => "http://www.irs.gov/efile", "xmlns" => "http://www.irs.gov/efile" }
+    end
+
+    def document
+      raise "SubmissionBuilder classes must implement their own document method that returns a Nokogiri::XML::Document object"
+    end
+
+    def build
+      errors = []
+      if @validate
```
bytheway875, line 26: The switch for validating against the schema allows us the option to validate each "built" piece against its own schema, but will allow us to bypass individual validation on the final "Return" object so that we can validate only at the top level and get all relevant xml errors in one response.

PR #1259, opened 2021-07-13
Service to standardize addresses on intakes

caitbot:
See [the story on Pivotal](https://www.pivotaltracker.com/story/show/178749757) for context (:


app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+  def self.standardize_address_on_intake(intake)
+    build_standardized_address(intake.street_address, intake.city, intake.state, intake.zip_code)
+  end
+
+  private
+
+  def self.build_standardized_address(street_address, city, state, zip_code)
+    usps_address_xml = get_usps_address_xml(street_address, city, state, zip_code)
+    {
+      street_address: usps_address_xml.xpath("//Address/Address2").text,
```
caitbot, line 16: @bytheway875 is this object format/set of keys acceptable given your use case? happy to change things around to be more easily usable by the service consuming it.

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    request_address_xml = Nokogiri::XML(builder.to_xml)
+    usps_request_address = "https://secure.shippingapis.com/ShippingAPI.dll?API=Verify&XML=#{request_address_xml}"
+
+    response = Net::HTTP.get_response(URI(usps_request_address))
+    response_xml = Nokogiri::XML(response.body)
+
+    if response_xml.xpath("//Error").present?
+      Rails.logger.error "Error returned from USPS Address API: #{response_xml.xpath("//Error/Description").text}"
```
caitbot, line 47: I'm tossing this to the Rails.logger but is there a better way to log these errors. My main concern with this path is noisy-ness in Slack or Sentry. I couldn't find much in the way of custom error logging around the codebase.

Maybe I should toss in a Datadog increment? 🤷🏻‍♀️

spec/services/standardize_address_service_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+describe StandardizeAddressService do
+  describe '#standardize_address_on_intake' do
+    before do
+      stub_request(:get, /.*secure\.shippingapis\.com.*/)
+        .to_return(status: 200, body: usps_api_response_body, headers: {})
+    end
+
+    let(:usps_api_response_body) { file_fixture("usps_address_validation_body.xml").read }
+
+    let(:intake) { create(:ctc_intake, street_address: "43 Vicksburg Street Unit B", city: city, state: state, zip_code: zip_code) }
```
caitbot, line 12: lol this is my current home address but I'm moving out and its kinda nice that it will forever be in a test case <3

spec/services/standardize_address_service_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      let(:usps_api_response_body) { file_fixture("usps_address_validation_error_body.xml").read }
+
+      let(:city) { nil }
+      let(:state) { nil }
+      let(:zip_code) { "44092" }
+
+      it 'logs an error and returns the description' do
+        expect(Rails.logger).to receive(:error).with("Error returned from USPS Address API: Address Not Found.  ")
```
caitbot, line 60: I don't love the extra white space at the end but that is actually what USPS sends 🤦🏻

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+      street_address: usps_address_xml.xpath("//Address/Address2").text,
+      city: usps_address_xml.xpath("//Address/City").text,
+      state: usps_address_xml.xpath("//Address/State").text,
+      zip_code: usps_address_xml.xpath("//Address/Zip5").text,
+      errors: usps_address_xml.xpath("//Error/Description").text
+    }
+  end
+
+  def self.get_usps_address_xml(street_address, city, state, zip_code)
+    usps_userid = "251CODEF8049" # EnvironmentCredentials.dig(:usps, :userid)
```
caitbot, line 25: lol I need to change this -- does anyone know how to stick this in our prod EnvCreds and want to pair with me on it?

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+      street_address: usps_address_xml.xpath("//Address/Address2").text,
+      city: usps_address_xml.xpath("//Address/City").text,
+      state: usps_address_xml.xpath("//Address/State").text,
+      zip_code: usps_address_xml.xpath("//Address/Zip5").text,
+      errors: usps_address_xml.xpath("//Error/Description").text
+    }
+  end
+
+  def self.get_usps_address_xml(street_address, city, state, zip_code)
+    usps_userid = EnvironmentCredentials.dig(:usps, :userid)
```
caitbot, line 25: I need to put this into EnvCreds.. does anyone know how and want to pair with me on it?

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
+require 'net/http'
+require 'uri'
+
+# PLEASE do NOT use this service to standardize addresses in bulk (do NOT use in backfills)
+# as that violates the terms of agreement with USPS and will result in access being revoked.
+class StandardizeAddressService
+  def self.standardize_address_on_intake(intake)
+    build_standardized_address(intake.street_address, intake.city, intake.state, intake.zip_code)
```
bytheway875, line 8: the new mocks had us add a "street_address2" -- a separate line to add apartment number. do we need to pass that in as well?

app/services/standardize_address_service.rb
bytheway875: i know how! and can pair with you on it.

app/services/standardize_address_service.rb
bytheway875: My PREFERENCE would be to create instances of this class and be able to get the response as method calls, but that's just a preference and not backed up by any specific need. However, it would allow us to normalize when there are error responses maybe?

app/services/standardize_address_service.rb
bytheway875: i think this actually will bypass Sentry and slack because its going to log errors into the rails logger rather than "raise" an error (which would become a Sentry ticket). It would be interesting to have it do both so that we can actively be notified of how often it happens and then deal with it later.

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def self.get_usps_address_xml(street_address, city, state, zip_code)
+    usps_userid = EnvironmentCredentials.dig(:usps, :userid)
+    builder = Nokogiri::XML::Builder.new do |xml|
+      xml.AddressValidateRequest(:USERID => usps_userid) {
+        xml.Revision 1
+        xml.Address(:ID => 0) {
+          xml.Address1
+          xml.Address2 street_address
```
caitbot, line 31: add street_address2 from intake to this

app/services/standardize_address_service.rb
caitbot: initializer - takes an intake and creates @street_address

@result = build_standardized_street_address

methods that are street_address that allow you to access the @result[street address]

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,52 @@
...
+          xml.City city
+          xml.State state
+          xml.Zip5 zip_code
+          xml.Zip4
+        }
+      }
+    end
+
+    request_address_xml = Nokogiri::XML(builder.to_xml)
+    usps_request_address = "https://secure.shippingapis.com/ShippingAPI.dll?API=Verify&XML=#{request_address_xml}"
```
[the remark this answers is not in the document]
caitbot: hahahahah yeah its kinda weird, they don't want it as params in the body, just in the url

app/services/standardize_address_service.rb
caitbot: oooh good to know, ok, do we have any rails logger error thresholds for sentry? I think thats a thing you can set up...

app/services/standardize_address_service.rb
```diff
@@ -0,0 +1,81 @@
+require 'net/http'
+require 'uri'
+
+# PLEASE do NOT use this service to standardize addresses in bulk (do NOT use in backfills)
+# as that violates the terms of agreement with USPS and will result in access being revoked.
+class StandardizeAddressService
+  def initialize(intake)
+    @street_address = [intake.street_address, intake.street_address2].join(" ")
```
bytheway875, line 8: i might prefix these with @_street_address etc to make it clear that these are the INPUTS (since the default behavior of many rails classes is to create attr_accessors that allow you to get the value of an instance variable as a method of the same name, and here you'll be subverting that expectation since the instance variables are the input and the methods are the output)

bytheway875:
Heres how you add to rails credentials:

```
$ EDITOR=vim rails credentials:edit --environment production
```

replacing production with the environment and the editor with wherever you want to edit it. 

Then since you want the key at :usps, :user_id, you'd just have :usps at the top level and user_id indented. 

You don't have to add it/shouldnt add it for test evironment since we'll just stub it in tests when necessary.

PR #1263, opened 2021-07-14
Test script for trying to get the efile xml to work at least one dang time

tgrathwell:
So far the best error I can ever get is this

```
<ErrorMessageTxt>The namespace declarations in the root element of the return ('Return' element) must be as follows:The default namespace shall be set to "http://www.irs.gov/efile" (xmlns= "http://www.irs.gov/efile").The namespace prefix "efile" shall be bound to the namespace "http://www.irs.gov" (xmlns:efile="http://www.irs.gov/efile").</ErrorMessageTxt>
```

but it's not clear that they're even reading submission.xml

### DIFF
--- a/app/services/gyr_efiler_service.rb
+++ b/app/services/gyr_efiler_service.rb
@@ -1,0 +1,79 @@
+class GyrEfilerService
+  CURRENT_VERSION = '1105349922f7be08cb3a2d8f1d1438eb72b9f5d2'
+
+  def self.run_efiler_command(*args)
+    raise StandardError.new("Cannot be used from the test environment") if Rails.env.test?
+
+    Dir.mktmpdir do |working_directory|
+      FileUtils.mkdir_p(File.join(working_directory, "output", "log"))
+      ensure_config_dir_prepared
+
+      # TODO: If the process blocks for >10 min, terminate it.
+      # TODO: Send process stdout to logs.
+      # TODO: Send output/logs/ to logs after process terminates.
+      classes_zip_path = Dir.glob(Rails.root.join("vendor", "gyr_efiler", "gyr-efiler-classes-#{CURRENT_VERSION}.zip"))[0]
+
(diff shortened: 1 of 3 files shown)


efile_submission_playground/do_stuff.rb
```diff
@@ -0,0 +1,68 @@
...
+        system("sed -i '' 's/REPLACEME/#{@submission_id}/g' manifest/manifest.xml")
+        system("zip -0 -r #{@submission_id}.zip manifest xml")
+        system("mv #{@submission_id}.zip #{__dir__}")
+      end
+    end
+  end
+
+  def submit
+    Dir.mktmpdir do |temp_dir|
+      Dir.chdir(temp_dir) do
```
alaroia-cfa, line 29: Dir.chdir isn't thread-safe, so let me see if I can revise this

app/services/gyr_efiler_service.rb
```diff
@@ -33,8 +39,31 @@ def self.get_single_file_from_zip(zipfile_path)
   private
 
   def self.ensure_config_dir_prepared
-    return if File.exists?(Rails.root.join("vendor", "gyr_efiler", "config"))
-    config_zip_path = Dir.glob(Rails.root.join("vendor", "gyr_efiler", "*classes*.zip"))[0]
+    config_dir = Rails.root.join("vendor", "gyr_efiler", "gyr_efiler_config")
+    return if File.exists?(File.join(config_dir, '.ready'))
+
+    config_zip_path = Dir.glob(Rails.root.join("vendor", "gyr_efiler", "gyr-efiler-config-#{CURRENT_VERSION}.zip"))[0]
+    system("unzip -o #{config_zip_path} -d #{Rails.root.join("vendor", "gyr_efiler")}")
```
tgrathwell, line 46: use ruby unzip or at least the array form but i guess ok

PR #1264, opened 2021-07-14
Restore active storage routing locally

bytheway875:
The catchall was catching too much! And blocking us from viewing active storage uploads in development.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -274,5 +274,7 @@ Rails.application.routes.draw do
     end
   end
 
-  get '*unmatched_route', to: 'public_pages#page_not_found'
+  get '*unmatched_route', to: 'public_pages#page_not_found', constraints: lambda { |req|
+    req.path.exclude? 'rails/active_storage'
+  }
 end


config/routes.rb
```diff
@@ -274,5 +274,7 @@ def scoped_navigation_routes(context, navigation, as_redirects: false)
     end
   end
 
-  get '*unmatched_route', to: 'public_pages#page_not_found'
+  get '*unmatched_route', to: 'public_pages#page_not_found', constraints: lambda { |req|
+    req.path.exclude? 'rails/active_storage'
```
alaroia-cfa, line 278: I'd like to see this constraint only apply in dev. either way lgtm

alaroia-cfa (approved):
yowza. Good eye

config/routes.rb
alaroia-cfa:
```
!Rails.env.development? || req.path.exclude? 'rails/active_storage'
```

I think that would accomplish my dream, in which the constraint is always true in non-dev.

alaroia-cfa:
```
if Rails.env.development?
  # ActiveStorage is considered unmatched; we only use this kind of active storage routes in dev
  req.path.exclude? 'rails/active_storage'
else
  true
end
```

might be clearer

alaroia-cfa:
```
if Rails.env.development?
  # In development, we the ActiveStorage .url route is "unmatched" and handled outside the app,
  # so we must skip rendering the 404 page.
  req.path.exclude? 'rails/active_storage'
else
  true
end
```

bytheway875:
because the active_storage urls are hit in prod and causing errors for the representation stuff (and creating a ton of errors in DD), I'm going to go ahead and just unconstrain the active storage routes in all enviornments to "restore" service? And then we can roll back around to doing it on a per environment basis should we decide that's preferred.

PR #1266, opened 2021-07-14
Creates class for EIP 1 calculation

miketlawrence:
We need to be able to calculate expected Economic Impact Payment (EIP) 1 value for filers. This class calculates expected EIP1 amount based on dependents and filing status.


miketlawrence:
@bytheway875

PR #1268, opened 2021-07-14
Creates class for EIP 3 calculation

miketlawrence:
We need to be able to calculate expected Economic Impact Payment (EIP) 3 value for filers. This class calculates expected EIP3 amount based on dependents and filing status.


miketlawrence:
@bytheway875

PR #1269, opened 2021-07-14
Create pages for filed-2020 and filed-2019 status checks on CTC intake #178806362

caitbot:


app/forms/ctc/filed_2019_form.rb
```diff
@@ -0,0 +1,15 @@
...
+    set_attributes_for :intake, :filed2019
+
+    def save
+      @intake.update(transform_attributes(attributes_for(:intake)))
+    end
+
+    private
+
+    def transform_attributes(attributes)
+      { filed_2019: attributes[:filed2019] }
```
caitbot, line 12: Rails magic makes the field on the form called `filed2019` (it doesn't consider 2019 to be its own word) but we didn't want to call it that on the `intake` - hence the transformation

spec/factories/addresses.rb
```diff
@@ -1,3 +1,18 @@
+# == Schema Information
```
caitbot, line 1: Hahah I guess this annotation was never created on the original addresses branch and snuck in here.. leaving it unless there are objections

app/forms/ctc/filed_2019_form.rb
alaroia-cfa: Similar to `Questions::HomebuyerCreditController`, Travis & I suggest using

```
   private
   
    def method_name
      "filed_2019"
    end
```

Check out `app/views/layouts/yes_no_question.html.erb` to see how `"#{method_name}"` is used. Given that the `method_name` system is part of our own framework in `vita-min`, it's nice if we can keep using it and stay consistent.

Not a blocker. Happy to show you what we are talking about over Tuple; not 100% sure that this got conveyed clearly.

app/forms/ctc/filed_2020_form.rb
```diff
@@ -0,0 +1,15 @@
...
+    set_attributes_for :intake, :filed2020
+
+    def save
+      @intake.update(transform_attributes(attributes_for(:intake)))
+    end
+
+    private
+
+    def transform_attributes(attributes)
+      { filed_2020: attributes[:filed2020] }
```
alaroia-cfa, line 12: same as the other comment about `method_name`

alaroia-cfa (approved):

app/forms/ctc/filed_2019_form.rb
caitbot: Oh very cool!! Thank you for pointing that out! I will update that!

PR #1274, opened 2021-07-15
Add /home to CTC intake

caitbot:
[The story!](https://www.pivotaltracker.com/story/show/178832375)


bytheway875 (approved):

bytheway875:
For another time: might make sense to have one field on intake called :offboarding_reason, and any time we offboard them and say they aren't qualified we persist that. However, its not necessary/asked for right now, and this fulfills the reporting need by sending the data to mixpanel.

PR #1284, opened 2021-07-15
Add life situations and ssn#178818538

coltborg:
[CTC intake flow: /life-situations2020 #178818538](https://www.pivotaltracker.com/story/show/178818538)


app/controllers/ctc/questions/life_situations2020_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      include AnonymousIntakeConcern
+
+      layout "intake"
+
+      private
+
+      def illustration_path; end
+
+      def next_path
+        @form.cannot_claim_me_as_a_dependent != "yes" ? questions_use_gyr_path : super
```
coltborg, line 14: This is how I achieve:
> - if "No one can claim me as a dependent" is NOT checked, route to /use-GYR
> - If "No one can claim me as a dependent" IS checked, route to /filing-status

app/views/ctc/questions/life_situations2020/edit.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <h1 class="h2"><%= @main_question %></h1>
+
+  <p class="spacing-below-15"><%= t("views.ctc.questions.life_situations_2020.p1") %></p>
+
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
+    <div class="form-card__stacked-checkboxes spacing-above-0 spacing-below-15">
+      <%= f.cfa_checkbox(:cannot_claim_me_as_a_dependent, t("views.ctc.questions.life_situations_2020.cannot_claim_me_as_a_dependent"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:member_of_the_armed_forces, t("views.ctc.questions.life_situations_2020.member_of_the_armed_forces"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+    </div>
+    <div class="reveal" style="max-width: 47.5rem;">
```
coltborg, line 15: I set this `max-width` to a magic number, but It looked weird for the reveal container not be the same size of the checkboxes on larger screens.

app/controllers/ctc/questions/life_situations2020_controller.rb
bytheway875: i would strongly recommend against using 2020 in the URL because then we have to go back and update urls when we start filing for 2021 instead of just being able to update the contents of the page. Some of the pages we added, the whole thing is super core to being from that year so its a grey area but this one doesnt feel that way to me -- /questions/life-situations should be sufficient.

bytheway875: ah, i guess there is the life situations 2019 controller. thats a real bummer.

app/forms/ctc/life_situations2020_form.rb
```diff
@@ -0,0 +1,9 @@
+module Ctc
+  class LifeSituations2020Form < QuestionsForm
```
bytheway875, line 2: its strange that the controller and form have different casing for the date. It feels like that should be consistent. either s_2020.rb or 2020.rb in all cases

app/models/intake/ctc_intake.rb
```diff
@@ -100,6 +101,7 @@
 #  locale                                               :string
 #  made_estimated_tax_payments                          :integer          default(0), not null
 #  married                                              :integer          default(0), not null
+#  member_of_the_armed_forces                           :integer          default("unfilled"), not null
```
bytheway875, line 104: are we going to ask a similar question about spouse and primary? if so, we might want to prefix with primary_member_of_the_armed_forces so that we can later do spouse_member_of_the_armed_forces

app/views/ctc/questions/life_situations2020/edit.html.erb
bytheway875: agreed. the varying widths of different components really frustrates me. I wish we could set it up so that everything by default takes up the size of its container, and then make things smaller if necessary.

bytheway875 (approved):
Nothing really blocking, but id definittely take a look at the whole primary_ spouse_ thing because that may become hairy later.

app/controllers/ctc/questions/life_situations2020_controller.rb
bytheway875: Oh! one thing. you can write a little helper method on the form if you want to if you think the != yes makes things kind of hard to read 

```
def cannot_be_claimed_as_dependent?
  cannot_be_claimed_as_dependent == "no"
end
```

app/views/ctc/questions/life_situations2020/edit.html.erb
coltborg: Yes and agreed, I would love to migrate to a "full width" of whatever container approach.

app/forms/ctc/life_situations2020_form.rb
coltborg: I tried to make all 2020 and 2019 be in the format of `life_situations_2020_controller.rb`, but it fails the tests because the auto loader is rails doesn't add an `_` at the start of the number.

I found a [post on stack overflow that points this out](https://stackoverflow.com/questions/9909596/how-do-i-name-a-rails-ruby-file-whose-class-name-has-numbers#answer-12719566), I'll move both 2020 and 2019 to not have an `_` before the number.

PR #1286, opened 2021-07-16
Add spec and fix content for service_name failure

bytheway875:
Missing tests burned me once, burned me twice.


kayline (approved):
Looks good, hopefully will save our future selves!

PR #1293, opened 2021-07-17
Access e-file schemas from Google Drive instead

alaroia-cfa:
Downloading things from S3 is too hard. I copied the file into Google Drive.


caitbot (approved):
I could access that link! yay! 🙌🏻

PR #1294, opened 2021-07-17
EIP and CTC Dependent Eligibility Validator

miketlawrence:
We need a way to determine if dependents are eligible for EIP1, EIP2, EIP3, or CTC payments. These class encapsulate the eligibility logic based on dependent age, and relationship to the filer.

### DIFF
--- a/app/lib/ctc_dependent_eligibility.rb
+++ b/app/lib/ctc_dependent_eligibility.rb
@@ -1,0 +1,26 @@
+class CtcDependentEligibility
+  CUT_OFF_BIRTHDATE = Date.new(2004, 1, 1)
+
+  def initialize(birthdate:, relationship:)
+    @birthdate = birthdate
+    @relationship = relationship
+  end
+
+  def self.eligible?(*args)
+    new(*args).eligible?
+  end
+
+  def eligible?
+    qualifying_relationship? && eligible_age?
+  end
+
+  private
+
+  def qualifying_relationship?
+    ChildRelationshipQualifier.qualifies?(relationship: @relationship)
+  end
+
+  def eligible_age?
+    CUT_OFF_BIRTHDATE <= @birthdate
+  end
+end
--- a/app/lib/eip_three_dependent_eligibility.rb
+++ b/app/lib/eip_three_dependent_eligibility.rb
@@ -1,0 +1,39 @@
+class EipThreeDependentEligibility
+  UNDER_NINETEEN_CUT_OFF_BIRTHDATE = Date.new(2002, 1, 1)
+  UNDER_TWENTY_FOUR_CUT_OFF_BIRTHDATE = Date.new(1997, 1, 1)
+
+  def initialize(dependent_birthdate:, disabled:, student:, filer_birthdate:)
+    @dependent_birthdate = dependent_birthdate
+    @disabled = disabled
+    @student = student
+    @filer_birthdate = filer_birthdate
+  end
+
+  def self.eligible?(*args)
+    new(*args).eligible?
+  end
+
+  def eligible?
+    return true if @disabled
+
+    return true if under_nineteen?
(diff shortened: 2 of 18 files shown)


app/lib/ctc_dependent_eligibility.rb
```diff
@@ -0,0 +1,9 @@
+require "date"
+
+class CtcDependentEligibility
+  CUT_OFF_BIRTHDATE = Date.new(2004, 1, 1)
+
+  def self.eligible?(birthdate:)
```
bytheway875, line 6: i believe we should add the qualifying relationship logic to this guy as well.

miketlawrence: Yeah for sure — still working but on the class that governs that

app/lib/eip_three_dependent_eligibility.rb
```diff
@@ -0,0 +1,32 @@
...
+class EipThreeDependentEligibility
+  UNDER_NINETEEN_CUT_OFF_BIRTHDATE = Date.new(2002, 1, 1)
+  UNDER_TWENTY_FOUR_CUT_OFF_BIRTHDATE = Date.new(1997, 1, 1)
+
+  def self.eligible?(dependent_birthdate:, disabled:, student:, filer_birthdate:)
+    return true if disabled
+
+    return true if self.under_nineteen?(birthdate: dependent_birthdate)
+
+    return false if !student
```
bytheway875, line 12: i wonder if this would be easier to read (or at least fall more in line with how the IRS defines the rule) as a one liner return false unless student && self.under_twenty_four(bd) && younger_than_filer?(bd, filer_bd)

bytheway875: I also love keyword args for the "public" methods so that you can get a glimpse at what the data is supposed to mean when implemented outside the class, but for "private" methods, just passing the param directly might be sufficient since the parameter name itself is going to be super indicative of what its purpose is.

bytheway875: Also, when passing the same parameters around in a class, sometimes it can be easier to create an instance of the class so that you can grab the shared instance variables. We have a little patten for adding a class method that wraps an instance method so that the 'public api' of the class is kept simple. Here that might look something like this:

```
class EipThreeDependentEligibility
  UNDER_NINETEEN_CUT_OFF_BIRTHDATE = Date.new(2002, 1, 1)
  UNDER_TWENTY_FOUR_CUT_OFF_BIRTHDATE = Date.new(1997, 1, 1)

  def initialize(dependent_birthdate:, disabled:, student:, filer_birthdate:)
    @dependent_birthdate = dependent_birthdate
    @disabled = disabled
    @student = student
    @filer_birthdate = filer_birthdate
  end 
  
   def self.eligible?(*args)
       new(*args).eligible?
   end

  private

   def under_19?
      ...
  end 
 
   def younger_than_filer?
      @filer_birthdate < @dependent_birthdate
  end

   def eligible?
      ... 
   end
end
```

Certainly not a required change as its something we don't have a MANDATED pattern for, but I think it makes it easier to read as well as test, because then you only have to test the public method (self.eligible?)

bytheway875: (technically, since all class methods are public, our testing guidelines would suggest that each class method has its own tests which don't really make sense here because they're only used in service of the self.eligible? implementation

app/lib/dependent_relationship_qualifier.rb
```diff
@@ -0,0 +1,19 @@
+class DependentRelationshipQualifier
```
bytheway875, line 1: let's call this something like "ChildRelationshipQualifier", since the other relationships can still be valid dependents but they aren't "children"

PR #1295, opened 2021-07-19
QuestionsController.to_path_helper uses url_for instead of path_helper_string

tgrathwell:
Previously, path_helper_string was used to compute the appropriate
url helper from a controller class (something like questions_ip_pin_path in CTC
or irs_letter_questions_path in GYR)

url_for can be used for this purpose and only requires a controller name
and action (which in our form navigation framework is always :edit)


bengolder (approved):
Looks good!

PR #1296, opened 2021-07-19
Efiler updates wip

bytheway875:
Some of these values are placeholders to ensure we can deal with an "Accepted" response, and will be updated with more accurate information soon.

### DIFF
--- a/app/jobs/build_submission_bundle_job.rb
+++ b/app/jobs/build_submission_bundle_job.rb
@@ -1,8 +1,9 @@
 class BuildSubmissionBundleJob < ApplicationJob
-  def perform(submission)
+  def perform(submission_id)
+    submission = EfileSubmission.includes(:intake, :dependents, :client, :address, :tax_return).find(submission_id)
     address_creation = submission.generate_irs_address
     unless address_creation.valid?
-      submission.transition_to!(:bundle_failure, error_message: address_creation.errors)
+      submission.transition_to!(:failed, error_message: address_creation.errors)
       return
     end
 
@@ -10,7 +11,7 @@ class BuildSubmissionBundleJob < ApplicationJob
     if response.valid?
       submission.transition_to!(:queued)
     else
-      submission.transition_to!(:bundle_failure, error_message: response.errors)
+      submission.transition_to!(:failed, error_message: response.errors)
     end
   end
 end
--- a/app/lib/submission_builder/return_header1040.rb
+++ b/app/lib/submission_builder/return_header1040.rb
@@ -32,10 +32,14 @@ module SubmissionBuilder
           }
           xml.SelfSelectPINGrp {
             xml.PrimaryBirthDt date_type(intake.primary_birth_date)
-            if intake.spouse_birth_date.present?
-              xml.SpouseBirthDt date_type(intake.spouse_birth_date)
+            if tax_return.filing_jointly?
+              xml.SpouseBirthDt date_type(intake.spouse_birth_date) if tax_return.filing_jointly?
             end
+            xml.PrimaryPriorYearAGIAmt 0
+            xml.SpousePriorYearAGIAmt 0 if tax_return.filing_jointly?
           }
+          xml.IdentityProtectionPIN intake.primary_ip_pin if intake.primary_ip_pin.present?
+          xml.SpouseIdentityProtectionPIN intake.spouse_ip_pin if tax_return.filing_jointly? && intake.spouse_ip_pin.present?
           xml.PINTypeCd "Self-Select On-Line"
           xml.JuratDisclosureCd "Online Self Select PIN"
           xml.PrimaryPINEnteredByCd "Taxpayer"
--- a/app/lib/submission_builder/return_header1040.rb
+++ b/app/lib/submission_builder/return_header1040.rb
@@ -28,7 +28,13 @@ module SubmissionBuilder
           xml.SoftwareId "11111111"
           xml.OriginatorGrp {
             xml.EFIN EnvironmentCredentials.dig(:irs, :efin)
-            xml.OriginatorTypeCd "OnlineFiler"
+            xml.OriginatorTypeCd "ERO"
+          }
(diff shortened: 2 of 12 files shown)


app/models/efile_submission.rb
```diff
@@ -47,14 +47,24 @@ def imperfect_return_resubmission?
   end
 
   def generate_irs_address
-    address_service = StandardizeAddressService.new(intake)
-    if address_service.valid?
+    if Rails.env.development?
```
bytheway875, line 50: @caitbot I was doing the same call over and over again and didn't want to hit the USPS api too many times, so I stubbed this out for development. There's probably a better pattern here... like maybe a config variable that gets passed to the IRS service itself? But wanted to check in whether you thought that stubbing calls to the API in dev is a good idea.

spec/factories/intakes.rb
```diff
@@ -238,7 +238,7 @@
   trait :primary_consented do
     primary_consented_to_service_at { 2.weeks.ago }
     primary_consented_to_service { "yes" }
-    primary_consented_to_service_ip { "192.168.2.1" } # IRS approved IP address
+    primary_consented_to_service_ip { "136.62.11.145" } # IRS approved IP address
```
bytheway875, line 241: The IRS's provided the test IP address on the left exploicitly as one that would "pass". However, it comes back as a reserved IP and fails on submission. 🤔  So I swapped it out with something that doesnt make the IRS complain

spec/factories/intakes.rb
```diff
@@ -298,6 +298,13 @@
     email_address { Faker::Internet.email }
   end
 
+  trait :with_address do
```
bytheway875, line 301: @alaroia-cfa @caitbot You don't have to do the address setting during testing anymore.

spec/jobs/build_submission_bundle_job_spec.rb
```diff
@@ -15,9 +15,9 @@
       let(:address_valid?) { false }
       let(:address_errors) { "usps error your zip code is a duck" }
 
-      it "transitions the submission into :bundle_failure" do
+      it "transitions the submission into :failed" do
```
bytheway875, line 18: i decided to remove the :bundle_failure status and transition to just using :failed status, and then we can use the metadata to create distinction between the different types of failures.

app/jobs/build_submission_bundle_job.rb
```diff
@@ -1,16 +1,17 @@
 class BuildSubmissionBundleJob < ApplicationJob
-  def perform(submission)
+  def perform(submission_id)
```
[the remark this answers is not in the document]
tgrathwell: I don't know about the logging part, but my understanding is when you enqueue a job with an ActiveRecord model as a parameter these days it enqueues it by its `to_global_id` representation, which is basically just the app name, class name, and id with some slashes.

```
irb(main):002:0> Intake.last.to_global_id.to_s
  Intake Load (1.2ms)  SELECT "intakes".* FROM "intakes" ORDER BY "intakes"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> "gid://vita-min/Intake::CtcIntake/58"
```

So it's just a quick way to avoid having to `Submission.find(...)` in the job and shouldn't have those downsides.

Can you show me where any filter-worthy params might be showing up in the logs? This is what I see:

```
09:21:38 worker.1  | [Worker(host:Traviss-MBP pid:21576)] Job BuildSubmissionBundleJob [35d38cff-4244-4d20-8e6e-7374d37e8549] from DelayedJob(default) with arguments: [{"_aj_globalid"=>"gid://vita-min/EfileSubmission/1"}] (id=40) (queue=default) RUNNING
```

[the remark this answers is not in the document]
caitbot: This is Shannon: Oh interesting. I didn't know that. Ran into an issue last year where when the jobs were failing, they were logging all of the information about the failed email/text jobs. But that may be a separate issue of failed jobs.

I haven't updated my understanding of how this works in a while and was running based off of "best practice" for implementing using delayedJob from years ago -- that's cool that its changed.

Here, I also was kind of relying on the fact that we had to load the object in order to eager load the associations before passing them in to the Builder so that the db calls were grouped to the beginning.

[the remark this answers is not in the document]
tgrathwell: I remember being scared away from passing models as arguments in the past for reasons like you describe, but as far as I can tell Rails has been doing the globalid thing since Rails 4.2 which dates back to 2014. I think Ben Sheldon got me in the habit of moving beyond the past and just passing the models without worry.

I saw the eager loading part; I'm not sure it makes enough of a difference in the case of a single item, since it should only effect whether the extra queries happen immediately or 5ms later when they're needed.

caitbot (approved):
reviewed as pair

app/lib/submission_builder/return_header1040.rb
```diff
@@ -71,13 +81,36 @@ def document
...
+                end
+              }
+            }
           }
           xml.FilingSecurityInformation {
+            xml.AtSubmissionCreationGrp {
+              xml.IPAddress {
+                xml.IPv4AddressTxt intake.primary_consented_to_service_ip
+              }
+              xml.DeviceId 9162213099514827927117083645386143446039
```
caitbot, line 101: Thrilled to look into this with you if you want another pair of eyes! - AL

bytheway875: @caitbot specifically the device id? i'm going to be working on getting real data for this. this is just placeholder.

spec/factories/intakes.rb
caitbot: Consider `8.8.8.8` or `1.1.1.1` both of which are public services run respectively by Google and Cloudflare, and therefore we'd know that the IP address here isn't meaningful. (Is this your current IP address, Shannon? :D) - AL

Not a blocker obviously. We might make this change.

PR #1298, opened 2021-07-19
Update tax return status after submission state transition

caitbot:


app/lib/tax_return_status.rb
```diff
@@ -43,7 +43,7 @@ def message_templates
 
     review_ready_for_qr: 301, review_reviewing: 302, review_ready_for_call: 303, review_signature_requested: 304, review_info_requested: 305,
 
-    file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406, file_hold: 450
+    efile_needs_review: 400, file_ready_to_file: 401, file_efiled: 402, file_mailed: 403, file_rejected: 404, file_accepted: 405, file_not_filing: 406, file_hold: 450
```
caitbot, line 46: @bytheway875 this was in the story descrip. but I guess I don't understand how this number got chosen. Can you clarify?

bytheway875, line 46: ohh i think i might have messed up in the story. it should probably be file_needs_review

bytheway875: i think theres something re: organizing them by "stage" that requires that the prefix is the same for all of them.

caitbot: hmm there isn't a spec but perhaps there should be one hahahaha

bytheway875 (approved):
Woohoo! This is amazing

app/lib/tax_return_status.rb
caitbot: ahh actually I take that back - there is a spec and I had to fix it. done done done.

bytheway875: basically, the "numbers" are how the stages are organized for metabase purposes (they do queries of 100-199 for intake stage, 400-499 for file stage) and then the prefix is what organizes things into stages in the dropdowns, organized in integer order.

So if we leave as efile_needs_review, it will show up nested under its own stage in the dropdown instead of along with the existing file stage statuses.

PR #1310, opened 2021-07-21
Capture source and save on intake

kayline:
Captures anything after the slash in CTC domain as `:source` param. `ApplicationController` already has a `set_source` method to capture this param and set it in the session. `FirstQuestionConcern` and `ConsentForm` now extract that value from session and save it on the current intake. Based on my understanding, existing Mixpanel behavior will include the `intake_source` attribute on appropriate Mixpanel events.

Feel free to reach out for real-time discussion if needed/desired.

[#178918905]

### DIFF
--- a/app/lib/ctc_question_navigation.rb
+++ b/app/lib/ctc_question_navigation.rb
@@ -5,7 +5,7 @@ class CtcQuestionNavigation
     # Basic info
     Ctc::Questions::OverviewController,
     Ctc::Questions::IncomeController,
-    Ctc::Questions::ConsentController,
+    Ctc::Questions::ConsentController,            # At this point we create the intake, client, and tax return
     Ctc::Questions::ContactPreferenceController,
     Ctc::Questions::CellPhoneNumberController,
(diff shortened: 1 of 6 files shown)


app/lib/ctc_question_navigation.rb
```diff
@@ -5,7 +5,7 @@ class CtcQuestionNavigation
     # Basic info
     Ctc::Questions::OverviewController,
     Ctc::Questions::IncomeController,
-    Ctc::Questions::ConsentController,
+    Ctc::Questions::ConsentController,            # At this point we create the intake, client, and tax return
```
alaroia-cfa, line 8: Helpful comment!

alaroia-cfa (approved):
This code is great, and @tgrathwell and I will merge it, in the interest of expediency!

We noticed something curious -- even though you're only calling `CtcIntake.new`, you're actually getting an intake persisted into the database. We took a look at why, and noticed that `ApplicationController#set_current_step` always saves the current_intake.

That's a TIL for us, so figured we'd share.

PR #1314, opened 2021-07-21
Add some spanish translations for ctc views

kayline:
[#178953171]


bengolder (approved):
Looks like YAML keys! Once we get them all in, we can better rely on automated testing to verify that we have all the keys we need.
Thanks!

PR #1316, opened 2021-07-22
Methods for determining qualified dependents

caitbot:


app/models/dependent.rb
```diff
@@ -101,8 +148,49 @@ def age_at_end_of_year(tax_year)
   end
 
   def eligible_for_child_tax_credit?(tax_year)
-    # add additional eligibility logic
-    age_at_end_of_year(tax_year) < 17
+    is_qualifying_child? && age_at_end_of_year(tax_year) < 17
```
jenny-heath, line 151: haha it seems optimistic and probably unnecessary to pass in tax year given that we only know the policy for one tax year. if in future we have other tax years to compute the CTC for, the rules might be different. not a big deal code-wise, but it might be simpler to remove the argument

jenny-heath: oh and now that i've read further, `is_qualifying_child?` is already hard-coding 2020, so it probably won't make sense to use that method in conjunction with an age cutoff that applies to any other year

app/models/dependent.rb
```diff
@@ -101,8 +148,49 @@ def age_at_end_of_year(tax_year)
...
+      (provided_over_half_own_support_no? && no_ssn_atin_no? && filed_joint_return_no?) &&
+      lived_with_less_than_six_months_no? && (can_be_claimed_by_other_no? or claim_regardless_yes?)
+  end
+
+  def possibly_qualifying_child?
+    qualifying_child_relationship? &&
+      ((!full_time_student_no? && age_at_end_of_year(2020) < 24) || !permanently_totally_disabled_no? || age_at_end_of_year(2020) < 19) &&
+      (!provided_over_half_own_support_yes? && !no_ssn_atin_yes? && !filed_joint_return_yes?) &&
+      !lived_with_less_than_six_months_yes? && (!can_be_claimed_by_other_yes? || !claim_regardless_no?)
+  end
```
jenny-heath, line 174: i wonder if there's a way for us to reuse some thing between this method and `qualifying_child?` - let's chat

app/models/dependent.rb
```diff
@@ -101,8 +148,49 @@ def age_at_end_of_year(tax_year)
...
+
+  def possibly_qualifying_relative?
+    !qualifying_child? &&
+      (((qualifying_child_relationship? && age_at_end_of_year(2020) >= 19 && !full_time_student_yes?) ||
+        (qualifying_child_relationship? && age_at_end_of_year(2020) > 24) ||
+        (qualifying_child_relationship? && ((!full_time_student_no? && age_at_end_of_year(2020) < 24) || !permanently_totally_disabled_no? || (age_at_end_of_year(2020) < 19)) && !filed_joint_return_no?) ||
+        qualifying_relative_relationship?) &&
+        !provided_over_half_own_support_yes? && !no_ssn_atin_yes? &&
+      !meets_misc_qualifying_relative_requirements_no?)
```
jenny-heath, line 193: same with these two methods - it gets pretty hard to read so i'm thinking we could extract some of the sets of conditions. that and i think a "possibly qualifying relative" might be `!possibly_qualifying_child?`? but that depends on the usage and i'm getting confused haha

app/models/dependent.rb
```diff
@@ -101,8 +148,54 @@ def age_at_end_of_year(tax_year)
...
+  def qualifying_child?
+    qualifying_child_relationship? &&
+      meets_qc_age_condition? &&
+      meets_qc_misc_conditions? &&
+      meets_qc_residence_condition? &&
+      meets_qc_claimant_condition?
+  end
+
+  def meets_qc_age_condition?
+    (full_time_student_yes? && age_at_end_of_year(2020) < 24) || permanently_totally_disabled_yes? || age_at_end_of_year(2020) < 19
```
bytheway875, line 171: does this make sense to be tax_year? it worries me that we're going to have to change this once we start filing 2021 returns.

PR #1321, opened 2021-07-22
Add special situation checkboxes to dependents info

caitbot:
For [this story](https://www.pivotaltracker.com/story/show/178819593)

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1362,6 +1362,9 @@ en:
             middle_initial: "Dependent’s middle initial"
             last_name: "Dependent’s legal last name"
             relationship_to_you: "What is their relationship to you?"
+            situations: "Select any situation that was true:"
+            full_time_student: "Dependent is a full-time student"
+            permanently_totally_disabled: "Dependent is permanently and totally disabled"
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -1353,6 +1353,9 @@ en:
             middle_initial: "Dependent’s middle initial"
             last_name: "Dependent’s legal last name"
             relationship_to_you: "What is their relationship to you?"
+            situations: "Select any situation that was true:"
+            full_time_student: "Dependent is a full-time student"
+            permanently_totally_disabled: "Dependent is permanently and totally disabled"
```
caitbot, line 1358: haha I realize that these were likely intended to actually specify the name, but given that its the page that we collect the name on.. theres some complexity in that (I mean we could do it for editing existing dependents but then you run into an issue of inconsistency if they decide to change the name, it might be confusing so ultimately I got rid of it)

caitbot: Spoke with Anu and got the ok on this copy!

kayline (approved):
Looks good! Thanks for clarifying the intent with the names.

PR #1327, opened 2021-07-23
Update dependent eligibility in xml

bytheway875:


app/lib/submission_builder/documents/adv_ctc_irs1040.rb
```diff
@@ -14,38 +14,34 @@ def dependent_xml(xml, dependent)
         xml.DependentDetail do
           xml.DependentFirstNm dependent.first_name
           xml.DependentLastNm dependent.last_name
-          xml.DependentNameControlTxt person_name_control_type(dependent.last_name)
+          xml.DependentNameControlTxt person_name_control_type(dependent.first_name)
```
bytheway875, line 17: Em and I realized during the test case scenario that the reason for the control text is so that across multiple forms you can identify the same person even if the name is provided in a slightly different format. For that reason, if def makes sense to have the "control" be the first name.

app/lib/submission_builder/documents/adv_ctc_irs1040.rb
```diff
@@ -14,38 +14,34 @@ def dependent_xml(xml, dependent)
...
           xml.DependentSSN dependent.ssn
           xml.DependentRelationshipCd dependent.relationship&.upcase
-          xml.EligibleForChildTaxCreditInd "X" if dependent.eligible_for_child_tax_credit?(2021)
```
bytheway875, line 20: I was originally confused about the year we were supposed to be comparing to for this determination.

app/models/dependent.rb
```diff
@@ -92,24 +92,24 @@ class Dependent < ApplicationRecord
   validates :ip_pin, ip_pin: true
 
   QUALIFYING_CHILD_RELATIONSHIPS = [
```
bytheway875, line 94: The casing on these relationships will cause mismatches. The relationships get stored in the database as all lowercased strings (see relationship helper spec for what the values are). We send them all to the IRS as upper case strings.

Choosing to upcase vs downcase was arbitrary.

spec/factories/dependents.rb
```diff
@@ -56,11 +56,31 @@
     was_student { "yes" }
     disabled { "no" }
     sequence(:ssn) do |n|
-      if intake.is_ctc?
+      if intake && intake.is_ctc?
```
bytheway875, line 59: i wanted to be able to make dependents that weren't associated with an intake.

app/lib/submission_builder/documents/adv_ctc_irs1040.rb
caitbot: Yeah, using 2020 struck me as odd at first too.

app/lib/submission_builder/return_header1040.rb
```diff
@@ -148,7 +148,7 @@ def document
             xml.TotalPreparationSubmissionTs total_preparation_submission_minutes
             xml.TotActiveTimePrepSubmissionTs total_active_preparation_minutes
             # TODO: Swap out with VendorControlNum that conforms to IRS standards -- first two digits will be provided by IRS
-            xml.VendorControlNum "xsdefedlsoenajsk"
+            xml.VendorControlNum "2P00000000000001"
```
caitbot, line 151: Does this confirm to IRS standards? Can we remove the comment? Is this what our control number will be going forward or is it another stand in?

spec/factories/dependents.rb
```diff
@@ -56,11 +56,31 @@
...
+      filed_joint_return { "no" }
+      lived_with_less_than_six_months { "no" }
+      can_be_claimed_by_other { "no" }
+      claim_regardless { "yes" }
+    end
+
+    trait :qualifying_relative do
+      relationship { "parent" }
+      birth_date { Date.new(1960, 5, 2) }
+      filed_joint_return { "no" }
```
caitbot, line 82: you shouldn't actually need this filed_joint_return answer

spec/models/tax_return_spec.rb
```diff
@@ -80,6 +80,22 @@
...
+      it "raises an error" do
+        expect {
+          tax_return.qualifying_dependents
+        }.to raise_error StandardError
+      end
+    end
+
+    context "when there are some qualifying dependents and some not qualifying dependents" do
+      let(:intake) { create :intake, :with_dependents, dependent_count: 4 }
+      let(:tax_return) { create :tax_return, year: 2020, intake: intake }
```
caitbot, line 95: this is missing an `it` block

caitbot (approved):
just need to update the spec + the relationship list values that we talked about and then this is good to go, other comments aren't blocking

PR #1330, opened 2021-07-23
Update dependents column to match question asked

caitbot:


app/controllers/ctc/questions/dependents/child_lived_with_you_controller.rb
```diff
@@ -15,8 +15,7 @@ def self.model_for_show_check(current_controller)
         end
 
         def method_name
-          # TODO: the column name seems to be the opposite of the question being asked on this page
-          'lived_with_less_than_six_months'
+          'lived_with_more_than_six_months'
```
caitbot, line 18: The "yes" and "no" answers to the question need to match the column intent. And the question is "Did the dependent live with you for more than six months?"

I spoke with Anu and we definitely want the happy path to be "yes" and so we needed to flip the column name/intent.

PR #1333, opened 2021-07-23
Add dependent claims pages to CTC intake

caitbot:
This completes [this story.](https://www.pivotaltracker.com/story/show/178966147)

I haven't added to the feature spec yet because these pages are gated by prior pages that are not yet in the flow.

Once this and the [disqualifiers page](https://www.pivotaltracker.com/story/show/178965959) is in, we can update the feature spec accordingly (update: oh wow [there is a story](https://www.pivotaltracker.com/story/show/178819620) for that)

### DIFF
--- a/app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
+++ b/app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
@@ -7,6 +7,7 @@ module Ctc
 
         def self.show?(dependent)
           return false unless dependent
+
           dependent.qualifying_child_relationship? &&
             dependent.meets_qc_age_condition_2020? &&
             dependent.meets_qc_misc_conditions? &&
@@ -18,7 +19,7 @@ module Ctc
         end
 
         def method_name
-          'can_be_claimed_by_other'
+          'cant_be_claimed_by_other'
         end
 
         private
--- a/app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
+++ b/app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
@@ -1,0 +1,32 @@
+module Ctc
+  module Questions
+    module Dependents
+      class ChildCanBeClaimedByOtherController < BaseDependentController
+        include AuthenticatedCtcClientConcern
+        layout "yes_no_question"
+
+        def self.show?(dependent)
+          return false unless dependent
+          dependent.qualifying_child_relationship? &&
+            dependent.meets_qc_age_condition_2020? &&
+            dependent.meets_qc_misc_conditions? &&
+            dependent.meets_qc_residence_condition_2020?
+        end
+
+        def self.model_for_show_check(current_controller)
+          current_resource_from_params(current_controller.visitor_record, current_controller.params)
+        end
+
+        def method_name
+          'can_be_claimed_by_other'
+        end
+
+        private
+
+        def illustration_path
+          "dependents.svg"
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/ctc/questions/dependents/claim_child_anyway_controller.rb
+++ b/app/controllers/ctc/questions/dependents/claim_child_anyway_controller.rb
@@ -1,0 +1,33 @@
+module Ctc
+  module Questions
+    module Dependents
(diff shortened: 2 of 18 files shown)


app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Ctc
+  module Questions
+    module Dependents
+      class ChildCanBeClaimedByOtherController < BaseDependentController
+        include AuthenticatedCtcClientConcern
+        layout "yes_no_question"
+
+        def self.show?(dependent)
+          return false unless dependent
```
tgrathwell, line 9: Trivial: Rubocop and Asheesh would suggest a blank line after a guard like this.

Not a blocker.

app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+            dependent.meets_qc_misc_conditions? &&
+            dependent.meets_qc_residence_condition_2020?
+        end
+
+        def self.model_for_show_check(current_controller)
+          current_resource_from_params(current_controller.visitor_record, current_controller.params)
+        end
+
+        def method_name
+          'can_be_claimed_by_other'
```
tgrathwell, line 21: This seems like it might be another situation where the text of the page is opposite of the column name:

the text is "Are you the only person who can claim %{dependent_name}?", if you say "yes" then we either want `can_be_claimed_by_other` to be FALSE, or to change the name to `cant_be_claimed_by_other` (?) or `can_only_be_claimed_by_me` (!) or `can_only_be_claimed_by_client` or `can_only_be_claimed_by_filer`

This seems like a substantial issue, and it seems worth fixing the column name.

app/controllers/ctc/questions/dependents/claim_child_anyway_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        include AuthenticatedCtcClientConcern
+        layout "yes_no_question"
+
+        def self.show?(dependent)
+          return false unless dependent
+          dependent.qualifying_child_relationship? &&
+            dependent.meets_qc_age_condition_2020? &&
+            dependent.meets_qc_misc_conditions? &&
+            dependent.meets_qc_residence_condition_2020? &&
+            dependent.can_be_claimed_by_other_yes?
```
tgrathwell, line 14: If we're right about the way yes/no buttons work on the previous page, then the wrong data is stored at the time you visit this page.

app/views/ctc/questions/dependents/claim_child_anyway/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :form_question, t("views.ctc.questions.dependents.claim_child_anyway.title", dependent_name: @form.dependent.first_name) %>
+<% content_for :form_help_text do %>
+  <%= t("views.ctc.questions.dependents.claim_child_anyway.help_text", dependent_name: @form.dependent.first_name) %>
+
+  <%= render('components/molecules/reveal', title: t("views.ctc.questions.dependents.claim_child_anyway.not_sure_reveal.title")) do %>
+    <p> <%= t("views.ctc.questions.dependents.claim_child_anyway.not_sure_reveal.p") %> </p>
+    <ul class="list--bulleted">
+      <li> <%= t("views.ctc.questions.dependents.claim_child_anyway.not_sure_reveal.li1") %> </li>
```
tgrathwell, line 8: Trivial suggestion: you can use a list in the en.yaml. See e.g. `eligibility.list` around L2248 of en.yml and how it gets consumed in `app/views/public_pages/stimulus.html.erb`.

Not a blocker.

tgrathwell (changes requested):
Asheesh and I think that the text and the database/form field names don't match up. Happy to discuss over Tuple.

With a rename to the fields, this is excellent.

app/controllers/ctc/questions/dependents/child_can_be_claimed_by_other_controller.rb
caitbot: Woooops good catch, thank you!

tgrathwell (approved):
yahoo!!!

PR #1335, opened 2021-07-23
Add qualifying relative page to CTC intake

caitbot:
For [this story](https://www.pivotaltracker.com/story/show/178956464)


tgrathwell (approved):
The `show?` is complex and therefore slightly hard to review, but it also matches the story, and Asheesh and I can make sense of it.

This is great. 🚢

PR #1338, opened 2021-07-26
Fully I18n-ize the CTC feature spec

tgrathwell:
incidentally fixes a bug with some links in error messages that
weren't getting rendered in the current locale


tgrathwell (commented):
Looks great! -AL

PR #1339, opened 2021-07-26
Resubmit submissions + submission refactor + some styling

bytheway875:
Screen Shot 2021-07-26 at 11 22 26 AM
Screen Shot 2021-07-26 at 11 22 51 AM
Screen Shot 2021-07-26 at 11 23 00 AM
Screen Shot 2021-07-26 at 11 23 07 AM

### DIFF
--- a/app/controllers/hub/efile_submissions_controller.rb
+++ b/app/controllers/hub/efile_submissions_controller.rb
@@ -2,7 +2,23 @@ module Hub
   class EfileSubmissionsController < ApplicationController
     include AccessControllable
     before_action :require_sign_in
-    load_and_authorize_resource
+    authorize_resource
+    load_resource only: [:resubmit]
     layout "admin"
+
+    def index
+      @efile_submissions = EfileSubmission.most_recent_by_tax_return.page(params[:page])
+    end
+
+    def show
+      @tax_return = TaxReturn.joins(:efile_submissions).find(params[:id])
+      @efile_submission = @tax_return.efile_submissions.last
+    end
+
+    def resubmit
+      @efile_submission.transition_to!(:resubmitted, { initiated_by_id: current_user.id })
+      flash[:notice] = "Resubmission initiated."
+      redirect_to hub_efile_submission_path(id: @efile_submission.tax_return)
+    end
   end
 end
--- a/app/controllers/hub/efile_submissions_controller.rb
+++ b/app/controllers/hub/efile_submissions_controller.rb
@@ -10,6 +10,10 @@ module Hub
       @efile_submissions = EfileSubmission.most_recent_by_tax_return.page(params[:page])
     end
 
+    # a little bit unexpectedly, the "show" page actually loads the tax return will all associated submissions
+    # instead of a single submission.
+    # However, efile_submission instance variable for most recent submission is used for access control and
+    # to display information about overall status on the show page.
     def show
       @tax_return = TaxReturn.joins(:efile_submissions).find(params[:id])
       @efile_submission = @tax_return.efile_submissions.last
--- a/app/models/efile_submission.rb
+++ b/app/models/efile_submission.rb
@@ -19,7 +19,6 @@ class EfileSubmission < ApplicationRecord
   has_many :dependents, through: :intake
   has_one :address, as: :record
   has_many :efile_submission_transitions, class_name: "EfileSubmissionTransition", autosave: false, dependent: :destroy
-
   has_one_attached :submission_bundle
 
   before_validation :generate_irs_submission_id
@@ -31,6 +30,12 @@ class EfileSubmission < ApplicationRecord
     transition_class: EfileSubmissionTransition,
     initial_state: EfileSubmissionStateMachine.initial_state,
   ]
+
+  scope :most_recent_by_tax_return, lambda {
+    joins(:tax_return).where("efile_submissions.created_at = (SELECT MAX(efile_submissions.created_at) FROM efile_submissions WHERE efile_submissions.tax_return_id = tax_returns.id)")
+  }
+
+  before_create :generate_irs_submission_id
 
   def state_machine
     @state_machine ||= EfileSubmissionStateMachine.new(self, transition_class: EfileSubmissionTransition)
@@ -54,7 +59,7 @@ class EfileSubmission < ApplicationRecord
   # there is an associated "previous submission" that needs to be reported to the IRS,
   # but only if that submission has been successfully transmitted to the IRS.
   def previously_transmitted_submission
-    tax_return.efile_submissions.joins(:efile_submission_transitions).where({ efile_submission_transitions: { to_state: "transmitted" } }).last
+    tax_return.efile_submissions.where("created_at < ?", created_at).last
   end
 
   def generate_irs_address
--- a/app/models/efile_submission.rb
+++ b/app/models/efile_submission.rb
@@ -35,8 +35,6 @@ class EfileSubmission < ApplicationRecord
     joins(:tax_return).where("efile_submissions.created_at = (SELECT MAX(efile_submissions.created_at) FROM efile_submissions WHERE efile_submissions.tax_return_id = tax_returns.id)")
   }
 
-  before_create :generate_irs_submission_id
-
   def state_machine
     @state_machine ||= EfileSubmissionStateMachine.new(self, transition_class: EfileSubmissionTransition)
   end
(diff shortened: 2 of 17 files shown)


app/state_machines/efile_submission_state_machine.rb
```diff
@@ -69,4 +67,13 @@ class EfileSubmissionStateMachine
       locale: client.intake.locale
     )
   end
+
+  after_transition(to: :resubmitted) do |submission|
+    if submission.efile_submission_transitions.where(to_state: :transmitted).count.zero?
```
bytheway875, line 72: This was kind of an odd decision that I'm still not sure I'm totally on board for. 

Essentially, if we've never transmitted the submission to the IRS, there's no real reason to create a new submission and generate a new IRS id, so those I'm transitioning back to the start. That allows me to know that something that has a "previous submission" needs to include the previous submission id node when building XML for the new submission.

If this is uncomfortable, we could always create new objects even if it hasnt already been transmitted, and update the logic for determining a "previously transmitted submission"

app/views/hub/efile_submissions/show.html.erb
```diff
@@ -1,22 +1,28 @@
...
 <% content_for :card do %>
-  <div class="slab slab--not-padded spacing-above-25">
-    <h1 class="h2 spacing-below-5">
-      Submission <%= @efile_submission.id %> <span class="small">(<%= @efile_submission.current_state %>)</span>
-    </h1>
-    <p class="spacing-above-5 spacing-below-5">Client: <%= link_to @efile_submission.tax_return.client.legal_name, hub_clients_path(id: @efile_submission.tax_return.client.id) %></p>
+  <div class="slab slab--not-padded spacing-above-25 submission">
+    <div style="display: flex;">
+      <h1 class="h2 spacing-below-5">
+        <%= link_to @tax_return.client.legal_name, hub_clients_path(id: @tax_return.client.id) %> (<%= @efile_submission.current_state %>)
```
alaroia-cfa, line 6: I think it's confusing to have the client be a link within a H1/H2, but if you think it makes sense, I am fine with it. (Not a blocker since this is just me expressing confusion.)

app/views/hub/efile_submissions/show.html.erb
```diff
@@ -1,22 +1,28 @@
...
+      </h1>
+      <div style="margin-left: 10px;">
+        <%= link_to "Resubmit return", resubmit_hub_efile_submission_path(id: @efile_submission.id), method: :patch, data: { confirm: "Have you updated the necessary information for #{@tax_return.client.legal_name}?" },  class: "button button--small" if @efile_submission.can_transition_to?(:resubmitted) %>
```
alaroia-cfa, line 9: ✨

alaroia-cfa: If the submission can't be transitioned to resubmitted, presumably the reason is that it's already been resubmitted. If so, it makes sense to me we'd provide a link to the next submission's EfileSubmission page.

Not a blocker, but curious if you agree.

app/views/hub/efile_submissions/show.html.erb
```diff
@@ -1,22 +1,28 @@
...
-          <li>
-            <%= timestamp transition.created_at %>
-            <%= transition.to_state %>
-            <%= transition.metadata["error_code"] %>
-            <%= transition.metadata["error_message"] %>
-          </li>
-        <% end %>
-      </ul>
+        <ul role="list" class="logs">
+          <%= render(partial: "log", collection: efile_submission.efile_submission_transitions, as: :transition) || "No logs yet" %>
```
alaroia-cfa, line 23: ✨

lib/tasks/create_test_efile_submissions.rake
```diff
@@ -10,7 +10,7 @@ namespace :efile do
 
     queued_group = preparing_group.take(75)
     queued_group.each do |submission|
-      submission.transition_to!(:queued, seeding: true)
+      submission.transition_to!(:queued)
```
alaroia-cfa, line 13: Does this mean that these rake-baed submissions will eventually try to get e-filed on my development workstation, if I have the credentials set up properly? If so, should we make a way to skip them, or should we just have fun sending them to the IRS?

spec/controllers/hub/efile_submissions_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
+require "rails_helper"
+
+describe Hub::EfileSubmissionsController do
+  describe '#index' do
+    it_behaves_like :an_action_for_admins_only, action: :index, method: :get
```
alaroia-cfa, line 5: ✨

alaroia-cfa (approved):
This is a good PR. I left some comments. I'm happy to see this merged. I'd love your thoughts on the comments sometime, but none are truly blockers.

Here are the important comments rephrased:

I believe that it'd be better change this table to be a list of TaxReturn records. Then we could simplify some access control logic, and we could show the logs of all EfileSubmissions related to that TaxReturn. That doesn't have to be part of this pull request. I'm curious if you agree with that general direction. I think it would make it easier for e.g. Client Support.

I think we should be more explicit about storing an ID of some kind to indicate that an EfileSubmission is a predecessor and/or successor to another one. I now see that's the purpose of `previously_transmitted_submission` -- I'd like to see that become more explicit.

I think we should call `generate_irs_submission_id` in at most one ActiveRecord callback, but this PR adds a second place.

I'd like to see an index on the field we're calling MAX on (but that's not a blocker).

Cheerio!

app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -2,7 +2,23 @@ module Hub
...
+    authorize_resource
+    load_resource only: [:resubmit]
     layout "admin"
+
+    def index
+      @efile_submissions = EfileSubmission.most_recent_by_tax_return.page(params[:page])
+    end
+
+    def show
+      @tax_return = TaxReturn.joins(:efile_submissions).find(params[:id])
```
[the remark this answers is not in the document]
bytheway875: ill consider moving the show pages to a tax returns controller, but would still list things on the index in terms of the "last" efile submission for a tax return on the index most likely

app/models/efile_submission.rb
```diff
@@ -32,6 +31,12 @@ class EfileSubmission < ApplicationRecord
     initial_state: EfileSubmissionStateMachine.initial_state,
   ]
 
+  scope :most_recent_by_tax_return, lambda {
+    joins(:tax_return).where("efile_submissions.created_at = (SELECT MAX(efile_submissions.created_at) FROM efile_submissions WHERE efile_submissions.tax_return_id = tax_returns.id)")
```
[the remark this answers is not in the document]
bytheway875: oh, yeah we should be able to do MAX efile_submissions.id since we have an incrementing db

app/models/efile_submission.rb
```diff
@@ -32,6 +31,12 @@ class EfileSubmission < ApplicationRecord
...
+  }
+
+  before_create :generate_irs_submission_id
```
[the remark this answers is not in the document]
bytheway875: this is a merge conflict, ill remove the before_create

app/models/efile_submission.rb
```diff
@@ -54,7 +59,7 @@ def resubmission?
   # there is an associated "previous submission" that needs to be reported to the IRS,
   # but only if that submission has been successfully transmitted to the IRS.
   def previously_transmitted_submission
-    tax_return.efile_submissions.joins(:efile_submission_transitions).where({ efile_submission_transitions: { to_state: "transmitted" } }).last
+    tax_return.efile_submissions.where("created_at < ?", created_at).last
```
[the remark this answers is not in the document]
bytheway875: New objects are only created if they have been transmitted, so you can assume that newer efile submissions are resubmissions of transmitted submissions.

app/views/hub/efile_submissions/show.html.erb
bytheway875: i agree it's weird and that there is no particular reason it needs to be done this way. however, it was a change requested by design and i already used my energy/goodwill negotiating for other changes that were out of scope on this page.

app/views/hub/efile_submissions/show.html.erb
bytheway875: all of the submissions are on this page. thats why it displays on a tax return basis.

lib/tasks/create_test_efile_submissions.rake
bytheway875: they fail because the test data isn't set up correctly because its a bit outdated, which im currently treating as a feature so that it prevents them from building and submitting properly 😬

app/views/hub/efile_submissions/_log.html.erb
```diff
@@ -0,0 +1,21 @@
+<li role="listitem">
+  <div class="timestamp"><%= timestamp transition.created_at %></div>
+  <div class="status <%= transition.to_state %>"><%= transition.to_state %></div>
+  <div class="details">
+    <% if transition.initiated_by.present? %>
+      <div>Initiated by: <%= transition.initiated_by.name %></div>
```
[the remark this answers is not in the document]
bytheway875: we store the id, so if we need additional info or want to link to the user at some point down the line that's entirely feasible!

app/state_machines/efile_submission_state_machine.rb
bytheway875: I updated to do this, which also helps me simplify grabbing the "previous submission" of a submission 👯

PR #1340, opened 2021-07-26
update intercom chat button#178761751

coltborg:
[Update Intercom chat button #178761751](https://www.pivotaltracker.com/story/show/178761751)

# What does this PR do?

- Adds a custom intercom button to display instead of the stock intercom button
- Both English and Spanish translations
- Adds a new intercom icon


app/assets/stylesheets/components/_buttons.scss
```diff
@@ -80,6 +80,49 @@
   }
 }
 
+.button--green {
```
coltborg, line 83: I thought that `button--green` can be re-used in the future if we want more money green buttons. And that `button--intercom` below has very specific styles for the intercom button.

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -80,6 +80,49 @@
...
+  }
+
+  @media screen and (max-width: $mobile-up) {
+    font-size: $font-size-15;
+    padding: $s5 !important;
+
+    img {
+      width: $s25;
+    }
+  }
```
coltborg, line 123: Made the button a little bit smaller on small device sizes to not take up a huge amount of space.

Screen Shot 2021-07-26 at 12 25 52

Screen Shot 2021-07-26 at 12 25 45

app/views/shared/_intercom.html.erb
```diff
@@ -1,12 +1,18 @@
 <% @app_id = EnvironmentCredentials.dig(:intercom, :app_id) %>
 
+<button type="button" class="open-intercom button button--small button--green button--intercom">
```
coltborg, line 3: The `open-intercom` class does not add any styles, it enables intercom to open on click. 

This is a class because it is being used on the [tax questions public page](https://demo.getyourrefund.org/en/tax-questions), so on that page there would be two intercom buttons (aka why it's not an `id`).

alaroia-cfa (approved):
Simple and good! I can handle follow-up changes if needed.

PR #1346, opened 2021-07-27
RRC calculation should 'zero out' over payments [#179017071]

bytheway875:
- Only claim RRC on the 1040 if the client wants to AND the amount to claim is greater than 0
- If the reported credit is greater than the expected credit from our calculation for either EIP1 or EIP, treat it as zero -- people have no responsibility to report 'excess' EIP payments, and can still claim underpayment on the other check if relevant.

### DIFF
--- a/spec/models/tax_return_spec.rb
+++ b/spec/models/tax_return_spec.rb
@@ -58,6 +58,65 @@ describe TaxReturn do
     end
   end
 
+  describe "#outstanding_recovery_rebate_credit" do
+    let(:tax_return) { create :tax_return, client: client, year: 2020 }
+    let(:client) { create :client, intake: create(:ctc_intake, eip1_amount_received: eip1_amount_received, eip2_amount_received: eip2_amount_received) }
+    let(:intake) { tax_return.intake }
+
+    context "when received credit is higher than expected for eip 1, and has some outstanding for eip2" do
+      let(:eip1_amount_received) { 4000 }
+      let(:eip2_amount_received) { 1000 }
+      before do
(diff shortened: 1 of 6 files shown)


spec/models/tax_return_spec.rb
```diff
@@ -58,6 +58,65 @@
...
+
+    context "when there is outstanding credit for eip1 and eip2" do
+      let(:eip1_amount_received) { 2300 }
+      let(:eip2_amount_received) { 1100 }
+      before do
+        allow(EconomicImpactPaymentOneCalculator).to receive(:payment_due).and_return(2400)
+        allow(EconomicImpactPaymentTwoCalculator).to receive(:payment_due).and_return(1200)
+      end
+
+      it "qualifies for outstanding eip1 amount" do
```
tgrathwell, line 100: should say something like "eip1 + eip2 amounts"

spec/models/tax_return_spec.rb
```diff
@@ -58,6 +58,65 @@
...
+    context "when there is no outstanding credit for either" do
+      let(:eip1_amount_received) { 2400 }
+      let(:eip2_amount_received) { 1200 }
+
+      before do
+        allow(EconomicImpactPaymentOneCalculator).to receive(:payment_due).and_return(2400)
+        allow(EconomicImpactPaymentTwoCalculator).to receive(:payment_due).and_return(1200)
+      end
+
+      it "qualifies for outstanding eip1 amount" do
```
tgrathwell, line 114: should say something like "has no outstanding rebate amount"

tgrathwell:
seems good just some comments on the test names

PR #1347, opened 2021-07-27
CTC intake: record eip entry method for better bookkeeping and back/forward flow

tgrathwell:
`eip1_entry_method` and `eip2_entry_method` can be the followng values:
* `unfilled`, initially
* `calculated_amount`, if the client said "Yes, I received this amount." on /stimulus-payment
* `did_not_receive`, if the client said "No" on one of the "did you receive...?" pages
* `manual_entry`, if the client said "Yes" on one of the "did you receive...?" pages and entered a number

Keeping the exact reason why the number got set allows us to do the appropriate
bookkeeping in `self.show?()` navigation methods while also keeping real values
in the eip1_amount_received field for calculations later


app/controllers/ctc/questions/stimulus_one_controller.rb
```diff
@@ -5,9 +5,14 @@ class StimulusOneController < QuestionsController
...
 
+      def self.show?(intake)
+        return false if intake.eip1_entry_method_calculated_amount?
+        true
+      end
+
       def update
-        current_intake.update!(eip1_amount_received: 0)
-        redirect_to questions_stimulus_two_path
+        current_intake.update!(eip1_amount_received: 0, eip1_entry_method: :did_not_receive)
```
bytheway875, line 14: I think doing update right in the controller definitely simplifies the implementation, but we miss out on the ability to leverage the automatic reporting of the answers to mixpanel by doing this. I'm not sure if the original implementor knew that we make that tradeoff when we override update.

Then again, since this will be in the db it call also be looked at in mixpanel.

bytheway875: This actually seems like the only page that doesn't use the form pattern... I'm wondering if its worth just converting it to use a form and default behavior of update.

bytheway875 (approved):
The change I'm requesting is secondary to the purpose of this PR.

bytheway875 (approved):

PR #1348, opened 2021-07-27
Use id for most recent tax return instead of created_at so that scope is based on an indexed column

bytheway875:
By design, does not need additional testing because the old tests for the same scope still pass


alaroia-cfa (approved):
Splendid! Should be faster.

PR #1349, opened 2021-07-27
CTC intake: feature test that a dependent can actually show on the dependents review

tgrathwell:
Fixes an issue whereby the ChildDisqualifiersController (and thereafter most of the
dependents questions) would not show up because it was trying to pull the current
dependent from the URL but the url still says /dependents/new.


tgrathwell (commented):
Wow! This is splendid! - AL

PR #1357, opened 2021-07-28
Irs1040pdf

bytheway875:
Screen Shot 2021-07-28 at 12 58 32 PM
Screen Shot 2021-07-28 at 12 58 41 PM
Screen Shot 2021-07-28 at 12 58 54 PM


app/controllers/ctc/questions/stimulus_received_controller.rb
```diff
@@ -7,13 +7,13 @@ class StimulusReceivedController < QuestionsController
 
       def self.show?(intake)
         return false if intake.eip1_amount_received.nil? || intake.eip2_amount_received.nil?
-
-        intake.tax_return(2020).outstanding_recovery_rebate_amount <= 0
+        intake.tax_return(2020).outstanding_recovery_rebate_credit.zero?
```
bytheway875, line 10: The logic for stimulus received displaying was flip flopped. "Stimulus Received" text says:
          title: "Based on your info, it looks like you’ve received your full stimulus payments."

It should display only when the outstanding amount is determined to be 0.

spec/controllers/ctc/questions/stimulus_two_controller_spec.rb
```diff
@@ -9,7 +9,7 @@
   end
 
   describe "#update" do
-    it "saves 0 as the amount for stimulus 2 and redirects to the stimulus-received" do
+    it "saves 0 as the amount for stimulus 2 and redirects to stimulus owed" do
```
bytheway875, line 12: Same as above -- by saying they did not receive stimulus and they have money they are eligible for, they should be sent to stimulus-owed to claim their owed stimulus should they so choose.

spec/models/dependent_spec.rb
```diff
@@ -154,7 +154,7 @@
     context "with a qualifying child" do
       let(:dependent) do
         build :dependent,
-              relationship: "niece",
+              relationship: "NIECE",
```
bytheway875, line 157: implicitly tests case insensitivity of qualifying relationship method

app/helpers/pdf_helper.rb
```diff
@@ -34,4 +34,10 @@ def output_file
     PdfForms.new.fill_form(source_pdf_path, pdf_tempfile.path, hash_for_pdf)
     pdf_tempfile
   end
+
+  def pdf_mask(string, unmasked_char_count = 0)
+    return string unless string.present?
+
+    string.gsub(/.(?=.{#{unmasked_char_count}})/, 'X')
```
embarnard, line 41: It'd be easier to understand if this were implemented with `.slice` e.g. `masked = string.slice(-unmasked_char_count...)` plus adding the "X"s back to the start.

We attempted to use https://regex101.com/ to find out exactly how the regex works, but we weren't able to get very far that way; we did test it manually in `rails c` and it seems to do the right thing.

Not a blocker.

app/models/dependent.rb
```diff
@@ -170,11 +170,11 @@ def eligible_for_eip2?
   end
 
   def qualifying_child_relationship?
-    QUALIFYING_CHILD_RELATIONSHIPS.include? relationship
+    QUALIFYING_CHILD_RELATIONSHIPS.include? relationship.downcase
```
embarnard, line 173: Would this be tidier as a validation on `relationship`? Why would we ever get non-downcase'd relationships, and are non-downcase'd relationships stored in the database? Can we avoid that?

Not a blocker.

app/models/intake.rb
```diff
@@ -319,6 +319,7 @@ def self.searchable_fields
     }
   }
 
+  # Memoize response into instance variable.
```
embarnard, line 322: It's not memoized because there's no `@` here. Is that intentional?

(Not a blocker, but it is confusing.)

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
+require "rails_helper"
+
+require "rails_helper"
```
embarnard, line 3: `require`ing it once should be enough! ;)

Not a blocker, but would be a good cleanup.

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+        submission.address.destroy!
+        submission.intake.bank_account.destroy!
+        submission.intake.dependents.destroy_all
+        submission.tax_return.update(filing_status: nil)
+        submission.reload
+      end
+
+      it "returns a pdf with default fields and values" do
+        output_file = pdf.output_file
+        result = non_preparer_fields(output_file.path)
```
embarnard, line 26: Why are we removing the intake_specialist fields? Would just using `filled_in_values` in PdfSpecHelper accomplish the same thing in this test?

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+          "TotalItemizedOrStandardDedAmt12" => "12400",
+          "TaxableIncomeAmt15" => "0",
+          "RecoveryRebateCreditAmt30" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "RefundableCreditsAmt32" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "TotalPaymentsAmt33" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "OverpaidAmt34" => "0",
+          "RefundAmt35" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "PrimarySignature" => "#{submission.intake.primary_first_name} #{submission.intake.primary_last_name}",
+          "PrimarySignatureDate" => "01/01/20",
+          "PrimaryIPPIN" => "12345",
```
embarnard, line 169: If we're masking the SSN, should we also mask the IP PIN? Maybe mask it down to 2 digits?

What's the purpose of the masking? When a client wants their full 1040 from an e-file preparer, I would assume it would contain the full information they provided so they can e.g. check for SSN errors.

The IP PIN is probably about as sensitive as the SSN, hence the question.

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+          "AdjustedGrossIncomeAmt11" => "1",
+          "TotalItemizedOrStandardDedAmt12" => "12400",
+          "TaxableIncomeAmt15" => "0",
+          "RecoveryRebateCreditAmt30" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "RefundableCreditsAmt32" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "TotalPaymentsAmt33" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "OverpaidAmt34" => "0",
+          "RefundAmt35" => submission.tax_return.claimed_recovery_rebate_credit.to_s,
+          "PrimarySignature" => "#{submission.intake.primary_first_name} #{submission.intake.primary_last_name}",
+          "PrimarySignatureDate" => "01/01/20",
```
embarnard, line 168: Will it be OK if we set this date to the server's timezone view of the time of signing, even if that ends up with a different date than what the client has in their timezone?

It does seem OK to have this date printed into the 1040 be with regard to the server's time zone, but we wanted to ask.

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+      before do
+        submission.intake.update(
+          spouse_first_name: "Randall",
+          spouse_last_name: "Rouse",
+          spouse_signature_pin_at: Date.new(2020, 1, 5),
+          spouse_ip_pin: "123456",
+          spouse_ssn: "123456789"
+        )
+        submission.tax_return.update(filing_status: "married_filing_jointly")
+        submission.reload
```
embarnard, line 189: We suspect this line isn't needed.

spec/lib/adv_ctc_irs1040_pdf_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+          "DependentSSN[0]" => "XXXXX6789",
+          "DependentCTCInd[0]" => "0", # checked
+          "DependentLegalNm[1]" => "Daniel Dob",
+          "DependentRelationship[1]" => "son",
+          "DependentSSN[1]" => "XXXXX6788",
+          "DependentCTCInd[1]" => "0", # checked
+          "DependentLegalNm[2]" => "Mother Dob",
+          "DependentRelationship[2]" => "parent",
+          "DependentSSN[2]" => "XXXXX5788",
+          "DependentCTCInd[2]" => "1", # checked
```
embarnard, line 250: "1" is unchecked! Change the comment?

embarnard (approved):
Approved by Asheesh and Em! Left some comments but no blockers :)

app/lib/adv_ctc_irs1040_pdf.rb
```diff
@@ -0,0 +1,83 @@
...
+    answers = {
+        FilingStatus: @tax_return.filing_status_code,
+        PrimaryFirstNm: @intake.primary_first_name,
+        PrimaryLastNm: @intake.primary_last_name,
+        PrimarySSN: pdf_mask(@intake.primary_ssn, 4),
+        AddressLine1Txt: @address&.street_address,
+        CityNm: @address&.city,
+        StateAbbreviationCd: @address&.state,
+        ZIPCd: @address&.zip_code,
+        VirtualCurAcquiredDurTYInd: false,
```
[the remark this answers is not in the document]
bytheway875: the intention is to fill these out the exact way we fill out the other ones -- one of the ONLY required fields on the whole return is this box, oddly enough. They've already approved it for the digital 1040, but need to visually show it here as well.

app/lib/adv_ctc_irs1040_pdf.rb
```diff
@@ -0,0 +1,83 @@
...
+        SpouseFirstNm: @intake.spouse_first_name,
+        SpouseLastNm: @intake.spouse_last_name,
+        SpouseSSN: pdf_mask(@intake.spouse_ssn, 4),
+        SpouseSignature: @intake.spouse_full_name,
+        SpouseSignatureDate: @intake.spouse_signature_pin_at.strftime("%m/%d/%y"),
+        SpouseIPPIN: @intake.spouse_ip_pin,
+    }
+  end
+
+  # TODO: The tax form only allows for 4 dependents. In the case where we have more than 4 dependents, we
```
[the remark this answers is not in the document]
bytheway875: No, that's not necessary nor accurate. We can submit digital PDFs with up to 100 dependents -- it's just the paper return that we're filling here that limits you to 4. This paper return is to provide a visual to testers and provide to filers as a record of our filing. It does not in any way restrict the functionality of the site. The limitation has already been communicated to product.

app/models/dependent.rb
bytheway875: no we can't avoid it because we use the same field to store free-form relationships on GYR intake and input from forms, etc. It's possible to create a conditional validation based on the type of intake the dependent is associated with, but that can still "silently" disqualify dependents in tests if they're associated to an intake after creation, or cause issues with updating existing dependents who have "invalid" dependent relationships based on newer logic

I think we should "fix" this by making relationship an enum field and remove the free form entry for dependents on GYR intake.

app/models/intake.rb
bytheway875: oh, i need to remove this note. I had tried doing something and then it caused issues so i removed it.

spec/lib/adv_ctc_irs1040_pdf_spec.rb
bytheway875: Rae indicated that the 1040s they get from TaxSlayer have important information masked and that is the "safe" way to generate these. Re: other sensitive fields - product will review and make additional requests if necessary once i can get this in front of them, but i started with the fields Rae explicitly requested to be masked.

PR #1361, opened 2021-07-29
Bug fixes

bytheway875:
Small bug fixes to fix:
- Going back to bank account information page was failing because incorrect method name was being called. Fixed + Added test
- Link from submission page to client page was routing to index instead of show page url for clients


embarnard (approved):
Looks good!

PR #1362, opened 2021-07-29
Add button to "not file" submissions and transition submission + tax return statuses.

bytheway875:
Screen Shot 2021-07-29 at 11 04 57 AM

### DIFF
--- a/app/controllers/hub/efile_submissions_controller.rb
+++ b/app/controllers/hub/efile_submissions_controller.rb
@@ -3,7 +3,7 @@ module Hub
     include AccessControllable
     before_action :require_sign_in
     authorize_resource
-    load_resource only: [:resubmit]
+    load_resource except: [:index, :show]
     layout "admin"
 
     def index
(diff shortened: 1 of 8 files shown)


app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -3,7 +3,7 @@ class EfileSubmissionsController < ApplicationController
     include AccessControllable
     before_action :require_sign_in
     authorize_resource
-    load_resource only: [:resubmit]
+    load_resource except: [:index, :show]
```
caitbot, line 6: I'm wondering why transition this to `except` vs `only` (and include the `:cancel`)?

I have a mild bias towards `only` just because every time I read an `except` I wonder what we aren't `except`ing and so it involves one more step of investigation but I've also not used `only` much so I don't know if it has limitations. This is non-blocking and just curiosity.

caitbot (approved):

app/controllers/hub/efile_submissions_controller.rb
bytheway875: i guess my thought process here was that like ... most of our controllers use load_and_authorize_resource by default, so switching it around to "except" seemed more clear that those were "exceptions" to our typical logic.

PR #1365, opened 2021-07-30
Generate pdf

bytheway875:
I realize now that the content of the PDF may not actually be different at any of these three creation times, and that we maybe could get away with just changing the filename or display name.


app/models/intake/ctc_intake.rb
```diff
@@ -345,14 +345,4 @@ def any_ip_pins?
   def filing_jointly?
     client.tax_returns.last.filing_status_married_filing_jointly?
   end
-
```
bytheway875, line 348: i moved this from intake over to tax return because there was no good way to memoize the tax_return(2020) query without the contents getting stale, and loading the intake onto the tax return is simpler than loading one of many tax_returns on relation to intake.

app/jobs/build_submission_bundle_job.rb
```diff
@@ -7,6 +7,13 @@ def perform(submission_id)
       return
     end
 
+    begin
+      submission.generate_form_1040_pdf("preparing")
+    rescue
```
jenny-heath, line 12: Often best to `rescue StandardError`

PR #1366, opened 2021-07-30
Rake flow_explorer:capture_screenshots will run various commands to get screenshots

tgrathwell:


tgrathwell (commented):
AL approves and paired on this

PR #1368, opened 2021-08-02
Generate PDF at preparing, accepted, and rejected stage

bytheway875:


app/jobs/build_submission_bundle_job.rb
```diff
@@ -7,6 +7,13 @@ def perform(submission_id)
       return
     end
 
+    begin
+      submission.generate_form_1040_pdf("preparing")
```
tgrathwell, line 11: Let's use the submission ID instead and only call this once in the lifecycle of an EfileSubmission. This is consistent with @bytheway875 's remark in #1365 that the PDF contents won't change.

bytheway875: we do actually "reprepare" the same submission until the point at which the bundle is transmitted successfully to the IRS. So we should leave it getting created here, but we can remove it on the accepted and rejected part.

PR #1371, opened 2021-08-02
EfileSubmission runbook

tgrathwell:
Trying to document how EfileSubmission state transition work
and what we might do when things become :failed

Leaving for review/discussion here; plan to merge around 1pm Tuesday.

### DIFF
--- a/docs/efile-submission-runbook.md
+++ b/docs/efile-submission-runbook.md
@@ -1,0 +1,43 @@
+# EfileSubmission runbook
+
+Date: 2021-08-02
+
+Authors: Asheesh Laroia, Travis Grathwell
+
+## Purpose
+Engineers might have to handle EfileSubmission records during on-call.
+This [runbook](https://en.wikipedia.org/wiki/Runbook) is a guide on how to do that.
+
+## Typical operation
+
+An EfileSubmission record is created during intake when clients have finished submitting the information needed to submit an Advance CTC return.
+
+The EfileSubmission has a state machine attached (EfileSubmissionStateMachine) to it. The standard flow is:
+
+- A record is created as `:new`.
+- When it is transitioned to `:preparing`, we create a BundleSubmissionJob and update the tax return status for Hub users. BundleSubmissionJob validates the client's address, generates a 1040 PDF for the client and Hub users to see, and creates a submission ZIP file. It transitions the EfileSubmission to `:queued`.
+- When it is transitioned to `:queued`, we create a SendSubmissionJob. This launches gyr-efiler to send the submission ZIP file to the IRS. This sets the state to `:transmitted`.
+- Over time, a cron job will run that polls the IRS for status of transmitted submissions. It launches gyr-efiler with the IRS submission IDs. When the IRS responds with the status, the cron job marks the EfileSubmission as `:accepted` or `:rejected` and updates the TaxReturn status.
+
+An EfileSubmission can transition to `:resubmitted` if it submission ZIP never reached the IRS.
+
+## Manual interventions in e.g. rails console
+
+Try to always use `transition_to!(...)` to transition an EfileSubmission's state, rather than launching delayed jobs directly.
+
+Get the current state with `EfileSubmission#current_state`.
+
+## When a submission is `:failed`
+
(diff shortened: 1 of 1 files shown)


docs/efile-submission-runbook.md
```diff
@@ -0,0 +1,43 @@
...
+
+Get the current state with `EfileSubmission#current_state`.
+
+## When a submission is `:failed`
+
+You can see all submissions, including failed ones, in the e-file dashboard at /en/hub/efile ([dev](http://localhost:3000/en/hub/efile), [demo](https://demo.getyourrefund.org/en/hub/efile), [prod](https://www.getyourrefund.org/en/hub/efile)).
+
+We transition the submission to `:failed` when something goes wrong. Depending on the situation, it should be handled by engineers or by client support. The state machine ensures that when the EfileSubmission is `:failed`, we change the tax return status to `:file_needs_review` aka "Needs review" in the Hub.
+
+When client support handles issues, typically they either edit the client's data directly in the Hub if the next step is clear, or alternatively they get in touch with the client to resolve the issue. Client support can also decide they don't want to fix the issue and mark the e-file submission as "Cancelled" within the dashboard.
```
bytheway875, line 36: i don't think the case is that client support decides they "don't want to fix the issue" -- the use case would be that a client has asked us not to refile, or the resolution to fix it is out of the bounds of the types of cases we handle. 

I suggest something like "To close out an e-file case and mark as no further action necessary, Client support can mark the e-file submission as "Cancelled" on the dashboard which will also transition the associated Tax Return object to status of Not Filing"

docs/efile-submission-runbook.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Manual interventions in e.g. rails console
+
+Try to always use `transition_to!(...)` to transition an EfileSubmission's state, rather than launching delayed jobs directly.
+
+Get the current state with `EfileSubmission#current_state`.
+
+## When a submission is `:failed`
+
+You can see all submissions, including failed ones, in the e-file dashboard at /en/hub/efile ([dev](http://localhost:3000/en/hub/efile), [demo](https://demo.getyourrefund.org/en/hub/efile), [prod](https://www.getyourrefund.org/en/hub/efile)).
```
bytheway875, line 32: While there is no way in the UI to filter right now, I added a param that allows you to filter down to a particular status to hold us over. `hub/efile?status=rejected` will filter to rejected state. Only one status at a time is supported at the moment.

docs/efile-submission-runbook.md
```diff
@@ -0,0 +1,43 @@
...
+An EfileSubmission record is created during intake when clients have finished submitting the information needed to submit an Advance CTC return.
+
+The EfileSubmission has a state machine attached (EfileSubmissionStateMachine) to it. The standard flow is:
+
+- A record is created as `:new`.
+- When it is transitioned to `:preparing`, we create a BundleSubmissionJob and update the tax return status for Hub users. BundleSubmissionJob validates the client's address, generates a 1040 PDF for the client and Hub users to see, and creates a submission ZIP file. It transitions the EfileSubmission to `:queued`.
+- When it is transitioned to `:queued`, we create a SendSubmissionJob. This launches gyr-efiler to send the submission ZIP file to the IRS. This sets the state to `:transmitted`.
+- Over time, a cron job will run that polls the IRS for status of transmitted submissions. It launches gyr-efiler with the IRS submission IDs. When the IRS responds with the status, the cron job marks the EfileSubmission as `:accepted` or `:rejected` and updates the TaxReturn status.
+
+An EfileSubmission can transition to `:resubmitted` if it submission ZIP never reached the IRS.
```
bytheway875, line 22: When clicking the "resubmit" button, the current EfileSubmission object will always transition to resubmitted. In the case that it was never transmitted to the IRS, we transition from resubmitted back to preparing (re-prepare the same submission). In the case that we have already transmitted the submission to the IRS, that IRS ID cannot be used again so we create a new EfileSubmission object. Both of these actions happen in the after_transition callback to :resubmitted.

alaroia-cfa:
@bytheway875 I've incorporated new text based on the feedback you provided. Let me know if you want more time before 1pm to comment further. Thank you for the useful feedback! Feel free to keep it coming.

I also added you as an author; we don't have a strong team norm on who we put in that line, and I figured that it's useful for people to know you've participated in the creation of this doc.

PR #1372, opened 2021-08-02
Rejection message

bytheway875:


app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -8,6 +8,7 @@ class EfileSubmissionsController < ApplicationController
 
     def index
       @efile_submissions = EfileSubmission.most_recent_by_tax_return.page(params[:page])
+      @efile_submissions = @efile_submissions.in_state(params[:status]) if params[:status].present?
```
bytheway875, line 11: very quick implementation to allow for filtering by status on efile submission page, otherwise the page becomes pretty useless. No UI -- that will come later.

app/models/efile_submission_transition.rb
```diff
@@ -31,16 +31,26 @@ def initiated_by
     User.find(metadata["initiated_by_id"])
   end
 
-  def error_message
-    metadata["error_message"]
-  end
+  def stored_errors
```
bytheway875, line 34: `errors` as a method name is reserved by ActiveModel which is a bummer because it would have been the perfect name for a method here.

app/lib/efile/error.rb
```diff
@@ -0,0 +1,11 @@
+module Efile
+  class Error
+    attr_accessor :code, :category, :severity, :message
+    def initialize(code: nil, message:, category: nil, severity: nil)
```
caitbot, line 4: `message: nil`?

app/services/replacement_parameters_service.rb
```diff
@@ -1,11 +1,12 @@
 class ReplacementParametersService
-  attr_accessor :body, :client, :preparer_user, :locale, :tax_return
+  attr_accessor :body, :client, :preparer_user, :locale, :tax_return, :intake
```
caitbot, line 2: looks like this change was left in from the portal work you were doing -- remove this before merging

spec/lib/efile/submission_rejection_parser_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+describe Efile::SubmissionRejectionParser do
+  let(:raw_response) { file_fixture("irs_acknowledgement_rejection.xml").read }
+
+  describe '#to_xml' do
+    it 'outputs the raw response a Nokogiri XML document' do
```
caitbot, line 7: missing a `to` in `outputs the raw response a Nokogiri XML document`

spec/services/replacement_parameters_service_spec.rb
```diff
@@ -66,6 +66,14 @@
         expect(subject.process).to eq "Log in here: http://test.host/es/portal/login"
       end
     end
+
+    context "with a ctc intake" do
+      let(:client) { create :client, intake: create(:ctc_intake, preferred_name: "Preferred Name"), tax_returns: [create(:tax_return)], vita_partner: (create :vita_partner, name: "Koala County VITA") }
```
caitbot, line 71: more portal things

spec/state_machines/efile_submission_state_machine_spec.rb
```diff
@@ -82,6 +91,8 @@
       end
     end
 
+
```
caitbot, line 94: maybe remove these?

caitbot (approved):
in general it looks good -- left a few comments

app/services/replacement_parameters_service.rb
bytheway875: This is relevant to this PR because we've got to send the "right" login link based on the type of intake you have. the <<Client.LoginLink>> needs to get replaced with one thing if it's being sent to a GYR intaker and another thing if its a CTC intaker. Otherwise, when they click the login link they'd always end up on GYR.

PR #1373, opened 2021-08-03
Remove "outreach" messaging from emails + remove flagging for negative states

bytheway875:
Looking for ways to streamline workload for customer service at launch, we are removing the outreach sentences from the emails. We also will not be turning on the "flag" for clients when rejected or failed -- this will allow customer service to keep focus on clients who have actively contacted us.

### DIFF
--- a/app/state_machines/efile_submission_state_machine.rb
+++ b/app/state_machines/efile_submission_state_machine.rb
@@ -38,14 +38,12 @@ class EfileSubmissionStateMachine
   end
 
   after_transition(to: :failed) do |submission|
-    submission.client.flag!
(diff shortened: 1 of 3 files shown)


alaroia-cfa (approved):
This is great!

I noticed that fewer translation keys were changed in `es.yml` than `en.yml`. Is that intentional?

If yes, then please merge! If not, feel free to make the appropriate deletions and merge without further review.

bytheway875:
@alaroia-cfa I believe it's because there was some mixed guidance on how to handle the email copy. For two of the emails we used google translate for the spanish so they're already in there, but there's a third email that was added to the ignore list for spanish. Kelly said they're all out for real translation now.

PR #1374, opened 2021-08-03
Update 2020 income requirement question

bytheway875:
Screen Shot 2021-08-02 at 8 11 03 PM


caitbot (approved):

bytheway875:
@caitbot Let it be known that both you and I saw that we were asked to add a ul to an h1 and complied. 😬

PR #1375, opened 2021-08-03
Client portal rejects

bytheway875:
Screen Shot 2021-08-03 at 10 13 07 AM
Screen Shot 2021-08-03 at 10 17 36 AM

### DIFF
--- a/app/assets/stylesheets/templates/_portal.scss
+++ b/app/assets/stylesheets/templates/_portal.scss
@@ -30,7 +30,6 @@
     .form-question {
       font-size: 1.9rem;
       margin-bottom: 2.5rem;
-      font-weight: normal;
     }
     .textarea {
       height: 200px;
--- a/app/controllers/ctc/portal/portal_controller.rb
+++ b/app/controllers/ctc/portal/portal_controller.rb
@@ -4,10 +4,13 @@ class Ctc::Portal::PortalController < ApplicationController
 
   def home
     if current_client.efile_submissions.any?
-      @status = current_client.efile_submissions.last.current_state
+      submission = current_client.efile_submissions.last
+      @status = submission.current_state
+      @errors = submission.last_transition&.stored_errors
       @current_step = nil
     else
       @status = "intake_in_progress"
+      @errors = nil
       @current_step = current_client.intake.current_step
     end
   end
--- a/spec/factories/efile_submissions.rb
+++ b/spec/factories/efile_submissions.rb
@@ -38,9 +38,8 @@ FactoryBot.define do
 
     trait :with_errors do
       after :create do |submission|
-        file = Pathname.new(File.join(Rails.root, "/spec/fixtures/files", "irs_acknowledgement_rejection.xml")).read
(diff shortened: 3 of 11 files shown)


app/assets/stylesheets/templates/_portal.scss
```diff
@@ -30,7 +30,6 @@
     .form-question {
       font-size: 1.9rem;
       margin-bottom: 2.5rem;
-      font-weight: normal;
```
alaroia-cfa, line 33: Out of curiosity, was this line basically doing nothing?

app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -4,10 +4,13 @@ class Ctc::Portal::PortalController < ApplicationController
 
   def home
     if current_client.efile_submissions.any?
-      @status = current_client.efile_submissions.last.current_state
+      submission = current_client.efile_submissions.last
+      @status = submission.current_state
+      @errors = submission.last_transition&.stored_errors
```
alaroia-cfa, line 9: The `EfileSubmissionTransition#stored_errors` method is cool! :D

app/assets/stylesheets/templates/_portal.scss
bytheway875: this line was making "what's on your mind" label normal weight instead of bold, because that's what had been asked for in the original mock for the portal messages page. In the mocks for the new update, that label is now bolded. The change will affect the current implementation on GYR portal as well as CTC portal. It is more consistent with what labels generally look like in our app, so seems like a good change.

spec/factories/efile_submissions.rb
```diff
@@ -35,5 +35,13 @@
         end
       end
     end
+
+    trait :with_errors do
+      after :create do |submission|
+        file = Pathname.new(File.join(Rails.root, "/spec/fixtures/files", "irs_acknowledgement_rejection.xml")).read
+        raw_response = Nokogiri::XML(file).to_xml
```
alaroia-cfa, line 42: Since you're storing the raw XML contents as a string in the `jsonb` Postgres field, I think it's more accurate to never parse the data with Nokogiri. Consider instead:

```
        raw_xml = File.read(File.join(Rails.root, "spec/fixtures/files", "irs_acknowledgement_rejection.xml"))
        submission.efile_submission_transitions.last.update(metadata: { raw_response: raw_xml })
```

Also, to my eyes, it's more common to do `spec/fixtures/files` as the subdirectory path, not `/spec/fixtures/files` as the subdirectory path; I use a leading slash to indicate a pathname is absolute, so I found leading slash confusing.

Not a blocker.

alaroia-cfa (approved):
This is great. I have some minor suggestions and questions; no blockers. Cheers!

app/controllers/ctc/portal/portal_controller.rb
bytheway875: Yeah, this is the little guy i have some questions about re: whether we should actually persist these so that we can expose rejection reasons to metabase in a more direct manner.

PR #1376, opened 2021-08-03
Expose efile_submission_transitions and efile_submissions in metabase

bytheway875:
Somewhat fulfills the needs of the metabase story -- rejection codes etc are hidden in the metadata. Also, is exposing the metadata going to be an issue? It could possibly leak PII because the IRS provides the contents of certain nodes.

### DIFF
--- a/db/create_analytics_views.sql
+++ b/db/create_analytics_views.sql
@@ -69,6 +69,14 @@ CREATE VIEW analytics.documents_requests AS
     SELECT id, created_at, intake_id, updated_at
     FROM public.documents_requests;
 
+CREATE VIEW analytics.efile_submission_transitions AS
+    SELECT id, created_at, efile_submission_id, metadata, most_recent, sort_key, to_state, updated_at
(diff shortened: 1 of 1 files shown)


bytheway875:
@alaroia-cfa I'm thinking that maybe exposing the raw metadata isn't a great idea, because it could possibly include some PII (if name field is incorrect, will show the name that was sent. If SSN field has issue, will show the contents of what was sent in the SSN field)

I'm currently reading the raw response to create "synthetic" error objects on the efile_submission_transition object. However, maybe it makes sense to persist those somewhere instead? Wondering if you have time to chat about that sometime today.

alaroia-cfa (changes requested):
I think it'd be good to merge this with efile_submission_transitions.metadata and efile_submissions.irs_submission_id removed. That way, we get easy stats in Metabase and the comfort of no sensitive data. I know you have follow-up work in progress to store the error codes; we can merge that too when it's time.

bytheway875:
Removed!

PR #1377, opened 2021-08-03
Format Name Line 1 for the IRS

caitbot:
This just adds better support for name line 1 of the 1040 header, as outlined on [page 189 of this doc](https://www.irs.gov/pub/irs-pdf/p4164.pdf)

To-do:
- Have a conversation with design and add validation?? for special characters (though we strip them all out anyways so maybe this step isn't entirely necessary but it could lead to heavy manipulation if we dont)
- Support for suffixes like JR, II
- Before doing that, take a peek at prod data for occurrences of those cases and what format they are usually in because who knows maybe we can simplify things

### DIFF
--- a/app/lib/submission_builder/formatting_methods.rb
+++ b/app/lib/submission_builder/formatting_methods.rb
@@ -15,8 +15,35 @@ module SubmissionBuilder
     end
 
     def name_line_1_type(primary_first, primary_middle, primary_last, spouse_first, spouse_middle, spouse_last)
+      name_line = build_name_line_1(primary_first, primary_middle, primary_last, spouse_first, spouse_middle, spouse_last)
+
+      # if the line > 35 characters then truncate according to the guidelines on page 189 of https://www.irs.gov/pub/irs-pdf/p4164.pdf
+      if name_line.size > 35 && spouse_last.present? # shorten spouse last
+        name_line = build_name_line_1(primary_first, primary_middle, primary_last, spouse_first, spouse_middle, spouse_last.first)
+      end
+      if name_line.size > 35 # shorten primary last
+        name_line = build_name_line_1(primary_first, primary_middle, primary_last.first, spouse_first, spouse_middle, spouse_last&.first)
+      end
+      if name_line.size > 35 && spouse_middle.present? # remove spouse middle
+        name_line = build_name_line_1(primary_first, primary_middle, primary_last.first, spouse_first, nil, spouse_last.first)
+      end
+      if name_line.size > 35 && primary_middle.present? # remove primary middle
+        name_line = build_name_line_1(primary_first, nil, primary_last.first, spouse_first, nil, spouse_last&.first)
+      end
+      if name_line.size > 35 && spouse_first.present? # shorten spouse first
+        name_line = build_name_line_1(primary_first, nil, primary_last.first, spouse_first.first, nil, spouse_last.first)
+      end
+      if name_line.size > 35 # shorten primary first
+        name_line = build_name_line_1(primary_first.first, nil, primary_last.first, spouse_first&.first, nil, spouse_last&.first)
+      end
+
+      name_line
+    end
+
+    # The IRS has very particular guidelines for what this line should look like and they are
+    # outlined on page 189 of https://www.irs.gov/pub/irs-pdf/p4164.pdf
+    def build_name_line_1(primary_first, primary_middle, primary_last, spouse_first, spouse_middle, spouse_last)
       # add in support for formatting JR/2nd/II
-      # add in support for 35+ char names
       name_line = formatted_first_name(primary_first)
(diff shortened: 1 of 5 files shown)


app/lib/submission_builder/return_header1040.rb
```diff
@@ -53,7 +53,7 @@ def document
           xml.Filer {
             xml.PrimarySSN intake.primary_ssn
             xml.SpouseSSN intake.spouse_ssn if tax_return.filing_jointly?
-            xml.NameLine1Txt person_name_type(client.legal_name)
+            xml.NameLine1Txt name_line_1_type(intake.primary_first_name, intake.primary_middle_initial, intake.primary_last_name, intake.spouse_first_name, intake.spouse_middle_initial, intake.spouse_last_name)
```
caitbot, line 56: let it be known that I do not love this -- its a lot of parameters. But it genuinely needs all of them and separately because it needs to rearrange them. And since it is only a formatting method, I didn't really want to pass it the entire intake either.

app/lib/submission_builder/formatting_methods.rb
```diff
@@ -14,22 +14,51 @@ def date_type(date)
...
     # phone number without country code or formatting
     # results in 10 digit number for transmitting to the IRS
     def phone_type(string)
       Phonelib.parse(string, "US").national(false)
     end
+
+    private
+
+    def formatted_first_name(name)
+      I18n.transliterate(name).upcase.gsub(/[^A-Z]/, '')
```
bytheway875, line 54: woah fancy -- good find!

bytheway875: i wonder if it would be helpful to add a comment to these methods, both to explain what we're stripping, and to make the difference between the first name and last name logic obvi without reading the regex.

app/lib/submission_builder/return_header1040.rb
bytheway875: lmao, yeah. pretty gross, but i guess its better to pass in the individual strings than the whole object. Although, maybe passing the intake wouldnt be the worst thing in the world... There's a tradeoff because if for some reason you changed the order of one of these parameters, the whole thing would be incorrect, right? so that's something to think about as well.

spec/lib/submission_builder/formatting_methods_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    context 'when there is a spouse name' do
+      let(:spouse_first) { "Cora" }
+      let(:spouse_middle) { "O" }
+
+      context "with a different last name" do
+        let(:spouse_last) { "Coconuts" }
+        it 'formats it correctly' do
+          expect(formatted_name).to eq("SAMANTHA<SEASHELLS<& CORA O COCONUTS")
```
bytheway875, line 53: Do you understanding the logic re: why you sometimes use the < and sometimes you don't? if so, can you explain it to me? Also, maybe we should add a link to the IRS PDF with a page number either as a comment in the tests or the implementation because this is like ... so specific and weird.

caitbot: Oh its kind of awful. My understanding is:
There is a maximum of 2 `<`
First name and middle initial are always concatenated without `<`.
Single filers:
Between a first/middle and the last name add a `<`.
Between last name and suffix add a `<`
Filers with same last name:
Concatenate their first/middles THEN add `<` then last name then `<` then suffix (so these last two steps are shared with the single filers)
Filers with different last names
Build single filer name, if no suffix then add `<` between last and `&`, otherwise if there is a suffix, you've reached your 2 `<` limit and so don't add one before the ampersand. Concatenate the first/middle/last of the spouse without special formatting.

I don't know why they decided to do this.

caitbot: I'll add a link to the pdf in a comment, I like that idea.

bytheway875: yeah, and even that logic there if you _wanted_ to, you could create one of those internal documentation things. OR make a chore to do it at some point in the future and link to this github convo here. OR do nothing.

app/lib/submission_builder/return_header1040.rb
caitbot: True, if we started to call this in more places than just here I would argue we should pass in the intake because copying over parameters is hard. But this should remain isolated and we have tests on the return 1040 header to catch any parameter swaps.

bytheway875:
@caitbot Based on the tests, it looks like we still need to restrict this to a max length of 35 characters total. Which seems somewhat odd because _how is it meaningful if it isn't a full name_? But them's the rules!

NameLine1Txt': [facet 'maxLength'] The value has a length of '61'; must be maximum 35 characters.

bytheway875 (approved):

PR #1382, opened 2021-08-03
Create file-full-return page in CTC intake flow

embarnard:


config/locales/en.yml
```diff
@@ -1551,6 +1551,18 @@ en:
               <li>tips <li>commissions </li>
               <li>self-employment or contract payments</li>
             </ul>
+        file_full_return:
+          title: If you earned income, you could collect additional benefits by filing a full return!
+          body_html: |
+            <p>By filing a full tax return, you could collect all the money you are owed.</p>
+            <p>Consider this option if:</p>
```
bytheway875, line 1558: i know we have some mixed paradigms on how we handle stuff like this, but think its somewhat risky and smelly to put so much html into the translation. That will means that the spanish translation needs to be formatted in the exact same way, and that the appearance of the spanish translation page may look different if we edit the content differently in the future. This is especially more important now that we are expecting the translators to provide us with translation files that we can just copy and paste into our app. 
Consider a format that leaves as much of the html in the view.
```
help_text: 
   - By filing...
   - Consider this option if...
list: 
  - You have all...
  - You have a paper copy...
file_for_free:   blah blah
```

```
...
<%= t("help_text").each do |help| %>
    <p><%= help %></p>
<% end %>

<ul class="list--bulleted">
   <%= t("list").each do |item| %>
      <li><%= item %></li>
   <% end %>
</ul>

bytheway875 (changes requested):
Please refactor some of the html out of the translation file to make spanish translation simpler and less error prone.

config/locales/en.yml
embarnard: @bytheway875 Hmm I'm not sure if I agree -- if the Spanish page will look different doesn't it make more sense to put the HTML in the YML where there can be different formatting for English vs Spanish then to leave it in the HTML where we would be tied to the structure of one or the other? But I could see how if we leave it in the YML and copied the en.yml to ask for Spanish translations it could confuse the translators to see html tags

bytheway875: the spanish page should use the same layout, but could end up looking different because the html is in the yml. If you left off the class bulleted list class or forgot to close a tag, the page would look really different. But yes, I agree that the primary reason to leave as much html out of the yml as possible is because it makes it more difficult to pass off the translation to non-engineering folks!

PR #1385, opened 2021-08-03
Rename total_session_active_seconds to previous_sessions_active_seconds

alaroia-cfa:
This trivial follow-up to #1378 improves a column name.

We used to call something a "total" that wasn't actually a total.


tgrathwell:
🤷

PR #1387, opened 2021-08-04
Move efile submission show page to the client profile

bytheway875:
Third (fourth?) times a charm in reconfiguring where this page is and what it displays. Better than ever!

All of the links on the efile index now link to this page, which simplifies things quite nicely. 

Screen Shot 2021-08-03 at 9 14 26 PM

### DIFF
--- a/app/views/shared/_tax_return_list.html.erb
+++ b/app/views/shared/_tax_return_list.html.erb
@@ -45,13 +45,7 @@
           <% end %>
         <% else %>
           <div>
-            <% if tax_return.has_submissions? %>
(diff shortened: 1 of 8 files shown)


app/views/shared/_tax_return_list.html.erb
```diff
@@ -45,13 +45,7 @@
           <% end %>
         <% else %>
           <div>
-            <% if tax_return.has_submissions? %>
```
bytheway875, line 48: Remove link from status to efile submission page, as it will be right below it.

alaroia-cfa (approved):
Looks great! 🚢

PR #1390, opened 2021-08-04
Capitalize device ID [#179119657]

alaroia-cfa:
I chose not to add validation at the model level because if someone's browser really submits junk data to us in this field, and we want to block further intake progress, we don't currently have a good way to indicate the reason to them, as far as I can tell.

### DIFF
--- a/app/javascript/lib/efile_security_information.js
+++ b/app/javascript/lib/efile_security_information.js
@@ -5,17 +5,17 @@ export function getEfileSecurityInformation() {
     let Fingerprint2 = initFingerprint2();
     let CryptoJS = initCryptoJS();
     Fingerprint2.get(function(components) {
-        var stringtohash = components.map(function (pair) { return pair.value }).join('###')
-        let encrypted_device_id =  CryptoJS.SHA1(stringtohash);
-        document.getElementById('ctc_consent_form_device_id').value = encrypted_device_id;
+        let concatenated = components.map(function (pair) { return pair.value }).join('###')
+        let concatenatedAndHashed = CryptoJS.SHA1(concatenated).toUpperCase();
+        document.getElementById('ctc_consent_form_device_id').value = concatenatedAndHashed;
     });
 
     document.getElementById('ctc_consent_form_user_agent').value = navigator.userAgent;
     document.getElementById('ctc_consent_form_browser_language').value = navigator.language;
     document.getElementById('ctc_consent_form_platform').value = navigator.platform;
-    var loadDate = new Date();
-    document.getElementById('ctc_consent_form_client_system_time').value = loadDate;
-    document.getElementById('ctc_consent_form_timezone_offset').value = loadDate.getTimezoneOffset();
+    let now = new Date();
+    document.getElementById('ctc_consent_form_client_system_time').value = now;
+    document.getElementById('ctc_consent_form_timezone_offset').value = now.getTimezoneOffset();
 }
 
 
--- a/app/models/client/efile_security_information.rb
+++ b/app/models/client/efile_security_information.rb
@@ -25,7 +25,7 @@
 class Client::EfileSecurityInformation < ApplicationRecord
   belongs_to :client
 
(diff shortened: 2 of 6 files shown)


app/javascript/lib/efile_security_information.js
```diff
@@ -5,17 +5,17 @@ export function getEfileSecurityInformation() {
     let Fingerprint2 = initFingerprint2();
     let CryptoJS = initCryptoJS();
     Fingerprint2.get(function(components) {
-        var stringtohash = components.map(function (pair) { return pair.value }).join('###')
-        let encrypted_device_id =  CryptoJS.SHA1(stringtohash);
-        document.getElementById('ctc_consent_form_device_id').value = encrypted_device_id;
+        let concatenated = components.map(function (pair) { return pair.value }).join('###')
+        let concatenatedAndHashed = CryptoJS.SHA1(concatenated).toUpperCase();
```
alaroia-cfa, line 9: I renamed this variable because SHA1 isn't an encryption function, but a [hashing function.](https://en.wikipedia.org/wiki/SHA-1)

app/javascript/lib/efile_security_information.js
```diff
@@ -5,17 +5,17 @@ export function getEfileSecurityInformation() {
...
+        document.getElementById('ctc_consent_form_device_id').value = concatenatedAndHashed;
     });
 
     document.getElementById('ctc_consent_form_user_agent').value = navigator.userAgent;
     document.getElementById('ctc_consent_form_browser_language').value = navigator.language;
     document.getElementById('ctc_consent_form_platform').value = navigator.platform;
-    var loadDate = new Date();
-    document.getElementById('ctc_consent_form_client_system_time').value = loadDate;
-    document.getElementById('ctc_consent_form_timezone_offset').value = loadDate.getTimezoneOffset();
+    let now = new Date();
```
alaroia-cfa, line 16: I decided to be bold and rename this variable to my taste.

app/models/client/efile_security_information.rb
```diff
@@ -25,7 +25,7 @@
 class Client::EfileSecurityInformation < ApplicationRecord
   belongs_to :client
 
-  # storing client_system_time as a string and then transforming it into DateTime for the return_header2040
+  # storing client_system_time as a string and then transforming it into DateTime for the return_header1040
```
alaroia-cfa, line 28: Fixed typo in comment

spec/controllers/ctc/questions/consent_controller_spec.rb
```diff
@@ -49,7 +49,7 @@
             phone_number: "831-234-5678",
             timezone: "America/Chicago",
             primary_tin_type: "ssn",
-            device_id: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+            device_id: "7BA1E530D6503F380F1496A47BEB6F33E40403D1",
```
alaroia-cfa, line 52: I wanted to use a realistic device ID, since the AAAA one is just to be used if the device ID is missing.

bytheway875:
i dont think we have to _validate_ it on the model, but it might make sense to move the data coercion/formatting to the model-- that way implementations on the javascript and ruby form objects don't diverge. 

Same for the coercion of adding the + on the timezone offset -- those both currently happen either in the JS OR on the form, both of which I think we'll have to duplicate and ensure the implementations stay in sync when capturing at the beginning and end of flow. If we coerced the data before we saved it into the db instead, we'd ensure that the coercion is always "in sync" regardless of where you attempt to create an object from.

The other option would be to make sure that we use the SAME javascript and Ruby form (or concern) to handle the coercion of the data every time we capture it.

bytheway875:
I think all the improvements made here are great, and I'm betting that as the work unfolds for adding this at the end of the flow, any refactors that make sense to ensure the implementations are consistent will bubble up.

bytheway875 (approved):

alaroia-cfa:
@bytheway875 Good feedback! As you said, I'll aim to incorporate that in future work on the same topic, coming soon.

PR #1395, opened 2021-08-04
Fix intake flow to ask eligibility questions before basic info

caitbot:
This required splitting `Can someone claim you as a dependent in 2020?` from `Were you an active member of the military in 2020?` The dependent one remains in the eligibility flow before consent and the military question was moved to the consent page.

image
image


app/controllers/ctc/questions/filed2020_controller.rb
```diff
@@ -11,6 +10,10 @@ def method_name
         "filed_2020"
       end
 
+      def next_path
+        @form.filed_2020? ? questions_filed2020_yes_path : super
```
bytheway875, line 14: is filed2020_yes a real path?!

app/controllers/ctc/questions/filed2020_yes_controller.rb
```diff
@@ -1,14 +1,9 @@
 module Ctc
   module Questions
     class Filed2020YesController < QuestionsController
-      include AuthenticatedCtcClientConcern
 
       layout "intake"
```
bytheway875, line 5: yes it is! haha

app/controllers/ctc/questions/life_situations2020_controller.rb
```diff
@@ -1,16 +1,18 @@
...
 
       def illustration_path; end
 
+      def method_name
+        "can_be_claimed_as_dependent"
+      end
+
       def next_path
-        @form.cannot_be_claimed_as_dependent? ? questions_use_gyr_path : super
+        @form.can_be_claimed_as_a_dependent? ? questions_use_gyr_path : super
```
bytheway875, line 15: oh woah, was this logic just backwards?

app/controllers/ctc/questions/filed2020_controller.rb
caitbot: hahahaha yes I made it one but I'm realizing now I can make it prettier

app/controllers/ctc/questions/life_situations2020_controller.rb
caitbot: no -- this is confusing and a clean up needs to happen but what we store on the intake (and is now an abandoned column) is `cannot_be_claimed_as_dependent` but it doesn't make sense to ask a question that is yes or no that is `Can you not be claimed as a dependent by anyone?` because double negatives, so I flipped it

bytheway875 (approved):
cool, i pulled this down and clicked through and works great!

app/controllers/ctc/questions/life_situations2020_controller.rb
bytheway875: ohh i see. makes some sense.

PR #1397, opened 2021-08-05
Design/Eng pair copy and design changes

embarnard:


app/controllers/ctc/questions/dependents/confirm_dependents_controller.rb
```diff
@@ -6,14 +6,17 @@ class ConfirmDependentsController < QuestionsController
 
         layout "intake"
 
+        # def self.show?(intake)
```
caitbot, line 9: If this is another story's work than you want to remove this + the spec for the sake of isolating changes.

caitbot (approved):
Looks good, I rebased and in doing so discovered that some of `consent` was renamed `legal_consent` and so I went around and consistently made it `legal_consent`. If that wasn't the intention then feel free to roll back my changes post-rebase. Otherwise ship it! 🎉

PR #1399, opened 2021-08-05
Expose efile errors to metabase

bytheway875:
Exposes the two new tables for tracking rejection reasons to metabase.


bytheway875:
merged this so we can expose to Max before code freezing for a couple hours.

caitbot (commented):
lgtm

PR #1400, opened 2021-08-05
Flow explorer: fall back to s3 image urls for missing images on localhost

tgrathwell:


app/controllers/flows_controller.rb
```diff
@@ -29,14 +29,25 @@ def show
 
   private
 
-  def screenshot_base
-    if Rails.env.development?
+  def screenshot_path(controller)
+    [screenshot_base, I18n.locale, "#{controller.name}.png"].join('/')
```
alaroia-cfa, line 33: I think it's better to check on disk then if not, provide the S3 URL, and only do that in dev.

alaroia-cfa (changes requested):

PR #1401, opened 2021-08-05
Flow explorer: magic intake creator for CTC intakes

tgrathwell:


alaroia-cfa (approved):
This will increase the rate of nonsense intakes on demo. That's probably a good thing. :ship:

Glad it's not enabled for prod! Glad we added a reveal/hide.

PR #1402, opened 2021-08-05
Fix EIP Go back bug by resetting values to defaults when clicking i havent received all of my stimulus

bytheway875:
This adds a new entry to QuestionNavigation. It is a no-op for edit (it never responds true to show), but we link to the update page from the link so that we can have a form action that clears the EIP data.

### DIFF
--- a/app/views/ctc/questions/stimulus_payments/edit.html.erb
+++ b/app/views/ctc/questions/stimulus_payments/edit.html.erb
@@ -29,7 +29,7 @@
         <%= t("views.ctc.questions.stimulus_payments.yes_received") %>
       <% end %>
 
-      <%= link_to questions_stimulus_one_path, class: "button button--wide text--centered" do %>
+      <%= link_to questions_stimulus_reset_path, method: :put, class: "button button--wide text--centered" do %>
         <%= t("views.ctc.questions.stimulus_payments.no_did_not_receive") %>
       <% end %>
     </div>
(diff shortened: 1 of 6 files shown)


embarnard (approved):
Looks good!

app/views/ctc/questions/stimulus_payments/edit.html.erb
```diff
@@ -29,7 +29,7 @@
         <%= t("views.ctc.questions.stimulus_payments.yes_received") %>
       <% end %>
 
-      <%= link_to questions_stimulus_one_path, class: "button button--wide text--centered" do %>
+      <%= link_to questions_stimulus_reset_path, method: :put, class: "button button--wide text--centered" do %>
```
tgrathwell, line 32: Since the StimulusPaymentsController is already accepting a form POST that can write to the `eip1_amount_received`/`eip2_amount_received` fields, I think it's better to  to adjust StimulusPaymentsForm to save different data to those fields depending on what button you clicked.

One way to do that is to have two `<%= f.button %>`s, one for yes and one for no, like this other CTC form erb:

https://github.com/codeforamerica/vita-min/blob/ca593acc4f0a6e1c7bfc32faab128cd1abde91a9/app/views/ctc/questions/stimulus_owed/edit.html.erb#L23-L26

In that case, `StimulusPaymentsForm#save` can decide which data to change.

Not a blocker - it's already merged - wanted to share a pattern we already use elsewhere.

alaroia-cfa: https://github.com/codeforamerica/vita-min/pull/1404 implements this approach.

PR #1404, opened 2021-08-05
CTC intake: remove StimulusResetController in favor of the two buttons on StimulusPayments doing different things

tgrathwell:


alaroia-cfa (approved):
This is a fine change. Glad that the feature spec is unchanged.

PR #1406, opened 2021-08-05
Update translations

caitbot:


config/locales/es.yml
```diff
@@ -784,28 +784,19 @@ es:
       married_filing_separately: Casado que presenta una declaración por separado (MFS)
       head_of_household: Cabeza de familia (HOH)
       qualifying_widow: Viudo que reúne los requisitos (QW)
-    dependent_relationships:
```
caitbot, line 787: These weren't used or in the `en.yml` so I removed them.

config/locales/es.yml
```diff
@@ -784,28 +784,19 @@ es:
...
     dependent_relationships:
       00_daughter: Hija
       01_son: Hijo
       02_parent: "Madre/Padre"
       03_grandchild: "Nieta/Nieto"
-      04_niece: Niece
-      05_nephew: Nephew
+      04_niece: Sobrina
+      05_nephew: Sobrino
       06_foster_child: Foster Child
```
bytheway875, line 794: i guess we're still waiting on translations for some of these?

config/locales/es.yml
```diff
@@ -1349,22 +1340,43 @@ es:
     final_tax_doc_and_unsigned_8879_have_tax_return: "%{document_type} debe estar asociado con un año fiscal"
   views:
     ctc:
+      portal:
+        home:
+          title: Â¡Gracias por presentar su declaraciÃ³n con GetCTC!
```
bytheway875, line 1345: oh no! all the special characters are fucked up.

config/locales/es.yml
```diff
@@ -2208,19 +2358,6 @@ es:
         see_faq: Consulte nuestras preguntas frecuentes
         support_html: Nuestro equipo está aquí para ayudarle. %{faq_link} o envíe una pregunta por chat, y nos comunicaremos con usted.
         title: Preguntas sobre los impuestos
-      sms_terms:
```
caitbot, line 2211: This was a duplicate of another `sms_terms` and so I just took the latest and assumed it was the most up to date

config/locales/es.yml
caitbot: I would assume? We didn't actually get any updates to relationships in the latest translation file though

caitbot: We might actually want to remove the untranslated ones in order to get translations

config/locales/es.yml
caitbot: I was wondering if that was the case or if `Ã³` means something -- its all over the translation file ):

caitbot: agh https://stackoverflow.com/questions/20543364/not-getting-spanish-characters-%C3%BA-%C3%AD-but-getting-like-%C3%83%C2%B3-%C3%83-on-ui

caitbot: Seems like an encoding error

caitbot: I can probably fix it..

alaroia-cfa: `Ã³` looks to me like we were given UTF-8 data and then re-interpreted the bytes as ISO-8859-1 aka Latin-1. See also https://en.wikipedia.org/wiki/Mojibake :D

PR #1408, opened 2021-08-05
Fix issue with efile 1040 values and bolster adv ctc 1040 tests

bytheway875:
The paper 1040 PDF was correct, but the efiled 1040 was providing the incorrect amount (0 instead of the rrc amount, thinking it was supposed to be a value for payments MADE instead of expected payments TO the client), which resulted in a rejected tax return.


bytheway875 (commented):
reviewed with CT

PR #1409, opened 2021-08-05
Emails: show 'GetCTC' in the header and have an @getctc.org reply address when relevant

tgrathwell:


config/environments/demo.rb
```diff
@@ -3,9 +3,13 @@
 Rails.application.configure do
   config.active_storage.service = :s3_demo
 
-  config.action_mailer.default_options = { from: 'hello@mg-demo.getyourrefund-testing.org' }
-  config.address_for_transactional_authentication_emails = 'no-reply@mg-demo.getyourrefund-testing.org'
   config.ctc_url = "https://ctc.demo.getyourrefund.org"
+  ctc_email_from_domain = "mg-demo-ctc.getyourrefund-testing.org"
```
bytheway875, line 7: are these configured in mailgun to work? I

bytheway875: Screen Shot 2021-08-09 at 10 35 22 AM


In order for this to work, we'd need to set up new sending domains for the staging and demo environments. It seems like we'd only be doing that for testing purposes, and that it would not correspond to how we handle things in production, where there would _not_ be an mg-ctc, as we handle that through mail fowarding rules set up by IT.

I'm wondering if we can set up mail forwarding rules through IT for demo and staging so that all of the environments are set up in the same manner.

PR #1412, opened 2021-08-09
Fix duplicated tax return problem on efile show

bytheway875:
BERFORE FIX:
Screen Shot 2021-08-09 at 11 03 27 AM

We were loading in tax returns joined to efile_submissions, which meant that every efile_submissions tax return was loaded. We really want it to be unique tax returns, for all tax returns that have submissions.

### DIFF
--- a/app/controllers/hub/efile_submissions_controller.rb
+++ b/app/controllers/hub/efile_submissions_controller.rb
@@ -15,7 +15,7 @@ module Hub
     # loops through the tax_returns that have efile_submissions.
     def show
       @client = Client.find(params[:id])
-      @tax_returns = @client.tax_returns.joins(:efile_submissions) # get all tax returns with submissions
+      @tax_returns = @client.tax_returns.joins(:efile_submissions).uniq # get all tax returns with submissions
       redirect_to hub_client_path(id: @client.id) and return unless @tax_returns.present?
     end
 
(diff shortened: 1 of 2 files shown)


caitbot (approved):

app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -15,7 +15,7 @@ def index
     # loops through the tax_returns that have efile_submissions.
     def show
       @client = Client.find(params[:id])
-      @tax_returns = @client.tax_returns.joins(:efile_submissions) # get all tax returns with submissions
+      @tax_returns = @client.tax_returns.joins(:efile_submissions).uniq # get all tax returns with submissions
```
tgrathwell, line 18: `.uniq` is an in-memory operation, so this will still retrieve however many EfileSubmission records there are in the DB and then discard them in Rubyland

It's not a big deal since any given client has <4 TaxReturn records, but if we were doing this in a paginated list or something it wouldn't be ideal

[I think the SQLier way is more like](https://stackoverflow.com/a/35508839)

`@client.tax_returns.includes(:efile_submissions).where.not(efile_submissions: { id: nil })`
or
`@client.tax_returns.joins(:efile_submissions).distinct`

bytheway875:
Ah yeah, I should probably use distinct and then eager load the efile
submissions because I really want those to be loaded so I can loop through
those in the view

PR #1420, opened 2021-08-09
Use FederalOriginalSubmissionIdDt for resubmission timestamp in 1040 XML header [#179170521]

alaroia-cfa:
This is important because the old XML tag name did not match the IRS schema,
which meant we could not re-submit a return if it had already been transmitted
once to the IRS.


alaroia-cfa:
This should allow us to re-submit https://demo.getyourrefund.org/en/hub/clients/34490/efile on demo.

bytheway875 (approved):
Ah lovely, good catch.

PR #1422, opened 2021-08-09
If submission generation raises an error, switch to failed [#179170521]

alaroia-cfa:

## Why this is needed

In the case of https://demo.getyourrefund.org/en/hub/clients/34487/efile#body , we ran into an exception while generating the submission bundle.

If this happens in the future, we'll have to look in Sentry for the details of the exception, so I added that remark here.

In this case, the codebase already has a fix for the issue; however, the EfileSubmission is in the wrong state. Once this pull request is approved, I'll mark the EfileSubmission as `:failed` which will allow it to be resubmitted within the Hub.

### DIFF
--- a/app/jobs/build_submission_bundle_job.rb
+++ b/app/jobs/build_submission_bundle_job.rb
@@ -9,16 +9,15 @@ class BuildSubmissionBundleJob < ApplicationJob
 
     begin
       submission.generate_form_1040_pdf
-    rescue
-      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL', raw_response: ['Engineers should look in Sentry for a crash'])
+    rescue StandardError => e
+      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL', raw_response: ["Engineers should look in Sentry for an exception", "EfileSubmission #{submission_id}", e.class.name])
       raise
     end
 
     begin
       response = SubmissionBundle.build(submission, documents: ["adv_ctc_irs1040"])
-    rescue
-      submission.transition_to!(:failed, error_code: 'BUNDLE-FAIL', raw_response: ['Engineers should look in Sentry for a crash'])
-      raise
+    rescue StandardError => e
+      submission.transition_to!(:failed, error_code: 'BUNDLE-FAIL', raw_response: ["Engineers should look in Sentry for an exception", "EfileSubmission #{submission_id}", e.class.name])
     end
 
     if response.valid?
--- a/app/jobs/build_submission_bundle_job.rb
+++ b/app/jobs/build_submission_bundle_job.rb
@@ -14,7 +14,13 @@ class BuildSubmissionBundleJob < ApplicationJob
       raise
     end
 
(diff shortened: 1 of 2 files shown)


app/jobs/build_submission_bundle_job.rb
```diff
@@ -10,11 +10,17 @@ def perform(submission_id)
     begin
       submission.generate_form_1040_pdf
     rescue
-      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL')
+      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL', raw_response: ['Engineers should look in Sentry for a crash'])
```
bytheway875, line 13: instead of adding to the raw_response here, you could actually pass in the error contents by doing like rescue e  and then raw_response e.message. I'm guessing you considered that and decided to go this route for a particular reason.

alaroia-cfa: Yeah - mostly I wanted to avoid there being two places to look into something. Sentry is going to have better information, and if there's any sanitization that we want to do, Sentry can do it, so I prefer directing people to look at Sentry.

bytheway875: yeahhh... but also will it be clear in sentry which one was causing the failure? I don't know if sentry will reliably link the id of the failed submission in a way that is exposed in the UI (the job uses submission id, and we only display the irs id in the hub. there's not really a reason that this needs to be the case). If we displayed the "main" raised error, it might help match things up in sentry faster but idk

bytheway875: maybe we can add the submission id or something else we can look for that would have been raised in sentry to identify?

alaroia-cfa: @bytheway875 Yeah - maybe a nice idea to log the exception class and our ID (not IRS submission ID) - I'll add that right now!

bytheway875 (approved):

app/jobs/build_submission_bundle_job.rb
```diff
@@ -9,12 +9,17 @@ def perform(submission_id)
 
     begin
       submission.generate_form_1040_pdf
-    rescue
-      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL')
+    rescue StandardError => e
+      submission.transition_to!(:failed, error_code: 'PDF-1040-FAIL', raw_response: ["Engineers should look in Sentry for an exception", "EfileSubmission #{submission_id}", e.class.name])
```
bytheway875, line 13: The contents in raw_response outputs "raw"-ly in a <pre><code> block, so i think this may get output as an array. doesn't really matter, and i guess we can also loop through items in raw response if we want to format things more nicely, so just an FYI and not a blocker.

alaroia-cfa: @bytheway875 oh hah! I will adjust this to be plain text, then merge.

alaroia-cfa: Actually I'll re-request your review just to make sure we're on the same page. Thanks for a fruitful discussion.

PR #1424, opened 2021-08-09
Touch last_seen_at when a client logs in

caitbot:
Fixes [this bug](https://www.pivotaltracker.com/story/show/179169548) 🐞

**What was going wrong?**
We want to figure out how long a client spent filling out questions and report it to the IRS. Each time a client signs in, Devise updates `last_sign_in_at`. Each time they answer a question on intake, we update/touch `last_seen_at`. 

When it comes time to either report session time to the IRS, we subtract `last_sign_in_at` from `last_seen_at` because we assume that we saw them sometime after their last login.

However, what if their last action was to login? This would mean that `last_seen_at` was still set to their LAST session's last action and we would get a negative number of milliseconds. The IRS doesn't like negative numbers. Woops.

**How does this fix it?**
We had two options - zero out `last_seen_at` on login or update/touch `last_seen_at` on login. The effect is about the same. I consider logging in to be a "seen" action and so I chose the latter. 

The one tiny downside to this is that there may actually be a few milliseconds between Devise setting their value and us setting ours. Fortunately, neither us nor the IRS requires that level of precision.

### DIFF
--- a/app/controllers/portal/client_logins_controller.rb
+++ b/app/controllers/portal/client_logins_controller.rb
@@ -66,6 +66,7 @@ module Portal
       if @form.valid?
         sign_in @form.client
         @form.client.accumulate_total_session_durations
+        @form.client.touch :last_seen_at
         redirect_to session.delete(:after_client_login_path) || portal_root_path
       else
         @clients.each(&:increment_failed_attempts)
--- a/spec/controllers/portal/client_logins_controller_spec.rb
+++ b/spec/controllers/portal/client_logins_controller_spec.rb
@@ -1,4 +1,5 @@
 require "rails_helper"
+include ActiveSupport::Testing::TimeHelpers
 
 RSpec.describe Portal::ClientLoginsController, type: :controller do
   let(:client) do
@@ -211,9 +212,9 @@ RSpec.describe Portal::ClientLoginsController, type: :controller do
           end
 
           it "updates the clients last_seen_at" do
-            Timecop.freeze do
-              expect { post :update, params: params }
-                .to change{ client.reload.last_seen_at }.to(Time.now)
+            freeze_time do
+              post :update, params: params
+              expect(client.reload.last_seen_at).to eq Time.zone.now
             end
           end
 
--- a/spec/controllers/portal/client_logins_controller_spec.rb
+++ b/spec/controllers/portal/client_logins_controller_spec.rb
@@ -193,6 +193,7 @@ RSpec.describe Portal::ClientLoginsController, type: :controller do
         before { allow_any_instance_of(ClientLoginService).to receive(:clients_for_token).and_return(client_query) }
 
         context "with a matching ssn/client ID" do
+          before { Timecop.freeze }
           let(:params) do
             {
               id: "raw_token",
@@ -206,6 +207,7 @@ RSpec.describe Portal::ClientLoginsController, type: :controller do
             post :update, params: params
 
             expect(subject.current_client).to eq(client)
+            expect(subject.current_client.last_seen_at).to eq(Time.now)
             expect(response).to redirect_to portal_root_path
           end
 
--- a/spec/controllers/portal/client_logins_controller_spec.rb
+++ b/spec/controllers/portal/client_logins_controller_spec.rb
@@ -1,11 +1,10 @@
 require "rails_helper"
-include ActiveSupport::Testing::TimeHelpers
 
 RSpec.describe Portal::ClientLoginsController, type: :controller do
   let(:client) do
     create(
       :client,
-      last_seen_at: DateTime.new(2021, 4, 20, 16, 21),
+      last_seen_at: nil,
       intake: create(
         :intake,
         email_address: "client@example.com",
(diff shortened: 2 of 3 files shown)


spec/controllers/portal/client_logins_controller_spec.rb
```diff
@@ -209,6 +210,13 @@
             expect(response).to redirect_to portal_root_path
           end
 
+          it "updates the clients last_seen_at" do
```
alaroia-cfa, line 213: Can you change the use of Timecop so that you freeze to a particular time e.g. 

```Timecop.freeze(DateTime.new(2021, 8, 1, 12, 0, 0)) do
  expect(...).to eq(DateTime.new(2021, 8, 1, 12, 0, 0
  end```
  
As written, your test is not a very idiomatic usage of Timecop, since you aren't specifying _what_ time to freeze to.

**Review by alaroia-cfa — CHANGES_REQUESTED** [review]
I'd like to see the timecop part of the test be clearer. I like this approach overall!

File: spec/controllers/portal/client_logins_controller_spec.rb (hunk 1 above)
Comment (alaroia-cfa) [reply spec/controllers/portal/client_logins_controller_spec.rb:213]: If you do that, you'll probably enjoy having a `fake_time` variable so you aren't repeating yourself.

Also FYI I edited the above comment like 10 times after saving; sorry if the previous 9 versions were just distractingly incorrect. I hope this version is adequately correct.

Comment (alaroia-cfa) [reply spec/controllers/portal/client_logins_controller_spec.rb:213]: https://frontdeveloper.pl/2020/03/how-we-migrated-from-timecop-to-built-in-rails-5-2-time-helpers/ is a good reference, and also it shows that we can get rid of Timecop entirely :P

Comment (caitbot) [reply spec/controllers/portal/client_logins_controller_spec.rb:213]: how do you feel about `freeze_time do ... test things... end`? It looks like `freeze_time` doesn't accept any arguments.

Comment (alaroia-cfa) [reply spec/controllers/portal/client_logins_controller_spec.rb:214]: I'm OK with this!

File: spec/controllers/portal/client_logins_controller_spec.rb [hunk 2]
```diff
@@ -1,9 +1,11 @@
 require "rails_helper"
+include ActiveSupport::Testing::TimeHelpers
```
alaroia-cfa, line 2: I'd like to see the `include` be lower down, e.g. on line 5, so it's within the `RSpec.describe ... do ... end`.

spec/controllers/portal/client_logins_controller_spec.rb
alaroia-cfa: @tgrathwell @bytheway875 Do you have an opinion on `ActiveSupport::Testing::TimeHelpers` vs `Timecop`?

It seems that Rails 5.2 integrates something very similar to Timecop, and that life will be simpler if we use it.

If you think it's sensible to use `ActiveSupport::Testing::TimeHelpers`, then I propose we do this:

```
RSpec.configure do |config|
  ....
  config.include ActiveSupport::Testing::TimeHelpers
  ....
end
```

in `spec/rails_helper.rb`

alaroia-cfa: @caitbot For now, for this pull request, consider making the above change to `spec/rails_helper.rb` and removing the `include` from this file. Not a blocker.

spec/controllers/portal/client_logins_controller_spec.rb
```diff
@@ -1,9 +1,11 @@
...
 
 RSpec.describe Portal::ClientLoginsController, type: :controller do
   let(:client) do
     create(
       :client,
+      last_seen_at: DateTime.new(2021, 4, 20, 16, 21),
```
alaroia-cfa, line 8: If this is intended to be an arbitrary time in the past, consider setting it to `nil` instead.

alaroia-cfa (approved):
This is a great change. I have some suggestions, none of which are blockers.

spec/controllers/portal/client_logins_controller_spec.rb
bytheway875: honestly, im An Old and haven't kept track of what has been added to rails to replace the things in my toolbox. if we're doing things that make external dependencies obsolete, im always cool w/ that.

However, if the code is valid and my Old way fits the pattern we'd been using before and we're two days away from a launch, i suggest we make a ticket to improve it across the board later and move forward with whatever fits the existing pattern we'd used before.

app/controllers/portal/client_logins_controller.rb
```diff
@@ -66,6 +66,7 @@ def update
       if @form.valid?
         sign_in @form.client
         @form.client.accumulate_total_session_durations
```
bytheway875, line 68: if both of these values are trying to achieve a similar goal, do we need both?

caitbot: `accumulate_total_session_durations` looks at the last sign in and last seen at, takes their diff, and adds it to the previous session time counter. All I want to do is touch last seen at on login. I chose to not put it in that method because technically that method could be called from anywhere -- not just at login, and to keep the update to last seen at as a clear side effect of login.

Also, I may have misinterpreted your question, but hopefully that gives useful context.

PR #1425, opened 2021-08-09
Add cancelled status description and asset for portal

bytheway875:


spec/features/ctc/portal_spec.rb
```diff
@@ -267,4 +267,38 @@
...
+      fill_in "Enter 6 digit code", with: code
+      click_on "Verify"
+      expect(page).to have_selector("h1", text: "Authentication needed to continue.")
+      fill_in "Client ID or Last 4 of SSN/ITIN", with: intake.client.id
+      click_on "Continue"
+
+      expect(page).to have_selector("h1", text: "Thank you for filing with GetCTC!")
+      expect(page).to have_text I18n.t("views.ctc.portal.home.status.cancelled.label")
+      expect(page).to have_text I18n.t("views.ctc.portal.home.status.cancelled.message")
+      click_on "Message my tax preparer"
```
alaroia-cfa, line 297: Since we already have a feature spec for the messaging behavior, I propose you simplify this feature spec to only validate that cancelled label & message appear and not validate the messaging behavior.

alaroia-cfa (approved):
This is a good change, and it's fine as-is.

I think there's room for simplification in the feature spec. Feel free to simplify and/or re-word as suggested and merge, or to merge as-is.

PR #1427, opened 2021-08-10
Sort the Status Logs by time

alaroia-cfa:

## Why

I was looking at https://demo.getyourrefund.org/en/hub/clients/34490/efile and stuff was all in arbitrary orders and it was confusing me so I figure the logs can be sorted by time.

### DIFF
--- a/app/models/efile_submission.rb
+++ b/app/models/efile_submission.rb
@@ -18,7 +18,7 @@ class EfileSubmission < ApplicationRecord
   has_one :client, through: :tax_return
   has_many :dependents, through: :intake
   has_one :address, as: :record
-  has_many :efile_submission_transitions, class_name: "EfileSubmissionTransition", autosave: false, dependent: :destroy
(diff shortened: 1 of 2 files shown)


bytheway875:
Appreciate this. I cant think of a time when this isnt the order we want to have things in though, and im actually surprised this isn't the default. Can you go move this to a default on the efile_submission_transitions table itself?

bytheway875:
Oh, you know what? Yesterday, I discovered that there is a method called "history" on efile_submission that comes from the state machine, and now that im thinking about it that's probably what we should be using here -- i bet that's already ordered?

Although, i really think because of the way we use the transition objects so heavily and store associated errors etc, it might make sense to add a default scope for ordering ...

alaroia-cfa:
@bytheway875 Let me know if this is better! I did add the `order` in three places, just to be sure, and I picked `id` to order by rather than `created_at` since `id` is already indexed by default, I believe.

bytheway875 (approved):

PR #1428, opened 2021-08-10
Action Text: efile edit and show

bytheway875:
Screen Shot 2021-08-10 at 10 14 36 AM
Screen Shot 2021-08-10 at 10 14 48 AM
Screen Shot 2021-08-10 at 10 15 02 AM

### DIFF
--- a/app/assets/stylesheets/_honeycrisp-compact.scss
+++ b/app/assets/stylesheets/_honeycrisp-compact.scss
@@ -13,6 +13,13 @@
     min-height: 18rem;
   }
 
+  ul {
+    list-style-type: initial;
+  }
+
+  ol {
+    list-style-type: decimal;
+  }
   hr {
     background-color: $color-grey;
     margin-top: $s35;
--- a/app/models/efile_error.rb
+++ b/app/models/efile_error.rb
@@ -13,7 +13,8 @@
 #  updated_at :datetime         not null
 #
 class EfileError < ApplicationRecord
-  def self.pdf_generation_error
-    create_or_find_by!(source: :internal, code: 'PDF-1040', message: 'Failed to generate PDF Form 1040.')
-  end
+  has_rich_text :description_en
+  has_rich_text :description_es
+  has_rich_text :resolution_en
+  has_rich_text :resolution_es
 end
--- a/app/assets/stylesheets/actiontext.scss
+++ b/app/assets/stylesheets/actiontext.scss
@@ -1,0 +1,404 @@
+/*
+Trix 1.3.1
+Copyright © 2020 Basecamp, LLC
+http://trix-editor.org/*/
+trix-editor {
+  border: 1px solid #bbb;
+  border-radius: 3px;
+  margin: 0;
+  padding: 0.4em 0.6em;
+  min-height: 5em;
+  outline: none; }
+trix-toolbar * {
+  box-sizing: border-box; }
+trix-toolbar .trix-button-row {
+  display: flex;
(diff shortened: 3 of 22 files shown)


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -13,6 +13,13 @@
     min-height: 18rem;
   }
 
+  ul {
```
bytheway875, line 16: i dont know why we remove the default list styling for ul/ol in honeycrisp. It's very odd and feels like an anti-pattern, as I'd much rather _remove_ bullet styling from a list that doesnt need it when we don't need it than globally not have it. 


For our honeycrisp compact overrides, I turned the list styles back on. Having them off would mean I'd need to target the ActionText editor's ul/ol to style them (because our global removal affects that component as well).

This means that there may be some instances of ul in the hub that have some bullets next to them now, but i didn't see any in my (cursory) look. ill take another pass and see what the client page looks like.

app/models/efile_error.rb
```diff
@@ -13,7 +13,8 @@
 #  updated_at :datetime         not null
 #
 class EfileError < ApplicationRecord
-  def self.pdf_generation_error
-    create_or_find_by!(source: :internal, code: 'PDF-1040', message: 'Failed to generate PDF Form 1040.')
-  end
+  has_rich_text :description_en
```
alaroia-cfa, line 16: The examples in the screenshot all are plain text. This is rich text. Would it suffice to use plain text here? If so, I encourage that.

ActionText is new and it scares me! I am willing to learn not to be scared, and this could be irrational on my part, but that's my feeling at this moment. There are a lot of moving pieces of our app, and my general anxiety level is higher now than average.

I see at https://github.com/basecamp/trix/releases that the most recent release involved a HTML sanitization fixes, which on average, means the next one will as well, and therefore it'll be important to update it frequently.

I also see there's support for images in here. What's the significance/purpose of that?

alaroia-cfa (commented):
I'm not ready to get excited about ActionText yet, but I'm open to it. Maybe we can Slack chat or Tuple about it, or if you have more info, I can read links if you can recommend some.

I am glad we're gaining a `show` page!

app/models/efile_error.rb
bytheway875: the examples in the screenshots are not indicative of the text that they want to put in -- they're multi-paragraph and have links to other places. It essentially needs to either be WSYIWIG or markdown, and any time i've added an admin feature for non-technical folks that uses markdown, they have hated me forever.

While ActionText new, Rails is _generally_ thoughtful and careful about promoting things into Rails, so i feel confident that it will be okay. Also, the implementation is limited enough that if we end up hating it, we can rip it out fairly easily and store the contents of these messages elsehow. 

The other option would be to grab another WYSIWYG editor and save that data into a field on the model. That's what I was expecting we'd have to do until I read about ActionText and took a shot at configuring it. The way it does storage is likely a little overkill for our use case, but it does get us up and running SO much faster than if we had to roll our own solution.

alaroia-cfa (approved):
@bytheway875 Thanks for answering! This makes sense to me now.

:ship:

I encourage you to:

- Remove the gallery support per https://stackoverflow.com/questions/61880076/is-there-any-way-to-disable-image-attachment-part-for-action-text-in-rails-6

- Limit the allowed tags to just A, DIV, P, UL, LI
- Limit the allowed attributes to just HREF

Those are not blocking.

PR #1429, opened 2021-08-10
Allow 000000 code for email/phone verification on demo and development

embarnard:


app/forms/ctc/phone_verification_form.rb
```diff
@@ -5,6 +5,8 @@ class PhoneVerificationForm < QuestionsForm
     validates_presence_of :verification_code
 
     def valid?
+      return true if verification_code = "000000" && (Rails.env.demo? || Rails.env.development?)
```
alaroia-cfa, line 8: I think this should be `==`. I think that the way it's written (`=` vs. `==`), it will accept any code on demo/dev.

alaroia-cfa (changes requested):
Looks good - simple clear code! One issue about `=` vs `==` - if you can fix that, then consider it approved.

alaroia-cfa:
@embarnard One other thing actually -- can you move this check _after_ the `Rails.env` check?

alaroia-cfa:
@embarnard Correction - ignore my suggestion about re-ordering the `&&`; this is fine to merge!

What I was going to suggest was: `return true if (Rails.env.demo? || Rails.env.development?) && verification_code == "000000"` - that is, environment check first, data check second.

For a split second, I thought it did matter with regard to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack) (click the link for a long fascinating rabbit hole). It doesn't actually matter with regard to timing attacks, so please feel free to ignore this comment.

alaroia-cfa (approved):
@embarnard Looks good to me, assuming it's green - thanks!

embarnard:
> @embarnard Correction - ignore my suggestion about re-ordering the `&&`; this is fine to merge!
> 
> What I was going to suggest was: `return true if (Rails.env.demo? || Rails.env.development?) && verification_code == "000000"` - that is, environment check first, data check second.
> 
> For a split second, I thought it did matter with regard to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack) (click the link for a long fascinating rabbit hole). It doesn't actually matter with regard to timing attacks, so please feel free to ignore this comment.

All good! I already changed it -- thanks for the review Asheesh!

PR #1430, opened 2021-08-10
Remove hyphens from ssn before validation

caitbot:


app/models/intake.rb
```diff
@@ -264,6 +264,12 @@ def self.searchable_fields
   validates :email_address, 'valid_email_2/email': true
   validates :phone_number, :sms_phone_number, allow_blank: true, e164_phone: true
   validates_presence_of :visitor_id
+
+  before_validation do
```
caitbot, line 268: the upside to doing it here is that we don't need to have this `before_validation` on each of the individual forms

the downside to doing it here is that it will do this before every save of intake, and we save intake often. however, we have other `before_validations` in the form of `auto_strip_attribute` so this ~kind of~ keeps in pattern

an alternative is to set up custom setter methods that strip out hyphens before calling `super`

spec/factories/intakes.rb
```diff
@@ -273,10 +273,10 @@
   end
 
   trait :with_ssns do
-    primary_ssn { "#{Faker::Number.number(digits: 3)}00#{Faker::Number.number(digits: 4)}".to_i } # 0s in 4th and 5th position are required for IRS test submissions
```
caitbot, line 276: I removed the `to_i` because it makes the stripping of the `-` unhappy and really we should never be passing these around as integers anyhow, we get them as strings and save them as strings

app/models/intake.rb
bytheway875: yes, interesting point.

You could use ActiveRecord dirty/change methods to only do it when it's changed:

`primary_ssn_changed?` or something like that

spec/factories/intakes.rb
bytheway875: yes makes sense

app/models/intake.rb
caitbot: but wouldn't we have to do that within `before_validation` which means it would still have to check `primary_ssn_changed?` on each save, but that is marginally better than scrubbing the string of `-` on each save so I'll add that in

PR #1431, opened 2021-08-10
Send sms opt-in message when someone selects text message for preferred contact info

embarnard:


config/locales/es.yml
```diff
@@ -1061,6 +1061,7 @@ es:
     default_subject: Actualización de GetYourRefund
     default_subject_with_service_name: Actualización de %{service_name}
     sms_opt_in: Welcome to GetYourRefund messages. Msg&data rates may apply.  Msg freq varies. Reply HELP for subscription/privacy info, STOP to stop.
+    ctc_sms_opt_in: Welcome to GetCTC messages. Msg&data rates may apply. Msg freq varies. Reply HELP for subscription/privacy info, STOP to stop.
```
bytheway875, line 1064: weird. i guess these are both supposed to be in english? maybe because its legal text?

bytheway875: totally optional, but since the content is the same for ctc and regular sms, you could just pass in the service name and use the same translation key.

bytheway875 (approved):

PR #1432, opened 2021-08-10
Do not send GYR survey to CTC clients

caitbot:


app/models/tax_return.rb
```diff
@@ -296,8 +296,8 @@ def system_change_status(new_status)
     self.status = new_status
   end
 
-  def send_client_completion_survey
-    if saved_change_to_status? && TaxReturnStatus::TERMINAL_STATUSES.map(&:to_s).include?(status)
+  def send_gyr_client_completion_survey
```
bytheway875, line 299: would it be easier or harder to read if was something like:

after_update_commit :send_gyr_client_completion_survey, unless :is_ctc?


I will leave that determination up to you.

bytheway875 (approved):

app/models/tax_return.rb
caitbot: I'm going to leave it as is since we are already checking other conditions and I find it easiest to read if they are all in one place.

PR #1434, opened 2021-08-10
CTC intake: ensure birth date and SSN are loaded in forms when coming back to edit

tgrathwell:
Also only require SSN confirmation when the SSN has changed (previously
only worked like this in the dependents flow)

This attempts to provide a consistent experience for entering birth date
and TIN no matter whether you are doing it for the primary, spouse, or
dependent. These fields should always show up when reloading persisted
data and not require confirmation unless the values have changed.

Previously, birth date and SSN were always blank when editing persisted
data, because `existing_attributes` relies on `model.attributes`
and the values don't show up there (for SSN, because they are encrypted,
and for birth date, because the form has more fields than the model)

### DIFF
--- a/app/controllers/ctc/questions/home_controller.rb
+++ b/app/controllers/ctc/questions/home_controller.rb
@@ -9,6 +9,10 @@ module Ctc
       def next_path
         @form.lived_in_territory_or_at_foreign_address? ? questions_use_gyr_path : super
       end
+
+      def tracking_data
+        @form.attributes_for(:misc)
+      end
     end
   end
 end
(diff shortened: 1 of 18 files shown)


bytheway875:
The birthday thing makes sense to me, but the SSN kind of worries me.

 im not totally confident that our session management is tight enough to ensure that you can't accidentally go back and see sensitive information of another person when you're on a public computer, for instance.

tgrathwell (commented):
Reviewed with AL

Kelly is cool with 'back' showing the client's SSN, and AL and I think the session management will be OK given the 10 min automatic logout. Jenny's fixing the logout button, too.

One option is to mask the SSN when going forward/back, like `*****1111`. That might be a nice improvement.

PR #1439, opened 2021-08-10
Remove unnecessary fields from CTC form

caitbot:
Fix for[ this story](https://www.pivotaltracker.com/story/show/179182369)


app/views/shared/_client_primary_info_fields.html.erb
```diff
@@ -16,17 +16,19 @@
...
     <% end %>
 
-    <%= f.cfa_select(
-            :preferred_interview_language,
-            t("hub.clients.fields.preferred_interview_language"),
-            language_options(only_locales: false),
-            include_blank: true,
-            selected: @form.preferred_interview_language
-    ) %>
+    <% if is_dropoff || !is_ctc %>
```
bytheway875, line 23: this means it will stay on ctc dropoff right? That seems to match the logic from the ticket, but wanted to make sure im understanding correctly.

bytheway875 (approved):
This looks good to me! So much better to be able to launch with a form we don't have to enter bad data into to get it to work. Woo.

app/views/shared/_client_primary_info_fields.html.erb
caitbot: Yeah if it is dropoff it always shows, or if it is GYR it should always show

PR #1442, opened 2021-08-10
Add tax return download link to client portal

bytheway875:
Screen Shot 2021-08-10 at 6 27 32 PM
Screen Shot 2021-08-10 at 6 27 51 PM

### DIFF
--- a/app/controllers/ctc/portal/portal_controller.rb
+++ b/app/controllers/ctc/portal/portal_controller.rb
@@ -9,10 +9,12 @@ class Ctc::Portal::PortalController < ApplicationController
       @status = latest_transition.present? ? latest_transition.to_state : EfileSubmissionStateMachine.initial_state
       @exposed_error = latest_transition.present? ? latest_transition.client_facing_errors.first : nil
       @current_step = nil
+      @pdf1040 = current_client.documents.find_by(tax_return: submission.tax_return, document_type: DocumentTypes::Form1040.key)
     else
       @status = "intake_in_progress"
       @exposed_error = nil
       @current_step = current_client.intake.current_step
+      @pdf1040 = nil
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -9,10 +9,12 @@ def home
       @status = latest_transition.present? ? latest_transition.to_state : EfileSubmissionStateMachine.initial_state
       @exposed_error = latest_transition.present? ? latest_transition.client_facing_errors.first : nil
       @current_step = nil
+      @pdf1040 = current_client.documents.find_by(tax_return: submission.tax_return, document_type: DocumentTypes::Form1040.key)
```
alaroia-cfa, line 12: Should we be concerned about the possibility that there's more than one, for example, if the Hub user edits their data, then resubmits?

Not a blocker.

app/views/ctc/portal/portal/home.html.erb
```diff
@@ -24,6 +24,12 @@
             <% if @current_step %>
               <%= link_to "Complete CTC form", @current_step, class: "button" %>
             <% end %>
+
+            <% if @pdf1040 %>
+              <%= link_to portal_document_path(id: @pdf1040.id), target: "_blank", rel: "no_opener", class: "button" do %>
```
alaroia-cfa, line 29: `rel: "noopener"` not `rel: "no_opener"` but not a blocker

alaroia-cfa (approved):
Two questions and no blockers. Cheers!

app/controllers/ctc/portal/portal_controller.rb
bytheway875: we use the same logic we use for creating new versions of the 136146, which is that we just say "if there is already a 1040 for this tax return, regenerate and reattach the file to the same document object". So there is only ever one!

PR #1447, opened 2021-08-11
Rename Efiler config keys

alaroia-cfa:
## Why

The config key name I used in prod was wrong. This makes it easier to get right. Also it fixes the credentials for prod & demo.


embarnard (approved):
Looks good!

PR #1457, opened 2021-08-11
CTC clients: fix crash on portal when last several transitions were 'waiting' or 'investigating'

tgrathwell:


bytheway875:
ohhh yeah this makes sense. we're now allowing toggling back and forth between waiting and investigating.

PR #1458, opened 2021-08-11
Add check payment status faq under common questions

embarnard:


embarnard:
Screen Shot 2021-08-11 at 6 47 06 PM

config/locales/en.yml
```diff
@@ -3159,6 +3159,15 @@ en:
             answer_html: |
               An email address, ID, proof of income and/or benefits, and your child's Social Security Number.<br/><br/>
               Children claimed for the Child Tax Credit must have a valid Social security number (SSN), however the tax filer (a parent, grandparent, or guardian) can have either a SSN or an Individual Taxpayer Identification Number (ITIN).
+          check_payment_status:
+            question: "Where can I check the status of my payments?"
```
embarnard, line 3163: I know there was some conversation about not including the html styling in the YML but I figured this was okay because the rest of answer_htmls include html, it would require me to change the layout and the spanish was already included

config/locales/en.yml
```diff
@@ -3159,6 +3159,15 @@ en:
...
+            answer_html: |
+              <p>You can check the status of your payments with the IRS:</p>
+              <ul class="list--bulleted">
+                <li>For Child Tax Credit monthly payments check the Child Tax Credit Update Portal</li>
+                <li>For stimulus payments 1 and 2 check Where’s My Refund</li>
```
tgrathwell, line 3168: i think things like "Where's My Refund" are intended to be links

PR #1465, opened 2021-08-12
Log metrics when polling for acks

caitbot:
AL and I decided on an approach for storing data about the acks cron job

- check for ACKs on all outstanding returns, but in batches of 100, log in datadog how many outstanding returns there are when the job runs and how many we heard back about from the IRS
- add a column for last_checked_for_ack_at on submission
- add datadog alert if the chron job hasn't run, so add a measurement/logging of the chron job running and then config the alert in datadog if it doesn't happen in some reasonable timeframe

I still have to add the datadog alert -- that'll happen after merging

### DIFF
--- a/app/models/intake/ctc_intake.rb
+++ b/app/models/intake/ctc_intake.rb
@@ -178,7 +178,7 @@
 #  spouse_active_armed_forces                           :integer          default("unfilled")
 #  spouse_auth_token                                    :string
 #  spouse_birth_date                                    :date
-#  spouse_can_be_claimed_as_dependent                   :integer          default(0)
+#  spouse_can_be_claimed_as_dependent                   :integer          default("unfilled")
 #  spouse_consented_to_service                          :integer          default(0), not null
 #  spouse_consented_to_service_at                       :datetime
 #  spouse_consented_to_service_ip                       :inet
--- a/app/services/efile/poll_for_acknowledgments_service.rb
+++ b/app/services/efile/poll_for_acknowledgments_service.rb
@@ -1,10 +1,27 @@
 module Efile
   class PollForAcknowledgmentsService
     def self.run
-      transmitted_submissions = EfileSubmission.in_state(:transmitted).limit(100)
-      response = Efile::GyrEfilerService.run_efiler_command(Rails.application.config.efile_environment, "acks", *transmitted_submissions.pluck(:irs_submission_id))
+      ack_count = 0
+      submission_ids = transmitted_submission_ids
+      submission_ids.each_slice(100) do |submission_ids|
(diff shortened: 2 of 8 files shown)


app/models/intake/ctc_intake.rb
```diff
@@ -178,7 +178,7 @@
 #  spouse_active_armed_forces                           :integer          default("unfilled")
 #  spouse_auth_token                                    :string
 #  spouse_birth_date                                    :date
-#  spouse_can_be_claimed_as_dependent                   :integer          default(0)
+#  spouse_can_be_claimed_as_dependent                   :integer          default("unfilled")
```
caitbot, line 181: I know this is unrelated but its friendly and just an annotation change and so I vote we leave it

app/services/efile/poll_for_acknowledgments_service.rb
```diff
@@ -1,10 +1,27 @@
 module Efile
   class PollForAcknowledgmentsService
     def self.run
-      transmitted_submissions = EfileSubmission.in_state(:transmitted).limit(100)
-      response = Efile::GyrEfilerService.run_efiler_command(Rails.application.config.efile_environment, "acks", *transmitted_submissions.pluck(:irs_submission_id))
+      ack_count = 0
+      submission_ids = transmitted_submission_ids
+      submission_ids.each_slice(100) do |submission_ids|
+        ack_count = _handle_response(Efile::GyrEfilerService.run_efiler_command(Rails.application.config.efile_environment, "acks", *submission_ids))
```
bytheway875, line 7: does this handle the error we were seeing when there are no submissions? I think it should, (the loop wont run if there aren't any items in the first slice, i assume?) but just wanted to make sure.

db/schema.rb
```diff
@@ -377,6 +377,7 @@
   create_table "efile_submissions", force: :cascade do |t|
     t.datetime "created_at", precision: 6, null: false
     t.string "irs_submission_id"
+    t.datetime "last_checked_for_ack_at"
```
bytheway875, line 380: i'd love to see this somewhere on the activity log list... maybe we could make a "synthetic log" to inject this onto the history list?

not as part of this work obviously, but sometime soon.

bytheway875 (approved):
knowing not only when the last poll was, but when the last poll _for each submission_ was is like ... more than i ever could have thought to ask for. 🙏

app/services/efile/poll_for_acknowledgments_service.rb
caitbot: oh yeah, that hadn't even occurred to me, but this now will not run if it is empty

PR #1469, opened 2021-08-13
When looking for testing-only config settings, use `.try()` to avoid crash outside of tests

alaroia-cfa:
Fixes https://sentry.io/organizations/codeforamerica/issues/2572625680/?referrer=alert_email&environment=demo


alaroia-cfa:
Gonna merge now in the interest of unblocking demo; I'd love post-merge review from @jenny-heath since we paired on the original implementation and both missed this :D

PR #1477, opened 2021-08-13
Forward portal messages to intercom

caitbot:
See [the ticket](https://www.pivotaltracker.com/story/show/179107827) for deets ✨

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -262,7 +262,7 @@ class Client < ApplicationRecord
   end
 
   def forward_message_to_intercom?
-    (tax_returns.pluck(:status).map(&:to_sym) & TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES).any?
+    !online_ctc? && tax_returns.pluck(:status).map(&:to_sym).all? { |status| TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES.include? status }
   end
 
   # TODO: Replace with accurate implementation
@@ -270,9 +270,12 @@ class Client < ApplicationRecord
     current_sign_in_ip
   end
 
+  def online_ctc?
+    intake.is_ctc? && intake.tax_returns.any? { |tr| tr.service_type == "online_intake" }
+  end
+
   def hub_status_updatable
-    ctc_online_intake = intake.is_ctc? && intake.tax_returns.any? { |tr| tr.service_type == "online_intake" }
-    !ctc_online_intake
+    !online_ctc?
   end
 
   private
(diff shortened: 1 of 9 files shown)


app/models/client.rb
```diff
@@ -262,17 +262,20 @@ def request_document_help(doc_type:, help_type:)
   end
 
   def forward_message_to_intercom?
-    (tax_returns.pluck(:status).map(&:to_sym) & TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES).any?
+    !online_ctc? && tax_returns.pluck(:status).map(&:to_sym).all? { |status| TaxReturnStatus::FORWARD_TO_INTERCOM_STATUSES.include? status }
```
caitbot, line 265: per Kellys request -- we want online CTC clients to be handled by client success, and we should only forward to intercom if ALL their tax returns are in the correct statuses

app/services/intercom_service.rb
```diff
@@ -1,6 +1,23 @@
...
+    contact_id_from_client_id = contact_id_from_client_id(client.id)
+    contact_id = contact_id_from_client_id.present? ? contact_id_from_client_id : create_or_update_intercom_contact(portal_message).id
+
+    if contact_id_from_client_id.present? && most_recent_conversation(contact_id).present?
+      reply_to_existing_intercom_thread(contact_id, portal_message.body)
+    else
+      create_new_intercom_thread(contact_id, portal_message.body)
+      if inform_of_handoff
+        contact_info = client.intake.contact_info_filtered_by_preferences
+        send_handoff_email(client) if contact_info[:email]
```
caitbot, line 15: I chose to send to ALL of their opted in contact methods since there isn't really a good way to prioritize one over the other.

**What will happen when the client responds via sms or email?**
That will fortunately be handled gracefully by `create_or_update_intercom_contact` which will find them based on `client.id` and update their Intercom entry with the contact method info.

app/models/client.rb
bytheway875: I think the CS people were saying that they didn't want _any_ CTC clients to forward to intercom. can we double check on that?

app/services/intercom_service.rb
bytheway875: nice. i kind of have a feeling that there's a lot more in common re: the three message types than they have different (each thing has a field called body, and then we end up looking at client_id to find the person in intercom, then send a message to the client).

Def not blocking, but something to think about if we end up ever adding another way to contact people 😬

bytheway875 (approved):
can we double check with Kelly + CS to make sure that they only want to skip forwarding for online intake?
Screen Shot 2021-08-13 at 4 29 59 PM

(I'm worried the double negative confused things)

PR #1478, opened 2021-08-13
Move navigator things into its own section

bytheway875:
Format output navigator-relevant fields was inconsistent with the established pattern for putting data into sections on the client show page. This attempts to group the data into a navigator section.

For "driving" purposes, I also am outputting the current step so that i can see where clients get stuck in the flow.

### DIFF
--- a/app/views/hub/clients/show.html.erb
+++ b/app/views/hub/clients/show.html.erb
@@ -101,37 +101,12 @@
             <div class="field-display">
               <span class="form-question"><%= t("general.dependents") %>:</span>
               <span class="label-value"><%= @client.intake&.dependents.count %></span>
-              <ul id="dependents-list">
+              <ul id="dependents-list" class="no-bullets">
                 <%= render @client.intake.dependents %>
               </ul>
             </div>
 
           </div>
-
-          <div class="client-profile__field-group">
-            <h2 class="text--bold"><%= t(".navigator_types") %></h2>
-            <div class="field-display">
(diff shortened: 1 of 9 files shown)


app/views/hub/clients/show.html.erb
```diff
@@ -251,30 +226,56 @@
...
+            <% if @client.intake.drop_off? %>
+              <div class="field-display">
+                <span class="form-question">Photo ID Used:</span>
+                <span class="label-value"><%= @client.intake.photo_id_display_names || "N/A"%></span>
+              </div>
+              <div class="field-display">
+                <span class="form-question">Taxpayer ID Used:</span>
+                <span class="label-value"><%= @client.intake.taxpayer_id_display_names || "N/A" %></span>
+              </div>
+            <% end %>
```
jenny-heath, line 277: seems like these fields are specific to drop off _and_ ctc

bytheway875: I originally had the same feeling, but it actually breaks if it's not scoped down to only ctc dropoff clients. These are not fields that get recorded for regular drop off clients (only for CTC ones), and when they arent scoped to just ctc dropoff clients they break because the methods are written onto Intake::CtcIntake.

bytheway875: oh duh. Yes! You are correct -- I'd originally had it scoped incorrectly (which you rightly pointed out) which caused the text failures, and only responded after I'd updated thinking you were suggesting the opposite.

PR #1479, opened 2021-08-13
Add sign out route to ctc routes

jenny-heath:
do i understand why these routes are for _user_ sessions even though what's signed in is a _client_? not exactly

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -328,6 +328,10 @@ Rails.application.routes.draw do
           resources :documents, only: [:show]
         end
 
+        devise_for :users, path: "hub", controllers: {
+          sessions: "users/sessions",
+        }
+
         # Any other top level slash just goes to home as a source parameter
         get "/:source" => "ctc_pages#source_routing", constraints: { source: /[0-9a-zA-Z_-]{1,100}/ }
       end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -328,6 +328,10 @@ Rails.application.routes.draw do
           resources :documents, only: [:show]
         end
 
+        # Devise logs out all logged in records, regardless of what model type they are
+        # It also generally wants a default Rails model to connect its routes to, which for us is :users
+        # These routes are used for logging in/out clients even though the paths are created for :users
+        # TODO: refactor to make more explicit session routes for clients
         devise_for :users, path: "hub", controllers: {
           sessions: "users/sessions",
         }
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,11 @@
 Rails.application.routes.draw do
+
+  devise_for :clients
+
+  devise_scope :client do
+    delete 'clients/sign_out', to: 'devise/sessions#destroy', as: :destroy_client_session
+  end
+
   def login_routes
     resources :client_logins, path: "login", only: [:new, :create, :edit, :update], path_names: { new: '', edit: '' } do
       get "locked", to: "client_logins#account_locked", as: :account_locked, on: :collection
@@ -118,12 +125,6 @@ Rails.application.routes.draw do
       get "/consent-to-disclose", to: "consent_pages#consent_to_disclose"
       get "/relational-efin", to: "consent_pages#relational_efin"
       get "/global-carryforward", to: "consent_pages#global_carryforward"
-
-      devise_for :client, skip: [:sessions]
-
-      devise_scope :client do
-        delete 'clients/sign_out', to: 'devise/sessions#destroy', as: :destroy_client_session
-      end
 
       namespace :portal do
         root "portal#home"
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -328,6 +328,10 @@ def scoped_navigation_routes(context, navigation)
           resources :documents, only: [:show]
         end
 
+        devise_for :users, path: "hub", controllers: {
```
bytheway875, line 331: You're right that users are in the hub (and you can even see that this is creating a namespaced hub path within the ctc area, which is also somewhat unexpected) and we're trying to logout clients, so i think we'll want to mount `devise_for :clients` instead.

Line 122 of the routes file shows how we're mounting the devise engine for gyr login/logout routes.

We might need to look at the logout logic in general -- it's possible that across the app it's configured through the routes for user which would be quite odd to me. 

Although it looks like line 122 specfiically routes everything but the sessions functionality, so potentially this is by design. Do you know anything about that @alaroia-cfa

bytheway875 (changes requested):

alaroia-cfa:
My recollection is that I read the Devise code, and logout logs out all
authenticated Devise objects, be they users or clients.

Users were the first to be able to be logged-in, so they were the first to
get logout. When we added clients, we noticed we didn't need a new logout.

Devise generally wants a Rails model to connect its routes to. I believe
that's configured here:

> config.default_scope = :user

in config/initializers/devise.rb

That's the other reason it's :users, I think.

bytheway875:
@alaroia-cfa That feels entirely unexpected and really strange to use hub/users logout path to log out clients. If we really want to share logout paths, can I propose that we create something generic? Something like /sessions/logout that kills all active sessions?

In the meantime, if this is actually the intended way to do this, we can potentially merge to fix the bug and file an improvement for later. If we do go that route, can we make sure that there is a comment somewhere that explains why there are hub paths in the ctc namespace?

jenny-heath:
@bytheway875 ok we left a comment by the route and I'll make a chore to revisit this. Lmk if this is good to merge!

Slightly unrelated Q: do you know why logging in was working for clients even though there were no `devise_for` routes in the ctc domain? (or at least it looks like there weren't)

bytheway875:
@jenny devise_for :clients is mounting the devise engine and likely enables writing routes for a whole suite of things, like default login forms. We don't use those default forms or behavior and write our own login routes. Those are included in ctc section.

I _really_ hope we prioritize fixing this other problem that writes a hub path into CTC. To me, at the very least, we should fix the part that requires posting to a /hub/ namespaced route on a subdomain that includes no other code about hub concerns as part of this bug ticket.

spec/features/ctc/complete_intake_spec.rb
```diff
@@ -11,7 +11,6 @@ def strip_html_tags(text)
 
   before do
     allow_any_instance_of(Routes::CtcDomain).to receive(:matches?).and_return(true)
-    allow_any_instance_of(Ctc::Questions::LegalConsentController)
```
caitbot, line 14: This is just unrelated cleanup (:

alaroia-cfa:
> @bytheway875 @alaroia-cfa would the two of you mind checking our approach to creating distinct routes for client sign out? It feels like a very important change and we just mucked our way through many attempts at getting the routes to be valid.

What I would do if possible is figure out a way to see what routes are mapped into the CTC domain. I'll see tomorrow if I can figure out how to do that!

config/routes.rb
```diff
@@ -119,7 +119,12 @@ def scoped_navigation_routes(context, navigation)
       get "/relational-efin", to: "consent_pages#relational_efin"
       get "/global-carryforward", to: "consent_pages#global_carryforward"
 
-      devise_for :clients, skip: [:sessions]
```
bytheway875, line 122: `devise_for :client`, in addition to (sometimes) writing routes, primary mounts devise as an engine (provides current_client functionality, etc). This writing of the helpers (and whatever else the engine is doing) seems to only need to happen once, not once per subdomain (which is why current_client was working even when we didn’t have devise_for defined under the :ctc subdomain.) 

In fact, I think we can continue to define devise_for :clients just once, and potentially even move it to the top of the routes file so that it’s clear it doesn’t “belong” solely to either subdomain.

If you check the docs for devise_for, it does also add some routes sometimes:
https://www.rubydoc.info/github/plataformatec/devise/ActionDispatch%2FRouting%2FMapper:devise_for

However, it only adds routes when you’re using certain supported modules. For example, it adds the sessions routes when :authenticatable is a devise module defined on Client. Since we don’t use that module, we can remove `skip: :sessions` since those routes won’t be written anyway. (You can test this out by toggling adding/removing `skip :sessions` and noticing that there are no additional routes written when removing.

See the difference between the devise modules used on the two Devise-powered classes: Client line 44 vs User line 51

bytheway875: I also see that we changed devise_for to use the singular :client instead of plural :clients. That makes logical sense to me given that devise_scope needs the singular :client, but while snooping around the docs (clicking on devise_for in Rubymine), I found this:

# Also be aware of that 'devise_scope' and 'as' use the singular form of the
# noun where other devise route commands expect the plural form. 

It’s possible that changing devise_for back to plural :clients could maybe save some random pain in the future, it is also possible that it will never affect us whatsoever!

PR #1481, opened 2021-08-16
Restrict UserAgent in XML to 150 characters to conform with schema

bytheway875:
A submission failed with this error, so we need to fix it. 

This simply trims the user_agent value to 150 characters when writing it into the XML.

(It likely provides the IRS with an invalid user_agent, but that is the fault of their validation TBH)


bytheway875:
@alaroia-cfa verbally approved over slack.

PR #1484, opened 2021-08-16
Add phone number to CTC client Update/New form

caitbot:
We send it to the IRS and ask it on the legal-consent page, so it should be edit-table on the form.

It fixes this issue (https://www.pivotaltracker.com/story/show/179238903)


bytheway875 (commented):
yay!

bytheway875 (approved):

PR #1496, opened 2021-08-17
Add Percy (#177084353)

coltborg:
See comments below


bengolder (commented):
This overall looks great!

Only blocking feedback is to either add documentation saying one screenshot: true spec per file or adjusting script to be able to identify multiple screenshots of the same page in the same file.

Other feedback that is non-blocking:
- it might be easier for team members to run the script if the secret could be added to the codebase (for example, by using the encrypted credentials file). Colt noted that this might require the script to access the rails environment in order to decrypt the credentials.
- The documentation is really helpful!
- The script for running percy is simple and clear, really appreciate that! Emojis are a delightful touch!
- I'm glad you included coverage for the home page since it's not included in the new_joint_filers spec.
- Are there other pages or specs that feel important for checking honeycrisp upgrades? Perhaps the hub?

coltborg:
We made some improvements:

- Add `PERCY_TOKEN` to `development.yml.enc`
- Update README to add an example to add screenshots to a feature spec
- Add index number to screenshots so that multiple of a single page can be taken
- Add screenshots to a small Hub feature spec

spec/support/helpers/feature_helpers.rb
```diff
@@ -5,8 +5,11 @@ def screenshot_after
     yield
 
     if @metadata_screenshot
+      @screenshot_index = defined?(@screenshot_index) ? @screenshot_index + 1 : 1
```
coltborg, line 8: This approach was taken from how [GCF numbers their screenshots](https://github.com/codeforamerica/gcf-backend/blob/4a41dbcd6af253bb3ea42474afa1eadd667e4ec4/spec/support/system_test_helpers.rb#L167).

spec/support/helpers/feature_helpers.rb
```diff
@@ -5,8 +5,11 @@ def screenshot_after
...
       example_text, spec_path = inspect.match(/"(.*)" \(\.\/spec\/features\/(.*)_spec\.rb/)[1, 2]
-      page.percy_snapshot("#{spec_path}/#{example_text.parameterize}#{current_path}")
+
+      screenshot_name = "#{format('%02d', @screenshot_index)}|#{spec_path}|#{example_text.parameterize}|#{current_path.parameterize}"
```
coltborg, line 11: Changed how the naming is laid out to be more readable, for example:

```
37|web_intake/new_joint_filers|new-client-filing-joint-taxes-with-spouse-and-dependents|en-dependents

```

coltborg:
# Goal of this PR

This PR adds Percy to GYR so that a developer can trigger a comparison build of screenshots between `main` and a new branch. Here will be the link to [vita-min in Percy](https://percy.io/Code-for-America/vita-min).

# What screenshot coverage is there in GYR?

- the home page
- every page in the `new_joint_filers_spec.rb` flow
- a few pages in the Hub covered by `edit_user_spec.rb`

# Once this is merged into main

After this PR is merged into `main`, we'll be able to run comparative screenshots for the honeycrisp update PR (#1496)

README.md
```diff
@@ -164,6 +164,35 @@ CHROME=y rspec
 COVERAGE=y rspec
 ```
 
+#### Percy visual diff tests
```
alaroia-cfa, line 167: Maybe this should move into a `doc/` file so that `README.md` is a little shorter. Not a blocker.

It can be without a datestamp since it's currently evergreen. Maybe `doc/visual-diff-testing.md`?

spec/support/helpers/feature_helpers.rb
alaroia-cfa: I would rather a different character to split these, not `|` which is used by bash, so it would require more special encoding.

I would suggest:

```
#{spec_path.parameterize}/#{example_text.parameterize}/#{format('%02d', @screenshot_index)}-#{current_path.parameterize}`

It seems to me there's a natural grouping - spec path first, then example name, then screenshot index (with current path being an optional suffix to screenshot index).

I think that it's best to avoid unnecessary `/` characters and to avoid `|`, but I don't know if these end up as filenames on disk, or what.

100% not a blocker.

alaroia-cfa (approved):
Thanks! This is a simple enough change and I'm happy to see it land. I'm merging now to avoid delay; if you want to make further changes, feel free to make a new PR!

PR #1507, opened 2021-08-18
Hub errors page: allow individual errors to be flagged as "Auto-cancel?"

tgrathwell:
Also sort the errors on that page by source then code

### DIFF
--- a/spec/features/hub/efile_errors_spec.rb
+++ b/spec/features/hub/efile_errors_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe "efile errors" do
+  context "as an authenticated user" do
+    context "as an admin" do
+      let(:current_user) { create :admin_user }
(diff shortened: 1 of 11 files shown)


spec/features/hub/efile_errors_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      describe 'editing an error' do
+        it 'allows an error to be classified as auto cancellable' do
+          visit hub_efile_errors_path
+          click_on efile_error.code
+          click_on I18n.t('general.edit')
+          check "Auto-cancel?"
+          click_on I18n.t('general.save')
+
+          expect(efile_error.reload).to be_auto_cancel
```
alaroia-cfa, line 19: Should we add sending/receiving data from the IRS within this feature spec? I think no, but it's a little odd standing alone like this.

bytheway875: this seems fine to me - theres no interaction with the IRS here. The only thing this does is change state on the internal version of our error.

PR #1511, opened 2021-08-18
CTC intake: "Add new dependent" button creates dependent in db instead of using :new path

tgrathwell:
Having a URL with :new in it caused trouble because when clients used the browser back
button it would create a new dependent instead of editing the one they just created.

This relaxes the in-app validations on Dependent such that a dependent without
first_name/last_name/relationship/birth_date is allowed in the DB. We should not
consider such a dependent "valid" in most senses of the word (it shouldn't show
on the confirm page or go out in someone's tax XML) but I'm sure it will cause
us some grief down the line nonetheless.


tgrathwell:
closing in favor of https://github.com/codeforamerica/vita-min/pull/1515

PR #1522, opened 2021-08-19
Capture IRS SDK audit_log.txt on gyr-efiler failure [#179291421]

alaroia-cfa:


app/services/efile/gyr_efiler_service.rb
```diff
@@ -1,44 +1,36 @@
...
-      Dir.mktmpdir do |working_directory|
-        FileUtils.mkdir_p(File.join(working_directory, "output", "log"))
-        ensure_config_dir_prepared
-
-        # TODO: If the process blocks for >10 min, terminate it.
-        # TODO: Send process stdout to logs.
-        # TODO: Send output/logs/ to logs after process terminates.
-        classes_zip_path = Dir.glob(Rails.root.join("vendor", "gyr_efiler", "gyr-efiler-classes-#{CURRENT_VERSION}.zip"))[0]
-        raise StandardError.new("You must run rails setup:download_gyr_efiler") if classes_zip_path.nil?
+      ensure_efiler_has_been_downloaded
```
bytheway875, line 8: 😍

bytheway875 (approved):

app/services/efile/gyr_efiler_service.rb
```diff
@@ -1,44 +1,36 @@
...
         raise StandardError.new("Process failed to exit?") unless $?.exited?
 
         exit_code = $?.exitstatus
-        raise StandardError.new("gyr-efiler failed; exited with #{exit_code}") if exit_code != 0
+        if exit_code == 0
+          return get_single_file_from_zip(Dir.glob(File.join(efiler_working_directory, "output", "*.zip"))[0])
+        end
 
-        get_single_file_from_zip(Dir.glob(File.join(working_directory, "output", "*.zip"))[0])
+        raise StandardError.new(File.read(File.join(efiler_working_directory, 'output/audit_log.txt')))
```
bytheway875, line 33: In the file that saves this into raw response, it might be nice to output only the message instead of the whole error object like i (think) we're doing at the moment (in another file). 

However, I bias towards a release so that we go into the weekend with this issue resolved than making that change now.

alaroia-cfa:
@bytheway875 In testing on staging, I just found there's a problem with how `unzip` is now called. @jenny-heath and I are going to make an even _more_ minimal version of this, and land that ASAP instead.

PR #1523, opened 2021-08-19
Add navigators page with markdown content

jenny-heath:
we made a page that renders markdown! the goal is that non-engineers will have an easier time of updating the contents (and we can workshop a way to get these changes in with minimal but sufficient eng review)

image


bytheway875:
It seems somewhat weird to me to add one way of adding content (using ActionText) and another way of adding/maintaining content (using Markdown), instead of just converging around one approach. Could we leverage a similar pattern that we use for the error messages and use action text? That would mean no code changes at all to edit the information on the page, and less overhead in terms of methodlogy for maintaining content.

PR #1525, opened 2021-08-23
Copy change for FAQ [#179324706]

alaroia-cfa:
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3191,7 +3191,7 @@ en:
           what_will_i_need_to_submit:
             question: "I don’t usually file taxes, what will I need to submit to sign up for the Child Tax Credit?"
             answer_html: |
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -3191,7 +3191,7 @@ en:
           what_will_i_need_to_submit:
             question: "I don’t usually file taxes, what will I need to submit to sign up for the Child Tax Credit?"
             answer_html: |
-              An email address, ID, proof of income and/or benefits, and your child's Social Security Number.<br/><br/>
+              Signing up for the Child Tax Credit is simple: you only need a working email address or phone number and the social security numbers of any children you want to claim.<br/><br/>
```
bytheway875, line 3194: do we need to update spanish translation as well?

alaroia-cfa: Nicole says that it's OK to roll this out in the absence of Spanish copy for now.

PR #1530, opened 2021-08-24
Add validation for presence of an answer choice to filed 2019 form

bytheway875:


alaroia-cfa (approved):
Maybe it'd be good to make this field non-null. That's complex and perhaps not worth it.

Anyway, 🚢  looks good!

PR #1531, opened 2021-08-24
Fix agi bug -- allow commas in money fields.

bytheway875:


tgrathwell:
ugh what an embarrassing bug

The stripping of the commas at type cast time is good, but is the decimal migration adding anything here? Are we sure the migration updates existing values successfully?

I expect AGI is "supposed" to be a whole dollar amount but I guess this would handle the case where someone enters 1234.00; though currently it just drops the fractional part which also seems fine

tgrathwell:
the `MoneyType` was my attempt to carry forward knowledge from GCF but somehow I completely missed the comma part which now makes me feel quite the fool

Screen Shot 2021-08-24 at 11 54 52 AM

bytheway875:
bleh, i was supposed to have reverted the change to decimal but i dont think i did.

spec/features/ctc/complete_intake_spec.rb
```diff
@@ -33,13 +33,15 @@ def strip_html_tags(text)
     expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.filed2020.title'))
     click_on I18n.t('general.negative')
     expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.filed2019.title'))
-    choose I18n.t('views.ctc.questions.filed2019.did_not_file')
+    choose I18n.t('views.ctc.questions.filed2019.filed_full')
```
tgrathwell, line 36: there's a separate "client who has filed in 2019" scenario in this file that would be a better place to put this, so we have one flow for a non-filer and another test that does the AGI for the primary and spouse

bytheway875: done!

PR #1535, opened 2021-08-24
Copy updates and bug fixes for CTC clarity

bytheway875:
- Change all copy as requested in ticket
- Makes buttons full width on offboarding pages because buttons with two lines of text skeeve me out
- consistently titleizes dependent name for polish
- Fixes bug where "Add a new dependent" on dependent offboarding page used an old pattern.


caitbot (approved):
✨ beautiful ✨

PR #1547, opened 2021-08-26
CTC dependent logic & copy updates

embarnard:


app/controllers/ctc/questions/dependents/child_lived_with_you_controller.rb
```diff
@@ -8,7 +8,7 @@ class ChildLivedWithYouController < BaseDependentController
         def self.show?(dependent)
           return false unless dependent&.relationship
 
-          dependent.qualifying_child_relationship? && dependent.meets_qc_age_condition_2020? && dependent.meets_qc_misc_conditions?
+          dependent.qualifying_child_relationship? && dependent.meets_qc_age_condition_2020? && dependent.meets_qc_misc_conditions? && !dependent.born_in_last_6_months_of_2020?
```
caitbot, line 11: whether or not they are a qualifying child relies on having this question answered, will this stop valid children from having this question filled out or is the child unqualified if they were born in the last 6 months of 2020?

app/controllers/ctc/questions/dependents/child_residence_exceptions_controller.rb
```diff
@@ -9,7 +9,7 @@ class ChildResidenceExceptionsController < BaseDependentController
         def self.show?(dependent)
           return false unless dependent&.relationship
 
-          dependent.qualifying_child_relationship? && dependent.meets_qc_age_condition_2020? && dependent.meets_qc_misc_conditions? && dependent.lived_with_more_than_six_months_no?
+          dependent.qualifying_child_relationship? && dependent.meets_qc_age_condition_2020? && dependent.meets_qc_misc_conditions? && dependent.lived_with_more_than_six_months_no? && !dependent.born_in_last_6_months_of_2020?
```
caitbot, line 12: same with this condition

app/controllers/ctc/questions/dependents/does_not_qualify_ctc_controller.rb
```diff
@@ -8,7 +8,7 @@ class DoesNotQualifyCtcController < BaseDependentController
         def self.show?(dependent)
           return false unless dependent&.relationship
 
-          !dependent.qualifying_child_2020? && !dependent.qualifying_relative_2020?
+          (!dependent.qualifying_child_2020? && !dependent.qualifying_relative_2020?) || dependent.birth_date.year == 2021
```
caitbot, line 11: we could instead add this condition to the qualifying_child_2020 and qualifying_relative_2020 methods

app/controllers/ctc/questions/dependents/qualifying_relative_controller.rb
```diff
@@ -9,8 +9,8 @@ def self.show?(dependent)
           return false unless dependent&.relationship
 
           # For family members like uncle, we need to show the qualifying relative page.
-          # For children, there are special rules that could make them not a qualifying relative, in which case we wouldn't show this page.
-          dependent.qualifying_relative_relationship? || dependent.disqualified_child_qualified_relative?
+          # For children, we show the page if they do not meet the age conditions
+          dependent.qualifying_relative_relationship? || (dependent.qualifying_child_relationship? && !dependent.meets_qc_age_condition_2020?)
```
caitbot, line 13: Should we be using `disqualified_child_qualified_relative` here? I'm worried that only checking that they do not meet age conditions might be oversimplifying it

app/forms/ctc/dependents/info_form.rb
```diff
@@ -51,6 +51,8 @@ def save
           birth_date: birth_date
         ))
         @dependent.save
+
+        @dependent.update!(lived_with_more_than_six_months: "yes") if @dependent.born_in_last_6_months_of_2020?
```
caitbot, line 55: ah I see, this answers my previous questions, so we answer the questions ahead of time.

app/controllers/ctc/questions/dependents/child_lived_with_you_controller.rb
caitbot: I see now that we do it in the `info form`

app/controllers/ctc/questions/dependents/child_residence_exceptions_controller.rb
caitbot: I see now that we do it in the `info form`

app/controllers/ctc/questions/dependents/does_not_qualify_ctc_controller.rb
embarnard: yeah you're right, I think that would make more sense!

app/controllers/ctc/questions/dependents/qualifying_relative_controller.rb
embarnard: dependents that filed a joint return should just be disqualified now so we should remove the `disqualified_child_qualified_relative` method altogether now

embarnard: previously dependents that were disqualified for filing jointly were still qualified relatives but now they are just offboarded

caitbot: sweeeet! makes sense, I appreciate the explanation!

caitbot (approved):

PR #1559, opened 2021-08-27
When IRS has an outage, calmly retry for up to 24 hours

alaroia-cfa:
Requesting design feedback

### DIFF
--- a/app/models/efile_submission.rb
+++ b/app/models/efile_submission.rb
@@ -105,6 +105,26 @@ class EfileSubmission < ApplicationRecord
     )
   end
 
+  def retry_send_submission
+    now = DateTime.now.utc
+    age = now.to_time - created_at.to_time
+    if age > 1.days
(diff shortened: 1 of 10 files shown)


app/models/efile_submission.rb
```diff
@@ -105,6 +105,26 @@ def generate_form_1040_pdf
     )
   end
 
+  def retry_send_submission
+    now = DateTime.now.utc
```
bytheway875, line 109: Could add `.last_retried_at`

spec/models/efile_submission_spec.rb
```diff
@@ -364,7 +364,7 @@
     end
   end
 
-  describe "imperfect_return_resubmission?" do
+  describe ".imperfect_return_resubmission?" do
```
bytheway875, line 367: should be #

bytheway875 (approved):

PR #1562, opened 2021-08-27
Replace &nbsp; with empty div

jenny-heath:
the profiler strikes again! i think we didn't notice there was a visible "\&nbsp;" because it was blocked

this is what demo looks like:
image

with this change it looks like:
image

but the empty div feels janky, please lmk if you have other ideas

### DIFF
--- a/app/views/public_pages/maintenance.html.erb
+++ b/app/views/public_pages/maintenance.html.erb
@@ -1,6 +1,6 @@
 <% content_for :page_title, t("views.ctc.questions.at_capacity.title") %>
 <% content_for :header do %>
-  <div></div>
+  <%= render 'shared/header', title: "GetCTC", home_link: ctc_root_path %>
 <% end %>
 <% content_for :footer do %>
   <footer class="footer">
@@ -18,7 +18,6 @@
       </div>
     </div>
   </footer>
-
 <% end %>
 
 <main role="main">
--- a/app/views/public_pages/maintenance.html.erb
+++ b/app/views/public_pages/maintenance.html.erb
@@ -1,5 +1,7 @@
 <% content_for :page_title, t("views.ctc.questions.at_capacity.title") %>
-<% content_for :header, "&nbsp;" %>
+<% content_for :header do %>
+  <div></div>
+<% end %>
 <% content_for :footer do %>
   <footer class="footer">
     <div class="slab slab--footer">


app/views/public_pages/maintenance.html.erb
```diff
@@ -1,5 +1,7 @@
 <% content_for :page_title, t("views.ctc.questions.at_capacity.title") %>
-<% content_for :header, "&nbsp;" %>
```
alaroia-cfa, line 2: Silly question - since a nbsp is a non-breaking space, can we use a regular space? If so:

```
<% content_for :header, " " %>
```

If we don't care about any contents, can we make it empty, like:

```
<% content_for :header do %>
<% end %>
```

Or even:

```
<% content_for :header, "" %>
```

...in fact, why do we even bother making the header at all? Can we remove the line entirely?

What you have is OK, too.

alaroia-cfa (approved):
What you have is fine; I suggested some other options.

app/views/public_pages/maintenance.html.erb
jenny-heath: the reason we bother with the content_for header is that we are trying to hide the default header because it has branding for getctc or getyourrefund and this is supposed to be a generic maintenance page for both. (in application.html.erb, it does a check for content_for and otherwise renders the GYR header)

i tried all of those options and it turns out that none of them successfully replace the header. i expected it for the empty block because i saw something on the internet about that but i'm mildly surprised that the regular space (" ") didn't do it 🤷🏻‍♀️

jenny-heath: i realized after typing that that in the absence of a header there's no way to translate the page into spanish, so i talked to kelly and we're gonna use a ctc header instead

jenny-heath:
new header who dis
image

app/views/public_pages/maintenance.html.erb
alaroia-cfa:
```
<% content_for :header do %>
&nbsp;
<% end %>
```

Also OK!

alaroia-cfa: Space probably doesn't work because it's an ERB/Rails thing, and it's probably using `.present?`, and

```
irb(main):001:0> ' '.present?
=> false
```

alaroia-cfa:
This version looks good too! Have you tested how people actually get to see this page? If you are expecting to show it when the service is down, as Aptible's maintenance page, then I think we have to pick between getctc or getyourrefund's maintenance page, since Aptible will only cache 1 URL per app: https://deploy-docs.aptible.com/docs/maintenance-page

I don't know the full context, but wanted to make sure to at least mention that.

jenny-heath:
@alaroia-cfa true! we do have to pick one. the context for the fix in this PR is [this commit](https://github.com/codeforamerica/vita-min/commit/1a1ac412be39a4809ce51d4f0446983d7c7ddfe8), in which we replace the content for the main gyr maintenance page with more generic content. the slightly bizarre part of _this_ PR is that https://www.getyourrefund.org/maintenance (which is what `MAINTENANCE_PAGE_URL` is set to in the aptible config) will show a getCTC header. this was a decision kelly and i made based on the fact that we expect more people to be using getCTC (and that the purpose of adding this more generic content was to look less like we messed up if getCTC goes down due to an influx of traffic from one outreach initiative or other)

PR #1563, opened 2021-08-30
Add 2019 Signature PIN to hub, and allow Return Header XML to use it

bytheway875:


app/lib/submission_builder/return_header1040.rb
```diff
@@ -39,8 +39,18 @@ def document
...
+            if intake.primary_prior_year_signature_pin.present?
+              xml.PrimaryPriorYearPIN intake.primary_prior_year_signature_pin
+            else
+              xml.PrimaryPriorYearAGIAmt intake.primary_prior_year_agi_amount || 0
+            end
+            if tax_return.filing_jointly?
+              if intake.spouse_prior_year_signature_pin.present?
+                xml.SpousePriorYearPIN intake.spouse_prior_year_signature_pin
+              else
+                xml.SpousePriorYearAGIAmt spouse_prior_year_agi(intake)
```
embarnard, line 51: should this be `xml.SpousePriorYearAGIAmt spouse_prior_year_agi(intake) || 0`

embarnard: Oh nevermind, I see the method below

app/lib/submission_builder/return_header1040.rb
```diff
@@ -193,19 +203,16 @@ def spouse_name_control(intake)
 
     private
 
-    def spouse_prior_year_agi(tax_return, intake)
-      return nil unless tax_return.filing_jointly?
+    def spouse_prior_year_agi(intake)
```
bytheway875, line 206: because of the fact that this method will now always add content if they're married filing jointly, we needed the "fall through" to return 0, so I refactored the order of the if statements.

app/forms/hub/update_ctc_client_form.rb
```diff
@@ -5,7 +5,10 @@ class UpdateCtcClientForm < ClientForm
                        :primary_first_name,
                        :primary_last_name,
                        :primary_suffix,
+                       :primary_prior_year_agi_amount,
```
embarnard, line 8: Should this form validate to either require the AGI amount or the PIN if filed 2019?

embarnard (approved):
Looks good! No blocking comments

app/forms/hub/update_ctc_client_form.rb
bytheway875: no, because its mostly used by customer service and they can just toggle based on what they need to do without breaking the form and needing to toggle whether the person filed in 2019. Removing all of the data from the form will result in us submitting 0 values, which is the same as the client telling us they did not file.

PR #1566, opened 2021-08-30
Confirm information edit

bytheway875:


app/controllers/ctc/update/base_intake_revision_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Ctc::Update::BaseIntakeRevisionController < Ctc::Portal::BaseIntakeRevisionController
```
caitbot, line 1: niiiice

app/controllers/ctc/update/dependents_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    render 'ctc/portal/dependents/confirm_remove'
+  end
+
+  def destroy
+    current_model.destroy!
+    flash[:alert] = "Family member / dependent removed from your tax return."
+    redirect_to questions_confirm_information_path
+  end
+
+  def update
```
caitbot, line 12: Could we have the other dependents controller inherit from `BaseRevisions` and then we wouldn't need this?

spec/controllers/ctc/questions/confirm_information_controller_spec.rb
```diff
@@ -92,27 +91,30 @@
         end
 
         context "without dependents" do
-          it "does not show dependents section" do
+          before do
+            # intake.dependents.destroy_all
```
caitbot, line 95: do we need this since it is commented out?

caitbot (approved):
looks good, left some questions/comments

app/controllers/ctc/update/dependents_controller.rb
bytheway875: i think we'd still have to do this because the base revisions controller expects to create a note and a couple other things, which we're bypassing here. I could have this controller inherit from baserevisions and then pull in the stuff that makes the dependents controller unique here instead, but this ended up being the more straightforward change.

PR #1567, opened 2021-08-30
Put launch button behind switch and add a banner

caitbot:
image
image


config/locales/en.yml
```diff
@@ -3303,8 +3303,10 @@ en:
...
             <br/>
             You can receive  payments of up to $300 per child every month starting in July 2021 through December 2021. You will need to file a tax return in 2022 to receive the rest of the payment.
-        why:
-          body_html: <strong>GetCTC is a <a rel="noopener nofollow" target="_blank" href="https://www.codeforamerica.org/">Code for America</a> product.</strong> Learn more about why we are doing this work <a rel="noopener nofollow" target="_blank" href="https://www.codeforamerica.org/news/ensuring-families-who-qualify-for-the-child-tax-credit-arent-left-behind/">here</a>.
+        affiliations:
+          cfa_product_html: GetCTC is a <a rel="noopener nofollow" target="_blank" href="https://www.codeforamerica.org/">Code for America</a> product.</br><a rel="noopener nofollow" target="_blank" href="https://www.codeforamerica.org/news/ensuring-families-who-qualify-for-the-child-tax-credit-arent-left-behind/">Why do we do this work?</a>
+          white_house: Supported by the White House and U.S. Department of Treasury.
```
caitbot, line 3308: I still need the link for this!

app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -18,10 +18,34 @@
...
   }
 
-  .slab--why {
+  .slab--affiliations {
     background-color: $color-grey-light;
   }
 
+  .affiliations {
+    display: flex;
+    justify-content: space-around;
```
bytheway875, line 27: what happens if you do space between here instead of space around? If the grid is restricted to a max-width already, you should get the outer two to align with edges a bit better.

app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -18,10 +18,34 @@
...
+    align-items: center;
+
+    .grid__item {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+    }
+    .grid__item.text {
+      width: 200px;
```
bytheway875, line 36: what happens if you pull this up to like 225? I noticed that the text under the first image has two lines with only one word each on it which looks a little weird. If each text section had a bit more width, we might be able to avoid orphaned words.

app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -18,10 +18,34 @@
     padding-bottom: 0;
   }
 
-  .slab--why {
+  .slab--affiliations {
     background-color: $color-grey-light;
   }
 
+  .affiliations {
```
bytheway875, line 25: What happens to this section on mobile? Do the sections stack on top of one another? I can't tell if there is something that changes the flex direction on a parent class or something.

bytheway875: oh, now that im looking at how its implemented, maybe the width one third class handles that?

app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -18,10 +18,34 @@
...
+    .grid__item {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+    }
+    .grid__item.text {
+      width: 200px;
+      text-align: center;
+    }
+    .grid__item.image {
```
bytheway875, line 39: if you make this a flex item, you can center the image tag within the div so that hopefully all of the images will look more balanced on the page -- it might pull down the code for america logo so that it aligns better with the center of the white house image.

bytheway875 (approved):
just some questions around the CSS.

If you've made sure that the new section collapses properly on mobile, i think this is good to go.

bytheway875:
Also, the "file your simplified return now" button text is significantly longer than the old button text -- can you check to see what that button looks like on mobile? we may need to reposition the image if they intersect.

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: Good point, I'll try raising it. For some reason, `product` always ends up by itself even when I raise it a LOT.

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: This happens :( 
image

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: So I already have them centered in the parent div. Putting an align-center on this div doesn't seem to do anything ):

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: It seems to be related to it being a link ):

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: It looks like this 
image

caitbot: I got it to work with a grid 
image

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: I made it look better using a grid.

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: Fixed it.

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: I just ended up using manually specified top padding.

PR #1569, opened 2021-08-31
Hub: ClientChange system notes now use the new table formatting from CTC portal notes

tgrathwell:
old ClientChange notes that baked all their presentation into the 'body'
attribute will still display how they used to

Also another commit is on here that tries to clean up the extra cruft you get when you edit an Intake::CtcIntake in the hub for the first time. Check it out!!

## bad

Screen Shot 2021-08-31 at 1 52 10 AM

## good ?

Screen Shot 2021-08-31 at 1 53 18 AM

### DIFF
--- a/app/forms/hub/update_client_form.rb
+++ b/app/forms/hub/update_client_form.rb
@@ -42,8 +42,6 @@ module Hub
     def initialize(client, params = {})
       @client = client
       super(params)
-      # parent Form class creates setters for each attribute -- won't work til super is called!
-      self.preferred_name = preferred_name.presence || "#{primary_first_name} #{primary_last_name}"
     end
 
     def self.existing_attributes(intake)
--- a/app/models/system_note/client_change.rb
+++ b/app/models/system_note/client_change.rb
@@ -22,23 +22,17 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class SystemNote::ClientChange < SystemNote
-  def self.generate!(intake:, initiated_by: )
-    return unless intake.saved_changes.present?
+  def self.generate!(original_intake:, intake:, initiated_by: )
+    interesting_changes = InterestingChangeArbiter.determine_changes(original_intake, intake)
+    return unless interesting_changes.present?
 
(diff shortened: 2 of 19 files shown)


app/models/system_note/client_change.rb
```diff
@@ -22,23 +22,17 @@
...
-    return unless intake.saved_changes.present?
+  def self.generate!(original_intake:, intake:, initiated_by: )
+    interesting_changes = InterestingChangeArbiter.determine_changes(original_intake, intake)
+    return unless interesting_changes.present?
 
-    changes_list = ""
-    intake.saved_changes.each do |k, v|
-      next if k == "updated_at"
-      next if k == "needs_to_flush_searchable_data_set_at"
-      next if k.include?("encrypted")
```
caitbot, line 32: It looks like we just skipped `encrypted` values for clients before, but now we should list them, niiice.

app/forms/hub/update_client_form.rb
```diff
@@ -42,8 +42,6 @@ class UpdateClientForm < ClientForm
     def initialize(client, params = {})
       @client = client
       super(params)
-      # parent Form class creates setters for each attribute -- won't work til super is called!
-      self.preferred_name = preferred_name.presence || "#{primary_first_name} #{primary_last_name}"
```
caitbot, line 46: Why remove this?

spec/support/feature_test_helpers.rb
```diff
@@ -0,0 +1,20 @@
+module FeatureTestHelpers
```
caitbot, line 1: nice

caitbot (approved):
looks good to me!

app/forms/hub/update_client_form.rb
caitbot: Ah upon further investigation, it looks like we do this in a `before validation` on the base form `ClientForm`

caitbot: And I just saw your commit message, hahahah good change

PR #1571, opened 2021-08-31
Allow clients to resubmit their own returns

bytheway875:
I moved security information entirely to client because passing the efile submission stuff through onto the efile submission, tracking whether its changed, etc. It simplifies things in a lot of ways and allows us to just grab first and last on the submission.

### DIFF
--- a/app/controllers/ctc/portal/portal_controller.rb
+++ b/app/controllers/ctc/portal/portal_controller.rb
@@ -16,15 +16,18 @@ class Ctc::Portal::PortalController < Ctc::Portal::BaseAuthenticatedController
     end
   end
 
-  def edit_info; end
+  def edit_info
+    @current_submission = current_intake.client.efile_submissions.last
+  end
 
   def resubmit
     @submission = current_client.efile_submissions.last
-    if @submission.can_transition_to?(:ready_to_resubmit)
-      @submission.transition_to(:ready_to_resubmit)
+    if @submission.can_transition_to?(:resubmitted)
+      current_client.efile_security_informations.create(params.require(:ctc_resubmit_form).permit!)
+      @submission.transition_to(:resubmitted)
       SystemNote::CtcPortalAction.generate!(
         model: @submission,
-        action: 'ready_to_resubmit',
+        action: 'resubmitted',
         client: current_client
       )
     end
--- a/app/lib/submission_builder/return_header1040.rb
+++ b/app/lib/submission_builder/return_header1040.rb
@@ -16,8 +16,8 @@ module SubmissionBuilder
       tax_return = submission.tax_return
       intake = submission.intake
       client = submission.client
-      creation_security_information = client.efile_security_information
-      filing_security_information = submission.efile_security_information
+      creation_security_information = client.efile_security_informations.first
+      filing_security_information = client.efile_security_informations.last
       address = submission.address
 
       Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
--- a/app/models/efile_security_information.rb
+++ b/app/models/efile_security_information.rb
@@ -26,8 +26,7 @@
 #  fk_rails_...  (efile_submission_id => efile_submissions.id)
 #
 class EfileSecurityInformation < ApplicationRecord
-  belongs_to :client, optional: true
-  belongs_to :efile_submission, optional: true
+  belongs_to :client
 
   # storing client_system_time as a string and then transforming it into DateTime for the return_header1040
   # b/c the db would record the date in UTC and we would lose the client's timezone
--- a/app/views/ctc/portal/portal/edit_info.html.erb
+++ b/app/views/ctc/portal/portal/edit_info.html.erb
@@ -9,7 +9,7 @@
       <%= link_to t("general.back"), :back %>
     </div>
     <div class="question-wrapper">
-      <%= form_with model: @form, url: ctc_portal_resubmit_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+      <%= form_with url: ctc_portal_resubmit_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
 
         <h1 class="h2"><%= @main_question %></h1>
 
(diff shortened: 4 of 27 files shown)


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -16,15 +16,18 @@ def home
     end
   end
 
-  def edit_info; end
+  def edit_info
+    @current_submission = current_intake.client.efile_submissions.last
+  end
 
   def resubmit
     @submission = current_client.efile_submissions.last
```
alaroia-cfa, line 24: In theory, there's a race condition here. A new EfileSubmission could be created between when the page is loaded vs. when the button is clicked.

Even though it's pedantic, I'd personally prefer if we sent the efile_submission ID to the server. Then we can detect this race and show an error message or something.

Not a blocker.

app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -16,15 +16,18 @@ def home
...
-    if @submission.can_transition_to?(:ready_to_resubmit)
-      @submission.transition_to(:ready_to_resubmit)
+    if @submission.can_transition_to?(:resubmitted)
+      current_client.efile_security_informations.create(params.require(:ctc_resubmit_form).permit!)
+      @submission.transition_to(:resubmitted)
```
alaroia-cfa, line 27: Should this be `!`? In the code as written, we call `.transition_to` which can return false in the case of an error (transition was impossible), but we're ignoring that error, which worries me.

app/lib/submission_builder/return_header1040.rb
```diff
@@ -16,8 +16,8 @@ def document
       tax_return = submission.tax_return
       intake = submission.intake
       client = submission.client
-      creation_security_information = client.efile_security_information
-      filing_security_information = submission.efile_security_information
+      creation_security_information = client.efile_security_informations.first
+      filing_security_information = client.efile_security_informations.last
```
alaroia-cfa, line 20: I worry about a scenario where there is just one EfileSecurityInformation for whatever reason - will that occur? If so, then this code will behave incorrectly, and our schema won't support us in identifying the incorrect behavior. I'm a person who's scared to live without strong support from my database schema!

Having said that, the scenario I'm worried about will not occur. The end-of-flow (Confirm Legal) form validates that it won't occur. 

So I can live with this.

app/models/efile_security_information.rb
```diff
@@ -26,8 +26,7 @@
 #  fk_rails_...  (efile_submission_id => efile_submissions.id)
 #
 class EfileSecurityInformation < ApplicationRecord
-  belongs_to :client, optional: true
-  belongs_to :efile_submission, optional: true
+  belongs_to :client
```
alaroia-cfa, line 29: I see the appeal of a non-optional `belongs_to`!

app/views/ctc/portal/portal/edit_info.html.erb
```diff
@@ -9,7 +9,7 @@
       <%= link_to t("general.back"), :back %>
     </div>
     <div class="question-wrapper">
-      <%= form_with model: @form, url: ctc_portal_resubmit_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+      <%= form_with url: ctc_portal_resubmit_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
```
alaroia-cfa, line 12: It looks like `@form` was nil, so `model: @form` was doing nothing. Thanks for deleting it!

spec/features/ctc/portal_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.feature "CTC Intake", active_job: true do
+RSpec.feature "CTC Intake", :js, :active_job do
```
alaroia-cfa, line 3: If you don't want `:js` here (`:js` can introduce slowness or flakiness), you could do something like:

```
RSpec.feature "CTC Intake", :efile_security, :active_job do
```

and then change `spec/rails_helper.rb` to be something like

```
diff --git a/spec/rails_helper.rb b/spec/rails_helper.rb
index 15fafcd5a..76d2167de 100644
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -138,7 +138,7 @@ RSpec.configure do |config|
       Capybara.page.current_window.resize_to(2000, 4000)
     end

-    unless Capybara.current_driver == Capybara.javascript_driver
+    if example.metadata[:efile_security] || Capybara.current_driver != Capybara.javascript_driver
       allow(Rails.application.config).to receive(:efile_security_information_for_testing).and_return(
         {
           device_id: "7BA1E530D6503F380F1496A47BEB6F33E40403D1",
```

(untested; just an idea)

spec/models/efile_submission_spec.rb
```diff
@@ -27,30 +27,6 @@
     allow(address_service_double).to receive(:city).and_return "Katy"
   end
 
-  context "validation" do
```
alaroia-cfa, line 30: I'm sad to lose this validation! However, I can totally deal.

alaroia-cfa (changes requested):
I think that this is a great feature and this is a good implementation. I left some non-blocking comments.

I think if we ship this as-is, it'll change the behavior of our existing EfileSubmissions. Does it make sense to do a data migration before we release this, perhaps, where we add `.client_id` to all the `EfileSecurityInformation` records that have `.efile_submission_id` on them, so that this code can find them in the right place?

The only reason I hit "requested changes" is to discuss the migration plan. No other changes are needed. If we assess that no migration plan is needed, then I'm happy to switch to "Approve"!

app/controllers/ctc/portal/portal_controller.rb
bytheway875: it didn't really make sense to me to block the resubmission if there was a problem with the JS, especially since we only added the banner that JS was required on that one page instead of across the app.

Previously, we weren't capturing the efile information here _at all_ (it was copying from the previous submission), so this is at least a step up from that... 

I could put the bang in and then we'll see an error get raised if it does become an issue where we can't get that data and see what happens and fix it if its an issue for people?

app/lib/submission_builder/return_header1040.rb
bytheway875: According to what the IRS told us, they want the information from the first time and last time the client touched the information before sending it to us. Grabbing from first and last should be fine, and if they click on confirm legal twice that's the last time they interacted with the site before they submitted, so we should, in fact, be using that object.

bytheway875:
Good points. Let's handle most of this as a followup, and we can discuss the migration plan once it gets merged to main. We can run the migration on production likely before we even do a release. My main goal is to make sure we can get this through acceptance and get the code in a good place for production deploy.

app/controllers/ctc/portal/portal_controller.rb
alaroia-cfa: When I asked about the `!`, I meant that `.transition_to(:resubmitted)` could become `.transition_to!(:resubmitted)`. I still think it's a good idea to either call `transition_to!(...)` or handle the failed transition somehow.

You seem to be talking about the `current_client.efile_security_informations.create(...)` call. I think it's a good idea to make that a `.create!` as well. I do think it's a good idea to block resubmission if JS is not working.

> we only added the banner that JS was required on that one page instead of across the app.

I'm not sure this is true - I think that we add the JS required banner on both pages that require the EfileSecurityInformation during intake; see spec/forms/ctc/confirm_legal_form_spec.rb and spec/forms/ctc/income_form_spec.rb

> I could put the bang in and then we'll see an error get raised if it does become an issue where we can't get that data and see what happens and fix it if its an issue for people?

I'm cool with this!

I'm also cool with no change if you think it's not worth it.

app/lib/submission_builder/return_header1040.rb
alaroia-cfa: Guess so!

alaroia-cfa (approved):
Via Slack DM

> For migration plan, i was planning on going in to aptible and just running a command that wrote the client id onto all of the efile security informations

Works for me!

PR #1575, opened 2021-08-31
Make the banner and buttons better on mobile

caitbot:
**BEFORE**
image

**AFTER**
image
image
image

### DIFF
--- a/app/assets/stylesheets/templates/_ctc-home.scss
+++ b/app/assets/stylesheets/templates/_ctc-home.scss
@@ -40,6 +40,11 @@
       .image.irs {
         padding-top: 45px;
       }
+      .image.whitehouse {
+        @media screen and (max-width: $mobile-up) {
+          margin-top: 40px;
+        }
+      }
     }
   }
 
--- a/app/assets/stylesheets/templates/_ctc-home.scss
+++ b/app/assets/stylesheets/templates/_ctc-home.scss
@@ -42,14 +42,17 @@
       }
       .image.whitehouse {
         @media screen and (max-width: $mobile-up) {
-          margin-top: 40px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -40,12 +40,20 @@
...
+      .image.whitehouse {
+        @media screen and (max-width: $mobile-up) {
+          margin-top: 35px;
+        }
+      }
     }
   }
 
-  .button--nowrap {
+  :lang(en) .button--locale-specific {
```
caitbot, line 51: I didn't know you could do this, so coooool! 🤩

app/assets/stylesheets/templates/_ctc-home.scss
```diff
@@ -40,12 +40,20 @@
...
     white-space: nowrap;
   }
+  :lang(es) .button--locale-specific {
+    text-align: center;
```
embarnard, line 55: is this needed? the text looks centered in both cases

app/assets/stylesheets/templates/_ctc-home.scss
embarnard: yeah so cool! I didn't know about this either

embarnard (commented):
Woohoo we love beautifying changes! Looks good ty catie 🥳

app/assets/stylesheets/templates/_ctc-home.scss
caitbot: so its only really needed when the text wraps in the mobile view, because otherwise the second line starts on the left hand side ):

PR #1576, opened 2021-08-31
Fix efile security on resubmission

bytheway875:
The +/- for timezone offset was happening at a form level, but this now does it in the JS so we only have to do it in one place.

It also adds model level validations for efile submission things since we ultimately require them anyway.

### DIFF
--- a/spec/controllers/ctc/portal/portal_controller_spec.rb
+++ b/spec/controllers/ctc/portal/portal_controller_spec.rb
@@ -52,7 +52,7 @@ describe Ctc::Portal::PortalController do
             "browser_language"=>"en-US",
             "platform"=>"MacIntel",
             "client_system_time"=>"Tue Aug 31 2021 11:46:22 GMT-0500 (Central Daylight Time)",
-            "timezone_offset"=>"300"
+            "timezone_offset"=>"+300"
           }
         }
       end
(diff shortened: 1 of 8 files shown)


spec/controllers/ctc/portal/portal_controller_spec.rb
```diff
@@ -77,6 +77,21 @@
...
+        before do
+          params[:ctc_resubmit_form]["device_id"] = nil
+        end
+        it "flashes an alert and does not save" do
+          expect {
+            put :resubmit, params: params
+          }.to change(client.efile_security_informations, :count).by 0
+          expect(flash[:alert]).to eq(I18n.t("general.enable_javascript"))
+
+          expect(submission.current_state).to eq("rejected") # transitions to resubmitted and then to preparing
```
caitbot, line 91: this comment confuses me, shouldn't it not transition to resubmitted or preparing after not saving, I imagine this might be left over from the test it was replicating

caitbot (approved):

spec/controllers/ctc/portal/portal_controller_spec.rb
bytheway875: ohh yes, that it true. i will remove that.

PR #1577, opened 2021-09-01
Auto re-submit on IRS data loss [#178855164]

alaroia-cfa:
For the two IRS errors that indicate that they had trouble parsing our data, let's resubmit. We saw these when there was an IRS outage.

* https://www.getyourrefund.org/en/hub/errors/33
* https://www.getyourrefund.org/en/hub/errors/34

It's probably good to only resubmit once like this, to avoid creating an infinite loop.


caitbot (approved):
looks great! ✨

PR #1579, opened 2021-09-01
Add confirm payments page with 2021 advance CTC, RRC and third stimulus payment amounts

embarnard:


app/controllers/ctc/questions/confirm_payment_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      layout "intake"
+
+      def edit
+        @ctc_amount = ChildTaxCreditCalculator.total_advance_payment_2021(current_intake.dependents)
+        @rrc_amount = current_intake.tax_return(2020)&.claimed_recovery_rebate_credit || 0
+        @third_stimulus_amount = current_intake.tax_return(2020)&.expected_recovery_rebate_credit_three || 0
+        @not_collecting = (@ctc_amount.zero? && @rrc_amount.zero? && @third_stimulus_amount.zero?) ? true : false
+
+        send_mixpanel_event(event_name: "ctc_efile_estimated_payments", data: {
```
bytheway875, line 14: I wonder if we can find another place to send this mixpanel event. for example, I'd love if we can tie sending the mixpanel event into some transition of the intake being completed.

We currently run into a ton of errors with sending the same email or text multiple times when we mix side effects into the controller get or post methods like this. sending this data to mixpanel multiple times may affect the quality of the data.

app/controllers/ctc/questions/confirm_payment_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Ctc
+  module Questions
+    class ConfirmPaymentController < QuestionsController
+      include AuthenticatedCtcClientConcern
+
+      layout "intake"
+
+      def edit
+        @ctc_amount = ChildTaxCreditCalculator.total_advance_payment_2021(current_intake.dependents)
```
bytheway875, line 9: Either the ChildTaxCreditCalculator OR the data that we pass in needs to be restricted to only qualifying children.

I suggest passing in all of the dependents on the intake as you are doing right now, and then narrowing it down to only dependents that respond `true` to `dependent.qualifying_child_2020?`

It may make your life easier to add a method to TaxReturn to make this calculation so that we can ensure that the interface is consistent as well as ensuring that people can't get an unexpected/incorrect result by passing in the incorrect set of dependents.

app/controllers/ctc/questions/confirm_payment_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        @rrc_amount = current_intake.tax_return(2020)&.claimed_recovery_rebate_credit || 0
```
bytheway875, line 10: since we're calling current_intake.tax_return(2020) multiple times here, I think you could improve readability by defining tax_return once and calling methods on it.

There should be no situation here where a client does _not_ have a 2020 tax return, and claimed_recovery_rebate_credit should respond with 0 when they aren't claiming credit, so I'd recommend removing the &. and || 0 to simplify the logic a bit.

There are cases I've seen where test data (and real data) can cause the calculation to error out, but I actually don't think that either of the conditions above would help in that situation (we need to fix it separately).

app/models/tax_return.rb
```diff
@@ -114,6 +114,13 @@ def expected_recovery_rebate_credit_two
     )
   end
 
+  def expected_recovery_rebate_credit_three
+    EconomicImpactPaymentThreeCalculator.payment_due(
+      filer_count: rrc_eligible_filer_count,
+      dependent_count: intake.dependents.count(&:eligible_for_eip3?)
```
bytheway875, line 120: ohh cool. i didn't know that you could do this with count.

app/views/ctc/questions/confirm_payment/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    <div class="spacing-above-35">
+      <p class="spacing-below-15"><strong><%= t('views.ctc.questions.confirm_payment.third_stimulus') %></strong></p>
+      <p class="spacing-below-15"><%= number_to_currency(@third_stimulus_amount, precision: 0, locale: :en) %></p>
+      <p class="spacing-below-15"><i><%= t('views.ctc.questions.confirm_payment.third_stimulus_help_text') %></i></p>
+      <p><%= t('views.ctc.questions.confirm_payment.third_stimulus_link_html') %></p>
+    </div>
+  </div>
+
+  <%= link_to t("general.continue"), next_path, class: "button button--primary button--wide text--centered spacing-above-60" unless @not_collecting%>
+  <%= link_to t('views.ctc.questions.confirm_payment.do_not_file'), root_path, class: "button button--primary button--wide text--centered spacing-above-60" if @not_collecting %>
```
bytheway875, line 30: In cases where the client selects do not file here, I think we might want to do a bit more to their intake object and in the interface to tell them whats happening. For example, this will interrupt their intake, but I think that it wont remove the current_intake id from the session object. 

When clicking this button, I think we should consider:
- Moving the tax return object to "not filing" 
- Flashing a flash notice telling the client "thanks for using GetCTC, we won't be filing your return" or some such
- Remove the intake_id from the current session

app/lib/child_tax_credit_calculator.rb
```diff
@@ -1,15 +1,18 @@
...
 
-    dependents_under_six_payment + dependents_over_six_payment
+    dependents_under_six_payment + dependents_six_and_over_payment
   end
 
-  def self.total_payment_due(dependents_under_six_count:, dependents_over_six_count:)
-    monthly_payment_due(dependents_under_six_count: dependents_under_six_count, dependents_over_six_count: dependents_over_six_count) * 12
+  # This is half of the total CTC due for 2021, but clients have to file in 2022 for the 2nd half
+  def self.total_advance_payment_2021(dependents)
+    monthly_payment_due(dependents) * 6
```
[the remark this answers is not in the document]
bytheway875: One additional thing -- Since we're filing 2020 tax returns, even though this is advance payment they're getting in 2021 it might be more consistent with naming conventions to use the tax year instead of the filing year here.

app/controllers/ctc/questions/confirm_payment_controller.rb
embarnard: Note to ourselves: Does Kelly have the capability to de-dupe these events for unique visitors in mixpanel? Otherwise we could put an event when we queue the submission... or something.

app/views/ctc/questions/confirm_payment/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+    <div class="spacing-above-35">
+      <p class="spacing-below-15"><strong><%= t('views.ctc.questions.confirm_payment.third_stimulus') %></strong></p>
+      <p class="spacing-below-15"><%= number_to_currency(@third_stimulus_amount, precision: 0, locale: :en) %></p>
+      <p class="spacing-below-15"><i><%= t('views.ctc.questions.confirm_payment.third_stimulus_help_text') %></i></p>
+      <p><%= t('views.ctc.questions.confirm_payment.third_stimulus_link_html') %></p>
+    </div>
+  </div>
+
+  <%= link_to t("general.continue"), next_path, class: "button button--primary button--wide text--centered spacing-above-60" unless @not_collecting%>
```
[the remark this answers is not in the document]
embarnard: We can raise this question to Kelly.

app/views/ctc/questions/confirm_payment/edit.html.erb
embarnard: > Moving the tax return object to "not filing"

This is a good idea. The story says we should "archive the client," but we don't have archiving the client yet. I asked Anu, and she and and Kelly will chat with us to advise.

> Flashing a flash notice telling the client "thanks for using GetCTC, we won't be filing your return" or some such

Might be a good idea. We can raise to Anu/Kelly.

> Remove the intake_id from the current session

Seems very wise; might be good to fully destroy the session; this is basically sign out.

app/controllers/ctc/questions/confirm_payment_controller.rb
embarnard: Kelly says data science will use this (if anyone) and the events will be connected to a `visitor_id` so data science will be able to de-dupe.

PR #1580, opened 2021-09-01
SSNs not valid for employment are still validated by SSN type

caitbot:


caitbot (commented):
Reviewed with SB as pairs

PR #1582, opened 2021-09-02
Spouse name control should use the spouse name by default

caitbot:
The IRS documentation specifies that SpouseNameControl should use the primary last name. This guidance seems to be outdated (both technically - and frankly, culturally) based on our interactions with their systems, ie we've run into a lot of issues with partners with different last names.

So the fix is to try and use the spouse's own last name by default. This matches the guidance provided by Courtney, as well.

This also flips the override to use the primary name, rather than the spouse name, and sets it to `false` by default.

### DIFF
--- a/app/forms/concerns/ctc_client_form_attributes.rb
+++ b/app/forms/concerns/ctc_client_form_attributes.rb
@@ -38,11 +38,5 @@ module CtcClientFormAttributes
         attributes[column] = intake.send(column)
       end
     end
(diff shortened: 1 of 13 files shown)


bytheway875 (approved):

bytheway875 (approved):
also love that this addresses the issues of form attribute inconsistency when there wasnt a default value by making sure there is a default value.

PR #1585, opened 2021-09-07
Trim input lengths in xml

bytheway875:
- PlatformTxt can only be 15 chars
- BrowserLanguageTxt can only be 5 chars
- Person name type can only be 20 chars
- CityNm can only be 22 chars

### DIFF
--- a/spec/lib/submission_builder/return_header1040_spec.rb
+++ b/spec/lib/submission_builder/return_header1040_spec.rb
@@ -359,17 +359,35 @@ describe SubmissionBuilder::ReturnHeader1040 do
     context "efile security information" do
       context "UserAgentTxt" do
         before do
-          submission.client.efile_security_informations.first.update(user_agent: "Mozilla/5.0 (Linux; Android 10; SAMSUNG SM-S205DL) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/12.1 Chrome/79.0.3945.136 Mobile Safari/537.36")
-          submission.client.efile_security_informations.last.update(user_agent: "Mozilla/5.0 (Linux; Android 10; SAMSUNG SM-S205DL) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/12.1 Chrome/79.0.3945.136 Mobile Safari/537.36")
+
+          submission.client.efile_security_informations.first.update(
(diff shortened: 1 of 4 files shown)


spec/lib/submission_builder/return_header1040_spec.rb
```diff
@@ -359,17 +359,35 @@
...
+          expect(submission.client.efile_security_informations.first.browser_language.length).to eq 6
+          expect(submission.client.efile_security_informations.last.browser_language.length).to eq 6
+          expect(submission.client.efile_security_informations.last.platform.length).to eq 17
+          expect(submission.client.efile_security_informations.last.platform.length).to eq 17
+
           response = SubmissionBuilder::ReturnHeader1040.build(submission)
           xml = Nokogiri::XML::Document.parse(response.document.to_xml)
           expect(xml.at("FilingSecurityInformation AtSubmissionFilingGrp UserAgentTxt").text.length).to eq 150
           expect(xml.at("FilingSecurityInformation AtSubmissionCreationGrp UserAgentTxt").text.length).to eq 150
+          expect(xml.at("FilingSecurityInformation AtSubmissionFilingGrp BrowserLanguageTxt").text.length).to eq 5
```
alaroia-cfa, line 387: I would prefer that this validate the truncated string, which I don't see. I also can live with it as-is.

alaroia-cfa (approved):
Thanks for doing this! Very important to do it! :ship:

I saw somewhere that you said the IRS gave us guidance that truncation is appropriate. I'd love if you can point me to it, since I like seeing primary sources if possible.

I'm trusting you that these are the actual max lengths! :D

bytheway875:
@alaroia-cfa 

See: 

https://www.irs.gov/pub/irs-pdf/p4164.pdf


1.4.4.1 Addresses

Street Address Line 1 ‘AddressLine1Txt’ cannot contain more than 35 characters, abbreviation of lengthy
address information may be required to provide all pertinent information.


12.5 Filer Name Format

Includes instructions for truncating Primary/Spouse names in specific ways, but no others


There is no explicit guidance I can find for the other fields, which worries me a little -- may be worth reaching out to the IRS to see if they have other recommendations, but I'm not sure what other strategy there is.

For person names, there is a strict 20 character rule for submission, but since it's all one name there's no way to know "what" to truncate, so truncating the end off seems like the only option. However, we can reach out to the IRS if we need more information. 

(I feel okay trimming the length because what the IRS primarily is concerned with is the name control value, not the entire name for dependents)

PR #1592, opened 2021-09-07
Add percy #177084353

bengolder:
**This is a re-opened PR:** The [previous PR was merged](https://github.com/codeforamerica/vita-min/pull/1496) and [then reverted](https://github.com/codeforamerica/vita-min/pull/1568) after the percy dependency caused [a problem with nodejs versions when deploying to Aptible](https://www.pivotaltracker.com/story/show/179411030).

This PR will attempt to fix the nodejs version issue so that the Percy screenshotting capability can be safely merged.


alaroia-cfa:
@bengolder Thanks for re-submitting it!

PR #1594, opened 2021-09-07
Edit bank info

bytheway875:
Screen Shot 2021-09-07 at 4 24 15 PM
Screen Shot 2021-09-07 at 4 25 29 PM


app/views/ctc/portal/bank_account/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          <%= image_tag "check_routing_and_account.png", class: "spacing-below-35" %>
+
+          <%= f.cfa_input_field(:routing_number, t("views.ctc.questions.routing_number.routing_number"), classes: ["form-width--long"]) %>
+          <%= f.cfa_input_field(:routing_number_confirmation, t("views.ctc.questions.routing_number.routing_number_confirmation"), classes: ["form-width--long"]) %>
+          <%= f.cfa_input_field(:account_number, t("views.ctc.questions.account_number.account_number"), classes: ["form-width--long"]) %>
+          <%= f.cfa_input_field(:account_number_confirmation, t("views.ctc.questions.account_number.account_number_confirmation"), classes: ["form-width--long"]) %>
+        </div>
+
+        <button class="button button--primary button--wide" type="submit">
+          <%=t("general.save") %>
```
alaroia-cfa, line 36: Trivial, but RubyMine auto-format (Cmd-Option-L) will convert `<%=t(...) %>` into `<%= t(...) %>`, adding a space, and I like the space.

Not a blocker at all.

app/views/ctc/portal/bank_account/edit.html.erb
```diff
@@ -0,0 +1,42 @@
+<% @main_question = t("views.ctc.portal.bank_account.title") %>
+
+<% content_for :page_title, @main_question %>
+
+<% content_for :card do %>
+  <div class="question-layout">
+    <div class="back">
+      <%= link_to t("general.back"), defined?(prev_path) ? prev_path : :back %>
```
alaroia-cfa, line 8: Consider explicitly defining `prev_path` in the controller and getting rid of the `defined?` check. Not a blocker.

app/views/ctc/questions/confirm_information/_bank_account_info.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="review-box__title">
+  <h2><%= t("views.ctc.questions.confirm_bank_account.bank_information") %></h2>
+</div>
+<div class="review-box__space-between bank-account-info">
+  <ul class="spacing-below-0 spacing-above-0">
+    <li class="review-box__name"><%= bank_account.bank_name %></li>
+    <li class="review-box__details"><%= t("general.type") %>: <%= bank_account.account_type&.titleize %></li>
```
alaroia-cfa, line 7: I think you don't need `&.` here; `.` would work fine.

Not a blocker.

spec/features/ctc/portal_spec.rb
```diff
@@ -235,6 +235,34 @@ def log_in_to_ctc_portal
...
+      fill_in I18n.t('views.ctc.questions.routing_number.routing_number_confirmation'), with: "123456789"
+      fill_in I18n.t('views.ctc.questions.account_number.account_number'), with: "123456789"
+      fill_in I18n.t('views.ctc.questions.account_number.account_number_confirmation'), with: "123456789"
+      click_on I18n.t("general.save")
+
+      within ".bank-account-info" do
+        expect(page).to have_text "Bank of Two Melons"
+        expect(page).to have_text "Type: Checking"
+        expect(page).to have_text "Routing number: 123456789"
+        expect(page).to have_text "Account number: ●●●●●6789"
```
alaroia-cfa, line 264: ⭐️

spec/forms/ctc/portal/bank_account_form_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    context "when there is an existing bank account object" do
+      before do
+        create(:bank_account, intake: intake)
+      end
+
+      it "updates the existing bank account object" do
+        expect {
+          described_class.new(intake, params).save
+          intake.reload
+        }.to change(BankAccount, :count).by(0)
```
alaroia-cfa, line 137: I would have said `not_to change(...)` but this is OK too!

spec/validators/account_number_validator_spec.rb
```diff
@@ -0,0 +1,62 @@
+require "rails_helper"
+class Validatable
```
alaroia-cfa, line 2: If possible, avoid defining classes like these in specs; when writing this way, they're not cleaned-up per test; since different spec files have `class Validatable`, due to Ruby open classes, the different test files are sharing one class, which is rather odd in my opinion.

If you can, try this approach:

https://github.com/codeforamerica/vita-min/blob/faed0d61efde3f616723e13c3505b38315b9adcc/spec/lib/submission_builder/formatting_methods_spec.rb#L3-L4

Not a blocker.

alaroia-cfa (approved):
Splendid. Well tested. I left some non-blocking comments.

app/controllers/ctc/portal/bank_account_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    render edit_template
+  end
+
+  private
+
+  def edit_template
+    "ctc/portal/bank_account/edit"
+  end
+
+  def form_class
```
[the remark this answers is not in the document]
alaroia-cfa: Oh darn it - I meant to edit the above comment before submitting it.

Honestly a question - does `form_class` figure out the right form here? I will go check now.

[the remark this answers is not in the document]
alaroia-cfa: OK - I thought this inherited from something that defined `form_class` but I see it does not, so please ignore this thread.

app/forms/ctc/account_number_form.rb
```diff
@@ -3,7 +3,7 @@ class AccountNumberForm < QuestionsForm
     set_attributes_for :bank_account, :account_number
     set_attributes_for :confirmation, :account_number_confirmation
 
-    validates :account_number, confirmation: true, presence: true
+    validates :account_number, presence: true, confirmation: true, length: { maximum: 17 }, account_number: true
```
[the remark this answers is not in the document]
bytheway875: i dont... im sure there's some reasonable minimum, but i couldnt find anything definitive online about it. i could probably add something like 5, as it would be absurd to have a bank account less than 5 digits.

bytheway875:
im going to push this and then open another PR for cleanup so that I can get product acceptance and revision started as well.

app/forms/ctc/account_number_form.rb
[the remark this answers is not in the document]
alaroia-cfa: I'm fine with no minimum in this case!

PR #1595, opened 2021-09-07
Update node version

bengolder:
- This bumps the Aptible node version from 10 to 12, which is required for being able to list percy/cli as a dev dependency.
- It also requires that anything running yarn has at least Node 12. This will require developers to upgrade node locally in rare cases

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "jest-junit": "^12.0.0",
     "webpack-dev-server": "^3.11.0"
   },
+  "engines": {
+    "node": ">=12.16"
+  },
   "scripts": {
     "test": "jest"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,6 +25,9 @@
     "jest-junit": "^12.0.0",
     "webpack-dev-server": "^3.11.0"
   },
+  "engines": {
+    "node": ">=14.17.6 <15"
+  },
```
bengolder, line 30: Alternatively, we could remove the `<15` requirement which would allow everyone to keep the newer node versions that they likely have installed.

alaroia-cfa:
Looks like CircleCI needs the same treatment, which will be annoying, but possible. Let me know if you want help with this. Thanks for taking a deep look at this! I dropped this ball a week or two ago and am grateful you're picking it up.

package.json
alaroia-cfa: I think the `<15` is worth it, so that everyone upgrades together, even though it'll be annoying. The one-time pain will be worth it imho.

alaroia-cfa (approved):
Great! Seems to work peacefully on CircleCI as well as Aptible. I'll merge. Thanks for taking care of this -- @coltborg , I had told you I'd do this, but I had dropped the ball, so I'm grateful it has moved forward.

PR #1596, opened 2021-09-08
Dependent has a default_scope to omit soft-deleted records

tgrathwell:
adds a concern that also provides scopes for .with_deleted
or .only_deleted

### DIFF
--- a/app/models/concerns/soft_deletable.rb
+++ b/app/models/concerns/soft_deletable.rb
@@ -1,0 +1,10 @@
+module SoftDeletable
+  extend ActiveSupport::Concern
+
+  included do
+    default_scope { where(soft_deleted_at: nil) }
+
+    scope :with_deleted, -> { unscope(where: :soft_deleted_at) }
+    scope :only_deleted, -> { unscope(where: :soft_deleted_at).where.not(soft_deleted_at: nil) }
+  end
+end
--- a/app/models/dependent.rb
+++ b/app/models/dependent.rb
@@ -47,6 +47,8 @@
 #
 
 class Dependent < ApplicationRecord
+  include SoftDeletable
+
   belongs_to :intake, inverse_of: :dependents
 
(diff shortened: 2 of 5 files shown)


app/models/dependent.rb
```diff
@@ -47,6 +47,8 @@
 #
 
 class Dependent < ApplicationRecord
+  include SoftDeletable
```
bytheway875, line 50: oh smart. this is great that we can use the same pattern across multiple tables.

app/models/tax_return.rb
```diff
@@ -67,7 +67,7 @@ class TaxReturn < ApplicationRecord
   def qualifying_dependents
     raise StandardError, "Qualifying dependent logic is only valid for 2020.  Define new rules for #{year}." unless year == 2020
 
-    intake.dependents.reject(&:soft_deleted_at).filter { |d| d.qualifying_child_2020? || d.qualifying_relative_2020? }
+    intake.dependents.filter { |d| d.qualifying_child_2020? || d.qualifying_relative_2020? }
```
bytheway875, line 70: The eip eligibility methods are currently using qualifying_dependents (so that they could take advantage of the soft deletion for the hot fix). I'm now wondering if we should revert to having those method call intake.dependents since they call the qualification logic in those methods, so we're kind of looping through the same set of dependents twice.

app/models/concerns/soft_deletable.rb
```diff
@@ -0,0 +1,10 @@
+module SoftDeletable
+  extend ActiveSupport::Concern
+
+  included do
+    default_scope { where(soft_deleted_at: nil) }
+
+    scope :include_deleted, -> { unscope(where: :soft_deleted_at) }
```
bytheway875, line 7: wondering if "with_deleted" may be better here. i think of "include" being a keyword in active record that does eager loading/joins

tgrathwell: i did it

PR #1599, opened 2021-09-08
Send email to clients who signed up for GetCTC updates [#179186159]

alaroia-cfa:


app/mailers/ctc_signup_mailer.rb
```diff
@@ -10,4 +10,14 @@ def beta_navigation(email_address:, name:)
       subject: @subject,
     )
   end
+
+  def launch_announcement(email_address:, name:)
+    @name = name
+    subject = "Thank you for signing up to receive updates regarding GetCTC! / ¡Gracias por registrarse para recibir las actualizaciones de GetCTC!"
```
bytheway875, line 16: do we not want people to open this email 😆

app/lib/ctc_signup_message.rb
```diff
@@ -38,4 +38,37 @@ def self.send(count)
...
+  def self.send_launch_announcement(count)
+    ctc_signups = CtcSignup.where(launch_announcement_sent_at: nil).take(count)
+    ctc_signups.each do |ctc_signup|
+      _send_one_launch_announcement(ctc_signup)
+    end
+  end
+
+  def self._send_one_launch_announcement(ctc_signup)
+    if ctc_signup.email_address.present?
+      CtcSignupMailer.launch_announcement(email_address: ctc_signup.email_address, name: ctc_signup.name).deliver_now
```
bytheway875, line 51: the choice in the CTC beta email to use deliver_now over deliver_later is something I regret. The text message call queues a job, so using deliver_later here to queue a job may also make sense.

app/lib/ctc_signup_message.rb
```diff
@@ -38,4 +38,37 @@ def self.send(count)
       ctc_signup.touch(:beta_email_sent_at)
     end
   end
-end
\ No newline at end of file
+
+  def self.send_launch_announcement(count)
```
bytheway875, line 42: the request for the beta emails was to send in chunks with specific counts, but here i think the request is to send to everyone on the list. what if passing nothing for the count got all of them?

app/mailers/ctc_signup_mailer.rb
alaroia-cfa: Good question. :) Kelly said it was fine!! :) I'll call and double-check.

app/lib/ctc_signup_message.rb
alaroia-cfa: I was thinking I'd pass in `10` first in production, make sure nothing blew up, then do them in batches of `100` or `1000` to make sure nothing blew up as a way to do all of them. Treating a missing `count` as infinity is fine too, shrug.

FWIW, I was also thinking that I'd use `def self._send_one_launch_announcement` to rehearse with a Kelly sign-up in production.

app/lib/ctc_signup_message.rb
```diff
@@ -38,4 +38,37 @@ def self.send(count)
...
+    if ctc_signup.email_address.present?
+      CtcSignupMailer.launch_announcement(email_address: ctc_signup.email_address, name: ctc_signup.name).deliver_now
+    end
+
+    text_email_template_file = File.read(File.join(Rails.root, "app/views/ctc_signup_mailer/launch_announcement.text.erb"))
+    text_email_template = ERB.new(text_email_template_file)
+    @name = ctc_signup.name
+    text_email = text_email_template.result(binding)
+
+    en_sms, es_sms = text_email.split('--------------------------')
```
alaroia-cfa, line 59: @bytheway875 FYI I'm going to have to change this to be two new hardcoded SMSs, based on a request from Kelly. I am under the impression you'd be OK with me making that change and merging, so I'll do that.

app/mailers/ctc_signup_mailer.rb
alaroia-cfa: Shorter would be better but in this case for expediency Kelly says we can leave it the way it is.

PR #1600, opened 2021-09-08
Fix screenshot test failure and point to target branch in percy

coltborg:
# What does this PR do?

- Fixes `new_joint_filers_spec.rb` test, since it was updated to use the `js` driver, the `upload` button needed to be pressed when uploading files.
- Improve `bin/percy` script to:
  - Target `main` branch
  - Associate a PR if it's already opened ([this is optional](https://docs.percy.io/docs/environment-variables#optional))

### DIFF
--- a/bin/percy
+++ b/bin/percy
@@ -10,24 +10,31 @@ def system!(*args)
 end
 
 FileUtils.chdir APP_ROOT do
+  puts "Enter the base branch (press enter to use main)"
+  target_branch = gets.chomp.presence || 'main'
+  puts "Enter a 4 digit PR number if available (or press enter)"
+  pr_number = gets.chomp
+
   system!({ "PERCY_TOKEN" => Rails.application.credentials.percy_token },
     '
+      set -e
       if [ -z $PERCY_TOKEN ]; then echo "== 🚨 Please download development.key to access PERCY_TOKEN =="; exit 1; fi
 
-      if [ -z "$(git status --porcelain)" ]; then
-        echo "== 📸 Take baseline images =="
-        git checkout main
-        bundle install
-        npx percy exec -- rspec --tag screenshot
-        echo "== 📸 Take new images =="
-        git checkout -
-        bundle install
-        npx percy exec -- rspec --tag screenshot
-      else
-       echo "== 🚨 Please commit changes first =="
-       exit 1
-      fi
-    '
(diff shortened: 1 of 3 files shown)


bin/percy
```diff
@@ -10,24 +10,24 @@ def system!(*args)
...
-        echo "== 📸 Take new images =="
-        git checkout -
-        bundle install
-        npx percy exec -- rspec --tag screenshot
-      else
-       echo "== 🚨 Please commit changes first =="
-       exit 1
-      fi
-    '
+      if [ -n "$(git status --porcelain)" ]; then echo "== 🚨 Please commit changes first =="; exit 1; fi
```
coltborg, line 20: I changed this a bit in that these if statements are gates. If they both pass, then the script runs. Instead of the previous if/else flow.

bin/percy
```diff
@@ -10,24 +10,24 @@ def system!(*args)
...
+
+      echo "== 📸 Take baseline images =="
+      git checkout main
+      bundle install
+      npx percy exec -- rspec --tag screenshot
+      echo "== 📸 Take new images =="
+      git checkout -
+      bundle install
+      PERCY_TARGET_BRANCH=main PERCY_PULL_REQUEST=%{pr_number} npx percy exec -- rspec --tag screenshot
```
coltborg, line 29: I apply these env variables only on the new branch execution. Since [these are optional](https://docs.percy.io/docs/environment-variables#optional), `PERCY_PULL_REQUEST` can be empty when running the script without giving it a PR number.

bin/percy
alaroia-cfa: Nice use of `git status --porcelain`! I don't think I knew about it!

```
❯ man git-status | grep -A 4 -F -- --porcelain

       --porcelain[=<version>]
           Give the output in an easy-to-parse format for scripts. This is
           similar to the short output, but will remain stable across Git
           versions and regardless of user configuration. See below for
           details.
```

alaroia-cfa: I guess it was here before this pull request, but anyway, it's cool. :)

alaroia-cfa (commented):
Nice script changes!

I'll look at the test failures now.

alaroia-cfa (changes requested):
Per our Tuple discussion, let's undo the changes in the spec file!

alaroia-cfa (approved):
🚢

PR #1602, opened 2021-09-09
Handle HTML entities like apostrophe reliably when parsing PDF form contents in tests

alaroia-cfa:
Also make sure we're using pdftk-java consistently everywhere.


Brewfile
```diff
@@ -1,8 +1,9 @@
-brew "rbenv"
```
alaroia-cfa, line 1: I sorted the file.

Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN apt-get update --allow-releaseinfo-change
 
 # System prerequisites
 RUN apt-get update \
- && apt-get -y install build-essential libpq-dev pdftk ghostscript poppler-utils curl \
```
alaroia-cfa, line 8: I irrelevantly switched from `pdftk` to `pdftk-java` -- we're already using `pdftk-java`; this makes it clearer.

spec/factories/intakes.rb
```diff
@@ -321,7 +321,7 @@
 
   trait :with_faker_contact_info do
     primary_first_name { Faker::Name.first_name }
-    primary_last_name { Faker::Name.last_name }
```
alaroia-cfa, line 324: I figured if we want to be sure we handle apostrophes reliably, we can choose to always add an apostrophe in there.

spec/support/pdf_spec_helper.rb
```diff
@@ -3,7 +3,8 @@ def filled_in_values(file_path)
     filled_in_fields = PdfForms.new.get_fields(file_path)
 
     filled_in_fields.each_with_object({}) do |field, hash|
-      hash[field.name] = field.value
+      # The pdf-forms gem returns HTML-escaped form values when using pdftk-java. Unescape.
```
alaroia-cfa, line 6: This general idea is alluded-to in the pdftk readme -- https://github.com/jkraemer/pdf-forms#field-names-with-html-entities -- although they're concerned about field names, not field values.

My guess is that the treatment of apostrophes is a behavior difference between `pdftk` and `pdftk-java` and the `pdf-forms` gem author doesn't know about this behavior difference.

Dockerfile
bytheway875: remember back in the day when we had an issue with PDF stuff and differences between what was running locally and what was running on CI and production? How can we make sure that what we install locally to our computers matches?

spec/factories/intakes.rb
tgrathwell: I find Faker really irritating because in feature-land someone will write a test that says `expect(page).to have_content(last_name)` and it will fail one out fifty times because what's on the page is `O&apos;Connor` or whatever

I don't see a point to making our test data this random, and I don't like having to care about apostrophes when making those kind of assertions, so if we're going to change `primary_last_name` to anything I think it should be a nice static string like "Smith"

spec/factories/intakes.rb
```diff
@@ -321,7 +321,7 @@
 
   trait :with_faker_contact_info do
```
alaroia-cfa, line 322: Let's try stopping using Faker! Let's use incrementing always. For names, let's try to spell it out so it's "Cherry One" and "Cherry Two" etc.

alaroia-cfa: And always apostrophe in the last name why not.

Dockerfile
alaroia-cfa: `brew bundle install` will install things from `Brewfile` which is what we want.

Brewfile
```diff
@@ -1,8 +1,9 @@
...
-brew "postgis"
+brew "ghostscript"
+brew "hub"
 brew "imagemagick"
+brew "pdftk-java"
```
alaroia-cfa, line 4: And I added this line.

PR #1603, opened 2021-09-09
Imperfect return indicator acceptance

bytheway875:
We learned last week that "Exception" acknowledgements indicate that the IRS has accepted the return, but with some conditions.

Move "exception" acknowledgement status into accepted instead of failed, and add messaging to explain the implications of this type of acceptance to the clients.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1426,6 +1426,9 @@ en:
             Reason for this rejection:
           next_step_title: |
             Next steps:
+          imperfect_return_indicator_text: |
+            Your return was accepted by the IRS, inlcuding a request to update one more more of your dependents' legal names in their records.
+            The IRS estimates that it will take 4-8 weeks to process your return. You may also receive a letter confirming your request to update the dependent's legal name.
           status:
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -1426,6 +1426,9 @@ en:
             Reason for this rejection:
           next_step_title: |
             Next steps:
+          imperfect_return_indicator_text: |
```
alaroia-cfa, line 1429: Typo: `inlcuding` != `including`

spec/services/efile/poll_for_acknowledgments_service_spec.rb
```diff
@@ -117,8 +117,9 @@
 
             it "changes the state from transmitted to investigating" do
```
alaroia-cfa, line 118: This `it "..."` line doesn't seem to describe what happens. What about `it "changes the state to accepted with the imperfect return acceptance flag"`?

alaroia-cfa (approved):
I have some non-blocking comments; feel free to fix them if you like!

PR #1605, opened 2021-09-09
Send a CtcExperienceSurvey 1 day after a CTC client's return enters accepted or rejected state

tgrathwell:

### DIFF
--- a/app/jobs/send_client_ctc_experience_survey_job.rb
+++ b/app/jobs/send_client_ctc_experience_survey_job.rb
@@ -1,0 +1,16 @@
+class SendClientCtcExperienceSurveyJob < ApplicationJob
+  def perform(client)
+    client.with_lock do
+      unless client.ctc_experience_survey_variant.present?
+        random_variant = 1 + rand(3)
+        client.update!(ctc_experience_survey_variant: random_variant)
+      end
+    end
+
+    SurveySender.send_survey(
+      client,
+      :ctc_experience_survey_sent_at,
+      AutomatedMessage::CtcExperienceSurvey
+    )
+  end
+end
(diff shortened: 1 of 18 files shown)


app/models/automated_message/ctc_experience_survey.rb
```diff
@@ -0,0 +1,25 @@
+module AutomatedMessage
+  class CtcExperienceSurvey
+    def self.name
+      'messages.surveys.ctc_experience'.freeze
+    end
+
+    def self.survey_link(client)
+      ctc_rejected_status = client.intake.tax_return(2020).status == "file_rejected" ? 'TRUE' : 'FALSE'
+
+      "https://codeforamerica.co1.qualtrics.com/jfe/form/SV_cHN2H3IWcxAEKPA?&ExternalDataReference=#{client.id}&ctcRejected=#{ctc_rejected_status}&expGroup=#{client.ctc_experience_survey_variant}"
```
bytheway875, line 10: if im understanding this correctly, its the same survey and we just send a different query param if its rejected?

bytheway875: i noticed the first query param has an & in front too `?&` -- likely doesn't affect anything, but i think the & is extraneous.

app/jobs/send_client_ctc_experience_survey_job.rb
```diff
@@ -0,0 +1,16 @@
+class SendClientCtcExperienceSurveyJob < ApplicationJob
+  def perform(client)
+    client.with_lock do
```
bytheway875, line 3: i feel like it would make more sense to me to have the survey sender deal with assigning a variant instead of doing it in this job since it's a survey-level concern. It's not immediately clear from reading this code where the survey_variant gets used, so having it live closer to where the data is consumed might help.

bytheway875 (commented):
i have a slight preference for moving the random variant generator into the send survey class, but i dont feel super strongly on it. consider changing or merge as-is

PR #1607, opened 2021-09-09
Fix 2019 agi logic so that clients who did non filer arent asked for agi

bytheway875:


spec/forms/ctc/spouse_filed2019_form_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      it "is not valid" do
+        expect(
+            described_class.new(intake, {})
+        ).not_to be_valid
+      end
+    end
+
+    context "when spouse_filed_2019 is not in the set" do
+      let(:filed_2019) { "on_the_moon" }
+      it "is not valid" do
```
alaroia-cfa, line 28: Consider `it "is valid at the form level but invalid at the model level" do`

to explain this behavior.

Not a blocker.

alaroia-cfa: If the model has testing for this validation, then consider removing this test.

spec/forms/ctc/spouse_filed2019_form_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        expect {
+          described_class.new(intake, params).save
+        }.to change(intake, :spouse_filed_2019).from("unfilled").to("did_not_file")
+        .and not_change(intake, :spouse_prior_year_agi_amount)
+      end
+    end
+
+    context "filed_non_filer_separate" do
+      let(:filed_2019) { "filed_non_filer_separate" }
+      it "updates spouse filed 2019, does not set agi" do
```
alaroia-cfa, line 79: Consider

`      it "updates spouse filed 2019, does set agi" do`

I think it **does** set the spouse AGI but the test description states that it does not set the AGI.

spec/forms/ctc/spouse_filed2019_form_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        expect {
+          described_class.new(intake, params).save
+        }.to change(intake, :spouse_filed_2019).from("unfilled").to("filed_non_filer_joint")
+         .and change(intake, :spouse_prior_year_agi_amount).from(nil).to(1)
+      end
+    end
+
+    context "filed_full_separate" do
+      let(:filed_2019) { "filed_full_separate" }
+      it "updates spouse filed 2019, sets AGI to 1" do
```
alaroia-cfa, line 59: Consider: `it "updates spouse filed 2019 and does not modify AGI"`.

I think this description is not accurate the way it is written - I can see in the test it does not change an AGI, but the `it ...` line says it does change an AGI.

spec/forms/ctc/spouse_filed2019_form_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+          described_class.new(intake, params).save
+        }.to raise_error ArgumentError
+      end
+    end
+  end
+
+  context "save" do
+    context "filed_full_joint" do
+      let(:filed_2019) { "filed_full_joint" }
+      it "updates spouse filed 2019, does not set agi" do
```
alaroia-cfa, line 39: These tests have a lot of boilerplate, which makes it hard to know what's unique about each test case. Consider a `before ... do` with `described_class.new(intake, params).save`.

I know that doing that makes it hard to use the `expect { ... }.to change(...)` syntax. I would rather see:

```
      let(:filed_2019) { "filed_full_joint" }
      it "updates spouse filed 2019, does not set agi" do
         intake.reload
         expect(intake.spouse_filed_2019).to eq("filed_full_joint")
         expect(intake.spouse_prior_year_agi_amount).to eq(nil)
      end
```

The tests are all so similar that I wish it weren't necessary to write the "it ..." description name at all. Maybe this is a good use for some of these ideas:

https://stackoverflow.com/questions/8754582/best-way-to-organize-tests-in-rspec-that-have-a-combinations-of-factor

Anyway, for now, please do fix the test description names, or come up with a better way to organize these! :)

alaroia-cfa (approved):
It's OK as it is! I have some quibbles about test descriptions which I would love to see addressed. They're not strictly blockers but I hope you will consider making those changes. Fine to merge as-is or with those changes.

spec/forms/ctc/spouse_filed2019_form_spec.rb
bytheway875: oh yeah, i just copy and pasted all of these tests. test descriptions are annoying. will change

bytheway875:
made test changes to descriptions, but not doing any further refactoring

PR #1612, opened 2021-09-10
Replace flaky Capybara `within('h1')... have_text` with `have_selector`

bensheldon:
I believe this is flaky because once the wrapping `within` block finds a match, it will not refresh if a navigation/refresh happens.

It looks like `have_selector` is the typical usage across the codebase, so these are just a few outliers.


alaroia-cfa (approved):
This is a great pattern to unify on. I approve. Thanks!!!

PR #1614, opened 2021-09-13
Show efile security informations in the E-file tab

embarnard:


embarnard:
Screen Shot 2021-09-13 at 11 35 35 AM

app/views/hub/clients/show.html.erb
```diff
@@ -294,9 +294,7 @@
           </div>
           <% if @client.intake.is_ctc? && !@client.intake.drop_off? %>
             <div class="client-profile__field-group">
-              <h2 class="text--bold">
-                Prior Year AGI Amount
-              </h2>
+              <h2 class="text--bold">Prior Year AGI Amount</h2>
```
embarnard, line 297: unrelated to this ticket but because this was on another line it was adding a space before the text

app/views/hub/efile_submissions/_efile_security_information.html.erb
```diff
@@ -0,0 +1,85 @@
...
+                <span class="form-question">User Agent:</span>
+                <span class="label-value"><%= esf_1.user_agent %></span>
+              </div>
+              <div class="field-display">
+                <span class="form-question">Device ID:</span>
+                <span class="label-value"><%= esf_1.device_id %></span>
+              </div>
+            </div>
+          </div>
+          <% if @client.efile_security_informations.size > 1 %>
```
embarnard, line 43: I'm assuming we want to show the EfileSecurityInformation that we capture at the beginning and end of the flow

alaroia-cfa: Yes - good to show all of them!

caitbot: It looks like we also create one when a client resubmits their information in the portal, so it might make sense to just display all present `EfileSecurityInformations`.

Otherwise this might show just the one we collect on the `income_form` and the one from resubmission (and skip the one created on the `legal_consent_form`)

I don't have a ton of context on how these are being used. But generally, I bias towards showing more data if it's going to be used for debugging, especially if whoever's using it assumes they will be shown everything.

bytheway875: i'd honestly prefer that we loop through all of them and display all of them instead of limiting it to just two... but i guess i could be convinced otherwise.

I also think that this is a great candidate for a partial where we can pass the efile_security_information object in!

bytheway875: @alaroia-cfa in the case that the first and last is not "all" of them, do you think we should be looping through all of them or just displaying the first and last.

bytheway875:
I really love the design consideration you gave to this -- it looks really really nice. I'd like to see a refactor to what's part of the partial and what's laid out on the parent html before merging, but also down to discuss the tradeoffs over tuple or slack!

app/views/hub/efile_submissions/_efile_security_information.html.erb
alaroia-cfa: I think we should show all of them. We're showing these to support our own manual investigation, so I think it's good to show all of them.

embarnard: Thanks for the feedback y'all! Okay I can add a loop for all of them. I did two because I went into prod and queried for recent random clients and found that most of them had 1-2 associations to EfileSecurityInformation. And I didn't use a loop because I wanted them to be in two different columns. I'll take the accordion out of the partial too.

bytheway875: @embarnard its possible to keep the column layout and loop! we can work on some css strategies for doing that this afternoon, if you want to pair on it for a bit!

PR #1615, opened 2021-09-13
Move confirm information so that people with unexposed errors can update their information

bytheway875:
Limiting showing this button to only people who have an exposed error means that the people affected by bank account bundle fails couldn't get to the update info page.


bytheway875:
I'm hoping that this logic can be refined so that the button's appropriately hidden from people in certain situations.

caitbot (approved):
this approach sounds a-ok to me

PR #1618, opened 2021-09-13
Record completed_at onto intake on the last step of intake

bytheway875:


app/forms/ctc/confirm_legal_form.rb
```diff
@@ -14,7 +14,9 @@ class ConfirmLegalForm < QuestionsForm
     validates_presence_of :device_id, :user_agent, :browser_language, :platform, :timezone_offset, :client_system_time, :ip_address
 
     def save
-      @intake.update(attributes_for(:intake))
+      intake_attributes = attributes_for(:intake)
+      intake_attributes[:completed_at] = DateTime.current unless @intake.completed_at.present?
+      @intake.update(intake_attributes)
```
caitbot, line 19: we could do `update!` here that way we know if something goes terribly wrong

caitbot: not a blocker though

caitbot (approved):
AL and I pair reviewed and this looks great

PR #1620, opened 2021-09-13
GyrEfiler: raise RetryableError if we get a redirect on login

tgrathwell:
historically this has happened over the weekend during some IRS downtime

one example is https://www.getyourrefund.org/en/hub/clients/582739/efile


caitbot (approved):
Reviewed with Asheesh as a pair, this looks good to us, we just had a comment on the conditions around identifying this particular IRS issue.

PR #1621, opened 2021-09-13
redirect to portal home when there are no efile submissions

bytheway875:


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -17,7 +17,11 @@ def home
   end
 
   def edit_info
-    @current_submission = current_intake.client.efile_submissions.last
+    @submission = current_client.efile_submissions.last
```
alaroia-cfa, line 20: +1 on the conversion to `current_client`

app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -17,7 +17,11 @@ def home
...
+    unless @submission.present?
+      Sentry.capture_message "Client #{current_client.id} unexpectedly lacks an efile submission."
```
alaroia-cfa, line 22: Based on https://sentry.io/organizations/codeforamerica/issues/2646400314/?project=1880321&referrer=slack , we don't get a stack trace for `.capture_message`; it'd be nice to get a stack trace. Also it's 100% not a blocker.

alaroia-cfa (approved):
Works for me!

(Pair-reveiwed with @caitbot )

app/controllers/ctc/portal/portal_controller.rb
bytheway875: @alaroia-cfa do you know in order to use capture_error, do I actually have to raise and then rescue an error? Or can I pass this same message

I kind of added this just as a tracker because i wanted to make sure we weren't breaking the page but would like to learn more about how people get in this state.

alaroia-cfa: @bytheway875 I haven't tried it, but it seems that every controller has a method called `capture_message` which is implemented as follows:

https://github.com/getsentry/sentry-ruby/blob/94c28fe9a7fd1c15a10eeeb0e1c4fcac099d4166/sentry-rails/lib/sentry/rails/controller_methods.rb#L4-L7

I think that would be a fine way to get more context in, and a full stack trace wouldn't be necessary.

PR #1622, opened 2021-09-14
Remove ready to submit state because we're no longer using it

bytheway875:


alaroia-cfa (approved):
🚢 

Code deleted is code tested.

Catie and I noticed there's still 1 in prod (was this the one in a weird state where it's half-deleted?)

image

Please handle that in prod before this gets deployed -- probably before you merge this, to avoid forgetting (I would probably personally forget unless I did it before I merged).

bytheway875:
its not actually going to hurt anything to let that efile submission remain, but i will delete it anyway.

PR #1624, opened 2021-09-14
Create explicit bank account errors

bytheway875:
We're still getting some bank account details are incorrect errors, even after adding validations. Bleh.

It's also just generally creating some other issues in the code to not have these recorded as a specific error type.

### DIFF
--- a/app/lib/default_error_messages.rb
+++ b/app/lib/default_error_messages.rb
@@ -30,5 +30,12 @@ class DefaultErrorMessages
       source: :internal,
       expose: false
     )
+    EfileError.find_or_create_by!(
+      code: "BANK-DETAILS",
+      message: "Some of the provided bank account details are incorrect or absent.",
+      source: :intake,
+      expose: true,
+      auto_wait: true
+    )
   end
 end
--- a/app/lib/efile/submission_error_parser.rb
+++ b/app/lib/efile/submission_error_parser.rb
@@ -6,15 +6,15 @@ module Efile
       @raw_response = transition.metadata["raw_response"]
     end
 
-    def to_xml
-      Nokogiri::XML(@raw_response)
-    end
-
-    def persist_efile_errors_from_transition_metadata
+    def persist_errors
       metadata = @transition.metadata
       # these errors are mostly failures on our side of things, like issues with bundling and addresses
       if metadata["error_code"].present?
         persist_errors_from_error_metadata
+      end
+
+      if @transition.to_state == "failed" && metadata["raw_response"].present?
+        persist_bundle_errors_from_raw_response_metadata
       end
 
       # there errors are responses from the IRS, like rejections or exceptions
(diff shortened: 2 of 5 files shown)


app/lib/default_error_messages.rb
```diff
@@ -30,5 +30,11 @@ def self.generate!
       source: :internal,
       expose: false
     )
+    EfileError.find_or_create_by!(
+      code: "BANK-DETAILS",
+      message: "Some of the provided bank account details are incorrect or absent.",
+      source: :client,
```
bytheway875, line 36: we don't have a source 'client' already. We have internal, usps, and irs. This doesn't truly feel like any of those, but I don't know if this makes sense either. Any suggestions?

app/lib/efile/submission_error_parser.rb
```diff
@@ -6,23 +6,25 @@ def initialize(transition)
       @raw_response = transition.metadata["raw_response"]
     end
 
-    def to_xml
-      Nokogiri::XML(@raw_response)
-    end
-
-    def persist_efile_errors_from_transition_metadata
+    def persist_errors
```
bytheway875, line 9: I did a little refactoring in this file. First, I made just this one method public and shortened its name -- I was getting really confused as to which method was being called in the state machine.

I also made the other methods that get called in persist_errors private and refactored the tests a bit.

app/lib/default_error_messages.rb
caitbot: Arguably the source could be `irs` because it is their schema validation that causes the failure but yeah, I see your point, this could also be `intake`? Since the data is more closely associated with the intake rather than client.

But this is also fine. :)

app/lib/efile/submission_error_parser.rb
```diff
@@ -6,23 +6,25 @@ def initialize(transition)
...
       metadata = @transition.metadata
       # these errors are mostly failures on our side of things, like issues with bundling and addresses
       if metadata["error_code"].present?
         persist_errors_from_error_metadata
       end
 
+      if @transition.to_state == "failed" && metadata["raw_response"].present?
```
caitbot, line 16: A comment explaining how this category of error is different from the one above might be nice. As AL and I understand it, it seems our methods of storing data on failed bundle issues diverged (one being `error_code` and the other being `raw_response`) and maybe it would make more sense to unify them into one schema.

caitbot: We looked in the code, it looks like the `usps` errors are the only ones that lack a `raw_response` and instead only have `error_code`

caitbot: An alternative that is a bit more idiomatic is to just check `error_code == BUNDLE-FAILURE`

and maybe rename the `persist_errors_from_error_metadata` to be something like `persist_usps_errors`

This is all nonblocking. (:

caitbot (approved):
AL + I approve and had some thoughts but they are non-blocking and this is good (:

PR #1625, opened 2021-09-14
Remove None of the above as an option for Spouse SSN

caitbot:
We had "None of the above" as an option for spouse identification. It caused a crash when you selected it. See [the story.](https://www.pivotaltracker.com/story/show/179542073)

### DIFF
--- a/app/views/ctc/questions/spouse_info/edit.html.erb
+++ b/app/views/ctc/questions/spouse_info/edit.html.erb
@@ -18,7 +18,7 @@
             help_text: t("hub.clients.show.date_of_birth_help"),
             classes: ["ctc-intake-date-text-input"]
           ) %>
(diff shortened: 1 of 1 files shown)


bytheway875:
WOWWWWWW

bytheway875:
wait... 

do we need to allow people to like offboard in some way if they have a spouse but the spouse doesnt have an SSN or ITIN?

My concern is people choose MFJ, get here, realize the spouse doesnt have this data, gets stuck.

Probably not to get fixed here, but is it something we need to worry about?

alaroia-cfa (approved):
@caitbot I'm happy to approve this good simple code change. If you want to communicate to Yvonne more, then fine with me; I'll hold off of merging.

PR #1629, opened 2021-09-15
Add capability to release a specific commit [#179592937]

alaroia-cfa:
This is not directly related to rollback. It allows easier incremental deploy of what's already on main.


alaroia-cfa:
I'm going to self-merge this since it doesn't affect the actual content of what's released to clients.

PR #1630, opened 2021-09-15
Validate SSN based on itin type on CTC client update form

caitbot:


embarnard (approved):
Looks good! Just remove that puts from the test

spec/forms/hub/update_ctc_client_form_spec.rb
```diff
@@ -54,22 +54,34 @@
...
-          form_attributes[:primary_ssn] = "111-22-7777"
-        end
-
         it "persists valid changes to ssn" do
           expect do
             form = described_class.new(client, form_attributes)
             form.save
             intake.reload
-          end.to change(intake, :primary_ssn).to("111227777").and change(intake, :primary_tin_type).to("ssn_no_employment")
+            puts form.errors.full_messages
```
embarnard, line 70: should we remove this puts message?

caitbot: woops good catch, thank you!

PR #1633, opened 2021-09-15
Fix portal button styling weirdness

caitbot:
image
image

### DIFF
--- a/app/views/ctc/portal/portal/home.html.erb
+++ b/app/views/ctc/portal/portal/home.html.erb
@@ -58,29 +58,27 @@
             <h3><%= t("views.ctc.portal.home.next_step_title") %></h3>
             <div class="text--small">
               <%= I18n.t("views.ctc.portal.home.imperfect_return_indicator_text") %>
-            <div>
+            </div>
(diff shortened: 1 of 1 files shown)


bytheway875:
oh, i like this much better anyway moving this second cta out of the box.

One thing to consider is that the root of this bug is that i moved the button location around in the code because the original location was relying on an `exposed_error` object in order to display the correct my information button. The OG thought being that if we're like "we dont know how to deal with this error or theres nothing for you to do" we dont show that button.

I moved the button to deal with the bug that didn't have a linked error for bank account stuff, so those clients were locked out of editing.

If we don't revert that change, people with bundle failures who failed because of timezone issues, or people who have sort of unexplored errors we mark as "dont expose to client" will still be able to update their information.

I don't have a clear answer on what is "better", and when I spoke to Yvonne she didn't have a clear way she wanted/needed it to work, so either approach is fine and I leave it to your judgement to decide which implementation we should go with. Feel free to tuple me if none of this made sense.

bytheway875 (approved):

PR #1636, opened 2021-09-16
Add recaptcha v3 to intake flow and efile security information.

caitbot:
image


bytheway875 (approved):
This looks good to me. I was going back and forth on thinking whether there should be some "placeholder" value for when we don't have a value for the recaptcha, but I think its fine as-is...

PR #1638, opened 2021-09-16
Add money claims to tax return 179583923

bytheway875:
Adds an accepted_tax_returns_analytics table where we can record finalized claims for accepted tax returns!

### DIFF
--- a/app/models/tax_return.rb
+++ b/app/models/tax_return.rb
@@ -4,6 +4,8 @@
 #
 #  id                  :bigint           not null, primary key
 #  certification_level :integer
+#  ctc_amount_cents    :bigint
+#  eip3_amount_cents   :bigint
 #  filing_status       :integer
 #  filing_status_note  :text
 #  internal_efile      :boolean          default(FALSE), not null
@@ -13,6 +15,7 @@
 #  primary_signed_at   :datetime
 #  primary_signed_ip   :inet
 #  ready_for_prep_at   :datetime
+#  refund_amount_cents :bigint
 #  service_type        :integer          default("online_intake")
 #  spouse_signature    :string
 #  spouse_signed_at    :datetime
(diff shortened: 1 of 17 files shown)


config/initializers/inflections.rb
```diff
@@ -13,4 +13,5 @@
 # These inflection rules are supported but not enabled by default:
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.acronym 'SLA'
+  inflect.irregular "analytics", "analytics"
```
embarnard, line 16: should one of these be "analytic"?

embarnard (approved):
Looks good! Have a few comments

app/models/tax_return.rb
```diff
@@ -143,6 +143,16 @@ def advance_to(new_status)
     update!(status: new_status) if TaxReturn.statuses[status.to_sym] < TaxReturn.statuses[new_status.to_sym]
   end
 
+  def record_expected_payments!
+    raise StandardError, "Cannot record payments on tax return that is not accepted" unless status == "file_accepted"
+
+    create_accepted_tax_return_analytics!(
```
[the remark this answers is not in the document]
bytheway875: money is an odd value, because they're technically decimal values, but they're decimal values that are always out of 100 and should never "round" the way that typical decimal values do.

By storing as cents, you're able to store the money value as an integer and then convert it to dollars for display values, which helps avoid peksy rounding errors when doing money calculations.

this is a pretty good resource for understanding why we do this from a computer science and engineering POV: https://blog.agentrisk.com/you-better-work-in-cents-not-dollars-2edb52cdf308

The reason we haven't run into this that much is that we actually don't store that many money values right now, and when we do they've typically been whole values.

With this table, if we intend to store refund data from returns that are not always whole number refunds (really anything but stimulus only, like these are), we'd need to store parts of the whole for accuracy.

config/initializers/inflections.rb
bytheway875: whats happening here is that im overriding the analytics singular to also be analytices, because the model name should indicate one SET of analytics.

PR #1640, opened 2021-09-16
CTC intake: combine the three direct deposit account info pages into one

tgrathwell:
Shares the same form class with the Portal version because they have the exact
same validations

Temporarily keeping the old controllers around so they can route to the new one


app/controllers/ctc/portal/bank_account_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Ctc::Portal::BankAccountController < Ctc::Portal::BaseIntakeRevisionController
-
   def edit
-    @form = form_class.new(current_intake)
+    # Deliberately present the form a blank bank account so as to not leak sensitive details
+    @form = form_class.from_bank_account(BankAccount.new)
```
embarnard, line 4: will it save the blank bank account for this client if they don't add any information?

app/controllers/ctc/questions/account_number_controller.rb
```diff
@@ -3,17 +3,21 @@ module Questions
     class AccountNumberController < QuestionsController
       include AuthenticatedCtcClientConcern
 
-      layout "intake"
+      def self.deprecated_controller?
```
embarnard, line 6: why not just get rid of the controllers instead of setting them as deprecated?

app/lib/ctc_question_navigation.rb
```diff
@@ -53,9 +53,13 @@ class CtcQuestionNavigation
 
     # Bank and mailing info
     Ctc::Questions::RefundPaymentController,
+
+    # DEPRECATED - remove soon
```
embarnard, line 57: okay-- I see, this is something we are going to do later?

app/views/ctc/bank_account/edit.html.erb
```diff
@@ -0,0 +1,42 @@
+<% @main_question = t("views.ctc.questions.bank_account.title") %>
+
+<% content_for :page_title, @main_question %>
+
+<% content_for :card do %>
+  <div class="question-layout">
+    <div class="back">
```
embarnard, line 7: hmm shouldn't this be included in most layouts?

spec/controllers/ctc/questions/bank_account_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  describe "#update" do
+    context "with no answer" do
+      let(:params) do
+        {}
+      end
+
+      it "re-renders the form with errors" do
+        post :update, params: params
```
embarnard, line 17: `post :update, params: {}` could we just do this and remove the `let(:params)` for this context

app/controllers/ctc/portal/bank_account_controller.rb
embarnard: Never mind! I see we have validation for not allowing blank fields

embarnard (approved):
Looks good! I'm just wondering why we are keeping the "deprecated" controllers instead of just removing them

PR #1653, opened 2021-09-20
Efiling: pad timezones like "-60" with zeroes to match schema

tgrathwell:
they must have at least three digits, e.g. "-060"

### DIFF
--- a/app/lib/submission_builder/formatting_methods.rb
+++ b/app/lib/submission_builder/formatting_methods.rb
@@ -96,6 +96,17 @@ module SubmissionBuilder
       Phonelib.parse(string, "US").national(false)
     end
 
+    # Timezones like '-60' need to be padded to '-060' to validate schema
+    # These timezones are all a little suspicious, but it's better to handle
+    # their suspiciousness somewhere else rather than crash on bundle
+    def time_zone_offset_type(string)
+      if (match = string.match(/\A([+-])([0-9]{1,2})\z/))
+        "#{match[1]}#{"%03d" % match[2]}"
+      else
+        string
+      end
+    end
+
     private
 
     def formatted_first_name(name)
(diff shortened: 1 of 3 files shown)


caitbot (approved):
This looks good to me but I wonder if this means we will be submitting "fraud-like" returns to the IRS (whereas at least now, we fail, albeit loudly). Does product have a plan for handling these suspicious looking values in the future?

tgrathwell:
there's a story a couple down in the backlog to put a fraud hold on anything outside of US timezones https://www.pivotaltracker.com/story/show/179544481

I dunno if anyone is tracking the distribution of timezones we're currently getting but here's a very very quick and dirty table of the most seen time zones so far

irb(main):007:0> countz = EfileSecurityInformation.group(:timezone_offset).count
irb(main):006:0> countz.sort_by { |k, v| v }.reverse.each { |pair| puts "| #{pair.first} | #{pair.last} |" }
| offset | count |
| - | - |
| +240 | 129524 |
| +300 | 82334 |
| +420 | 52008 |
| +360 | 11110 |
| +600 | 1363 |
| -600 | 405 |
| +480 | 337 |
| -60 | 229 |
| -480 | 118 |
| -180 | 93 |
| -120 | 78 |
| +0 | 57 |
| -420 | 45 |
| +180 | 36 |
| -540 | 24 |
| -330 | 23 |
| -300 | 11 |
| +660 | 10 |
| -060 | 8 |
| -240 | 7 |
| -720 | 4 |
| -660 | 2 |
| +60 | 2 |
| +540 | 2 |
| -345 | 2 |
| -840 | 1 |
| -360 | 1 |
| +154 | 1 |
| +150 | 1 |
| +137 | 1 |

a lot of those we maybe ought to be treating with equal suspicion as the `-60` but we haven't been looking at them much because they don't crash

PR #1658, opened 2021-09-21
Add fraud indicator to security tab

caitbot:
Style follow up for [this story](https://www.pivotaltracker.com/story/show/179367886)

I also updated the recaptcha display on the tab to display separate from device data so that its a little prettier.

image


config/locales/en.yml
```diff
@@ -668,7 +668,7 @@ en:
         client_messages: Messages
         client_notes: Notes
         client_profile: Client Profile
-        client_security: Security
+        client_security_html: "<div class='security-wrapper'>Security %{fraud_icon}</div>"
```
caitbot, line 671: This is kinda weird. I would prefer not to do it this way but since we have to use that `tab_navigation_link` to keep styling similar to other tabs and it doesn't allow you to provide other classes, this seemed like the simplest way.

I initially attempted to wrap the tab navigation link in this div, but that made the styling weird. ):

app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -301,6 +301,12 @@
 
   .client-navigation {
     @include no-bleed();
+    .security-wrapper{
```
jenny-heath, line 304: the tiniest nitpick ever is that there should be a space between the class name and the bracket! 🦐

app/views/hub/clients/_navigation.html.erb
```diff
@@ -7,6 +7,7 @@
     <%= tab_navigation_link(t(".client_efile"), efile_hub_client_path(id: @client, anchor: "body")) %>
   <% end %>
   <% if @client.efile_security_informations.size.nonzero? %>
-    <%= tab_navigation_link(t(".client_security"), security_hub_client_path(id: @client)) %>
+    <% fraud_icon = image_tag("security-notification.svg") if @client.efile_security_informations.any?(&:fraud_suspected?) %>
```
jenny-heath, line 10: typically i prefer to keep logic out of views. in part i find it easier to read but it also then means you have the option to write a unit test without using render_views. not blocking!

config/locales/en.yml
jenny-heath: i wonder if `tab_navigation_link` could be refactored into a partial instead of a helper? i haven't thought about that too hard and it's a little out of scope so i'd say this is fine unless you feel particularly enthused about pursuing a refactor

jenny-heath (commented):
left a few nonblocking comments! the only blocking thing is that this adds new functionality that includes some logic (though very simple) so it would be good to have a test (no preference re: feature or unit test)

app/views/hub/clients/_navigation.html.erb
caitbot: oh I like that idea, I'll move it to the helpers.

PR #1661, opened 2021-09-22
Add sentry alert for IncomingEmails with blank bodies and no attachments

caitbot:
Investigating [this story](https://www.pivotaltracker.com/story/show/178990723), found that we have gotten 101 IncomingEmails with blank bodies and no attachments since this event started firing. This should quiet that event and add better logging for when we get a blank email. I am curious to see if this event will completely eliminate the previous event, but occur at a similar pace.


embarnard (approved):
Looks good!

PR #1666, opened 2021-09-23
Disable the resubmit button after the initial button press

caitbot:
This just makes it so that after the user presses the button, they can't press it again while it loads the next page.

We already did this to the button on the confirm-legal page, going to do it here because it is causing timeout/duplicate errors for Recaptcha (no user impact, but noisy on Sentry)


tgrathwell (approved):

tgrathwell:
I am in favor of any `disable_with` additions

I thought there was something on GCF we did to ensure that every button had a `disable_with` but all I can find is this ancient commit in the form builder that adds it to _some_ buttons https://github.com/codeforamerica/gcf-backend/commit/071ae817894c9ca0884725e86db69c731654b1ce so who knows

PR #1676, opened 2021-09-27
Update still needs help flow to fit need for october 2021 closing

bytheway875:


config/locales/en.yml
```diff
@@ -1109,6 +1109,15 @@ en:
         sent_at: Sent at %{datetime}
       title: Invitations
   messages:
+    closing_soon:
+      email:
+        body: 'We are closing GetYourRefund for the tax season on October 15th. We will need all required documentation by October 8th; otherwise, we will not be able to prepare your taxes. Please login and respond by October 8th. If we do not hear from you by then, we will close out your case and you can file with us next year.
```
alaroia-cfa, line 1114: Dumb question - are we sure that we want the trailing newlines in this string? Not a blocker; just curious.

```
irb(main):003:0> I18n.t("messages.closing_soon.email.body", locale: "en").last(1)
=> "\n"
```

app/services/still_needs_help_service.rb
```diff
@@ -5,7 +5,26 @@ def self.must_show_still_needs_help_flow?(client)
...
+
+  def self.send_still_need_help_notification(client)
+    ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
+      client: client,
+      locale: client.intake.locale,
+      message: AutomatedMessage::ClosingSoon.new
+    )
+  end
+
+  def self.may_need_help_statuses
```
alaroia-cfa, line 21: `intake_ready` seems to have disappeared from this list. Is that intentional?

app/services/still_needs_help_service.rb
```diff
@@ -5,7 +5,26 @@ def self.must_show_still_needs_help_flow?(client)
...
-    client.tax_returns.where(status: %w[intake_ready intake_in_progress]).update(status: :file_not_filing)
+    client.tax_returns.where(status: may_need_help_statuses).update(status: :file_not_filing)
     client.save
+    send_still_need_help_notification(client)
+  end
+
+  def self.send_still_need_help_notification(client)
+    ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
+      client: client,
+      locale: client.intake.locale,
```
alaroia-cfa, line 16: I'd like to see a change someday where `send_system_message_to_all_opted_in_contact_methods` always grabs `locale` from `client.intake.locale`. Is there a scenario where that would be a mistake?

Not a blocker for this PR, naturally. Might be a nice refactor after this gets merged.

Related: I'd like to see `send_system_message_to_all_opted_in_contact_methods` use `I18n.with_locale(client.intake.locale) do ... end`, per https://translation.io/blog/rails-i18n-with-locale

app/services/still_needs_help_service.rb
```diff
@@ -5,7 +5,26 @@ def self.must_show_still_needs_help_flow?(client)
...
+      message: AutomatedMessage::ClosingSoon.new
+    )
+  end
+
+  def self.may_need_help_statuses
+    %w[intake_info_requested intake_in_progress intake_greeter_info_requested intake_needs_doc_help]
+  end
+
+  def self.clients_who_still_may_need_help(intake_type: nil)
```
alaroia-cfa, line 25: This is only used within a Rake task. I'd like to see it inlined into the Rake task.

If you don't like that idea, consider converting `intake_type` into a mandatory parameter, since the only call site always provides it; consider covering the method with a unit test.

Not a blocker.

alaroia-cfa (approved):
I have a couple of comments I'd like your input on; could Tuple or Slack or whatever; none are strictly blockers.

This is a good change!

app/services/still_needs_help_service.rb
bytheway875: per their ticket, yes.

app/services/still_needs_help_service.rb
bytheway875: not sure, but that feels outside the scope of the ticket here.

config/locales/en.yml
bytheway875: This is what normalize does. I'm not sure why I can't use the pipe to create a multiline string without trailing whitespace now that we're using normalize.

PR #1678, opened 2021-09-28
Add timezone to efile security informations

bytheway875:
Adds timezone to efile security informations so that we can capture timezone in a format we'll be able to interpret as a part of `TimeZone.us_timezones` or whatever the method rails provides for timezone support.

### DIFF
--- a/app/views/ctc/questions/income/edit.html.erb
+++ b/app/views/ctc/questions/income/edit.html.erb
@@ -9,18 +9,3 @@
     <% end %>
   </ul>
 <% end %>
-
-<% content_for :script do %>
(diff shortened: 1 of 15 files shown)


app/views/ctc/questions/income/edit.html.erb
```diff
@@ -9,18 +9,3 @@
     <% end %>
   </ul>
 <% end %>
-
-<% content_for :script do %>
```
bytheway875, line 13: Don't need this anymore because we're picking up the timezone on the efile security information, and then grabbing from there to set on the intake.

PR #1679, opened 2021-09-28
Redirect GYR.org/ca and GYR.org/cdss need to redirect to DIY

embarnard:

### DIFF
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -55,6 +55,16 @@ class PublicPagesController < ApplicationController
 
   def diy; end
 
+  def cdss
+    session[:source] = "cdss"
+    redirect_to diy_path
+  end
+
+  def ca
+    session[:source] = "ca"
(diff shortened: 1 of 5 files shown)


app/controllers/public_pages_controller.rb
```diff
@@ -55,6 +55,16 @@ def sms_terms; end
 
   def diy; end
 
+  def cdss
```
embarnard, line 58: For whoever is reviewing, we were not sure this was the right protocol for adding a source. Do we want to save DIY sources separately from homepage sources?

bytheway875: Setting the source directly like this is a little strange to me. I think it would make more sense to _not_ add these new routes, but to rather detect the source param in "home" and redirect to "diy" if source param is "cdss" or "ca"

mquadri23: @bytheway875 good idea! How does this change look now? Is it okay that we added it to set_source?

PR #1682, opened 2021-09-28
Improve release script output

alaroia-cfa:
We use `--quiet` for all git operations, since we don't need the usual git output unless something goes wrong.

`git push --quiet` will print info if an error occurs [1]. 

For `git checkout` being used to switch branch, it's OK if we see less info [2]. Same for `git merge` [3]. It will still fail if it needs to fail, and failures will cause the script to exit due to [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

I also removed some newlines.

1. https://git-scm.com/docs/git-push#Documentation/git-push.txt---quiet
2. https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---quiet
3. https://git-scm.com/docs/git-merge#Documentation/git-merge.txt---quiet

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -86,8 +86,14 @@ function cleanup() {
 trap cleanup EXIT
 
 echo "Updating local release branch."
-git checkout release
-git merge --ff-only "${new_release}" || {
+git checkout --quiet release
+git merge --quiet --ff-only "$origin_git_remote"/release || {
+  # merge failed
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
Self-merged b/c it doesn't affect anything on the backend or for clients.

PR #1683, opened 2021-09-29
Use lack of timezone or international timezone as a fraud marker.

bytheway875:


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -45,6 +45,7 @@ def efile_security_params
                                               :browser_language,
                                               :platform,
                                               :timezone_offset,
+                                              :timezone,
```
bytheway875, line 48: I missed adding the timezone to the ESI when we capture it in the portal.

embarnard (approved):

PR #1684, opened 2021-09-29
Do not flag ctc intakes just because they are completed

bytheway875:
We do this for GYR intakes, but shouldnt do it for CTC intakes.


embarnard (approved):
Looks good!

PR #1685, opened 2021-09-29
Fix word breaks on review box globally

bytheway875:
Screen Shot 2021-09-29 at 1 08 37 PM
Screen Shot 2021-09-29 at 1 14 01 PM
Screen Shot 2021-09-29 at 1 18 23 PM
Screen Shot 2021-09-29 at 1 18 32 PM


alaroia-cfa (approved):
Super rad.

PR #1686, opened 2021-09-29
Offseason 2022

bytheway875:
https://www.pivotaltracker.com/story/show/179703410

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -231,7 +231,7 @@ class ApplicationController < ActionController::Base
   end
 
   def open_for_intake?
-    true
+    !Rails.configuration.offseason
   end
 
   private
--- a/app/controllers/ctc/questions/questions_controller.rb
+++ b/app/controllers/ctc/questions/questions_controller.rb
@@ -2,6 +2,8 @@ module Ctc
   module Questions
     class QuestionsController < ::Questions::QuestionsController
       helper_method :wrapping_layout
+      skip_before_action :unauthenticated_redirect_in_offseason
+
 
       private
 
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -1,7 +1,7 @@
 module Questions
   class QuestionsController < ApplicationController
+    before_action :unauthenticated_redirect_in_offseason
     before_action :set_current_step
-
     delegate :form_name, to: :class
     delegate :form_class, to: :class
 
@@ -87,6 +87,12 @@ module Questions
     end
 
     private
+
+    def unauthenticated_redirect_in_offseason
+      if Rails.configuration.offseason && !current_client
+        redirect_to root_path and return
+      end
+    end
 
     def after_update_success; end
 
--- a/config/environments/demo.rb
+++ b/config/environments/demo.rb
@@ -13,7 +13,7 @@ Rails.application.configure do
   }
   config.action_mailer.default_url_options = { host: 'demo.getyourrefund.org' }
   config.action_mailer.asset_host = config.gyr_url
-  config.offseason = false
+  config.offseason = true
   config.hide_ctc = false
 
   Rails.application.default_url_options = config.action_mailer.default_url_options
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -98,7 +98,9 @@ Rails.application.routes.draw do
 
       resources :dependents, only: [:index, :new, :create, :edit, :update, :destroy]
 
-      get "/sign-up", to: redirect('/')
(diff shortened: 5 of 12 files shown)


app/controllers/questions/questions_controller.rb
```diff
@@ -88,6 +88,12 @@ def form_navigation
 
     private
 
+    def unauthenticated_redirect_in_offseason
+      if Rails.configuration.offseason && !current_client
+        redirect_to root_path and return
```
alaroia-cfa, line 93: `and return` isn't really necessary here, I think -- it's the final expression of a method, so it'll be a `return` anyway.

app/controllers/application_controller.rb
```diff
@@ -231,7 +231,7 @@ def show_offseason_banner?
   end
 
   def open_for_intake?
```
alaroia-cfa, line 233: I would prefer to remove this method, and inline it into the call sites. I find it confusing to have _both_ the config option _and_ the method, and when reviewing this, I've had to jump back and forth a lot to try to remember if this method has any special logic, but it doesn't.

Not a blocker.

config/routes.rb
```diff
@@ -98,7 +98,9 @@ def scoped_navigation_routes(context, navigation)
 
       resources :dependents, only: [:index, :new, :create, :edit, :update, :destroy]
 
-      get "/sign-up", to: redirect('/')
+      resources :signups, only: [:new, :create], path: "sign-up", path_names: { new: '' } do
```
alaroia-cfa, line 101: What's the significance of this change and the change on the line below?

config/environments/demo.rb
```diff
@@ -13,7 +13,7 @@
   }
   config.action_mailer.default_url_options = { host: 'demo.getyourrefund.org' }
   config.action_mailer.asset_host = config.gyr_url
-  config.offseason = false
+  config.offseason = true
```
alaroia-cfa, line 16: I'd like to see us set `config.offseason = true` for the test environment, too. That way, the test environment is close to demo/dev.

Maybe that'll be too much of a hassle? OK, maybe forget it.

app/controllers/questions/questions_controller.rb
alaroia-cfa: I think I'd like to see some unit test coverage of this method.

app/controllers/ctc/questions/questions_controller.rb
```diff
@@ -2,6 +2,8 @@ module Ctc
   module Questions
     class QuestionsController < ::Questions::QuestionsController
       helper_method :wrapping_layout
+      skip_before_action :unauthenticated_redirect_in_offseason
```
alaroia-cfa, line 5: Will this code prevent clients from signing in? `Ctc::Portal::BaseAuthenticatedController` may need the same treatment.

alaroia-cfa (commented):
I'm concerned that `unauthenticated_redirect_in_offseason` goes too far and prevents sign-in. I haven't validated that fear. Let me know what you think. I have some other comments, too.

app/controllers/questions/questions_controller.rb
bytheway875: 1) my understanding was that explicit returns after a redirect were required in before_actions so that you're indicating that you shouldn't move forward with the behavior in the next method. redirect_to does not automatically stop the other controller action from being called.

config/routes.rb
bytheway875: It adds back the gyr signup flow routes instead of redirecting signups to root path.

config/environments/demo.rb
bytheway875: I think that would break some tests, because the whole point of this is to keep testing the "on season flow" and then add tests that check for the offseason.

app/controllers/ctc/questions/questions_controller.rb
bytheway875: Only if this before action happens _before_ we look to see if someone is authenticated and redirect them. My thought was that would already be happening at a higher level.

bytheway875:
@alaroia-cfa Some context here -- I picked up this ticket after expressing to the team that I would be out today, but didn't seem like anyone else was in a rush to pick it up either. It needs to be done and released tomorrow for close. Can you please ensure that if there are blocking changes that need to get done that someone on the team picks it up and makes those changes today?

app/controllers/application_controller.rb
bytheway875: That's fine -- I'm trying to make the fewest changes necessary based on the original implementation because there's a tight timeline to get this done.

alaroia-cfa:
Great! Getting this tested on demo will be super helpful. I hugely appreciate that you set us up for success by working on this before you are gone! :D I also appreciate your answers to my pull request comments.

PR #1691, opened 2021-10-04
Retry unauthorized efile #179805956

jenny-heath:
[Sentry alert](https://sentry.io/organizations/codeforamerica/issues/2618857779/events/d6df8ee982d640a8a05eabd24a6f2baa/?project=1880327&referrer=pivotal_plugin)

### DIFF
--- a/app/services/efile/gyr_efiler_service.rb
+++ b/app/services/efile/gyr_efiler_service.rb
@@ -42,6 +42,8 @@ module Efile
             raise RetryableError, log_contents
           elsif log_contents.match(/connect timed out - Fault Code: soap:Server/)
             raise RetryableError, log_contents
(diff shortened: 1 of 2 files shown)


alaroia-cfa (approved):
🌟

app/services/efile/gyr_efiler_service.rb
```diff
@@ -3,6 +3,11 @@ module Efile
   class GyrEfilerService
     CURRENT_VERSION = 'f5eeb816f6c919fff8d5c742062e664b1f4cd13a'
     POSTGRES_LOCK_PREFIX = 1640661264
+    RETRYABLE_LOG_CONTENTS = [
```
bytheway875, line 6: As this list grows, my suggestion would be to create a .yml file for this and load it in! (See: the way we create an array for timezone overrides and/or the way we load in zip codes.

PR #1692, opened 2021-10-04
Consecutive spaces bundle failure 179785272

mquadri23:
We considered the following options for this:
- removing extra spaces in each form where primary/spouse legal names are updated
- removing extra spaces in a `before_save` hook on the model level (<- we went with this one)
- presenting the user with a validation error when they include extra spaces
- removing extra spaces just before sending the submission bundle

With both the `before_save` and the modifying at bundle-time, it seems like there's a risk of this being "hidden" from the developer / hub user. Would love to hear thoughts on minimizing the risk of creating an invisible step when troubleshooting either a bug or a particular user's case (this seems like it should be the main consideration here).

Re: the actual lines of code: we could do something less fancy than the current loop (for example, one line per attribute) that would make it more readable. Also welcome thoughts on this!

### DIFF
--- a/app/models/intake/ctc_intake.rb
+++ b/app/models/intake/ctc_intake.rb
@@ -288,16 +288,13 @@ class Intake::CtcIntake < Intake
   end
 
   before_save do
-    # TODO: only normalize spacing when these attributes are being assigned/changed
-    [:primary_first_name, :primary_last_name, :spouse_first_name, :spouse_last_name].each do |attribute|
+    attributes_to_change = self.changes_to_save.keys
+    name_attributes = ["primary_first_name", "primary_last_name", "spouse_first_name", "spouse_last_name"]
+
+    (attributes_to_change & name_attributes).each do |attribute|
       new_value = self.send(attribute).split(/\s/).filter { |s| !s.empty? }.join(" ")
       self.assign_attributes(attribute => new_value)
     end
-
-    # self.primary_first_name = self.primary_first_name.split(/\s/).filter { |s| !s.empty? }.join(" ")
-    # self.primary_last_name = self.primary_last_name.split(/\s/).filter { |s| !s.empty? }.join(" ")
-    # self.spouse_first_name = self.spouse_first_name.split(/\s/).filter { |s| !s.empty? }.join(" ")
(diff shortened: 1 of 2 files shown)


app/models/intake/ctc_intake.rb
```diff
@@ -287,6 +287,18 @@ class Intake::CtcIntake < Intake
     self.spouse_ssn = self.spouse_ssn.remove(/\D/) if spouse_ssn_changed? && self.spouse_ssn
   end
 
+  before_save do
+    attributes_to_change = self.changes_to_save.keys
+    name_attributes = ["primary_first_name", "primary_last_name", "spouse_first_name", "spouse_last_name"]
+
+    (attributes_to_change & name_attributes).each do |attribute|
+      if self.send(attribute).present?
```
bytheway875, line 295: is the presence check here to guard against nil? because seems like attributes_to_change is already filtering to present attributes. 

_If_ the reason is to prevent against running this on attributes that are blank or nil, maybe `unless self.send(attribute).blank?` might be more clear, not sure.

app/models/intake/ctc_intake.rb
```diff
@@ -287,6 +287,18 @@ class Intake::CtcIntake < Intake
     self.spouse_ssn = self.spouse_ssn.remove(/\D/) if spouse_ssn_changed? && self.spouse_ssn
   end
 
+  before_save do
```
alaroia-cfa, line 290: I would rather this be a `before_validation` hook. That allows you to change the data as early as possible, which means the transformed data will also be what any validation sees. Not a blocker.

alaroia-cfa (approved):
This is great; I left one non-blocking comment.

app/models/intake/ctc_intake.rb
alaroia-cfa: Can this be `self.attributes[attribute]` rather than `self.send(attribute)`? I have a preference for minimizing use of `send` if something else will do the job. Not a blocker.

alaroia-cfa:
> Re: the actual lines of code: we could do something less fancy than the current loop (for example, one line per attribute) that would make it more readable. Also welcome thoughts on this!

I'm 51% in favor of the loop, and 49% in favor of one line per attribute.

PR #1694, opened 2021-10-04
Upgrade gnutls and root certificates [#179809550]

alaroia-cfa:
Based on https://github.com/nodesource/distributions/issues/1266 and https://letsencrypt.org/certificates/ and the current CA chain of nodesource.com, this would fix it. I've also tested that staging couldn't deploy before this change, and can after this change.

It makes sense to me why ca-certificates had to be updated. I'm surprised gnutls also had to be, but oh well.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update --allow-releaseinfo-change
 
 # System prerequisites
 RUN apt-get update \
- && apt-get -y install build-essential libpq-dev pdftk-java ghostscript poppler-utils curl \
+ && apt-get -y install ca-certificates libgnutls30 build-essential libpq-dev pdftk-java ghostscript poppler-utils curl \
  && curl -sL https://deb.nodesource.com/setup_12.x | bash - \
  && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
  && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \


bytheway875:
Looks good! Wondering how you were able to debug this and determine what needed to be updated? I did a google search for the error that was in red text on CircleCI but it didn't really get me anywhere specific...

alaroia-cfa:
I saw this message in the CircleCI logs:

> Certificate verification failed: The certificate is NOT trusted. The certificate chain uses expired certificate. Could not handshake: Error in the certificate verification. [IP: 23.222.5.155 443]

I saw it was in the context of processing the APT package repo at https://deb.nodesource.com/node_12.x .

> remote: INFO -- : Err:5 https://deb.nodesource.com/node_12.x buster Release        
> remote:   Certificate verification failed: The certificate is NOT trusted. The certificate chain uses expired certificate.  Could not handshake: Error in the certificate verification. [IP: 23.222.5.155 443]        
> remote: INFO -- : Reading package lists...        
> remote: INFO -- :         
> remote: INFO -- : E: The repository 'https://deb.nodesource.com/node_12.x buster Release' does not have a Release file.        

This told me that there was something wrong with the TLS certificate setup for that specific URL.

I wondered about a few things:

* Does the website itself have an expired certificate?
* Does the _signer_ of the website's cert have an expiry problem? What about the signer of the signer? I know that a website's certificate must be signed by a trusted entity, and sometimes the trusted entity actually signs an "intermediate" certificate which signs the website's certificate.
* Could I perceive this in my web browser? I know that browsers will show a broken padlock situation and let me navigate the hierarchy of the site's certificate, its signer, and any hierarchical signers all the way up to a trusted root.
* Does this relate to a [thing I read about on Hacker News a week ago, namely that Let's Encrypt has a root certificate expiry thing going on](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/)?

I visited https://deb.nodesource.com/ and things looked totally normal in the browser -- solid padlock. This told me the website itself didn't have an expired certificate.

I re-read the error message, and I saw that the "chain" has expiry problems. This suggests to me that a signer, or a signer's signer, is expired.

I looked again at what I saw in Chrome on macOS. The root is not expired, and there is an intermediate cert, but it's not expired either.

I didn't have a super solid theory yet, but I also knew that `ca-certificates` in Debian/Ubuntu is the package that contains the list of root certificates used by the operating system. Presumably, APT (aka apt-get) uses this for its trust roots. I knew that there are a few (OK probably like 100) root certificates, and that a certificate's expiration can be changed without the key changing. I also know that when you make a HTTPS connection, websites typically don't send the root certificate they use; they instead send the intermediate and their own. I wondered if this website used a root had an expiry in the past, and it had gotten renewed, and Debian had an updated ca-certificates package except we happened not to install it as part of our build process. Based on this theory, I added this patch to the `Dockerfile`:

```
diff --git a/Dockerfile b/Dockerfile
index 2171a9fbe..f91628595 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM ruby:2.6.5

 # The Docker environment is based on Debian buster, which used to be called stable Debian, but is now called oldstable.
 RUN apt-get update --allow-releaseinfo-change
+# Ensure root certificates are fresh #179809550
+RUN apt-get install ca-certificates

 # System prerequisites
 RUN apt-get update \
```

To test it, I did a force push to staging. I also launched Docker desktop on my workstation so I could test locally.

This didn't seem to be enough -- the error persisted. I then went and read [more info from Let's Encrypt](https://letsencrypt.org/docs/certificate-compatibility/), and saw that the situation wasn't that a root certificate had expired _per se_, but that the intermediate (aka R3) is signed by two certificates -- one root that's now expired, and another root. I didn't go check the name of the Debian release we use, but the behavior we were seeing was consistent with a situation where we were failing to use the _valid_ root. Perhaps the valid root wasn't in ca-certificates until we updated it... but then, why would APT not be satisfied?

My primary goal was to solve the problem. I Googled [nodesource "Certificate verification failed: The certificate is NOT trusted."] and found this discussion here: https://github.com/nodesource/distributions/issues/1266

At that discussion, I did a Cmd-F for `ca-certificates`, hoping that other people had gone down the path I had gone down. Some had, and one person said they updated `libgnutls30` as well. I tried that with a build in `docker build .` and saw it got past APT, so I did a push to staging and prepared a pull request while waiting for those to fully succeed.

## Appendix

After writing the above analysis, it struck me that the bad behavior could be caused by a bug in the TLS library. One possible scenario is where the TLS library, when given the website's certificate plus the intermediate, greedily finds the first root certificate with a trust chain (presumably root -> intermediate -> website cert). If the root it found in that trust chain is an expired root, then the TLS library would declare failure. This would be buggy behavior because a better option is to look for _all_ trust chains, filtering down to just the valid ones, and then declaring failure if there are _no_ valid ones.

To test this theory, I did `git checkout HEAD^` and made sure `docker build .` printed an error message. Good! Then I added this patch to `Dockerfile`.

```
diff --git a/Dockerfile b/Dockerfile
index 2171a9fbe..ea92003a6 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update --allow-releaseinfo-change

 # System prerequisites
 RUN apt-get update \
- && apt-get -y install build-essential libpq-dev pdftk-java ghostscript poppler-utils curl \
+ && apt-get -y install libgnutls30 build-essential libpq-dev pdftk-java ghostscript poppler-utils curl \
  && curl -sL https://deb.nodesource.com/setup_12.x | bash - \
  && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
  && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
```

After this, I was able to complete the build! This suggests that upgrading libgnutls30 alone does the job. I checked with `DOCKER_BUILDKIT=false docker build .` that `ca-certificates` wasn't being upgraded by this procedure; it wasn't. (Feel free to ask for more details on this.)

I wondered if there was info in the Debian changelog for libgnutls30 about this. I went to https://packages.debian.org/libgnutls30 and clicked on "oldstable" and clicked on "changelog" and found myself at https://metadata.ftp-master.debian.org/changelogs//main/g/gnutls28/gnutls28_3.6.7-4+deb10u7_changelog .

After extensive study, I believe this changelog entry fixed it:

```
gnutls28 (3.6.7-4+deb10u5) buster; urgency=medium

  * 42_rel3.6.11_10-session-tickets-parse-extension-during-session-resum.patch
    from GNUTLS 3.6.11: Fix TL1.2 resumption errors. Closes: #956649
  * 47_rel3.6.13_10-session_pack-fix-leak-in-error-path.patch from GNUTLS
    3.6.14: One line fix for memory leak. Closes: #958704
  * Rename
    44_rel3.6.14_01-stek-differentiate-initial-state-from-valid-time-win.patch
    (security upload) to 44_rel3.6.14_90_... to be able to pull earlier fixes
    from 3.6.14 and have correct patch filename order.
  * 44_rel3.6.14_10-Update-session_ticket.c-to-add-support-for-zero-leng.patch
    from GnuTLS 3.6.14: Handle zero length session tickets, fixing connection
    errors on TLS1.2 sessions to some big hosting providers. (See LP 1876286)
  * 44_rel3.6.14_15-_gnutls_pkcs11_verify_crt_status-check-validity-agai.patch
    44_rel3.6.14_16-x509-trigger-fallback-verification-path-when-cert-is.patch
    44_rel3.6.14_17-tests-add-test-case-for-certificate-chain-supersedin.patch
    backported from GnuTLS 3.6.14: Fix verification error with alternate
    chains. Closes: #961889

 -- Andreas Metzler <ametzler@debian.org>  Sun, 07 Jun 2020 07:45:55 +0200
```

The last bullet point talks "chains", and refers to https://bugs.debian.org/961889 which is titled, "Fails building chains with expired intermediate regardless of trust store".

So there you go. This story has a lesson: despite being extremely important, TLS libraries are generally of mediocre quality.

PR #1695, opened 2021-10-04
Make birth date non-nullable on dependents table [#179379352]

alaroia-cfa:
On production, there are 0 such dependents:

```
db=# select count(*) from dependents where birth_date is null;
 count
-------
     0
(1 row)
```

Same on demo:

```
irb(main):002:0> Dependent.where(birth_date: nil).count
=> 0
```

This would have prevented the strange situation that caused https://sentry.io/organizations/codeforamerica/issues/2690836784/?project=1880327&query=is%3Aunresolved&sort=freq&statsPeriod=14d

### DIFF
--- a/db/migrate/20211004220516_make_birth_date_non_null_on_dependents.rb
+++ b/db/migrate/20211004220516_make_birth_date_non_null_on_dependents.rb
@@ -1,0 +1,5 @@
+class MakeBirthDateNonNullOnDependents < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :dependents, :birth_date, false
+  end
+end
--- a/app/models/dependent.rb
+++ b/app/models/dependent.rb
@@ -3,7 +3,7 @@
 # Table name: dependents
 #
 #  id                                          :bigint           not null, primary key
-#  birth_date                                  :date
+#  birth_date                                  :date             not null
 #  born_in_2020                                :integer          default("unfilled"), not null
 #  cant_be_claimed_by_other                    :integer          default("unfilled"), not null
 #  claim_anyway                                :integer          default("unfilled"), not null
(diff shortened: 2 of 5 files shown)


app/models/tax_return.rb
```diff
@@ -4,8 +4,6 @@
 #
 #  id                  :bigint           not null, primary key
 #  certification_level :integer
-#  ctc_amount_cents    :bigint
```
caitbot, line 7: weird schema things happened!

spec/models/intake/gyr_intake_spec.rb
```diff
@@ -1,3 +1,255 @@
+# == Schema Information
+#
+# Table name: intakes
+#
+#  id                                                   :bigint           not null, primary key
+#  additional_info                                      :string
+#  adopted_child                                        :integer          default("unfilled"), not null
```
caitbot, line 7: this is also unrelated, but I suppose it isn't bad to have..

db/migrate/20211004220516_make_birth_date_non_null_on_dependents.rb
```diff
@@ -0,0 +1,5 @@
+class MakeBirthDateNonNullOnDependents < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :dependents, :birth_date, false
```
caitbot, line 3: huh, there's a method for every kind of column change

caitbot (approved):
the dependents-not-null related things look good. there are some schema notations that should probably be removed before merging (:

app/models/tax_return.rb
alaroia-cfa: Fixed in the rebase!

PR #1700, opened 2021-10-05
Rename "CTC Admin Tools" to "Admin Tools"

embarnard:


spec/features/ctc/admin_tools_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.feature "CTC Admin Tools", active_job: true do
```
alaroia-cfa, line 3: The `#rename` comment should probably be deleted.

alaroia-cfa (approved):
Please remove the new `#rename` comment, then merge this! :D Looks great. Thanks for keeping the admin-focused app copy accurate!

spec/features/ctc/admin_tools_spec.rb
embarnard: oops good catch!

PR #1702, opened 2021-10-05
When Twilio raises an error, mark as failed in the Hub [#178990861]

alaroia-cfa:

### DIFF
--- a/app/jobs/send_outgoing_text_message_job.rb
+++ b/app/jobs/send_outgoing_text_message_job.rb
@@ -5,11 +5,16 @@ class SendOutgoingTextMessageJob < ApplicationJob
   def perform(outgoing_text_message_id)
     outgoing_text_message = OutgoingTextMessage.find(outgoing_text_message_id)
 
-    message = TwilioService.send_text_message(
-      to: outgoing_text_message.to_phone_number,
-      body: outgoing_text_message.body,
-      status_callback: outgoing_text_message_url(outgoing_text_message, locale: nil)
-    )
+    message = begin
+                TwilioService.send_text_message(
+                  to: outgoing_text_message.to_phone_number,
+                  body: outgoing_text_message.body,
+                  status_callback: outgoing_text_message_url(outgoing_text_message, locale: nil)
+                )
+              rescue Twilio::REST::RestError
+                outgoing_text_message.update(twilio_status: "twilio_error")
+                raise
+              end
 
     outgoing_text_message.update(
       twilio_status: message.status,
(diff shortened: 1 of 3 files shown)


app/jobs/send_outgoing_text_message_job.rb
```diff
@@ -5,11 +5,16 @@ class SendOutgoingTextMessageJob < ApplicationJob
...
-    )
+    message = begin
+                TwilioService.send_text_message(
+                  to: outgoing_text_message.to_phone_number,
+                  body: outgoing_text_message.body,
+                  status_callback: outgoing_text_message_url(outgoing_text_message, locale: nil)
+                )
+              rescue Twilio::REST::RestError
+                outgoing_text_message.update(twilio_status: "twilio_error")
+                raise
```
tgrathwell, line 16: is there any capability for these to be retried?

In GCF, we handle this the following way:
* [Most jobs have a nonzero `retry_attempts`](https://github.com/codeforamerica/gcf-backend/blob/main/app/jobs/intercom_to_sms_job.rb)
* [Jobs that use Twilio or Mailgun rescue from any of a list of ephemeral-seeming errors, and reraise as 'NetworkError'](https://github.com/codeforamerica/gcf-backend/blob/main/app/jobs/application_job.rb#L37)
* [sentry.rb ignores `NetworkError`](https://github.com/codeforamerica/gcf-backend/blob/main/config/initializers/sentry.rb#L34)

not saying this is the only or best way to do it, but it's **a** way to do it, and it hopefully automatically clears up some failures cause by network tomfoolery

Though we'd have to be careful if we have a nonzero `retry_attempts` on some jobs that they're actually idempotent and whatnot; e.g. if we had this job
```
TwilioService.send_sms('hello');
DatadogService.track('sent_sms');
```

... and it crashed from an error in Datadog land, we don't want it to trigger a retry and keep sending the same SMS over and over

alaroia-cfa: Our top RestError comes from messages being too long (over 1600), so retrying would just be bad. We'd want to distinguish "we did something dumb" from "Twilio is sad right now", someday.

My goal right now is to make sure we're communicating these failures to Hub users in some kind of visible way, since this week we had about 170 of them.

I think that following your plan is basically a good idea. A good option for later would be to handle the 1600 chars issue separately. I want to roll-out this patch ASAP if you will let me, though.

tgrathwell: huh, I wonder how many times the 1600 character thing occurs in GCF? I guess it just retries a bunch and gives up without any additional fanfare which I guess is bad

yeah you can merge it I guess

alaroia-cfa: :D

PR #1703, opened 2021-10-05
Always validate that dependents have a birth_date [#179379352]

alaroia-cfa:
Upon save, you now see a useful validation error.

image

### DIFF
--- a/app/models/dependent.rb
+++ b/app/models/dependent.rb
@@ -82,8 +82,7 @@ class Dependent < ApplicationRecord
   validates_presence_of :first_name
   validates_presence_of :last_name
 
-  # Allow birth date to be blank when we first create dependents in the CTC intake flow, but nowhere else
-  validates_presence_of :birth_date, unless: -> { intake&.is_ctc? }
+  validates_presence_of :birth_date
   validates_presence_of :birth_date, on: :ctc_valet_form
 
   validates_presence_of :relationship, on: :ctc_valet_form
--- a/app/models/dependent.rb
+++ b/app/models/dependent.rb
@@ -83,7 +83,6 @@ class Dependent < ApplicationRecord
   validates_presence_of :last_name
 
   validates_presence_of :birth_date
-  validates_presence_of :birth_date, on: :ctc_valet_form
 
   validates_presence_of :relationship, on: :ctc_valet_form
 


app/models/dependent.rb
```diff
@@ -82,9 +82,7 @@ class Dependent < ApplicationRecord
   validates_presence_of :first_name
   validates_presence_of :last_name
 
-  # Allow birth date to be blank when we first create dependents in the CTC intake flow, but nowhere else
-  validates_presence_of :birth_date, unless: -> { intake&.is_ctc? }
-  validates_presence_of :birth_date, on: :ctc_valet_form
```
tgrathwell, line 87: this diff is fine, but is there a reason this conditional validation wasn't being triggered when reproducing the situation where you can somehow get a dependent without a birthdate into the system?

Because that might mean we're setting the context incorrectly sometimes which is also bad for validation

alaroia-cfa: Good question. I'll look into that question soon (tomorrow morning hopefully).

tgrathwell: looks like the `dependent_validation_context` is not being set on update because it's set in `UpdateClientForm` but now `UpdateCtcClientForm`. which probably dates back to the creation of `UpdateCtcClientForm` -- at the time, I had been trying to cram all the ctc/gyr differences into `UpdateClientForm` but someone decided to fork them apart and it has solved some problems and caused other ones

here is a quick and dirty fix but it doesn't have tests or anything https://github.com/codeforamerica/vita-min/pull/1704

if we want to implement that we may need to do something about any existing CTC dependents that don't have a 'relationship' or 'ssn' set

alaroia-cfa: > if we want to implement that we may need to do something about any existing CTC dependents that don't have a 'relationship' or 'ssn' set

Originally, when I read this comment, I thought you meant that it would be good to make relationship or SSN mandatory. I see that we have lot of such dependents in the database. I now remember that dependents are created bit by bit, data-wise. Therefore, I now think what you were saying is that if a Hub user edits a client, if the client has any dependents without a relationship set, the Hub user won't be able to save because the client will fail validation.

Do I understand you correctly? If so, it seems like a substantial problem.

What would you think about, if a dependent is unchanged, we don't run validation on it? (Is that how it is intended to work, already?)

I can move this comment to the other PR. I figured I'd reply here since we have this thread already going.

tgrathwell: Currently the CTC intake flow and the CTC valet form both require relationship and SSN on create; anyone who doesn't have it was created before the pages in CTC intake were combined, or because of this bug

We **could** have some special validation code, specific to the CTC valet update form, that still considers a dependent 'valid' if it was already persisted and the relationship/SSN has not changed. But they probably need relationship+SSN to actually file, so I dunno if that's worth doing

PR #1704, opened 2021-10-05
Hub valet form: disallow setting dependent 'relationship' or 'ssn' to null on update

tgrathwell:
the validation was only being applied on create, not update

### DIFF
--- a/app/forms/hub/update_ctc_client_form.rb
+++ b/app/forms/hub/update_ctc_client_form.rb
@@ -117,6 +117,10 @@ module Hub
       new(client, existing_attributes(intake, attribute_keys))
     end
 
+    def dependent_validation_context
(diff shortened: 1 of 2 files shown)


tgrathwell:
See additional discussion elsewhere: https://github.com/codeforamerica/vita-min/pull/1703#pullrequestreview-771843421

For what it's worth, no CTC dependents seem to have a blank 'relationship' but a number of them have blank SSN:
```
irb(main):005:0> Dependent.includes(:intake).where(intakes: { type: 'Intake::CtcIntake' }).where(relationship: nil).count
=> 0
```

```
irb(main):008:0> Dependent.includes(:intake).where(intakes: { type: 'Intake::CtcIntake' }).where(encrypted_ssn: nil).count
=> 2746
```

so if we deployed this as-is, someone editing those intakes with the valet form could not save until they provided an SSN

PR #1711, opened 2021-10-08
State machine spike

bytheway875:
Currently, we only store information about the latest status and we don’t have access to the history. For a recent task, I was asked to change the status back to something when some action happened if the previous status was something else. We’ve been asked to do something like that before, and we accomplished it using a field that stored when that transition happened. However, there’s also talk of wanting to monitor how long returns are in different statuses for SLA or monitoring purposes next season, and we do a lot of “state machine” type stuff with tax returns but we don’t really have a state machine.

This is a minimal implementation of transitioning to using statesman state machine. It will require additional refactor work to completely get rid of the old implementation of TaxReturnStatuses and to move some of the after transition logic into the state machine, but this at least moves us in the right direction with backwards compatibility for data science and for our clients. Looking for feedback.

### DIFF
--- a/app/state_machines/tax_return_state_machine.rb
+++ b/app/state_machines/tax_return_state_machine.rb
@@ -34,7 +34,7 @@ class TaxReturnStateMachine
   end
 
   after_transition(after_commit: true) do |tax_return, transition|
-    tax_return.status = transition.to_state # save the integer version, too, for now.
+    tax_return.status = transition.to_state # save the integer version, too, for now. (Backwards compatability for data science reports)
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
@bytheway875 Have you evaluated the impact of this change on performance for the client table? I realized this morning that I'm nervous about this change triggering a slowdown. I'm sure that if it does trigger a slowdown, we can find some techniques to speed things up to the old speed or faster. Using the 'rack-mini-profiler' in local dev would be a fine way to start.

PR #1727, opened 2021-10-14
i18n_spec now checks whether each key has consistent usage of HTML across locales

tgrathwell:
The intention is to fail the suite if a Transifex PR comes in where HTML from
the `en` string was not preserved in `es`, but unfortunately there's a lot of
preexisting cases where they already don't match up. So the test exists but is
'pending' for now


alaroia-cfa:
@tgrathwell this is awesome, and I've validated the most recent commit.

In the interest of latency reduction, I'm going to merge this now.

PR #1743, opened 2021-10-20
[#180018695] Fix nil multiplication error for expected payments

bytheway875:


bytheway875:
In the case that the tax return was submitted, the value was submitted and it was submitted as 0. If they're nil and they shouldn't be and they're getting through, that's def a bug that we should file and fix but I don't think it makes sense to enshrine the bug into another step by not recording the values since we did indeed file the return.

at one point, these submissions would fail -- we can look at the submission logic to see why these values are able to be nil and submit to the IRS.

tgrathwell:
ok https://www.pivotaltracker.com/story/show/180019733

PR #1758, opened 2021-10-22
Add focus to toggle element on client-header

embarnard:


embarnard:
This [story](https://www.pivotaltracker.com/story/show/178471695) requests that we add a focus to the toggle button in the client profile header for accessibility reasons. The current PR does accomplish adding a focus when the toggle is clicked but when we test it using a screen reader and hit enter it does not change the toggle. Therefore there needs to be more work to make meet our standard of accessibility before merging. Closing for now and hoping to resurface later.

PR #1766, opened 2021-10-29
Add job to delete unconsented GYR intakes that were created more than 2 weeks ago

embarnard:


caitbot (approved):
This looks good to me. 

If you're at all worried about performance/testing in prod, you could consider adding a `limit` param to the job and running it on a small subset before running the whole thing. The only reason I suggest that is it looks like there are a WHOLE bunch of records that are `dependent: :destroy` on Client, which might make a `destroy` on Client slow. Or maybe just run this at a non-peak time of day, in case it causes high db load. 

I might be being overly cautious due to recent issues 😂

bytheway875:
why'd we choose 2 weeks? I feel like we could get away with a much shorter time period -- 2 days maybe

embarnard:
> This looks good to me.
> 
> If you're at all worried about performance/testing in prod, you could consider adding a `limit` param to the job and running it on a small subset before running the whole thing. The only reason I suggest that is it looks like there are a WHOLE bunch of records that are `dependent: :destroy` on Client, which might make a `destroy` on Client slow. Or maybe just run this at a non-peak time of day, in case it causes high db load.
> 
> I might be being overly cautious due to recent issues 😂

Oh good point! Do you think lowering the batch size could also address the high db load? I did this intentionally because one of the requirements was to delete all associated records. I also did a check of what kinds of records we created between the backtaxes and consent page and it should only be `Client`, `Intake` and `TaxReturn` records.

embarnard:
> why'd we choose 2 weeks? I feel like we could get away with a much shorter time period -- 2 days maybe
@bytheway875 
It was a frame of time that was confirmed by product and chosen arbitrarily because it felt like a safe amount of time to assume the client would not return -- but happy to change it to a lower frame of time. I can check in with product again.

bytheway875:
ah... since the session ends when you close or navigate away from the browser, it seems like we'd be safe with a much shorter time to delete. If someone comes back to a dead tab 2 weeks later, I'd almost expect the form to fail and to have to start over. but if that's what was requested, its also fine

PR #1768, opened 2021-10-29
Flag phone numbers for reuse

bytheway875:
Branches off of flagging bank account for reuse.

### DIFF
--- a/spec/services/fraud_indicator_service_spec.rb
+++ b/spec/services/fraud_indicator_service_spec.rb
@@ -177,7 +177,7 @@ describe FraudIndicatorService do
           create :ctc_intake, phone_number: "+15124441234"
         end
 
-        it "marks for fraud" do
+        it "doesnt mark for fraud" do
           expect(FraudIndicatorService.new(intake.client).hold_indicators).to eq []
         end
       end
@@ -191,7 +191,7 @@ describe FraudIndicatorService do
           end
         end
 
-        it "marks for fraud" do
+        it "doesnt mark for fraud" do
           expect(FraudIndicatorService.new(intake.client).hold_indicators).to eq []
         end
       end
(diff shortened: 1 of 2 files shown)


spec/services/fraud_indicator_service_spec.rb
```diff
@@ -155,5 +155,46 @@
...
+      end
+
+      context "when there are fewer than 3 CTC intakes with the same phone number" do
+        let(:intake) { create :ctc_intake, phone_number: "+15124441234" }
+
+        before do
+          create :ctc_intake, phone_number: "+15124441234"
+        end
+
+        it "marks for fraud" do
```
jenny-heath, line 180: should say "doesn't mark for fraud"

spec/services/fraud_indicator_service_spec.rb
```diff
@@ -155,5 +155,46 @@
...
+      context "when there are GYR intakes with the same phone number" do
+        let(:intake) { create :ctc_intake, phone_number: "+18324658840" }
+
+        before do
+          3.times do
+            create :intake, phone_number: "+18324758840", type: "Intake::GyrIntake"
+          end
+        end
+
+        it "marks for fraud" do
```
jenny-heath, line 194: ditto

jenny-heath (approved):
looks good! nit: a couple of spec descriptions that say the opposite of what they mean. also i'm not familiar with the DeduplificationService but assuming that was reviewed earlier today (the [PR](https://github.com/codeforamerica/vita-min/pull/1763) doesn't indicate a pair or review but these things don't always get documented)

bytheway875:
@jenny-heath good catch! fixed -- yes, had been working on that with asheesh. we tuple reviewed and revised.

PR #1770, opened 2021-11-01
Only send fraud OR prep message, not both

bytheway875:
Because the new -> preparing logic was in a separate callback, it was getting triggered even when we were passing through preparing into fraud hold.

I'm able to remove the check for submission count because the email sending logic checks to see if the email has already been sent before sending.


alaroia-cfa (approved):
Nice! Cool that we can rely on `MessageTracker` et al to simplify the code here.

PR #1772, opened 2021-11-01
Move admin layout into its own layout

bytheway875:
To prepare for some upcoming changes to the layout of the hub, we need to extract the admin layout to its own thing. This leaves the same structure in place but inlines certain things that were being configured at the existing admin template layer.


alaroia-cfa (approved):
🚢

PR #1774, opened 2021-11-01
Remove stray punctuation in HTML title on the Hub

alaroia-cfa:
Before -- notice the window title has an extra `%>` in it

image

After -- it's cleaned-up

image


alaroia-cfa:
Whoever reviews this -- if you like it, please merge!

PR #1778, opened 2021-11-02
Ensure no continued intake 180019127

bytheway875:
Screen Shot 2021-11-02 at 5 45 04 PM

This PR also prevents using "go back" and browser back to make changes to a completed submission.

### DIFF
--- a/app/controllers/concerns/ctc/after_verification_concern.rb
+++ b/app/controllers/concerns/ctc/after_verification_concern.rb
@@ -7,6 +7,7 @@ module Ctc
     def after_verification_actions
       send_mixpanel_event(event_name: "ctc_contact_verified")
       sign_in current_intake.client
+      session.delete("intake_id")
       current_intake.tax_returns.last.advance_to("intake_in_progress")
 
       ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
--- a/app/controllers/ctc/ctc_pages_controller.rb
+++ b/app/controllers/ctc/ctc_pages_controller.rb
@@ -79,5 +79,7 @@ module Ctc
 
     def how_do_i_get_an_itin
     end
+
+    def completed; end
   end
 end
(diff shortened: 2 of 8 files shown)


tgrathwell:
`redirect_if_completed_intake_present` is sorely needed and will be a great boon to the people of the earth

not sure what the part about deleting intake from the session is... did it just seem like something that ought to be cleaned up, or does it solve a particular problem?

app/controllers/concerns/ctc/after_verification_concern.rb
```diff
@@ -7,6 +7,7 @@ module AfterVerificationConcern
     def after_verification_actions
       send_mixpanel_event(event_name: "ctc_contact_verified")
       sign_in current_intake.client
+      session.delete("intake_id")
```
alaroia-cfa, line 10: I like this, for clarity.

app/controllers/clients/sessions_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Clients::SessionsController < Devise::SessionsController
+  def destroy
+    session[:intake_id] = nil
```
alaroia-cfa, line 3: I'd like to know, on logout -- do we clear the entire session? If so, this isn't needed. Maybe we can Tuple and watch the network console? Harmless to keep it.

alaroia-cfa (approved):
One question; no blockers; yay.

app/controllers/clients/sessions_controller.rb
bytheway875: yeah, i actually think maybe this is not needed when we remove the intake_id when someone is logged in. I can also remove this code and leave the test and see if it clears.

bytheway875:
@alaroia-cfa I tested to ensure that the entire session cookie gets deleted which it does (which def makes sense), so the custom destroy action was redundant like you said. I removed it, but I left the deletion of intake_id from the session when someone logs in because i think it makes it more straightforward that you need to look at the current_client's intake in cases where current_client is present.

PR #1779, opened 2021-11-03
Add irs_submission_id to create analytics views

bytheway875:
Allows people with access to replica to run reports with submission ids we can send to the irs


bytheway875 (commented):
reviewed w catie

PR #1780, opened 2021-11-03
cron task for send some stats to datadog every minute

tgrathwell:
* current queue length (Delayed::Job.count excluding failed or future-run jobs)
* current state of each efile submission ('failed', 'transmitted', 'accepted' etc)


app/models/efile_submission.rb
```diff
@@ -46,6 +46,21 @@ def state_machine
   delegate :can_transition_to?, :current_state, :history, :last_transition, :last_transition_to,
            :transition_to!, :transition_to, :in_state?, to: :state_machine
 
+  def self.state_counts(except: [])
```
bytheway875, line 49: I think we should add new to the except list by default.

 If we don't skip new, it will always be included and it will always be 0 because as the "initial state", it never gets stored as a persisted transition to_state. I think it's somewhat misleading to report that there are 0 in the new state, because the number in the "new" state would actually accurately be the number EfileSubmissions that have 0 associated transition objects.

bytheway875 (approved):

PR #1781, opened 2021-11-03
delayed_jobs table stores job_class and job_object_id

tgrathwell:
This is accomplished by adding a hook to Delayed::Worker.plugins
which copies properties from the serialized job handler onto the
Delayed::Job record.

The job_class part in particular makes it easier to determine how
many of a given job are trying to run. Otherwise you have to pull
all the jobs into memory and check `job.payload_object.handler['job_class']`

The job_object_id part is occasionally useful if you wanted to see
how many SuchAndSuchJobs are queued up for a particular intake or
submission or whatever other ActiveRecord model the might be most
aligned with.

The active_job_id part is just because we did it that way in GCF
and may occasionally be useful when digging through logs? Who could say.

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,2 +1,21 @@
 Delayed::Worker.destroy_failed_jobs = false
 Delayed::Worker.max_attempts = 1
+
+module Delayed
+  module Plugins
+    class AddExtraJobAttributes < Plugin
+      callbacks do |lifecycle|
+        lifecycle.before(:enqueue) do |delayed_job|
+          serialized_job = delayed_job.payload_object.job_data
+          delayed_job.assign_attributes(
+            job_class: serialized_job['job_class'],
+            job_object_id: serialized_job['job_object_id'],
+            active_job_id: serialized_job['job_id']
+          )
+        end
+      end
+    end
+  end
+end
+
+Delayed::Worker.plugins << Delayed::Plugins::AddExtraJobAttributes
--- a/spec/jobs/after_transition_tasks_for_rejected_return_job_spec.rb
+++ b/spec/jobs/after_transition_tasks_for_rejected_return_job_spec.rb
@@ -62,4 +62,18 @@ describe AfterTransitionTasksForRejectedReturnJob do
       end
     end
   end
+
+  describe '#job_object_id' do
+    let(:submission) { create(:efile_submission, :transmitted) }
+    let(:efile_error) { create(:efile_error, code: "IRS-ERROR", expose: true, auto_wait: false, auto_cancel: false) }
+
+    before do
+      submission.transition_to!(:rejected, error_code: efile_error.code)
+    end
+
+    it 'returns the id of the EfileSubmission record' do
+      job = described_class.perform_later(submission, submission.last_transition)
+      expect(job.job_object_id).to eq(submission.id)
+    end
+  end
 end
--- a/spec/jobs/application_job_spec.rb
+++ b/spec/jobs/application_job_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+
+describe ApplicationJob do
+  before do
+    SampleJobClass = Class.new(ApplicationJob) do
+      def perform(object_id)
+        return true
+      end
+    end
+  end
+
+  around do |example|
+    ActiveJob::Base.queue_adapter = :delayed_job
+    example.run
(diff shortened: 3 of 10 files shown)


config/initializers/delayed_job.rb
```diff
@@ -1,2 +1,21 @@
 Delayed::Worker.destroy_failed_jobs = false
 Delayed::Worker.max_attempts = 1
+
+module Delayed
```
alaroia-cfa, line 4: Trivial -- why put this in `Delayed::Plugins` when we can instead make our own class in like `lib` or `services` called `ExtraJobAttributes`, then do `Delayed::Worker.plugins << ExtraJobAttributes`? I'd prefer the class definition to be outside the initializer.

Not a blocker.

spec/jobs/after_transition_tasks_for_rejected_return_job_spec.rb
```diff
@@ -62,4 +62,18 @@
...
+
+  describe '#job_object_id' do
+    let(:submission) { create(:efile_submission, :transmitted) }
+    let(:efile_error) { create(:efile_error, code: "IRS-ERROR", expose: true, auto_wait: false, auto_cancel: false) }
+
+    before do
+      submission.transition_to!(:rejected, error_code: efile_error.code)
+    end
+
+    it 'returns the id of the OutgoingEmail record' do
```
alaroia-cfa, line 74: Confused -- this talks about an `OutgoingEmail record` but really it's looking for an `EfileSubmission` record, right?

spec/jobs/application_job_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+describe ApplicationJob do
```
alaroia-cfa, line 3: Yay, fine spec

spec/jobs/send_outgoing_email_job_spec.rb
```diff
@@ -19,4 +19,13 @@
       expect(outgoing_email.sent_at).to eq fake_time
     end
   end
+
+  describe '#job_object_id' do
```
alaroia-cfa, line 23: Why bother with this spec here? I'd rather see the two cases covered in `ApplicationJobSpec`.

Not a blocker.

spec/support/active_job.rb
```diff
@@ -4,7 +4,7 @@
   config.include ActiveJob::TestHelper, type: :feature
 
   config.before(:each) do |example|
-    if [:job, :feature].include? example.metadata[:type] || example.metadata[:active_job]
+    if (ActiveJob::Base.queue_adapter.respond_to?(:perform_enqueued_jobs)) && ([:job, :feature].include?(example.metadata[:type]) || example.metadata[:active_job])
```
alaroia-cfa, line 7: I don't know what this is really about, and would like to know.

alaroia-cfa (approved):
This is fine as-is. Some suggestions and comments, but honestly none are strictly blockers.

spec/jobs/after_transition_tasks_for_rejected_return_job_spec.rb
tgrathwell: that's a good ol' copy paste for ya dawg

spec/jobs/send_outgoing_email_job_spec.rb
tgrathwell: two reasons
* I made this before ApplicationJobSpec because I was lazy
* there is perhaps some value in testing two real implementations (one where the first arg is an integer, one where it is an object) rather than setting up fake classes that behave like that, in case the fake classes are too fake

alaroia-cfa: I still have my preference, but I see your point, and can totally live with this testing approach.

spec/support/active_job.rb
tgrathwell: for the tests where we flop over to `queue_adapter = :delayed_job` to test the delayed job activerecord stuff, `clear_enqueued_jobs` and `clear_performed_jobs` don't exist

I guess it could be alternately be written as
```
if whatever
  clear_enqueued_jobs if respond_to?(:clear_enqueued_jobs)
end
```

(does that work? just typing off the cuff here)

but that seemed a little weirder because if `clear_enqueued_jobs` somehow stopped existing because it got renamed that vestigal line would be there forever

PR #1784, opened 2021-11-04
When validating phone number, don't crash if it's non-UTF-8 garbage

alaroia-cfa:


alaroia-cfa:
I think we are getting binary garbage because scumbag robots are submitting non-UTF-8 data.

PR #1788, opened 2021-11-05
Fix bash syntax for default port

bensheldon:


alaroia-cfa:
Boldly merging this non-client-facing config change solely for Heroku.

PR #1794, opened 2021-11-09
Fix styling on text message character limiter [#180196444]

alaroia-cfa:
When empty

image

When big but not too big

image

When too big

image

When fixed by hitting backspace

image


caitbot (approved):
reviewed as pair

PR #1795, opened 2021-11-09
specs: Add a MockDogapi concern to consistently replace Dogapi::Client with a mock

tgrathwell:
previously, if a test did this, it was required to unset the memoized
@dogapi_client in the DatadogApi class, otherwise later tests could
complain that they were using stale mocks from another test.

now it will be more automatically cleaned up


alaroia-cfa (approved):
OK! Why not!

Looks good to me.

PR #1803, opened 2021-11-10
VitaPartner refactor: introducing Site and Organization

caitbot:
### Ways to approach
The first five commits are important and so I would recommend viewing by commit. Beyond that, it is spec changes that are mostly trivial (updating `creates` to be for orgs or sites, not vita partners). The important test changes are those related to the `organizations_controller` and the `sites_controller`.

### Changes to pay attention to
- `VitaPartner` now has two sub classes, `Site` and `Organization`. Wow.
- This means there is now a `type` column on VitaPartner
- There was some weirdness about `oring` because of Site and Organization having different shapes, so now it is `+`. My googling skills failed me in telling me what the big differences are between the two. My main concern is performance.
- To limit the breadth of changes, I mocked our old scopes `VitaPartner.organizations` and `VitaPartner.sites`. I will need to follow up to replace those with `Organization/Site.all`. 
- The site controller and organization controller now refer to their particular subclasses rather than a `VitaPartner`
- Minor updates to `ability` and `accessible_vita_partners` to account for the two new classes.

### Deployment notes
Please, merge and release the migrations (#1804) first. THEN rebase this branch on main and merge it, test it, release it. 

I recommend heavy poking in demo (on org/site related features) prior to releasing, we have specs, but manual testing is nice too.

PR #1804, opened 2021-11-10
VitaPartner refactor: just the migrations

caitbot:
The whole thing lives over here #1803

I pulled out the migrations, we should release these first (before and separate from the class changes) to prevent downtime.

### DIFF
--- a/db/migrate/20211109200705_backfill_type_on_vita_partner.rb
+++ b/db/migrate/20211109200705_backfill_type_on_vita_partner.rb
@@ -1,12 +1,13 @@
 class BackfillTypeOnVitaPartner < ActiveRecord::Migration[6.0]
-  disable_ddl_transaction!
 
   def up
-    orgs = VitaPartner.where(parent_organization_id: nil)
-    orgs.update_all(type: "Organization")
+    ActiveRecord::Base.connection.execute(
+      "UPDATE vita_partners SET type='Organization' WHERE parent_organization_id IS NULL"
+    )
 
-    sites = VitaPartner.where.not(parent_organization_id: nil)
-    sites.update_all(type: "Site")
+    ActiveRecord::Base.connection.execute(
+      "UPDATE vita_partners SET type='Site' WHERE parent_organization_id IS NOT NULL"
+    )
   end
 
   def down
--- a/db/migrate/20211109200705_backfill_type_on_vita_partner.rb
+++ b/db/migrate/20211109200705_backfill_type_on_vita_partner.rb
@@ -1,0 +1,15 @@
+class BackfillTypeOnVitaPartner < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def up
+    orgs = VitaPartner.where(parent_organization_id: nil)
+    orgs.update_all(type: "Organization")
+
+    sites = VitaPartner.where.not(parent_organization_id: nil)
+    sites.update_all(type: "Site")
+  end
+
+  def down
+    VitaPartner.all.update_all(type: nil)
+  end
+end
--- a/app/models/vita_partner.rb
+++ b/app/models/vita_partner.rb
@@ -11,6 +11,7 @@
 #  national_overflow_location :boolean          default(FALSE)
 #  processes_ctc              :boolean          default(FALSE)
 #  timezone                   :string           default("America/New_York")
+#  type                       :string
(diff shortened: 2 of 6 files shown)


db/migrate/20211109200705_backfill_type_on_vita_partner.rb
```diff
@@ -0,0 +1,15 @@
+class BackfillTypeOnVitaPartner < ActiveRecord::Migration[6.0]
```
caitbot, line 1: Looking at this now, my one thought is that there will be a period of time between this going out and the class changes going out, which made me concerned about there being a partner that is created in the in-between, that then lacks a `type`.

But in typing this out, I remember that we don't actually create new orgs or sites often (or ever? except the start of seasons?) and so this really shouldn't be a problem.

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
```diff
@@ -0,0 +1,15 @@
...
+  disable_ddl_transaction!
+
+  def up
+    orgs = VitaPartner.where(parent_organization_id: nil)
```
alaroia-cfa, line 5: It can be bad form to use a model in a migration: https://jakeyesbeck.com/2021/04/10/avoid-models-in-migrations/

I suppose it doesn't matter in this case necessarily, so it's not a blocker. If you are willing to change it, that's excellent -- what do you think?

I encourage you to use raw SQL to 
`db/migrate/20201222174316_move_user_id_from_role_to_user.rb` is an example.

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
```diff
@@ -0,0 +1,15 @@
+class BackfillTypeOnVitaPartner < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
```
alaroia-cfa, line 2: I encourage you to keep transactions enabled. `disable_ddl_transaction!` is required for `algorithm: :concurrently`, but you're not doing that, so let the transactions flow.

Not a blocker... but would you be willing to change it?

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
alaroia-cfa: I guess I agree! If you're the one to do the deploy of the PR after this, you could visit `rails c` in prod and check for `VitaPartner.where(type: nil).count` before rolling that out.

alaroia-cfa (changes requested):
Personally, I would like to see:

* Another migration after this that makes `type` `NOT NULL` to avoid any confusion down the road. It should be in this pull request, in my opinion.
* A change to the `up` to use raw SQL (I'm open to push-back on this).
* The removal of `disable_ddl_transaction!`
* The knowledge that you've manually tested all that on your workstation.

I'm actually open to push-back on basically any of the above if you think it's important.

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
caitbot: My understanding was that this just stops the outer transaction from being created (wrapping the migration) but in the case of backfilling data, we don't need to erase our progress if something goes wrong saving one of the models.

I was following this https://github.com/ankane/strong_migrations#backfilling-data

What do you think?

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
caitbot: I'll do that!

db/migrate/20211109200705_backfill_type_on_vita_partner.rb
```diff
@@ -0,0 +1,16 @@
...
+      "UPDATE vita_partners SET type='Organization' WHERE parent_organization_id IS NULL"
+    )
+
+    ActiveRecord::Base.connection.execute(
+      "UPDATE vita_partners SET type='Site' WHERE parent_organization_id IS NOT NULL"
+    )
+  end
+
+  def down
+    VitaPartner.all.update_all(type: nil)
```
alaroia-cfa, line 14: Ho ho ho, this could be raw SQL and that'd be better, except we don't use `down` in prod anyway, so it's no problem to leave it this way.

PR #1806, opened 2021-11-12
Intake persists and indexes canoncial_email_address and email_domain

tgrathwell:
For gmail addresses, canonical_email_address removes any dots before
the domain. e.g., john.q.public@gmail.com -> johnqpublic@gmail.com

Will be backfilled some time after this lands

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -12,6 +12,7 @@
 #  bought_energy_efficient_items                        :integer
 #  bought_health_insurance                              :integer          default(0), not null
 #  cannot_claim_me_as_a_dependent                       :integer          default(0), not null
+#  canonical_email_address                              :string
 #  city                                                 :string
 #  claim_owed_stimulus_money                            :integer          default("unfilled"), not null
 #  claimed_by_another                                   :integer          default(0), not null
@@ -49,6 +50,7 @@
 #  eip_only                                             :boolean
 #  email_address                                        :citext
 #  email_address_verified_at                            :datetime
+#  email_domain                                         :string
 #  email_notification_opt_in                            :integer          default("unfilled"), not null
 #  encrypted_bank_account_number                        :string
 #  encrypted_bank_account_number_iv                     :string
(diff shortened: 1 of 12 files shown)


app/models/intake.rb
```diff
@@ -276,6 +280,10 @@ def self.searchable_fields
 
   before_save do
     self.needs_to_flush_searchable_data_set_at = Time.current
+    if email_address.present?
```
alaroia-cfa, line 283: Will this cause unnecessary writes? i.e., when saving irrelevant columns, will Postgres try to update the `self.email_domain` column?

app/models/intake.rb
```diff
@@ -568,4 +576,13 @@ def new_dependent_token
     verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
     verifier.generate(SecureRandom.base36(24))
   end
+
+  def compute_canonical_email_address
+    if email_domain == 'gmail.com'
+      username, domain = email_address.split('@')
```
alaroia-cfa, line 582: In the unlikely event there's two `@` characters, do you want to handle that? I guess not because that's handled by our email validator. OK! Ignore this comment.

app/models/intake.rb
```diff
@@ -568,4 +576,13 @@ def new_dependent_token
     verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
     verifier.generate(SecureRandom.base36(24))
   end
+
+  def compute_canonical_email_address
+    if email_domain == 'gmail.com'
```
alaroia-cfa, line 581: In theory, gmail.com is definitely not the only domain name with this treatment. If we were to try to do an excellent job at this, we'd do a DNS lookup on the email domain and check that it's operated by Google. That approach would probably mean this should be in a background job, and would probably mean that we'd want it in another table called like `IntakeCanonicalEmail` which `belongs_to :intake`, so we could fill it asynchronously and separately.

Related, we may want to canonicalize the gmail-equivalent domains -- @googlemail.com is treated the same as @gmail.com. See e.g. https://www.gmass.co/blog/domains-gmail-com-googlemail-com-and-google-com/

alaroia-cfa (approved):
This approach is nice and simple, and I approve.

I left some comments as ideas for other directions if we need more complexity later.

PR #1810, opened 2021-11-13
Replace dependent DOB dropdown on CTC valet form with text inputs

jenny-heath:
- also add date formatting help text to all 3 DOB fields on that page

### DIFF
--- a/spec/features/hub/edit_client_spec.rb
+++ b/spec/features/hub/edit_client_spec.rb
@@ -116,27 +116,27 @@ RSpec.describe "a user editing a clients intake fields" do
         fill_in "Legal first name", with: "Laura"
         expect(find_field("hub_update_client_form[dependents_attributes][0][first_name]").value).to eq "Laura"
         fill_in "Legal last name", with: "Peaches"
-        select "December", from: "Month"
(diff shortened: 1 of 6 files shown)


spec/features/hub/edit_client_spec.rb
```diff
@@ -159,8 +159,8 @@
 
       within "#dependents-fields" do
         new_field_id = all(".dependent-form").last.first("input")["id"].tr('^0-9', '')
-        fill_in "hub_update_client_form_dependents_attributes_#{new_field_id}_last_name", with: "Chung"
-        select "2019", from: "hub_update_client_form_dependents_attributes_#{new_field_id}_birth_date_year"
+        fill_in "hub_update_client_form_dependents_attributes_#{new_field_id}_last_name", with: "Cranberry"
+        # fill_in "hub_update_client_form_dependents_attributes_#{new_field_id}_birth_date_year", with: "2019"
```
jenny-heath, line 163: Why commented-out?

jenny-heath: If the point is to test editing, let's edit this -- let's make it be 2018 and adjust the spec below.

PR #1813, opened 2021-11-15
rake stats:track_metrics reports queue length by job_class

tgrathwell:
now that we are storing job_class on Delayed::Job this is possible

### DIFF
--- a/lib/tasks/stats.rake
+++ b/lib/tasks/stats.rake
@@ -1,7 +1,13 @@
 namespace :stats do
   desc "Send custom application metrics to Datadog"
   task track_metrics: :environment do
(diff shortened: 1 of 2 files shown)


bensheldon:
What do you think of also capturing point-in-time queuing latency/wait?

```ruby
select_sql = Delayed::Job.sanitize_sql(["job_class, MAX(? - run_at) AS latency", Time.now])
Delayed::Job.select(select_sql).where(failed_at: nil).where('run_at <= ?', Time.now).group(:job_class).as_json.each do |job_class_data|
  latency = job_class_data['latency']
  job_class = job_class_data['job_class']
  DatadogApi.gauge('delayed_job.job_latency', latency, tags: ["job_class:#{job_class}"])
end
```

PR #1817, opened 2021-11-16
Avoid inconsistent local test failures

alaroia-cfa:
On my workstation, I was inconsistently getting test failures like:

```
RuntimeError: failed to get urandom
```

https://stackoverflow.com/questions/59890432/rack-error-runtimeerror-failed-to-get-urandom-in-a-rails-app-rails-5-0-6-ru is a Q&A where someone else had the same problem. This pull request follows their strategy -- change the maximum permissible number of open files so that opening the `/dev/urandom` file does not fail.


tgrathwell:
okay sure!! weird!!

PR #1819, opened 2021-11-16
Rename VitaPartnerState to StateRoutingTarget

embarnard:


tgrathwell:
instead of this route we've chosen to add the new tables `StateRoutingTarget` and `VitaPartnerStateRoutingFraction` incrementally in the stories where the Coalition/Organization/Routing interface is changed

we can then choose to copy over data from the (then unused) VitaPartnerState tables onto the new tables, or just start fresh

PR #1824, opened 2021-11-17
Close CTC login on Friday, November 19, at midnight PST

caitbot:
Tested locally with the time set to November 15th and it worked as expected.


app/controllers/application_controller.rb
```diff
@@ -233,7 +234,11 @@ def open_for_intake?
   end
 
   def open_for_ctc_intake?
-    app_time <= CTC_CLOSING_TIME
+    app_time <= CTC_INTAKE_CLOSING_TIME
+  end
+
+  def open_for_ctc_login?
```
caitbot, line 240: The other option is to have an environment variable that we update. That is what we do for GYR.

The downside to the variable is that someone needs to go in and update it to shut things down. This one doesn't need that, but will require a code change to bring the site back up.

People seemed to really like the initial "low touch" requirement of this approach, so I went with it. If someone feels strongly, I would love to hear more.

PR #1825, opened 2021-11-18
rake stats:track_metrics reports queue latency by job_class

bensheldon:
Adds an additional gauge that is queue latency: for unfinished jobs, the number of seconds that have elapsed since it was expected to be run. Because this rake task is run every minute, this metric will be unreliable for values less than 60 seconds.

Why? There is a stat that tracks queue length, but that doesn't necessarily provide insight into quality of service: e.g. having 100 outbound SMS messages enqueued might not be a problem because they will each be finished very quickly; but having 1 outbound SMS message that is enqueued behind a very, very, slow job would be a problem because it would have a long wait to be sent. 

Measuring this with a query is a different strategy than GCF, which enqueues a Heartbeat Job, and then measures its latency. A benefit on vita-min is that it measures by job class, not by queue (though measurement by queue could also be done with this same strategy)

Followup to #1813


tgrathwell:
sure let's try it!!

PR #1829, opened 2021-11-19
Clean up DNS when pull requests are closed [#180153801]

alaroia-cfa:
This (untested) code should clean up Route 53 DNS when pull requests are closed in Heroku. The reason it's untested is that our Heroku app seems to have gone missing.

### DIFF
--- a/lib/tasks/heroku.rake
+++ b/lib/tasks/heroku.rake
@@ -70,6 +70,45 @@ namespace :heroku do
     Rails.logger.info("Done setting up Heroku review app DNS")
   end
 
+  task review_app_predestroy: :environment do
+    # Delete this app's hostnames from Route 53
+    gyr_hostname = Rails.application.config.gyr_url.sub(/^https:\/\//, '')
+    ctc_hostname = Rails.application.config.ctc_url.sub(/^https:\/\//, '')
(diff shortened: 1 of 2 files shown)


lib/tasks/heroku.rake
```diff
@@ -70,6 +70,45 @@ namespace :heroku do
...
+    route53_client = Aws::Route53::Client.new(
+      access_key_id: ENV["HEROKU_DNS_AWS_ACCESS_KEY_ID"],
+      secret_access_key: ENV["HEROKU_DNS_SECRET_ACCESS_KEY"],
+      region: 'us-east-1',
+    )
+    [gyr_hostname, ctc_hostname].each do |hostname|
+      cname_target = heroku_client.domain.info(heroku_app_name, hostname)["cname"]
+      Rails.logger.info("Deleting AWS CNAME with cname_target=#{cname_target} fully_qualified_domain=#{hostname}")
+
+      route53_client.change_resource_record_sets(
```
alaroia-cfa, line 90: Per https://docs.aws.amazon.com/cli/latest/reference/route53/change-resource-record-sets.html :

> To delete a resource record set, you must specify all the same values that you specified when you created it.

Seems like a lot of work to delete something, but OK.

PR #1871, opened 2021-11-30
Fix crash on getctc spanish privacy policy page

tgrathwell:
Fix escaped HTML in getctc privacy policy page


caitbot (approved):
wow so much fixing.

this is cool, its a little sneaky that it includes the Heroku fix as well, but I am ok with both as they are small.

PR #1874, opened 2021-11-30
Use full SSNs on GYR

bytheway875:
Switch to TIN Type + Full SSN on spouse and primary legal consent pages in GYR.

Also, now we will update the last 4 any time the intake is persisted with the full ssn updated.

### DIFF
--- a/app/forms/consent_form.rb
+++ b/app/forms/consent_form.rb
@@ -10,13 +10,30 @@ class ConsentForm < QuestionsForm
     :birth_date_day,
     :primary_first_name,
     :primary_last_name,
-    :primary_last_four_ssn
+    :primary_ssn,
+    :primary_tin_type
   )
+  set_attributes_for :confirmation, :primary_ssn_confirmation
+  set_attributes_for :virtual, :ssn_no_employment
 
   validates_presence_of :primary_first_name
   validates_presence_of :primary_last_name
-  validates_length_of :primary_last_four_ssn, maximum: 4, minimum: 4
   validate :valid_birth_date
+  validates :primary_tin_type, presence: true
+  validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
+  validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin" }
+  validates_presence_of :primary_ssn
+
+  with_options if: -> { (primary_ssn.present? && primary_ssn != intake.primary_ssn) || primary_ssn_confirmation.present? } do
+    validates :primary_ssn, confirmation: true
+    validates :primary_ssn_confirmation, presence: true
+  end
+
+  before_validation do
+    if ssn_no_employment == "yes" && primary_tin_type == "ssn"
+      self.primary_tin_type = "ssn_no_employment"
+    end
+  end
 
   def save
     attributes = attributes_for(:intake)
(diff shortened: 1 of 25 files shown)


app/forms/consent_form.rb
```diff
@@ -10,13 +10,24 @@ class ConsentForm < QuestionsForm
     :birth_date_day,
     :primary_first_name,
     :primary_last_name,
-    :primary_last_four_ssn
+    :primary_ssn,
+    :primary_tin_type
   )
+  set_attributes_for :confirmation, :primary_ssn_confirmation
+  set_attributes_for :virtual, :ssn_no_employment
```
embarnard, line 17: What is this used for? Do we need to save it as a tin_type value before validating like in `app/forms/ctc/dependents/info_form.rb:37`

embarnard: And if we need to include it we should add the checkbox into the view and make the spouse_consent form + view consistent

app/views/questions/consent/edit.html.erb
```diff
@@ -9,14 +9,9 @@
...
-      <%= f.cfa_input_field(
-              :primary_last_four_ssn,
-              t("views.questions.consent.primary_last_four_ssn"),
-              prefix: "XXX-XX-",
-              type: :tel,
-              classes: ["form-width--name field--last-four-ssn"],
-              options: { maxlength: 4 }
-          ) %>
+    <%= f.cfa_select(:primary_tin_type, t('general.tin_type'), tin_options_for_select(include_itin: true)) %>
+    <%= f.cfa_input_field(:primary_ssn, t("attributes.primary_ssn"), classes: ["form-width--long"], options: { maxlength: 11 }) %>
```
embarnard, line 13: is the maxlength 11 to include the 2 dashes?

embarnard (commented):
Looks good! 🤠 Just had a few questions

app/models/intake.rb
```diff
@@ -284,6 +284,8 @@ def self.searchable_fields
       self.email_domain = email_address.split('@').last.downcase
       self.canonical_email_address = compute_canonical_email_address
     end
+    self.primary_last_four_ssn = primary_ssn&.last(4) if encrypted_primary_ssn_changed?
```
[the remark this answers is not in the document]
bytheway875: hmm... i think because of the fact that we don't expect it to be an action anyone but an engineer or console access would be able to do AND we never worried about keeping SSNs in sync before, it seems like adding complexity for a use case that hasn't ever been a problem so I'd probably hold off on it. I can't see a legitimate reason for anyone doing this in good faith.

spec/models/intake_spec.rb
```diff
@@ -317,6 +317,74 @@
     end
   end
 
+  describe "keeping last 4 ssn in sync with ssn" do
+    context "when creating the object" do
+      let(:intake) { create :intake, primary_ssn: "12345678", spouse_ssn: "2345677777" }
```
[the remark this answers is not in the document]
bytheway875: its not intentional -- its actually a good catch that we CAN persist bad SSNs to an intake object. We probably could validate SSNs and or ITINs at the model level

app/views/questions/consent/edit.html.erb
bytheway875: yep!

spec/models/intake_spec.rb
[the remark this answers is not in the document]
bytheway875: changed!

PR #1877, opened 2021-12-01
Update invitations page selection + user edit tweaks

bytheway875:
Based on pairing with design:
- Updates role names to be in a consistent order in each implementation of roles as a list (User.roles)
- Title cases roles for display purposes
- Uses dropdown for invitation index select instead of buttons for each role
- Removes resend invitation button if the invitation has already been accepted
- Defaults select in hub to be 100% width of parent container (like text-input), which can then be limited by the form width classes.

### DIFF
--- a/app/views/hub/users/edit.html.erb
+++ b/app/views/hub/users/edit.html.erb
@@ -1,7 +1,7 @@
 <% content_for :page_title, @user.name %>
 <% content_for :card do %>
   <div class ="slab slab--padded">
-    <div class="client-container">
+    <div>
       <h1><%= @user.name %></h1>
 
       <% if current_user.admin? %>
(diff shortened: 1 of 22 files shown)


app/views/hub/users/edit.html.erb
```diff
@@ -16,15 +16,22 @@
         <% end %>
       <% end %>
 
-      <%= form_with model: [:hub, @user], method: :put, local: true, builder: VitaMinFormBuilder, html: { class: 'form-card form-card--long' } do |f| %>
```
bytheway875, line 19: form-card--long changes the size of the labels on the inputs (this class provides an override to make labels _smaller_ than the default enormous ones on customer-facing honeycrisp, so it makes them _bigger_ than our default admin styles when applied here.

app/views/hub/users/edit.html.erb
```diff
@@ -16,15 +16,22 @@
...
-        <p>
-          <strong><%= t("general.email") %>:</strong> <%= @user.email %>
-          <%= link_to t("general.resend_invitation"), user_profile_resend_invitation_path(user_id: @user.id), class: "button button-small", method: :put %>
-        </p>
-
-        <%= f.cfa_input_field(:name, t("general.name")) %>
-        <%= f.cfa_input_field(:phone_number, t("general.phone_number"), type: "phone") %>
-        <%= f.cfa_select(:timezone, t("general.timezone"), timezone_select_options) %>
+      <%= form_with model: [:hub, @user], method: :put, local: true, builder: VitaMinFormBuilder, html: { class: 'form-card' } do |f| %>
+        <div>
```
bytheway875, line 20: This replaces the email (that cannot be edited) with a disabled input with greyed/not-allowed cursor AND only displays the resend invite button if the client has not already accepted the invite.

app/views/hub/users/edit.html.erb
```diff
@@ -36,17 +43,16 @@
...
-            <li><%= link_to t("general.coalition_lead"), edit_role_hub_user_path({user: {role: CoalitionLeadRole::TYPE}}) %></li>
-            <li><%= link_to t("general.organization_lead"), edit_role_hub_user_path({user: {role: OrganizationLeadRole::TYPE}}) %></li>
-            <li><%= link_to t("general.site_coordinator"), edit_role_hub_user_path({user: {role: SiteCoordinatorRole::TYPE}}) %></li>
-            <li><%= link_to t("general.team_member"), edit_role_hub_user_path({user: {role: TeamMemberRole::TYPE}}) %></li>
-          </ul>
-        <% end %>
+        <div style="width: 375px">
+          <%= render('components/molecules/reveal', title: t(".reassign")) do %>
+            <ul class="list--bulleted">
+              <% User.roles.each do |role_type| %>
```
bytheway875, line 49: This limits the display of the roles you can change to to the ones that the current user is allowed to manage. Before, any user could click to try to change a User they could see to be an admin, even if they did not have permissions to actually do that.

app/views/hub/users/edit_role.html.erb
```diff
@@ -10,41 +10,59 @@
 
   <%= form_for(@user, method: :patch, url: update_role_hub_user_path, local: true, builder: VitaMinFormBuilder) do |f| %>
 
-    <%= f.hidden_field :role, value: params.dig(:user, :role) %>
-    <% if params.dig(:user, :role) == AdminRole::TYPE %>
+    <%= f.hidden_field :role, value: params[:role] %>
```
bytheway875, line 13: The nested {user: { role: } param didn't really serve a functional purpose, so I simplified it down to just coming in as params[:role] without nesting in all implementations.

app/views/invitations/index.html.erb
```diff
@@ -1,77 +1,28 @@
...
-        <% if can? :manage, TeamMemberRole %>
-          <%= link_to new_user_invitation_path(user: {role: TeamMemberRole::TYPE}), class: "button" do %>
-            <%= t(".actions.new_team_member_invite") %>
-          <% end %>
-        <% end %>
+    <%= form_for :invitation, url: "/hub/invitation/new", method: "get" do |f| %>
+    <div class="form-group">
+      <%= label_tag(:role, "What type of user do you want to invite?", class: "form-question") %>
+      <div class="select form-width--long">
+        <%= select_tag :role, options_for_select(User.roles.map {|role_type| [readable_role_from_role_type(role_type), role_type] if can?(:manage, role_type.constantize)}.compact), class: "select__element" %>
```
caitbot, line 8: sweet this is nice

caitbot (approved):
looks great ✨

PR #1878, opened 2021-12-01
Add sidebar toggle that collapses and expands #179744221

caitbot:
It toggles!

image

image

### DIFF
--- a/app/javascript/hub/MainMenuComponent.js
+++ b/app/javascript/hub/MainMenuComponent.js
@@ -1,8 +1,21 @@
 export default function MainMenuComponent() {
-    const mainMenu = document.querySelector('[data-component="MainMenuComponent"]')
+    const mainMenu = document.querySelector('[data-component="MainMenuComponent"]');
+    highlightSelectedPageNavigation(mainMenu);
+    setSidebarExpandCollapse(mainMenu);
+}
+
+function highlightSelectedPageNavigation(mainMenu) {
     mainMenu.querySelectorAll('.menu-item').forEach((item) => {
         if (item.href === window.location.href) {
             item.classList.toggle('selected');
         }
+    });
+}
+
+function setSidebarExpandCollapse(mainMenu) {
+    mainMenu.querySelector('.toggle').addEventListener("click", () => {
+        const classes = mainMenu.classList;
+        classes.toggle("expanded");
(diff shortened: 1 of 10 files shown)


app/views/layouts/hub.html.erb
```diff
@@ -46,7 +46,7 @@
 
 <body class="admin honeycrisp-compact hub">
 <div class="columns">
-  <div class="columns--child">
```
caitbot, line 49: Removing this because the `overflow-y: auto;` results in the toggle icon being half-hidden.

caitbot: Discussed w/ Suran and removing the scroll bar on the main menu is ok.

app/javascript/hub/MainMenuComponent.js
```diff
@@ -1,8 +1,27 @@
...
         if (item.href === window.location.href) {
             item.classList.toggle('selected');
         }
     });
 }
+
+function toggleSidebarExpandCollapse(mainMenu) {
+    mainMenu.querySelector('.toggle').addEventListener("click", () => {
+        const classes = mainMenu.classList;
+        classes.toggle("expanded");
```
bytheway875, line 18: this is kind of nitty, but i feel like it would be simpler if we let one of the states be the "default" appearance of the component and then only toggle ONE class on or off

app/views/layouts/hub.html.erb
bytheway875: is this removing the vertical scroll? On the external screen I use, I do need to scroll to see all of the current options. My worry is that as we keep adding options, we'll run into a situation where you can't see all of the links on the screen if we don't account for overflow. BUT maybe this is another overflow.

caitbot: It removes the vertical scroll on the sidebar, yeah. Because when it was present, you couldn't toggle the sidebar. Suran and I checked and our impression was you had to make the browser pretty small to start getting rid of buttons.

Would you mind sharing a screenshot? Cause we might have totally missed something.

app/javascript/hub/MainMenuComponent.js
caitbot: I like that, I'll clean it up

PR #1884, opened 2021-12-03
Add a data mask to SSN and ITIN input fields

caitbot:
Adds a javascript plugin to allow data-masking on fields, which adds hyphens into the field as you type.

For [this story](https://www.pivotaltracker.com/story/show/179062568)

### DIFF
--- a/app/forms/ctc/base_primary_filer_form.rb
+++ b/app/forms/ctc/base_primary_filer_form.rb
@@ -9,7 +9,10 @@ module Ctc
     validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
     validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin" }
 
-    before_validation { primary_ssn&.remove!("-") }
+    before_validation do
+      primary_ssn&.remove!("-")
+      primary_ssn_confirmation&.remove!("-")
+    end
     with_options if: -> { (primary_ssn.present? && primary_ssn != intake.primary_ssn) || primary_ssn_confirmation.present? } do
       validates :primary_ssn, confirmation: true
       validates :primary_ssn_confirmation, presence: true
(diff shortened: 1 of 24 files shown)


bytheway875:
I _think_ there's something built into honeycrisp that will do this as well, but I honestly prefer us relying on whatever the popular JS implementation is for this at the moment because I know people had talked about the honeycrisp one being kind of wonky.

embarnard (changes requested):
Looks good! But looks like a test is failing at spec/features/ctc/portal_spec.rb:254 and giving a validation error that the SSN or ITIN field "Can't be blank".

app/forms/ctc/base_spouse_form.rb
```diff
@@ -9,6 +9,7 @@ class BaseSpouseForm < QuestionsForm
     validates :spouse_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include?(spouse_tin_type) && spouse_ssn.present? }
     validates :spouse_ssn, individual_taxpayer_identification_number: true, if: -> { spouse_tin_type == "itin" }
 
+    before_validation { spouse_ssn&.remove!("-") }
```
embarnard, line 12: looks good!

embarnard (approved):
Woohoo!! 😊

app/forms/ctc/base_primary_filer_form.rb
```diff
@@ -9,6 +9,10 @@ class BasePrimaryFilerForm < QuestionsForm
     validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
     validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin" }
 
+    before_validation do
```
bytheway875, line 12: I feel like we're going to want to make sure this is consistent across all implementations of SSN persistance. Maybe we can have the SSN/ITIN validator handle scrubbing when validating, and then add the removal to the before_save on intake when these fields are present. 

(I also thought that when we discussed length of 11 SSN fields, Travis said that this scrubbing was already being done somewhere consistently to ensure only numbers got into the DB?)

caitbot: Agreed, I am having trouble finding where the scrubbing is happening though

caitbot: Line 290 in ctc_intake

PR #1885, opened 2021-12-03
Implement a custom GyrDiskService for ActiveStorage to workaround CircleCI issues

tgrathwell:
As seen in this issue, certain versions of Docker and certain versions of the Linux
kernel manifest a problem where creating tempfiles wiht `IO.copy_stream` doesn't work:
https://github.com/docker/for-linux/issues/1015

CircleCI seems to have upgraded their linux kernels and now we're seeing this problem.
A temporary workaround is to change the ActiveStorage code to not use IO.copy_stream
by making a custom service (since the Disk service is only for test anyway so we can
do whatever we want)

### DIFF
--- a/app/lib/active_storage/service/gyr_disk_service.rb
+++ b/app/lib/active_storage/service/gyr_disk_service.rb
@@ -1,0 +1,20 @@
+require "active_storage/service/disk_service"
+
+class ActiveStorage::Service::GyrDiskService < ActiveStorage::Service::DiskService
+  def upload(key, io, checksum: nil, **)
+    instrument :upload, key: key, checksum: checksum do
+      # This line from the real DiskService fails on certain docker + kernel combos, so we will do it a different way so our CircleCI builds work:
+      # IO.copy_stream(io, make_path_for(key))
+
+      # We should be able to remove this if CircleCI starts using a different linux kernel version (currently 5.4.0-1060-aws)
+
+      if io.respond_to?(:path)
+        File.write(make_path_for(key), File.read(io.path))
+      else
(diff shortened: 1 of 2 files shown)


app/lib/active_storage/service/gyr_disk_service.rb
```diff
@@ -0,0 +1,17 @@
...
+class ActiveStorage::Service::GyrDiskService < ActiveStorage::Service::DiskService
+  def upload(key, io, checksum: nil, **)
+    instrument :upload, key: key, checksum: checksum do
+      # This line from the real DiskService fails on certain docker + kernel combos, so we will do it a different way so our CircleCI builds work:
+      # IO.copy_stream(io, make_path_for(key))
+
+      # We should be able to remove this if CircleCI starts using a different linux kernel version (currently 5.4.0-1060-aws)
+
+      File.write(make_path_for(key), io.read)
+      io.rewind
```
caitbot, line 12: why do rewind? is this because it is the state that `IO.copy_stream` leaves `io` in?

caitbot (approved):
this looks good to me and makes sense as a work-around

app/lib/active_storage/service/gyr_disk_service.rb
tgrathwell: I have some paranoia about IO files getting angry if someone tries to `read` them twice, though I didn't actually see that problem in real life yet for this code

everything I've tried so far has made some tests pass but others fail; I've done another revision where it only tries to `File.write` if the `io` object is an actual file, let's see if that fares any better

PR #1887, opened 2021-12-06
Full Ssn on hub GYR intake and edit 180137844

bytheway875:


app/forms/hub/create_client_form.rb
```diff
@@ -52,6 +55,16 @@ class CreateClientForm < ClientForm
     validate :at_least_one_tax_return_present
     validates :state_of_residence, inclusion: { in: States.keys }
     validates :preferred_interview_language, presence: true, allow_blank: false
+    validates :primary_tin_type, presence: true
+    validates_confirmation_of :primary_ssn
+    validates_presence_of :primary_ssn_confirmation, if: :primary_ssn
+    validates_presence_of :spouse_ssn_confirmation, if: :spouse_ssn
+    validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
+    validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin"}
```
alaroia-cfa, line 63: Can/should we only allow these three tin types for primary SSN? I ask b/c as-is, it looks like if they enter ATIN as the tin type, they can skip all validation, which just seems bad.

app/forms/hub/update_client_form.rb
```diff
@@ -39,13 +41,23 @@ class UpdateClientForm < ClientForm
...
 
+    validates :primary_tin_type, presence: true
+    validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
+    validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin"}
+
+    validates_confirmation_of :spouse_ssn, if: -> { filing_joint == "yes" }
+    validates :spouse_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include?(spouse_tin_type) && filing_joint == "yes"}
+    validates :spouse_ssn, individual_taxpayer_identification_number: true, if: -> { spouse_tin_type == "itin" && filing_joint == "yes" }
+
+    attr_accessor :client
```
alaroia-cfa, line 52: Dumb question: What's the value of adding this?

app/views/shared/_client_primary_info_fields.html.erb
```diff
@@ -48,33 +48,26 @@
...
-              :primary_ssn_confirmation,
-              "Re-enter SSN/ITIN",
-              type: :tel,
-              classes: ["form-width--name"],
-              options: { maxlength: 11 }
-          ) %>
-      <% end %>
+      ) %>
+    <% end %>
+    <% if is_ctc %>
```
alaroia-cfa, line 69: It seems to me that this `if is_ctc` isn't related to this pull request, but is also definitely a good idea. Is that accurate? No change requested, just curious.

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -862,8 +872,13 @@
           "timezone" => [nil, "America/Chicago"],
           "primary_last_name" => [intake.primary_last_name, "Name"],
           "primary_first_name" => [intake.primary_first_name, "Updated"],
-          "state_of_residence" => [nil, "CA"],
           "primary_last_four_ssn" => ["[REDACTED]", "[REDACTED]"],
+          "state_of_residence" => [nil, "CA"],
+          "primary_ssn" => ["[REDACTED]", "[REDACTED]"],
+          "primary_tin_type" => [nil, "ssn"],
+          "spouse_last_four_ssn" => ["[REDACTED]", "[REDACTED]"],
```
alaroia-cfa, line 879: Should these `last_four_ssn` properties disappear?

alaroia-cfa (approved):
This is good as-is. I have some questions and suggestions in the comments.

A general question: Do you think we should copy the validation into the intake/gyrintake/ctcintake model? I'm inclined to say yes, although I know that duplicated validation between forms and models is a constant paint-point for us, so feel free to ignore that suggestion for now and we can see if we want to do it later.

app/forms/hub/create_client_form.rb
bytheway875: ATIN is not an option on the dropdown for primary and spouse so I think the implementation (which is copy and pasted from the other implementations) doesn't account for that but we could make sure that primary and spouse tin_types must be within the set of approved types?

For in intake steps, I noticed that we were doing this "if its this if its that" without validating that tin_type was present, which shouldnt really be possible based on how the form was working but I think it was somehow causing some of the primary_snn.last_4 bugs we saw. I have added validates_presence_of :primary_tin_type in all places for that reason

app/forms/hub/update_client_form.rb
bytheway875: in the "ctc only" part of the view (that form already includes this line), we call @form.client to get some logic around whether the confirmation field for ssn should exist. Now that we're reusing that logic, we need to add this.

app/views/shared/_client_primary_info_fields.html.erb
bytheway875: You need to use the expander to see how the full diff was affected here. There is an unchanged line for "is_ctc" that starts on line 45. i've now added an end on line 51 so that only that part is is_ctc. Now, the full ssn section is shared between both types of intakes. But we do not need IP PIN for GYR, so that is wrapped in a is_ctc section.

bytheway875:
@alaroia-cfa Adding validations at the model level to intake is really difficult because we push content on to these objects incrementally. "Validate something this way if this thing exists" would likely just adds more complexity because we'd need validations on forms and on the model.

However, we talked briefly at computer gardening next week about returning clients and looking at content that is relevant to a client for all tax years vs a specific tax year and we talked about maybe splitting off information that sustains through tax years onto client vs onto intake, or splitting the intake up into smaller pieces and that change might be worthwhile then.

PR #1888, opened 2021-12-06
Update ruby to 2.6.9

alaroia-cfa:


tgrathwell:
I'm gonna be the guy and ask, if it's gonna take all these commits to bump it up, why not 3.x?

alaroia-cfa:
The main reason I didn't go to 3.x is that I thought it'd be quick. In fact it wasn't quick -- there are tests failing for build/environment/dependency/etc. reasons that I haven't diagnosed yet, but seem important to diagnose.

I figured that 3.x would be more work, but now that I see all this, it's not clear to me that 3.x would be more work. So maybe that's a better idea.

Maybe I'll just close this now and say it was a failed experiment.

PR #1889, opened 2021-12-06
Update filed 2020 to already_filed everywhere

bytheway875:


caitbot (approved):
This looks good to me.

I was surprised to see that `already_filed` is a column we have had since June and it looks like it isn't currently being used.

I like this moving away from being year specific though, this is nice.

bytheway875:
@caitbot Yeah I was trying to remember what was going on there. Already_filed was only on GYR intake. I'm wondering if there was a ticket to add already_filed to GYR around the same time we were doing CTC stuff, we didn't know we had it or that the implementations overlapped so we didnt use it. It was also interesting that locally i had an attribtue called filed_2020 on intake, but when i went into aptible production I _didnt_ have a field called filed_2020 on intake?

PR #1890, opened 2021-12-06
Create StateRoutingTarget

caitbot:
For work on https://www.pivotaltracker.com/story/show/180351602 and subsequent stories

### DIFF
--- a/db/migrate/20211206193923_create_state_routing_target.rb
+++ b/db/migrate/20211206193923_create_state_routing_target.rb
@@ -1,0 +1,9 @@
+class CreateStateRoutingTarget < ActiveRecord::Migration[6.1]
+  def change
+    create_table :state_routing_targets do |t|
+      t.string :state_abbreviation, null: false, index: true
+      t.references :target,null: false, polymorphic: true, index: true
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20211206193923_create_state_routing_target.rb
```diff
@@ -0,0 +1,9 @@
+class CreateStateRoutingTarget < ActiveRecord::Migration[6.1]
+  def change
+    create_table :state_routing_targets do |t|
+      t.string :state_abbreviation, null: false
+      t.references :target,null: false, polymorphic: true, index: true
```
bytheway875, line 5: nitty, but missing a space after the :target on line 5 😂

bytheway875: does polymorphic:true automatically create both the target_id and target_type attrs? thats cool. i dont think i knew that

db/schema.rb
```diff
@@ -900,6 +900,15 @@
     t.index ["vita_partner_id"], name: "index_source_parameters_on_vita_partner_id"
   end
 
+  create_table "state_routing_targets", force: :cascade do |t|
+    t.datetime "created_at", precision: 6, null: false
+    t.string "state_abbreviation", null: false
+    t.bigint "target_id", null: false
```
bytheway875, line 906: indeed it does! cool

bytheway875 (approved):

db/migrate/20211206193923_create_state_routing_target.rb
caitbot: lol, good catch, and yupp, it is cool

PR #1891, opened 2021-12-06
Move existing intakes to archived_intakes_2021

embarnard:

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -221,8 +221,8 @@
 #  with_limited_english_navigator                       :boolean          default(FALSE)
 #  with_unhoused_navigator                              :boolean          default(FALSE)
 #  zip_code                                             :string
-#  created_at                                           :datetime
-#  updated_at                                           :datetime
+#  created_at                                           :datetime         not null
+#  updated_at                                           :datetime         not null
 #  client_id                                            :bigint
 #  visitor_id                                           :string
 #  vita_partner_id                                      :bigint
--- a/spec/controllers/hub/clients_controller_spec.rb
+++ b/spec/controllers/hub/clients_controller_spec.rb
@@ -409,6 +409,22 @@ RSpec.describe Hub::ClientsController do
             expect(attrib.strip).to eq(message_summary.strip)
           end
         end
+
+        context "when there are clients with no current intakes (clients from previous tax years)" do
+          let!(:former_year_client) { create :client, vita_partner: organization, intake: build(:intake, :filled_out) }
+          let!(:former_year_tax_return) { create :tax_return, :intake_in_progress, client: former_year_client, year: 2020, assigned_user: assigned_user }
+
+          before do
+            # In reality this intake would be moved to the `archived_intakes_2021` table, but removing it from the DB is good enough for our purposes
+            former_year_client.intake.destroy
+          end
+
+          it "does not show those clients in the list" do
+            get :index
+
+            expect(assigns(:clients)).to match_array([george_sr, michael, tobias])
+          end
+        end
       end
 
       context "sorting and ordering" do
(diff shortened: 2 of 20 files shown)


app/models/documents_request.rb
```diff
@@ -2,18 +2,21 @@
...
-#  id           :bigint           not null, primary key
-#  completed_at :datetime
-#  created_at   :datetime         not null
-#  updated_at   :datetime         not null
-#  intake_id    :bigint
+#  id                       :bigint           not null, primary key
+#  completed_at             :datetime
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#  archived_intakes_2021_id :bigint
```
jenny-heath, line 9: i still don't see the point of having "archived" in the name but i'm just about ready to stop feeding that fed horse

app/models/intake.rb
```diff
@@ -253,6 +253,7 @@
 #
 # Foreign Keys
 #
+#  fk_rails_...  (bank_account_id => bank_accounts.id)
```
jenny-heath, line 256: not following this new bank account foreign key...maybe it will become apparent as i keep reading

db/migrate/20211206191929_archive_2021_intakes.rb
```diff
@@ -0,0 +1,28 @@
+class Archive2021Intakes < ActiveRecord::Migration[6.1]
```
jenny-heath, line 1: will want to chat about this file, mostly because it seems consequential

db/migrate/20211206191930_create_intakes.rb
```diff
@@ -0,0 +1,242 @@
...
+      t.boolean :with_vita_approved_taxpayer_id, default: false
+      t.string :zip_code
+
+      t.timestamps
+    end
+
+    set_pk_sequence!(
+      'intakes',
+      '(SELECT MAX(id) FROM archived_intakes_2021)'
+    )
```
jenny-heath, line 240: tried googling this but probably still need an explanation

db/migrate/20211207212323_allow_documents_requests_to_be_on_new_or_old_intakes.rb
```diff
@@ -0,0 +1,7 @@
+class AllowDocumentsRequestsToBeOnNewOrOldIntakes < ActiveRecord::Migration[6.1]
+  def change
+    rename_column :documents_requests, :intake_id, :archived_intakes_2021_id
+    add_foreign_key :documents_requests, :archived_intakes_2021
+    add_reference :documents_requests, :intake, foreign_key: true, index: true
```
jenny-heath, line 5: is this the only table we have to do this for?

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -409,6 +409,22 @@
...
           end
         end
+
+        context "when there are clients with no current intakes (clients from previous tax years)" do
+          let!(:former_year_client) { create :client, vita_partner: organization, intake: build(:intake, :filled_out) }
+          let!(:former_year_tax_return) { create :tax_return, :intake_in_progress, client: former_year_client, year: 2020, assigned_user: assigned_user }
+
+          before do
+            # In reality this intake would be moved to the `archived_intakes_2021` table, but removing it from the DB is good enough for our purposes
+            former_year_client.intake.destroy
```
jenny-heath, line 419: i assume you destroyed it instead of just not creating it in the first place because of some factory rule or something

app/models/intake.rb
tgrathwell: it was just missing... we had an association between intakes and bank accounts but no foreign key to guarantee you can't put 9999 in there or some other arbitrary value

db/migrate/20211206191930_create_intakes.rb
tgrathwell: want to make sure that the first id in the newly-created `intakes` table carries over from where `archived_intakes_2021` left off. prevents us from making mistakes where we confuse a 2022 intake id for a 2021 intake id

app/models/intake.rb
tgrathwell: on further investigation this column should be dropped... we have columns in the DB for both `bank_accounts.intake_id` and `intakes.bank_account_id` but only the former is populated

PR #1893, opened 2021-12-07
Use node 16 in bin/setup

alaroia-cfa:
Without this change, my system wants to use node 17, which seems to have problems for us.


embarnard (approved):
Looks good! :)

PR #1896, opened 2021-12-07
Don't show count when it is too large to be quickly counted

alaroia-cfa:

### DIFF
--- a/app/controllers/hub/clients_controller.rb
+++ b/app/controllers/hub/clients_controller.rb
@@ -10,10 +10,20 @@ module Hub
     load_and_authorize_resource except: [:new, :create]
     layout "hub"
 
+    MAX_COUNT = 1000
+
     def index
       @page_title = I18n.t("hub.clients.index.title")
       # @tax_return_count HAS to be defined before @clients, otherwise it will cause SQL errors
-      @tax_return_count = TaxReturn.where(client: filtered_clients.with_eager_loaded_associations.without_pagination).size
+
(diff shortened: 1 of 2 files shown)


alaroia-cfa:
On pages like this:

image

You can click the checkbox to select all, and you see a message at the bottom:

image

This shows how it would behave if we merge this pull request: In the case that there are more than 1000 total returns matching the search query, we show "> 1000" instead of the precise number. This is faster to compute.

Note that this is a mockup -- here there are in fact only 7 clients matching the query. Pretend there are more than 1000 total returns matching the query. :)

alaroia-cfa:
Further work to do:

* Make the same change in other places that compute `@tax_return_count`, probably by adding a method to `ApplicationRecord` called `count_up_to_threshold`, e.g. AssignedClientsController
* See if the same change is needed in TaxReturnSelectionsController

tgrathwell:
A couple of things that occur to me:

* We should be doing this max-counting on the actual paginated list; in the corner it should say "Displaying 1-25 of 1000+ clients" when appropriate. That converts another full table scan into a mere... some..table..scan
* We could defer doing the `@tax_return_count` at all until someone clicks the checkbox to bulk message people. Some kinda javascript may be helpful or required.
* We should probably prevent bulk messaging over a certain limit. What's the bulkiest bulk message we ever expect someone to send? Should it be disabled for CTC clients?

alaroia-cfa:
- When it's >1000 just show the empty string as the count (goes from 999 to "")
- Keep the count on the bulk actions page itself

PR #1908, opened 2021-12-08
Add spec for StateRoutingTarget model [#180351533]

alaroia-cfa:
In the process, I needed to change the implementation of some validation.


caitbot (approved):
sweeet

PR #1909, opened 2021-12-08
Add coalition edit and create page

caitbot:
image


config/locales/es.yml
```diff
@@ -1090,6 +1090,11 @@ es:
       already_taken: "%{zip_code} ya está enrutado a <a href=%{vita_partner_path}>%{vita_partner_name}</a>."
       confirm: "¿Estás segura de que quieres eliminar? %{code} de %{name}?"
       success: Se agregó %{code} a %{name}.
+    coalitions:
```
caitbot, line 1093: Is this how we are doing translations? My assumption is this will show up in Transifex but I don't actually know our current process... I can also just google translate this since it is in the Hub.

app/controllers/hub/coalitions_controller.rb
```diff
@@ -0,0 +1,45 @@
+module Hub
+  class CoalitionsController < ApplicationController
+    include AccessControllable
+    before_action :require_sign_in
+    load_and_authorize_resource :coalition, parent: false
+
+    layout "hub"
+
+    def new
+      @coalition = Coalition.new
```
caitbot, line 10: Probably comes for free from cancancan

app/controllers/hub/coalitions_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def create
+      @form = CoalitionForm.new(@coalition, coalition_params.merge(state_routing_target_params))
```
caitbot, line 14: could be moved to a before_action shrug

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,29 @@
+module Hub
+  class CoalitionForm < Form
+
+    def initialize(coalition, form_params)
+      @coalition = coalition
+      @params = form_params
```
caitbot, line 6: Attributes are more common for our forms. It'd be better IMHO that way but it doesn't actually matter.

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def save
+      ActiveRecord::Base.transaction(joinable: false, requires_new: true) do
+        @coalition.name = @params[:name]
+        return false unless @coalition.save
+
+        existing_states = @coalition.state_routing_targets.map(&:state_name)
```
caitbot, line 14: If we normalize `form_states` into `form_state_abbreviations` we can remove the `state_name` method from `StateRoutingTarget` and use `.pluck`.

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,29 @@
...
+        form_states = @params[:states].split(",") # states are comma delimited in a string, i.e. "Ohio,California"
+
+        new_states = form_states - existing_states
+        new_states.each do |state|
```
caitbot, line 18: @coalition.state_routing_targets.build(state_abbreviation: state_abbreviation)

and then remove the transaction, calling save once at the end of all of this

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,29 @@
...
+          raise ActiveRecord::Rollback unless StateRoutingTarget.create(state_abbreviation: States.key_for_name(state), target: @coalition)
+        end
+
+        states_to_delete = existing_states - form_states
+        state_abbrs = states_to_delete.map { |state| States.key_for_name(state) }
```
caitbot, line 23:
```
@coalition.state_routing_targets.each { |t| t.destroy unless form_states.include(t.state_abbreviation) }
```

don't plan what to do - just do it

caitbot: actually the build approach is better nm

app/models/coalition.rb
```diff
@@ -14,4 +14,5 @@
 class Coalition < ApplicationRecord
   validates :name, uniqueness: true
   has_many :organizations, class_name: "Organization", foreign_key: "coalition_id"
+  has_many :state_routing_targets, class_name: "StateRoutingTarget", foreign_key: "target_id"
```
caitbot, line 17: `as: :target`

app/models/state_routing_target.rb
```diff
@@ -15,12 +15,18 @@
 #
 class StateRoutingTarget < ApplicationRecord
   belongs_to :target, polymorphic: true
+  alias_attribute :state, :state_abbreviation
```
caitbot, line 18: 🗑️

app/models/state_routing_target.rb
```diff
@@ -15,12 +15,18 @@
...
+
   validate :valid_state_abbreviation
 
   def routing_percentage
     (routing_fraction * 100).to_i
   end
 
+  def state_name
```
caitbot, line 26: trash can

app/views/hub/coalitions/_form.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <%= form_with model: @coalition, url: url, method: http_method, local: true, builder: VitaMinFormBuilder, id: "coalition-form" do |f| %>
+    <h1 class="form-card__title">
+      <%= @title %>
+    </h1>
+
+    <%= f.cfa_input_field :name, t("general.name") %>
+      <div class="form-group state-select">
+        <label for="state-select" class="form-question"><%= t("general.state") %></label>
+        <div>
+          <%= text_field_tag("state_routing_targets[states]", @coalition.state_routing_targets.map(&:state_name).to_json.html_safe, id: "state-select", class: "text-input multi-select-state") %>
```
caitbot, line 11: pass them through the helper anyways it'll be a little longer winded

app/views/hub/coalitions/_form.html.erb
```diff
@@ -0,0 +1,24 @@
+<div>
+  <%= form_with model: @coalition, url: url, method: http_method, local: true, builder: VitaMinFormBuilder, id: "coalition-form" do |f| %>
```
caitbot, line 2: `local_assigns[:http_method]`

config/locales/es.yml
```diff
@@ -720,6 +720,11 @@ es:
           status: actualizó estado
           success: 'Éxito: Acción completo! %{action_list}.'
           text_message: envió mensjaje de texto
+    coalitions:
```
caitbot, line 723: google translate these things

spec/controllers/hub/coalitions_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        }
+      }
+    end
+
+    it_behaves_like :a_post_action_for_admins_only, action: :create
+
+    context "as an authenticated admin user" do
+      before { sign_in admin_user }
+
+      it "creates the coalition and associated routing targets, and redirects to the organizations page" do
```
caitbot, line 27: `context "with valid params" do`

then add invalid params

spec/controllers/hub/coalitions_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
+require "rails_helper"
+
+RSpec.describe Hub::CoalitionsController, type: :controller do
+  let(:admin_user) { create :admin_user }
+
+  describe "#new" do
```
caitbot, line 6: mock the form's save to return true/false

spec/controllers/hub/coalitions_controller_spec.rb
caitbot: but the mocking still you know

spec/controllers/hub/coalitions_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+          expect(coalition.state_routing_targets.pluck(:state_abbreviation)).to match_array ["AL", "OH", "UT"]
+
+          expect(flash[:notice]).to eq "Changes saved"
+          expect(response).to redirect_to edit_hub_coalition_path(id: coalition.id)
+        end
+      end
+
+      context "when the coalition object is not valid" do
+        before do
+          allow_any_instance_of(Coalition).to receive(:save).and_return false
```
caitbot, line 86: when you mock the form you can ditch `allow_any_instance_of` yay b/c rubocop complains about this so yay phew saved

spec/factories/state_routing_targets.rb
```diff
@@ -0,0 +1,17 @@
+# == Schema Information
+#
+# Table name: coalitions
+#
+#  id         :bigint           not null, primary key
```
caitbot, line 5: This is the wrong place, but state routing target can/should grow a uniqueness constraint on the combination of target_type + target_id + state_abbreviation. Could be a Rails validation too so then we get nice errors.

app/controllers/hub/coalitions_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    include AccessControllable
+    before_action :require_sign_in
+    load_and_authorize_resource :coalition, parent: false
+
+    layout "hub"
+
+    def new; end
+
+    def create
+      @form = CoalitionForm.new(coalition_params.merge(state_routing_target_params).merge({ coalition: @coalition }))
```
alaroia-cfa, line 12: Trivial nit: This can be `.merge(coalition: @coalition)`:

```
irb(main):001:0> {}.merge(hi: 'there')
=> {:hi=>"there"}
```

app/controllers/hub/coalitions_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        redirect_to hub_organizations_path
+      else
+        render :new
+      end
+    end
+
+    def edit; end
+
+    def update
+      @form = CoalitionForm.new(coalition_params.merge(state_routing_target_params).merge({ coalition: @coalition }))
```
alaroia-cfa, line 23: Same as above about `merge`.

app/controllers/hub/coalitions_controller.rb
alaroia-cfa, line 12: Trivial nit:

Since you're asking cancancan to `load_and_authorize_resource` with `:coalition`, it will already call `coalition_params` as part of creating `@coalition`. Given that, I wonder if we can simplify the form to not take `name` as param, since it'll be part of the `@coalition` already. In that case, it might look like:

```
CoalitionForm.new({coalition: @coalition}.merge(state_routing_target_params))
```

This would allow us to remove `@coalition.name = name` from the form, too.

Another form that I would prefer even more would be:

```
CoalitionForm.new(@coalition, state_routing_target_params[:states])
```

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,18 @@
+module Hub
+  class CoalitionForm < Form
+    include ActiveModel::Model
+
+    attr_accessor :coalition, :name, :states
+
+    def save
+      coalition.name = name
+      form_states = states.split(",").map { |abbr| States.key_for_name(abbr) } # states are comma delimited in a string, i.e. "Ohio,California"
```
alaroia-cfa, line 9: Trivial nit:

You're using the variable name `abbr`, but it's not a state abbreviation. It's a full state name, so a better form might be `|name| States.key_for_name(name)`.

I suppose even slicker might be

```
['Ohio', 'New York'].map(&(States.method(:key_for_name)))
```

But actually that took me like 15 minutes of experimentation to figure out, so maybe it will be hard for people to read?

app/helpers/states.rb
```diff
@@ -19,6 +19,10 @@ def self.name_value_pairs
     self.states
   end
 
+  def self.html_safe_names
+    self.states.map(&:first).to_json.html_safe
```
alaroia-cfa, line 23: Trivial nit: It took me a little while to understand this. I would have found `hash.values.to_json.html_safe` easier to read.

app/models/state_routing_target.rb
```diff
@@ -15,6 +15,8 @@
 #
 class StateRoutingTarget < ApplicationRecord
   belongs_to :target, polymorphic: true
+
+  validates :state_abbreviation, uniqueness: { scope: [:target] }
```
alaroia-cfa, line 19: This is good.

Nit: I would also like to see the corresponding SQL constraint.

app/views/hub/coalitions/edit.html.erb
```diff
@@ -0,0 +1,11 @@
+<% @title = t(".title") %>
```
alaroia-cfa, line 1: Trivial nit: This `@title = ...` pattern is a little weird. I know it's copy-pasta, but if this view works fine without modifying instance variables, it would make 1% happier.

spec/forms/hub/coalition_form_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect(coalition.name).to eq "A New New Name"
+      end
+    end
+
+    context "removing states" do
+      let(:coalition) { create(:coalition, name: "A New Coalition") }
+      let(:params) { { name: "A New Coalition", states: "Utah,California", coalition: coalition } }
+
+      before do
+        create(:state_routing_target, state_abbreviation: "OH", target: coalition)
```
alaroia-cfa, line 48: Trivial nit: If you can do this in the `let(:coalition)` it's easier to read. e.g.

```
create(:coalition, name: "A New Coalition", state_routing_targets: [build(:state_routing_target, state_abbreviation: "OH")])
```

alaroia-cfa (approved):
🚢 

I discovered some more trivial nits, which may be mildly interesting. None are blockers, and I'm going to merge this in the interest of latency reduction.

app/controllers/hub/coalitions_controller.rb
bytheway875: I'll second the suggestion to send the params in as separate arguments! It seems much clearer in terms of looking at what the form is going to be responsible for.

app/forms/hub/coalition_form.rb
bytheway875: I wonder if we can set up the tagify implementation to display the full state name but pass the abbreviation to the backend for us so that we don't have to finagle this in the form. I feel like as often as possible the form should pass the "real" values we want to persist into the db.

app/forms/hub/coalition_form.rb
```diff
@@ -0,0 +1,18 @@
...
+      existing_states = coalition.state_routing_targets.pluck(:state_abbreviation)
+
+      new_states = form_states - existing_states
+      new_states.each { |state| coalition.state_routing_targets.build(state_abbreviation: state, target: @coalition) }
+      coalition.state_routing_targets = coalition.state_routing_targets.select { |t| form_states.include?(t.state_abbreviation) }
+      coalition.save
```
bytheway875, line 15: Did you try using Active Record's `replace` here? It will delete the ones that arent included in the collection but if they are included in the collection I think it is smart enough to use the existing record. It probably looks a lot like what you're doing in lines 10 - 14.

```
routing_targets = form_states.map { |state| StateRoutingTarget.new(state_abbreviation: state, target: @coalition) }
coalition.state_routing_targets.replace(routing_targets)
```

The tradeoff is I think you'd have to do two db transactions to persist the data -- one for any changes to the name of the coalition and another for changes to the states included.

bytheway875: https://apidock.com/rails/ActiveRecord/Associations/CollectionProxy/replace

bytheway875: If we don't want to use AR replace, I'd def suggest pulling the state mapping and replacing logic on 10 -15 into a private method in the form since it's a super specific chunk of functionality.

app/views/hub/coalitions/_form.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <h1 class="form-card__title">
+      <%= @title %>
+    </h1>
+
+    <%= f.cfa_input_field :name, t("general.name") %>
+      <div class="form-group state-select">
+        <label for="state-select" class="form-question"><%= t("general.state") %></label>
+        <div>
+          <%= text_field_tag("state_routing_targets[states]",
+                             @coalition.state_routing_targets.map { |srt| States.name_for_key(srt.state_abbreviation) }.to_json.html_safe,
```
bytheway875, line 12: It seems like we're converting from state abbreviation to full name of the state and vice versa pretty often. Maybe we can add a method to the StateRoutingTarget class.

`@coalition.state_routing_targets.map { |srt| srt.full_state_name }`

app/views/hub/coalitions/_form.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      </div>
+    <div>
+      <%= f.submit t("general.save"), class: "button button--primary button--wide" %>
+    </div>
+  <% end %>
+</div>
+
+<% content_for :script do %>
+  <script>
+      window.taggableItems = <%= States.html_safe_names %>;
```
bytheway875, line 24: I'm now seeing that it may be the case that we might be doing the name conversion back and forth because its possibly simpler to display the already selected states properly when reloading ... 😬 

I still think if it's possible to deal with displaying the full name but passing the abbreviation around as values, that would be nice.

app/controllers/hub/coalitions_controller.rb
caitbot: Ty ty for these suggestions, I agree that this is much nicer.

app/forms/hub/coalition_form.rb
caitbot: Ooooh, @alaroia-cfa thats fancy.

@bytheway875, I totally agree. I was struggling to get Tagify to accept various formats, but I can do another pass. Especially since this will be re-used.

app/controllers/hub/coalitions_controller.rb
caitbot: Ok but the part where I got confused is providing them as two arguments but still using `attr_accessor` because what I was doing before was using an `initialize` with two separate arguments, then I switched to `attr_accessor` and I'm unsure how that should work.. I'll figure it out.

app/forms/hub/coalition_form.rb
caitbot: It appears `replace` isn't smart enough ):

I wish it was, that way would be much nicer. I pulled out the logic instead.

app/views/hub/coalitions/_form.html.erb
caitbot: Hahaha, I did this then got rid of it in first iteration of code review. We cleaned up a lot of the places that are doing this though!

app/views/hub/coalitions/_form.html.erb
caitbot: We were able to pass the abbrv as value! Good call!

PR #1910, opened 2021-12-08
FactoryBot: default any unspecified enums to 'no'

tgrathwell:
We were seeing some behavior where a test would flake 1/4 of the time
because the factories would randomly choose a `filing_joint` value between
['unfilled', 'no', 'yes', 'unsure'] and the test happened to fail
when it chose 'yes'

randomness in tests can cause these kind of shenanigans so lets try to
tone down the randomness


alaroia-cfa (approved):
I think this is a totally fine approach, and I'm enthusiastic about the loss of `.sample`.

PR #1911, opened 2021-12-08
Remove intakes.bank_account_id

tgrathwell:
It's always null, we seem to use bank_accounts.intake_id instead


alaroia-cfa (approved):
I confirmed this against prod's database and I agree. 🚢

alaroia-cfa (approved):

PR #1915, opened 2021-12-09
Provide standard deduction for 2021

bytheway875:
Because of the updates to how we're programmatically looking for standard deduction values I can't update the config tax year without updating standard deduction first.


alaroia-cfa (approved):
Sweet

PR #1917, opened 2021-12-09
Update application config tax year

bytheway875:


app/forms/backtaxes_form.rb
```diff
@@ -1,5 +1,5 @@
 class BacktaxesForm < QuestionsForm
-  set_attributes_for :intake, :needs_help_2017, :needs_help_2018, :needs_help_2019, :needs_help_2020, :source, :referrer, :locale, :visitor_id
```
caitbot, line 2: what happened to 2017?

spec/controllers/ctc/portal/portal_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe Ctc::Portal::PortalController do
   let(:intake) { create :ctc_intake, current_step: "/en/last/question" }
-  let(:client) { create :client, intake: intake, tax_returns: [create(:tax_return, year: 2020)] }
+  let(:client) { create :client, intake: intake, tax_returns: [create(:tax_return, year: 2021)] }
```
caitbot, line 5: would it make sense to set this to `TaxReturn.current_year`?

spec/controllers/ctc/questions/confirm_information_controller_spec.rb
```diff
@@ -48,7 +48,7 @@
 
         context "when not filing joint" do
           let(:intake) { create :ctc_intake }
-          let!(:tax_return) { create :tax_return, year: 2020, filing_status: "single", client: intake.client }
+          let!(:tax_return) { create :tax_return, year: 2021, filing_status: "single", client: intake.client }
```
caitbot, line 51: same here, do we want this to update each year?

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -623,7 +623,7 @@
 
       context "ordering tax returns" do
         let(:client) { (create :intake).client }
-        let!(:tax_return_2020) { create :tax_return, :intake_in_progress, client: client, year: 2020 }
+        let!(:tax_return_2020) { create :tax_return, :intake_in_progress, client: client, year: 2021 }
```
caitbot, line 626: I would keep this in sync with the test expectation

caitbot (approved):
This looks good to me!

Non-blocking thought -- but in a lot of places in the tests, you update `2020` to `2021` on the tax return. Is it possible we could replace that with `TaxReturn.current_tax_year` and save ourselves from having to do a very similar update next year? I noticed you used it in other places, but was unsure if there was a reason for hardcoding `2021` in other places.

bytheway875:
will merge and then fix some of the manual instances

PR #1918, opened 2021-12-09
Refactor some of the state multiselect and coalition form

caitbot:
A followup from #1909 

Make it better thanks to code review from AL and SB!

This was done while pairing w/ AL, but I forgot to add him to git duet ):


alaroia-cfa (approved):
I'm in

PR #1920, opened 2021-12-13
Make reveals that come after narrower content match the width of that content

jenny-heath:
disclaimer: the way i did this is maybe questionable but it didn't seem like there was a way to write a CSS rule that figured out which things to apply itself to. if that sentence doesn't make sense please call me!

summary: Anu and i looked at every reveal and adjusted their widths based on what looked right to her (hence the custom wrapper with the punny name)

screenshots!
whereas before, the reveal on this page would have been wider than the inputs above it, it is now the same width:
image

### DIFF
--- a/app/assets/stylesheets/components/_reveal.scss
+++ b/app/assets/stylesheets/components/_reveal.scss
@@ -9,9 +9,7 @@
 // this is used in places where we want the reveal to match the width of the form inputs above it
 .reveal-shrink-wrapper {
   .reveal {
-    @media screen and (min-width: $mobile-up) {
-      width: 475px;
-    }
+    max-width: $width-form-long;
   }
 }
 
--- a/app/assets/stylesheets/components/_reveal.scss
+++ b/app/assets/stylesheets/components/_reveal.scss
@@ -6,6 +6,15 @@
   padding: .2em .5em .3em .7em;
 }
 
+// this is used in places where we want the reveal to match the width of the form inputs above it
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/components/_reveal.scss
```diff
@@ -6,6 +6,15 @@
   padding: .2em .5em .3em .7em;
 }
 
+// this is used in places where we want the reveal to match the width of the form inputs above it
+.reveal-shrink-wrapper {
+  .reveal {
+    @media screen and (min-width: $mobile-up) {
+      width: 475px;
```
tgrathwell, line 13: it seems like this is trying to match the width of classes like `.form-width--long`, which is [actually defined as `$width-form-long` or 25em](https://github.com/codeforamerica/honeycrisp-gem/blob/main/app/assets/stylesheets/honeycrisp/atoms/_form-widths.scss#L10)

so we should probably put `$width-form-long` here instead

tgrathwell: `.form-width--long` also defines it as a `max-width` instead of a `width`, which is helpful for a small range of screen sizes (between `$mobile-up` (=481px) and ~500px)

Screen Shot 2021-12-14 at 11 25 02 AM

it's possible if we set it as `max-width` we can drop the media query

jenny-heath: oh wow that's much better

PR #1924, opened 2021-12-14
Remove source param logic that caused reporting problems overcounting organic google

bytheway875:
We pushed some code in October that attempted to escape hatch if session[:source] was already set, but it was actually incorrectly allowing session[:source] organic_google to be written in the case that the session value was already set and the client was entering from google.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -107,9 +107,11 @@ class ApplicationController < ActionController::Base
     session[:source]
   end
 
+  # Allow session[:source] to re-set on every interaction so we can track all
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -107,9 +107,11 @@ def source
     session[:source]
   end
 
+  # Allow session[:source] to re-set on every interaction so we can track all
+  # unique source entry points in Mixpanel.
```
bytheway875, line 111: This may end up changing to only set once and not override, but for now will revert to how it was originally set up.

PR #1925, opened 2021-12-15
Add some new destroy dependencies for client and update tests

bytheway875:
As we've changed the scope of what is attached to a client or a client's dependencies, we haven't always made sure that they get wiped when the client gets wiped. This attempts to fix that.

### DIFF
--- a/app/models/tax_return.rb
+++ b/app/models/tax_return.rb
@@ -50,7 +50,7 @@ class TaxReturn < ApplicationRecord
   has_many :assignments, class_name: "TaxReturnAssignment", dependent: :destroy
   has_many :tax_return_selection_tax_returns, dependent: :destroy
   has_many :tax_return_selections, through: :tax_return_selection_tax_returns
-  has_many :efile_submissions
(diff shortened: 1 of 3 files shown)


alaroia-cfa (approved):
🚢

app/models/tax_return.rb
```diff
@@ -50,7 +50,7 @@ class TaxReturn < ApplicationRecord
   has_many :assignments, class_name: "TaxReturnAssignment", dependent: :destroy
   has_many :tax_return_selection_tax_returns, dependent: :destroy
   has_many :tax_return_selections, through: :tax_return_selection_tax_returns
-  has_many :efile_submissions
+  has_many :efile_submissions, dependent: :destroy
```
tgrathwell, line 53: not to open any cans of wormses, but compliance-wise do we need to retain some kind of record of efile submissions even for clients who are deleted?

I don't know what normally triggers clients to be deleted so maybe this is not an issue but wondering if it's something we need to treat a little differently

PR #1926, opened 2021-12-15
Suppress Ruby warnings when running test suite

alaroia-cfa:
Before:

image

After:

image

It's at least _less_ noisy.


alaroia-cfa:
A better fix would be to use the `warnings` gem to silence the particular _path_ that's generating the warning.

tgrathwell:
> A better fix would be to use the `warnings` gem to silence the particular _path_ that's generating the warning.

okay i tried https://github.com/codeforamerica/vita-min/pull/1932

PR #1930, opened 2021-12-15
Expose archived dependents

bytheway875:


tgrathwell:
Are we going to do this for `intakes` too, for next year's data?

If we were being more paranoid it might be better to compute something like `age_in_2020` rather than directly exposing `birth_date` of all the dependents but I dunno if it matters enough to change

tgrathwell (approved):

bytheway875:
im not sure -- it seems like what they really want to know is if the person is eligible and if not why not. it may make more sense for us to persist that when we do the calculation to expose since keeping their eligibility reqs in line with ours might be hard...

PR #1932, opened 2021-12-17
Use the `warning` gem to hide regexp warnings generated by device_detector

tgrathwell:
until that gem is changed, someday

The warning ignoring part happens in a Rails initializer, so you'll still see the warnings if you run a test that loads device_detector but not Rails (do we have one of those?)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,7 @@ group :development, :test do
   gem 'parallel_tests'
   gem 'turbo_tests'
   gem 'timecop'
+  gem 'warning', require: false
 end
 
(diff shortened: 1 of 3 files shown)


alaroia-cfa:
@tgrathwell This sure does the job. 🚢 

Do you have an interest in fixing the Selenium/Capybara/`:browser_options` warning too, or nah? Either way feel free to merge.

tgrathwell:
@alaroia-cfa sure here for you https://github.com/codeforamerica/vita-min/pull/1933/files

PR #1933, opened 2021-12-17
Fix deprecation error with Capybara::Selenium::Driver

tgrathwell:
We used to conditionally initialize Capybara::Selenium::Driver with either :options
or :capabilities depending on the version, but the version matching was comparing
strictly against `~> 4.0`. Now that there's 4.1 we would have to update that
conditional, but it doesn't matter because we should just always send `capabilities`


alaroia-cfa:
@tgrathwell 😮 🙏

PR #1934, opened 2021-12-20
Change max message length to 900

bytheway875:


alaroia-cfa (approved):
@bytheway875 This is a great change. Consider updating tests in `text_message_length_limiter.spec.js` (not a blocker). Feel free to merge now or make those spec changes and then merge.

PR #1935, opened 2021-12-21
Move from Phonelib to Phony to add support for Puerto Rico phone numbers

alaroia-cfa:


app/services/phone_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  def self.formatted_phone_number(raw_phone_number)
+    return nil if raw_phone_number.nil?
+    return "" if raw_phone_number == ""
+
+    phony_normalized = Phony.normalize(raw_phone_number, cc: '1')
+    if Phony.plausible?(phony_normalized)
+      Phony.format(phony_normalized, format: :national)
```
caitbot, line 20: what does it mean for it to be `format: :national` as that seems to be what we mean by `formatted_phone_number`.

it could be useful to provide a comment with an example of the return format above `formatted_phone_number` and `with_country_code`

app/models/vita_provider.rb
```diff
@@ -87,11 +87,11 @@ def phone_number
   end
 
   def formatted_phone_number
-    phone_data.local_number
+    PhoneParser.formatted_phone_number(phone_number)
   end
 
   def sanitized_phone_number
```
caitbot, line 93: I am surprised to find that `sanitized_phone_number` means `with_country_code` but perhaps I will learn more reading the PhoneParser

app/views/questions/phone_verification/edit.html.erb
```diff
@@ -6,7 +6,7 @@
   <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
     <h1 class="h2"><%= @main_question %></h1>
     <p>
-      <%= t("views.questions.verification.body") %> <strong><%= Phonelib.parse(current_intake.sms_phone_number).local_number %></strong>
+      <%= t("views.questions.verification.body") %> <strong><%= PhoneParser.formatted_phone_number(current_intake.sms_phone_number) %></strong>
```
caitbot, line 9: can we use the helper's method `formatted_phone_number` instead to match the pattern in the other `erb` files

app/services/phone_parser.rb
caitbot: nevermind, these questions and concerns were answered by the specs, which provides the documentation of formatting that I wanted. yay yay yay

caitbot (approved):
sweet! this looks good to me, I left a few non-blocking comments/questions.

app/models/vita_provider.rb
caitbot: I suppose it can be called `sanitized` because it does not contain symbols or special characters, that makes enough sense.

alaroia-cfa:
I merged this to get it in ASAP because this thing has bit-rotted before and I'm fearful of that. I'll see if I can make a follow-up PR responding to your comments!

PR #1936, opened 2021-12-21
Normalize not sending emails or texts when clients cannot or have not opted into receive them

bytheway875:
This PR makes it so that we should not get errors raised because someone is opted in to a communication method but doesn't have a corresponding contact value.

This should not be very possible after these changes, but I guess could still happen in rare circumstances. I have it logging to a Datadog counter so that we can see if something's going wrong since the point of this ticket was to stop raising errors. However, it seems like to me that it SHOULD raise an error if you are opted in and dont have an email or sms phone number as we will no longer officially support this as a "workaround".

- Removes workaround for allowing people to opt in to a contact method without providing a valid contact method.
- Allows in person signature option to create a return without providing contact method

- Emails and texts sent through ClientMessagingService cannot be sent if the client has not opted into the contact method. If a client has opted in but has no valid contact method, we will increment a datadog counter.


app/forms/hub/client_form.rb
```diff
@@ -76,16 +77,26 @@ def dependent_validation_context
...
+        errors.add(:email_address, I18n.t("forms.errors.hub.communication_opt_in"))
+        errors.add(:sms_phone_number, I18n.t("forms.errors.hub.communication_opt_in"))
+      end
+    end
+
+    def contact_method_if_opted_in
+      if email_notification_opt_in == "yes" && email_address.blank?
+        errors.add(:email_address, I18n.t("forms.errors.hub.contact_method_required", attribute: "email"))
+      end
+      if sms_notification_opt_in == "yes" && sms_phone_number.blank?
```
jenny-heath, line 98: line 36 (`validates :sms_phone_number, presence: true, allow_blank: false, if: -> { opted_in_sms? }`) can probably be removed now that this is here

app/services/client_messaging_service.rb
```diff
@@ -62,12 +77,9 @@ def send_message_to_all_opted_in_contact_methods(client:, user:, body:, locale:
       args = { client: client, user: user, body: body }
       args[:tax_return] = tax_return if tax_return.present?
       args[:locale] = locale if locale.present?
-      if client.intake.email_notification_opt_in_yes? && client.email_address.present?
-        message_records[:outgoing_email] = send_email(**args)
-      end
-      if client.intake.sms_notification_opt_in_yes? && client.sms_phone_number.present?
-        message_records[:outgoing_text_message] = send_text_message(**args)
-      end
+      # returns nil unless client opted in to contact method
```
jenny-heath, line 80: does this comment refer to what `send_email` and `send_text_message` return? if so it looks like they actually return false

app/views/hub/clients/new.html.erb
```diff
@@ -8,10 +8,11 @@
           <h1 class="form-card__title">
             <%= @title %>
           </h1>
-          <%= f.cfa_select(:service_type, t("hub.clients.fields.service_type"), [["Drop off", "drop_off"]], options: { selected: "drop_off" } ) %>
-          <%= f.cfa_select(:signature_method, t("hub.clients.fields.signature_method"), signature_methods_for_select, options: { selected: "online" }) %>
-          <%= f.cfa_select(:vita_partner_id, t(".assign_to_label"), grouped_vita_partner_options) %>
-
+          <div class="hub-form__card card-small">
```
jenny-heath, line 11: for the future: with visual changes it's always nice to have a screenshot!

spec/features/hub/send_messages_spec.rb
```diff
@@ -11,8 +11,10 @@
           :intake,
           preferred_name: "Tobias",
           email_address: "tfunke@example.com",
+          email_notification_opt_in: "yes",
           phone_number: "+14155551212",
-          sms_phone_number: "+14155551212"
+          sms_phone_number: "+14155551212",
+          sms_notification_opt_in: "yes"
```
jenny-heath, line 17: huh, were we allowing people to email clients who hadn't opted in? maybe i'm not understanding these tests

spec/forms/hub/create_client_form_spec.rb
```diff
@@ -213,6 +213,62 @@
         end
       end
 
+      context "contact opt in -- no opted into methods" do
```
jenny-heath, line 216: since several forms (all children of `client_form`) changed as part of this PR, it would probably be good to test those forms in a similar way. it could be a bit redundant so maybe a shared example would work? this can be a follow-up task also

jenny-heath (approved):

alaroia-cfa (commented):
Need to review more later; wanted to hit Submit on this partial review first

app/forms/hub/client_form.rb
```diff
@@ -76,16 +77,26 @@ def dependent_validation_context
...
 
     def at_least_one_contact_method
       unless opted_in_email? || opted_in_sms?
-        errors.add(:communication_preference, I18n.t("forms.errors.need_one_communication_method"))
+        errors.add(:email_address, I18n.t("forms.errors.hub.communication_opt_in"))
+        errors.add(:sms_phone_number, I18n.t("forms.errors.hub.communication_opt_in"))
+      end
+    end
+
+    def contact_method_if_opted_in
```
jenny-heath, line 94: rename to `validate_notification_opt_ins` maybe

jenny-heath: `validate :opt_in_checkboxes`
`validate :opt_in_checkboxes_have_corresponding_information`

spec/features/hub/send_messages_spec.rb
jenny-heath: Turns out yes, and we can discuss that with program/product/design at some point.

jenny-heath:
here's what the new client form looks like: 
image

i updated the new ctc client form to also have that box around the top fields:
image

PR #1954, opened 2021-12-22
Simplify PhoneParser and rely on PhoneNumberHelper more in ERB

alaroia-cfa:
Based on review feedback in #1935


caitbot (approved):
nice! this looks really good!

PR #1957, opened 2021-12-22
Add a script to add a primary keys to all the tables in the anon2020 schema

tgrathwell:
They weren't created in the initial copy_anonymized_data script, but we
need the primary keys for Aptible's logical replication postgres upgrade
strategy to work.


tgrathwell:
I approve [this is Asheesh]

PR #1958, opened 2021-12-22
Update staging's database connection string to new pg13 version

tgrathwell:


tgrathwell:
approved by asheesh yay

PR #1959, opened 2021-12-23
New sla columns 180265589

jenny-heath:
image

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -45,7 +45,7 @@ module ClientSortable
   private
 
   def setup_sortable_client
-    @default_order = { "first_unanswered_incoming_interaction_at" => "asc" }
+    @default_order = { "last_outgoing_communication_at" => "asc" }
     delete_cookie if params[:clear]
     @sort_column = clients_sort_column
     @sort_order = clients_sort_order
--- a/app/helpers/time_helper.rb
+++ b/app/helpers/time_helper.rb
@@ -39,7 +39,7 @@ module TimeHelper
     datetime.strftime(format)
   end
 
-  def formatted_business_days_ago(time)
+  def business_days_ago(time)
     return unless time
 
     converted_time = Date.parse(time.utc.to_s)
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -263,7 +263,6 @@ class Client < ApplicationRecord
       UserNotification.create(notifiable_type: "SystemNote::DocumentHelp", notifiable_id: note.id, user: user)
     end
     tax_returns.each { |tax_return| tax_return.transition_to(:intake_needs_doc_help) }
-    flag!
   end
 
   def forward_message_to_intercom?
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -126,12 +126,6 @@ class Client < ApplicationRecord
       select(column_names).joins(:intake).merge(Intake.order(Hash[column, direction])).distinct
     else
       order = {column => direction}
-      if column == "first_unanswered_incoming_interaction_at" && direction == "desc"
-        order = "clients.first_unanswered_incoming_interaction_at DESC NULLS LAST"
-      elsif column == "last_outgoing_communication_at" && direction == "desc"
-        order = "clients.last_outgoing_communication_at DESC NULLS LAST"
-      end
-
       includes(:intake).order(order).distinct
     end
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -188,7 +188,8 @@ Rails.application.routes.draw do
         end
         resources :automated_messages, only: [:index]
         resources :clients do
-          get "/sla-breaches", to: "unattended_clients#index", on: :collection, as: :sla_breaches
+          # TODO: make SLA dashboard work again
+          # get "/sla-breaches", to: "unattended_clients#index", on: :collection, as: :sla_breaches
           get "/organization", to: "clients/organizations#edit", on: :member, as: :edit_organization
           patch "/organization", to: "clients/organizations#update", on: :member, as: :organization
           patch "/unlock", to: "clients#unlock", on: :member, as: :unlock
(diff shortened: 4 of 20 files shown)


app/helpers/time_helper.rb
```diff
@@ -39,7 +39,7 @@ def formatted_datetime(datetime, year: false)
     datetime.strftime(format)
   end
 
-  def formatted_business_days_ago(time)
+  def business_days_ago(time)
```
jenny-heath, line 42: this name change is a nitpick - the method just returns the number of days, which to me doesn't seem "formatted"

db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
```
jenny-heath, line 1: hm just noticed this - not sure how it happened but i'll change it back

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -356,10 +356,10 @@
           expect(tobias_2018_assignee).to have_text "Lindsay"
         end
 
-        context "when a client needs a response" do
+        context "when a client is flagged" do
           before { tobias.touch(:flagged_at) }
 
-          it "adds the needs response icon into the DOM" do
+          it "adds the flagged icon into the DOM" do
```
jenny-heath, line 362: this is just preemptive description-changing. in a future commit i'll completely decouple the flag from any SLA-related fields

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -736,6 +736,50 @@
           end
         end
       end
+
+      context "SLA columns" do
```
jenny-heath, line 740: this controller spec mostly checks the DOM, which normally i'd do in a feature spec but the one we have (`clients_searching_sorting_and_filtering_spec`) is kind of a mess imo so i found it easier to start somewhere else where i could keep it focused on the two table columns i was trying to test

db/schema.rb
caitbot: very odd

app/controllers/concerns/client_sortable.rb
```diff
@@ -45,7 +45,7 @@ def has_search_and_sort_params?
   private
 
   def setup_sortable_client
-    @default_order = { "first_unanswered_incoming_interaction_at" => "asc" }
+    @default_order = { "last_outgoing_communication_at" => "asc" }
```
caitbot, line 48: double check with design

this prioritizes last contact

app/models/client.rb
```diff
@@ -126,12 +126,6 @@ def self.sortable_intake_attributes
       select(column_names).joins(:intake).merge(Intake.order(Hash[column, direction])).distinct
     else
       order = {column => direction}
-      if column == "first_unanswered_incoming_interaction_at" && direction == "desc"
```
caitbot, line 129: looks like this is only used in client_sortable and thats exactly the kind of thing we want to change so this is good

config/routes.rb
```diff
@@ -188,7 +188,8 @@ def scoped_navigation_routes(context, navigation)
         end
         resources :automated_messages, only: [:index]
         resources :clients do
-          get "/sla-breaches", to: "unattended_clients#index", on: :collection, as: :sla_breaches
+          # TODO: make SLA dashboard work again
```
caitbot, line 191: I would ask design/product if we want to keep this around long-term

caitbot: Bring it to CGC? Ask Shannon?

caitbot: Or see if its fixable in like 10 minutes

spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -560,13 +560,13 @@
         end
 
         context "with no or bad params" do
-          let!(:first_id) { create :client, :with_return, vita_partner: organization, intake: create(:intake), first_unanswered_incoming_interaction_at: 1.day.ago }
-          let!(:second_id) { create :client, :with_return, vita_partner: organization, intake: create(:intake), first_unanswered_incoming_interaction_at: 2.days.ago }
+          let!(:first_id) { create :client, :with_return, vita_partner: organization, intake: create(:intake), last_outgoing_communication_at: 1.day.ago }
+          let!(:second_id) { create :client, :with_return, vita_partner: organization, intake: create(:intake), last_outgoing_communication_at: 2.days.ago }
 
           it "defaults to sorting by first_unanswered_incoming_interaction_at, asc by default" do
```
caitbot, line 566: defaults to `last_outgoing_communication_at` in test descrip.

spec/features/hub/clients_searching_sorting_and_filtering_spec.rb
```diff
@@ -182,33 +182,34 @@
...
+        expect(page.all('.client-row')[0]).to have_text(alan_intake_in_progress.preferred_name)
+        expect(page.all('.client-row')[0]).to have_text("Response")
+        expect(page.all('.client-row')[1]).to have_text(betty_intake_in_progress.preferred_name)
+        expect(page.all('.client-row')[1]).to have_text("Response")
+        expect(page.all('.client-row')[2]).to have_text(patty_prep_ready_for_call.preferred_name)
+        expect(page.all('.client-row')[2]).to have_text("Response")
+        expect(page.all('.client-row')[3]).to have_text(marty_ctc.preferred_name)
+        expect(page.all('.client-row')[3]).to have_text("Response")
+        expect(page.all('.client-row')[4]).to have_text(zach_prep_ready_for_call.preferred_name)
+        expect(page.all('.client-row')[4]).to have_text("Update")
```
caitbot, line 211: consider testing that this floats to the top when sorting order is reversed, at least manually

spec/models/client_spec.rb
```diff
@@ -178,8 +178,9 @@
     end
   end
 
+  # before deleting these tests, check how we want nulls to be sorted
```
caitbot, line 181: or replace with tests that check that null floats to the top

caitbot: or just delete them thats ok too

caitbot: because reversing order is normal and expected

PR #1961, opened 2021-12-23
Add job to delete GYR demo data

embarnard:


app/jobs/delete_gyr_demo_info_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    Client.includes(:intake).where(intake: { type: "Intake::GyrIntake" })
+          .where.not(id: exempted_client_ids).map(&:destroy)
+
+    # Delete all non-admin users
+    User.where.not(role_type: AdminRole::TYPE).map(&:destroy)
+
+    # Delete sites GYR sites with no clients
+    Site.where(national_overflow_location: false, processes_ctc: false)
+        .where.not(name: "GetCTC.org (Site)")
+        .where.missing(:clients).map(&:destroy)
```
bytheway875, line 13: ohh ive never seen this .missing -- does that find things where there are 0 of that association? that's magical.

bytheway875:
this looks good to me. I'm wondering why you chose a job over a rake task? Also, I think its somewhat concerning to have this in the codebase because if you accidentally run it on production, it can really mess things up. Should we add a line of code that exits if it is in production environment to avoid accidents?

app/jobs/delete_gyr_demo_info_job.rb
embarnard: right! its a new 6.1 thing I think ✨

embarnard:
> this looks good to me. I'm wondering why you chose a job over a rake task? Also, I think its somewhat concerning to have this in the codebase because if you accidentally run it on production, it can really mess things up. Should we add a line of code that exits if it is in production environment to avoid accidents?

True! I probably could have done a rake test instead. Great idea thought about adding a production guard! I added a return statement if the environment is production

PR #1962, opened 2021-12-23
Remove anonymized csv extracts code

bytheway875:
We did not use this last tax year, and the purpose of this code path is redundant with what we do in a more sophisticated way now (using metabase for data science and reporting)

After this is merged, I will do a follow up PR to drop the tables.


alaroia-cfa (approved):
It's a Christmas miracle.

PR #1963, opened 2021-12-23
Clear tax returns' assigned user before deleting user for the DeleteGyrDemoInfoJob

embarnard:


spec/jobs/delete_gyr_demo_info_job_spec.rb
```diff
@@ -74,5 +74,18 @@
...
+    context "users with assigned tax returns of exempted clients" do
+      let!(:tax_return_2) { create :tax_return, :file_ready_to_file, assigned_user: site_coordinator_user, client: gyr_client, year: 2020 }
+      let!(:tax_return_3) { create :tax_return, :file_ready_to_file, assigned_user: greeter_user, year: 2021 }
+
+      it "deletes non-admin users" do
+        described_class.perform_now([gyr_client.id])
+
+        expect(admin_user.reload).to be_present
+        expect { site_coordinator_user.reload }.to raise_error(ActiveRecord::RecordNotFound)
+        expect { greeter_user.reload }.to raise_error(ActiveRecord::RecordNotFound)
```
alaroia-cfa, line 87: It would be nice to validate that the `tax_return` still exists and has a `nil` `assigned_user`.

alaroia-cfa (approved):
lgtm; 1 comment; might be an easy thing to fix before merging; ship either way

spec/jobs/delete_gyr_demo_info_job_spec.rb
embarnard: good point!

PR #1966, opened 2021-12-23
Use Postgres 13 in CI [#180014967]

alaroia-cfa:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ executors:
         environment:
           PGHOST: 127.0.0.1
           PGUSER: root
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
We upgraded production to use Postgis 13.4 (from 11.x) so this is now in Circle.

bytheway875:
i wonder if in the future it makes sense to try to bump circle before we do the aptible upgrade so that we can try to catch any test failures?

bytheway875 (approved):

alaroia-cfa:
> i wonder if in the future it makes sense to try to bump circle before we do the aptible upgrade so that we can try to catch any test failures?

This is obviously a very good idea!

PR #1970, opened 2022-01-06
New GYR triage flow

tgrathwell:


app/forms/intake_form.rb
```diff
@@ -0,0 +1,17 @@
+class IntakeForm < Form
```
tgrathwell, line 1: TriageForm??

tgrathwell: yay fixed locally

PR #1977, opened 2022-01-12
Add job to send message about saving CTC letter

embarnard:


config/locales/en.yml
```diff
@@ -1279,6 +1279,20 @@ en:
...
         It looks like you stepped away before finishing our application. Our preparers at <<Client.AssignedOrganization>> look forward to working with you! To complete our form, simply login here: <<Client.LoginLink>>
         Thanks!
+    save_ctc_letter:
+      email:
+        body: |
+          Hello,
+
+          You will receive what’s called a 6419 Letter from the IRS in January 2022. This letter will help you and your tax preparer correctly claim your Child Tax Credit for you and your family. The letter will be have 6419 printed on it and will include information about the Child Tax Credit payments that you received in 2021.
+
+          Please save this letter for when you file your taxes. Whether you file with GetYourRefund or any other service, this will help you collect any missing payments and receive any refund due without delay.
```
bytheway875, line 1289: Do we want "GetYourRefund" to be swapped out with GetCTC if that's what they used before or no?

app/jobs/send_client_save_ctc_letter_message_job.rb
```diff
@@ -0,0 +1,15 @@
+class SendClientSaveCtcLetterMessageJob < ApplicationJob
+  def perform(client)
+    return if client.intake.email_notification_opt_in != "yes" && client.intake.sms_notification_opt_in != "yes"
+
+    service_name = client.intake.is_ctc? ? "GetCTC" : "GetYourRefund"
```
bytheway875, line 5: This may be overkill since we only use the service_name in here, but it might make sense to use MultiTenetService to achieve rendering the appropriate name based on the service if we're doing things like this in the future (would have been especially helpful if this email had a link or something)

bytheway875:
I don't see code that triggers sending the messages. Are you planning on writing another PR for that or executing in the console directly?

config/locales/en.yml
embarnard: Hmm the copy is the same for that part on both of the copies in the ticket description

embarnard:
> I don't see code that triggers sending the messages. Are you planning on writing another PR for that or executing in the console directly?

Yeah! I was planning to run it on the console directly --> I think I would just pass all Clients through in batches, do you think it would be better just to have the job run the whole thing instead?

app/jobs/send_client_save_ctc_letter_message_job.rb
embarnard: Oh good point! Making the change to include `MultiTenantService.new(:ctc).service_name` for the name instead of hard coding it. The ticket didn't specify to make it a link but I can run it by Yvonne

app/jobs/send_client_save_ctc_letter_message_job.rb
```diff
@@ -0,0 +1,33 @@
+class SendClientSaveCtcLetterMessageJob < ApplicationJob
+  def perform(number_of_clients: 0)
+    total_sent = 0
+
+    puts "\n==Sending messages to #{number_of_clients} clients=="
+    Archived::Intake2021.all.limit(number_of_clients).find_each do |intake|
```
alaroia-cfa, line 6: In theory the `.all` isn't necessary -- the `.limit` is delegated to `.all` by the model, as it happens. Not a blocker.

app/jobs/send_client_save_ctc_letter_message_job.rb
```diff
@@ -0,0 +1,33 @@
...
+      message_tracker = MessageTracker.new(client: intake.client, message: AutomatedMessage::SaveCtcLetter)
+      next if message_tracker.already_sent?
+
+      messages = send_message(intake)
+      puts "Client ##{intake.client.id} sent messages: #{messages}" if messages.present?
+      total_sent += 1 if messages.present?
```
alaroia-cfa, line 12: Could be something like `+= messages.count`. Not a blocker.

app/models/client.rb
```diff
@@ -255,6 +255,10 @@ def preferred_language
     intake.locale || intake.preferred_interview_language
   end
 
+  def intake_or_archived_intake
```
alaroia-cfa, line 258: Consider inlining this to avoid creating a new thing that others may depend on but perhaps mistakenly. Might be good to check in with e.g. Shannon to see if we can rely entirely on ClientMessagingService.contact_methods and move this logic into that method.

app/services/client_messaging_service.rb
```diff
@@ -2,23 +2,24 @@ class ClientMessagingService
...
     # only sends email if the client can receive emails
     def send_email(client:, user:, body:, attachment: nil, subject: nil, locale: nil, tax_return: nil, to: nil)
-      return unless client.intake.email_notification_opt_in_yes?
+      intake = client.intake_or_archived_intake
+      return unless intake.email_notification_opt_in_yes?
 
-      if client.intake.email_notification_opt_in_yes? && !client.email_address.present?
-        DatadogApi.increment('clients.missing_email_for_email_opt_in')
+      if intake.email_notification_opt_in_yes? && !intake.email_address.present?
+        DatadogApi.increment('clients.missing_email_for_email_opt_in') unless client.intake.nil?
```
alaroia-cfa, line 9: I encourage you to remove the new `unless`

app/services/client_messaging_service.rb
```diff
@@ -44,17 +45,18 @@ def send_system_email(client:, body:, subject:, tax_return: nil, locale: nil, to
...
     # only sends text message if client can receive texts
     def send_text_message(client:, user:, body:, tax_return: nil, locale: nil, to: nil)
-      return unless client.intake.sms_notification_opt_in_yes?
+      intake = client.intake_or_archived_intake
+      return unless intake.sms_notification_opt_in_yes?
 
-      if client.intake.sms_notification_opt_in_yes? && !client.sms_phone_number.present?
-        DatadogApi.increment('clients.missing_sms_phone_number_for_sms_opt_in')
+      if intake.sms_notification_opt_in_yes? && !intake.sms_phone_number.present?
+        DatadogApi.increment('clients.missing_sms_phone_number_for_sms_opt_in') unless client.intake.nil?
```
alaroia-cfa, line 52: I encourage you to remove the new `unless`

app/services/replacement_parameters_service.rb
```diff
@@ -30,10 +30,10 @@ def process_replacements_hash(replacements_hash)
 
   def replacements
     {
-        "Client.PreferredName": client&.preferred_name&.titleize,
+        "Client.PreferredName": @intake&.preferred_name&.titleize,
         "Preparer.FirstName": preparer_first_name,
         "Documents.List": documents_list,
-        "Client.LoginLink": intake.is_ctc? ? new_ctc_portal_client_login_url(locale: @locale, host: Rails.configuration.ctc_url) : new_portal_client_login_url(locale: @locale),
+        "Client.LoginLink": @intake.is_ctc? ? new_ctc_portal_client_login_url(locale: @locale, host: Rails.configuration.ctc_url) : new_portal_client_login_url(locale: @locale),
```
alaroia-cfa, line 36: Consider removing the `@` from `@intake` since the `attr_accessor` lets you. Not a blocker.

config/locales/es.yml
```diff
@@ -1240,6 +1240,20 @@ es:
         Hola <<Client.PreferredName>>,
         Parece que no a terminado nuestra aplicación. ¡Nuestros preparadores en <<Client.AssignedOrganization>> esperamos trabajar con usted! Para completar nuestro formulario, simplemente inicie sesión aquí: <<Client.LoginLink>>
         ¡Gracias!
+    save_ctc_letter:
```
alaroia-cfa, line 1243: https://docs.google.com/document/d/12q4jh324MMqepHCS7gabdXyZzfUKV7JzpYbZsQvLKg4/edit# "For copy that must have human translations first"

spec/factories/archived/intake/ctc_intake_2021.rb
```diff
@@ -3,6 +3,7 @@
     preferred_name { "Cherry" }
     type { 'Intake::CtcIntake' }
     primary_first_name { "Cher" }
+    association :client, factory: :ctc_client
```
alaroia-cfa, line 6: continue reviewing from here

alaroia-cfa (commented):
No blockers yet. Looks good but see comments (plus I need to finish reviewing)

PR #1978, opened 2022-01-12
Route returning clients 180665573

bytheway875:


app/models/archived/intake_2021.rb
```diff
@@ -298,8 +298,6 @@ def self.searchable_fields
         self.email_domain = email_address.split('@').last.downcase
         self.canonical_email_address = compute_canonical_email_address
       end
-      self.primary_last_four_ssn = primary_ssn&.last(4) if encrypted_primary_ssn_changed?
-      self.spouse_last_four_ssn = spouse_ssn&.last(4) if encrypted_spouse_ssn_changed?
```
bytheway875, line 302: This was added for behavior that is specific to this tax season and moving forward, but we hadn't archived last year's intake model yet. This logic caused some issues when trying to set up archived intakes in the tests without a full ssn, and since its not relevant to how last year's intakes worked, im removing it.

app/controllers/questions/at_capacity_controller.rb
```diff
@@ -3,6 +3,12 @@ class AtCapacityController < QuestionsController
     include AnonymousIntakeConcern
     layout "application"
 
+    def self.show?(intake)
+      return true if intake.client.routing_method_at_capacity?
+
+      intake.vita_partner&.at_capacity?
```
bytheway875, line 9: the old at capacity page routing was a bit weird, because it seemed to let through the people who did not have a vita partner assigned, but we actually will not assign a vita partner when we're at capacity....

I'm considering moving the second line intake.vita_partner.at_capacity? from here as well, because we shouldnt assign them to that vita partner on the previous page until there IS capacity. Does that seem right?

app/controllers/questions/at_capacity_controller.rb
```diff
@@ -11,13 +17,5 @@ def update
       current_intake.update(continued_at_capacity: true)
```
bytheway875, line 17: i dont really think that this is relevant code, because we no longer invite people to continue once we're at capacity -- we redirect them to tax slayer.

app/controllers/questions/optional_consent_controller.rb
```diff
@@ -1,6 +1,5 @@
 module Questions
   class OptionalConsentController < QuestionsController
```
bytheway875, line 2: Removed requirememnt for being logged in here because we don't log in AtCapacity people (we dont want them to accidentally be able to continue through the process) but we do have them do the optional consent pages.

spec/models/intake/gyr_intake_spec.rb
```diff
@@ -276,4 +276,22 @@
...
+      let(:intake) { create :intake, primary_birth_date: Date.new(1929, 3, 22), primary_first_name: "Seth", primary_last_name: "Strawberry", primary_ssn: "12341234" }
+
+      it "returns the Archived::Intake2021 object" do
+        expect(intake.probable_previous_year_intake).to eq archived_intake
+      end
+    end
+
+    context "when there is not a matching intake" do
+      it "returns nil" do
+
```
caitbot, line 293: oops missing expectations

caitbot (approved):
Sweet, this looks really good.

I left a few comments. All non-blocking except the ones on specs.

app/controllers/questions/at_capacity_controller.rb
caitbot: That sounds right to me. I was also really surprised by this. Might be worth checking with product why we did this exactly and confirm that we do not want to let them through. My thought was maybe we planned on messaging them or assigning them when there was capacity?

caitbot: Ok based on other comments it looks like the decision is just to send them to taxslayer. Got it.

app/controllers/questions/optional_consent_controller.rb
bytheway875: It was kind of an arbitrary decision to show the at capacity people the second optional consent page and i think you're right -- it doesnt really make sense to have them get this far if we can't serve them.

I THINK the reason it ended up working this way is because we take in birthday and SSN on the legal consent page, and we cant check to see if they're duplicated until we have that information.

I could move the at capacity page before this optional consent, but that also felt odd to me. 

I feel like this whole flow could use some more thought because essentially whats happening here is that these people are consenting to service so they will show up in the hub as not assigned to any vita partner and we're telling the clients we can't help them (show them the DIY link for tax slayer), but I'm worried it will be confusing to see them in the hub. Maybe we shouldnt mark them as consented to service if we cant route them either
but i also kind of hate that we're making people do like 10 things (validate phone number or email, read these consent forms) when we wont be doing anything with their info.

app/models/archived/intake_2021.rb
```diff
@@ -280,7 +280,7 @@ def self.searchable_fields
     has_many :dependents, -> { order(created_at: :asc) }, inverse_of: :intake, dependent: :destroy, class_name: 'Archived::Dependent2021', foreign_key: 'archived_intakes_2021_id'
     belongs_to :client, optional: true
     has_many :tax_returns, through: :client
-    belongs_to :vita_partner, optional: true
+    has_one :vita_partner, through: :client
```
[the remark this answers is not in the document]
bytheway875: We used to dual write to intake vita partner when we implemented this last tax season. However, anything that was previously relying on vita partner from intake is gone and we use vita partner off of client in the hub everywhere.

app/models/vita_partner.rb
```diff
@@ -68,4 +68,9 @@ def site?
   def organization?
     type == Organization::TYPE
   end
+
+  # A VitaPartner is considered active if they have a StateRoutingFraction that is greater than 0 for any state.
+  def active?
+    StateRoutingFraction.where(vita_partner_id: id).where(StateRoutingFraction.arel_table[:routing_fraction].gt(0.0)).exists?
```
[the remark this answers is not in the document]
bytheway875: it looks like this: ```SELECT 1 AS one FROM "state_routing_fractions" WHERE "state_routing_fractions"."vita_partner_id" = $1 AND "state_routing_fractions"."routing_fraction" > 0.0 LIMIT $2  [["vita_partner_id", 1], ["LIMIT", 1]]```

app/services/partner_routing_service.rb
```diff
@@ -25,6 +29,17 @@ def determine_partner
 
   private
 
+  def previous_year_partner
+    return false unless @intake
```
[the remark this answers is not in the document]
bytheway875: all of the other sub-methods in this file use `return false`, at the top, but it is true that its not consistent with the later return value. I guess they should all be updated.

PR #1979, opened 2022-01-14
Remove /questions/file-with-help from intake flow

jenny-heath:


app/controllers/questions/triage_controller.rb
```diff
@@ -36,7 +36,7 @@ def form_attributes
     end
 
     def redirect_if_matching_source_param
-      redirect_to file_with_help_questions_path if SourceParameter.find_vita_partner_by_code(session[:source]).present?
+      redirect_to backtaxes_questions_path if SourceParameter.find_vita_partner_by_code(session[:source]).present?
```
caitbot, line 39: Is this just a placeholder for now? I assume so.

caitbot (approved):
This looks good to me. It’s a little funny that we are replacing the links to it with backtaxes, as we will be removing that too. 

Depending on how product feels (about this intermediate state), you could consider bundling the removal of backtaxes in as well, since this is rather small on its own.

I don’t feel terribly strongly either way.

PR #1995, opened 2022-01-20
Fix national routing capabilities

bytheway875:
instead of turning off routing to national sites, the code for national routing was commented out at the end of last season.

### DIFF
--- a/spec/features/web_intake/routing_spec.rb
+++ b/spec/features/web_intake/routing_spec.rb
@@ -183,9 +183,51 @@ feature "Intake Routing Spec", :flow_explorer_screenshot, :active_job do
 
     expect(page.html).to have_text "Our team at Hogwarts is here to help!"
   end
+  context "at capacity but overflow site exists" do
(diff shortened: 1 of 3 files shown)


spec/features/web_intake/routing_spec.rb
```diff
@@ -183,9 +183,51 @@ def fill_out_notification_preferences
...
+      fill_in "Do you have any time preferences for your interview phone call?", with: "During school hours"
+      click_on "Continue"
+
+      fill_out_notification_preferences
+
+      expect(page.html).to have_text "Default Organization is here to help"
+    end
+  end
 
   context "vita partner is at capacity" do
```
caitbot, line 229: is this a context for when the national overflow partner is at capacity? it isn't entirely clear to me

caitbot: I think I am confused because the national overflow above is named the same, but here it is not national overflow

caitbot (approved):
looks great, thanks for adding this back so quickly

PR #1997, opened 2022-01-21
Stimulus and ctc question 180829676

bytheway875:
Screen Shot 2022-01-20 at 3 03 52 PM

### DIFF
--- a/app/forms/arp_payments_form.rb
+++ b/app/forms/arp_payments_form.rb
@@ -40,7 +40,7 @@ class ArpPaymentsForm < QuestionsForm
   end
 
   def advance_ctc_amount_present_or_unsure
-    if received_advance_ctc_payment != "unsure" && advance_ctc_amount_received.to_i.blank?
+    if received_advance_ctc_payment != "unsure" && advance_ctc_amount_received.blank?
       errors.add(:received_advance_ctc_payment, I18n.t("views.questions.arp_payments.errors.advance_child_tax_credit"))
     end
   end
--- a/app/forms/arp_payments_form.rb
+++ b/app/forms/arp_payments_form.rb
@@ -20,7 +20,7 @@ class ArpPaymentsForm < QuestionsForm
       end
     end
     unless received_advance_ctc_payment == "unsure"
-      intake.received_advance_ctc_payment = advance_ctc_amount_received.zero? ? "no" : "yes"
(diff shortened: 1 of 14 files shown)


app/forms/arp_payments_form.rb
```diff
@@ -0,0 +1,47 @@
+class ArpPaymentsForm < QuestionsForm
+  set_attributes_for :intake,
+                     :received_stimulus_payment,
+                     :eip1_amount_received,
```
tgrathwell, line 4: should add a numericality validator like on the other AGI / stimulus payment forms, e.g.
```
validates :primary_prior_year_agi_amount, gyr_numericality: true, if: :not_blank?
```

`gyr_numericality` mostly ensures the value is less than postgres max of `2**31 - 1` so jokers that come in an hammer out big numbers on their keyboard don't cause a 500 exception

app/forms/arp_payments_form.rb
```diff
@@ -0,0 +1,47 @@
...
+                     :eip3_amount_received,
+                     :advance_ctc_amount_received,
+                     :received_advance_ctc_payment
+  validate :eip_amounts_present_or_unsure
+  validate :advance_ctc_amount_present_or_unsure
+
+  def save
+    intake.assign_attributes(attributes_for(:intake))
+
+    unless received_stimulus_payment == "unsure"
```
tgrathwell, line 15: just noting that as written this will still accept/persist any numerical values the client enters even when they also check `unsure`

we could make it a truly exclusive thing (provide numbers *or* leave them blank and click 'unsure') but I guess that's up to whatever design wants

tgrathwell (approved):

tgrathwell:
approved but with a comment about the `gyr_numericality` validator that could save us exception noise in the future

app/forms/arp_payments_form.rb
bytheway875: its fine that they add numbers and check unsure -- they provide a value, but they're not quite sure. This indicates to the tax preparer that they may need to reconfirm those numbers with the client on the intake call.

app/forms/arp_payments_form.rb
bytheway875: good call -- thanks.

bytheway875:
added the numericalityvalidation

PR #1998, opened 2022-01-21
Triage: client with any missing tax filing years who meets other criteria goes directly to full service

tgrathwell:
previously they went to the "income types" page (which asks if they have farm/rental
income) but the outcome was the same no matter what they chose, so we'll take them
directly to full service I guess.

### DIFF
--- a/spec/support/helpers/feature_helpers.rb
+++ b/spec/support/helpers/feature_helpers.rb
@@ -34,19 +34,39 @@ module FeatureHelpers
     end
   end
 
-  def answer_gyr_triage_questions(
-    screenshot_method: nil,
-    income_level: "hh_1_to_25100_html",
-    id_type: "have_paperwork",
-    doc_type: "all_copies_html",
-    filed_past_years: [
-      TaxReturn.current_tax_year - 3,
-      TaxReturn.current_tax_year - 2,
-      TaxReturn.current_tax_year - 1,
-    ],
-    assistance_options: ['chat'],
-    income_type_options: ['none_of_the_above']
-  )
+  class TriageChoices
+    def initialize(choices)
+      @choices = choices
+    end
+
+    private
+
+    def method_missing(symbol, *args)
+      if @choices[symbol]
+        @choices[symbol]
+      else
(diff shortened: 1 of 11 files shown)


spec/support/helpers/feature_helpers.rb
```diff
@@ -34,27 +34,47 @@ def seen_pages
...
-    income_type_options: ['none_of_the_above']
-  )
+  class TriageChoices
+    def initialize(choices)
+      @choices = choices
+    end
+
+    private
+
+    def method_missing(symbol, *args)
```
alaroia-cfa, line 44: If you're gonna support `method_missing`, why switch to `"#{choices.doc_type"`? Mostly I'm confused; not a blocker.

alaroia-cfa (approved):
Give the people what they want.

🚢 

One non-blocking question about the refactor.

spec/support/helpers/feature_helpers.rb
tgrathwell: my goal was to convert the simple list of keyword params for `answer_gyr_triage_questions` into an object such that every test in `features/triage_spec.rb` is forced to provide every single piece of information that will be used.

prior to this, any missing kwargs would revert to the default argument, which is (arguably) confusing because you have to look in two places to find the exact options that will be used

but the `method_missing` thing isn't the only way to do it. now it has the new problem that you could initialize `TriageChoices` with an option that doesn't exist and it wouldn't complain, unlike the kwargs. I mostly also wanted to keep it concise.

tgrathwell: I get it now [hi this is asheesh] `choices = TriageOptions.new(...)`

alaroia-cfa:
There's a fun (?) i18n HTML error

tgrathwell:
> There's a fun (?) i18n HTML error

it turned out to be not very fun!! https://github.com/codeforamerica/vita-min/pull/1998/files#diff-51474a8cda735bb507e5736cdba89bde6ed714d8cd2c6ddd3bd5494c7d8de107R42

PR #2001, opened 2022-01-21
Send weekly client-not-ready emails during 2022 season and minimize repeated work for future season open/close [#180995822]

alaroia-cfa:

PR #2002, opened 2022-01-21
Ensure soft-launch and full-launch are on different days on demo [#180995822]

alaroia-cfa:
In #2001 I had configured demo's soft launch time & full launch time to be the same. This made it impossible to do acceptance testing for soft launch (unique URLs only). This fixes that.


alaroia-cfa:
@tgrathwell I would appreciate you can review this, even though I have already merged it.

tgrathwell:
it apparently doesn't let me see the 'approve' radio anymore but i approve

PR #2008, opened 2022-01-24
Align tax returns with backtaxes years and ensure we can try to route clients who originally got message of at capacity

bytheway875:


app/controllers/questions/consent_controller.rb
```diff
@@ -14,29 +14,31 @@ def form_params
     def after_update_success
       GenerateRequiredConsentPdfJob.perform_later(current_intake)
 
-      # client has not yet been routed
-      unless current_intake.client.routing_method.present?
+      # client has not yet been routed, or was previousl determined to have been at capacity
```
embarnard, line 17: *previously

embarnard (approved):
LGTU! Reviewed by TRavis and Em :)

app/controllers/questions/consent_controller.rb
bytheway875: my y key intermittantly stops working, but worked every time here in this ephemeral comment where it wouldnt matter if it was broken. lol

PR #2015, opened 2022-01-25
Show already started link on homepage

embarnard:


caitbot (approved):
This works!

I found myself thinking for a while if there is a way to do this that would be more idiomatic/allow an easier reversal when the season closes again.

There is `Rails.configuration.start_of_unique_links_only_intake` that is used in `open_for_intake?` and looks for the presence of `cookies[:used_unique_link]` and could be used to make a similar method called `open_for_soft_launch?` which uses the same Rails config variable, but doesn't look for the presence of the cookie.

This isn't necessary, but if we want the sign in link to disappear at the end of the season, it is preferable and worth checking with product.

If we need to just get something out now because this is blocking people then ship it!

But I am surprised that `open_for_intake?` isn't returning true for visitors that came to the site with a unique url and were thus able to start an intake. In that case, their cookie should be set and so `open_for_intake?` should be evaluated based on the soft launch date. But 🤷🏻‍♀️. Might be worth investigating a bit more.

app/controllers/portal/client_logins_controller.rb
```diff
@@ -115,7 +115,7 @@ def redirect_to_portal_if_client_authenticated
     end
 
     def redirect_gyr_in_offseason
-      redirect_to root_path if Routes::GyrDomain.new.matches?(request) && !open_for_intake?
+      redirect_to root_path if Routes::GyrDomain.new.matches?(request) && !open_for_soft_launch?
```
caitbot, line 118: This is the spot that I think it might be weird once the season really opens because what we really mean is "soft launch AND normal season intake", hence the recommendation to add a parameter to `open_for_intake`

caitbot (approved):
Rather than duplicating, I suggest we pass a parameter into the old method to specify that we do not require a cookie for soft launch in some cases. 

Non-blocking, since I know we likely need to get this out and this also works!

Thanks for working on this! (:

PR #2019, opened 2022-01-26
create verification attempts table and model

bytheway875:

### DIFF
--- a/app/models/verification_attempt.rb
+++ b/app/models/verification_attempt.rb
@@ -1,0 +1,7 @@
+class VerificationAttempt < ApplicationRecord
+  belongs_to :client
+
+  # I'm guessing we will need to link somehow to Active_storage to define what selfie and photo_id are?
+  has_one_attached :selfie
+  has_one_attached :photo_id
+end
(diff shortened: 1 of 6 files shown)


app/models/verification_attempt.rb
```diff
@@ -0,0 +1,7 @@
+class VerificationAttempt < ApplicationRecord
+  belongs_to :client
+
+  # I'm guessing we will need to link somehow to Active_storage to define what selfie and photo_id are?
+  has_one_attached :selfie
+  has_one_attached :photo_id
```
bytheway875, line 6: we'll def want to update this to call it "photo_identification" -- photo_id will cause confusion, thinking its an id reference to a table called photos /model named Photo 😬

db/migrate/20220125232709_create_verification_attempts.rb
```diff
@@ -0,0 +1,12 @@
+class CreateVerificationAttempts < ActiveRecord::Migration[6.1]
+  def change
+    create_table :verification_attempts do |t|
+
```
bytheway875, line 4: your note is exactly right -- just make sure you move lines 11 and 12 inside of the create_table block, and when you run `rails db:migrate`, it will add those references to the schema for the verification_attempts table.

(if you've already run `rails db:migrate`, you can run `rails db:rollback` to reverse the migration and then `rails db:migrate again once you've added the new fields)

PR #2021, opened 2022-01-26
Triage: implement income-level-only triage

tgrathwell:
i think the commit messages basically sum up everything that is done here. the one less than ideal part of this is that i left a lot of skipped tests in `triage_spec` and commented code in `triage_result_service` because it seemed like more effort than it was worth to delete it and then immediately recreate it in another form. if we care a lot about having a clean commit history (or if there's another reason it's good to delete it), i can do that.

image


app/views/questions/triage_income_types/edit.html.erb
```diff
@@ -9,13 +9,13 @@
...
+      <%= f.cfa_checkbox(:income_type_farm, t('.income_types.farm'), options: { checked_value: "yes", unchecked_value: "no" }) %>
       <%= f.cfa_checkbox(:none_of_the_above, t("general.none_of_the_above"), options: { id: "none__checkbox", checked_value: "yes", unchecked_value: "no" }) %>
     </div>
-  <% end %>
 
-  <div class="spacing-above-60">
-    <%= link_to t('general.continue'), next_path, class: "button button--primary button--wide text--centered" %>
-  </div>
+    <div class="spacing-above-60">
+      <%= f.submit t('general.continue'), class: "button button--primary button--wide text--centered" %>
```
jenny-heath, line 18: @tgrathwell dagnabit sorry we missed this one when you and I were working on this story - Asheesh

alaroia-cfa (approved):
This implements income-only triage! https://www.pivotaltracker.com/story/show/181074077 I'll merge and send it to demo.

PR #2022, opened 2022-01-26
Update triage flow copy

caitbot:


spec/forms/triage_assistance_form_spec.rb
```diff
@@ -18,8 +18,7 @@
       let(:params) do
         {
           assistance_in_person: "no",
-          assistance_chat: "yes",
```
alaroia-cfa, line 21: This removes the one `yes` that we were saving into the model. I'd rather save at least 1 `yes`. Not a blocker.

spec/support/helpers/feature_helpers.rb
```diff
@@ -61,7 +61,7 @@ def answer_gyr_triage_questions(screenshot_method: nil, **options)
           TaxReturn.current_tax_year - 2,
           TaxReturn.current_tax_year - 1,
         ],
-        assistance_options: ['chat'],
```
alaroia-cfa, line 64: It's simpler to select only one option here, e.g. `phone_review_english` (which is probably the common case). Not a blocker.

spec/features/web_intake/triage_spec.rb
```diff
@@ -138,7 +138,7 @@
         TaxReturn.current_tax_year - 1,
       ],
       income_type_options: ['none_of_the_above'],
-      assistance_options: ['chat'],
```
alaroia-cfa, line 141: It's simpler to select only one option here, e.g. `phone_review_english` (which is probably the common case). Not a blocker.

alaroia-cfa (approved):
lgtm

This is going to create a conflict with the other work Jenny and I were trying to do on triage cases, so we'd better merge this soon. :)

PR #2024, opened 2022-01-26
Verification attempts controller

mquadri23:
- created a verification controller under the hub folder
- created a show.html.erb for the verification controller 
- created an index.html.erb
- modified the routes.rb file to create a route for verification_attempts


bytheway875:
Awesome!! This is looking so good. We can walk through and answer any questions you have tomorrow morning, or feel free to work with Em on it.

The next step is to start adding data to the view so that we can view individual verification attempts and a list of all verification attempts!

PR #2025, opened 2022-01-27
Hub: fix crash on /messages page for some archived clients

tgrathwell:


bytheway875 (approved):

app/models/incoming_email.rb
```diff
@@ -42,8 +42,4 @@ def body
   def datetime
     received_at
   end
-
```
alaroia-cfa, line 45: This method removal to be causing a test failure. Consider restoring the method but giving it a null guard on `client.intake`.

PR #2029, opened 2022-01-28
Only display the years the client has not filed on backtaxes

caitbot:


app/controllers/questions/backtaxes_controller.rb
```diff
@@ -2,12 +2,28 @@ module Questions
...
     layout "intake"
 
     def current_intake
       Intake::GyrIntake.new
     end
 
+    def edit
+      # skip this question if they answered that they only haven't filed for the current year
+      # and save on the intake that they need help for the current year
+      if current_triage.present? && only_current_tax_year_not_filed(current_triage)
```
bytheway875, line 15: i looked at this a couple of times and kept thinking of alternate ways to approach it, and each time ended up coming back to understanding why you selected this approach.

app/forms/backtaxes_form.rb
```diff
@@ -3,7 +3,15 @@ class BacktaxesForm < QuestionsForm
   validate :at_least_one_year
 
   def save
-    @intake.update(attributes_for(:intake).merge(client: Client.create!))
+    # some of the needs_help fields are not shown, based on what years the client has filed
+    # but the intake does not allow nil values
+    needs_help_params = {
+      needs_help_2018: needs_help_2018.nil? ? "unfilled" : needs_help_2018,
```
bytheway875, line 9: trying to think of whether there are any implications of leaving these as unfilled further on down the line, but this actually seems to be the most respectful to the actions the client has taken in terms of altering data.

app/controllers/application_controller.rb
```diff
@@ -248,6 +248,7 @@ def show_offseason_banner?
   end
 
   def open_for_intake?
+    return true
```
bytheway875, line 251: I'm guessing this is why the tests are failing at the moment.

bytheway875 (changes requested):
need to remove the override in open_for_intake?, but other than that looks great!

app/controllers/questions/backtaxes_controller.rb
caitbot: lol this is validating, thank you. yeah, I spent a long time noodling on it and really couldn't come up with anything better..... just ended right back here, not the prettiest thing in the world but it works!

app/controllers/application_controller.rb
caitbot: hahahah I completely forgot I put that in, good catch

PR #2030, opened 2022-01-28
Fix SendOutgoingTextMessageJob to not crash for an Invalid Phone Number error

tgrathwell:
since it is no longer raising an error, the calling code needs to be aware
that sometimes send_text_message will not return a message

### DIFF
--- a/spec/jobs/send_outgoing_text_message_job_spec.rb
+++ b/spec/jobs/send_outgoing_text_message_job_spec.rb
@@ -1,6 +1,8 @@
 require "rails_helper"
 
 RSpec.describe SendOutgoingTextMessageJob, type: :job do
+  include MockTwilio
+
   describe "#perform" do
     let(:user) { create(:user) }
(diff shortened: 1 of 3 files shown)


spec/jobs/send_outgoing_text_message_job_spec.rb
```diff
@@ -31,18 +27,30 @@
...
-        }.to raise_error(Twilio::REST::RestError)
-        outgoing_text_message.reload
-        expect(outgoing_text_message.twilio_status).to eq "twilio_error"
+          outgoing_text_message.reload
+          expect(outgoing_text_message.twilio_status).to eq "twilio_error"
+        end
+      end
+
+      context "for other errors" do
+        let(:status_code) { 8675309 }
```
alaroia-cfa, line 45: 🎵

alaroia-cfa (approved):
🚢

PR #2033, opened 2022-01-28
Update triage to send clients with lower income to choice between gyr and express

jenny-heath:
the feature spec here is def overkill, but i don't feel like redoing it as part of this commit so will pare it down later


caitbot (approved):
hahaha agree about the feature spec but this looks great otherwise, thanks for getting this change in so quickly!

PR #2075, opened 2022-01-31
Rename cmd to web in Procfile

alaroia-cfa:
All Aptible services called `cmd` are now scaled to zero, so I think this is peaceful.


tgrathwell (approved):
I looked in the Aptible dashboard and I didn't see any endpoints configured for `cmd` anymore so this seems like it is the way to go

thanks for finishing what I was unable to get finished!!

alaroia-cfa:
> thanks for finishing what I was unable to get finished!!

:D team work makes the bleem! work

or however the saying goes!!!

PR #2077, opened 2022-01-31
[#181077447] Update GYR.org/DIY copy

alaroia-cfa:
## Before

image

## After

image


bytheway875:
@alaroia-cfa This is really nitpicky, but I think the margin between the paragraph and the button has increased on this implementation from the original.

bytheway875 (approved):

PR #2087, opened 2022-01-31
Hub fix crash on security page for clients with archived CTC intakes

tgrathwell:
`duplicate_bank_client_ids` should now work, more or less, though it will
only show dupes in the archived table if you're looking at an archived
client, or dupes in the `intakes` table if you're looking at a current client.

This required implementing `.sti_name` in the two leaf archived intake classes,
because otherwise if you do `archived_bank_account.intake` it will not be able
to find anything because the inverse association is configured with a `class_name`
of `Archived::CtcIntake2021` which will make the database query look for a
`type: 'Archived::CtcIntake2021'` when the values that actually live in the DB
are `type: 'Intake::CtcIntake'`

is this the last we've seen of archiving + STI problems? probably not!!


app/controllers/hub/security_controller.rb
```diff
@@ -17,7 +17,7 @@ def show
 
     def duplicate_bank_client_ids
       return [] unless @client.intake.bank_account.present?
-      Intake.where(id: @client.intake.bank_account.duplicates.pluck(:intake_id)).pluck(:client_id)
+      @client.intake.bank_account.duplicates.map { |ba| ba.intake.client_id }.uniq
```
caitbot, line 20: curious for my own understanding -- it doesn't seem like this is doing anything functionally different than what we had before. is it more efficient to do this than the active record query?

caitbot (approved):
this looks good, thanks for the write up and for the fix!

app/controllers/hub/security_controller.rb
tgrathwell: it's definitely not more efficient, probably severely inefficient for any cases where the array of `duplicates` has more than a couple things in it (not sure what the dupiest bank account we have is)

I was just trying as a quick fix to normalize between `BankAccount` objects which have an `intake_id` and `Archived::BankAccount2021` objects which have an `archived_intakes_2021_id`. If we just fetch the `intake` (which is configured correctly on both models to use the right keys/tables) it will get the data we need, though it's less efficient that we need to `map` over all the bank accounts and fetch intakes one-by-one to do so.

I could've also probably tried to run two big queries that tried to fetch all the dupes from the `bank_accounts` table and all the dupes from the `archived_intakes_2021` table but I didn't want to have to mess with the definition of #duplicates at all so I left it in this odd state

caitbot: ah yes, I forgot about the class weirdness of archived intakes vs non-archived intakes. ok fair point. this makes sense then. 

couldn't you do something like
```
Intake.where(id: @client.intake.bank_account.duplicates.pluck(:intake_id).and(Intake.where(id: @client.intake.bank_account.duplicates.pluck(:archived_intakes_2021_id))
``` 
then `pluck(:client_id)` from those?

I understand not wanting to mess with the definition of duplicates. This odd state will do. If it becomes unwieldly on cases of severe fraud then we can revisit it.

PR #2088, opened 2022-01-31
Hub: fix crash on messages page w/archived clients if there was an outbound call message

tgrathwell:


caitbot (approved):
I believe that this fixes it, and I can see that client is available to us, but super quick: why was this breaking? outbound_call.client was nil for archived clients?

tgrathwell:
It's not very clear from this code alone, but the SyntheticNote is being initialized with a `Hub::ClientsController::HubClientPresenter` which is aware that `client.intake` is sometimes an `Intake` and sometimes an `Archived::Intake2021`

`outbound_call.client` just returns a plain old client which has no such awareness

my goal, such as it is, has been to avoid teaching the `Client` model itself about `Archived::Intake2021` and try to leave that up to controllers and presenters and other things closer to the view, but it definitely has a sisyphean feel to it

PR #2092, opened 2022-02-01
Mark a client as needing ITIN help depending on their triage answer

caitbot:
image
image

### DIFF
--- a/app/views/hub/clients/_client_header.html.erb
+++ b/app/views/hub/clients/_client_header.html.erb
@@ -20,6 +20,10 @@
         <% if @client.access_locked? %>
           <%= link_to t("hub.unlock_account"), unlock_hub_client_path(id: @client), method: :patch, class: "button button--danger button--small", data: { confirm: t("hub.unlock_confirmation", name: @client.preferred_name) } %>
         <% end %>
+
+        <% if @client.needs_itin_help_yes? %>
+          <span class="label"><%= t("hub.itin_applicant") %></span>
+        <% end %>
       </div>
 
       <div class="grid-flex align-center">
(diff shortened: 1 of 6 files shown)


app/views/hub/clients/_client_header.html.erb
```diff
@@ -33,6 +33,9 @@
         </div>
       </div>
 
+      <% if @client.needs_itin_help_status == I18n.t("general.affirmative") %>
```
caitbot, line 36: Is this my favorite thing in the world? No.

An alternative is to have `needs_itin_help` return a boolean or `nil` in place of the translation string. Then the translation string can be put directly in the `erb`. Opinions?

My hesitation was having a method that returns a boolean or nil, which doesn't seem to be a pattern we have and so I opted for a consistent return value (always a string)

alaroia-cfa: People worry about [tri-state booleans](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem). You have something even weirder: you have a four-state boolean! :D

In case it's useful, here are some ideas.

- You can return `true`, `false`, `nil`, or `:n_a` from `needs_itin_help_status`.
- You could return `:yes` `:no` `:missing` or `:n_a` from `needs_itin_help_status`.
- You could move `needs_itin_help_status` out of `Client` and into `HubClientPresenter`, and then you could examine the status and return the right I18n string.

Of these, I like that last option the most. Not a blocker.

alaroia-cfa: Update: Most of what I said above was nonsensical but I tuple'd with Catie.

caitbot: We discussed on tuple, going to keep the translation string returning method in the presenter and also write a `needs_itin_help_yes?` method for determining whether or not to show the label

PR #2093, opened 2022-02-01
Persist tax return status as string called state

bytheway875:
I got most of the way through this refactor and then realized maybe it would make more sense to have created a method on TaxReturn called

```
def status
  state || latest_transition.current_state
end
```

But instead I took the approach of using `.state` everywhere....


alaroia-cfa:
I am fearful of how statesman seems (iirc?) to turn states into numbers, too, like those ‘duplicate key in index’ errors. Is this something you’ve been able to test? Is the general idea with statesman that we add states to the of its list, even if they’re semantically earlier states? One way to test might be, on this branch, try to make an intake, then add a new state (“intake_screener_review”) into the code, then transition an intake manually between a few states and see if you hit conflicts. It might be revealing, and if there are problems, we might be able to figure out how to resolve them.

This is not a general approval/rejection, just a comment.

bytheway875:
@alaroia-cfa i tested this on the branch and dont run into issues when inserting a new state.

Maybe it is because I allow all states to transition to any other state, whereas with the efile submissions restrict which statuses you can transition to. Although the error that gets raised is a db level error.

app/state_machines/tax_return_state_machine.rb
```diff
@@ -34,9 +35,64 @@ class TaxReturnStateMachine
...
+  STATES_BY_STAGE = begin
+                      stages = {}
+                      statuses = states.without("intake_before_consent")
+                      statuses.map do |status, _|
+                        stage = status.to_s.split("_")[0]
+                        stages[stage] = [] unless stages.key?(stage)
+                        stages[stage].push(status)
+                      end
+                      stages
+                    end.freeze
```
jenny-heath, line 59: it looks like these two blocks of code (`determine_statuses_by_stage` and `STATES_BY_STAGE`) are doing the same thing - do we need both?

app/models/bulk_tax_return_update.rb
```diff
@@ -25,7 +27,7 @@ class BulkTaxReturnUpdate < ApplicationRecord
   belongs_to :tax_return_selection
   belongs_to :assigned_user, class_name: "User", optional: true
 
-  enum status: TaxReturnStatus::STATUSES, _prefix: :status
+  enum status: TaxReturnStateMachine.states, _prefix: :status
```
jenny-heath, line 30: if we're understanding this line (we = me + @tgrathwell) it looks like it changes the enum definition from `{ intake_before_consent: 100 ...` to `["intake_before_consent", ...` which seems like it won't have the same meaning for btru's before and after this commit.

are we trying to preserve this enum for Max? (i'm assuming so since this adds the redundant `state` column) if so we should probably leave the enum definition the same (as in `tax_return.rb`)

bytheway875: yes that's true -- it is correctly maintained in tax_return model, but got changed incorrectly here. I will revert it. Also, the goal is to eventually drop status, but only once we've done the migrations to state and informed Max of the transition. We have to take a stepped approach.

app/controllers/hub/bulk_actions/change_assignee_and_status_controller.rb
```diff
@@ -73,10 +73,10 @@ def create_change_assignee_and_status_notifications!
         bulk_update = BulkTaxReturnUpdate.create!(
           tax_return_selection: @selection,
           assigned_user: @form.assigned_user,
-          status: status_action == BulkTaxReturnUpdate::KEEP ? nil : @form.status,
+          state: status_action == BulkTaxReturnUpdate::KEEP ? nil : @form.status,
           data: {
             assigned_user: assignment_action,
-            status: status_action
+            state: status_action
```
jenny-heath, line 79: should we also be updating the old `status` column so it's in sync with the new one? that appears to be what we're doing with `TaxReturn`s

app/controllers/portal/still_needs_helps_controller.rb
```diff
@@ -13,7 +13,7 @@ def update
 
       if current_client.update(update_params)
         if current_client.still_needs_help_yes?
-          current_client.tax_returns.where(status: "file_not_filing").each { |tax_return| tax_return.update!(status: "file_hold") }
+          current_client.tax_returns.where(state: "file_not_filing").each { |tax_return| tax_return.transition_to(:file_hold) }
```
jenny-heath, line 16: does this mean that some people could have gotten into a state where their tax return status was out of sync with their state machine transitions? not sure who was filling out this form and when (maybe no one!) in any case, if people's records did get out of sync, do we need to fix that somehow?

bytheway875: i dont think we do because this feature wont affect any "active" tax returns that are around this season. i also dont think it would have that much of a reporting impact because we didn't update to use the state machine for tax return statuses until the tax season was closed.

app/forms/hub/take_action_form.rb
```diff
@@ -88,8 +88,8 @@ def set_default_locale
       @locale = client.intake.locale
     end
 
-    def status_has_changed
-      errors.add(:status, I18n.t("forms.errors.status_must_change")) if status == tax_return&.current_state
+    def state_has_changed
+      errors.add(:state, I18n.t("forms.errors.status_must_change")) if state == tax_return&.current_state
```
jenny-heath, line 92: is there a particular reason to be getting `tax_return&.current_state` instead of `tax_return&.state`? do we always get state from one or the other (the transition or the tax return model)?

app/models/bulk_tax_return_update.rb
```diff
@@ -40,8 +42,8 @@ def data_stored_appropriately
 
   def updates
     updates = {}
-    if status.present?
-      updates["status"] = TaxReturnStatusHelper.status_translation(status)
+    if state.present?
+      updates["status"] = TaxReturnStatusHelper.status_translation(state)
```
jenny-heath, line 46: i _think_ this means that for any notifications that contain old btru's(that have a status but not a state), it'll look on the notifications tab like they didn't update the status/state

bytheway875: hmm... you are probably correct. i think i need to go through the bulk tax return updates part of things and make sure we read from status OR state until we do the backfill, then push another PR to only look at state once the state is done.

Tbh, this review is making me remember that the bulk updates area needs more attention in this update. thanks!

app/models/tax_return.rb
```diff
@@ -59,24 +61,22 @@ class TaxReturn < ApplicationRecord
...
+    ## backwards compatible with current implementation while TaxReturn#state can be nil
+    state || state_machine.current_state
   end
 
-  before_save do
-    if status == "prep_ready_for_prep" && status_changed?
-      self.ready_for_prep_at = DateTime.current
-    end
+  def ready_for_prep_at
+    tax_return_transitions.order(:created_at).find_by(to_state: "prep_ready_for_prep")&.created_at
```
jenny-heath, line 79: should we, for old records, delegate this to the old `ready_for_prep_at` column? (less important: not sure what this is used for, but maybe we can eventually get rid of the old column? not urgent though)

db/migrate/20220125014850_add_state_to_bulk_tax_return_updates.rb
```diff
@@ -0,0 +1,6 @@
+class AddStateToBulkTaxReturnUpdates < ActiveRecord::Migration[6.1]
+  def change
+    add_column :bulk_tax_return_updates, :state, :string
+    add_index :bulk_tax_return_updates, :state
```
jenny-heath, line 4: do we look up these records by state a lot? just wondering what the reason behind the index is

lib/tasks/tax_return_status_to_state_migration.rake
```diff
@@ -0,0 +1,26 @@
...
+  task tax_return_migrate: [:environment] do
+    TaxReturn.where(state: nil).where.not(status: "intake_before_consent").find_each do |tr|
+      tr.update(state: tr.status)
+    end
+  end
+
+  task tax_return_low_priority_migrate: [:environment] do
+    TaxReturn.where(state: nil).where(status: "intake_before_consent").find_each do |tr|
+      tr.update(state: tr.status)
+    end
```
jenny-heath, line 24: doing a `where ... find_each ... update` is going to be really slow, in these situations the fastest way might be
```
TaxReturn.where(...).find_in_batches do |batch|
  updates = batch.map { |tax_return| { id: tax_return.id, state: tax_return.status } }
  TaxReturn.upsert_all(updates)
end
```

spec/controllers/hub/bulk_actions/change_assignee_and_status_controller_spec.rb
```diff
@@ -61,12 +61,9 @@
...
 
-          expect(tax_return_1.reload.status).to eq new_status
-          expect(tax_return_1.current_state).to eq new_status
-          expect(tax_return_2.reload.status).to eq new_status
-          expect(tax_return_2.current_state).to eq new_status
-          expect(tax_return_3.reload.status).to eq new_status
-          expect(tax_return_3.current_state).to eq new_status
+          expect(tax_return_1.reload.state).to eq new_status
+          expect(tax_return_2.reload.state).to eq new_status
+          expect(tax_return_3.reload.state).to eq new_status
```
jenny-heath, line 66: since we're still updating both, should we continue to check status?

spec/services/mixpanel_service_spec.rb
```diff
@@ -254,154 +254,158 @@
...
+                "filing_rejected",
+                hash_excluding(
+                    {
+                        user_id: user.id,
+                        user_site_name: site.name,
+                        user_site_id: site.id,
+                        user_organization_name: organization.name,
+                        user_organization_id: organization.id,
+                        user_coalition_name: coalition.name,
+                        user_coalition_id: coalition.id,
```
jenny-heath, line 328: if you feel so inclined, it might be better to make this a positive rather than a negative test (so, like the above, would just have the hash that _is_ sent). it is a bit longer / redundant though and it was already like this so totally fine if you don't feel like changing it!

spec/state_machines/tax_return_state_machine_spec.rb
```diff
@@ -21,7 +21,7 @@
```
jenny-heath, line 21: sorry if i just lost track of this and it does exist somewhere but: is there a test that checks that the transition stays in sync with `tax_return.status` and `tax_return.state`? for some reason was expecting to see it here but fine if it lives somewhere else

app/forms/hub/take_action_form.rb
jenny-heath: having now read the rest of the PR, it looks like it's probably best to be reading `tax_return.current_state` since it is a method that checks both! can't remember whether there are places that read directly from the column, just came here to say: using `current_state` is good

app/models/tax_return.rb
jenny-heath: now that we've looked it up and see that this is only used for the mixpanel event at the end, it's definitely not important for it to delegate to the column version

and in fact we should someday delete that column

db/migrate/20210104224344_migrate_statuses_for_tax_returns.rb
```diff
@@ -1,9 +0,0 @@
-class MigrateStatusesForTaxReturns < ActiveRecord::Migration[6.0]
-  def up
-    # MigrateStatuses.migrate_all
```
caitbot, line 3: I see, this was just used to backfill. I am wary of deleting migrations but I understand that there is no reversing this. Should we keep it for documentation purposes?

db/schema.rb
```diff
@@ -797,6 +799,37 @@
     t.datetime "updated_at", precision: 6, null: false
   end
 
+  create_table "idme_users", force: :cascade do |t|
```
caitbot, line 802: should this be included? I assume this is an unrelated change.

db/schema.rb
```diff
@@ -840,6 +873,26 @@
     t.index ["created_at"], name: "index_incoming_text_messages_on_created_at"
   end
 
+  create_table "intake_site_drop_offs", force: :cascade do |t|
```
caitbot, line 876: Same with this

caitbot: There is actually a number of unrelated changes in this schema file. Best not to include them here.

db/migrate/20210104224344_migrate_statuses_for_tax_returns.rb
caitbot: I see, it's being replaced by a new rake task. Deleting it is fine.

app/controllers/portal/portal_controller.rb
```diff
@@ -28,11 +26,11 @@ def current_intake
     # Once we've started preparing their taxes, we don't want to prompt them through the intake flow, but instead
     # show their tax return status information.
     def show_tax_returns?
-      current_client.intake.completed_at? || current_client.tax_returns.map(&:status_before_type_cast).any? { |status| status >= 102 }
+      current_client.intake.completed_at? || current_client.tax_returns.map(&:current_state).any? { |state| TaxReturnStateMachine.states.index(state) >= TaxReturnStateMachine.states.index("intake_ready") }
```
caitbot, line 29: this is much nicer than random numbers

caitbot (approved):
I like this change a lot. It makes certain areas much more readable.

I left some comments. There are unrelated changes to schema that should be removed but other than that this looks great.

db/migrate/20210104224344_migrate_statuses_for_tax_returns.rb
bytheway875: i dont think this migration ever worked or did anything, but the pre-existing code was def deleted.

PR #2098, opened 2022-02-02
Add Customer Support role to have accessible users

bytheway875:


caitbot (approved):
yay thank you for fixing, looks great.

alaroia-cfa:
@bytheway875 If I understand this change correctly, it grants client success `:manage` access over `Client` and `EfileSubmission` records in all VitaPartners. I think that change makes sense.

I think it's best to also add a section to `ability_spec.rb` accordingly to help other engineers later who may wonder about our intended permissions. What do you think?

PR #2101, opened 2022-02-03
Fix active client count for organization capacities

caitbot:
Addresses [this story](https://www.pivotaltracker.com/story/show/181129414).

Also Asheesh paired on this whole thing with me and was a big help, I'm just bad at using git-duet. We learned a lot about joins.

### User impacting issue
Organization capacity active client counts are much higher than expected. There is a discrepancy between what is shown as the current client count and the number of clients pulled up via filtering on the all clients page.

### Why is this happening
Organization capacity, as calculated in `organization_capacities_v03.sql`, does not reference intakes at all. It loads any clients who have a tax return in a set of "non-terminal" statuses.

This means that if last years clients did not have all of their tax returns updated to a "terminal" status (that does not count against capacity) then they are still showing up in the current capacity count.

Contrast this with the clients page, which does an inner join on intake in order to exclude all clients with archived intakes. And so the numbers on the clients page are much lower than those shown for the active client capacity count.

### Solution
Similar to the inner join on intake in `client_sortable.rb`, we do an inner join on intake for client capacity so that we take only clients with this years intake into account.

### Implications
Clients without intakes will not be included in the active client count for an organization.

### DIFF
--- a/db/views/organization_capacities_v04.sql
+++ b/db/views/organization_capacities_v04.sql
@@ -1,0 +1,22 @@
+WITH
+    organization_id_by_vita_partner_id AS (
+        select id, (CASE WHEN parent_organization_id IS NULL THEN id ELSE parent_organization_id END) AS organization_id
+        from vita_partners
+    ),
+    client_ids AS
+    (
+        select distinct tax_returns.client_id
+        from tax_returns
+        INNER JOIN intakes ON (intakes.client_id) = tax_returns.client_id
+
+        where tax_returns.status >= 102 AND tax_returns.status <= 404 AND tax_returns.status != 403 AND tax_returns.status != 106 AND tax_returns.status != 130
+    ),
+    partner_and_client_counts AS (
+        SELECT organization_id, count(clients.id) as active_client_count
+        FROM organization_id_by_vita_partner_id
+        LEFT OUTER JOIN clients ON organization_id_by_vita_partner_id.id = clients.vita_partner_id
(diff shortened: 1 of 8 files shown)


db/views/organization_capacities_v04.sql
```diff
@@ -0,0 +1,22 @@
+WITH
+    organization_id_by_vita_partner_id AS (
+        select id, (CASE WHEN parent_organization_id IS NULL THEN id ELSE parent_organization_id END) AS organization_id
+        from vita_partners
+    ),
+    client_ids AS
+    (
+        select distinct tax_returns.client_id
+        from tax_returns
+        INNER JOIN intakes ON (intakes.client_id) = tax_returns.client_id
```
caitbot, line 10: this is the new line!

spec/models/organization_capacity_spec.rb
```diff
@@ -17,7 +17,7 @@
       context "for #{status} status" do
         let(:organization) { create :organization }
         before do
-          create :client_with_status, status: status[0], vita_partner: organization
+          create :client_with_status, status: status[0], vita_partner: organization, intake: create(:intake)
```
bytheway875, line 20: i wonder if we should just add creating an intake to this factory since its pretty weird to have clients without an intake

bytheway875:
this is so awesome and simple! good job ya'll

spec/models/organization_capacity_spec.rb
caitbot: I was thinking the same thing, I'll do that in a follow up

PR #2102, opened 2022-02-03
Duplicate guard checks for full primary ssn instead of email and phone number

embarnard:


app/helpers/duplicate_intake_guard.rb
```diff
@@ -1,17 +1,9 @@
...
-         (arel_table[:email_address].eq(email_address)
-           .and(arel_table[:email_address].not_eq(nil)))
-         .or(arel_table[:phone_number].eq(phone_number)
-           .and(arel_table[:phone_number].not_eq(nil)))
-      ).exists?
-  end
-
-  private
+    ClientLoginService.new(:gyr).accessible_intakes.each do |intake|
+      return true if DeduplificationService.duplicates(intake, :hashed_primary_ssn).exists?
```
bytheway875, line 4: Yay! I love that the deduplification sevice could be useful here.. However, this code would loop through every accessible intake in the application, and there will eventually potentially be millions! I think we need to refactor the Deduplication service query so that we can chain it onto accessible intakes and make these guys compatible. We can pair on that if you'd like?

bytheway875:
@embarnard Let me know if this feedback makes sense. We can pair on it today OR we can do a call and just chat about some different ways to accomplish this or however else you'd like to approach it.

bytheway875:
@embarnard i also see that its marked WIP so maybe you already knew this and have it covered haha

app/helpers/duplicate_intake_guard.rb
embarnard: @bytheway875 Oh yeah just saw this comment, thank you! But yes I caught this and have been trying to resolve it by doing this 

```
dupes = DeduplificationService.duplicates(intake, :hashed_primary_ssn)

if intake.type == "Intake::GyrIntake" && intake.drop_off?
      dupes.exist?
    elsif intake.type == "Intake::GyrIntake" && !intake.drop_off?
      dupes.exists?(primary_consented_to_service: "yes")
    else
      dupes.exists?('sms_phone_number_verified_at != nil OR email_address_verified_at != nil OR navigator_has_verified_client_identity != nil')
    end
```

embarnard: ugh but have a failing test because its not creating tax returns on my intake for some reason 🙄😭

bytheway875: @em i think we can make something work that is a little more straightforward.

embarnard: Ooo yeah i updated it

embarnard: maybe this could be more straight forward too
```
def has_duplicate?
    intake = Intake.find(id)
    return if intake.primary_ssn == nil

    dupes = DeduplificationService.duplicates(intake, :hashed_primary_ssn)

    if intake.type == "Intake::GyrIntake"
      dupes.any?(&:drop_off?) ? dupes.exists? : dupes.exists?(primary_consented_to_service: "yes")
    else
      dupes.exists?('sms_phone_number_verified_at != nil OR email_address_verified_at != nil OR navigator_has_verified_client_identity != nil')
    end
  end
```

bytheway875:
@embarnard i updated with a new approach.

You can now call `intake.duplicates` to get back all duplicates _of that service type_ and call `intake.has_duplicate?` to get back the boolean of whether there are duplicates or not.

If the ticket you're working on for the UI update also requires that you look for CTC duplicates on GYR tickets or GYR duplicates on CTC tickets, you can do something like this:

```
# gets CTC duplicates of a GYR Intake
gyr_intake = Intake::GyrIntake.last
DeduplificationService.duplicates(gyr_intake, :hashed_primary_ssn, from_scope: Intake::CtcIntake.accessible_intakes)

ctc_intake = Intake::CtcIntake.last
DeduplificationService.duplicates(ctc_intake, :hashed_primary_ssn, from_scope: Intake::GyrIntake.accessible_intakes)
```

^^ That will look at the hashed primary ssn on the GYR intake, and look for CTC intakes with the same information on it.

bytheway875:
Also, I think we need to include a rake task to backfill the hashed primary ssns on existing intakes. Can you do that?

embarnard:
> Also, I think we need to include a rake task to backfill the hashed primary ssns on existing intakes. Can you do that?
hmm this should be doing that `lib/tasks/backfill_hashed_intake_primary_ssn.rake`, is that what you were talking about?

app/models/intake/ctc_intake.rb
```diff
@@ -286,7 +288,12 @@ class Intake::CtcIntake < Intake
   enum has_primary_ip_pin: { unfilled: 0, yes: 1, no: 2 }, _prefix: :has_primary_ip_pin
   enum has_spouse_ip_pin: { unfilled: 0, yes: 1, no: 2 }, _prefix: :has_spouse_ip_pin
   enum consented_to_legal: { unfilled: 0, yes: 1, no: 2 }, _prefix: :consented_to_legal
-
+  scope :accessible_intakes, -> do
```
embarnard, line 291: Sleeeekk niceee 😎

app/models/intake/ctc_intake.rb
```diff
@@ -336,6 +343,24 @@ class Intake::CtcIntake < Intake
     }
   }
 
+  def duplicates
```
embarnard, line 346: Beautiful

app/models/intake/gyr_intake.rb
```diff
@@ -392,6 +398,16 @@ def bank_account
     @bank_account ||= BankAccount.new(account_type: type, bank_name: bank_name, account_number: bank_account_number, routing_number: bank_routing_number)
   end
 
+  def duplicates
+    return Model.none unless hashed_primary_ssn.present?
```
embarnard, line 402: @bytheway875 oh I'm unfamiliar with `Model.none`-- why wouldn't we just return nil?

bytheway875: the intent was to return an empty active record relation instead of nil in the hope that it might be easier to deal with a reponse of the same "type." Second guessing whether that's useful, so def down to have you change it to return nil

PR #2097, opened 2022-02-02
Remove references to old routing model (VitaPartnerState)

alaroia-cfa:
These were preventing Heroku deploys and otherwise making life strange.

Co-authored-by: Asheesh Laroia <alaroia@codeforamerica.org>

PR #2103, opened 2022-02-03
Cover db:seed with a test

alaroia-cfa:
In https://github.com/codeforamerica/vita-min/pull/2097 we fixed a problem where `db:seed` was failing.

With this pull request, we run `db:seed` to the test suite so we notice these problems earlier.

You can validate that this pull request works by adding:

```
should_crash = 1 / 0
```

to `db/seeds.rb`  and see that this test fails.

### DIFF
--- a/spec/tasks/seeds_spec.rb
+++ b/spec/tasks/seeds_spec.rb
@@ -1,0 +1,8 @@
+require 'rails_helper'
+
+describe 'db:seed' do
+  it 'runs without error' do
+    Rails.application.load_tasks
+    Rake::Task['db:seed'].invoke
+  end
+end


caitbot (approved):
haha this is great, no more broken db seeds!

tgrathwell:
this is good and valuable

I wish there was also something we could do about the Github/Heroku interaction such that deployments for PRs that fail `postdeploy` are marked as failed in some way

but this PR at least solves our immediate known cases why they might fail (and gets us to fix up the seeds sooner)

alaroia-cfa:
> I wish there was also something we could do about the Github/Heroku interaction such that deployments for PRs that fail `postdeploy` are marked as failed in some way

I agree! If you know one, let me know!!!

PR #2104, opened 2022-02-03
Update years on separated, divorced, widowed questions in GYR

caitbot:


bytheway875:
since the options are the same in all three of those, it might also make sense to create one helper that uses all of the years. That way, if we end up wanting to change the "before 2014" in the future or something like that, we dont have to track down all of the implementations

embarnard (approved):
Looks good! 🟢 @caitbot

PR #2108, opened 2022-02-03
Replace references to last year's backtaxes years with calculation based on TaxReturn.current_tax_year

jenny-heath:
i would screenshot every place where this changed in both languages but that would be excessive imo...
image

### DIFF
--- a/app/views/public_pages/faq.html.erb
+++ b/app/views/public_pages/faq.html.erb
@@ -22,7 +22,7 @@
             <p>
               <%= t(
                 "views.public_pages.faq.getting_started.non_filer.a",
-                years: [TaxReturn.current_tax_year - 3, TaxReturn.current_tax_year - 2, TaxReturn.current_tax_year - 1].to_sentence(locale: locale)
+                years: TaxReturn.backtax_years.reverse.to_sentence(locale: locale)
               ) %>
             </p>
           </div>
--- a/app/views/public_pages/faq.html.erb
+++ b/app/views/public_pages/faq.html.erb
@@ -19,7 +19,12 @@
             <h3><%=t("views.public_pages.faq.getting_started.non_filer.q") %></h3>
           </a>
           <div class="accordion__content" id="a1">
-            <p><%=t("views.public_pages.faq.getting_started.non_filer.a") %></p>
+            <p>
+              <%= t(
+                "views.public_pages.faq.getting_started.non_filer.a",
+                years: [TaxReturn.current_tax_year - 3, TaxReturn.current_tax_year - 2, TaxReturn.current_tax_year - 1].to_sentence(locale: locale)
+              ) %>
+            </p>
           </div>
         </div>
 
--- a/app/views/public_pages/home.html.erb
+++ b/app/views/public_pages/home.html.erb
@@ -103,7 +103,7 @@
         <p class="spacing-below-0">
           <%= t(
             "views.public_pages.home.backtaxes.body",
-            years: [TaxReturn.current_tax_year - 3, TaxReturn.current_tax_year - 2, TaxReturn.current_tax_year - 1].to_sentence(locale: locale)
+            years: TaxReturn.backtax_years.reverse.to_sentence(locale: locale)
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -3062,7 +3062,7 @@ en:
                   data-track-attribute-position="stimulus-how-faq3">
                 Get My Payment</a> tool will show a status of “more information needed” and you will have the opportunity to provide new bank information.
             - title: What if I haven’t filed federal taxes?
-              content_html: We recommend that you file your taxes in order to access the Economic Impact Payments and any additional tax credits you may be eligible for. You can file 2017, 2018, 2019, and 2020 returns for free with GetYourRefund.
+              content_html: We recommend that you file your taxes in order to access the Economic Impact Payments and any additional tax credits you may be eligible for. You can file 2018, 2019, 2020, and 2021 returns for free with GetYourRefund.
```
jenny-heath, line 3065: i was sad to learn that you can't use interpolation for these list-y things

app/views/public_pages/faq.html.erb
```diff
@@ -19,7 +19,12 @@
             <h3><%=t("views.public_pages.faq.getting_started.non_filer.q") %></h3>
           </a>
           <div class="accordion__content" id="a1">
-            <p><%=t("views.public_pages.faq.getting_started.non_filer.a") %></p>
+            <p>
+              <%= t(
+                "views.public_pages.faq.getting_started.non_filer.a",
+                years: [TaxReturn.current_tax_year - 3, TaxReturn.current_tax_year - 2, TaxReturn.current_tax_year - 1].to_sentence(locale: locale)
```
bytheway875, line 25: Can also use `TaxReturn.filing_years.without(TaxReturn.current_tax_year)`. If you need them with oldest year first, append `.reverse`

bytheway875: Actually, there's a method on TaxReturn that does this already:

`TaxReturn.backtax_years.reverse.to_sentence(locale: locale)`

config/locales/en.yml
bytheway875: You can -- you can just pass the variable in at the top level:

```
<% t("whatever.content_html", years: years) do |content| %>
  <p><%= content %></p>
<% end %>  


```

app/views/public_pages/home.html.erb
```diff
@@ -187,7 +192,7 @@
         <ul class="list--bulleted text--small text--semibold">
           <li class="spacing-below-0"><%= t("views.public_pages.home.features.free") %></li>
           <li class="spacing-below-0"><%= t("views.public_pages.home.features.chat") %></li>
-          <li class="spacing-below-0"><%= t("views.public_pages.home.features.backtaxes") %></li>
+          <li class="spacing-below-0"><%= t("views.public_pages.home.features.backtaxes", least_recent_bt_year: TaxReturn.current_tax_year - 3, most_recent_bt_year: TaxReturn.current_tax_year - 1) %></li>
```
bytheway875, line 195: TaxReturn.backtax_years.first, TaxReturn.backtax_years.last maybe?

config/locales/en.yml
jenny-heath: oh hm i tried that and it just displayed %{years} on the page. do we have an example of this somewhere in the code? i couldn't find anything on the internet (with a cursory search anyway)

jenny-heath: there aren't many upvotes here but if this is to be believed, it looks like it's not supported: https://stackoverflow.com/questions/21574900/interpolation-in-i18n-array

bytheway875: Ah -- the issue is that you can send things into arrays, but the implementation on that page has arrays with then a nested array of attributes, and we're accessing the data on that nested array directly instead of using the translation string. 

```
<% t("whatever.content_html", years: years).each do |content| %> <p><%= content[:item] %></p> <% end %>
```
That content[:item] part cant access the variables

PR #2109, opened 2022-02-04
Remove a bunch of unused stimulus-related triage views

jenny-heath:
looks like we don't use these. also they are very old


bytheway875 (approved):
🥳

PR #2117, opened 2022-02-08
Dupe check full ssns 180352601

bytheway875:


spec/models/ctc_intake_spec.rb
```diff
@@ -219,7 +219,73 @@
...
+    end
+
+    context "when neither phone number nor email are present" do
+      let(:intake) { create :ctc_intake, email_address: nil, sms_phone_number: nil }
+      it "responds with an empty ActiveRecord relation" do
+        expect(intake.duplicates).to eq described_class.none
+      end
+    end
+  end
+>>>>>>> Add tests for duplicate functionality
```
embarnard: we should remove this

PR #2121, opened 2022-02-08
Use Rails 6.2 defaults instead of Rails 6.0 defaults

tgrathwell:
so there's hopefully less steps to get to 7.x someday


tgrathwell:
oh we're actually on 6.1? well i'll be

PR #2123, opened 2022-02-08
Assists itin applicants 181132598

bytheway875:


app/models/user.rb
```diff
@@ -34,7 +34,7 @@
 #
 # Indexes
 #
-#  index_users_on_email                  (email)
+#  index_users_on_email                  (email) UNIQUE
```
bytheway875, line 37: I'm pretty sure this SHOULD be here -- i keep having to remove it from my PRs when it gets generated, but it is in fact a unique index ?

db/schema.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
caitbot, line 1: I forget why this gets put here. I've seen it before in others commits, but its always removed. I don't remember why.

All that being said, maybe we remove it

caitbot (approved):
looks good and simple enough to me.

I don't feel terribly strongly about the unique index thing. as long as it is that way on main. maybe it was a schema update that ever got committed?

PR #2124, opened 2022-02-08
Flow explorer: fix flake in returning_filer_spec

tgrathwell:
another occasion where we need to put dashes in the SSNs if it runs in a real chrome

also adds an incredibly heavy-handed patch for `fill_in` that will raise
an error if we ever try to fill in an SSN that doesn't match the expected format


alaroia-cfa:
🏋️ 🤝  = 💎 🤲

PR #2125, opened 2022-02-08
Do not ask clients who need help getting an ITIN for SSN/ITIN and block duplicates

jenny-heath:
image

### DIFF
--- a/app/controllers/questions/personal_info_controller.rb
+++ b/app/controllers/questions/personal_info_controller.rb
@@ -1,6 +1,12 @@
 module Questions
   class PersonalInfoController < QuestionsController
     include AnonymousIntakeConcern
+
+    def edit
+      @show_ssn_field = current_intake&.triage&.id_type_need_help? ? false : true
+      super
+    end
+
     def illustration_path; end
 
     def tracking_data
--- a/app/controllers/questions/personal_info_controller.rb
+++ b/app/controllers/questions/personal_info_controller.rb
@@ -3,7 +3,7 @@ module Questions
     include AnonymousIntakeConcern
 
     def edit
-      @show_ssn_field = current_intake&.triage&.id_type_need_help? ? false : true
+      @hide_ssn_field = current_intake.itin_applicant?
       super
     end
 
--- a/app/forms/consent_form.rb
+++ b/app/forms/consent_form.rb
@@ -2,9 +2,6 @@ class ConsentForm < QuestionsForm
   include BirthDateHelper
   set_attributes_for(
     :intake,
-    :primary_consented_to_service,
-    :primary_consented_to_service_ip,
-    :primary_consented_to_service_at,
     :birth_date_year,
     :birth_date_month,
     :birth_date_day,
@@ -21,8 +18,6 @@ class ConsentForm < QuestionsForm
       .except(:birth_date_year, :birth_date_month, :birth_date_day)
       .merge(
         primary_birth_date: parse_birth_date_params(birth_date_year, birth_date_month, birth_date_day),
-        primary_consented_to_service: "yes",
-        primary_consented_to_service_at: DateTime.now
       )
     intake.update(attributes)
   end
--- a/app/forms/personal_info_form.rb
+++ b/app/forms/personal_info_form.rb
@@ -7,8 +7,8 @@ class PersonalInfoForm < QuestionsForm
   validates :preferred_name, presence: true
   validates :phone_number, presence: true, confirmation: true
   validates :phone_number_confirmation, presence: true
-  validates_presence_of :primary_tin_type
-  validates_presence_of :primary_ssn
+  validates_presence_of :primary_tin_type, unless: :itin_applicant?
+  validates_presence_of :primary_ssn, unless: :itin_applicant?
   validates :primary_ssn, social_security_number: true, if: -> { ["ssn", "ssn_no_employment"].include? primary_tin_type }
   validates :primary_ssn, individual_taxpayer_identification_number: true, if: -> { primary_tin_type == "itin" }
 
@@ -25,5 +25,11 @@ class PersonalInfoForm < QuestionsForm
   def save
     state = ZipCodes.details(zip_code)[:state]
     @intake.update(attributes_for(:intake).except(:phone_number_confirmation).merge(state_of_residence: state))
+  end
+
+  private
+
+  def itin_applicant?
+    @intake&.triage&.id_type_need_help?
   end
 end
--- a/app/forms/personal_info_form.rb
+++ b/app/forms/personal_info_form.rb
@@ -30,6 +30,6 @@ class PersonalInfoForm < QuestionsForm
   private
 
   def itin_applicant?
-    @intake&.triage&.id_type_need_help?
+    @intake.itin_applicant?
   end
 end
--- a/spec/controllers/questions/consent_controller_spec.rb
+++ b/spec/controllers/questions/consent_controller_spec.rb
@@ -30,11 +30,13 @@ RSpec.describe Questions::ConsentController do
         allow(MixpanelService).to receive(:send_event)
       end
 
-      it "saves the answer, along with a timestamp and ip address" do
+      it "saves the answer, along with a timestamp and ip address and consent fields" do
         post :update, params: params
 
         intake.reload
         expect(intake.primary_consented_to_service_ip).to eq ip_address
+        expect(intake.primary_consented_to_service_at).not_to eq nil
+        expect(intake.primary_consented_to_service).to eq "yes"
       end
 
       it "authenticates the client and clears the intake_id from the session" do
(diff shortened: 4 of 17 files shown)


app/controllers/questions/personal_info_controller.rb
```diff
@@ -1,6 +1,12 @@
 module Questions
   class PersonalInfoController < QuestionsController
     include AnonymousIntakeConcern
+
+    def edit
+      @show_ssn_field = current_intake&.triage&.id_type_need_help? ? false : true
```
bytheway875, line 6: Since nil evaluates to false, I feel like you can skip the ternary here and just let this evaluate to the result of `id_type_need_help?`

jenny-heath: yeah i can't remember why i did this....maybe was feeling paranoid? i would just test it out locally to make sure it works as expected?

app/forms/personal_info_form.rb
```diff
@@ -26,4 +26,10 @@ def save
     state = ZipCodes.details(zip_code)[:state]
     @intake.update(attributes_for(:intake).except(:phone_number_confirmation).merge(state_of_residence: state))
   end
+
+  private
+
+  def itin_applicant?
```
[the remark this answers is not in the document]
jenny-heath: i didn't think about that! agree

app/controllers/questions/personal_info_controller.rb
bytheway875: note that per the PR open for updates to itin in hub, this is not the correct method name for getting to whether they are an itin applicant.

app/forms/consent_form.rb
```diff
@@ -2,9 +2,6 @@ class ConsentForm < QuestionsForm
   include BirthDateHelper
   set_attributes_for(
     :intake,
-    :primary_consented_to_service,
```
bytheway875, line 5: Consenting has been removed into the after update steps so that we don't end up consenting clients that we're offboarding.

spec/controllers/questions/consent_controller_spec.rb
```diff
@@ -202,6 +202,31 @@
         expect(error_messages[:primary_last_name].first).to eq "Please enter your last name."
       end
     end
+
+    context "with an itin applicant that has a duplicate" do
+      before do
+        allow(intake).to receive(:has_duplicate?).and_return true
+        allow(intake).to receive(:itin_applicant?).and_return true
+      end
```
alaroia-cfa, line 210: Trivial & stylistic, non blocking: Our codebase tends to have a blank line between a `before do...end` and a `let...end` and the start of an `it do...end`

spec/models/intake/gyr_intake_spec.rb
```diff
@@ -292,23 +292,65 @@
...
+
+      context "when only sms is present" do
+        let(:intake) { create :intake, primary_birth_date: Date.tomorrow, email_address: nil, sms_phone_number: "+18324658840", triage: triage }
+        it "responds with duplicates from sms" do
+          intake.duplicates
+          expect(DeduplificationService).to have_received(:duplicates).exactly(1).times.with(intake, :sms_phone_number, :primary_birth_date, from_scope: described_class.accessible_intakes)
+        end
+      end
+
+      context "when both email and sms are present" do
```
alaroia-cfa, line 318: Should you check `phone_number` in addition to `email` and `sms_phone_number`? I think so.

spec/models/intake_spec.rb
```diff
@@ -1041,4 +1041,27 @@
       expect(intake.relevant_intake_document_types).to eq doc_types
     end
   end
+
+  describe "#itin_applicant?" do
```
alaroia-cfa, line 1045: Not a blocker for this pull request:

At some point in the near future (1-5 days), since [we are going to have to let Hub users edit the ITIN-applicant nature of a client](https://cfa.slack.com/archives/GT1B45BS5/p1644533164143279), I think we should change `itin_applicant?` to be a boolean property on `Intake`. That way, Hub users would be editing Intake data rather than Triage data. In that case, we'd need to copy the triage answer onto the Intake, similar to how the backtaxes years work.

Would you agree with that?

This is not a request for a change on this pull request.

spec/models/intake/gyr_intake_spec.rb
```diff
@@ -292,23 +292,65 @@
...
-      let(:intake) { create :intake, primary_ssn: nil }
+    context "when an itin applicant" do
+      let(:dupe_double) { double }
+      let(:triage) { create(:triage, id_type: "need_itin_help" )}
+      before do
+        allow(DeduplificationService).to receive(:duplicates).and_return dupe_double
+        allow(dupe_double).to receive(:or)
+      end
+      context "when only email is present" do
+        let(:intake) { create :intake, primary_birth_date: Date.tomorrow, email_address: "mango@example.com", sms_phone_number: nil, triage: triage }
```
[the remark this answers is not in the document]
alaroia-cfa: It's also weird to me that we're querying email addresses without necessarily being 100% sure the email addresses have been confirmed. Should we add a filter on `email_address_verified_at` being non-nil?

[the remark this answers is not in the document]
alaroia-cfa: (Same idea for sms_phone_number)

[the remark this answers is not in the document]
alaroia-cfa: (And then of course we _don't_ verify the phone_number. Maybe that's a reason not to dedup against it.)

spec/controllers/questions/consent_controller_spec.rb
```diff
@@ -202,6 +202,31 @@
         expect(error_messages[:primary_last_name].first).to eq "Please enter your last name."
       end
     end
+
```
alaroia-cfa, line 205: I'd like to see a test covering the fact that `ConsentController` now is responsible for adding `primary_consented_to_service_at`.

spec/controllers/questions/personal_info_controller_spec.rb
```diff
@@ -7,6 +7,28 @@
...
   end
 
+  describe "#edit" do
+    context "when a client said they need help getting an ITIN during triage" do
+      let(:intake) { create :intake, triage: create(:triage, id_type: "need_itin_help") }
+
+      it "sets hide_ssn_field to be truthy" do
+        get :edit
+
+        expect(assigns(:hide_ssn_field)).to be_truthy
```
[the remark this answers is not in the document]
bytheway875: hmm i actually changed it from show_ssn_field to hide_ssn_field because I believe that the `hide` naming more accurately describes that show is the default and the exceptional case is to hide it.

spec/features/web_intake/itin_applicant_filer_spec.rb
```diff
@@ -0,0 +1,99 @@
+require "rails_helper"
+
+RSpec.feature "A client who wants help getting an ITIN" do
+  include MockTwilio
+  before do
+    # duplicated itin app
+    create(:intake, primary_birth_date: Date.new(1971, 3, 5), email_address: "gary.gardengnome@example.green", triage: (create :triage, id_type: "need_itin_help"))
+  end
+
+  scenario "the client fills out triage and makes it past the personal info page" do
```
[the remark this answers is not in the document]
alaroia-cfa: Oops I forgot to finish typing the above comment but now finished typing it.

spec/models/intake/gyr_intake_spec.rb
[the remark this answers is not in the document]
bytheway875: Also, everyone has to provide us an email for GYR but we only verify it if we haven't verified your sms phone number. To me, it wasn't asked in the ticket to de-dupe only against verified contact info, so this implementation fits the product ask.

spec/models/intake_spec.rb
bytheway875: the intake is not created at the time at which someone answers the triage question, so theres a logic change that will be needed to make that happen. I don't disagree with you, but that feels out of scope for this ticket.

spec/controllers/questions/consent_controller_spec.rb
bytheway875: i expected that it already did, but it only tested the ip_address getting saved. I'll add that.

spec/models/intake/gyr_intake_spec.rb
bytheway875: no, i dont think so.... phone number is where people take the intake call. Many people use shared phones for this.

alaroia-cfa: This is a very good point. Thanks for the reminder. :)

spec/models/intake_spec.rb
alaroia-cfa: Agreed, out of scope for this ticket, just offering a heads-up to you and looking to validate the idea. Sounds like you would consider such a change peaceful to do in the future.

spec/models/intake/gyr_intake_spec.rb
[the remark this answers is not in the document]
alaroia-cfa: This all makes sense to me. Thanks for the conversation!

PR #2126, opened 2022-02-08
Update org/coalition lead permissions to edit users in their org/coalition

caitbot:


app/views/hub/users/edit.html.erb
```diff
@@ -4,7 +4,7 @@
     <div>
       <h1><%= @user.name %></h1>
 
-      <% if current_user.admin? %>
+      <% if can?(:edit, @user) %>
```
caitbot, line 7: the funny thing is this determines if you see the link to the user edit page on the user index, so really, we could just remove it. I decided to leave it be, in case something changes to how we display that link, we let you SEE the page if you have a URL, you just can't save. this keeps the mutations and their intended use more protecteed.

bytheway875: i dont think i quite understand the distinction, but sounds fine haha

bytheway875 (approved):

PR #2130, opened 2022-02-09
Update the "potential duplicates" check in Hub to match on full SSN instead of contact info

embarnard:


app/models/client.rb
```diff
@@ -231,23 +231,12 @@ def generate_login_link
     Rails.application.routes.url_helpers.portal_client_login_url(id: raw_token)
   end
 
-  def clients_with_dupe_contact_info(is_ctc)
```
caitbot, line 234: Does it make sense to leave this in for @bytheway875 to use in the ticket where we check for ITIN dupes (https://www.pivotaltracker.com/story/show/181132531)? Since we won't be able to use SSN to check for dupes. I see that its in progress now, so might be worth syncing really quick with her (if you haven't already -- maybe you already have)

spec/models/client_spec.rb
```diff
@@ -458,78 +458,86 @@
...
 
-        context "with a client who hasn't reached consent" do
-          let!(:client_before_consent) { create :client_with_tax_return_state, intake: create(:intake, email_address: "fizzy_pop@example.com"), state: "intake_before_consent" }
+      context "looking for GYR matches" do
+        let(:display_type) { Intake::GyrIntake }
 
-          it "does not return the client who hasn't consented" do
-            expect(client.clients_with_dupe_contact_info(false)).not_to include(client_before_consent.id)
-          end
+        it "returns accessible CTC clients with the same ssn" do
```
caitbot, line 487: this should probably read `returns accessible GYR clients with the same ssn`

currently says CTC

spec/models/client_spec.rb
```diff
@@ -458,78 +458,86 @@
...
 
-        context "with a client who hasn't reached consent" do
-          let!(:client_before_consent) { create :client_with_tax_return_state, intake: create(:ctc_intake, email_address: "fizzy_pop@example.com"), state: "intake_before_consent" }
+      context "looking for GYR matches" do
+        let(:display_type) { Intake::GyrIntake }
 
-          it "does not return the client who hasn't consented" do
-            expect(client.clients_with_dupe_contact_info(true)).not_to include(client_before_consent.id)
-          end
+        it "returns accessible CTC clients with the same ssn" do
```
caitbot, line 509: same here, should be GYR rather than CTC

caitbot (approved):
This looks good. Some of the spec descriptions need changed before merging to main.

I like that it won't block people who do not have an SSN. That buys us time until Shannon puts in the contact matching dupe detector.

PR #2134, opened 2022-02-10
Strip the leading and trailing whitespace on emails on login

caitbot:


caitbot (commented):
Paired with Shannon on this! Reviewed too.

PR #2141, opened 2022-02-12
Change state column to current state

bytheway875:
Max objected to the column name of state, so I offered to change it to current_state.

This is likely better anyway because now we can access state from tax return and submissions using the same method.

In order to prevent downtime during the column name change, queries are now looking at status column instead of the column name that will change. That does mean I'll need to do ANOTHER follow up PR to fix that.

I also decided to add some safeguards that prevent us from writing to current_state or status directly, which meant some refactoring of other parts of the code AND some places in the tests where I was still writing directly to TaxReturn#status.

### DIFF
--- a/app/controllers/hub/tax_returns_controller.rb
+++ b/app/controllers/hub/tax_returns_controller.rb
@@ -67,7 +67,7 @@ module Hub
     end
 
     def load_assignable_users
-      @client ||= @tax_return&.client
+      @client ||= @tax_return.client
       @assignable_users = assignable_users(@client, [current_user, @tax_return&.assigned_user].compact)
     end
 
--- a/app/views/hub/tax_returns/new.html.erb
+++ b/app/views/hub/tax_returns/new.html.erb
@@ -2,7 +2,7 @@
 <% content_for :page_title, @page_title %>
 
 <% content_for :card do %>
-  <div class="slab">
+  <div class="slab slab--padded">
     <div class="grid">
       <h1 class="spacing-below-0"><%= @page_title %></h1>
       <p><%= t(".current_years") %> <%= @form.tax_return_years.join(", ") %> </p>
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
@@ -11,7 +9,9 @@
 # migrations use external dependencies or application code.
 #
 # It's strongly recommended that you check this file into your version control system.
+
 ActiveRecord::Schema.define(version: 2022_02_09_191119) do
+
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "plpgsql"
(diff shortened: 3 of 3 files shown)


app/controllers/hub/tax_returns_controller.rb
```diff
@@ -62,25 +62,21 @@ def update
...
     private
 
+    def load_client
+      @client = Client.find(params[:client_id])
+    end
+
     def load_assignable_users
-      @client ||= @tax_return.client
-      @assignable_users = assignable_users(@client, [current_user, @tax_return.assigned_user])
+      @client ||= @tax_return&.client
```
caitbot, line 70: in what cases are tax_return nil? are we seeing that? not that I disagree with this nil check, just curious

app/models/tax_return.rb
```diff
@@ -71,8 +71,16 @@ def state_machine
...
 
   def current_state
-    ## backwards compatible with current implementation while TaxReturn#state can be nil
-    state || state_machine.current_state || status
+    ## backwards compatible with current implementation while cutting over column name to current_state
+    read_attribute(:current_state) || read_attribute(:state) || state_machine.current_state || status
+  end
+
+  def current_state=(_)
+    raise("Avoid writing to TaxReturn#current_state directly. Instead, use #transition_to or #transition_to!")
```
caitbot, line 79: beautiful

spec/controllers/hub/unattended_clients_controller_spec.rb
```diff
@@ -13,31 +13,29 @@
...
+      let!(:client_within_sla) { create :client_with_intake_and_return, tax_return_state: "intake_ready", first_unanswered_incoming_interaction_at: 2.business_days.ago, last_internal_or_outgoing_interaction_at: 11.business_days.ago, vita_partner: site }
+      let!(:four_day_completed_status) { create :client_with_intake_and_return, tax_return_state: "file_accepted", first_unanswered_incoming_interaction_at: 4.business_days.ago, last_internal_or_outgoing_interaction_at: 11.business_days.ago, vita_partner: site}
+      let!(:four_day_breach_client) { create :client_with_intake_and_return, tax_return_state: "intake_ready", first_unanswered_incoming_interaction_at: 4.business_days.ago, last_internal_or_outgoing_interaction_at: 11.business_days.ago, vita_partner: site }
+      let!(:six_day_breach_client) { create :client_with_intake_and_return, tax_return_state: "intake_ready", first_unanswered_incoming_interaction_at: 6.business_days.ago, last_internal_or_outgoing_interaction_at: 11.business_days.ago, vita_partner: site }
       let!(:ten_day_breach_client_done_filing) do
-        create :client,
+        c = create :client,
```
caitbot, line 21: did you run into issues doing the nested build pattern that existed before? just curious, don't really have a preference but wondering if its something to watch out for

caitbot (approved):
Nice! Ignore my comments about `current_state`, from our slack conversation I understand that will be updated in a later PR to avoid downtime.

All this looks good, I would update the test description before shipping but other than that, no blocking comments.

spec/controllers/hub/unattended_clients_controller_spec.rb
bytheway875: yeah, unfortunately now that we're not able to write to current_state directly, the tax_return needs to be persisted in order for current_state to be present.

PR #2142, opened 2022-02-12
Consent drop off clients

bytheway875:
Mark as primary_consented_to_service_at

Also, use primary_consented_to_service_at instead of the enum.

Need to backfill on prod for existing drop offs when we release.

### DIFF
--- a/app/forms/hub/create_client_form.rb
+++ b/app/forms/hub/create_client_form.rb
@@ -107,7 +107,9 @@ module Hub
     def default_intake_attributes
       {
         type: "Intake::GyrIntake",
-        visitor_id: SecureRandom.hex(26)
+        visitor_id: SecureRandom.hex(26),
+        primary_consented_to_service: "yes",
+        primary_consented_to_service_at: DateTime.now
       }
     end
 
--- a/app/models/intake/gyr_intake.rb
+++ b/app/models/intake/gyr_intake.rb
@@ -351,11 +351,7 @@ class Intake::GyrIntake < Intake
   enum widowed: { unfilled: 0, yes: 1, no: 2 }, _prefix: :widowed
   enum wants_to_itemize: { unfilled: 0, yes: 1, no: 2, unsure: 3 }, _prefix: :wants_to_itemize
   enum received_advance_ctc_payment: { unfilled: 0, yes: 1, no: 2, unsure: 3 }, _prefix: :received_advance_ctc_payment
-  scope :accessible_intakes, -> do
-    online_consented = joins(:tax_returns).where({ tax_returns: { service_type: "online_intake" } }).where(primary_consented_to_service: "yes")
-    drop_off = joins(:tax_returns).where({ tax_returns: { service_type: "drop_off" } })
-    online_consented.or(drop_off)
-  end
+  scope :accessible_intakes, -> { where.not(primary_consented_to_service_at: nil) }
   after_save do
     if saved_change_to_completed_at?(from: nil)
       InteractionTrackingService.record_incoming_interaction(client) # client completed intake
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -540,10 +540,10 @@ describe Client do
     let(:ctc_client_inaccessible_ssn_match) { create :client, intake: create(:ctc_intake, primary_ssn: primary_ssn, sms_phone_number_verified_at: nil, email_address_verified_at: nil, navigator_has_verified_client_identity: nil) }
     let(:ctc_client_accessible_no_ssn_match) { create :client, intake: create(:ctc_intake, primary_ssn: "123456789", sms_phone_number_verified_at: DateTime.now) }
 
-    let(:gyr_client_accessible_ssn_match_online) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: primary_ssn, primary_consented_to_service: "yes"), tax_returns: [(create :tax_return, service_type: "online_intake")] }
-    let(:gyr_client_accessible_ssn_match_drop_off) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: primary_ssn, primary_consented_to_service: "unfilled"), tax_returns: [(create :tax_return, service_type: "drop_off")] }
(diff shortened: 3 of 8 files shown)


spec/models/client_spec.rb
```diff
@@ -540,10 +540,10 @@
...
-    let(:gyr_client_accessible_ssn_match_drop_off) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: primary_ssn, primary_consented_to_service: "unfilled"), tax_returns: [(create :tax_return, service_type: "drop_off")] }
-    let(:gyr_client_inaccessible_ssn_match) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: primary_ssn, primary_consented_to_service: "unfilled"), tax_returns: [(create :tax_return, service_type: "online_intake")] }
-    let(:gyr_client_accessible_no_ssn_match) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: "123456789"), tax_returns: [(create :tax_return, service_type: "drop_off")] }
+    let(:gyr_client_accessible_ssn_match_online) { create :client_with_tax_return_state, intake: create(:intake, primary_ssn: primary_ssn, primary_consented_to_service_at: DateTime.now), tax_returns: [(create :tax_return, service_type: "online_intake")] }
```
alaroia-cfa, line 543: IMHO the factory should now do this, perhaps depending on `tax_return_state` being `after_consent`.

app/models/intake/gyr_intake.rb
```diff
@@ -351,11 +351,7 @@ class Intake::GyrIntake < Intake
   enum widowed: { unfilled: 0, yes: 1, no: 2 }, _prefix: :widowed
   enum wants_to_itemize: { unfilled: 0, yes: 1, no: 2, unsure: 3 }, _prefix: :wants_to_itemize
   enum received_advance_ctc_payment: { unfilled: 0, yes: 1, no: 2, unsure: 3 }, _prefix: :received_advance_ctc_payment
-  scope :accessible_intakes, -> do
-    online_consented = joins(:tax_returns).where({ tax_returns: { service_type: "online_intake" } }).where(primary_consented_to_service: "yes")
-    drop_off = joins(:tax_returns).where({ tax_returns: { service_type: "drop_off" } })
-    online_consented.or(drop_off)
-  end
+  scope :accessible_intakes, -> { where.not(primary_consented_to_service_at: nil) }
```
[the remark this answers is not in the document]
alaroia-cfa: I also personally aesthetically prefer `primary_consented_to_service: "yes"` vs the non-`nil` check, but that's a very low priority and I have no rational basis for it.

spec/services/client_login_service_spec.rb
```diff
@@ -97,14 +97,6 @@
           it "returns a blank set" do
             expect(subject.clients_for_token("raw_token")).to match_array []
           end
-
```
alaroia-cfa, line 100: I can't say I precisely understand right now the purpose of this test block that got deleted. What was it doing, and why is it safe to delete? I could help figure out what it was for, if that's useful.

alaroia-cfa (commented):
I think we need a backfill before we can deploy this code; since I don't see a backfill PR yet, I'll wait to "Approve" in GitHub, but if a backfill already happens/happened, then you can consider it approved.

See comments for some additional comments and suggestions.

app/forms/hub/create_client_form.rb
```diff
@@ -107,7 +107,9 @@ def self.permitted_params
     def default_intake_attributes
       {
         type: "Intake::GyrIntake",
-        visitor_id: SecureRandom.hex(26)
+        visitor_id: SecureRandom.hex(26),
+        primary_consented_to_service: "yes",
```
alaroia-cfa, line 111: The IP address field... can/should this be the Hub user IP address at the time the Hub user adds them? I would suppose yes. What do you think?

Daydreaming a little more, we could make this a be a fourth enum option `primary_consented_to_service: "dropoff"`. That's perhaps too much hassle, so ignore this enum comment unless it sparks joy.

The IP address thing I do want your opinion on.

spec/models/client_spec.rb
bytheway875: thats not the way the code works though :(

spec/services/client_login_service_spec.rb
```diff
@@ -128,13 +120,13 @@
...
     let(:sms_notification_opt_in) { "yes" }
-    let(:primary_consented_to_service) { "yes" }
+    let(:primary_consented_to_service_at) { DateTime.current }
     let(:service_type) { "online_intake" }
 
     context "service_type is :gyr" do
       subject { described_class.new(:gyr) }
 
-      let(:intake) { (create :intake, sms_phone_number: phone_number, primary_consented_to_service: primary_consented_to_service, sms_notification_opt_in: sms_notification_opt_in)}
+      let(:intake) { (create :intake, sms_phone_number: phone_number, primary_consented_to_service_at: primary_consented_to_service_at, sms_notification_opt_in: sms_notification_opt_in)}
```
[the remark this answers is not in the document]
bytheway875: `_at` should be the only thing necessary moving forward, to my mind. i want to drop the "yes/no", and IP is only used when the client is executing.

app/forms/hub/create_client_form.rb
bytheway875: Hmm... i dont know. i think it makes sense to leave the IP nil for drop offs, and to use information about the intake and tax returns that show that it is a drop off instead of muddying with another enum option. Especially because then one would need to maintain knowledge that "yes" and "dropoff" are both valid options, and the point of this PR is to decrease overhead of "unique" data states.

spec/services/client_login_service_spec.rb
bytheway875: there used to be unique behavior for drop off clients. the behavior is now the same for clients/intakes of all types, so it does not make sense to maintain tests that test the unique logic of a certain type of return.

app/models/intake/gyr_intake.rb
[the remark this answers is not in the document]
bytheway875: Also, I backfilled the existing dropoffs on Friday and planned to backfill the others right before we deploy. There are always drop off clients coming in, so i want to make sure none are skipped.

app/forms/hub/create_client_form.rb
bytheway875: the `_ip` field implies that the consent was done online, whereas dropoff clients are essentially consenting verbally or in writing / signature to the hub user on-site, depending on the VITA process at that location.

PR #2143, opened 2022-02-14
Allow a hub user to mark ITIN-specific type documents [#181132742]

alaroia-cfa:


tgrathwell:
Closing this PR because we've decided we cannot currently validate Form1040s have an associated tax return (some 81 records in the DB right now have a 1040 without a tax return)

PR #2147, opened 2022-02-15
Update user name display to include role

caitbot:


bytheway875 (approved):
Nice! I love how it includes making sure places that were already using something similar are using the same method!!

PR #2150, opened 2022-02-16
Update the user's name display in a few more places

caitbot:


caitbot (commented):
Reviewed as pair with Travis

PR #2152, opened 2022-02-16
Remove stimulus and full-service landing pages

caitbot:


bytheway875:
This LGTM -- wondering if its clear to the requesters that the source params that used to have this unique behavior will now direct straight to the home page. Assuming yes!

bytheway875 (approved):

PR #2154, opened 2022-02-16
Route ITIN applicants to ITIN-VITA sites

embarnard:


app/services/partner_routing_service.rb
```diff
@@ -62,6 +65,26 @@ def vita_partner_from_source_param
...
+    vita_partners_ids_in_state = StateRoutingTarget.preload(:target).where(state_abbreviation: state, target_type: "VitaPartner").pluck(:target_id)
+
+    vita_partners_in_state_itin_enabled = VitaPartner.where(id: vita_partners_ids_in_state, accepts_itin_applicants: true)
+
+    if vita_partners_in_state_itin_enabled.present?
+      @routing_method = :itin_enabled
+      return vita_partners_in_state_itin_enabled.order(Arel.sql('RANDOM()')).first
+    elsif vita_partners_itin_enabled = VitaPartner.where(accepts_itin_applicants: true)
+      @routing_method = :itin_enabled
+      return vita_partners_itin_enabled.order(Arel.sql('RANDOM()')).first if vita_partners_itin_enabled.present?
```
embarnard, line 84: Do we also need to check if the vita partner active? But then why does only `from_previous_year_partner` do this check, should we also put it everywhere else?

app/services/partner_routing_service.rb
```diff
@@ -62,6 +65,30 @@ def vita_partner_from_source_param
...
+                                                            .where("routing_fraction > ?", 0).pluck(:vita_partner_id)
+
+    active_vita_partners_in_state_itin_enabled = VitaPartner.where(id: active_vita_partners_ids_in_state, accepts_itin_applicants: true)
+
+    if active_vita_partners_in_state_itin_enabled.present?
+      @routing_method = :itin_enabled
+      return active_vita_partners_in_state_itin_enabled.order(Arel.sql('RANDOM()')).first
+    end
+
+    active_vita_partners_itin_enabled = VitaPartner.joins(:state_routing_fractions)
```
caitbot, line 83: might be valuable to leave a comment here that is like `# route them to any ITIN enabled partner if none are available in state` because it took me a second to realize this wasn't just duplicate code 😆 

I keep trying to think if there is a way to refactor this to make it so it does less queries, but this gets the job done.

caitbot (approved):
looks good to me!

PR #2157, opened 2022-02-16
Create routes for ajax_mixpanel_events in CtcDomain

jenny-heath:
2 things:

1) i forgot that the click handler appends "click_" to the events so i removed the redundant "click_" from the property name
2) more importantly, these events were not getting to mixpanel from the ctc domain because we didn't create the routes. i think. i'm still trying to get the events to show up in the dev environment on mixpanel.

this makes me think we should have better tests for the mixpanel events that are fired using the JS handler


jenny-heath:
i tried testing this in several ways and just got so stuck you can't imagine

jenny-heath:
however, in the time i was trying to write a test, the dang ctc click events finally showed up in mixpanel so that's good i guess

image

tgrathwell (approved):

PR #2158, opened 2022-02-17
Redirect the correct stimulus page

caitbot:
Oops removed the CTC one originally and we actually want to ~~remove~~ redirect the Stimulus FAQ to the beginning of intake.

This PR also includes changes for a more aggressive redirect straight to backtaxes for people coming in with the "full-service" source because client success wants to use it to send people directly into intake.


app/controllers/questions/triage_controller.rb
```diff
@@ -32,7 +32,7 @@ def form_attributes
     end
 
     def redirect_if_matching_source_param
-      redirect_to backtaxes_questions_path if SourceParameter.find_vita_partner_by_code(session[:source]).present? || session[:source] == "full-service"
```
caitbot, line 35: don't need this since redirecting directly on home

app/views/ctc/ctc_pages/stimulus_home.html.erb
```diff
@@ -0,0 +1,203 @@
+<% intake_public_access = ENV['CTC_INTAKE_PUBLIC_ACCESS'] == 'true' %>
+
```
caitbot, line 2: this just adds back what I previously deleted 🤦🏻

app/controllers/ctc/ctc_pages_controller.rb
```diff
@@ -17,6 +22,17 @@ def help
       render :home
     end
 
+    def stimulus
```
caitbot, line 25: adding back

app/views/public_pages/home.html.erb
```diff
@@ -73,11 +73,6 @@
           <strong><%= t("views.public_pages.home.stimulus_faq.info") %></strong>
         </p>
       </div>
-      <div class="grid__item width-one-third button-container">
```
caitbot, line 76: want to keep links and just do the redirect

app/controllers/public_pages_controller.rb
```diff
@@ -18,7 +18,11 @@ def source_routing
     redirect_to root_path
   end
 
-  def home; end
+  def home
+    if session[:source] == "full-service"
```
alaroia-cfa, line 22: If they come in via `/full-service` then happily go to intake, and then they decide to learn more about GYR, will this change mean that when they view the homepage, they get redirected back into intake?

config/locales/en.yml
```diff
@@ -2559,6 +2559,49 @@ en:
           sub_header: All you need to do is provide us with your name and either an email address or phone number, and we'll email or text you when there are updates.
           submit: Keep me updated
           title: Sign Up
+      stimulus_home:
```
caitbot, line 2562: add it back

app/controllers/public_pages_controller.rb
```diff
@@ -50,6 +54,11 @@ def page_not_found
 
   def tax_questions; end
 
+  def stimulus
```
alaroia-cfa, line 57: The story https://www.pivotaltracker.com/n/projects/2409240/stories/181142024/comments/229655892 seems to say that visitors to demo.getyourrefund.org/stimulus should be redirected to home, but this redirects them to the beginning of intake. Does it make more sense to remove this `PublicPagesController#stimulus` method entirely?

spec/controllers/ctc/ctc_pages_controller_spec.rb
```diff
@@ -36,7 +36,13 @@
       [
         %w( cactc /en/help ),
         %w( fed   /en/help ),
-      ].each do |source, location|
+        %w( eip /en/stimulus-navigator ),
```
caitbot, line 39: add it back

spec/controllers/ctc/ctc_pages_controller_spec.rb
```diff
@@ -57,6 +63,33 @@
     end
   end
 
+  describe "#stimulus" do
```
caitbot, line 66: add it back

alaroia-cfa (changes requested):
I left two comments which I think merit changes, unless I don't understand the behavior in this pull request or I don't understand the story. Let me know what you think.

app/controllers/public_pages_controller.rb
caitbot: hahahaha i suppose they will, let's see, I can keep the route, but have it redirect them to `backtaxes` and set the session source?

alaroia-cfa: I like that!

app/controllers/public_pages_controller.rb
alaroia-cfa: Oh cool yay

spec/controllers/public_pages_controller_spec.rb
```diff
@@ -87,6 +87,23 @@
     end
   end
 
+  describe "#stimulus" do
+    it "redirects to the beginning of intake" do
+      get :stimulus
```
alaroia-cfa, line 92: Consider testing that the `session[:source]` is set. Not a blocker.

app/controllers/public_pages_controller.rb
```diff
@@ -50,6 +50,16 @@ def page_not_found
...
+    # the copy on this page is wrong and until we get it updated, redirect to beginning of intake
+    redirect_to_beginning_of_intake
+  end
+
+  def full_service
+    session[:source] = "full-service"
```
alaroia-cfa, line 59: The other things I've seen only set the `session[:source]` if it's not already set. Should this stick to that behavior, too?

alaroia-cfa (approved):
Two minor comments that embody suggestions; no blockers; feel free to merge, with or without addressing those comments.

app/controllers/public_pages_controller.rb
caitbot: I'm pretty sure we actually do re-set it everytime, see this comment
````  # Allow session[:source] to re-set on every interaction so we can track all
  # unique source entry points in Mixpanel.```

alaroia-cfa: Fascinating. Thanks!

spec/controllers/public_pages_controller_spec.rb
caitbot: Based on the acceptance criteria, I do not think that we want the session source to be set by going to this page. I tested it in dev and it does not set the `session[:source]`. This makes sense to me because this is actually a FAQ page, not really so much a landing page, and thus it is linked to from home (among other places) so if it set session source, it might overwrite the clients actual session source.

alaroia-cfa: Ah hah. Thanks!

PR #2161, opened 2022-02-17
Do not send Mixpanel event for /healthcheck

bensheldon:
Result of brief investigation as to why Mixpanel's `GetYourRefund.Dev` account is emitting a significant number of events: 30%+ of the activity of `GetYourRefund.Prod`.  

This may not be the only source of extra events, but seems like a good place to start,


alaroia-cfa:
I resolved a merge conflict (please update your local repo to the latest `origin/main`).

lgtm; I'll wait for tests to agree and I'll merge. Thanks @bensheldon !

alaroia-cfa (approved):

PR #2164, opened 2022-02-17
Protect against nil intake on client when intake is archived in GYR portal dashboard

embarnard:


alaroia-cfa (commented):
Dumb question -- how does someone with only an archived intake even log in? I'd rather see us prevent that from happening.

@jenny-heath @tgrathwell I think that you two were the chief intake archivists. Would you agree?

@embarnard you might want to get review from those two, as I seem to have more questions than I can answer.

app/controllers/portal/portal_controller.rb
```diff
@@ -26,11 +27,11 @@ def current_intake
     # Once we've started preparing their taxes, we don't want to prompt them through the intake flow, but instead
     # show their tax return status information.
     def show_tax_returns?
-      current_client.intake.completed_at? || current_client.tax_returns.map(&:current_state).any? { |state| TaxReturnStateMachine.states.index(state) >= TaxReturnStateMachine.states.index("intake_ready") }
+      current_intake.completed_at? || current_client.tax_returns.map(&:current_state).any? { |state| TaxReturnStateMachine.states.index(state) >= TaxReturnStateMachine.states.index("intake_ready") }
     end
 
     def ask_for_answers?
-      !current_client.intake.completed_at? && current_client.tax_returns.map(&:current_state).all? { |state| (TaxReturnStateMachine::STATES_BY_STAGE["intake"]).include?(state) }
+      !current_intake.completed_at? && current_client.tax_returns.map(&:current_state).all? { |state| (TaxReturnStateMachine::STATES_BY_STAGE["intake"]).include?(state) }
```
bytheway875, line 34: This is going to ask people to go back into the intake if their _archived_ intake was not marked as completed. They will likely get an error when they click to reenter the intake because it will be looking for a 2022 intake, which they will not have.

My recommendation is that instead of trying to bend an archived intake to look like the client's current year intake, we show something to the client that's more useful than the page breaking but also indicates that they don't have a current year intake and either a) prompt them to start an intake for this year or b) tell them to reach out to customer service.

The real bug to me here seems to be that clients who don't have current year intakes shouldn't even be able to log in... are we somehow loading a client associated with an archived intake instead of the this-year version of the client when there is duplicated contact info for multiple client objects?

PR #2168, opened 2022-02-17
Upgrade to Ruby 2.7.5 from 2.6.5

bensheldon:
- Updated Ruby Version. _~Originally went to 2.7.5 but @alaroia-cfa asked me to go all the way to 3.0.3 🎸~ `attr_encrypted` gem is [not compatible with Ruby](https://github.com/attr-encrypted/attr_encrypted/issues/398) 3 😭_ 
- Have the Gemfile read the ruby version dynamically from the `.ruby-version` file so there's one less place to get out of sync.
- Updated 1 test that asserts on an ArgumentError message (TIL, it changed across ruby versions)
- Successfully built the `Dockerfile` locally up until the Aptible-specific steps of adding `.aptive.env`; it's now necessary to add the `.ruby-version` before bundling. 

Next steps for the reviewer:

- Verify that this successfully builds on Aptible by deploying to a staging environment


tgrathwell:
i am currently trying to push this to staging but it is taking a while

bensheldon:
Please don't deploy the v3.0 commit. It doesn't pass in CI and yes, I imagine trying to access a model (or write to one) using `attr_encrypted` would crash.

PR #2170, opened 2022-02-18
Preload :child_sites to optimize taggable_vita_partners [#178761000]

alaroia-cfa:
Before this change, taggable_vita_partners seems to do a lot of SQL queries (N+1), and http://localhost:3000/en/hub/clients shows me a lot of queries if I have a lot of `organizations` in the database. After, it only does two, due to the preload.

### DIFF
--- a/app/helpers/tagging_helper.rb
+++ b/app/helpers/tagging_helper.rb
@@ -1,10 +1,11 @@
 module TaggingHelper
   def taggable_vita_partners(vita_partners)
     taggable_vita_partners = []
-    vita_partners.organizations.each do |vita_partner|
-      taggable_vita_partners << { id: vita_partner.id, name: vita_partner.name, value: vita_partner.id }
-      vita_partner.child_sites.each do |site|
-        taggable_vita_partners << { id: site.id, name: site.name, parentName: vita_partner.name, value: site.id }
+    # We query via the Organization model to trigger auto-loading of child_sites per Organization model's default scope.
+    Organization.where(id: vita_partners.organizations).each do |organization|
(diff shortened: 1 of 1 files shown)


app/helpers/tagging_helper.rb
```diff
@@ -1,10 +1,11 @@
 module TaggingHelper
   def taggable_vita_partners(vita_partners)
     taggable_vita_partners = []
-    vita_partners.organizations.each do |vita_partner|
-      taggable_vita_partners << { id: vita_partner.id, name: vita_partner.name, value: vita_partner.id }
-      vita_partner.child_sites.each do |site|
-        taggable_vita_partners << { id: site.id, name: site.name, parentName: vita_partner.name, value: site.id }
+    # We query via the Organization model to trigger auto-loading of child_sites per Organization model's default scope.
```
caitbot, line 4: I wondered if it would be better to do this where we load vita partners in the controller (since this is only used on the client index) but looks like that relies on the application controller to load vita partners and isn't worth the trouble.

I like this, it wouldn't have immediately occurred to me to query for them again, but makes sense that it avoids the n+1. neato!

caitbot (approved):
nice! thanks for improving!

app/helpers/tagging_helper.rb
alaroia-cfa: Good idea. I'll keep that in mind if we see other places that might need this treatment. One oddity is that `@vita_partners` is an access-controlled collection of sites _and_ organizations; if it were `@organizations` I'd definitely agree it would make sense to do the preload there.

PR #2175, opened 2022-02-24
copy changes and remove outdated faq to ctc homepage

bytheway875:


app/views/ctc/ctc_pages/home.html.erb
```diff
@@ -8,7 +8,7 @@
 
 <%= render 'shared/facebook_pixel' %>
 <main role="main" class="ctc-home">
-  <%= render 'ctc_offseason_warning' %>
```
bytheway875, line 11: i dont know whether to completely remove this, or leave it and comment it out so that its easier to toggle if we end up wanting to use it again during/after the season?

alaroia-cfa: I'm inclined to say remove it fully. It'll be easy to bring back. Not a blocker.

alaroia-cfa: This reminds me that I'm not sure we have written down the open/close procedures for GetCTC and what the expectations are on different dates. Do you want to talk with Yvonne/Kelly about that? Want me to?

alaroia-cfa (approved):
Approved! I left some comments; feel free to merge with those comments addressed, or not, at your option.

PR #2176, opened 2022-02-24
CircleCI: retry failed rails tests once every build

tgrathwell:
breaking the "in case of emergency, retry tests" glass because
we already retry them all the time in the CircleCI UI anyway

obviously the better answer is to fix all the tests, maybe someday!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,9 +106,15 @@ jobs:
       - run: |
           EAGER_LOAD=1 RAILS_CACHE_CLASSES=1 bundle exec rspec \
               --format RspecJunitFormatter \
-              --out ~/test-results/rspec/rspec.xml \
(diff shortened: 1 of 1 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -1044,4 +1044,12 @@ def hub_path; end
       end
     end
   end
+
+  it "fails this test half the time" do
```
alaroia-cfa, line 1048: Maybe we can rewrite this so that it succeeds if rspec is running failures, so it deterministically passes the second time.

alaroia-cfa:
@tgrathwell I think this PR is ready to go except I think we should remove the two new tests, which (I think) were just added to test re-testing. Do you agree?

PR #2178, opened 2022-02-24
Sticky js

bytheway875:
im using a small sticky plugin ive used in the past to accomplish the stickiness because flex and sticky dont always play well together (was having trouble getting it to work here) and im not really interested in refactoring the entire component in order to make this header sticky.

### DIFF
--- a/app/views/shared/_service_comparison.html.erb
+++ b/app/views/shared/_service_comparison.html.erb
@@ -1,6 +1,6 @@
 <div class="slab">
   <div class="grid">
-    <div class="service-comparison">
+    <div class="service-comparison data-sticky-container">
       <h2 class="title">
         <%= t("views.shared.service_comparison.title_html") %>
       </h2>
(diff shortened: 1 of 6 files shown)


app/views/shared/_service_comparison.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="slab">
   <div class="grid">
-    <div class="service-comparison">
+    <div class="service-comparison data-sticky-container">
```
caitbot, line 3: where does this class get used? looks like there is only a new Sticky for mobile, which makes sense to me. is this a special class defined by the sticky library?

caitbot (approved):
sweet, nice and simple, looks good.

app/views/shared/_service_comparison.html.erb
bytheway875: the sticky library looks for the class passed to new Sticky() instance _inside of_ the data-sticky-container. Data-sticky-container becomes the reference point for what the class you designate gets "stuck" to. It's kind of a weird implementation since a part of it is so explicit and the other is so inferred.

PR #2180, opened 2022-02-25
Add an index to archived intakes table to speed up #probable_previous_year_intake

tgrathwell:
currently the ConsentController is slower than it ought to be
because it's making a non-indexed query on the very large archived
intakes table


alaroia-cfa:
I like <s>traffic lights</s> indexes

alaroia-cfa (approved):

alaroia-cfa:
@tgrathwell I'm gonna boldly merge; let me know if that isn't what you wanted; you're gray on Tuple and this seems harmless and great.

PR #2183, opened 2022-02-25
Hub: validate presence of code in source param form

tgrathwell:
so it doesn't crash and annoy us in sentry!

# now it is fixed 
Screen Shot 2022-02-25 at 3 01 23 PM


alaroia-cfa (approved):
I approve, but since we're scheduled to pair, I propose we Tuple before merging it.

alaroia-cfa (approved):
Oops I forgot TG is out this week; merging.

PR #2191, opened 2022-02-28
Fix Kernel#open deprecation warning by using URI#open

bytheway875:
Saw a deprecation warning when running tests locally, decided to update it.


caitbot (approved):
sweeeet!

PR #2192, opened 2022-02-28
Add mixpanel_token to heroku credentials so we can test mixpanel events on demo apps

bytheway875:
uses the same token as staging


caitbot (approved):
niiiice

PR #2196, opened 2022-03-01
Split personal info page #181316172

jenny-heath:
image

image

### DIFF
--- a/app/controllers/questions/ssn_itin_controller.rb
+++ b/app/controllers/questions/ssn_itin_controller.rb
@@ -3,7 +3,7 @@ module Questions
     include AnonymousIntakeConcern
 
     def self.show?(intake)
-      intake.itin_applicant? ? false : true
+      !intake.itin_applicant?
     end
 
     def illustration_path; end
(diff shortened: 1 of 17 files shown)


app/controllers/questions/ssn_itin_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Questions
+  class SsnItinController < QuestionsController
+    include AnonymousIntakeConcern
+
+    def self.show?(intake)
+      intake.itin_applicant? ? false : true
```
caitbot, line 6: it looks like before this was relying on the method in the form, but there is no method in the new form. where is this defined? on intake?

caitbot (approved):
This looks great to me!

app/controllers/questions/ssn_itin_controller.rb
bytheway875: If the return value of the method is a boolean, it's redundant to have a ternary returning a boolean as well. can this be refactored to `intake.itin_applicant?`

jenny-heath: oops haha good point @bytheway875 - at first i didn't realize there was a method on intake so the ternary made more sense, i just forgot to replace it when i finally found the method.

@caitbot the method on the form looked like this:
```
def itin_applicant?
    @intake.itin_applicant?
  end
```
so, yes, it's on intake

PR #2198, opened 2022-03-01
Verification attempt don't block approved people

bytheway875:
Also moves No dependents over to be a fraud indicator and refactors some stuff around determining presence of qualifying dependents to not require an efile submission to determine.


mquadri23 (approved):
Thanks for making the update to the splat!

PR #2200, opened 2022-03-02
Add gyr-ids page to the flow and gyr-diy for later triaging

caitbot:
image
image

### DIFF
--- a/app/controllers/questions/triage_gyr_ids_controller.rb
+++ b/app/controllers/questions/triage_gyr_ids_controller.rb
@@ -1,0 +1,17 @@
+module Questions
+  class TriageGyrIdsController < TriageController
+    layout "intake"
+
+    def self.show?(triage)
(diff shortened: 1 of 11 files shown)


app/controllers/questions/triage_gyr_ids_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Questions
+  class TriageGyrIdsController < TriageController
+    layout "intake"
+
+    def self.show?(triage)
```
caitbot, line 5: checking with Yvonne but I think it makes sense to start showing this page when people get triaged to GYR with the current triage

caitbot: ok this is cool 👍🏻

jenny-heath (approved):
looks pretty good!

PR #2201, opened 2022-03-02
Don't show tax return count on hub client index page when it is greater than 1000

embarnard:

### DIFF
--- a/spec/controllers/hub/clients_controller_spec.rb
+++ b/spec/controllers/hub/clients_controller_spec.rb
@@ -614,11 +614,25 @@ RSpec.describe Hub::ClientsController do
           }
         end
 
-        it "shows the full amount of tax returns" do
-          get :index, params: params
+        context "when there are less than 1000 tax returns" do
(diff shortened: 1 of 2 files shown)


spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -614,11 +614,25 @@
...
         end
 
-        it "shows the full amount of tax returns" do
-          get :index, params: params
+        context "when there are less than 1000 tax returns" do
+          it "shows the full amount of tax returns" do
+            get :index, params: params
 
-          expect(assigns(:tax_return_count)).to eq 50
+            expect(assigns(:tax_return_count)).to eq 50
```
alaroia-cfa, line 621: Presumably the right hand side of the equality check needs the same `.to_s` that the implementation got!

alaroia-cfa (approved):
Looks great!

I encourage you to merge, then try to see in Datadog APM if the hub index page is faster.

PR #2203, opened 2022-03-03
add most recent verification attempt in security tab

mquadri23:


app/controllers/hub/security_controller.rb
```diff
@@ -11,6 +11,15 @@ def show
       @security_events = (
         @client.efile_security_informations + @client.recaptcha_scores
       ).sort_by(&:created_at)
+
+      @verification_attempt = VerificationAttempt.includes(:client, :transitions).find(params[:id])
```
bytheway875, line 15: There will not be a params ID coming through here that corresponds to a verification attempt, so you will get very unepected behavior here.

bytheway875: How would you grab the first or last item from the DB from a model?

We'd use `VerificationAttempt.first` or `VerificationAttempt.last`

Try applying similar logic to the client's set of verification attempts, knowing that you already have the client loaded and the client has an association to many verification_attempts.

PR #2204, opened 2022-03-03
Update service routing pages #181350984

caitbot:
For [this ticket](https://www.pivotaltracker.com/story/show/181350984)

`gyr-express`
image

`diy`
image

`gyr`
image


app/views/questions/triage/_gyr_tile.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <div class="refund-indicator">
+      <div class="spacing-below-10"><%= t('questions.triage.refund') %></div>
+      <div class="symbols"><%= image_tag("icons/money.svg") %> <%= image_tag("icons/money.svg") %> <%= image_tag("icons/money.svg") %></div>
+    </div>
+  </div>
+  <p class="text--red-bold">
+    <%= t('.warning') %>
+  </p>
+  <div class="spacing-above-35">
+    <%= link_to t('.choose_gyr'), next_path, class: "button button--wide text--centered", "data-track-click": click_tracking_event_name %>
```
jenny-heath, line 27: maybe this is just paranoia, but something about splitting this view into a shared partial makes me worried that `next_path` is going to be different depending on where this is used (not that it currently links to the wrong place, just that it could in the future if rendered on a different page). a couple of options are:
- replace `next_path` with something more explicit, like the other tile views
- pass in the path/url from the view that renders the partial (and that view can pass in `next_path`)
- leave it the way it is, it might be totally fine

caitbot: thats a totally valid concern! Thinking about it from a readability standpoint, using next path here isn't optimal because you have to go searching for all the callers of this partial. 

I'll make it an explicit path to `gyr-ids`!

jenny-heath (approved):
looks good! left one little comment. i'd make sure that whoever acceptance tests this actually uses the mixpanel dev environment to make sure the click tracking events still work

PR #2206, opened 2022-03-07
Update filters to take a range

bytheway875:
Change SLA definition to six days, and allow the scope to take a range.

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -25,7 +25,7 @@ module ClientSortable
     clients = clients.where(tax_returns: { service_type: @filters[:service_type] }) if @filters[:service_type].present?
     clients = clients.where(intake: Intake.where(had_unemployment_income: "yes")) if @filters[:unemployment_income].present?
     clients = clients.where(vita_partner: VitaPartner.allows_greeters) if @filters[:greetable].present?
-    clients = clients.first_unanswered_incoming_interaction_communication_breaches(@filters[:sla_breach_date]) if @filters[:sla_breach_date].present?
+    clients = clients.first_unanswered_incoming_interaction_between(...@filters[:sla_breach_date]) if @filters[:sla_breach_date].present?
     clients = clients.where(intake: Intake.where(with_general_navigator: true).or(Intake.where(with_incarcerated_navigator: true)).or(Intake.where(with_limited_english_navigator: true)).or(Intake.where(with_unhoused_navigator: true))) if @filters[:used_navigator].present?
 
     if @filters[:vita_partners].present?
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/client_sortable.rb
```diff
@@ -25,7 +25,7 @@ def filtered_clients
     clients = clients.where(tax_returns: { service_type: @filters[:service_type] }) if @filters[:service_type].present?
     clients = clients.where(intake: Intake.where(had_unemployment_income: "yes")) if @filters[:unemployment_income].present?
     clients = clients.where(vita_partner: VitaPartner.allows_greeters) if @filters[:greetable].present?
-    clients = clients.first_unanswered_incoming_interaction_communication_breaches(@filters[:sla_breach_date]) if @filters[:sla_breach_date].present?
+    clients = clients.first_unanswered_incoming_interaction_between(...@filters[:sla_breach_date]) if @filters[:sla_breach_date].present?
```
alaroia-cfa, line 28: `between` may be a little confusing. Maybe `first_unanswered_incoming_interaction_within` is a better name? Not a blocker.

app/models/client.rb
```diff
@@ -96,8 +96,8 @@ def self.sortable_intake_attributes
   scope :with_eager_loaded_associations, -> { includes(:vita_partner, :intake, :tax_returns, tax_returns: [:assigned_user]) }
   scope :sla_tracked, -> { distinct.joins(:tax_returns).where.not(tax_returns: { status: TaxReturnStateMachine::EXCLUDED_FROM_SLA }) }
 
-  scope :first_unanswered_incoming_interaction_communication_breaches, ->(breach_threshold_datetime) do
-    sla_tracked.where(arel_table[:first_unanswered_incoming_interaction_at].lteq(breach_threshold_datetime))
+  scope :first_unanswered_incoming_interaction_between, ->(range) do
```
alaroia-cfa, line 99: Given the relative simplicity of the scope, I think it's better to remove the scope entirely, and have the callsites directly query `Client.sla_tracked.where(first_unanswered_incoming_interaction_at: ...@breach_date)`. Bold suggestion, but that's what comes to mind. Not a blocker; I can live with the scope existing.

app/views/hub/metrics/index.html.erb
```diff
@@ -8,7 +8,7 @@
       </div>
 
       <div class="spacing-above-15">
-        SLA Breach: Clients that haven't been contacted in 5 business days (since <%= long_formatted_datetime(@report.breached_at)%>)
+        SLA Breach: Clients that haven't been contacted in 6 business days (since <%= long_formatted_datetime(@report.breached_at)%>)
```
alaroia-cfa, line 11: Is this text change something product/design approves of? I guess it's an internal page anyway, but it's probably good to make sure they like whatever we end up with on this page. Not a blocker.

alaroia-cfa (approved):
Some questions, no blockers! Feel free to merge as-is or handle the comments.

app/models/client.rb
bytheway875: its called in multiple places, and it's important that the implementations are the same.

app/views/hub/metrics/index.html.erb
bytheway875: The text change is accurate to the change they requested.

PR #2207, opened 2022-03-07
Fix dynamic take action changes

bytheway875:
Regression from updating column name to current_state -- the dynamic changes on the take action page weren't being applied. Decided to change what it's called on the take action page instead since the take action javascript is shared with the bulk auto update.

### DIFF
--- a/app/controllers/hub/clients_controller.rb
+++ b/app/controllers/hub/clients_controller.rb
@@ -98,7 +98,7 @@ module Hub
         @client,
         current_user,
         tax_return_id: params.dig(:tax_return, :id)&.to_i,
-        current_state: params.dig(:tax_return, :current_state),
+        status: params.dig(:tax_return, :current_state) || params.dig(:tax_return, :status),
         locale: params.dig(:tax_return, :locale)
       )
     end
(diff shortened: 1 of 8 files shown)


app/controllers/hub/clients_controller.rb
```diff
@@ -98,7 +98,7 @@ def edit_take_action
         @client,
         current_user,
         tax_return_id: params.dig(:tax_return, :id)&.to_i,
-        current_state: params.dig(:tax_return, :current_state),
+        status: params.dig(:tax_return, :current_state) || params.dig(:tax_return, :status),
```
bytheway875, line 101: Because the field is called current_state on the tax return drop down in the client header, we want to load from current_state there. But when we change on the page, we need to call it status.

app/forms/hub/take_action_form.rb
```diff
@@ -9,9 +9,9 @@ class TakeActionForm < Form
                   :internal_note_body,
                   :action_list,
                   :current_user,
-                  :client,
-                  :current_state
-    validates_presence_of :current_state
+                  :client
+    attr_accessor :state_transition, :status
```
tgrathwell, line 13: is this `state_transition` needed? seems like `:current_state` -> `:status` was the necessary part

app/controllers/hub/clients_controller.rb
tgrathwell, line 101: it might be nice if this form consumed the parameter as `updated_status` because that's what it says on the form label

but not a blocker I guess

tgrathwell (approved):

PR #2222, opened 2022-03-09
Dependent relationships and some dependent rules refactoring

bytheway875:
I moved the definition of which relationships are eligible and what IRS relationship value they map to into a yml file. It seems easier to manage additions, subtractions, and reorders in one place like this, and can provide instant visibility to business stakeholders to make sure we're all on the same page.

I made a bit of an opinionated change to the Dependent::Rules footprint to take in a dependent object instead of the arguments. I found the implementation with 11 arguments a bit difficult to reason about when trying to understand the existing logic. Because of the ultimate implementation of the relationship stuff, this refactor wasn't strictly necessary within this PR. Open to discussion about tradeoffs between the implementations.


app/models/archived/dependent_2021.rb
```diff
@@ -202,8 +202,8 @@ def rules_2021
       rules(2021)
     end
 
-    def rules(year)
-      Dependent::Rules.new(birth_date, year, full_time_student_yes?, permanently_totally_disabled_yes?, ssn.present?, qualifying_child_relationship?, qualifying_relative_relationship?, meets_misc_qualifying_relative_requirements_yes?, meets_qc_residence_condition_generic?, meets_qc_claimant_condition?, meets_qc_misc_conditions?)
+    def rules(tax_year)
+      Dependent::Rules.new(self, tax_year)
```
bytheway875, line 206: I found the implementation that passed in all of the dependent attributes individually a bit difficult to reason about and extend upon. This is a bit of an opinionated change, so I'm open to talking about tradeoffs.

db/relationships.yml
```diff
@@ -0,0 +1,68 @@
+relationships:
```
bytheway875, line 1: Consolidating the values stored in the db, the relationship, and the value we send to the irs into one file is helpful because we can provide a link to program to make sure that what we're defining aligns with their expectations. It also gives us a way to deal with irs_enum values that are NOT in a consistent pattern (which is the case with these newer relationships we're adding)

config/locales/en.yml
```diff
@@ -264,24 +264,29 @@ en:
...
-      '09_sister': Sister
-      10_brother: Brother
-      11_other: Other
-      12_stepchild: Stepchild
-      13_stepbrother: Stepbrother
-      14_stepsister: Stepsister
-      15_half_brother: Half brother
-      16_half_sister: Half sister
-      17_grandparent: Grandparent
+      aunt: Aunt
```
bytheway875, line 267: order of these keys is now established by the order that the items are listed in the yml file.

app/models/dependent.rb
```diff
@@ -196,6 +182,13 @@ def mixpanel_data
   # add a default year with no prefix.
   delegate :age, :born_in_final_6_months?, :disqualified_child_qualified_relative?, :meets_qc_age_condition?, :meets_qc_residence_condition?, :qualifying_child?, :qualifying_relative?, to: :rules_2020, prefix: :yr_2020
   delegate :age, :born_in_final_6_months?, :disqualified_child_qualified_relative?, :meets_qc_age_condition?, :meets_qc_residence_condition?, :qualifying_child?, :qualifying_relative?, to: :rules_2021, prefix: :yr_2021
+  delegate :qualifying_child_relationship?, :qualifying_relative_relationship?, to: :relationship_info
+
+  def relationship_info
+    return unless relationship.present?
+
+    @_relationship_info ||= Efile::Relationships.new(relationship)
```
bytheway875, line 190: will memoizing this cause a problem if someone changes the relationship on a dependent?

app/lib/efile/relationships.rb
```diff
@@ -0,0 +1,41 @@
+module Efile
+  class Relationships
+    RELATIONSHIPS = %w[qualifying_relative qualifying_child ineligible].freeze
+    @@relationships ||= IceNine.deep_freeze!(
+      (YAML.load_file(Rails.root.join("db/relationships.yml"))['relationships']).to_h do |relationship|
```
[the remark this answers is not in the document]
bytheway875: interesting -- i put it where i did because thats where the other file that gets consumed as data is (db/states.yml).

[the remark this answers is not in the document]
bytheway875: i moved it

app/lib/efile/relationships.rb
```diff
@@ -0,0 +1,41 @@
+module Efile
+  class Relationships
+    RELATIONSHIPS = %w[qualifying_relative qualifying_child ineligible].freeze
+    @@relationships ||= IceNine.deep_freeze!(
+      (YAML.load_file(Rails.root.join("app/lib/efile/relationships.yml"))['relationships']).to_h do |relationship|
```
alaroia-cfa, line 5: YAML file could have a comment at the top that order matters.

bytheway875:
Tuple reviewed and revised with @alaroia-cfa

PR #2223, opened 2022-03-09
Verification attempt history 181407759

bytheway875:


app/controllers/hub/security_controller.rb
```diff
@@ -8,6 +8,7 @@ class SecurityController < ApplicationController
     def show
       @client = Hub::ClientsController::HubClientPresenter.new(Client.find(params[:id]))
       @duplicate_bank_client_ids = duplicate_bank_client_ids
+      @most_recent_verification_attempt = @client.verification_attempts&.last
```
bytheway875, line 11: unnecessary &.

app/state_machines/verification_attempt_state_machine.rb
```diff
@@ -11,15 +11,15 @@ class VerificationAttemptStateMachine
   transition from: :escalated, to: [:approved, :denied, :requested_replacements]
 
   after_transition(to: :approved, after_commit: true) do |verification_attempt, transition|
-    verification_attempt.client.update(identity_verified_at: transition.created_at)
+    verification_attempt.client.update(identity_verified_at: transition.created_at, identity_verification_denied_at: nil)
```
mquadri23, line 14: Why do we have to set the identity_verification_denied_at to nil?

app/state_machines/verification_attempt_state_machine.rb
```diff
@@ -11,15 +11,15 @@ class VerificationAttemptStateMachine
...
     efile_submission = verification_attempt.client.efile_submissions&.last
     if efile_submission.present? && efile_submission.in_state?(:fraud_hold)
       efile_submission.transition_to(:resubmitted, { initiated_by_id: transition.metadata[:initiated_by_id] })
     end
   end
 
   after_transition(to: :denied, after_commit: true) do |verification_attempt, transition|
-    verification_attempt.client.update(identity_verification_denied_at: transition.created_at)
+    verification_attempt.client.update(identity_verification_denied_at: transition.created_at, identity_verified_at: nil)
```
mquadri23, line 22: Same comment as above. Since the variable is nil anyway why do even have to define it?

spec/state_machines/verification_attempt_state_machine_spec.rb
```diff
@@ -6,21 +6,28 @@
     let(:user) { create :admin_user }
 
     context "to approved" do
+      before do
+        verification_attempt.client.touch(:identity_verification_denied_at)
```
mquadri23, line 10: how does .touch work?

db/migrate/20220309052057_add_client_bypass_note_to_verification_attempt.rb
```diff
@@ -0,0 +1,5 @@
+class AddClientBypassNoteToVerificationAttempt < ActiveRecord::Migration[6.1]
+  def change
+    add_column :verification_attempts, :client_bypass_request, :text
```
mquadri23, line 3: so this is adding a new column to the verification attempts table correct?

app/views/hub/verification_attempts/show.html.erb
```diff
@@ -19,18 +19,26 @@
...
       </ul>
 
-      <div class="grid-flex">
-        <% if @verification_attempt.selfie.present? %>
-          <div class="item-15r">
-            <%= image_tag transient_storage_url(@verification_attempt.selfie.blob), id: "selfie" %>
+      <% if @verification_attempt.client_bypass_request.present? %>
+        <div class="note">
+          <div class="note__body">
+            <%= simple_format(@verification_attempt.client_bypass_request) %>
```
mquadri23, line 25: trying to figure out what this looks like in the UI...is it just a note that says something like bypass?

db/seeds.rb
```diff
@@ -159,4 +159,5 @@
 
 if defined?(FactoryBot)
   FactoryBot.create :verification_attempt, :with_fraud_hold_efile_submission
+  FactoryBot.create :verification_attempt, :with_fraud_hold_efile_submission, client_bypass_request: "I don't have an ID but I'd like submit my taxes."
```
mquadri23, line 162: *to submit my taxes

app/state_machines/verification_attempt_state_machine.rb
bytheway875: only one of the options should be filled in at a time, or else you'd be able to be both globally approved and denied which is weird.

app/state_machines/verification_attempt_state_machine.rb
bytheway875: this is wiping out a value if it had been previously set

spec/state_machines/verification_attempt_state_machine_spec.rb
bytheway875: .touch updates this column in the db to be the current timestamp.

db/migrate/20220309052057_add_client_bypass_note_to_verification_attempt.rb
bytheway875: correct!

app/views/hub/verification_attempts/show.html.erb
bytheway875: This will show whatever text the client submitted as to why they should have their request approved without providing photos. So in the case that they don't have an ID, they might say something like "I currently don't have access to my government issued ID."

"I actually am unhoused right now and I don't have access to a device with a camera"

mquadri23 (approved):

PR #2224, opened 2022-03-09
GYR homepage changes

embarnard:


app/assets/stylesheets/components/_slabs.scss
```diff
@@ -37,6 +37,10 @@
   background-color: $color-grey-light;
 }
 
+.slab--green {
+  background-color: #E8FDF7;
```
bytheway875, line 41: seems like we're using this in a couple places now -- wondering if we should turn this into a variable?

app/assets/stylesheets/templates/_home.scss
```diff
@@ -216,45 +216,110 @@
...
-  .slab--testimonials {
-    background-color: white;
-
+  .slab--rating {
+    padding-bottom: 3.5rem;
+    .title {
+      font-size: 2.7rem;
+    }
+    .text--small {
+      font-size: 1.4rem;
```
bytheway875, line 225: I pulled down the code and this subtitle text looks a little "tight" to the title above it. The line height in figma is set to 20px ... maybe we can try that or add a bit of top spacing?

app/assets/stylesheets/templates/_home.scss
```diff
@@ -216,45 +216,110 @@
...
+        text-align: center;
+        padding: 0 1.5rem 0 1.5rem;
+      }
     }
+  }
 
-    #testimonial-right {
-      padding-left: 6rem;
+  .slab-habla-espanol {
+    padding-top: 0;
```
bytheway875, line 246: I noticed that the h2 in the mock has less padding between the p than the implementation does. I think that's probably fine instead of overriding the h2 spacing, though.

app/assets/stylesheets/templates/_home.scss
```diff
@@ -216,45 +216,110 @@
...
+        background-color: #FFD29C;
+      }
+    }
+    #testimonial-right {
+      padding-left: 3rem;
+      .rounded-box {
+        background-color: #C2C2FF;
+      }
+    }
+    @media screen and (max-width: $tablet-up) {
```
bytheway875, line 293: I have a preference for nesting the media query logic into each selector's logic so that it's easier to compare the differences, but that's just my personal preference. It seems like we do a mix of both in the repo.

bytheway875:
I noticed that the contained width of contents in the mock is 1090px, but the "grid" wrapper we have on everything is containing  the width to 960px... I'm not sure whether that's something we should address, but it does create some minor layout issues (see: Claim your credits easily with our trusted support where "support" is being cut onto a second line)

That being said, the service component with is _wider_ than what I'm seeing in figma (I guess I did something inadvertently to make it wider than the design shows 😬 ) and design didn't have an issue with the width not being pixel perfect so it is perhaps totally fine for us to leave the width constrained to 960.

bytheway875 (approved):
This looks really really nice! I left a couple of comments but nothing to me looks blocking, unless you want to check in with design on the constrained width of the page.

Ohh also, I like to run SVGs through this service:

https://jakearchibald.github.io/svgomg/

It compresses the size of the SVG a little and takes out some of the extraneous markup that isn't needed for web implementations of an SVG.

PR #2226, opened 2022-03-09
Perform Datadog metrics calls in Concurrent::Future background thread

bensheldon:
Wraps `DatadogApi.guage` and `DatadogApi.increment` calls in a `Concurrent::Future` that can be force-resolved with `value!` inside of tests.

There is no error handling because I'm not sure what would be appropriate (suggestions please!).

If this pattern is acceptable, I'll subsequently look at Mixpanel calls or anything else that could appropriately be fire-and-forget.

I noticed a pattern of unit testing against the underlying adapter; I replaced those by testing against the wrapper.

### DIFF
--- a/app/lib/datadog_api.rb
+++ b/app/lib/datadog_api.rb
@@ -62,7 +62,9 @@ module DatadogApi
   def self.emit_point(label:, tags:, value:, type:)
     tags << "env:#{configuration.env}"
     Concurrent::Future.execute do
-      self.client.emit_point(self.apply_namespace(label), value, { tags: tags, type: type })
+      client.emit_point(self.apply_namespace(label), value, { tags: tags, type: type })
+    rescue Net::OpenTimeout
+      nil
     end
   end
   private_class_method :emit_point
--- a/app/lib/datadog_api.rb
+++ b/app/lib/datadog_api.rb
@@ -1,9 +1,8 @@
 module DatadogApi
-
   METRIC_TYPES = {
-      count: "count".freeze,
-      gauge: "gauge".freeze,
-      rate: "rate".freeze,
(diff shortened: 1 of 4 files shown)


tgrathwell:
@jenny-heath and I made a little script trying to play with Concurrent::Future.execute to dispel any fears that it may throw messages into the void

```
puts "Hello, i am outside a thread"

Concurrent::Future.execute do
  puts "I am inside a thread! When will I execute?"
end

Concurrent::Future.execute do
  sleep 2
  puts "I had to sleep before I printed this!"
end

puts "it's me outside the thread again!!"

sleep 1

puts "ready to be done with this script! BYEEEE"
```

unfortunately it did not, because the still-sleeping thread seemed abandoned when the script executed

output:
```
Hello, i am outside a thread
it's me outside the thread again!!
I am inside a thread! When will I execute?
ready to be done with this script! BYEEEE
```

there may be ways to get around this (maybe grab the appropriate thread pool and wait for it to complete at some point?) but i'm not sure how is best to do that

bensheldon:
Futures get created on [`Concurrent.global_io_executor`](https://github.com/ruby-concurrency/concurrent-ruby/blob/385156117c110808875e1e87d5267147f696f796/lib/concurrent-ruby/concurrent/future.rb#L137)

You can force a shutdown of that executor:

```ruby
executor = Concurrent.global_io_executor
executor.shutdown
executor.kill unless executor.wait_for_termination(30.seconds)
```

Another option:

```ruby
Timeout::timeout(30.seconds) do
  ObjectSpace.each_object(Concurrent::Future) { |f| f.value unless f.unscheduled? }
end
```

PR #2230, opened 2022-03-10
Only include clients with intakes in the sla_tracked_scope

bytheway875:
Join to intakes to exclude clients without current year intakes.

Wondering if there's a way to do this without joining to the full intake table when really I just need to know if an intake _exists_. Can anyone help me out with that?


alaroia-cfa:
Since the query starts on the `clients` table, I think it has to be a join onto intakes -- I think that's the only way in SQL to use data from the intakes table that I can think of.

I'd be curious to know if Postgres considers this query basically fast or basically slow. The lowest-effort way to do that is to get this merged and on demo, execute the query somehow (is it queried directly on the Hub Metrics page? a clients list page? I forget), then look in Datadog and see how long it took. :) Of course, that won't be a perfect representation of what happens in prod.

The most precise way is to call `Client.sla_tracked.count` in `rails c`, look at what SQL it printed out, then run that in a `rails dbconsole` with `explain analyze`, and put that into a Postgres query visualizer like https://explain.dalibo.com/ and/or look at it and think really hard.

My hunch is the JOIN won't be a huge pain.

bytheway875:
@alaroia-cfa does that mean i can merge this?

alaroia-cfa:
@bytheway875 I just right now saw your email asking if this was OK to merge -- answer is yes! Earlier today when I wrote a review, I left a comment and forgot to hit "Approve" -- that means it looks like I only just now wrote the comment. Oops.

PR #2235, opened 2022-03-14
Do not create a new client when one already exists

bytheway875:


app/forms/personal_info_form.rb
```diff
@@ -28,12 +28,9 @@ def normalize_phone_numbers
...
   def save
     state = ZipCodes.details(zip_code)[:state]
-    @intake.update(
-      attributes_for(:intake)
-        .except(:phone_number_confirmation)
-        .merge(client: Client.create!)
-        .merge(state_of_residence: state)
-    )
+    attributes = attributes_for(:intake).merge(state_of_residence: state)
+    attributes[:client] = Client.create! unless @intake.client.present?
```
alaroia-cfa, line 32: Looking at this, it makes me wonder if we should `Client.build` instead. Then ActiveRecord transactions will avoid creating an empty client.

Not at all a blocker.

alaroia-cfa (approved):
No blockers. One comment.

Also, a question: Does this mean the steps to reproduce are to go back in the intake flow and change your phone number? Can you validate that on demo, and leave a comment in the bug ticket?

app/forms/personal_info_form.rb
bytheway875: yes, good call. i was trying to remember how to do something like that. will update now

PR #2237, opened 2022-03-14
Remove fallback creation of intake

bytheway875:
Job count still has vestiges of a WIP intake creation from when it was the first page created.


tgrathwell:
@embarnard @bytheway875 this is a funny one! It seems harmless because we were already checking the client was authenticated. You probably already know that, but now Travis and Asheesh (we) know that too.

PR #2239, opened 2022-03-14
Fix timedout devise gap

bytheway875:
devise uses flash to store temporary state. this is why sometimes we see "true" on the page or, now that the background color of the hero has changed, a white gap at the top of the page.

Devise does not want to stop using flash to store this state even though people have told them to use session becuase this is weird since about 2012. To fix, we can just skip from outputting the value of the timedout flash onto the page.
Screen Shot 2022-03-14 at 11 55 41 AM
Screen Shot 2022-03-14 at 11 55 46 AM


tgrathwell:
it is perhaps the destiny of any Rails project that uses Devise to add this line eventually https://github.com/search?l=&q=timedout+filename%3Aapplication.html.erb&type=code

PR #2253, opened 2022-03-16
Qualifying relatives

bytheway875:


app/controllers/ctc/questions/dependents/child_qualifiers_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Ctc
   module Questions
     module Dependents
-      class ChildQualifiesController < BaseDependentController
+      class ChildQualifiersController < BaseDependentController
```
bytheway875, line 4: make this match the relative qualifiers naming

app/lib/efile/relationships.yml
```diff
@@ -3,66 +3,88 @@ relationships:
   - id: daughter
     relationship: qualifying_child
     irs_enum: DAUGHTER
+    skip_relative_household_test: true
```
bytheway875, line 6: i added this here because it kind of reflects the format that IRS provides this data to us in which is each relationship either needs this test or not. however, should i invert the logic? relative_household_test: false (i invert the logic in the relationship.rb that consumes this file.)

app/services/dependent_flow_service.rb
```diff
@@ -31,6 +31,21 @@ def show?
     when "Ctc::Questions::Dependents::ClaimChildAnywayController"
       eligibility = child_qualification(except: :claimable_test)
       dependent.cant_be_claimed_by_other_no? && eligibility.qualifies?
+    when "Ctc::Questions::Dependents::RelativeMemberOfHouseholdController"
+      return false if child_qualification.qualifies?
+
+      eligibility = relative_qualification(except: [:residence_test, :financial_support_test, :claimable_test])
+      eligibility.qualifies? && eligibility.requires_member_of_household_test?
+    when "Ctc::Questions::Dependents::RelativeFinancialSupportController"
+      return false if child_qualification.qualifies?
```
bytheway875, line 40: To see qualifying relative pages, you'll need to have failed the qualifying child test.

bytheway875: This is what ensures that the "other descendant of sibling" will see this page if they fail the age qualification.

spec/lib/efile/dependent_eligibility/base_spec.rb
```diff
@@ -15,6 +24,26 @@
                                                                                   )
     end
 
+    describe ".age" do
```
bytheway875, line 27: I just moved these tests to the base class.

app/lib/efile/relationships.yml
caitbot: `requires_relative_household_test: false` if youre always inverting it on consumption, I saw invert it here

spec/lib/efile/relationship_spec.rb
```diff
@@ -45,4 +45,8 @@
       end
     end
   end
+
+  describe "qualifying_relative_requires_member_of_household_test?" do
```
bytheway875, line 49: This branch still needs some testing. If you think its helpful to have this stuff, though, in order to get going on confirm_dependents, we can chunk it.

caitbot (approved):

PR #2254, opened 2022-03-16
adjust styling for selfie and photo id

mquadri23:


app/assets/stylesheets/_honeycrisp-compact.scss
```diff
@@ -1,6 +1,17 @@
 
 // likely needed in main repo
 .honeycrisp-compact {
+  .container{
```
mquadri23, line 4: rename to something like .verification-attempts--photo-container  and move for verification-attempst.scss , the .item class can stay the same because you're nesting it and it compiles to .verification-attempts--photo-container .item { 

This will prevent these styles from applying to ANY class .container in all of the hub

db/schema.rb
```diff
@@ -1567,7 +1567,7 @@
            SELECT DISTINCT tax_returns.client_id
              FROM (tax_returns
                JOIN intakes ON ((intakes.client_id = tax_returns.client_id)))
-            WHERE ((tax_returns.current_state)::text <> ALL ((ARRAY['intake_before_consent'::character varying, 'intake_in_progress'::character varying, 'intake_greeter_info_requested'::character varying, 'intake_needs_doc_help'::character varying, 'file_mailed'::character varying, 'file_accepted'::character varying, 'file_not_filing'::character varying, 'file_hold'::character varying, 'file_fraud_hold'::character varying])::text[]))
```
bytheway875, line 1570: do you think you can remove these changes from the commit? I'm still not sure why this happens... @alaroia-cfa does this happen when you use the UI to run migrations instead of rails db:migrate or vice versa 🤔

bytheway875 (approved):

bytheway875 (approved):

PR #2255, opened 2022-03-16
Update dependent logic for 2022

caitbot:


app/controllers/ctc/portal/dependents_controller.rb
```diff
@@ -18,7 +18,7 @@ def destroy
       action: 'removed',
       client: current_client
     )
-    if current_intake.dependents.filter(&:eligible_for_child_tax_credit_2020?).length.zero?
+    if current_intake.dependents.filter(&:eligible_for_child_tax_credit?).length.zero?
```
bytheway875, line 21: i feel like this will need the tax year?

Also, it might make more sense to instantiate the eligibility classes in this file? 🤔

app/lib/efile/dependent_eligibility/qualifying_child.rb
```diff
@@ -41,15 +41,17 @@ def qualified_student?
       end
 
       def over_qualifying_age_limit?
-        age > 18
-      end
+        return age >= 17 if tax_year == 2020
 
-      private
+        age >= 18
```
bytheway875, line 46: hmm... this isnt quite right. The qualifyng age limit for qualifying children is being less than 19.

The rules for CTC are more specific -- only a subset of qualifying children are qualified for getting the CTC.

I think we need a separate rules file for the CTC that defines a more strict rule for age instead of overwriting the qualifying child age rules.

bytheway875: you could also create _another_ method thats like age_qualifies_for_ctc? that defines this logic

bytheway875: It might seem like overkill, but I was imagining classes like

Efile::DependentEligibility::Ctc for each benefit

where self.rules looks like:

```
def self.rules
{ qualifying_child_test: :qualifying_child?,
  age_test: :under_ctc_age_test?,
  tin_test: :ssn_or_atin?
}

def qualifying_payment
  return 0 unless qualifies?
  $$$ if age_during(tax_year) < 6
  $$$ if age_during(tax_year) > 6
end
```

app/lib/efile/dependent_eligibility/qualifying_relative.rb
```diff
@@ -37,6 +37,13 @@ def requires_member_of_household_test?
 
       private
 
+      def filer_financially_supported?
+        # this is not defined or required on archived dependents
```
bytheway875, line 41: nitty, but because we changed the name of the method "this" is a little murky in terms of what it refers to.

app/services/dependent_flow_service.rb
```diff
@@ -42,6 +42,7 @@ def show?
       eligibility = relative_qualification(except: [:financial_support_test, :claimable_test])
       eligibility.qualifies?
     when "Ctc::Questions::Dependents::RelativeQualifiersController"
+      puts "hello"
```
bytheway875, line 45: woops, we left this puts in there

bytheway875 (approved):
So much deleted code. Awesome refactor. The only thing I noticed is we left a puts hello in the dependent_flow_service that we should probably take out before merging.

PR #2258, opened 2022-03-16
Fix the layout for service component in spanish by truncating content

bytheway875:
The layout for the service component is broken in spanish because the content is too long. Product has approved truncating some of the content so that they layout is not broken.


alaroia-cfa (approved):
This is a totally good PR.

Do we need to make the same change in Transifex so it "sticks"? I suspect yes, but don't know, so I'll just leave it as a question for you, not a blocker.

PR #2261, opened 2022-03-16
Add dash mask to dependent SSNs

bytheway875:
Dependents didn't get the same treatment as our other inputs.


caitbot (approved):
I <3 data-masks

PR #2263, opened 2022-03-17
Submission dependents 181592627

bytheway875:
Store dependents for each submission onto submission_dependents table along with some eligibility data


app/lib/efile/dependent_eligibility/eligibility.rb
```diff
@@ -0,0 +1,79 @@
...
+      def qualifying_child?
+        @qualifying_child ||= child_eligibility.qualifies?
+      end
+
+      def qualifying_relative?
+        @qualifying_relative ||= !child_eligibility.qualifies? && relative_eligibility.qualifies?
+      end
+
+      def qualifying_ctc?
+        @qualifying_ctc ||=ctc_eligibility.qualifies?
```
caitbot, line 35: super nit but add a space after `=` here and in `qualifying_eip3?`

app/models/dependent.rb
```diff
@@ -149,13 +149,6 @@ def eligible_for_eip1?(tax_year = TaxReturn.current_tax_year)
     child_qualifiers.qualifies? && child_qualifiers.age < 17 && [:ssn, :atin].include?(tin_type&.to_sym)
   end
 
-  def eligible_for_eip2?(tax_year = TaxReturn.current_tax_year)
```
caitbot, line 152: why remove this now?

app/models/tax_return.rb
```diff
@@ -102,7 +102,8 @@ def standard_deduction
   def outstanding_recovery_rebate_credit
     return nil unless intake.eip1_amount_received && intake.eip2_amount_received
 
-    [expected_recovery_rebate_credit_one - intake.eip1_amount_received, 0].max + [expected_recovery_rebate_credit_two - intake.eip2_amount_received, 0].max
+    benefits = Efile::BenefitsEligibility.new(tax_return: self, dependents: intake.dependents)
```
bytheway875, line 105: im tempted to move the calculations for outstanding credits and all that jazz into the benefitsEligibility calculation files as well... the reason for this is that when I look at the 1040 and 8812 files, there is a lot of calculation in place happening which feels ... not great. we have to keep that in sync between the pdf version, the xml version, and what we show to clients. when we do it in this file, we need to instantiate a new instance of the benefits eligibility for each method but if we do it in the benefitseligibility class, each place that does any benefits calculations can instantiate one object  and grab all the relevant data from it...

caitbot: But I would also vote for doing it in a follow up PR

spec/features/ctc/complete_intake_spec.rb
```diff
@@ -204,8 +204,8 @@
 
     # =========== RECOVERY REBATE CREDIT ===========
     expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.stimulus_payments.title'))
-    expect(page).to have_selector(".first-stimulus", text: "$2,900")
-    expect(page).to have_selector(".second-stimulus", text: "$1,800")
+    expect(page).to have_selector(".first-stimulus", text: "$0")
```
caitbot, line 207: this is just due to the tax year being 2021, yeah? basically, these will likely be removed with the new RRC/EIP pages?

spec/features/ctc/rrc_intake_questions_spec.rb
```diff
@@ -37,10 +37,10 @@
       expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.stimulus_two_received.title'))
       fill_in I18n.t('views.ctc.questions.stimulus_two_received.eip2_amount_received_label'), with: "600"
       click_on I18n.t('general.continue')
-      expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.stimulus_owed.title'))
+      expect(page).to have_selector("h1", text: "It looks like you’ve received the full amount of your first two stimulus payments.")
       expect(page).to have_text("EIP 1: $1,200")
       expect(page).to have_text("EIP 2: $600")
-      expect(page).to have_selector('.stimulus-owed', text: "$1,800")
+      # expect(page).to have_selector('.stimulus-owed', text: "$0")
```
caitbot, line 43: either remove this or uncomment it

spec/jobs/build_submission_bundle_job_spec.rb
```diff
@@ -41,6 +41,37 @@
...
+          described_class.perform_now(submission.id)
+        }.to change(EfileSubmissionDependent, :count).by 2
+      end
+
+      context "when objects already exist for some dependents" do
+        before do
+          EfileSubmissionDependent.create(dependent: submission.intake.dependents.first, efile_submission: submission)
+        end
+
+        it "does a thing" do
```
caitbot, line 66: hahahah update this its

caitbot (approved):
this looks good, left a handful of comments but none of them are for major changes or are blocking. ✨

app/controllers/ctc/portal/dependents_controller.rb
```diff
@@ -18,7 +18,8 @@ def destroy
       action: 'removed',
       client: current_client
     )
-    if current_intake.dependents.filter(&:eligible_for_child_tax_credit?).length.zero?
+    ctc_eligibility = Efile::DependentEligibility::ChildTaxCredit
```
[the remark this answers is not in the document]
bytheway875: yeah -- ill change it back to inline. the names of these classes are so long

app/lib/efile/benefits_eligibility.rb
```diff
@@ -0,0 +1,36 @@
+module Efile
+  class BenefitsEligibility
+    attr_accessor :year, :eligible_filer_count, :dependents
+    def initialize(tax_return:, dependents:)
+      @eligible_filer_count = tax_return.rrc_eligible_filer_count
```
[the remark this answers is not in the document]
bytheway875: that probably does make sense because we'd also be able to store the amounts there, just like we do for the dependents.

app/models/tax_return.rb
bytheway875: yes, that makes sense to me -- i still need to rebase this against the 1040 work that was being done at the same time and want to do that ASAP

spec/features/ctc/complete_intake_spec.rb
bytheway875: yep, i think all of these pages are going to be removed

PR #2267, opened 2022-03-18
Hub: use a static hub.js instead of a dynamic hub.js.erb

tgrathwell:
For whatever reason, compiling hub.js.erb seems to fail on CircleCI
now and again. This could be because `rails-erb-loader` and all
of the technology it requires is flaky under some conditions.

Since I don't know what conditions these are, maybe we could just
not do the ERB part for now?


alaroia-cfa (approved):
I recall co(?)authoring these files and think your change is a good one. Wanna merge it?

PR #2268, opened 2022-03-18
Hub: add 'last contacted' filter and two quick filter links

tgrathwell:


app/views/hub/clients/index.html.erb
```diff
@@ -4,8 +4,20 @@
 <% end %>
 <% content_for :card do %>
   <section class="slab clients-header">
-    <h1> <%= @page_title %> </h1>
     <div>
+      <h1> <%= @page_title %> </h1>
+      <div class="text--no-wrap quick-filters">
+        <label>Quick Filters:</label>
+        <% ClientSortable::QUICK_FILTERS.each do |(filter_value, filter_label)| %>
```
bytheway875, line 11: feels a little different than the pattern we normally use to have the labels inside ClientSortable service. i guess maybe we dont generally do things like that because we use translation files and we don't translate in the hub anymore. But it does feel like maybe there should be a helper method for the labels instead, idk.

app/views/hub/clients/index.html.erb
```diff
@@ -4,8 +4,20 @@
...
+          <% selected = filtering_only_by?(:last_contact, filter_value) %>
+          <%= link_to selected ? hub_clients_path(clear: true) : hub_clients_path(last_contact: filter_value), class: "button button--quick-filter  #{selected ? "selected" : ""}" do %>
+            <%= filter_label %>
+            <i class="clear-filter icon-close"></i>
```
bytheway875, line 15: My component-oriented mind makes me think this would make sense as like a partial where we pass in the data it needs, or maybe even to style the icon as an ::after element instead of having the css for the icon right here. Something feels weird about building such a complex little UI element right on the index page because then it's not really reusable as a UI element

app/controllers/hub/clients_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
       @page_title = I18n.t("hub.clients.index.title")
 
       # Compute the count of tax returns, up to a maximum amount. Postgres is slow at computing counts if they are very large.
-      tax_return_count = TaxReturn.where(client: filtered_clients.with_eager_loaded_associations.without_pagination).limit(MAX_COUNT + 1).size
```
tgrathwell, line 19: I would love to see the crash but that's just a curiosity. - AL

app/controllers/concerns/client_sortable.rb
```diff
@@ -81,11 +102,35 @@ def filters_from(source)
...
   end
 
+  def filtering_only_by?(filter_values)
+    @filters.select { |_k, v| v.present? } == filter_values.transform_values { |v| v.to_s }
+  end
+
   def search_and_sort_params
-    [:search, :status, :unassigned, :assigned_to_me, :flagged, :unemployment_income, :year, :vita_partners, :assigned_user_id, :language, :service_type, :greetable, :sla_breach_date, :used_navigator, :ctc_client]
+    [
+      :search,
```
tgrathwell, line 116: Wonder if it's possible to refactor this to share the list of keys with filtered_clients but obvi not a showstopper -al

tgrathwell (commented):
approved by al

app/lib/seeder.rb
```diff
@@ -0,0 +1,217 @@
...
+    )
+    TaxReturn.find_or_create_by(year: 2021, client: verifying_no_bypass_yet_intake.client).transition_to("intake_ready")
+    VerificationAttempt.find_or_create_by(client: verifying_no_bypass_yet_intake.client)
+
+    verifying_with_bypass_intake = find_or_create_intake_and_client(
+      Intake::CtcIntake,
+      primary_first_name: "VerifierTwo",
+      primary_last_name: "Smith",
+      primary_consented_to_service_at: DateTime.current,
+      visitor_id: "test_visitor_id_verifier1",
```
tgrathwell, line 194: gen

app/lib/seeder.rb
```diff
@@ -0,0 +1,217 @@
...
+    )
+    TaxReturn.find_or_create_by(year: 2021, client: verifying_with_bypass_intake.client).transition_to("intake_ready")
+    VerificationAttempt.find_or_initialize_by(client: verifying_with_bypass_intake.client).update(client_bypass_request: "I don't have an ID but I'd like to submit my taxes.")
+  end
+
+  def find_or_create_intake_and_client(intake_type, attributes)
+    attributes[:preferred_name] = attributes[:primary_first_name] if attributes[:preferred_name].blank?
+
+    finder_columns = [:primary_first_name, :primary_last_name, :preferred_name]
```
tgrathwell, line 203: maybe just first 2

PR #2270, opened 2022-03-21
Update 1040 for 2021

embarnard:


app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
```diff
@@ -42,33 +42,42 @@ def document
...
+
+            # Line 30: remaining amount of RRC they are claiming (as determined in flow). If they are not claiming anything, the Revenue Procedure is ambiguous as to whether this should be $0 or blank; but I recall last year having an issue with blank, so I would advocate $0. Rev Proc is clear IRS will not correct $0.
             xml.RecoveryRebateCreditAmt tax_return.claimed_recovery_rebate_credit # 30
-            xml.RefundableCreditsAmt tax_return.claimed_recovery_rebate_credit # 32
-            xml.TotalPaymentsAmt tax_return.claimed_recovery_rebate_credit # 33
-            xml.OverpaidAmt tax_return.claimed_recovery_rebate_credit # 34
-            xml.RefundAmt tax_return.claimed_recovery_rebate_credit # 35a
+
+            # Line 32, 33, 34, 35a: Line 28 + Line 30
+            total_payments = (claimed_child_tax_credit.to_i + tax_return.claimed_recovery_rebate_credit.to_i).to_s
```
bytheway875, line 61: i think this should be eip3 amount instead of claimed_recovery_rebate_credit -- they can't get eip1 and 2 on thisreturn.

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
```diff
@@ -42,33 +42,42 @@ def document
...
+            xml.RefundableCreditsAmt total_payments # 32
+            xml.TotalPaymentsAmt total_payments # 33
+            xml.OverpaidAmt total_payments # 34
```
bytheway875, line 64: it doesnt seem right to me that overpaid amount would be filled out like this. was this in the list of attributes gabe said to fill in?

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
```diff
@@ -42,33 +42,42 @@ def document
...
+            xml.TotDedCharitableContriAmt tax_return.standard_deduction # 12c
+            xml.TotalDeductionsAmt tax_return.standard_deduction # 14
             xml.TaxableIncomeAmt 0 # 15
+
+            # Line 28: remaining amount of CTC they are claiming (as determined in flow and listed on 8812 14i
+            claimed_child_tax_credit = "0" # TODO: eventually replace this amount with the adv_ctc_amount_received when this is set in the RRC stories
+            xml.RefundableCTCOrACTCAmt claimed_child_tax_credit # 28
+            # maybe add this to be part of the stories for RRC
+
+            # Line 30: remaining amount of RRC they are claiming (as determined in flow). If they are not claiming anything, the Revenue Procedure is ambiguous as to whether this should be $0 or blank; but I recall last year having an issue with blank, so I would advocate $0. Rev Proc is clear IRS will not correct $0.
```
bytheway875, line 57: and this is limited to RRC 3, correct? because 1 and 2 are not claimable in 2021 tax year.

bytheway875:
hmm im not sure why some of the vendor files are showing up into version control. can you remove them?

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
```diff
@@ -42,33 +42,42 @@ def document
...
           }
         end.doc
       end
 
       def build
-        unless ENV['TEST_SCHEMA_VALIDITY_ONLY'] == 'true'
-          raise NotImplementedError, "SubmissionBuilder::TY2021::AdvCtcIrs1040 is for testing purposes only and does not currently conform to the 2021 revenue procedure."
-        end
-
+        # unless ENV['TEST_SCHEMA_VALIDITY_ONLY'] == 'true'
```
bytheway875, line 78: i think you can fully remove this commented out code.

bytheway875: hmm actually you know what... we've still got a lot of placeholders here. i think what we actually need to do is add an around filter to the tests to set the env variable and revert to having this warning show up if you try to run the code directly. that way we don't forget that there are things that need to be updated in here.

bytheway875 (changes requested):
def remove the extra vendor files and set up the tests so that we can have the error raise on "real" builds since the file isn't quite ready for prime time yet.

Can you also make sure that there is a follow up ticket for making sure the right amounts get calculated in this file?

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
```diff
@@ -42,33 +42,42 @@ def document
               dependent_xml(xml, dependent)
```
bytheway875, line 42: The name of this file (adv_ctc_irs_1040.rb) reflected the fact that last year's revenue procedure was to claim advanced ctc money. I think you can change this to under_standard_deduction_irs_1040.rb or lapsed_filer_irs_1040.rb and it would better reflect what this is for this year.

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
bytheway875: nevermind... this is what we did last year so i guess this likely is correct even though i dont quite get it. i guess because its a rebate but idk.

bytheway875:
also, i needed to rebase this with the work i just pushed to main so make sure you pull before you make any additional changes

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
embarnard: I was using `claimed_recovery_rebate_credit` here because I think `outstanding_recovery_rebate_credit` (which is what determines claimed_recovery_rebate_credit) will get updated in a future RRC story to only make the calculations with EIP-3 [#181452309](https://www.pivotaltracker.com/story/show/181452309)

app/lib/submission_builder/ty_2021/adv_ctc_irs1040.rb
embarnard: Renamed! :)

PR #2272, opened 2022-03-22
Obfuscation of fraud rules

bytheway875:
If not much makes sense here without additional context, that's a good thing.

I sent Asheesh the .yml file that we'd seed into the db -- I guess we'd need to host it somewhere like we do the schemas. 🤔  so that we could import it into different environments. (Ask Asheesh to share it with you if you're interested in matching the code to the rules while I'm out.)

This creates a pattern for each of the existing rules as well as some of the upcoming ones (all of them except for the really custom patternized rules). 

This does NOT hook up the results of the test to the FraudIndicatorService, or provide points or anything like that yet... just trying to nail down the obfuscation patterns.

It's executing a lot of queries based on what's stored in the database, but I'm pretty sure there's not anywhere that we're passing in something that could attempt to execute as pure SQL... I could be wrong, though, and that's what I'm looking for some feedback on.

Also I couldn't figure out how to create a class in the test file to get something that responded to "list" without just opening a class and adding it to the Global Namespace... if there are any recommendations there, let me know


app/lib/efile/timezone.rb
```diff
@@ -0,0 +1,21 @@
...
+        overrides = IceNine.deep_freeze!(YAML.load_file("lib/timezone_overrides.yml"))
+        overrides + us_timezones
+      end
+    end
+
+    def self.all
+      @@timezones
+    end
+
+    def self.list
```
bytheway875, line 17: We'll implement a list method on all of the classes that get used as list_class_names in a fraud rule

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+#  name                 :string
+#  points               :integer
+#  reference            :string
+#  source_table_name    :string
+#  threshold            :decimal(, )
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+class FraudIndicator < ApplicationRecord
+  validates :indicator_type, presence: true, inclusion: { in: FraudIndicator.instance_methods - ApplicationRecord.instance_methods }
```
bytheway875, line 20: I'm planning to add tests for some of these validations, because I think having the validations be accurate per test type can save us from mistakes that would ultimately make the code not work later.

bytheway875:
@alaroia-cfa I think you have the most context for being able to provide feedback on this PR. If it's something you have time to do this week while I'm gone, great. If not, can you put some time on my cal for Monday morning central time to pair review + discuss? I want to make sure that I'm able to make revisions and/or think through alternate approaches next week.

alaroia-cfa:
@bytheway875 Calendar'd so I don't forget. I may be able to review before then, too.

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  private
+
+  def scoped_query(references)
+    self_reference = source_table_name.downcase == reference.underscore
+    scope = self_reference ? { id: references[reference].id } : { reference => references[reference] }
+    source_table_name.constantize.where(scope)
+  end
+
+  def comparison_list
```
alaroia-cfa, line 99: safelist and denylist methods

app/lib/efile/timezone.rb
```diff
@@ -0,0 +1,21 @@
+module Efile
+  class Timezone
```
alaroia-cfa, line 2: -> activerecord

spec/models/fraud_indicator_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+#  reference            :string
+#  source_table_name    :string
+#  threshold            :decimal(, )
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+require "rails_helper"
+
+describe FraudIndicator do
+  class FraudIndicator::Bunny
```
alaroia-cfa, line 22:
```
  before do
    stub_const("FraudIndicator::Bunny", Class.new do
      def self.list
        ["Bunny", "Peter", "Rabbit"]
      end
    end)
  end
```

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+#
+class FraudIndicator < ApplicationRecord
+  validates :indicator_type, presence: true, inclusion: { in: FraudIndicator.instance_methods - ApplicationRecord.instance_methods }
+  validates :points, presence: true
+  validates :list_table_name, presence: true, if: -> { indicator_type.in? ["safelist", "denylist"] }
+  validates :reference, presence: true, inclusion: { in: ["client", "intake", "efile_submission", "bank_account"] }
+  validates :name, presence: true
+  validates :source_table_name, :list_table_name, class_name: true
+  validates :threshold, numericality: true, if: -> { indicator_type.in? ["average_threshold", "duplicates"] }
+  validates :indicator_attributes, length: { is: 1 }, if: -> { indicator_type.in? ["average_threshold", "safelist", "denylist", "missing"] }
```
alaroia-cfa, line 27: `not_in_safelist`
`in_denylist`

maybe

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+class FraudIndicator < ApplicationRecord
+  validates :indicator_type, presence: true, inclusion: { in: FraudIndicator.instance_methods - ApplicationRecord.instance_methods }
+  validates :points, presence: true
+  validates :list_table_name, presence: true, if: -> { indicator_type.in? ["safelist", "denylist"] }
+  validates :reference, presence: true, inclusion: { in: ["client", "intake", "efile_submission", "bank_account"] }
+  validates :name, presence: true
+  validates :source_table_name, :list_table_name, class_name: true
```
alaroia-cfa, line 25: ? `source_model_name`

`source` of the bad data

alaroia-cfa: Maybe `query_model_name`

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+#  reference            :string
+#  source_table_name    :string
+#  threshold            :decimal(, )
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+class FraudIndicator < ApplicationRecord
+  validates :indicator_type, presence: true, inclusion: { in: FraudIndicator.instance_methods - ApplicationRecord.instance_methods }
+  validates :points, presence: true
+  validates :list_table_name, presence: true, if: -> { indicator_type.in? ["safelist", "denylist"] }
```
alaroia-cfa, line 22: `list_model_name` or `list_class_name`

app/models/fraud_indicator.rb
```diff
@@ -0,0 +1,102 @@
...
+  validates :reference, presence: true, inclusion: { in: ["client", "intake", "efile_submission", "bank_account"] }
+  validates :name, presence: true
+  validates :source_table_name, :list_table_name, class_name: true
+  validates :threshold, numericality: true, if: -> { indicator_type.in? ["average_threshold", "duplicates"] }
+  validates :indicator_attributes, length: { is: 1 }, if: -> { indicator_type.in? ["average_threshold", "safelist", "denylist", "missing"] }
+  validates :indicator_attributes, length: { is: 2 }, if: -> { indicator_type.in? ["particular_value"] }
+  validates :indicator_attributes, length: { minimum: 1 }, if: -> { indicator_type.in? ["duplicates"] }
+  validates :multiplier, presence: true, if: -> { indicator_type.in? ["duplicates"] }
+
+  default_scope { where.not(active_at: nil) }
```
alaroia-cfa, line 32: `activated_at`?

db/migrate/20220321205520_create_fraud_indicators.rb
```diff
@@ -0,0 +1,17 @@
...
+    create_table :fraud_indicators do |t|
+      t.string :name
+      t.string :indicator_type
+      t.string :source_table_name
+      t.decimal :threshold
+      t.string :reference
+      t.string :list_table_name
+      t.string :indicator_attributes, array: true, default: []
+      t.integer :points
+      t.decimal :multiplier
```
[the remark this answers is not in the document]
alaroia-cfa: `duplicate_multiplier`

app/models/fraud_indicator.rb
alaroia-cfa: `average_below` instead of `average_threshold` maybe

app/services/fraud_indicator_service.rb
```diff
@@ -24,7 +24,7 @@ def fraud_suspected?
   def acceptable_timezone_strings
     @acceptable_timezone_strings ||= begin
       us_timezones = ActiveSupport::TimeZone.us_zones.map { |tz| [tz.name, tz.tzinfo.name].uniq }.flatten.freeze
-      overrides = IceNine.deep_freeze!(YAML.load_file("lib/timezone_overrides.yml"))
+      overrides = ["America/Indianapolis"]
```
bytheway875, line 27: removed the overrides file in anticipation of using the timezone table -- left this implementation here so that i dont break tests.

app/validators/class_name_validator.rb
```diff
@@ -0,0 +1,10 @@
+class ClassNameValidator < ActiveModel::EachValidator
+  def validate_each(record, attr_name, value)
+    return false if value.nil?
+    begin
+      Object.class_eval(value)
```
bytheway875, line 5: I thought about updating this to be more specific to model names, but the problem is that `ApplicationRecord.descendants` only returns a list of classes that have already been loaded, which would mean that the validator would need to eager load the rest of the models in non-production environments in order for it to work properly. This seems to be a simpler way to ensure that at least what's provided is indeed a class inside the application, even though it's less specific.

bytheway875:
@alaroia-cfa I build off of this in a more recent PR. If you want to take a look at this and see if it's good to merge as is right now, that may make the following PR smaller and easier to grep since it will show updates from this original implementation.

This still has the true/false responses, which are updated with the tuple-ish responses in the following PR

PR #2273, opened 2022-03-22
Make client_id index on intakes unique

jenny-heath:


db/schema.rb
```diff
@@ -1584,7 +1584,7 @@
            SELECT DISTINCT tax_returns.client_id
              FROM (tax_returns
                JOIN intakes ON ((intakes.client_id = tax_returns.client_id)))
-            WHERE ((tax_returns.current_state)::text <> ALL ((ARRAY['intake_before_consent'::character varying, 'intake_in_progress'::character varying, 'intake_greeter_info_requested'::character varying, 'intake_needs_doc_help'::character varying, 'file_mailed'::character varying, 'file_accepted'::character varying, 'file_not_filing'::character varying, 'file_hold'::character varying, 'file_fraud_hold'::character varying])::text[]))
+            WHERE ((tax_returns.current_state)::text <> ALL (ARRAY[('intake_before_consent'::character varying)::text, ('intake_in_progress'::character varying)::text, ('intake_greeter_info_requested'::character varying)::text, ('intake_needs_doc_help'::character varying)::text, ('file_mailed'::character varying)::text, ('file_accepted'::character varying)::text, ('file_not_filing'::character varying)::text, ('file_hold'::character varying)::text, ('file_fraud_hold'::character varying)::text]))
```
jenny-heath, line 1587: no idea why this changed with the migration. it doesn't look related to this PR but also doesn't look like a simple formatting thing

lib/tasks/dedup_clients_with_multiple_intakes.rake
```diff
@@ -0,0 +1,12 @@
+namespace :clients_with_many_intakes do
```
jenny-heath, line 1: this should only have to be run in demo

PR #2274, opened 2022-03-22
Refactor RRC calculations into BenefitsEligibility

caitbot:
I began working on the EIP changes and realized a refactor to align with the new pattern of benefit calculations would be beneficial before continuing down this path. 

What this PR does:
Moves `outstanding_recovery_rebate_credit`, `claimed_recovery_rebate_credit`, and `rrc_eligible_filer_count` from `TaxReturn` into `BenefitsEligibility`

Updates `outstanding_recovery_rebate_credit` to depend on the tax year. If 2020, it is EIP1 + EIP2, if 2021, it is EIP3.

Most of this was moving the tests.

Since I changed the RRC logic but didn't update the UI, it made some of the tests into nonsense (the flow is incorrect right now) and so I deleted them. They will be made better in subsequent PRs which contain the new RRC flow. I didn't feel so bad doing this since no one can see the RRC flow now.


spec/lib/efile/benefits_eligibility_spec.rb
```diff
@@ -107,7 +109,181 @@
       it "calculates amount" do
         # 1 qualified child under age 6 @ 3600 ea
         # 1 qualified child over 6 @ 3000 each
-        expect(Efile::BenefitsEligibility.new(tax_return: intake.tax_returns.last, dependents: intake.dependents).ctc_amount).to eq 6600
+        expect(subject.ctc_amount).to eq 6600
```
caitbot, line 112: this file is _mostly_ just specs moved over from tax return and updated a tiny bit to fit the new params

app/lib/efile/benefits_eligibility.rb
```diff
@@ -53,5 +66,35 @@ def odc_amount
...
+    def claimed_recovery_rebate_credit
+      return 0 if intake.claim_owed_stimulus_money_no?
+
+      outstanding_recovery_rebate_credit
+    end
+
+    private
+
+    def rrc_eligible_filer_count
+      return intake.primary_tin_type == "ssn" ? 1 : 0 if tax_return.filing_status_single?
```
embarnard, line 91: Should we also check for the presence of an ssn in addition to tin_type; Idk are there scenarios where someone has a tin_type but no ssn?

app/models/efile_submission_dependent.rb
```diff
@@ -21,7 +21,7 @@ class EfileSubmissionDependent < ApplicationRecord
   belongs_to :dependent
   belongs_to :efile_submission
 
-  delegate :age, :first_name, :last_name, :ssn, :irs_relationship_enum, :birth_date, to: :dependent
+  delegate :age, :first_name, :last_name, :ssn, :irs_relationship_enum, :birth_date, :tin_type_ssn?, :tin_type_atin?, to: :dependent
```
embarnard, line 24: nice

app/views/hub/clients/show.html.erb
```diff
@@ -149,11 +149,12 @@
               <% end %>
               <% unless @client.intake.drop_off? %>
                 <% tax_return = @client.tax_returns.find_by(year: TaxReturn.current_tax_year) %>
+                <% benefits = Efile::BenefitsEligibility.new(tax_return: tax_return, dependents: @client.intake.dependents) if tax_return %>
                 <div class="field-display">
                   <span class="form-question"> Outstanding RRC / Claimed RRC: </span>
-                  <span class="label-value"><%= "$#{tax_return&.outstanding_recovery_rebate_credit}" if tax_return&.outstanding_recovery_rebate_credit %></span>
+                  <span class="label-value"><%= "$#{benefits.outstanding_recovery_rebate_credit}" if benefits&.outstanding_recovery_rebate_credit %></span>
```
embarnard, line 155: Should we show a 0 if benefits&.outstanding_recovery_rebate_credit/claimed_recovery_rebate_credit == false? Or protect it from the showing the span or "/" at all?

spec/controllers/ctc/questions/stimulus_payments_controller_spec.rb
```diff
@@ -19,7 +19,7 @@
       intake.reload
       expect(intake).to be_eip1_entry_method_calculated_amount
       expect(intake).to be_eip2_entry_method_calculated_amount
-      expect(response).to redirect_to questions_stimulus_received_path
+      # TODO: correct redirect to next path for new RRC flow
```
embarnard, line 22: Hmm like we will add this back in with the correct redirect in the future?

embarnard (approved):
Looks good! Left a few comments but no blocking change requests

app/views/hub/clients/show.html.erb
caitbot: I was thinking the same thing, thats probably a reasonable thing to do. I just wanted to make it so that there weren't any changes. I'm going to re-examine this in the UI ticket (which calls for making sure this display makes sense)

app/lib/efile/benefits_eligibility.rb
caitbot: I'm pretty sure we validate the presence of ssn so this shouldn't be an issue/this is the same logic as before and it hasn't been an issue so I'm going to leave it as it is, but good call out.

PR #2275, opened 2022-03-22
Update client success abilities so that they can view, edit, and help clients without an organization assigned

bytheway875:
This was reported as an issue with archived clients, but I actually think that customer success has not been able to view any clients who weren't assigned to an organization...

In order to allow access to clients with nil organizations (as well as the rest of their cascading stuff), I needed to add and then taketh away certain features from the CS role. It's not ideal, but I couldn't figure out another way to allow nil vita partners...


app/lib/ability.rb
```diff
@@ -42,7 +56,7 @@ def initialize(user)
 
     can :manage, EfileSubmission, tax_return: { client: { vita_partner: accessible_groups } }
 
-    cannot :index, EfileSubmission unless user.role_type == AdminRole::TYPE
```
bytheway875, line 45: This feels like something client success should be able to see but idk

bytheway875:
@alaroia-cfa @caitbot This PR addresses the open problem where Client Success can't view certain clients. If this looks good, feel free to merge while I'm gone or take over the work. Or leave it for Monday, whatever makes sense!

app/lib/ability.rb
caitbot: That seems reasonable to me!

caitbot (approved):
This seems good to me and should solve the problem, I see no problem with this

PR #2277, opened 2022-03-23
Try hobby-basic Heroku database tier in review apps

alaroia-cfa:
This may address an issue where hobby-dev doesn't allow enough rows.

### DIFF
--- a/app.json
+++ b/app.json
@@ -48,5 +48,10 @@
   },
   "addons": [
     "heroku-postgresql"
-  ]
+  ],
+  "environments": {
+    "review": {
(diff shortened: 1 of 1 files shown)


alaroia-cfa:
https://stackoverflow.com/questions/60724956/specify-postgres-database-plan-in-heroku-review-apps says we might need to contact support.

I found that we are not getting the "hobby-basic" databases, so I contacted support.

tgrathwell:
i dunno if heroku support ever got back to us but i'm closing this assuming it doesn't solve the problem (but maybe it does??)

PR #2279, opened 2022-03-23
Updates from design-eng pairing

caitbot:
**What we achieved:**
- Added help text to `SpouseIssuedIdentityPin`
- Updated the icon on the `MarriedController` (from confused and distant little peoples to close and cohabitated little peoples)
- Updated the size of the labels on the`ArpPayments` page (they were giant because the form was missing a class that makes them small)
- Update `LifeSituations` and `SpouseLifeSituations` to require the client to select something (anything!). Note that the design of the error display is not optimal, but we did what we could with the time we had. We will iterate on this at future sessions or in future stories.
image
image
image
image

### DIFF
--- a/app/forms/life_situations_form.rb
+++ b/app/forms/life_situations_form.rb
@@ -1,7 +1,16 @@
 class LifeSituationsForm < QuestionsForm
   set_attributes_for :intake, :was_full_time_student, :was_on_visa, :had_disability, :was_blind
+  set_attributes_for :confirmation, :no_life_situations_apply
+  validates :no_life_situations_apply, at_least_one_or_none_of_the_above_selected: true
 
   def save
     @intake.update(attributes_for(:intake))
+  end
+
+  def at_least_one_selected
+    was_full_time_student == "yes" ||
+      was_on_visa == "yes" ||
+      had_disability == "yes" ||
+      was_blind == "yes"
   end
 end
--- a/app/forms/spouse_life_situations_form.rb
+++ b/app/forms/spouse_life_situations_form.rb
@@ -1,7 +1,16 @@
 class SpouseLifeSituationsForm < QuestionsForm
   set_attributes_for :intake, :spouse_was_full_time_student, :spouse_was_on_visa, :spouse_had_disability, :spouse_was_blind
+  set_attributes_for :confirmation, :no_life_situations_apply
+  validates :no_life_situations_apply, at_least_one_or_none_of_the_above_selected: true
 
   def save
     @intake.update(attributes_for(:intake))
+  end
+
+  def at_least_one_selected
+    spouse_was_full_time_student == "yes" ||
+      spouse_was_on_visa == "yes" ||
+      spouse_had_disability == "yes" ||
+      spouse_was_blind == "yes"
   end
 end
--- a/app/controllers/questions/married_controller.rb
+++ b/app/controllers/questions/married_controller.rb
@@ -9,7 +9,7 @@ module Questions
     end
 
     def illustration_path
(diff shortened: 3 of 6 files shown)


app/forms/life_situations_form.rb
```diff
@@ -1,7 +1,20 @@
...
+
+  private
+
+  def at_least_one_option_selected
+    chose_one = was_full_time_student == "yes" ||
+      was_on_visa == "yes" ||
+      had_disability == "yes" ||
+      was_blind == "yes" ||
+      no_life_situations_apply == "yes"
+    errors.add(:no_life_situations_apply, I18n.t("forms.errors.at_least_one_option")) unless chose_one
```
tgrathwell, line 18: we've tried to standardize on a patern for this where you implement `at_least_one_selected` to be true/false and the `errors.add` is done by a `at_least_one_or_none_of_the_above_selected` validator

see https://github.com/codeforamerica/vita-min/blob/main/app/forms/ctc/dependents/relative_qualifiers_form.rb#L8 https://github.com/codeforamerica/vita-min/blob/main/app/validators/at_least_one_or_none_of_the_above_selected_validator.rb

caitbot: 😭 beautiful, thank you for this

db/schema.rb
```diff
@@ -1584,7 +1584,7 @@
            SELECT DISTINCT tax_returns.client_id
              FROM (tax_returns
                JOIN intakes ON ((intakes.client_id = tax_returns.client_id)))
-            WHERE ((tax_returns.current_state)::text <> ALL ((ARRAY['intake_before_consent'::character varying, 'intake_in_progress'::character varying, 'intake_greeter_info_requested'::character varying, 'intake_needs_doc_help'::character varying, 'file_mailed'::character varying, 'file_accepted'::character varying, 'file_not_filing'::character varying, 'file_hold'::character varying, 'file_fraud_hold'::character varying])::text[]))
+            WHERE ((tax_returns.current_state)::text <> ALL (ARRAY[('intake_before_consent'::character varying)::text, ('intake_in_progress'::character varying)::text, ('intake_greeter_info_requested'::character varying)::text, ('intake_needs_doc_help'::character varying)::text, ('file_mailed'::character varying)::text, ('file_accepted'::character varying)::text, ('file_not_filing'::character varying)::text, ('file_hold'::character varying)::text, ('file_fraud_hold'::character varying)::text]))
```
embarnard, line 1587: Hmmm was this change intentional?

app/forms/spouse_life_situations_form.rb
```diff
@@ -1,7 +1,18 @@
 class SpouseLifeSituationsForm < QuestionsForm
   set_attributes_for :intake, :spouse_was_full_time_student, :spouse_was_on_visa, :spouse_had_disability, :spouse_was_blind
+  set_attributes_for :misc, :no_life_situations_apply
+  validate :at_least_one_option_selected
 
   def save
     @intake.update(attributes_for(:intake))
   end
+
```
embarnard, line 9: Probably want to do Travis' suggested change here as well

embarnard (approved):
Looks good! Confused what the schema change was for in design facing terms and probably want to address Travis' comment but otherwise looks good to go!

db/schema.rb
caitbot: oops no I will remove

app/forms/spouse_life_situations_form.rb
caitbot: Yes def!

PR #2280, opened 2022-03-23
Remove commented-out line in Devise config

alaroia-cfa:
Devise::SecretKeyFinder looks for the Rails secret_key_base, so we still have a key.


tgrathwell:
🤷  the comment above looks a little weird without this line, could also consider replacing with something less obvious
`config.secret_key = 'not-a-real-key'` or deleting the rest of the big comment

but either way it is very low steaks

tgrathwell (approved):

alaroia-cfa:
@tgrathwell Good call! I've updated the comment. When it goes green, I'll merge.

PR #2283, opened 2022-03-24
Remove StimulusOne and StimulusTwo, add StimulusThree

caitbot:
I am chunking the work on RRC by what is described in the stories, but I do not think it makes sense to acceptance test until all three stories are finished. My intention is to merge this to main and have acceptance testing done there, but it should not block releases to production as it is not visible to clients.

That being said...
**This PR aims to:**
- Remove the `StimulusOne`, `StimulusTwo`, `StimulusOneReceived`, `StimulusTwoReceived` pages and their corresponding forms and controllers.

- Creates a `StimulusThree` page, controller and form, which asks the client to enter the amount they received for their third stimulus payment. It is saved on intake. This page looks most like the old `StimulusNumberReceived` pages but the designs called for an update to the naming scheme.

- Remove the use of EIP1 and EIP2 on the `StimulusReceived` page, this page parrots back to the client what _we understand_ them to have received and asks them to confirm. Now, we are only confirming EIP3.

- Updates the client page to show either EIP1 and EIP2 OR EIP3, depending on what values are present. 

- `xdescribe` the RRC feature spec file, because at this point, the RRC flow for CTC is nonsensical (and non-reachable, with the initial question commented out, there is no path to it). Fear not, feature tests are coming -- but I won't add them until I've got all the page updates in place. 🙃 

Stimulus Three page
image

Stimulus Received page
image


app/views/ctc/questions/stimulus_received/edit.html.erb
```diff
@@ -8,10 +8,9 @@
   <div class="review-box spacing-below-60">
     <div class="review-box__title spacing-below-15">
       <h2><%= t("views.ctc.questions.stimulus_received.amount_received") %></h2>
-      <%= link_to t("general.edit").downcase, questions_stimulus_one_path, class: "review-box__edit" %>
+      <%= link_to t("general.edit").downcase, questions_stimulus_three_path, class: "review-box__edit" %>
     </div>
-    <p class="spacing-below-15"><%= t("views.ctc.questions.stimulus_received.eip_one") %>: <%= calculated_or_provided_dollar_amount(current_intake.eip1_amount_received) %></p>
-    <p class="spacing-below-0"><%= t("views.ctc.questions.stimulus_received.eip_two") %>: <%= calculated_or_provided_dollar_amount(current_intake.eip2_amount_received) %></p>
+    <p class="spacing-below-0"><%= t("views.ctc.questions.stimulus_received.eip_three") %>: <%= number_to_currency(current_intake.eip3_amount_received) %></p>
```
embarnard, line 13: should we round this number and use `calculated_or_provided_dollar_amount`? otherwise includes "$.00"

app/controllers/ctc/questions/stimulus_three_controller.rb
```diff
@@ -1,16 +1,10 @@
 module Ctc
   module Questions
-    class StimulusOneController < QuestionsController
+    class StimulusThreeController < QuestionsController
       include AuthenticatedCtcClientConcern
 
       layout "intake"
```
embarnard, line 7: I guess we'll have to remember to add a skip function later--- or could we do it depending on if they set an amount on the stimulus-payments page?

embarnard: but I guess we dont have the stimulus payments page adjusted for EIP-3

embarnard (approved):
looks good!

app/views/ctc/questions/stimulus_received/edit.html.erb
caitbot: ahh yes, I wasn't sure if using that method was necessary, but I missed that point. I'll put it back.

app/controllers/ctc/questions/stimulus_three_controller.rb
caitbot: yeah exactly, i figured when i add the feature specs back, this will get sorted pretty easily (doing that now)

PR #2284, opened 2022-03-24
CTC AdvCTC intake flow: create /adv-ctc and /adv-ctc-received pages

embarnard:


embarnard:
This PR:
* Adds the `/advance-ctc page`, corresponding controller, form and tests
* Adds the  `/advance-ctc-received` page, corresponding controller and test
* Adds the `advance_ctc_entry_method` column to intake
* Edits feature test `complete_intake_spec`

`Advance CTC page`
Screen Shot 2022-03-29 at 1 58 14 PM

`Advance CTC Received page`
Screen Shot 2022-03-29 at 1 58 52 PM

app/controllers/ctc/questions/advance_ctc_received_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        @ctc_owed = benefits.outstanding_ctc_amount
+        super
+      end
+
+      private
+
+      def illustration_path; end
+
+      def initialized_edit_form
+        nil
```
caitbot, line 26: Why do we need to override this?

caitbot: Is it possible that we could do 
```
      def form_class
        NullForm
      end
```
with the same results?

Thats more of a pattern that we have in the codebase already

app/forms/ctc/advance_ctc_form.rb
```diff
@@ -0,0 +1,28 @@
+module Ctc
+  class AdvanceCtcForm < QuestionsForm
+    set_attributes_for(
+      :no_model,
+      :advance_ctc_received_choice,
+      )
+
+    def save
```
caitbot, line 8: hahah I just did something very similar for RRC, I opted for a case but this is a-ok too!

caitbot (approved):
Looks good to me!

PR #2286, opened 2022-03-24
Flag clients if they do an incoming interaction (email, text, portal messages, or docs)

jenny-heath:
This is a partial revert of 5ad7bfd64da6f3dcb1e465ecfb9ba6310fd3bdf2
but only ever sets the flag on, it doesn't turn the flag off for
outgoing interactions

It also doesn't set the flag when intake is completed


alaroia-cfa:
@jenny-heath @tgrathwell I can merge this now so that Yvonne can acceptance-test on Demo.

PR #2295, opened 2022-03-31
CTC AdvCTC intake flow: create /advance-ctc-amount

embarnard:


embarnard:
This PR aims to create an `advance-ctc-amount` page that allows the client to manually enter the amount of Advance CTC they received

Screen Shot 2022-03-31 at 5 10 52 PM

app/views/ctc/questions/stimulus_three/edit.html.erb
```diff
@@ -12,7 +12,7 @@
     <%= t("views.ctc.questions.stimulus_three.reveal_content") %>
   <% end %>
 
-  <%= form_with model: @form, url: questions_stimulus_three_path, local: true, method: "put", builder: VitaMinFormBuilder, class: "stimulus-received-form" do |f| %>
```
caitbot, line 15: it looks like we removed the styling sheet for this class but I would bet this class is still being used on the `third-stimulus` page. Mind confirming if we should get rid of the spreadsheet? If it isn't doing anything then remove the classes from `third-stimulus` as well.

caitbot (approved):
Sweet this looks great, it's worth checking if the stimulus-received class is still in use on the `third-stimulus` page, but other than that, this is good to go

app/views/ctc/questions/stimulus_three/edit.html.erb
embarnard: Oooo good point! I also found the styling class used on the `prior_tax_year_agi` and `spouse_prior_tax_year_agi` pages so I took it out and put it in the `_form-cards.scss` file instead and changed the styling class name to `amount-collection` and `amount-collection-form` instead

PR #2296, opened 2022-04-01
Add shadow ban capability, create new submissions on every resubmission, and break preparing and bundling into two states

bytheway875:
This replaces the new fraud scoring and no longer leverages the fraud service in EfileSubmissionStateMachine

However, the verification attempts still show the list of broken rules based on the old fraud service AND the fraud_hold status on the client's efile submission list is not yet hooked up to show the fraud indicators based on the fraud score object, so nothing will show there. I still need to rip out the Fraud Indicator Service and swap implementations over to using Fraud::Score.

I wish that I had pulled the verification attempt changes into a separate PR so that this one wasn't so large, but it is currently included as part of this PR.

### DIFF
--- a/app/jobs/build_submission_bundle_job.rb
+++ b/app/jobs/build_submission_bundle_job.rb
@@ -1,7 +1,6 @@
 class BuildSubmissionBundleJob < ApplicationJob
   def perform(submission_id)
     submission = EfileSubmission.includes(:intake, :qualifying_dependents, :address, :tax_return, client: :efile_security_informations).find(submission_id)
-    persist_dependent_records(submission)
 
     address_creation = submission.generate_irs_address
     unless address_creation.valid?
@@ -9,6 +8,7 @@ class BuildSubmissionBundleJob < ApplicationJob
       return
     end
 
+    # TODO: Add IRS submission id generation here.
     begin
       submission.generate_form_1040_pdf
     rescue StandardError => e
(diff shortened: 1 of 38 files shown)


app/jobs/build_submission_bundle_job.rb
```diff
@@ -1,14 +1,14 @@
 class BuildSubmissionBundleJob < ApplicationJob
   def perform(submission_id)
     submission = EfileSubmission.includes(:intake, :qualifying_dependents, :address, :tax_return, client: :efile_security_informations).find(submission_id)
-    persist_dependent_records(submission)
```
bytheway875, line 4: Creating the dependents has moved into the 'preparing' step in the efile submission state machine.

app/jobs/deny_restricted_verification_attempt_job.rb
```diff
@@ -0,0 +1,7 @@
+class DenyRestrictedVerificationAttemptJob < ApplicationJob
+  def perform(verification_attempt)
+    return unless verification_attempt.current_state == "restricted"
```
bytheway875, line 3: By delaying this for a set period of time, we can make sure that restricted verification attempts automatically transition to denied (and they get the denial email) after a certain amount of time.

We could also leave the status in denied and delay only the sending of the denial _email_ if it feels better to have these stay in a different state forever, but I couldn't find a compelling reason to do that...

PR #2298, opened 2022-04-04
Heroku: delete unused `spatial_ref_sys` rows in `postdeploy` task

tgrathwell:
Heroku has a 10,000 row limit for Hobby postgres DBs,
and this table has 8500 rows in it. Deleting all those rows
will make us much less likely to get emails about how we're
gonna hit the row limit.


tgrathwell:
FYI I didn't test this properly before merging and it seems like it didn't work 😢  -- there's still 8500 rows.
Not sure exactly what went wrong... couldn't find any angry output from `heroku:postdeploy` in the logs before the thing shut down

PR #2300, opened 2022-04-04
Heroku: log the count of spatial_ref_sys rows before and after deletion

tgrathwell:
seems like the previous commit trying to delete them didn't actually work


tgrathwell:
This delete can't work right now because of Heroku Postgres' permissions, see https://github.com/orcasound/orcasite/pull/74

PR #2302, opened 2022-04-04
Create restrictions page in CTC flow

caitbot:
**This PR:**
- Adds the `RestrictionsController` and corresponding page.
- Removes the `question-mark.svg` because it is the same as `ineligible.svg` and `ineligible` is what it is called in Figma, so I chose to go with that. It also updates the one place that was using it to now use `ineligible`.
- Fixes up a lot of feature specs to include this new page, `complete_intake` clicks on the offboarding link then goes back, for test coverage's sake.

TIL if you click on the Export tag in figma with an SVG selected, it will tell you the name of the svg (and then it is easy to find in assets or export if its new). I guess the little Figma element tree does this as well. Very handy.

image


embarnard (approved):
Looks great!

PR #2307, opened 2022-04-05
CTC AdvCTC intake flow: create /adv-ctc-claimed and /not-filing

embarnard:


app/controllers/ctc/questions/advance_ctc_claimed_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Ctc
+  module Questions
+    class AdvanceCtcClaimedController < QuestionsController
+      include AuthenticatedCtcClientConcern
+
+      layout "intake"
+
+      def self.show?(intake)
+        return false if intake.advance_ctc_amount_received.nil? || intake.dependents.count(&:qualifying_ctc?).zero?
```
embarnard, line 9: A question for Gabe has been left in Pivotal Tracker, neato

app/views/ctc/questions/advance_ctc_received/edit.html.erb
```diff
@@ -10,7 +10,7 @@
       <h2><%= t("views.ctc.questions.advance_ctc_received.already_received") %></h2>
       <%= link_to t("general.edit").downcase, questions_advance_ctc_path, class: "review-box__edit review-box__edit-button" %>
     </div>
-    <p class="spacing-below-15"><%= t("views.ctc.questions.advance_ctc_received.total_adv_ctc", amount: current_intake.advance_ctc_amount_received) %></p>
+    <p class="spacing-below-15"><%= t("views.ctc.questions.advance_ctc_received.total_adv_ctc", amount: calculated_or_provided_dollar_amount(current_intake.advance_ctc_amount_received)) %></p>
```
embarnard, line 13: Unrelated to this pull request: Can we remove the `fallback_value` from calculated_or_provided_dollar_amount?

app/views/ctc/questions/not_filing/edit.html.erb
```diff
@@ -0,0 +1,15 @@
+<% @main_question = t("views.ctc.questions.not_filing.title") %>
```
embarnard, line 1: @caitbot Do you think we should adjust the position in the I18N tree such that `t('.title')` works? We'd seemingly have to do that for all CTC questions.

embarnard (commented):
Asheesh approves

app/views/ctc/questions/advance_ctc_received/edit.html.erb
caitbot: it seems like we can since it is never used (I recall thinking the same thing - but then not doing it)

app/views/ctc/questions/not_filing/edit.html.erb
caitbot: Is that how it is for GYR? I don't feel strongly. If it's a simple find and replace then sure, but if it's going to be more effort, don't bother. 

In some ways I prefer having the more descriptive path anyhow.

PR #2309, opened 2022-04-05
Write backfill for copying triage answers to intake

jenny-heath:
in the cases of `triage_filing_frequency` and `triage_vita_income_ineligible`, we are condensing multiple columns into one, which makes for some complicated-ish edge cases to consider. that said, i chose not to consider too heavily the cases that seem very unlikely to occur in the wild (for example, when two questions were asked on one form but one of them has an answer and the other is "unfilled"). in other words, i really leaned into the idea that "we don't know what we don't know", so you'll see the expected cases laid out in their own if clauses and everything else goes in the "unfilled" bucket. if that doesn't make sense it would probably be faster to talk on the phone.

this was probably way too much thought to have put into backfilling a few columns that are at this point just for historical data analysis, but i was feeling too sleepy not to write a test.

### DIFF
--- a/lib/tasks/backfill_triage_data.rake
+++ b/lib/tasks/backfill_triage_data.rake
@@ -1,0 +1,39 @@
+namespace :backfill_triage_data do
+  desc "Backfill data onto Triage that lived on Intake"
+  task intake_triage_fields: :environment do
+    Triage.includes(:intake).where('intake_id IS NOT NULL').find_each(batch_size: 100) do |triage|
+      filing_frequency = [
+        triage.filed_2018,
+        triage.filed_2019,
+        triage.filed_2020,
+        triage.filed_2021,
+      ]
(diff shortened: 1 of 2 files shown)


caitbot (approved):
I agree with your philosophy around this backfill (if we lack an answer or the ability to resolve data -- leave it unfilled). This looks right and good to me. Go for it.

lib/tasks/backfill_triage_data.rake
```diff
@@ -0,0 +1,43 @@
+namespace :backfill_triage_data do
+  desc "Backfill data onto Triage that lived on Intake"
+  task intake_triage_fields: :environment do
+    Triage.includes(:intake).where('intake_id IS NOT NULL').find_each(batch_size: 100) do |triage|
```
caitbot, line 4: beautiful `find_each` 😭

PR #2312, opened 2022-04-06
Refactor first intake page CTC concerns and update some copy.

caitbot:
**This PR:**
- Updates the copy on the following pages: `Income`, `Overview`
- Moves `FilingStatus` to before `Income`
- Refactors the concerns of the first form + controller of intake (taking security snapshot, initializing client, tax return, and intake) into `FirstPageOfCtcIntakeConcern` and `InitialCtcFormAttributes`. This way you can just move over the `includes` statements and the corresponding `shared_examples` and `shared_context` if we need to change the first page again. This is like the third time we've done this, so I figured it might be nice, and it was fun to do. TIL about `shared_context`s.
- Updates the `index.js`, this was sneaky and is something else you have to do when the order gets moved around.

TBH most of this was shifting specs around and then getting the `shared_context` to work.

I'll follow up with the rest of the copy changes, I didn't want this PR to get too cumbersome.

Here are the updated pages.
image
image

_Bonus section of fun things I learned about while implementing this_
- `shared_context`, I wonder if we could use it elsewhere.. I didn't know it existed and it let me keep some of the concerns of initial intake pages abstracted from the particular page/form that is the current initial page.
- `included do` in a mixin allows you to define instance methods in the mixin (which was necessary for the form). so fun.
- I probably could have made the form mixin into a service class instead, but was clearly over-enthused about concerns after refactoring the controller


spec/controllers/ctc/questions/filing_status_controller_spec.rb
```diff
@@ -1,48 +1,35 @@
...
+    include_context :first_page_of_ctc_intake_update_context, form_name: :ctc_filing_status_form, additional_params: { filing_status: "single" }
+    it_behaves_like :first_page_of_ctc_intake_update
   end
 
   describe '#update' do
     context "with no answer" do
-      let(:params) do
-        {}
-      end
+      include_context :first_page_of_ctc_intake_update_context, form_name: :ctc_filing_status_form, additional_params: { filing_status: nil }
```
caitbot, line 11: Unfortunately, I had to include the context in multiple places due to the dynamic nature of `additional_params`, a variable which is created with a `let` is only available in a `before` or `it` block, whereas `include_context` is only available in a `describe` or `context` block.

config/locales/es.yml
```diff
@@ -2037,16 +2037,32 @@ es:
...
+          income_qualifications:
+          - you earned less that %{standard_deduction} in income from any source
+          - you earned less that $400 in self employment income
+          - you did not recieve the Advanced Premium Tax Credit
+          title: 'In %{current_tax_year}, are the following statements correct:'
+          what_is_aptc_reveal:
+            content:
+              p1: You would not have received the Advanced Premium Tax Credit if you are on Medicaid, Medicare,  get healthcare coverage through an employer, or you don’t have healthcare coverage.
+              p2: You probably would have received the Advanced Premium Tax Credit if you purchased market-rate insurance in a health insurance marketplace under the Affordable Care Act.
+            title: What is the Advanced Premium Tax Credit?
```
tgrathwell, line 2065: do we want english copy on the spanish page? or was this supposed to be a placeholder?

spec/controllers/ctc/questions/filing_status_controller_spec.rb
tgrathwell: it seems like, if you really want to only include the context once, you could (instead of passing in additional params) just modify `params` in a before block like so:

```
before do
  params[:ctc_filing_status_form][:filing_status] = nil
end
```

but, this doesn't seem any worse tbh

spec/controllers/ctc/questions/filing_status_controller_spec.rb
```diff
@@ -1,48 +1,35 @@
...
-        {
-          ctc_filing_status_form:
-            { filing_status: "single" }
-        }
-      end
+      include_context :first_page_of_ctc_intake_update_context, form_name: :ctc_filing_status_form, additional_params: { filing_status: "single" }
+
       it "updates the tax return's filing status" do
-        put :update, params: params
+        post :update, params: params
```
tgrathwell, line 24: why post? probably effectively the same, just curious why the change

spec/controllers/ctc/questions/filing_status_controller_spec.rb
```diff
@@ -1,48 +1,35 @@
...
     context "with no answer" do
-      let(:params) do
-        {}
-      end
+      include_context :first_page_of_ctc_intake_update_context, form_name: :ctc_filing_status_form, additional_params: { filing_status: nil }
 
       it "re-renders the form with errors" do
         put :update, params: params
         expect(response).to render_template :edit
         expect(assigns(:form).errors).not_to be_blank
```
tgrathwell, line 16: could be a slightly better test if it cared *which* errors the form had, not just that they're not blank

app/lib/ctc_question_navigation.rb
```diff
@@ -24,8 +25,7 @@ class CtcQuestionNavigation
     Ctc::Questions::PhoneVerificationController,  # Same sign in behavior as previous controller, but verified through SMS instead of email
 
     # Filing Status
```
tgrathwell, line 27: this section is probably now "spouse info"

spec/features/ctc/complete_intake_spec.rb
```diff
@@ -79,10 +83,6 @@
     click_on I18n.t("views.ctc.questions.verification.verify")
 
     # =========== FILING STATUS ===========
```
tgrathwell, line 85: this section is probably now "spouse info"

tgrathwell (approved):

spec/controllers/ctc/questions/filing_status_controller_spec.rb
caitbot: I figured since its creating the things (intake, client, tax return) in the db rather than updating them, it was more idiomatically correct, but really, the funny thing is that its both creating and updating anyhow, and yeah, its effectively the same.

app/lib/ctc_question_navigation.rb
caitbot: good call

spec/controllers/ctc/questions/filing_status_controller_spec.rb
caitbot: hahahaah good point, I updated

config/locales/es.yml
caitbot: oops I meant to google translate those, thanks!

PR #2319, opened 2022-04-08
Log client ids during every Client.create! to diagnose why we're still making some clients with no intake

tgrathwell:


tgrathwell:
this implementation doesn't make it easy to tie the created `Client` id to a `visitor_id` which is unhelpful :|

not sure if there's ever a way to log `visitor_id` from model-land, though

PR #2322, opened 2022-04-11
Consent all clients (getCTC wasn't consented) and use that value to show clients

bytheway875:
This should resolve a bug wherein clients without tax returns (ie all clients that had routing method of :at_capacity) were not discoverable using search, and it also mirrors the logic we use for allowing clients to log in.

Also I found a bug where the greeters were not able to see clients who were assigned to them but had been transitioned out of greetable states. I have not seen this reported as a bug, but it's a clear regression in the code introduced in this PR: https://github.com/codeforamerica/vita-min/pull/1147 . The code as written with the || would never execute the left hand side.

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -88,7 +88,7 @@ class Client < ApplicationRecord
   end
 
   delegate *delegated_intake_attributes, to: :intake
-  scope :after_consent, -> { distinct.joins(:tax_returns).merge(TaxReturn.where.not(status: "intake_before_consent")) }
+  scope :after_consent, -> { joins(:intake).where("intakes.primary_consented_to_service_at IS NOT NULL") }
   scope :greetable, -> do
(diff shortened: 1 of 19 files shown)


app/models/client.rb
```diff
@@ -88,7 +88,7 @@ def self.sortable_intake_attributes
   end
 
   delegate *delegated_intake_attributes, to: :intake
-  scope :after_consent, -> { distinct.joins(:tax_returns).merge(TaxReturn.where.not(status: "intake_before_consent")) }
+  scope :after_consent, -> { joins(:intake).where("intakes.primary_consented_to_service_at IS NOT NULL") }
```
bytheway875, line 91: Screen Shot 2022-04-11 at 12 56 26 PM

Screen Shot 2022-04-11 at 12 56 52 PM

I got this SQL error when i had it written the more Railsy way. I THINK the search gem we use was interfering with the joins on intake?

bytheway875:

PR #2325, opened 2022-04-11
Revert datadog async

tgrathwell:


tgrathwell:
this commit turned out to be innocent

PR #2328, opened 2022-04-11
GYR: Update copy on spouse prior tax year and prior year agi

jenny-heath:
image

image


bytheway875 (approved):
👍

jenny-heath:
@bytheway875 yes it is still the case that the transifex translations will wipe out what we have, but in this case what we have was made up by me so that is okay 👌

PR #2332, opened 2022-04-12
Submission bundle 2021

bytheway875:
Hook it all together so that we build the right bundle


app/lib/submission_builder/ty2020/adv_ctc_irs1040.rb
```diff
@@ -1,5 +1,5 @@
 module SubmissionBuilder
-  module TY2020
+  module Ty2020
```
bytheway875, line 2: when i changed from ty_2020 to ty2020, it started not liking the all upper cased casing on the TY2020 module name. Not sure if that's a reason to keep it the other way...?

PR #2335, opened 2022-04-13
Intake: avoid creating orphaned Client records on personal info page

tgrathwell:
previously, if someone filled out PersonalInfoForm with a phony phone
number like 0000000000, it would leave a `Client` record in the DB
with no associated `Intake`. We would like to more confidently
assert that every Client has an associated Intake, so this commit
re-arranges the `.save` logic to try to assure that

it also makes phone numbers like `0000000000` show as a validation
error rather than trying to save them to the model where they
are silently rejected

### DIFF
--- a/spec/forms/personal_info_form_spec.rb
+++ b/spec/forms/personal_info_form_spec.rb
@@ -1,7 +1,8 @@
 require "rails_helper"
 
 RSpec.describe PersonalInfoForm do
-  let(:intake) { create :intake }
+  let(:intake) { Intake::GyrIntake.new }
+
   let(:valid_params) do
     {
       preferred_name: "Greta",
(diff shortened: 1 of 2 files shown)


spec/forms/personal_info_form_spec.rb
```diff
@@ -1,7 +1,8 @@
 require "rails_helper"
 
 RSpec.describe PersonalInfoForm do
-  let(:intake) { create :intake }
+  let(:intake) { Intake::GyrIntake.new }
```
jenny-heath, line 4: is `build :intake` the same thing?

spec/forms/personal_info_form_spec.rb
```diff
@@ -29,6 +30,21 @@
...
+        weird_params = valid_params.merge(additional_params).merge(phone_number: '0000000000', phone_number_confirmation: '0000000000')
+        form = described_class.new(intake, weird_params)
+
+        expect do
+          expect do
+            expect do
+              form.save
+            end.to raise_error(ActiveRecord::RecordInvalid)
+          end.not_to change { Client.count }
+        end.not_to change { Intake.count }
```
jenny-heath, line 44: just curious, is this the same as `.and not_to change { }` ?

spec/forms/personal_info_form_spec.rb
```diff
@@ -29,6 +30,21 @@
       end
     end
 
+    context "if we're trying to .save data that cannot make it into the db (mismatch between model and form validation)" do
+      it "creates neither a client nor an intake" do
+        weird_params = valid_params.merge(additional_params).merge(phone_number: '0000000000', phone_number_confirmation: '0000000000')
```
jenny-heath, line 35: i thought this example was no longer a mismatch between the model and form validation?

jenny-heath (approved):
couple of small comments, mostly concerned about the "mismatch" one (test still seems valid, just not sure it's testing what it says it's testing in the description)

PR #2339, opened 2022-04-14
CTC: Limit the number of resubmissions

embarnard:


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -37,6 +37,7 @@ def resubmit
         flash[:alert] = I18n.t("general.enable_javascript")
         return redirect_back(fallback_location: ctc_portal_edit_info_path)
       end
+      return redirect_back(fallback_location: ctc_portal_edit_info_path) unless @submission.tax_return.under_submission_limit?
```
alaroia-cfa, line 40: I know elsewhere in this controller action we `redirect_back`, but I personally prefer to specifically redirect to where we need to go.

e.g. `ApplicationController#redirect_to_intake_after_triage` has these lines:

```
  def redirect_to_intake_after_triage
    if Rails.env.production?
      redirect_to Questions::OverviewController.to_path_helper
    else
      redirect_to Questions::EnvironmentWarningController.to_path_helper
    end
  end
```

If you know where to go, why go "back" with a "fallback" when you can go there? :)

app/models/tax_return.rb
```diff
@@ -241,6 +241,10 @@ def enqueue_experience_survey
     end
   end
 
+  def under_submission_limit?
```
alaroia-cfa, line 244: So simple!

spec/controllers/ctc/portal/portal_controller_spec.rb
```diff
@@ -104,6 +104,21 @@
           expect(response).to redirect_to Ctc::Portal::PortalController.to_path_helper(action: :edit_info)
         end
       end
+
+      context "when the client has resubmitted 20 times" do
```
alaroia-cfa, line 108: This is a good simple test. Excellent.

spec/controllers/ctc/portal/portal_controller_spec.rb
```diff
@@ -104,6 +104,21 @@
...
+        let!(:submissions) { create_list :efile_submission, 20, :rejected, tax_return: client.tax_returns.first }
+
+        before do
+          client.tax_returns.first.update(efile_submissions: submissions)
```
alaroia-cfa, line 112: Usually, it's possible to avoid writing this kind of `before` block by carefully selecting what you put into the factory. If it's unavoidable that's fine, but I strongly encourage you to figure out how to remove it, just from a code cleanliness perspective. Not a blocker.

spec/models/tax_return_spec.rb
```diff
@@ -974,4 +974,24 @@
...
+
+      it "returns true" do
+        expect(tax_return.under_submission_limit?).to eq true
+      end
+    end
+
+    context "when the tax return has been resubmitted 20 times" do
+      let!(:efile_submissions) { create_list :efile_submission, 19, tax_return: tax_return }
+
+      it "returns false" do
```
alaroia-cfa, line 992: The test on the line below asserts it `eq true` but here the `it` says `false`. Which one is it? :)

alaroia-cfa (changes requested):
The last test in the pull request needs a change. Largely looks great. Some non-blocking comments, too.

PR #2340, opened 2022-04-14
Generate 1040 & Schedule 8812 PDF

embarnard:


app/models/efile_submission_dependent.rb
```diff
@@ -21,7 +21,7 @@ class EfileSubmissionDependent < ApplicationRecord
   belongs_to :dependent
   belongs_to :efile_submission
 
-  delegate :first_name, :last_name, :ssn, :irs_relationship_enum, :birth_date, :tin_type_ssn?, :tin_type_atin?, to: :dependent
+  delegate :age, :first_name, :last_name, :full_name, :ssn, :irs_relationship_enum, :birth_date, :tin_type_ssn?, :tin_type_atin?, to: :dependent
```
caitbot, line 24: oops we probably don't need this anymore (was due to a misunderstanding in conversation with JH + WS), remove `age`

app/lib/irs1040_pdf.rb
```diff
@@ -1,15 +1,15 @@
-class AdvCtcIrs1040Pdf
```
bytheway875, line 1: I sort of wish we'd followed the pattern like we did for the schemas where we kept last years stuff around, namespaced them to the year, and created a new one... just in case we do file backtaxes this season, but I feel like that likely won't happen so this is cool.

app/lib/irs8812_pdf.rb
```diff
@@ -0,0 +1,44 @@
+class Irs8812Pdf
```
bytheway875, line 1: I def think its worth namespacing these to a year, though, just to create a precedent moving forward and keep it consistent (even if we don't keep last years PDF for the 1040 around). It would make it more clear in the naming that the implementation is super specific to the tax year. The appearance and content on the 8812 PDF changed a lot from 2020 to 2021 tax years, for example.

app/models/efile_submission.rb
```diff
@@ -128,9 +128,15 @@ def generate_irs_address
 
   def generate_form_1040_pdf
```
bytheway875, line 129: since this is now the entire tax return (includes more than just the 1040, maybe we can update this to "generate_tax_return_pdf" or "generate_filing_pdf"

app/models/efile_submission.rb
```diff
@@ -128,9 +128,15 @@ def generate_irs_address
...
     filename = "IRS 1040 - TY #{tax_return.year} - #{irs_submission_id}.pdf"
-
+    temp_1040_pdf = Irs1040Pdf.new(self).output_file
+    temp_8812_pdf = Irs8812Pdf.new(self).output_file
+    combined_temp_file = Tempfile.new(
```
bytheway875, line 133: I wonder too if we should update the name of the filename to make it clear that it also includes the 8812. Although i think i recall that the ticket wanted to continue using the 1040 document type so maybe its okay to keep the naming conventions.

bytheway875 (approved):
this looks good to me... just a couple non blocking comments on naming conventions.

app/lib/irs8812_pdf.rb
caitbot: totally, I'll add the year to the name

app/models/efile_submission.rb
caitbot: yeah, since we are labeling it as the 1040 document, I prefer to keep it the same, even though it now includes the 8812 attachment.

PR #2344, opened 2022-04-15
Fix IP pin page, none of the above option

caitbot:
When using the `at_least_one_or_none_of_the_above` validator, you cannot include the `none_of_the_above` attribute in the form implemented `at_least_one_selected`. I added a test to this form to prevent this in the future (for this form). 

It is unfortunately, something you need to be aware of when using this validator, otherwise bad behavior occurs (selecting none of the above will cause the error message to show, incorrectly).

### DIFF
--- a/app/forms/ctc/ip_pin_form.rb
+++ b/app/forms/ctc/ip_pin_form.rb
@@ -40,7 +40,6 @@ module Ctc
     end
 
     def at_least_one_selected
-      no_ip_pins == "yes" ||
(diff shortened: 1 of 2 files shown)


wschaefer (approved):
This looks good! Jenny and I looked at the validator and thought its kind of hard to reason about. If you felt like making changes there too while you're in here, we thought it would be clearer if the validator argument "value" was renamed to "none_of_the_above_value"

PR #2345, opened 2022-04-15
CTC portal allows removing dependents again

wschaefer:
But only when the client will still have a positive CTC after the
dependent is removed


alaroia-cfa (approved):
lgtm

alaroia-cfa:
I'm going to merge this to main so Yvonne can accept on demo; see Pivotal Tracker story.

PR #2348, opened 2022-04-18
CTC consent: replace 'i agree' button with checkbox and 'continue' button

jenny-heath:
didn't do any fancy styling on the checkbox since there wasn't anything in the story

image


alaroia-cfa (approved):
The code looks simple and great!

PR #2352, opened 2022-04-19
Allow hub users to save ITIN applicants without an SSN

caitbot:


app/models/intake.rb
```diff
@@ -525,8 +525,7 @@ def new_dependent_token
   end
 
   def itin_applicant?
-    # TODO: remove the triage part once we have migrated Triage#id_type to Intake#need_itin_help
-    need_itin_help_yes? || self&.triage&.id_type_need_itin_help?
+    need_itin_help_yes?
```
bytheway875, line 528: nice! glad to see we're going back to pull out the triage-related TODOs :D

bytheway875 (commented):
One question about the @client thing, that is potentially blocking because I'm not sure of the potential impact.

I'm guessing that the method that we need to use is only on the "presented" version of the object, but if that's the case to me it seems like the abstraction of that method into the presenter may be a poor one and we should instead move it up onto the model itself... it really weirds me out to have our form acting on saving a client from a presenter object instead of the client itself, but I'm open to being convinced that this is copacetic.

PR #2355, opened 2022-04-19
Display fraud score data in places

bytheway875:
Screen Shot 2022-04-19 at 12 33 40 PM
Screen Shot 2022-04-19 at 12 42 02 PM
Screen Shot 2022-04-19 at 12 42 20 PM


app/views/hub/security/_fraud_score_snapshot.html.erb
```diff
@@ -0,0 +1,8 @@
+<tr class="index-table__row index-table__vertical_align">
```
caitbot, line 1: I find it mildly confusing that this is named `fraud_score_snapshot` and calls another partial called `fraud_score_snapshot`, consider naming this `fraud_score`?

app/views/hub/efile_submissions/_log.html.erb
```diff
@@ -11,8 +11,19 @@
...
+    <% if transition.to_state.in? ["bundling", "fraud_hold"] %>
+      <% fraud_score = transition.efile_submission&.fraud_score %>
+      <div> Fraud Score: <%= fraud_score.score %></div>
+      <% if fraud_score.snapshot.present? %>
+        <div class="accordion spacing-above-10">
+          <a href="#" class="accordion__button" aria-expanded="true" aria-controls="a2">
+            <h3>Details</h3>
+          </a>
+          <div class="accordion__content" id="a2">
+            <%= render partial: "shared/fraud_score_snapshot", locals: { fraud_score: transition.efile_submission.fraud_score, fraud_indicators: Fraud::Indicator.all } %>
```
caitbot, line 23: I notice that you load `fraud_indicators`once in the controller on the security page where it is likely to be listed multiple times. Since this is also a partial, would it make sense to do that here? 

Is there a higher level place where that makes sense to be done and can be passed in? Not a big deal if not, I was just curious, otherwise on the page with many efile logs we may load the indicators many times.

caitbot (approved):
Sweet! None of my comments are blocking, just some suggestions/questions. This looks good to me. Excited to see this go live, this will be helpful!

app/views/hub/efile_submissions/_log.html.erb
```diff
@@ -11,8 +11,19 @@
           <%= error.message %>
         </div>
     <% end %>
-    <% if transition.metadata["indicators"].present? %>
-      <div> Fraud indicators: <%= transition.metadata["indicators"].map(&:humanize).join(", ") %></div>
+    <% if transition.to_state.in? ["bundling", "fraud_hold"] %>
```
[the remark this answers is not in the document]
bytheway875: every client will be assigned a fraud score, but some may not exceed the threshold to put them on hold. For example, if an international timezone is there but its the only indicator, their score would only be 20 points so they get passed to bundling.

Each submission will only ever be in bundling OR fraud hold, because we'll create another submission when they resubmit.

app/views/hub/efile_submissions/_log.html.erb
bytheway875: i dont think that it will run multiple queries, because the query gets called only once because it's cached to run the query only once per page. But yes, it probably makes more sense just to load them once.

Since we're going to be grabbing every entry in this table often, I was thinking about using redis cache to store this but we don't have a cache store set up on production so im going to defer that work for now.

PR #2358, opened 2022-04-19
Remove "Received unemployment income" filter

tgrathwell:
It hasn't been working for a nearly a year, maybe it isn't needed.


bytheway875 (approved):
👍 

There's also a UI checkmark on the client index per-client. Wonder if we can check with product to see whether that is also an artifact of a previous tax season thing that can now be removed?

PR #2359, opened 2022-04-19
Consent ctc clients to service, and show consented at time on client profile

bytheway875:
Screen Shot 2022-04-19 at 3 16 14 PM


alaroia-cfa (approved):
Code looks simple and nice!

PR #2360, opened 2022-04-19
Fix crash submitting a CTC without a bank account

wschaefer:
The fraud check expected there would always be one, but sometimes it is a check instead

Also the flows generates an intake with refund_payment_method of check now

### DIFF
--- a/app/models/fraud/indicator.rb
+++ b/app/models/fraud/indicator.rb
@@ -70,6 +70,9 @@ module Fraud
     end
 
     def duplicates(references)
+      # skip this rule if we can't check against the reference object
+      return false if references[reference].blank?
+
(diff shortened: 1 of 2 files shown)


app/models/fraud/indicator.rb
```diff
@@ -70,6 +70,9 @@ def in_riskylist(references)
     end
 
     def duplicates(references)
+      # skip this rule if we can't check against the reference object
+      return false if references[reference].blank?
```
bytheway875, line 74: nice, im going to update this to be [0, []] since the expected return value of each of these methods is a tuple where the first item is the # of points from the test result and the second is some info about what made it fail. i ran into an issue where the points ends up getting set to false which breaks things in a different way.

PR #2361, opened 2022-04-19
Configure open and close times for getctc season

bytheway875:
Screen Shot 2022-04-19 at 4 25 17 PM

---

Last year, I think we were auto-redirecting into the flow during the beta because the homepage wasn't set up to have any links to get started. However, now that we're able to toggle between open and closed states, we can change the home page during the beta state instead of redirecting into the flow directly. This seems to me to be more straightforward, so I removed the redirect into the flow for the beta (another reason I removed it is that I actually didn't fully understand how the ctc app was consuming the open_for_ctc_intake? and open_for_ctc_login? methods. Previously these were used in conjunction with additional environment variables, but we don't have to rely on those variables any more for this approach to work.

IF we want to make it clear to clients that we're in beta while it's beta time, maybe we can add a banner when the beta cookie is present and we haven't passed the full launch date yet...

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,10 +1,7 @@
 class ApplicationController < ActionController::Base
-  CTC_INTAKE_CLOSING_TIME = Time.find_zone('America/Los_Angeles').parse('2021-11-15 23:59:59')
-  CTC_LOGIN_CLOSING_TIME = Time.find_zone('America/Los_Angeles').parse('2021-11-19 23:59:59')
-
   include ConsolidatedTraceHelper
   around_action :set_time_zone, if: :current_user
-  before_action :redirect_to_getyourrefund, :set_visitor_id, :set_source, :set_referrer, :set_utm_state, :set_navigator, :set_sentry_context, :set_collapse_main_menu
+  before_action :set_ctc_beta_cookie, :set_visitor_id, :set_source, :set_referrer, :set_utm_state, :set_navigator, :set_sentry_context, :set_collapse_main_menu
   around_action :switch_locale
   before_action :check_maintenance_mode
   after_action :track_page_view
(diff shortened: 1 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -325,12 +337,6 @@ def redirect_or_add_flash
     end
   end
 
-  def redirect_to_getyourrefund
-    if request.get? && request.host.include?("vitataxhelp.org")
```
bytheway875, line 329: we don't have the app on vitataxhelp.org at all anymore, so this code is totally stale.

app/controllers/concerns/ctc/can_begin_intake_concern.rb
```diff
@@ -9,13 +9,10 @@ module CanBeginIntakeConcern
     private
 
     def check_ctc_intake_cookie
-      return unless Rails.env.production?
+      return if open_for_ctc_intake?
```
bytheway875, line 12: all the beta stuff gets handled in the open_for_intake? method already, so we can simplify this implementation a little.

app/controllers/ctc/ctc_pages_controller.rb
```diff
@@ -1,11 +1,6 @@
 module Ctc
   class CtcPagesController < CtcController
     def home
-      if params[:ctc_beta] == "1" && ENV['DISABLE_CTC_BETA_PARAM'].blank?
```
bytheway875, line 4: I believe this redirect was added because before launch last year, we didn't have links to start intake on the home page "toggling" for when intake is open. Now we can toggle the appearance what links to start intake or sign up for more info show to people, so we can simplify by not having a redirect in the controller.

caitbot (approved):
Reviewed as pair with SB, this looks good to me and seems like a good clean up. Def should be heavily acceptance tested, because I lack context on the intention of the previous implementation. But this looks great to my naive eyes!

Just have to fix some tests.

PR #2363, opened 2022-04-20
Add new validation text data for GYR EV certificate [#181128168]

alaroia-cfa:

### DIFF
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -77,8 +77,18 @@ class PublicPagesController < ApplicationController
 
   def pki_validation
     # Used for Identrust annual EV certificate validation
+    website_text =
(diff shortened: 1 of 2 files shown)


alaroia-cfa:
Validate by visiting http://localhost:3000/.well-known/pki-validation/9061966.txt and comparing against the Pivotal Tracker story

caitbot (commented):
I am receiving `qrk3...` for both the CTC domain and standard domain when I test locally with this branch. Based on my reading of this code and the story, that doesn't seem like the intention? But I may be missing something as I am not an expert in these EV certificate things.

caitbot (approved):
Ignore my prior comment, I was working off an outdated branch. This is great. Tested as expected! Yay!

alaroia-cfa:
@caitbot Thanks for the careful review!

PR #2368, opened 2022-04-20
Add a nil check to the fraud score section on the client efile page

caitbot:


app/views/hub/efile_submissions/_log.html.erb
```diff
@@ -11,7 +11,7 @@
           <%= error.message %>
         </div>
     <% end %>
-    <% if transition.to_state.in? ["bundling", "fraud_hold"] %>
+    <% if transition.to_state.in? ["bundling", "fraud_hold"] && transition.efile_submission&.fraud_score.present? %>
       <% fraud_score = transition.efile_submission&.fraud_score %>
```
bytheway875, line 15: nit, but if we're already checking for the &. in the if then we dont need it when we assign fraud_score on 15

bytheway875 (approved):

PR #2369, opened 2022-04-21
Client verification attempts 181901434

bytheway875:


app/controllers/ctc/portal/portal_controller.rb
```diff
@@ -78,4 +75,16 @@ def ensure_current_submission
       redirect_to(action: :home) and return
     end
   end
+
+  def load_current_submission
+    @submission = current_client.efile_submissions.last
```
bytheway875, line 80: this ensures that we dont have to load ALL the submissions then narrow it down to the last one -- does a SQL query to just get the last one from the db or returns nil

app/controllers/ctc/portal/verification_attempts_controller.rb
```diff
@@ -0,0 +1,51 @@
+class Ctc::Portal::VerificationAttemptsController < Ctc::Portal::BaseAuthenticatedController
+  layout "intake"
+  helper_method :prev_path, :illustration_path, :illustration_folder
```
bytheway875, line 3: i need to define these methods as helper methods so that i can use the a shared layout without directly inheriting from questions controller.

app/views/ctc/portal/portal/home.html.erb
```diff
@@ -69,9 +69,6 @@
       <div class="spacing-above-35">
         <%= link_to t('views.ctc.portal.home.contact_us'), new_ctc_portal_message_path, class: "button button--full-width" %>
       </div>
-      <% if @can_submit_documents %>
```
bytheway875, line 72: removing this because we only added this last season to facilitate the upload of identity documents.

bytheway875: im actually going to check with product to see if it should be removed or not

app/views/ctc/questions/file_full_return/edit.html.erb
```diff
@@ -6,14 +6,12 @@
 
   <p><%= t("views.ctc.questions.file_full_return.help_text", current_tax_year: current_tax_year)%></p>
 
-  <div class="reveal-shrink-wrapper">
```
bytheway875, line 9: This was just a random thing I came across where there was a full width button on the page and then this reveal that was awkwardly shorter than the button. Removing this class makes it the same width as the button.

app/controllers/ctc/portal/verification_attempts_controller.rb
caitbot: what does defining them as `helper_methods` do?

app/controllers/ctc/portal/verification_attempts_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  # allows for deletion of photos from an existing attempt
+  def destroy
+    @verification_attempt = current_client.verification_attempts.find_by(id: params[:id])
+    if @verification_attempt.blank? || !params[:photo_type].in?(["selfie", "photo_identification"])
+      flash["alert"] = I18n.t("general.access_denied")
+      redirect_to action: :edit and return
+    end
+    @verification_attempt.send(params[:photo_type]).purge_later
```
caitbot, line 32: huh til about `purge_later`

app/models/efile_submission.rb
```diff
@@ -76,7 +76,16 @@ def accepted_as_imperfect_return?
...
-    history.reverse.find do |transition|
+    # TODO: Simplify logic here so that we can show appropriate next steps for all cancelled returns
+    transitions = history.reverse
+    # Allow showing of cancelled state for clients who transitioned from fraud hold
+    if transitions[0]&.to_state == "cancelled" && transitions[1]&.to_state == "fraud_hold"
+      return transitions[0]
+    end
+
+    # We don't show the cancelled status HERE because it hides the instructions
+    # for how to handle your reject through cpaper filing.
```
caitbot, line 87: super nit, type in `paper filing`

app/models/verification_attempt.rb
```diff
@@ -34,5 +34,14 @@ def state_machine
   delegate :can_transition_to?,
            :current_state, :history, :last_transition, :last_transition_to,
            :transition_to!, :transition_to, :in_state?, to: :state_machine
+
+  scope :open, -> { in_state(:new, :pending, :escalated, :restricted) }
```
caitbot, line 38: what is a `restricted` verification attempt?

app/views/hub/verification_attempts/show.html.erb
```diff
@@ -57,8 +57,8 @@
                     <% end %>
                   </div>
                 </div>
-                <div class="grid-flex justify-flex-end">
-                  <div class="spacing-below-0"><%= link_to "#{timestamp(attempt.created_at)}", hub_verification_attempt_path(id:attempt.id), id: "time" %></div>
+                <div class="grid-flex column-column justify-flex-end">
```
caitbot, line 60: `column-column` is a bit of a silly class name, my guess is its for a nested column? 🙃

app/views/hub/verification_attempts/show.html.erb
```diff
@@ -57,8 +57,8 @@
...
+                  <div class="spacing-below-0" style="font-size: smaller; text-align: right;"><%= link_to "#{timestamp(attempt.created_at)}", hub_verification_attempt_path(id:attempt.id), id: "time" %></div>
```
caitbot, line 61: did you mean for this to be a class or want to keep it inline?

spec/features/hub/review_verification_attempts_spec.rb
```diff
@@ -160,7 +122,7 @@
     visit hub_verification_attempts_path
 
     # check info in table
-    within "#verification-attempt-#{verification_attempt_2.id}" do
+    within "#verification-attempt-#{catie_verification_attempt.id}" do
```
caitbot, line 125: hahahah, a cameo!

caitbot (approved):
Nice specs! There were a good amount of them!

Screenshots included on the PR would be nice since this adds a good bit of UI and its always nice to see it. 

This all looks good to me. No blocking comments, some questions, but thats it.

app/views/ctc/portal/verification_attempts/edit.html.erb
```diff
@@ -0,0 +1,95 @@
...
+          <%= f.label(:photo_identification, class: "button button--wide button--icon js-only", style: "display: none !important;") do %>
+              <div style="display:flex;">
+                <%= image_tag "icons/plus-circle.svg", alt: "" %>
+                <%= t("views.ctc.portal.verification.id_label") %>
+              </div>
+          <% end %>
+        <% end %>
+        <% unless @verification_attempt.selfie.present? %>
+          <%= f.file_field(:selfie, class: "form__documentuploader file-input", data: { "upload-immediately" => true}) %>
+          <%= f.label(:selfie, class: "button button--wide button--icon js-only", style: "display: none !important;") do %>
```
caitbot, line 35: consider making the inline style into a class? unless it needs to be inlined for importance

app/controllers/ctc/portal/verification_attempts_controller.rb
bytheway875: it lets you use them as methods in the template. the intake layout requires that these methods are defined in order to display the back button and icons

app/models/verification_attempt.rb
bytheway875: when someone has a fraud score that's super duper high, we will still let them submit a verification attempt but it's in the restricted status, so we deal with it a little bit differently.

app/views/hub/verification_attempts/show.html.erb
bytheway875: its a flex method - there's methods like column-column, row-column, column-row so that you can create different layouts for mobile and larger screens. since the default flex direction is row, row-row would be the implied default.

This change makes sure the two elements are stacked on top of one another instead of aligned side by side.

PR #2379, opened 2022-04-21
CTC intake: remove spouse can be claimed as dependent question

jenny-heath:
I checked production and `Intake.where.not(spouse_can_be_claimed_as_dependent: 0).count` is 0, which is why it seems safe to remove the column!

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_04_01_183023) do
+ActiveRecord::Schema.define(version: 2022_04_21_222721) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -1620,7 +1619,7 @@
            SELECT DISTINCT tax_returns.client_id
              FROM (tax_returns
                JOIN intakes ON ((intakes.client_id = tax_returns.client_id)))
-            WHERE ((tax_returns.current_state)::text <> ALL ((ARRAY['intake_before_consent'::character varying, 'intake_in_progress'::character varying, 'intake_greeter_info_requested'::character varying, 'intake_needs_doc_help'::character varying, 'file_mailed'::character varying, 'file_accepted'::character varying, 'file_not_filing'::character varying, 'file_hold'::character varying, 'file_fraud_hold'::character varying])::text[]))
```
caitbot, line 1623: This isn't related but also doesn't seem harmful?

caitbot (approved):
Yay more removal

db/schema.rb
bytheway875: this happens either because the UI was used to run a migration or vice versa... I've never been able to figure out which changes it to which

jenny-heath: oops this slipped under the radar! i'll remove it so it doesn't thrash

PR #2381, opened 2022-04-22
 Unmask SSNs on PDF and move PDF generation to the bundling step.

bytheway875:
- This change unmasks the SSNs on the PDF and removes bank account information entirely.

It also moves PDF generation into a job in the preparing step instead of in the bundling step.

This allows the PDF to be available to clients who need to paper mail their returns because they don't have ID documents or we can't verify their identity by moving it into the preparing step instead of waiting until the bundling step. 

--------

Because we want to provide Paper 1040s to clients who either :
a) get denied on a verification attempt
b) do not have ID documents that they can upload

we need to now create PDFs earlier in the flow (we currently check for fraud in the “preparing” step and then generate the PDF right before we create the submission bundle in the “bundling” step, which comes after.)

There are a couple of concerns to keep in mind here:
- We currently do the fraud score generation in the FOREGROUND because we want to make sure that we’re able to use the results of the fraud score to send the client to the right spot in the flow. We _could_ move this to a background job, but it would delay what we do for fraud and there might be a gap of time before clients are prompted to come back and upload photos.
- However, we create the PDF in the _background_. This means that if we transition the efile submission state to “failed” if the PDF generation fails, clients may come back later and see “failed” messaging in the application instead of the fraud flow.
- Also, we know that PDF failures are one of the “unhandled” driving concerns from last season. Not sure why PDF generation would fail … was it a PDF gem issue? Data issues? 

- Always use clients entered address on the paper PDF, since we verify their address in the bundling phase and its fine to send a paper PDF to the IRS with an address in any format
    - If we use the verified address from last pass and fall through to the raw address if the verified one doesn’t exist, we run into a situation where the PDF might use the verified address from the LAST submission instead of whatever the client recently updated their address to be. So just going with the client-submitted address on the paper return seems a lot safer.

- What should we do when PDF generation fails?
    - Currently we transition them into another state, but that seems like overkill now because we don’t want PDF generation to interrupt their ability to move foward in the flow (if we change their status to failed, and they’re actually in a fraud state, it will remove them from the verification flow.
    - I’m not actually sure what type of PDF failures we saw last season, so I think we’d need to look into that.

- How will we mitigate PDF failures for clients?
    - The raised error on PDF generation may report to sentry. If not, we should make sure it does.
    - We can report the count to Datadog?
    - We’ll need a way to figure out how we can generate PDFs successfully where it fails in the first place.
        - Retry?
        - Add button in hub for regeneration of PDF?
        - Allow client to download blank PDF if we couldn’t generate one?
        - Add a note or indicator to client profile if it fails? (exposure for customer service, I guess)
        - Create some sort of dashboard to investigate PDF generation failures.


bytheway875:
After discussion with @caitbot , the approach will be: attempt creation of pdf in the backround. When client clicks on download PDF button, let them download the existing pdf if there is one. otherwise, create the pdf and have them download that one. if there is an error in foreground PDF creation, show the client an error message that instructs them to reach out to customer service.

caitbot (approved):
this looks good to me. I have some thoughts about using the `intake.address` on the pdf, but I don't feel that they are blocking. happy to discuss over tuple if you want.

app/lib/irs1040_pdf.rb
```diff
@@ -10,7 +10,7 @@ def initialize(submission)
     @tax_return = submission.tax_return
     @intake = submission.intake
     @qualifying_dependents = submission.qualifying_dependents
-    @address = @submission.address
+    @address = @intake.address
```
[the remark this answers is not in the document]
bytheway875: i started writing this reponse about how i explored that and it wouldnt work, but i mistakenly thought that the address was associated with the INTAKE and not the submission, so i thought there was going to be some out-of-sync-ness. But, this does make sense. However, because of the way we do the PDF generation, the vast majority of clients WONT have a verified address when we create the PDF because we do the address verification later... i can change that logic around, but it might get a little messy...

app/jobs/create_submission_pdf_job.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSubmissionPdfJob < ApplicationJob
+  def perform(submission_id)
+    submission = EfileSubmission.includes(:intake, :qualifying_dependents, :verified_address, :tax_return).find(submission_id)
+
+    begin
+      submission.generate_filing_pdf
+    rescue StandardError => e
+      DatadogApi.increment('clients.pdf_generation_failed')
+      raise
```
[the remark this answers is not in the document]
bytheway875: in the docs it says: If an exception from a job is not rescued, then the job is referred to as "failed".

app/models/efile_submission.rb
```diff
@@ -117,10 +117,8 @@ def previously_transmitted_submission
     previous_submission if previous_submission.last_transition_to("transmitted").present?
   end
 
-  def generate_irs_address
-    if address.present? && address.skip_usps_validation
-      return OpenStruct.new(valid?: true)
-    end
+  def generate_verified_address
+    return verified_address if verified_address.present?
```
bytheway875, line 121: the active record object will respond to .valid?

caitbot: oh sneaky, I love it, this makes sense to me but I had to go look at the call site to understand the context (the context being that we call this method and expect an `address_creation` object instead of the actual `address` object). It might be nice to leave a comment here saying more or less what this comment says.

app/jobs/create_submission_pdf_job.rb
```diff
@@ -2,6 +2,12 @@ class CreateSubmissionPdfJob < ApplicationJob
   def perform(submission_id)
     submission = EfileSubmission.includes(:intake, :qualifying_dependents, :verified_address, :tax_return).find(submission_id)
 
+    begin
```
caitbot, line 5: the `generate_verified_address` service doesn't raise though, right? it just creates an object with errors on it?

app/models/efile_submission.rb
bytheway875: yeah, ill revert to using an openstruct because i dont want there to be unexpected return values if we end up adding validations that can fail on addresses at some point or something like that

app/jobs/create_submission_pdf_job.rb
bytheway875: it can also raise if there is an issue connecting to the service for whatever reason, but i guess that the comment is sort of misleading.

PR #2382, opened 2022-04-22
Instead of raising, use 0 as the fallback value for advanced CTC amount received

caitbot:
Raising was causing [this sentry](https://sentry.io/organizations/codeforamerica/issues/3216566515/?project=1880321&referrer=slack) when creating the 8812 PDF and no qualifying CTC dependents were present on the intake (clients don't fill out `advance_ctc_amount_received` if there are no qualifying dependents).

Investigating the use of this value, it seems that it cannot return nil because it is used to compute outstanding amount, but using 0 as a fallback should be ok.

After checking with product, it turns out listing 0 or blank are the same, and so listing 0 if they haven't actually filled out the value is totally valid. Going to go with that for simplicities sake.

### DIFF
--- a/app/lib/irs8812_ty2021_pdf.rb
+++ b/app/lib/irs8812_ty2021_pdf.rb
@@ -8,7 +8,7 @@ class Irs8812Ty2021Pdf
   def initialize(submission)
     @submission = submission
     @qualifying_dependents = submission.qualifying_dependents
-    @ctc_qualifying_dependents =  @qualifying_dependents.select {|d| d.qualifying_ctc? }
+    @ctc_qualifying_dependents =  @qualifying_dependents.select(&:qualifying_ctc?)
     @benefits = Efile::BenefitsEligibility.new(tax_return: submission.tax_return, dependents: @qualifying_dependents)
   end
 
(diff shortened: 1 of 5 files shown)


app/lib/irs8812_ty2021_pdf.rb
```diff
@@ -35,7 +36,7 @@ def hash_for_pdf
       Line14c: 0, #14c
       Line14d: 0, #14d
       TotalCtcAmt14e: 0, #14e
-      AdvCtcReceived14f: @benefits.advance_ctc_amount_received, #14f
+      AdvCtcReceived14f: @ctc_qualifying_dependents.none? && @benefits.advance_ctc_amount_received.zero? ? nil : @benefits.advance_ctc_amount_received, #14f
```
caitbot, line 39: per yvonne, if there are no qualifying dependents, they will not have answered the question about advanced CTC and so we should leave it blank

app/lib/irs8812_ty2021_pdf.rb
```diff
@@ -35,7 +36,7 @@ def hash_for_pdf
       Line14c: 0, #14c
       Line14d: 0, #14d
       TotalCtcAmt14e: 0, #14e
-      AdvCtcReceived14f: @benefits.advance_ctc_amount_received, #14f
+      AdvCtcReceived14f: @ctc_qualifying_dependents.none? ? nil : @benefits.advance_ctc_amount_received, #14f
```
caitbot, line 39: per @yvonnefong's suggestion, if there are no qualifying dependents, they will not have answered whether there was advanced ctc received and so we should leave it blank

bytheway875 (approved):
👍

PR #2384, opened 2022-04-22
AGI should be 0 [#181973806]

alaroia-cfa:


alaroia-cfa:
It seems that the PDF needed this change but that the e-file XML data did not.

Last year's 1040 XML shows taxable interest, income, and AGI of 1:

https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/ty2020/documents/irs1040.rb#L43-L45

This year's 1040 shows income of 0: https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/ty2021/documents/irs1040.rb#L46

That shows me that the XML data was fine but the PDF was not.

embarnard:
> It seems that the PDF needed this change but that the e-file XML data did not.
> 
> Last year's 1040 XML shows taxable interest, income, and AGI of 1:
> 
> https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/ty2020/documents/irs1040.rb#L43-L45
> 
> This year's 1040 shows income of 0: https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/ty2021/documents/irs1040.rb#L46
> 
> That shows me that the XML data was fine but the PDF was not.

I think I'm confused by what you mean when you say the PDF needed to change. Do you mean the actual pdf or the irs1040_pdf.rb file? Like you're saying what we are submitting to the IRS through the XML is fine but what is getting displayed on the pdf created through irs1040_pdf.rb is not?

embarnard (approved):

alaroia-cfa:
> Do you mean the actual pdf or the irs1040_pdf.rb file?

The PDF that the code stores in the client's documents as the 1040 has the wrong information filled-in. To fix it, I changed the irs1040_pdf.rb Ruby file.

> Like you're saying what we are submitting to the IRS through the XML is fine but what is getting displayed on the pdf created through irs1040_pdf.rb is not?

Exactly.

Hope that helps; happy to discuss more.

PR #2385, opened 2022-04-22
[#181973860] PDF: 12a should be the same as 12c

alaroia-cfa:
From what I can tell, the PDF was missing some data, but the XML works a different way and didn't have this problem.


embarnard:
> From what I can tell, the PDF was missing some data, but the XML works a different way and didn't have this problem.

Do we need to add `TotalAdjustmentsToIncomeAmt12c: @tax_return.standard_deduction,` to the `irs1040_pdf.rb` answers hash in `has_for_pdf`?

embarnard (approved):

PR #2386, opened 2022-04-22
PDF: Add line 28 is missing and update lines 32, 33, 34 and 35a

embarnard:


app/lib/irs1040_pdf.rb
```diff
@@ -30,11 +31,12 @@ def hash_for_pdf
         AdjustedGrossIncomeAmt11: 1,
         TotalItemizedOrStandardDedAmt12a: @tax_return.standard_deduction,
         TaxableIncomeAmt15: 0,
+        AdditionalChildTaxCreditAmt28: @benefits.outstanding_ctc_amount,
         RecoveryRebateCreditAmt30: @benefits.claimed_recovery_rebate_credit,
-        RefundableCreditsAmt32: @benefits.claimed_recovery_rebate_credit,
-        TotalPaymentsAmt33: @benefits.claimed_recovery_rebate_credit,
-        OverpaidAmt34: @benefits.claimed_recovery_rebate_credit,
-        RefundAmt35: @benefits.claimed_recovery_rebate_credit,
+        RefundableCreditsAmt32: refundable_credits,
```
bytheway875, line 36: since these values are calculated on both the PDF and on the XML itself, it likely makes sense to create a calculated value on the benefits object called 'total_refundable_credits'.

spec/lib/submission_builder/ty2021/documents/irs1040_spec.rb
```diff
@@ -94,15 +94,15 @@
         expect(xml.at("TaxableIncomeAmt").text).to eq "0"
 
         # Line 28: remaining amount of CTC they are claiming (as determined in flow and listed on 8812 14i
-        expect(xml.at("RefundableCTCOrACTCAmt").text).to eq "0" # TODO: replace this when we calculate this number
+        expect(xml.at("RefundableCTCOrACTCAmt").text).to eq "3600"
```
alaroia-cfa, line 97: This 3600 number seems arbitrary. Maybe it makes sense to mock `Efile::BenefitsEligibility` in this spec so the 3600 can come from somewhere; same strategy as 1040_pdf_spec. Not a blocker, but I think a very good idea. What do you think?

alaroia-cfa (approved):
I'd like to see more clarity in the irs1040_spec about some data that's in there. No blockers!

spec/lib/submission_builder/ty2021/documents/irs1040_spec.rb
embarnard: ooo makes sense, added mocks for BenefitsEligibility!

PR #2390, opened 2022-04-25
Update copy for "SSN not valid for employment" checkbox in all places where it appears

jenny-heath:
image


bytheway875:
Is this copy change in place of the other recent change that had a reveal with more details?

jenny-heath:
@bytheway875 yeah it is, the original PR had the reveal and then when we asked some clarifying questions we found out the reveal was no longer wanted

PR #2397, opened 2022-04-26
Create table for safe and denied domains

bytheway875:


app/models/archived/intake/ctc_intake_2021.rb
```diff
@@ -198,7 +198,7 @@
 #  spouse_signature_pin_at                              :datetime
 #  spouse_suffix                                        :string
 #  spouse_tin_type                                      :integer
-#  spouse_was_blind                                     :integer          default(0), not null
```
bytheway875, line 201: im betting that these migrations ran for the implementer before the enums were added, which is why me running the migrations here changed the values

caitbot (approved):
I wasn't sure what a domain was but based on the specs, this seems to indicate email domains, with that context, this looks good to me.

I am curious about what `activated_at` specifies

bytheway875:
@caitbot activated_at allows us to toggle a domain rule off or on without deleting the entry completely. Essentially a soft delete.

PR #2402, opened 2022-04-26
Add rake task: backfill missing service_type on tax_returns

jenny-heath:


tgrathwell:
chose to do `TaxReturn.where(service_type: nil).update_all(service_type: 'online_intake')` instead

because the nature of the bug that caused this meant that it would only have been happening for `online_intake` clients

PR #2403, opened 2022-04-26
Fraud updates

bytheway875:
Some cleanup of fraud things to fix some issues I've encountered with it:

- canonical email should not fail when email is nil
- we need to be able to use Intake::CtcIntake to scope down comparing phone numbers to only other ctc intakes instead of comparing to Gyr intakes too. need to update some of the model comparison code to allow for that.
- duplications allow for a lower threshold before we start penalizing based on points, but that was not implemented.

### DIFF
--- a/app/models/fraud/indicator.rb
+++ b/app/models/fraud/indicator.rb
@@ -111,7 +111,7 @@ module Fraud
     end
 
     def calculate_points_from_count(count)
-      return 0 unless count >= threshold
+      return 0 if threshold.present? && count < threshold
       return points unless multiplier.present?
 
       applied_count = (count - 1)
--- a/app/models/fraud/indicator.rb
+++ b/app/models/fraud/indicator.rb
@@ -72,7 +72,6 @@ module Fraud
     def duplicates(references)
       # skip this rule if we can't check against the reference object
       return passing_response if references[reference].blank?
-
       duplicate_ids = DeduplificationService.duplicates(references[reference], *indicator_attributes, from_scope: query_model_name.constantize).pluck(:id)
       points = calculate_points_from_count(duplicate_ids.count)
       duplicate_ids.present? ? response(points, duplicate_ids.uniq) : passing_response
@@ -97,7 +96,8 @@ module Fraud
     private
 
     def scoped_records(references)
-      self_reference = query_model_name.underscore == reference
+      compare_model_name = query_model_name.split("::")[0]
+      self_reference = compare_model_name.underscore == reference
       scope = self_reference ? { id: references[reference].id } : { reference => references[reference] }
       query_model_name.constantize.where(scope)
     end
--- a/app/models/fraud/indicator.rb
+++ b/app/models/fraud/indicator.rb
@@ -96,7 +96,7 @@ module Fraud
     private
 
     def scoped_records(references)
-      compare_model_name = query_model_name.split("::")[0]
+      compare_model_name = query_model_name.include?("Intake::") ? "Intake" : query_model_name # match from child intake type
       self_reference = compare_model_name.underscore == reference
       scope = self_reference ? { id: references[reference].id } : { reference => references[reference] }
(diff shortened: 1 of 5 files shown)


app/models/fraud/indicator.rb
```diff
@@ -73,7 +73,8 @@ def duplicates(references)
       # skip this rule if we can't check against the reference object
       return passing_response if references[reference].blank?
 
-      duplicate_ids = DeduplificationService.duplicates(references[reference], *indicator_attributes, from_scope: query_model_name.constantize).pluck(:id)
+      from_scope = query_model.respond_to?(:accessible_intakes) ? query_model.accessible_intakes : query_model
```
bytheway875, line 76: for duplicate checks on intakes, we want to narrow things down to only accessible_intakes, or else we'll be overcounting duplicates by including intakes where the client entered a phone number or email but never verified it, for example.

app/views/shared/_fraud_score_snapshot.html.erb
```diff
@@ -9,8 +9,9 @@
   <div class="field-display spacing-below-5">
     <span class="form-question"><%= key.humanize %>:</span>
     <span class="label-value"><%= value["points"] %> points</span>
-    <br>
-    <span style="font-size: smaller">
+    <% if value["data"].present? %>
```
bytheway875, line 12: this prevents the descriptor for the values from displaying if there is no data

app/models/fraud/indicator.rb
```diff
@@ -96,10 +97,15 @@ def equals(references)
 
     private
 
+    def query_model
+      query_model_name.constantize
+    end
+
     def scoped_records(references)
-      self_reference = query_model_name.underscore == reference
+      compare_model_name = query_model_name.include?("Intake::") ? "Intake" : query_model_name # match from child intake type
```
bytheway875, line 105: We often have checks that are scoped down to Intake::CtcIntake, which is still a self-reference to intake

app/models/fraud/indicator.rb
```diff
@@ -111,6 +117,7 @@ def riskylist
     end
 
     def calculate_points_from_count(count)
+      return 0 if threshold.present? && count < threshold
```
bytheway875, line 120: This ensures that we only attribute points for a duplicate infraction if the number of infractions is greater than the indicated threshold

app/models/fraud/indicator.rb
caitbot: that makes sense

app/models/fraud/indicator.rb
caitbot: So in the database the model name is `Intake` and so we want child classes to match that?

spec/models/fraud/indicator_spec.rb
```diff
@@ -66,6 +66,22 @@ def self.riskylist
...
+      let(:intake) { create :intake }
+      let(:query_double) { double }
+      before do
+        indicator.reference = "intake"
+        indicator.query_model_name = "Intake::CtcIntake"
+        allow(Intake::CtcIntake).to receive(:where).and_return query_double
+        allow(query_double).to receive(:average)
+      end
+
+      it "builds the scoped query as a self-reference" do
```
caitbot, line 79: cool this makes sense and clarifies things for me

caitbot (approved):
the tests were so helpful for understanding these changes, this all makes sense, thanks for including good test coverage

app/models/fraud/indicator.rb
bytheway875: the object passed into the references block in execute is always going to be "intake", but the reference table is sometimes/often going to be Intake::CtcIntake because we only want to look at Ctc intakes. This would blow up if we tried to compare intake_ctc_intake to intake by attempting to create a query that looks like Intake::CtcIntake.where(intake: intake), but it needs to be Intake::CtcIntake.where(id: intake.id)

PR #2404, opened 2022-04-26
Fix eligibility head of household bug

bytheway875:
When toggling from married filing jointly to head of household, the eligibility calculations would still use married filing jointly logic to determine the calculation (tin type and some of the spouse questions would be answered). This will raise an error for the statuses we do not support (so that we dont run into this error again in the future if we add another filing status), and makes the logic more explicit.

### DIFF
--- a/app/lib/efile/benefits_eligibility.rb
+++ b/app/lib/efile/benefits_eligibility.rb
@@ -86,7 +86,11 @@ module Efile
     private
 
     def rrc_eligible_filer_count
-      return intake.primary_tin_type == "ssn" ? 1 : 0 if tax_return.filing_status_single?
+      raise "unsupported filing type" unless tax_return.filing_status.in? %w[married_filing_jointly head_of_household single]
+      
+      unless tax_return.filing_status_married_filing_jointly?
+        return intake.primary_tin_type == "ssn" ? 1 : 0
+      end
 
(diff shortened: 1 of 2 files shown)


app/lib/efile/benefits_eligibility.rb
```diff
@@ -86,7 +86,11 @@ def claimed_recovery_rebate_credit
     private
 
     def rrc_eligible_filer_count
-      return intake.primary_tin_type == "ssn" ? 1 : 0 if tax_return.filing_status_single?
+      raise "unsupported filing type" unless tax_return.filing_status.in? %w[married_filing_jointly head_of_household single]
```
alaroia-cfa, line 89: It took me a while to understand this code. I think it'd be clearer with a `case` on the outside, like

```
case tax_return.filing status
when "head_of_household", "single"
  something
when "married_filing_jointly"
  something
else
  raise "unsupported filing type"
end
```

100% not a blocker, just a FYI.

alaroia-cfa (approved):
Yay - good fix! One non blocking comment about readability.

alaroia-cfa:
lol the poor test suite got rekt

app/lib/efile/benefits_eligibility.rb
bytheway875: good call -- i can do that quickly :D

PR #2405, opened 2022-04-26
Add driver's license pages to CTC intake

caitbot:
**This PR:**
- adds the controllers, forms, and views for `DriversLicense` and `SpouseDriversLicense` pages in the intake.
- adds the model `DriversLicense` which stores all the data from those pages
- adds two foreign keys to CTC intakes, which reference the drivers license table, one for primary and one for spouse

**Here are what the pages look like**
image
image


app/models/intake/ctc_intake.rb
```diff
@@ -308,6 +312,11 @@ class Intake::CtcIntake < Intake
   has_one :bank_account, inverse_of: :intake, foreign_key: :intake_id, dependent: :destroy
   accepts_nested_attributes_for :bank_account
 
+  belongs_to :primary_drivers_license, class_name: "DriversLicense", optional: true
+  belongs_to :spouse_drivers_license, class_name: "DriversLicense", optional: true
+  accepts_nested_attributes_for :primary_drivers_license
```
bytheway875, line 317: i think we can inline all of the accepts_nested_attributes things into one line with multiple arguments, if we want to

app/models/drivers_license.rb
```diff
@@ -0,0 +1,16 @@
...
+#  expiration_date :date             not null
+#  issue_date      :date             not null
+#  license_number  :string           not null
+#  state           :string           not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+class DriversLicense < ApplicationRecord
+  has_one :intake_as_primary, class_name: "Intake", foreign_key: "primary_drivers_license_id"
+  has_one :intake_as_spouse, class_name: "Intake", foreign_key: "spouse_drivers_license_id"
```
caitbot, line 15: I would love feedback on this approach. it feels like this is maybe a weird thing to do, but I'm not expecting someone to go from driver's license to intake, as much as I'm expecting someone to look at an intake and want the data on their driver's license.

if anyones got better ideas for this, or thinks this is fine and great, let me know. 💕

app/models/intake/ctc_intake.rb
caitbot: yes yes, we def can, I went ahead and did that. for some reason rubymine complains about the arguments and so I wasnt sure if it was possible, but looks like it is.

app/controllers/ctc/questions/drivers_license_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Ctc
+  module Questions
+    class DriversLicenseController < QuestionsController
+      include AuthenticatedCtcClientConcern
+
+      layout "intake"
+
+      def show?; end
```
bytheway875, line 8: I feel like this would make the return value of this method nil and thus make it never show... this overrides the super value of true. Is this purposeful, and if so how do clients get to this page?

app/forms/ctc/base_spouse_form.rb
```diff
@@ -1,6 +1,6 @@
 module Ctc
   class BaseSpouseForm < QuestionsForm
-    include BirthDateHelper
+    include DateHelper
```
bytheway875, line 3: nice

bytheway875:
@caitbot This is such a nitty little nitpick, but it always bothers me how much additional space there is between the help text and the title and the inputs on the date fields -- its a lot more than the provided mocks in every instance, as are the gaps between the boxes themselves. I wonder if we should file another ticket just to clean up the styling to match the Figma -- that's something I would be interested in taking care of sometime soon.

app/forms/ctc/drivers_license_form.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def expiration_date
+      parse_date_params(expiration_date_year, expiration_date_month, expiration_date_day)
+    end
+
+    def issue_date_is_valid_date
+      valid_text_date(issue_date_year, issue_date_month, issue_date_day, :issue_date)
+    end
+
+    def expiration_date_is_valid_date
```
bytheway875, line 58: i wonder if its worth adding to this error something like "if a day is not provided, please specify 1" since we do seem to think that some places only use a month and year

app/forms/ctc/spouse_drivers_license_form.rb
```diff
@@ -0,0 +1,62 @@
+module Ctc
```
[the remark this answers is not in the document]
bytheway875: im trying to think about how product is going to specify changes to these pages in the future. i feel like they're probably juist going to say "drivers license page" and its going to be up to the implementor to understand that the spouse and primary pages need to match. that's probably fine

app/models/drivers_license.rb
bytheway875: i actually like this. like you said i double we'll be using it much but it makes things fairly clear to me

app/views/ctc/questions/spouse_drivers_license/edit.html.erb
```diff
@@ -0,0 +1,33 @@
+<% content_for :page_title, t("views.ctc.questions.spouse_drivers_license.title") %>
```
bytheway875, line 1: i do actually think that it makes sense to maybe use a partial for this, but its not a sticking point.

bytheway875 (approved):
Approving! this looks great

Just a couple of questions, mostly about UI

app/controllers/ctc/questions/drivers_license_controller.rb
caitbot: huh yeah thats a good point, I don't know why I did that. the funny thing is the show method is actually `def self.show?` and so the reason this was working is that ultimately it was meaningless hahahaha

app/forms/ctc/drivers_license_form.rb
caitbot: @yvonnefong was adamant that most (all?) states put a day, but I can follow up with that suggestion! thats a pretty simple way to handle it

PR #2406, opened 2022-04-26
CTC: only include 8812 form bundled with 1040 when claiming CTC

wschaefer:


bytheway875:
@wschaefer I think this is only going to remove the 8812 from the PDF, but we also need to change the logic so that the XML only includes the 1040 as well.

Thats happening in the SubmissionBuilder::Ty2021::Return1040 attached documents class method, but I'm wondering if we're going to need to refactor how that code works since the class method doesn't have any knowledge of the individual filing situation...

bytheway875:
@wschaefer i cant think off the top of my head how i'd accomplish refactoring it so that if works... we can look at it together at 2:30 if you'd like to?

spec/models/efile_submission_spec.rb
```diff
@@ -363,13 +363,31 @@
...
+        expect(doc.document_type).to eq(DocumentTypes::Form1040.key)
+        expect(doc.tax_return).to eq(submission.tax_return)
+        expect(doc.upload.blob.download).not_to be_nil
+      end
+    end
+
+    context "the filer is not claiming CTC (line 28 = $0 or missing)" do
+      it "generates and stores just the 1040 (does not generate the 8812)" do
+        expect { submission.generate_filing_pdf }.to change(Document, :count).by(1)
+        expect(Irs8812Ty2021Pdf).not_to have_received(:new)
```
embarnard, line 384: yeah i wonder if there is a way to count the number of pages in the doc as well but this is probably sufficient as a check

embarnard (approved):
Looks good!

wschaefer (commented):
paired w/shannon

PR #2407, opened 2022-04-26
Bug fix to all current relationships shown as free form option

embarnard:


app/helpers/relationships_helper.rb
```diff
@@ -3,7 +3,7 @@ module RelationshipsHelper
   # to the list if necessary.
   def dependent_relationship_options(current_relationship: nil)
     options = Efile::Relationship.all.map { |relationship| [I18n.t("general.dependent_relationships.#{relationship.id}"), relationship.id] }
-    if current_relationship.present? && Efile::Relationship.find(current_relationship.to_sym).nil?
+    if current_relationship.present? && Efile::Relationship.find(current_relationship).nil?
```
embarnard, line 6: Efile::Relationship actually returns an array of strings so it was never matching to symbols and therefore always returning nil

alaroia-cfa: Egad

alaroia-cfa (approved):
Less is more!

PR #2412, opened 2022-04-26
Use ssn to check for duplicates on ctc clients

bytheway875:
Convert CTC clients to using SSN for duplicate check, but leave accessible intakes to those that have verified contact info.

This means that clients WILL be able to create multiple intakes with the same SSN, but they will not show up in the hub or block clients from creating another intake until they are also able to log in.

### DIFF
--- a/app/controllers/concerns/anonymous_intake_concern.rb
+++ b/app/controllers/concerns/anonymous_intake_concern.rb
@@ -7,10 +7,6 @@ module AnonymousIntakeConcern
 
   private
 
-  def redirect_if_duplicate_ctc_client
-    redirect_to questions_returning_client_path if current_intake.has_duplicate?
-  end
-
   def set_show_client_sign_in_link
     @show_client_sign_in_link = true
   end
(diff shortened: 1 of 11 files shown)


caitbot (approved):
The description was really helpful -- to restate my understanding of the behavior change, now clients will be able to access intake, regardless of duplication, up until the point of legal consent. At legal consent, we either log them in, or let them know they are returning. BEFORE we were always blocking people from filling out intake as soon as they got far enough along to have entered any duplicated data (or for us to detect they were a duplicate).

I like this approach. It makes sense that we would only block upon login. Sweet!

bytheway875:
@caitbot sort of -- last year, we used to tell people they were a duplicate right after they entered a duplicated email or phone number.

But now, we want to base it on SSN. The problem, though, is that even if someone enters a duplicated SSN as someone did on another intake, some of those intakes with duplicate SSNs might have been abandoned before they gave us an email or a phone number. And if that's the case, if we told them "hey we have a duplicate, go login", some people would be "stuck." This change says "only look at and consider as potential duplicates intakes where the client provided information to login with"

PR #2414, opened 2022-04-27
Only display spouse indicators if MFJ

bytheway875:
The other spouse indicators on the header and in the 1040 seem to already have the if filing_status guard

### DIFF
--- a/app/lib/submission_builder/shared/return_header1040.rb
+++ b/app/lib/submission_builder/shared/return_header1040.rb
@@ -34,9 +34,7 @@ module SubmissionBuilder
           }
           xml.SelfSelectPINGrp {
             xml.PrimaryBirthDt date_type(intake.primary_birth_date)
-            if tax_return.filing_jointly?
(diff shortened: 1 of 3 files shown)


app/lib/submission_builder/shared/return_header1040.rb
```diff
@@ -34,9 +34,7 @@ def document
           }
           xml.SelfSelectPINGrp {
             xml.PrimaryBirthDt date_type(intake.primary_birth_date)
-            if tax_return.filing_jointly?
```
caitbot, line 37: hahhaha was this a double check for `filing_jointly?` thats kind of funny

caitbot (approved):
lgtm

app/lib/submission_builder/shared/return_header1040.rb
bytheway875: yes and then i gitblamed and both were committed by me in the same commit. its very silly

PR #2416, opened 2022-04-27
Rename column from deny to risky per fraud rule

bytheway875:
The existing implementation to create rules requires "riskylist" but I accidentally called it deny


db/migrate/20220427204338_change_deny_to_risky_on_fraud_indicator_domain.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDenyToRiskyOnFraudIndicatorDomain < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured { rename_column :fraud_indicators_domains, :deny, :risky }
```
caitbot, line 3: love to see it

caitbot (approved):
nice

PR #2429, opened 2022-04-29
CTC copy updates

wschaefer:


jenny-heath (approved):

app/controllers/ctc/questions/overview_controller.rb
```diff
@@ -17,7 +17,9 @@ def after_update_success
         send_mixpanel_event(event_name: "ctc_started_flow")
       end
 
-      def illustration_path; end
+      def illustration_path
+        "wages.svg"
```
bytheway875, line 21: interestingly, this is a duplciate of hand-holding-check, but it's blue. I wondering if we should call it hand-holding-check-blue so that its easier to find and compare later, but then again none of the svgs are really easy to recover later haha

PR #2430, opened 2022-04-29
Refactor controller to have a base controller so we can build off of it, and add controller tests

bytheway875:


caitbot (approved):
Nice refactor, this makes sense and looks good to me.

PR #2431, opened 2022-04-29
Hub all clients table: wrap cell contents without making all cells too narrow

jenny-heath:
this styling is a little far-reaching (as it applies to most table cells in the hub) and as such feels like it has the potential for some unexpected consequences when applied to real data. i tried to think of edge cases but would appreciate it if any reviewers look for ones i may have missed!

before:
Screen Shot 2022-04-29 at 3 12 03 PM

after:
Screen Shot 2022-04-29 at 3 11 46 PM

### DIFF
--- a/app/assets/stylesheets/components/_index-table.scss
+++ b/app/assets/stylesheets/components/_index-table.scss
@@ -119,7 +119,7 @@ It expects to be the primary content of the given page (full width).
 
   &__cell {
     padding: 0.8rem $s25/2;
-    white-space: nowrap;
+    min-width: 15rem;
 
     &:first-child {
       padding-left: 3rem;
@@ -131,6 +131,10 @@ It expects to be the primary content of the given page (full width).
     &--center {
       text-align: center;
     }
+  }
+
+  &__cell.narrow {
+    min-width: unset;
   }
 
   &__vertical_align {
--- a/app/assets/stylesheets/components/_index-table.scss
+++ b/app/assets/stylesheets/components/_index-table.scss
@@ -119,7 +119,6 @@ It expects to be the primary content of the given page (full width).
 
   &__cell {
     padding: 0.8rem $s25/2;
-    min-width: 15rem;
 
     &:first-child {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/components/_index-table.scss
```diff
@@ -119,7 +119,7 @@ It expects to be the primary content of the given page (full width).
 
   &__cell {
     padding: 0.8rem $s25/2;
-    white-space: nowrap;
+    min-width: 15rem;
```
bytheway875, line 122: this feels like its going to have the opposite affect of making certain cells larger than they were before on other implementations of the index table... im somewhat concerned about that.

bytheway875: is it possible to set a max width on the narrow tabs instead, since thats where we want the behavioral change?

bytheway875: oih nevermind, i understand whats going on here. the small ones are what now have narrow applied, which allow them to shrink. Instead of calling it "narrow" can you call it like "shrink-to-size" or something like that, so that it indicates there is no explicit width?

I still wonder if it's possible to employ a max-width on the cell and achieve the same thing, then add a class called `grow` that allows cells with that class to have an unset max-width.

bytheway875: Hmm, I think removing the min-width altogether would work as well, and then just style any `th` elements where their children need more space as needed (th class="index=table__header timestamp") where the timestamp class sets a min width, since that seems to be the one mostly affected by the th being shorter than the contained, wrappable content

bytheway875: OR! add a timestamp class to the timestamp data that indicates nowrap, since really we dont want to break between the time and AM/PM

OR! See if design would be okay with us changing the timestamp to be Feb 10 5:39PM without a space between the time and the AM/PM, so the wrapping doesnt make that field break.

PR #2437, opened 2022-05-02
Fix dependent eligibility bug

bytheway875:
The EfileSubmissionDependent 'prequalifying' logic looked at the wrong attribute. Instead of looking at qualifying_ctc, it was looking at qualifying_child.

### DIFF
--- a/app/lib/efile/dependent_eligibility/child_tax_credit.rb
+++ b/app/lib/efile/dependent_eligibility/child_tax_credit.rb
@@ -30,13 +30,11 @@ module Efile
       private
 
       def is_qualifying_child?
-        return dependent.qualifying_child if dependent.is_a? EfileSubmissionDependent
-
         (@child_eligibility || Efile::DependentEligibility::QualifyingChild.new(dependent, tax_year)).qualifies?
       end
 
       def prequalifying_attribute
-        "qualifying_child"
+        "qualifying_ctc"
       end
     end
   end
--- a/app/lib/efile/dependent_eligibility/eip_one.rb
+++ b/app/lib/efile/dependent_eligibility/eip_one.rb
@@ -27,6 +27,8 @@ module Efile
       end
 
       def is_qualifying_child?
+        return dependent.qualifying_child if dependent.is_a? EfileSubmissionDependent
+
         (@child_eligibility || Efile::DependentEligibility::QualifyingChild.new(dependent, tax_year)).qualifies?
       end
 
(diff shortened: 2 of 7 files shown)


app/lib/efile/dependent_eligibility/child_tax_credit.rb
```diff
@@ -30,13 +30,11 @@ def under_ctc_age_limit?
       private
 
       def is_qualifying_child?
```
caitbot, line 32: the confusion here was probably that this is called `is_qualifying_child?` which is a slightly different concept than `is_ctc_qualifying_child?` consider renaming to that for precision/clarities sake. 

it will be mildly redundant with the service name, and depending on how we are calling eligibility might not entirely work, so this is a non-blocking suggestion.

app/lib/efile/dependent_eligibility/eip_one.rb
```diff
@@ -27,6 +27,8 @@ def tax_year_2020?
       end
 
       def is_qualifying_child?
+        return dependent.qualifying_child if dependent.is_a? EfileSubmissionDependent
```
caitbot, line 30: why is it possible to add this here? is EIP1 qualification only that you have to be a QC?

caitbot: might be worth a comment?

caitbot (approved):
approved! left a non-blocking naming suggestion, good catch on the birthday

app/lib/efile/dependent_eligibility/child_tax_credit.rb
bytheway875: this line is ONLY to see if the client is a qualifying child, flat out. The line that was removed was intended to look at the ESD to see if they are a qualifying_child, which is sort of redundant since this logic will never run if we're looking at the qualifying attribute (qualifying_ctc) when it is an efile submission dependent object.

app/lib/efile/dependent_eligibility/eip_one.rb
bytheway875: because if it is an efile submission dependent, the qualifying_child attribute will have been set to true or false based on a previous run of the qualifying logic -- we can look to that instead of running the calculation again

app/lib/efile/dependent_eligibility/child_tax_credit.rb
caitbot: gotcha! i read this improperly, that makes more sense

PR #2439, opened 2022-05-02
Add safe domains and timezones pages

bytheway875:


app/helpers/fraud_indicators_helper.rb
```diff
@@ -11,11 +11,21 @@ def link_to_indicator_list(fraud_indicator)
...
     table = fraud_indicator.list_model_name.split("::").last.downcase
+    path_with_type = "hub_#{type}_#{table}s_path"
+    path_without_type = "hub_#{table}s_path"
 
-    if respond_to?("hub_#{type}_#{table}s_path")
-      link_to send("hub_#{type}_#{table}s_path") do
-        content_tag :i, "list_alt", class: "icon-"
-      end
-    end
+    matching_path = nil
```
bytheway875, line 17: the timezones page isn't prefixed with risky vs safe, so this needed a bit of a logic update

app/models/fraud/indicators/timezone.rb
```diff
@@ -18,6 +18,10 @@ class Timezone < ApplicationRecord
       def self.safelist
         all.pluck(:name).push(nil)
       end
+
+      def active
```
bytheway875, line 22: The form checkbox helper requires a boolean-esque value -- using activated_at directly doesnt work correctly. This and the domain model has this method implemented.

app/helpers/fraud_indicators_helper.rb
```diff
@@ -11,11 +11,21 @@ def link_to_indicator_list(fraud_indicator)
...
+    matching_path = path_with_type if respond_to?(path_with_type)
```
caitbot, line 18: ok this all makes sense to me, but I was surprised when I looked at line 6 in this file and saw that the `indicator_type` `not_in_safelist` means the `type` here is `safe`. am I missing some context?

app/views/hub/fraud_indicators/_form.html.erb
```diff
@@ -1,6 +1,6 @@
-<h4 class="spacing-above-35 spacing-below-15">New <%= @resource_name %></h4>
+<h4 class="spacing-above-35 spacing-below-15">New <%= resource_name %></h4>
 
-<%= form_for @resource, url: url_for(locale: :en), builder: VitaMinFormBuilder, remote: true do |f| %>
+<%= form_for resource, url: url_for(locale: :en), builder: VitaMinFormBuilder, remote: true do |f| %>
```
caitbot, line 3: so on the BaseController it looks like `resource` is still an instance variable (`@resource`) and `resource_name` is a helper method. should that be updated here as well (ie `resource`)?

caitbot, line 3: so on the BaseController it looks like `resource` is still an instance variable (`@resource`) and `resource_name` is a helper method. should that be updated here as well?

spec/controllers/hub/fraud_indicators/safe_domains_controller_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    context "when logged in" do
+
+      before do
+        sign_in user
+        Fraud::Indicators::Domain.create(risky: true, name: "something.com", activated_at: DateTime.now)
+        Fraud::Indicators::Domain.create(safe: true, name: "something-else.com", activated_at: nil)
+        Fraud::Indicators::Domain.create(safe: true, name: "another-thing.com", activated_at: DateTime.now)
+      end
+
+      it "assigns a list of all risky Fraud::Indicator::Domain objects" do
```
caitbot, line 18: should be `all safe Fraud::Indicators...`

caitbot, line 18: should be `all safe Fraud::Indicators...`

spec/controllers/hub/fraud_indicators/timezones_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      { id: indicator.id }
+    end
+    it_behaves_like :a_post_action_for_authenticated_users_only, action: :update
+
+    context 'when logged in' do
+      before do
+        sign_in user
+      end
+
+      it "toggles from activated to not activated" do
```
caitbot, line 78: technically this is toggling from `not activated to activated`

caitbot, line 78: technically this is toggling from `not activated to activated`

caitbot (approved):
Cool! For the most part this looks good, I left some comments and questions.

app/helpers/fraud_indicators_helper.rb
bytheway875: if the indicator type is not in safelist, you have to maintain a list of safe things that it can check against to see if the value is not in that list.

PR #2440, opened 2022-05-02
Update GYR homepage

caitbot:
The commit messages tell the story more granularly. But here is a series of before/afters.

### Top of the page
**Before**
image

**After**
image

### Service comparison
**Before:**
image

**After:**
image

### FAQ
**Before:**
image

**After:**
image

### Footer
**Before:**
image

**After:**
image

### DIFF
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -89,7 +89,7 @@ module Questions
     private
 
     def redirect_in_offseason
-      redirect_to root_path unless open_for_intake?
+      redirect_to root_path unless open_for_gyr_intake?
     end
 
     def redirect_if_completed_intake_present
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -2,7 +2,7 @@ module Questions
   class QuestionsController < ApplicationController
     before_action :redirect_in_offseason
     before_action :redirect_if_completed_intake_present
-    before_action :set_current_step
+    before_action :set_current_step, :set_in_intake_flow
     delegate :form_name, to: :class
     delegate :form_class, to: :class
 
@@ -96,6 +96,10 @@ module Questions
       if current_intake && current_intake.completed_at.present?
         redirect_to portal_root_path
       end
+    end
+
+    def set_in_intake_flow
+      @in_intake_flow = true
     end
 
     def after_update_success; end
--- a/app/views/public_pages/home.html.erb
+++ b/app/views/public_pages/home.html.erb
@@ -3,15 +3,7 @@
 <% end %>
 
 <% content_for :notice do %>
-  <% if open_for_intake? %>
-    <div class="slab slab--banner">
-      <div class="grid">
-        <div class="grid__item">
-          <%= t('views.public_pages.home.documents_warning_html') %>
-        </div>
-      </div>
-    </div>
-  <% else %>
+  <% unless open_for_intake? %>
     <div class="slab slab--banner">
       <div class="grid">
         <div class="grid__item">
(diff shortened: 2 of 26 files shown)


app/controllers/questions/questions_controller.rb
```diff
@@ -98,6 +98,10 @@ def redirect_if_completed_intake_present
       end
     end
 
+    def set_in_intake_flow
+      @in_intake_flow = true
```
bytheway875, line 102: what does this do?

bytheway875: also, you could maybe define this as a helper method instead 

```
helper_methods :in_intake_flow

def in_intake_flow
  true
end
```

then, if this method is trying to get used somewhere and its not defined it will raise an error

app/views/public_pages/home.html.erb
```diff
@@ -36,7 +41,11 @@
         </h1>
         <p>
           <strong>
-            <%= open_for_intake? ? t("views.public_pages.home.subheader") : t("views.public_pages.home.subheader_offseason") %>
+            <% if open_for_intake? %>
```
bytheway875, line 44: it might be helpful to rename `open_for_intake?` to `open_for_gyr_intake?` so that this is less confusing

app/views/public_pages/home.html.erb
```diff
@@ -116,17 +125,16 @@
...
+        <h2 class="h3 spacing-below-0"><%= t("views.public_pages.home.faq.header") %></h2>
+        <ul class="list--bulleted">
+          <li class="spacing-below-15"><%= link_to t("views.public_pages.faq.question_groups.stimulus.how_do_i_get_the_stimulus_payments.question"), faq_question_url(host: MultiTenantService.new(:gyr).host, section_key: :stimulus, question_key: "how_do_i_get_the_stimulus_payments"), class: "link--subtle" %></li>
+          <li class="spacing-below-15"><%= link_to t("views.public_pages.faq.question_groups.should_i_file_a_tax_return.what_are_the_potential_benefits_of_filing_a_tax_return.question"), faq_question_url(host: MultiTenantService.new(:gyr).host, section_key: :should_i_file_a_tax_return, question_key: "what_are_the_potential_benefits_of_filing_a_tax_return"), class: "link--subtle" %></li>
+          <li class="spacing-below-0"><%= link_to t("views.public_pages.faq.question_groups.should_i_file_a_tax_return.am_i_a_nonfiler.question"), faq_question_url(host: MultiTenantService.new(:gyr).host, section_key: :should_i_file_a_tax_return, question_key: "am_i_a_nonfiler"), class: "link--subtle" %></li>
```
bytheway875, line 132: oh, i actually didnt know you could specify a host on a named route like this. that's nice and a bit cleaner than having to build it completely from `url_for`

bytheway875: but ... if this link is on GYR I feel like you don't need to specify the host at all?

bytheway875 (approved):
A couple of suggestions and questions

app/views/public_pages/home.html.erb
caitbot: I might not need to, I'll give it a try

app/views/public_pages/home.html.erb
caitbot: I like that, it was a touch confusing

app/controllers/questions/questions_controller.rb
caitbot: I don't really want it to be defined everywhere, what I really need is a way to specify that the client has started the questions flow by clicking "Getting started" because now we show "Getting started" in the top navigation bar, and we don't want it to show up there if they are already IN the questions flow. 

However, theres no consistency between ctc or gyr around creating objects (or creating them on the first page) so you can't really check for presence of data, so I decided to create this little method to specify that youre on a question controller page.

app/views/public_pages/home.html.erb
caitbot: You do not!

PR #2446, opened 2022-05-02
Use upcase_first instead of titleize for dependent name

bytheway875:


bytheway875:
titleize will take something like "bartQC" and turn it into "Bart QC", which I guess we do not want.

I think this might be better as a helper method so that we can do it consistently, but doing upcase_first if we were to send a full name like "bart simpson" in would end up with "Bart simpson", so it just seemed most clear to make the change in place.

caitbot (approved):
This makes a lot of sense, good change. I agree with your comment about maybe having a helper for this in the future

PR #2448, opened 2022-05-02
CTC: update spouse prior tax year questions

wschaefer:

### DIFF
--- a/app/forms/ctc/spouse_filed_prior_tax_year_form.rb
+++ b/app/forms/ctc/spouse_filed_prior_tax_year_form.rb
@@ -6,12 +6,6 @@ module Ctc
 
     def save
       attributes = attributes_for(:intake)
-      attributes[:spouse_prior_year_agi_amount] = 1 if %w[
-        filed_non_filer_separate
-      ].include?(spouse_filed_prior_tax_year)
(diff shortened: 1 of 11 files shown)


bytheway875:
@wschaefer can you take a look at the method `spouse_prior_tax_year_agi` as used on the return_header1040.rb and confirm that that logic is still correct based on the updates in this PR?

jenny-heath:
possible changes to make to the portal (later?):
- display spouse prior year AGI if filing jointly _this year_
- allow client to edit spouse prior year AGI if filed separately and spouse filed full return _last year_ (in other words, only allow to edit when we are not inferring the amount ourselves)

wschaefer:
@bytheway875 Thanks! We also liked the idea of removing the duplication but ended up going the other direction (not saving extra values for the amount) instead of removing the method that calculates the amount.

Our thought was that it would be harder to keep the data accurate if we updated spouse_prior_year_agi_amount with calculated amounts (ie if we set it to be primary_prior_year_agi_amount and then the primary amount changes we'd have to remember to update the spouse amount). If we just keep spouse_prior_year_agi_amount as the amount the client enters (if they do) and keep spouse_filed_prior_tax_year then its easy to always get the right amount. We moved that logic into a method on the intake so we can use it more places.

PR #2453, opened 2022-05-03
Fix small CTC bugs

embarnard:


config/locales/en.yml
```diff
@@ -2303,7 +2303,7 @@ en:
           - you want to file Form 8332 in order to claim a child who does not live with you
           - you are claiming a qualifying relative under a “multiple support agreement” as defined by the IRS
           - you are not claiming any children for the Child Tax Credit this year, but you received Advance Child Tax Credt payments in 2021
-          - you bought or sold cryptocurrency
+          - you bought or sold cryptocurrency in 2021
```
bytheway875, line 2306: this should be %{current_tax_year}

config/locales/en.yml
```diff
@@ -2303,7 +2303,7 @@ en:
           - you want to file Form 8332 in order to claim a child who does not live with you
           - you are claiming a qualifying relative under a “multiple support agreement” as defined by the IRS
           - you are not claiming any children for the Child Tax Credit this year, but you received Advance Child Tax Credt payments in 2021
```
bytheway875, line 2305: can you add an i to "Credt" on this line while we're at it?

PR #2454, opened 2022-05-03
Fix dupe bug when nil for duplification element

bytheway875:


app/models/fraud/indicator.rb
```diff
@@ -72,6 +72,7 @@ def in_riskylist(references)
     def duplicates(references)
       # skip this rule if we can't check against the reference object
       return passing_response if references[reference].blank?
+      return passing_response if indicator_attributes.any? { |attr| references[reference].send(attr).nil? }
```
alaroia-cfa, line 75: It's a little weird to me that this would be an `any` here -- would we want to check that all are nil? Or would we want to filter out the `indicator_attributes` whose value for the reference object is nil? Like

```
indicator_attributes = indicator_attributes.filter { |attr| references[reference].send(attr).nil? }
```

?

Let me know what you think.

alaroia-cfa (approved):
I have a question, but I don't think it's a blocker.

app/models/fraud/indicator.rb
alaroia-cfa: One idea is that it'd be nice to see a test for the scenario where there are multiple indicator_attributes, and one is nil but the other isn't, and specify the behavior for it that way.

bytheway875: i thought about this re: our current implementation of the bank account duplication. if there are multiple things we're looking to check for and one is nil, its like ... that data isnt really valid. So like if you had a bank account with only a routing number and no account number, it wouldn't be "right" to mark all of the routing number duplicates. We don't have any other instances where there are more than one indicator attribute right now, so im not sure what we'd want to do with those, so its flexible to update it in the future if needed

PR #2455, opened 2022-05-03
Change copy for acceptance email and text

bytheway875:
Copy change for email and text message when efile is accepted


alaroia-cfa (approved):
Copy that. lgtm

PR #2458, opened 2022-05-03
More GYR homepage updates from acceptance testing

caitbot:
Makes the FAQ section into a card rather than a banner, makes the "Get started" button smaller.
image

Update some copy
image

Round some corners
image

Make the sign in link less subtle
image


alaroia-cfa (approved):
Copy that.

PR #2464, opened 2022-05-04
Only count states of efile submissions that have tax returns in the current year

embarnard:


embarnard:
Screen Shot 2022-05-04 at 5 59 43 PM
This PR is a followup from this [PR](https://github.com/codeforamerica/vita-min/pull/2452/files) which only shows efile submissions in the dashboard table of submissions with tax returns in the current tax year. This story also updates the counts at the top of the `/hub/efile` page that counts the states of efile submissions so that we only count it for ones in the current tax year.

jenny-heath (approved):
gonna guess this page is still v slow after this but we'll see? would be nice to have a test but it looks like there wasn't one to begin with and it seems like the content that matters most is in the table(?), so, up to you!

PR #2465, opened 2022-05-04
Add cron job to track longest periods of time that efile submissions have been in preparing/bundling/queued states in DataDog

jenny-heath:
Discussed the strategy with Asheesh, solo'd on writing the code. Two things:
1. I chose to send a number of minutes somewhat arbitrarily, maybe it should be seconds or hours or something? Not sure. The threshold the alert will be set at is a day, but that could change in future. Also I'm somewhat concerned that "min" will be read as "minimum" so maybe I should write out "minutes" if it stays minutes.
2. The cron job running hourly is also somewhat arbitrary because I can't remember what Asheesh and I said over tuple, I just know it was supposed to be less frequent than every minute (which is how often the stats:track_metrics task runs)

UPDATE: "min" is now "minutes" and the cron job runs every 5 min

### DIFF
--- a/crontab
+++ b/crontab
@@ -1,7 +1,7 @@
 0 0 * * 0 bundle exec rake vita_providers:scrape_vita_providers
 * * * * * bundle exec rake search:refresh
 * * * * * bundle exec rake stats:track_metrics
-0 * * * * bundle exec rake stats:monitor_delayed_efile_submissions
+*/5 * * * * bundle exec rake stats:monitor_delayed_efile_submissions
 20 21 * * 1-5 bundle exec rake client_surveys:send_client_in_progress_surveys
 15 20 * * * bundle exec rake not_ready:remind
 */10 * * * * bundle exec rake efile:poll_and_get_acknowledgments
--- a/crontab
+++ b/crontab
@@ -1,6 +1,7 @@
 0 0 * * 0 bundle exec rake vita_providers:scrape_vita_providers
 * * * * * bundle exec rake search:refresh
 * * * * * bundle exec rake stats:track_metrics
+0 * * * * bundle exec rake stats:monitor_delayed_efile_submissions
 20 21 * * 1-5 bundle exec rake client_surveys:send_client_in_progress_surveys
 15 20 * * * bundle exec rake not_ready:remind
 */10 * * * * bundle exec rake efile:poll_and_get_acknowledgments
--- a/lib/tasks/stats.rake
+++ b/lib/tasks/stats.rake
@@ -19,4 +19,13 @@ namespace :stats do
       DatadogApi.gauge('efile_submissions.state_counts', count, tags: ["current_state:#{state}"])
     end
   end
+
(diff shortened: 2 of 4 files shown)


lib/tasks/stats.rake
```diff
@@ -19,4 +19,13 @@ namespace :stats do
       DatadogApi.gauge('efile_submissions.state_counts', count, tags: ["current_state:#{state}"])
     end
   end
+
+  desc "Monitor the longest periods of time that any efile submission has been in preparing, bundling, and queued"
+  task monitor_delayed_efile_submissions: :environment do
+    [:preparing, :bundling, :queued].each do |state|
+      oldest_transition_to = EfileSubmissionTransition.where(to_state: state).sort_by(&:created_at).first
```
bytheway875, line 26: I think you want to scope this to EfileSubmissionTransition where most_recent is true, or else this is going to give you the same oldest transition on every run.

jenny-heath: ohhh good catch!

crontab
```diff
@@ -1,6 +1,7 @@
 0 0 * * 0 bundle exec rake vita_providers:scrape_vita_providers
 * * * * * bundle exec rake search:refresh
 * * * * * bundle exec rake stats:track_metrics
+*/5 * * * * bundle exec rake stats:monitor_delayed_efile_submissions
```
bytheway875, line 4: does this mean every half minute?

lib/tasks/stats.rake
```diff
@@ -19,4 +19,13 @@ namespace :stats do
       DatadogApi.gauge('efile_submissions.state_counts', count, tags: ["current_state:#{state}"])
     end
   end
+
+  desc "Monitor the longest periods of time that any efile submission has been in preparing, bundling, and queued"
+  task monitor_delayed_efile_submissions: :environment do
+    [:preparing, :bundling, :queued].each do |state|
+      oldest_transition_to = EfileSubmissionTransition.where(most_recent: true, to_state: state).sort_by(&:created_at).first
```
bytheway875, line 26: gah, sorry to take another shot at this _but_ i wonder if there is a performance differnce by doing an enumerable sort_by and getting the first item vs doing a .order(:created_at).limit(1), which would only retrieve the first item by created_at out of the db instead of all of them and then sorting them.

I think we're likely fine to merge, but it might be an interesting query to try to benchmark and see which is more performant.

bytheway875 (approved):

crontab
jenny-heath: i think it means every 5 minutes, but i got it from here: https://crontab.guru/#*/5_*_*_*_*

jenny-heath: that link totally didn't work lol

lib/tasks/stats.rake
jenny-heath: mm, not sure! i honestly don't know how to test that but would be happy to pair on it if you want? i might merge this first anyway but we can refactor in a follow-up commit

PR #2466, opened 2022-05-05
Set es.yml currency unit formatting

embarnard:


alaroia-cfa (approved):
This is a good change. I wonder if we should write our own template helper called `number_to_currency` that is something like

```
def number_to_usd(number, options = {})
  ActionView::Helpers::NumberHelper.new.number_to_currency(number, options.merge(locale: :en))
end
```

so that we can avoid having to write the `locale: :en` thing everywhere.

Not a blocker. If you like the idea, feel free to run with it; if not, OK.

embarnard:
> This is a good change. I wonder if we should write our own template helper called `number_to_currency` that is something like
> 
> ```
> def number_to_usd(number, options = {})
>   ActionView::Helpers::NumberHelper.new.number_to_currency(number, options.merge(locale: :en))
> end
> ```
> 
> so that we can avoid having to write the `locale: :en` thing everywhere.
> 
> Not a blocker. If you like the idea, feel free to run with it; if not, OK.

Oooo yeah that's a good idea!

bytheway875:
There's a built in way to deal with this using locales, without having to pass in the locale. we just need to alter the currency format in the locales file: https://stackoverflow.com/questions/7237592/number-to-currency-locale-converting

bytheway875 (changes requested):
convert to using the locale file currency keyword so that we can use number_to_currency without overriding any behavior

alaroia-cfa (approved):
Flawless.

PR #2468, opened 2022-05-05
CTC flow, confirm claim dependent page: update button text

jenny-heath:
image

### DIFF
--- a/app/views/ctc/questions/dependents/child_claim_anyway/edit.html.erb
+++ b/app/views/ctc/questions/dependents/child_claim_anyway/edit.html.erb
@@ -1,20 +1,33 @@
-<% content_for :form_question, t("views.ctc.questions.dependents.child_claim_anyway.title", name: @form.dependent.first_name&.upcase_first) %>
-<% content_for :form_help_text do %>
-  <p class="text--bold"><%= t("views.ctc.questions.dependents.child_claim_anyway.help_text.p1", name: @form.dependent.first_name&.titleize) %></p>
-  <ul class="list--bulleted">
-    <% t("views.ctc.questions.dependents.child_claim_anyway.help_text.list", current_tax_year: current_tax_year, name: @form.dependent.first_name&.titleize).each do |item| %>
-      <li> <%= item %> </li>
+<% @main_heading = t("views.ctc.questions.dependents.child_claim_anyway.title", name: @form.dependent.first_name&.upcase_first) %>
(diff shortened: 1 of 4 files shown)


app/views/ctc/questions/dependents/child_claim_anyway/edit.html.erb
```diff
@@ -1,20 +1,33 @@
...
+    <%= render('components/molecules/reveal', title: t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.title")) do %>
+      <p> <%= t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.p1", current_tax_year: current_tax_year, name: @form.dependent.first_name&.titleize) %> </p>
+      <p> <%= t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.p2", name: @form.dependent.first_name&.titleize) %> </p>
+    <% end %>
 
-  <%= render('components/molecules/reveal', title: t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.title")) do %>
-    <p> <%= t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.p1", current_tax_year: current_tax_year, name: @form.dependent.first_name&.titleize) %> </p>
-    <p> <%= t("views.ctc.questions.dependents.child_claim_anyway.not_legal_parent_reveal.p2", name: @form.dependent.first_name&.titleize) %> </p>
+    <%= f.button :submit, name: "ctc_dependents_child_claim_anyway_form[claim_anyway]", value: "yes", class: "button button--full-width text--centered spacing-above-60 button--primary", "aria-describedby": "main-question" do %>
```
bytheway875, line 26: it seems like we;re doing this on several pages now ... it might make sense to turn this into a partial where we can set up these buttons in this way for other questions if we have time

jenny-heath: hm, i looked around for examples but could only find ones where these buttons are links, not submit buttons. for those ones it might make sense to have a shared partial but i think it might be a little out of scope for this story

jenny-heath: it's also possible i just didn't find the places where we have these wide submit buttons! lmk if you know of some

embarnard (approved):
Looks good!

PR #2469, opened 2022-05-05
Adjust PDF 1040 generation to rely on XML generation

wschaefer:


app/lib/irs1040_pdf.rb
```diff
@@ -1,92 +1,87 @@
...
-    @address = @submission.verified_address || @intake.address
-    @benefits = Efile::BenefitsEligibility.new(tax_return: @tax_return, dependents: @qualifying_dependents)
+    # For some PDF fields, use values from the database b/c the XML values are truncated or missing.
+    intake = submission.intake
+    @primary_first_name = intake.primary_middle_initial.present? ? "#{intake.primary_first_name} #{intake.primary_middle_initial}" : intake.primary_first_name
+    @primary_last_name = intake.primary_last_name
+    @spouse_first_name = intake.spouse_middle_initial.present? ? "#{intake.spouse_first_name} #{intake.spouse_middle_initial}" : intake.spouse_first_name
+    @spouse_last_name = intake.spouse_last_name
+    address = submission.verified_address || intake.address
+    @street_address = address.street_address
```
bytheway875, line 16: to me, it would make sense to just call all of these methods on the address object when using them instead of instantiating the additional instance variables.

app/lib/irs1040_pdf.rb
```diff
@@ -1,92 +1,87 @@
...
   def initialize(submission)
-    @submission = submission
-    @tax_return = submission.tax_return
-    @intake = submission.intake
-    @qualifying_dependents = submission.qualifying_dependents
-    @address = @submission.verified_address || @intake.address
-    @benefits = Efile::BenefitsEligibility.new(tax_return: @tax_return, dependents: @qualifying_dependents)
+    # For some PDF fields, use values from the database b/c the XML values are truncated or missing.
+    intake = submission.intake
+    @primary_first_name = intake.primary_middle_initial.present? ? "#{intake.primary_first_name} #{intake.primary_middle_initial}" : intake.primary_first_name
```
bytheway875, line 11: it also feels like these make more sense inline where they're consumed rather than defining as instance variables.

app/lib/irs1040_pdf.rb
wschaefer: Asheesh and I talked about it and felt like individually pulling out each specific entry that we take not from xml would help make it more clear that the default way we do it should be through xml

app/lib/irs1040_pdf.rb
```diff
@@ -1,92 +1,87 @@
...
+      FilingStatus: @xml_document.at("IndividualReturnFilingStatusCd")&.text,
+      PrimaryFirstNm: @primary_first_name,
+      PrimaryLastNm: @primary_last_name,
+      PrimarySSN: @xml_document.at("PrimarySSN")&.text,
+      AddressLine1Txt: @street_address,
+      CityNm: @city,
+      StateAbbreviationCd: @state,
+      ZipCd: @zip_code,
+      VirtualCurAcquiredDurTYInd: @xml_document.at("VirtualCurAcquiredDurTYInd")&.text,
+      PrimaryBlindInd: bool_from_xml(@xml_document.at("PrimaryBlindInd")) ? "1" : "Off",
```
bytheway875, line 36: xml_check_to_bool() may make more sense as a method name here, since we're specifically looking for the presence of the X character. Also noodling on whether it would make sense to continue having a helper to create the "right" check vs Off value for the PDF vs doing a ternary in multiple places

app/lib/irs1040_pdf.rb
```diff
@@ -1,92 +1,87 @@
...
-    @qualifying_dependents.first(4).each_with_index do |dependent, index|
-      answers["DependentLegalNm[#{index}]"] = dependent.full_name
-      answers["DependentRelationship[#{index}]"] = dependent.irs_relationship_enum
-      answers["DependentSSN[#{index}]"] = dependent.ssn
-      answers["DependentCTCInd[#{index}]"] = dependent.qualifying_ctc ? 1 : 0
+    dependent_nodes.first(4).each_with_index do |dependent, index|
+      answers["DependentLegalNm[#{index}]"] = [dependent.at("DependentFirstNm").text, dependent.at("DependentLastNm").text].join(" ")
+      answers["DependentRelationship[#{index}]"] = dependent.at("DependentRelationshipCd").text
+      answers["DependentSSN[#{index}]"] = dependent.at("DependentSSN").text
+      answers["DependentCTCInd[#{index}]"] = bool_from_xml(dependent.at("EligibleForChildTaxCreditInd")) ? 1 : 0
```
bytheway875, line 79: its really weird that this is 1 or 0 and the others are 1 or off ... im now seeing why you dont have a helper for it. This seems like an inconsistency that could be fixed in the implementation of the PDF mapping if we thought it was necessary

spec/lib/irs1040_pdf_spec.rb
```diff
@@ -8,26 +8,6 @@
   let(:tax_year) { 2021 }
   let(:submission) { create :efile_submission, :ctc, tax_year: tax_year }
 
-  describe "initialization" do
```
bytheway875, line 11: i feel like this test ensured that there was a fallback to the intake address when there was not a persisted verified address object to refer to and I dont know that that behavior is otherwise tested in the spec.

wschaefer: yup we're going to add another way of testing this

PR #2470, opened 2022-05-05
Fix crash on stats reporting

bytheway875:


lib/tasks/stats.rake
```diff
@@ -23,8 +23,8 @@ namespace :stats do
   desc "Monitor the longest periods of time that any efile submission has been in preparing, bundling, and queued"
   task monitor_delayed_efile_submissions: :environment do
     [:preparing, :bundling, :queued].each do |state|
-      oldest_transition_to = EfileSubmissionTransition.where(most_recent: true, to_state: state).sort_by(&:created_at).first
-      min_since_transition = ((Time.now - oldest_transition_to.created_at)/60).to_i
+      oldest_transition_to = EfileSubmissionTransition.where(most_recent: true, to_state: state).order(&:created_at).limit(1)
+      min_since_transition = oldest_transition_to.present? ? ((Time.now - oldest_transition_to.created_at)/60).to_i : nil
```
jenny-heath, line 27: thanks for fixing this! do you think we should just avoid sending the event altogether if there's no transition?

jenny-heath (approved):

PR #2473, opened 2022-05-06
Add risky routing number list

bytheway875:
This needs to be a multistep process because due to the soft launch we already have 11 BankAccount objects on prod.

- we cannot just add an attribute called routing_number to the db without causing funky name collisions with the attr_encrypted gem that uses routing_number. Thus, I will dual write to `_routing_number` and leave all other things equal (that is included in this PR)
- Then I will run a backfill for the Bank Accounts that already exist in the db so that all elements have raw routing number written. (rake task included in this pr)
- Then, I will write another PR. In this one, I will remove the `attr_encrypted :routing_number` on bank account and then. there will be a method on bank account called `routing_number` that reads from `_routing_number` and will probably need to update some of the `_changed?` logic for the hashed routing number (must continue to read from hashed routing number because i need hashed_routing_number to exist for the purposes of the current fraud check, which i guess I could update to look at raw_routing_number temporarily but it removes a step to just rely on hashed_routing_number. It will also include adding an attribute called routing_number (that PR is here: https://github.com/codeforamerica/vita-min/pull/2474)

- Once that PR is out and the code is live, I will be able to update the bank_account fraud rule to look at `[:routing_number, :hashed_account_number]` instead of the hashed routing number.
- Once that is done on demo and production, I will do another PR that drops `:_routing_number` and `:hashed_routing_number` so that everything reads from routing_number and we have no vestiges of content that has raw and hashed or encrypted versions on the table.

### DIFF
--- a/app/models/bank_account.rb
+++ b/app/models/bank_account.rb
@@ -12,6 +12,7 @@
 #  encrypted_routing_number_iv :string
 #  hashed_account_number       :string
 #  hashed_routing_number       :string
+#  routing_number              :string
 #  created_at                  :datetime         not null
 #  updated_at                  :datetime         not null
 #  intake_id                   :bigint
@@ -34,7 +35,7 @@ class BankAccount < ApplicationRecord
   attr_encrypted :account_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
   # Enum values are acceptable BankAccountType values to be sent to the IRS (See efileTypes.xsd)
   enum account_type: { checking: 1, savings: 2 }
-  before_save :hash_data
+  before_save :hash_account_number
 
   # map string enum value back to the corresponding integer
   def account_type_code
@@ -42,10 +43,13 @@ class BankAccount < ApplicationRecord
   end
 
   def duplicates
-    DeduplificationService.duplicates(self, :hashed_routing_number, :hashed_account_number, from_scope: self.class)
+    DeduplificationService.duplicates(self, :routing_number, :hashed_account_number, from_scope: self.class)
   end
 
   def hash_data
+    # WIP because we need to get everything written onto the raw version before we can proceed.
+    # Then, we will remove the attr_encrypted and change the name of the column from raw_routing_number to routing_number
+    self.raw_routing_number = routing_number if routing_number_changed?
     [:routing_number, :account_number].each do |attr|
       if send("#{attr}_changed?") && send(attr).present?
         assign_attributes("hashed_#{attr}" => DeduplificationService.sensitive_attribute_hashed(self, attr))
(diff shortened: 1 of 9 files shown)


alaroia-cfa:
I think now's a good time to convert to storing the raw routing number!

(or if you really really don't want, hash these at save time, so that the hashing doesn't have to happen repeatedly)

lib/tasks/backfill_routing_numbers.rake
```diff
@@ -0,0 +1,8 @@
+namespace :routing_number do
+  desc "backfill routing numbers"
+  task backfill: :environment do
+    BankAccount.where(_routing_number: nil).find_each do |ba|
```
alaroia-cfa, line 4: This might be good to write in SQL for performance and to avoid unnecessarily updating `updated_at`. Not a blocker.

app/models/bank_account.rb
```diff
@@ -46,6 +47,9 @@ def duplicates
   end
 
   def hash_data
+    # WIP because we need to get everything written onto the _routing_number version before we can proceed.
+    # Then, we will remove the attr_encrypted and change the name of the column from _routing_number to routing_number
+    self._routing_number = routing_number if routing_number_changed?
```
alaroia-cfa, line 52: Not a blocker, but I would find it easier to read if this were its own method e.g. `before_save :dual_write_routing_number, :hash_data` since this isn't strictly related to hashing the data.

app/models/fraud/indicators/routing_number.rb
```diff
@@ -0,0 +1,31 @@
...
+# Table name: fraud_indicators_routing_numbers
+#
+#  id             :bigint           not null, primary key
+#  activated_at   :datetime
+#  bank_name      :string
+#  routing_number :string
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#
+module Fraud
```
alaroia-cfa, line 12: Not a blocker, but given that the dual write stuff is subtle, I would prefer to split that into its own PR, get it deployed, remove the ugly dual write stuff, _then_ add the new Fraud::Indicators.

alaroia-cfa (commented):
I left some comments. I think I need to eat lunch before I can understand this full plan. Could be interesting to Tuple review.

app/models/bank_account.rb
```diff
@@ -46,6 +47,9 @@ def duplicates
...
     [:routing_number, :account_number].each do |attr|
```
alaroia-cfa, line 53: `:routing_number` can now be deleted from here!

lib/tasks/backfill_routing_numbers.rake
alaroia-cfa: this cannot work because Rails is decrypting!

PR #2484, opened 2022-05-09
Rescue URI parsing errors

caitbot:
Other parties have begun security scanning us (presumably for nefarious purposes). And it creates a whole bunch of noise. We should rescue some of these raises rather than alerting on them. This exception was the biggest contributor to noisy alerts.

### DIFF
--- a/app/controllers/ajax_mixpanel_events_controller.rb
+++ b/app/controllers/ajax_mixpanel_events_controller.rb
@@ -24,12 +24,17 @@ class AjaxMixpanelEventsController < ApplicationController
 
   def controller_and_path_data
     controller, action = event_params[:controller_action].split("#", 2)
+    begin
+      path = URI(event_params[:full_path]).path
+    rescue URI::InvalidURIError
+      path = ""
+    end
     {
-      full_path: event_params[:full_path],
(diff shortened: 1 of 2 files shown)


app/controllers/ajax_mixpanel_events_controller.rb
```diff
@@ -24,12 +24,16 @@ def all_required_params_present?
...
   def controller_and_path_data
     controller, action = event_params[:controller_action].split("#", 2)
-    {
-      full_path: event_params[:full_path],
-      path: URI(event_params[:full_path]).path,
-      controller_action: event_params[:controller_action],
-      controller_action_name: action,
-      controller_name: controller.sub("Controller", ""),
-    }
+    begin
```
caitbot, line 27: going to precompute `path` and set to empty string for exception

caitbot, line 27: going to precompute `path` and set to empty string for exception

caitbot: and add test

caitbot: and add test

PR #2485, opened 2022-05-09
Update ddtracer, enabling log-trace correlation

caitbot:
Turns out we were using a pretty old version of `ddtracer` (0.42.0), and newer versions include log-trace correlation by default (if using `lograge`)

More details found [here](https://docs.datadoghq.com/tracing/setup_overview/setup/ruby/#for-logging-in-rails-applications)


bytheway875:
what does the tracing do?

caitbot (commented):
Reviewed as a pair with @alaroia-cfa

PR #2489, opened 2022-05-09
Hub efile submissions index: add column for error codes and messages

jenny-heath:
2 things:
1) the code in the view itself isn't super tested, i wonder if there are some real world cases that would cause an undefined method for nil class type of error. i did manage to locally create a submission with no errors attached and that didn't cause any problems but idk if there are other realistic cases i should have tried out.
2) i did a nowrap on the error table cell because the messages are typically pretty long. i figured this was better than wrapping it and having fat rows? i'll attach a screenshot without the nowrap in a comment.

image

### DIFF
--- a/app/views/hub/efile_submissions/index.html.erb
+++ b/app/views/hub/efile_submissions/index.html.erb
@@ -44,6 +44,9 @@
         <th scope="col" class="index-table__header sortable">
           Last updated
         </th>
+        <th scope="col" class="index-table__header">
+          Errors
+        </th>
       </tr>
       </thead>
       <tbody class="index-table__body submission-table">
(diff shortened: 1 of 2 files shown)


jenny-heath:
without `white-space: nowrap;` on the error
image

bytheway875:
I know it would be more complex, but i wonder if we could make it so that we show like a truncated version and then have the hover thing with the black box like we do for messages on the client index? 🤔

jenny-heath:
@bytheway875 changed it to this! i thought it would be quick but i got confused by the tests haha

image

jenny-heath:
small update: i showed yvonne the most recent design for this and she says it's good

bytheway875 (approved):

PR #2490, opened 2022-05-09
Ignore `TaxReturn#status` column, which we will soon delete

wschaefer:
The information previously stored as a integer in this column
is now stored as a string in `TaxReturn#current_state`


bytheway875:
@wschaefer @tgrathwell i think client_sortable is still looking at tax_return.status also

PR #2492, opened 2022-05-09
Drop encrypted columns real

bytheway875:


db/schema.rb
```diff
@@ -425,22 +425,17 @@
...
     t.string "encrypted_bank_name_iv"
-    t.string "encrypted_routing_number"
-    t.string "encrypted_routing_number_iv"
     t.string "hashed_account_number"
-    t.string "hashed_routing_number"
     t.bigint "intake_id"
     t.string "routing_number"
     t.datetime "updated_at", null: false
     t.index ["hashed_account_number"], name: "index_bank_accounts_on_hashed_account_number"
-    t.index ["hashed_routing_number"], name: "index_bank_accounts_on_hashed_routing_number"
```
alaroia-cfa, line 443: We're dropping an index here. It looks like we didn't add a similar index for `routing_number`. Might be good to add the `routing_number` index.

app/models/bank_account.rb
```diff
@@ -21,15 +17,13 @@
...
 #  index_bank_accounts_on_hashed_account_number  (hashed_account_number)
-#  index_bank_accounts_on_hashed_routing_number  (hashed_routing_number)
 #  index_bank_accounts_on_intake_id              (intake_id)
 #
 # Foreign Keys
 #
 #  fk_rails_...  (intake_id => intakes.id)
 #
 class BankAccount < ApplicationRecord
-  self.ignored_columns = [:encrypted_routing_number, :encrypted_routing_number_iv, :hashed_routing_number, :_routing_number]
```
alaroia-cfa, line 32: I'm 49% tempted to say don't delete `ignored_columns`.

I vote 51% in favor of "it's OK to remove it". The database will be in a state where the columns are gone when the code is executed, so I guess it's OK to remove it!

alaroia-cfa (approved):
This seems perfect (except for the index comment); lgtm either way.

I'm nervous because we've been so bad at this before... but I think this time we're doing it right.

db/schema.rb
bytheway875: good call! added

bytheway875: good call! i added the index

PR #2493, opened 2022-05-09
Scope stats to this year #182029118

jenny-heath:
First commit: makes sure we don't catch submissions from last year in our DD latency stats

Second commit: i was trying to transition a submission from last year that is in `preparing` to `cancelled` and couldn't because the clearest path is through `fraud_hold`, but `after_transition(to: :fraud_hold)` throws a nil class error when trying to get the locale from the nonexistent intake. (i tried some other paths to `cancelled` but they proved to be complicated.

### DIFF
--- a/lib/tasks/stats.rake
+++ b/lib/tasks/stats.rake
@@ -23,7 +23,11 @@ namespace :stats do
   desc "Monitor the longest periods of time that any efile submission has been in preparing, bundling, and queued"
   task monitor_delayed_efile_submissions: :environment do
     [:preparing, :bundling, :queued].each do |state|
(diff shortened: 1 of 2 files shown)


bytheway875:
I think for these "stale" old commits, we dont actually need to transition them through other states ... we can just update the to_state to cancelled (which might get rid of some of the history, but thats kind of okay because we know these are all people who were not able to file). The change you made means that we're going to send messages to people about the status change, and we definitely don't want to do that!

jenny-heath:
@bytheway875 ohhh i didn't realize you could do it the other way! i'll do that instead

jenny-heath:
@bytheway875 since the one problem submission is now fixed in prod this PR isn't as necessary. do you think we should add this filter anyway or just scrap it?

PR #2495, opened 2022-05-10
Remove duplicated submission status note for verification attempt

bytheway875:
When we first implemented verification attempts, there was not a persisted transition for the creation of a submission. Now there is, so we have duplicated entries! We can remove one of them.
Screen Shot 2022-05-09 at 5 55 01 PM


alaroia-cfa:
@bytheway875 I think your screenshot didn't finish submitting. Can you re-upload?

bytheway875:
@alaroia-cfa reuploaded

PR #2496, opened 2022-05-10
Move recaptcha commit before the transition to preparing to ensure its always committed

bytheway875:


app/models/efile_security_information.rb
```diff
@@ -28,6 +28,7 @@
 #  fk_rails_...  (efile_submission_id => efile_submissions.id)
 #
 class EfileSecurityInformation < ApplicationRecord
+  self.ignored_columns = [:recaptcha_score]
```
bytheway875, line 31: i guess we recently learned this doesnt do anything...

spec/controllers/ctc/questions/confirm_legal_controller_spec.rb
```diff
@@ -53,7 +53,6 @@
           efile_submission = client.reload.tax_returns.last.efile_submissions.last
           expect(efile_submission.current_state).to eq "bundling"
           expect(client.efile_security_informations.last.ip_address).to eq ip_address
-          expect(client.efile_security_informations.last.recaptcha_score).to eq 0.9
```
caitbot, line 56: should we test that it is adding the `ntake.client.recaptcha_scores` here instead?

caitbot (approved):
cool! this looks good to me

spec/controllers/ctc/questions/confirm_legal_controller_spec.rb
```diff
@@ -53,7 +53,6 @@
...
           recaptcha_score = client.recaptcha_scores.last
           expect(recaptcha_score.score).to eq 0.9
```
bytheway875, line 57: thats happening here, right? @caitbot

caitbot: ahh yes, it definitely is

PR #2497, opened 2022-05-10
Autofocus on adding item to list in fraud indicators

bytheway875:
This allows us to add domains/entries much faster than having to click into the box again, since we're often adding many entries in succession.


bytheway875:
btw, its not focused when you first load the page but just after you add one or there is an error with the one you tried to add

caitbot (approved):
Nice! This is really cool

jenny-heath (approved):

PR #2506, opened 2022-05-10
Handle USPS blank responses and add address validation

caitbot:

### DIFF
--- a/app/services/standardize_address_service.rb
+++ b/app/services/standardize_address_service.rb
@@ -11,6 +11,7 @@ class StandardizeAddressService
     @_zip_code = intake.zip_code
 
     @result = build_standardized_address
+    log_errors
   end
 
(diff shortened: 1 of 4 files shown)


app/services/standardize_address_service.rb
```diff
@@ -49,6 +50,16 @@ def valid?
 
   private
 
+  def log_errors
+    if !valid?
+      Rails.logger.error "Error returned from USPS Address API: #{@result[:error_message]}"
+    elsif @result[:zip_code].blank?
+      Rails.logger.error "Error from USPS Address API: Response had no data."
+      @result[:error_code] = "-MISSING-DATA"
+      @result[:error_message] = "USPS API response had no data"
```
bytheway875, line 59: ohhh does this get prefixes with USPS when it gets consumed in the job?

PR #2507, opened 2022-05-10
Fix not ready reminder reference

bytheway875:


spec/tasks/not_ready_reminders_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+Rails.application.load_tasks
+
+describe "not_ready:remind" do
+  after(:each) do
```
caitbot, line 6: huh til about this `after(:each)` and `reenable` pattern

source: https://tasdikrahman.me/2020/10/20/testing-rake-tasks-with-rspec/

caitbot (approved):
cool, looks good, thanks for adding a spec, much improved

PR #2518, opened 2022-05-11
Update fraud_suspected? so that it returns true when the score equals the hold threshold, and add tests

bytheway875:


spec/models/client_spec.rb
```diff
@@ -91,6 +91,45 @@
...
+      stub_const("Fraud::Score::HOLD_THRESHOLD", 50000)
+    end
+
+    context "with at least one fraud score over the threshold" do
+      before do
+        client.fraud_scores.create(score: 130000, efile_submission: efile_submission)
+      end
+
+      it "returns true" do
+        puts Fraud::Score::HOLD_THRESHOLD
```
jenny-heath, line 107: remove this?

jenny-heath (approved):

PR #2544, opened 2022-05-11
Add additional validations for drivers license year - Drivers license years must be 4 digits and start with either 19 or 20 for issue date or 20 for expiration date

bytheway875:
We're checking the string version of the provided date in some of these validations which, to my knowledge, should be fine since they're always coming in from http request form params.

### DIFF
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -25,22 +25,37 @@ module DateHelper
   def valid_text_birth_date(birth_date_year, birth_date_month, birth_date_day, key = :birth_date)
     parsed_birth_date = parse_date_params(birth_date_year, birth_date_month, birth_date_day)
     unless parsed_birth_date.present?
-      self.errors.add(key, I18n.t('errors.attributes.birth_date.blank'))
+      errors.add(key, I18n.t('errors.attributes.birth_date.blank'))
       return false
     end
 
     if parsed_birth_date.year < 1900 || parsed_birth_date.year > Date.today.year
-      self.errors.add(key, I18n.t('errors.attributes.birth_date.blank'))
+      errors.add(key, I18n.t('errors.attributes.birth_date.blank'))
       return false
     end
 
     true
   end
 
+  def valid_expiration_date(date_year, date_month, date_day, key = :expiration_date)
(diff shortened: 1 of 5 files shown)


app/helpers/date_helper.rb
```diff
@@ -25,22 +25,35 @@ def valid_birth_date
...
+  def valid_expiration_date(date_year, date_month, date_day, key = :expiration_date)
+    if date_year.present? && date_year.starts_with?("19")
+      errors.add(key, I18n.t('errors.attributes.expiration_date.format'))
+      return false
+    end
+    valid_text_date(date_year, date_month, date_day, key)
+  end
+
   def valid_text_date(date_year, date_month, date_day, key = :date)
+    if date_year.present? && (!date_year.length == 4 && !["19", "20"].includes?(date_year[0..1]))
```
bytheway875, line 49: this was tricky, but yvonne wanted the issue year to only allow 19 or 20 as the start dates, but also require that the expiration date is in the 2000s

jenny-heath: should this be an or statement? as in:
`!date_year.length == 4 || !["19", "20"].includes?(date_year[0..1])`
because being either not four digits or not in the last century would make it invalid?

jenny-heath: wait, i got it wrong. this is hard to think about haha

spec/forms/ctc/drivers_license_form_spec.rb
```diff
@@ -42,6 +42,68 @@
...
+            expiration_date_month: "12",
+            expiration_date_day: "13",
+            state: "CA",
+            license_number: "YT12345",
+        }
+      }
+
+      it "includes the correct error attributes" do
+        form.valid?
+        puts form.errors.inspect
```
jenny-heath, line 62: leftover print statement?

spec/forms/ctc/drivers_license_form_spec.rb
```diff
@@ -42,6 +42,68 @@
...
```
jenny-heath, line 83: same

jenny-heath (approved):
logic feels much easier to read now!

PR #2546, opened 2022-05-11
Ctc intake flow: add puerto rico as home option and direct to sign up page for updates

jenny-heath:
image

image


app/forms/ctc/home_form.rb
```diff
@@ -7,5 +7,9 @@ def save; end
     def lived_in_territory_or_at_foreign_address?
       ["lived_in_us_territory", "lived_at_foreign_address"].include? home_location
     end
+
+    def lived_in_puerto_rico?
+      home_location == "lived_in_puerto_rico"
+    end
```
jenny-heath, line 13: was surprised to find out that we don't save the answer to this question

config/locales/en.yml
```diff
@@ -1785,6 +1785,13 @@ en:
           - To check the status of your %{current_tax_year} tax return you can log into your <a href='https://www.irs.gov/payments/your-online-account'>IRS online account</a>.
           - If you need additional support, you can reach out to a Volunteer Income Tax Assistance (VITA) site or the Taxpayer Advocate Service.
           title: Unfortunately, you can’t use GetCTC because you already filed a %{current_tax_year} tax return.
+        puerto_rico_sign_up:
+          help_text_html: |
+            <p>GetCTC is currently live for residents of the 50 states and D.C. and will be available for residents of Puerto Rico later this year.</p>
```
bytheway875, line 1790: Its weird that one paragraph is in a p tag and the other one isn't. The structure of this will be:

```
<p>
  <p>GetCTC is currently live ...</p>
  Get notified...
</p>
```

If we want both to be in their own p tag, we should either remove the p tag from the layout and add a p tag to the "Get notified" test in the yml OR, preferably, create a help_text_items array with text-only yml and loop through it from the template:

```
<% t("help_text_items").each do |item| %>
  <p><%= item %></p>
<% end %>
```

config/routes.rb
```diff
@@ -320,6 +320,7 @@ def scoped_navigation_routes(context, navigation)
       get "/questions/use-gyr", to: "ctc/questions/use_gyr#edit", as: :questions_use_gyr
       get "/questions/returning-client", to: "ctc/questions/returning_client#edit", as: :questions_returning_client
       get "/already-filed", to: "ctc/offboarding/already_filed#show", as: :offboarding_already_filed
+      get "/cant-use-getctc", to: "ctc/offboarding/puerto_rico_sign_up#show", as: :offboarding_cant_use_getctc
```
bytheway875, line 323: its weird to have this url be so generic when it is actually very specific to puerto rico. Is it possible that we can do `/cant-use-getctc-pr` instead?

bytheway875 (changes requested):
Please update the pattern for paragraph tags in the template, yaml, or both

config/routes.rb
jenny-heath: agree that it's weird! i was just using the one specified in the story, but can ask yvonne if it's ok to change it

config/locales/en.yml
jenny-heath: oops good catch! i originally didn't notice the two paragraphs so this was a bit of an oversight

bytheway875 (approved):
Perfection!

PR #2547, opened 2022-05-12
Fix prev tax year (which was always displaying as 2020)

bytheway875:


app/views/hub/clients/show.html.erb
```diff
@@ -328,15 +328,15 @@
               <% end %>
             <% end %>
           </div>
-          <% if @client.intake.is_ctc? && !@client.intake.drop_off? %>
```
caitbot, line 331: we shouldn't have any more CTC drop off clients this season, right?

if we did, I feel like we would actually want to keep this, because if we had drop off clients we want to efile for, it would be useful to be able to change their AGI.

but since we don't (IIRC) then I don't feel strongly about this, but also don't feel that its a necessary inclusion.

caitbot (approved):
This looks good, I'm in favor of removing the `!@client.intake.drop_off?` check but don't feel that it is blocking.

app/views/hub/clients/show.html.erb
bytheway875: we deleted everything having to do with creating drop off clients for CTC before the season

app/views/hub/clients/show.html.erb
```diff
@@ -328,15 +328,15 @@
...
+          <% if @client.intake.is_ctc? %>
             <div class="client-profile__field-group">
               <h2 class="text--bold">Prior Year AGI Amount</h2>
               <div class="field-display">
-                <span class="form-question">Primary Prior Year (2019) AGI:</span>
+                <span class="form-question">Primary Prior Year (<%= @client.tax_returns.last.year - 1 %>) AGI:</span>
```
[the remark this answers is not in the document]
bytheway875: yeah, i actually ran into this because all of the tests were for archived intakes!

PR #2556, opened 2022-05-12
Add the ability to reprocess errors to the admin error index

caitbot:
image

### DIFF
--- a/app/controllers/hub/efile_errors_controller.rb
+++ b/app/controllers/hub/efile_errors_controller.rb
@@ -24,13 +24,11 @@ module Hub
 
     def reprocess
       if @efile_error.present? && (@efile_error.auto_wait || @efile_error.auto_cancel)
-        submission_ids = EfileSubmissionTransitionError.includes(:efile_error, :efile_submission_transition).where(efile_error: @efile_error, efile_submission_transitions: { most_recent: true, to_state: ["rejected", "failed"] }).pluck(:efile_submission_id)
-        submissions = EfileSubmission.where(id: submission_ids)
-
         auto_transition_to_state = @efile_error.auto_wait ? :waiting : :cancelled
-        submissions.find_each { |submission| submission.transition_to(auto_transition_to_state) }
+        submission_ids = EfileSubmissionTransitionError.includes(:efile_error, :efile_submission_transition).where(efile_error: @efile_error, efile_submission_transitions: { most_recent: true, to_state: ["rejected", "failed"] }).pluck(:efile_submission_id)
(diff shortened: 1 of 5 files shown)


app/controllers/hub/efile_errors_controller.rb
```diff
@@ -22,6 +22,21 @@ def update
...
+      if @efile_error.present? && (@efile_error.auto_wait || @efile_error.auto_cancel)
+        submission_ids = EfileSubmissionTransitionError.includes(:efile_error, :efile_submission_transition).where(efile_error: @efile_error, efile_submission_transitions: { most_recent: true, to_state: ["rejected", "failed"] }).pluck(:efile_submission_id)
+        submissions = EfileSubmission.where(id: submission_ids)
+
+        auto_transition_to_state = @efile_error.auto_wait ? :waiting : :cancelled
+        submissions.find_each { |submission| submission.transition_to(auto_transition_to_state) }
+
+        flash[:notice] = "Successfully reprocessed #{submissions.count} submission(s) with #{@efile_error.code} error!"
+      else
+        flash[:notice] = "Could not reprocess #{@efile_error.code}. Try again."
```
caitbot, line 33: Argh! I tried to make this an `error` notice but for some odd reason it was not showing up correctly on the page. It would create space, but it would not be red, nor would you see any text. 

Given that this is an admin page and Yvonne wants it ASAP, I decided not to waste too much time figuring out why.

app/views/hub/efile_errors/index.html.erb
```diff
@@ -21,16 +23,16 @@
         <th scope="col" width="500" class="index-table__header">
           Error Message
         </th>
-        <th scope="col" width="50" class="index-table__header index-table__cell--center">
+        <th scope="col" width="75" class="index-table__header index-table__cell--center">
```
caitbot, line 26: Otherwise they are smushed!

bytheway875:
I slightly wonder whether the reprocess errors button should be on the show page for the error instead of on the index in order to reduce the risk of pressing the wrong button. In the case that we leave the button on the index, I recommend swapping the placement of source to the end of the row so that the error code displays next to the button to diminish the risk of fat fingering or eyes not lining up to the correct line.

app/views/hub/efile_errors/index.html.erb
```diff
@@ -21,16 +23,16 @@
         <th scope="col" width="500" class="index-table__header">
```
bytheway875, line 23: another thing i wanted to try on this table was to see if it was as easy as adding a class to make the table header sticky... the check boxes lose context once you scroll past the top of the page and now weve got lots of error messages.

caitbot: ooo thats a nice idea

PR #2557, opened 2022-05-13
Copy change for Married filing separately

bytheway875:


caitbot (approved):
Huh why was the rake preventing a merge?

This looks a-ok to me.

bytheway875:
whitney and i did a follow up commit yesterday that ignores the sent_followup column because were planning on dropping it and instead adding a more specific column name

PR #2558, opened 2022-05-13
Really send open email 181037257

bytheway875:
This creates two new rake tasks -- one for sending a specific message to signups and another for deleting all people who we've messaged with a specific message.

Instead of a generic sent_followup boolean, we added a pattern where you need to add a field that matches the name of the email that you want to send. This will allow us to additional send messages to subsets of these signups later if we need to, I guess? (The whole Peurto Rico signups will need to get another message later kind of messed with us.

Now you can create an AutomatedMessage object and specify that and it should work.


lib/tasks/signup.rake
```diff
@@ -0,0 +1,24 @@
+namespace :signup do
+  desc 'Deal with messaging signups'
```
caitbot, line 2: lol at the language "deal with"

caitbot (approved):
reviewed as pair-ish

PR #2560, opened 2022-05-16
Prevent double-submit on GYR consent form [#182199495]

alaroia-cfa:

## Before

If you click "I agree", the form button continues to appear, and if you click it repeatedly _while waiting for the new page to load_ you'd actually ask the app to submit _twice_, and then you'd see a crash screen because our app is bad at doing this action twice in parallel. (Clients could presumably resolve the issue for themselves by hitting back and resubmitting a _third_ time, but yeesh.)

## After

If you click "I agree", the form button changes label to "Please wait...", and if you click it repeatedly _while waiting for the new page to load_ the click is ignored, and you have to just keep waiting.

The video doesn't show the fact that we do actually navigate you to the next page. That's b/c of my error when editing the video! Please picture that.

Screen Recording 2022-05-16 at 1 09 49 PM

## Other

Should we do this for all buttons? Seems like yes. Thoughts? @bytheway875 your thoughts especially welcome here.

Anu, are you OK with this behavior? It's a situation where eng has a need to change app behavior, which is backwards in theory, but in this case I think is a net win for clients.

GetCTC already has this behavior, at least in the client portal.

### DIFF
--- a/app/views/ctc/portal/portal/edit_info.html.erb
+++ b/app/views/ctc/portal/portal/edit_info.html.erb
@@ -62,14 +62,7 @@
             <%= t("views.ctc.portal.edit_info.help_text") %>
           </p>
           <%= recaptcha_v3(action: 'resubmit') %>
-          <%= button_tag(
-                class: "button button--primary button--full-width",
-                type: "submit",
-                data: { disable_with: t("general.please_wait") },
-                disabled: !@submit_enabled
-              ) do %>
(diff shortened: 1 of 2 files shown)


bytheway875 (approved):

caitbot (approved):
I love it, do it everywhere.

I double love the PR description, so good.

wschaefer:
Closing so we can make a new one where Heroku review apps will work hopefully

PR #2562, opened 2022-05-16
Update experience survey send out

caitbot:
Per a conversation with Yvonne, `rejected` isn't really considered a terminal status for CTC tax returns, and so we want to update to send out the experience survey once a CTC tax return enters the `not_filing` status (which appears as cancelled in the efile dashboard)

Along with this change, we want to maintain sending out experience surveys to `rejected` GYR clients, as it IS used as a terminal status for GYR.

Finally, it was decided that we want to begin sending experience surveys to `not_filing` tax returns clients for GYR as well.

### DIFF
--- a/app/state_machines/tax_return_state_machine.rb
+++ b/app/state_machines/tax_return_state_machine.rb
@@ -65,8 +65,13 @@ class TaxReturnStateMachine
   end
 
   after_transition(to: "file_rejected") do |tax_return, _|
-    tax_return.enqueue_experience_survey
+    tax_return.enqueue_experience_survey unless tax_return.is_ctc
     MixpanelService.send_file_completed_event(tax_return, "filing_rejected")
+  end
+
+  after_transition(to: "file_not_filing") do |tax_return, _|
+    tax_return.enqueue_experience_survey
+    MixpanelService.send_file_completed_event(tax_return, "not_filing")
   end
 
   after_transition(to: "file_mailed") do |tax_return, _|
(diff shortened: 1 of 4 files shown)


app/state_machines/tax_return_state_machine.rb
```diff
@@ -65,10 +65,15 @@ class TaxReturnStateMachine
   end
 
   after_transition(to: "file_rejected") do |tax_return, _|
-    tax_return.enqueue_experience_survey
+    tax_return.enqueue_experience_survey unless tax_return.is_ctc
     MixpanelService.send_file_completed_event(tax_return, "filing_rejected")
```
caitbot, line 69: I imagine Max is using this statistic, so I am maintaining it, even though in CTC land this isn't really considered "completed"

spec/state_machines/tax_return_state_machine_spec.rb
```diff
@@ -76,9 +76,10 @@
 
   context "transitions" do
     let(:tax_return) { create(:tax_return) }
+
     context "to file_accepted" do
       before do
-        allow_any_instance_of(TaxReturn).to receive(:enqueue_experience_survey)
+        allow(tax_return).to receive(:enqueue_experience_survey)
```
caitbot, line 82: oo yay we can get rid of `allow_any_instance` pretty easily here

bytheway875:
straightforward! love it

bytheway875 (approved):

PR #2563, opened 2022-05-16
Fix bug where we send en to SendAutomatedMessage

bytheway875:
This bug was kind of caused by having this "go between method" in ClientMessagingService. I wonder if I should spend the time to just switch everything over to SendAutomatedMessage directly...

### DIFF
--- a/app/services/client_messaging_service.rb
+++ b/app/services/client_messaging_service.rb
@@ -84,7 +84,7 @@ class ClientMessagingService
       message_records
     end
 
-    def send_system_message_to_all_opted_in_contact_methods(client:, message:, tax_return: nil, locale: "en", body_args: {})
+    def send_system_message_to_all_opted_in_contact_methods(client:, message:, tax_return: nil, locale: nil, body_args: {})
       SendAutomatedMessage.new(
         client: client,
         message: message,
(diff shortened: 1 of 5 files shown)


app/views/hub/verification_attempts/show.html.erb
```diff
@@ -57,7 +57,7 @@
                     <% end %>
                   </div>
                 </div>
-                <div class="grid-flex column-column justify-flex-end">
+                <div class="grid-flex column-column">
```
alaroia-cfa, line 60: Seems unrelated, but I don't object.

app/services/client_messaging_service.rb
```diff
@@ -84,7 +84,7 @@ def send_message_to_all_opted_in_contact_methods(client:, user:, body:, locale:
       message_records
     end
 
-    def send_system_message_to_all_opted_in_contact_methods(client:, message:, tax_return: nil, locale: "en", body_args: {})
+    def send_system_message_to_all_opted_in_contact_methods(client:, message:, tax_return: nil, locale: nil, body_args: {})
```
alaroia-cfa, line 87: Thrilled to see this removal of `en` default!

alaroia-cfa (approved):
Nice! There's the CSS thing that may be unrelated; I approve anyway.

alaroia-cfa:
> I wonder if I should spend the time to just switch everything over to SendAutomatedMessage directly.

If you find yourself interested in a refactor, and it's less than 3 hours of work, it's probably worth it.

PR #2566, opened 2022-05-17
Fix pagination error and add user name search

bytheway875:
Check params, and also make it less likely people need to mess with this param at all by adding search to the page.
Screen Shot 2022-05-16 at 5 08 03 PM
Screen Shot 2022-05-16 at 5 08 11 PM


alaroia-cfa (approved):
Nice tests & simple code!

PR #2568, opened 2022-05-17
Fix page param on user page

bytheway875:


alaroia-cfa (approved):
Very simple.

It's sorta too bad `will_paginate` makes us add all this defensive code everywhere.

Have I also mentioned my general fear that `will_paginate` uses slow `LIMIT OFFSET` queries? https://www.eversql.com/faster-pagination-in-mysql-why-order-by-with-limit-and-offset-is-slow/

So maybe one day we can shop for an alternative to `will_paginate` that uses 'query set pagination' (much speedier) and it'll have this param cleaning built-in, or something.

PR #2569, opened 2022-05-17
[#182141745] Allow incoming Twilio messages with blank message & no attachments

alaroia-cfa:
```
> FactoryBot.create(:incoming_text_message, body: '', client: Client.last)
```

After this change, this is what a blank message looks like.

Screen Shot 2022-05-27 at 12 06 57 PM

### DIFF
--- a/spec/features/hub/send_messages_spec.rb
+++ b/spec/features/hub/send_messages_spec.rb
@@ -15,9 +15,9 @@ RSpec.feature "Read and send messages to a client", js: true do
       )
     end
     let(:client) { create(:client, vita_partner: user.role.organization, intake: intake) }
-
     before do
       login_as user
+      create(:incoming_text_message, body: "", client: client)
     end
 
     scenario "I can view a client's information and send them a message" do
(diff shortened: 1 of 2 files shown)


bytheway875:
My sense is that there should be some visual distinction here, because its confusing that the text looks exactly what the text would look like if a client actually sent it.

alaroia-cfa:
@bytheway875 Interesting, yeah. I wondered about that, too.

I called Yvonne to bring that up, and we added Anu, and Anu said she'd think about what's best. We discussed 3 options: (1) Leave this as-is (2) Change the copy to something that's clearer (3) Change how it appears overall to indicate that it's a weird situation. There was some discussion about how if it's rare, maybe it makes sense to optimize for something low-lift.

I'll mark this PR as a draft, awaiting input.

FYI Anu:

image

caitbot (commented):
From a code standpoint, this look good to me. I agree with SB's take that it would be nice to have some visual distinction if there is a low-lift option. My eyes ignored the message since it had a body and then I came back around and realized its intent.

bytheway875:
Screen Shot 2022-05-27 at 12 06 57 PM

@alaroia-cfa

bytheway875:
@alaroia-cfa Do you already have a client on demo set up to test the empty message where you remember the id? or should i create a new message on someones profile?

alaroia-cfa (commented):
lgtm (needs a test removed)

alaroia-cfa:
> Do you already have a client on demo set up to test the empty message where you remember the id? or should i create a new message on someones profile?

I don't have one lying around, so you'll have to create one. Furthermore... before this pull request, demo can't _have_ any empty messages, so I think none exist. :)

PR #2571, opened 2022-05-17
Do not create duplicate error when the error message changes

caitbot:
See the slack for more details: https://cfa.slack.com/archives/GT1B45BS5/p1652718087681559

When clearing out the reject queue, we discovered that an error, IND-664-02, was duplicated.

This duplication is due to the error message being updated by the IRS. We do a `find_or_create` on `EfileError` and the error message is one of the attributes that we match on. When it didn't match we created a new error.

Going forward we should not create a new error, but instead update the error message of the existing error when the IRS updates the error message copy.

### DIFF
--- a/app/lib/efile/submission_error_parser.rb
+++ b/app/lib/efile/submission_error_parser.rb
@@ -30,9 +30,10 @@ module Efile
       metadata = @transition.metadata
 
       attrs = { code: metadata["error_code"] }
-      attrs[:message] = metadata["error_message"] if metadata["error_message"].present?
       attrs[:source] = metadata["error_source"] if metadata["error_source"].present?
       efile_error = EfileError.find_or_create_by!(attrs)
+
+      efile_error.update(message: metadata["error_message"]) if metadata["error_message"].present? && efile_error.message != metadata["error_message"]
 
       @transition.efile_submission_transition_errors.create(efile_submission_id: efile_submission.id, efile_error: efile_error)
     end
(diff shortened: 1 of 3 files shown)


app/lib/efile/submission_error_parser.rb
```diff
@@ -30,10 +30,11 @@ def persist_errors_from_error_metadata
       metadata = @transition.metadata
 
       attrs = { code: metadata["error_code"] }
-      attrs[:message] = metadata["error_message"] if metadata["error_message"].present?
       attrs[:source] = metadata["error_source"] if metadata["error_source"].present?
```
wschaefer, line 33: I remember talking about the error code as what we should base it on, but I don't remember us talking about source - do you know what that is? Is it possible we'll have the same dup problems with source that we had with message? I know that's not what this story was about but since we're in here I was curious :)

app/lib/efile/submission_error_parser.rb
```diff
@@ -45,11 +46,13 @@ def persist_errors_from_raw_response_metadata
       as_xml.search("ValidationErrorGrp").each do |error_group|
         error = EfileError.find_or_create_by!(
           code: error_group.at("RuleNum")&.text,
-          message: error_group.at("ErrorMessageTxt")&.text,
           category: error_group.at("ErrorCategoryCd")&.text,
           severity: error_group.at("SeverityCd")&.text,
           source: "irs"
         )
+        message = error_group.at("ErrorMessageTxt")&.text
```
wschaefer, line 53: So just for my own learning: these are errors coming back from the irs and the other place you changed are errors coming back from other integrations (like the usps or whatever we do for addresses)? So theoretically any of them could have the code stay the same but the message text change?

wschaefer (approved):

app/lib/efile/submission_error_parser.rb
caitbot: Ooo another good question. In short, yes. And to be totally honest, checking for message changes in the metadata is way less important than in this method. Most of the metadata errors are due to addresses or bundling issues (and so are on our side of things), like you said. 

I did both since we never needed to be matching on messages to begin with and it felt like it left less cracks. But we shouldn't really expect the metadata messages to change.

PR #2573, opened 2022-05-17
Update links to GYR/DIY on file-full-return and stimulus-payments

caitbot:


config/locales/en.yml
```diff
@@ -2151,7 +2151,7 @@ en:
             body_html:
             - The IRS issued three rounds of stimulus payments during the COVID pandemic. Most families received these in April 2020, December 2020, and March 2021.
             - You can only claim all or part of the third stimulus payment using GetCTC.
-            - If you missed any of the first or second stimulus payments, you have to file a 2020 tax return to claim them. You can file for free at GetYourRefund.
+            - If you missed any of the first or second stimulus payments, you have to file a 2020 tax return to claim them. You can file for <a target="_blank" rel="noopener nofollow" href="https://www.getyourrefund.org/en/diy?source=gctc_ref">free at GetYourRefund.</a>.
```
wschaefer, line 2154: I wasn't sure if there was a reason this url is hardcoded to getyourrefund.org and the other one is generated with a `url_for`?

caitbot: Since there is a source param included in this URL, I just put it inline rather than generate it with `url_for`. There might be a clever way to include a source param in `url_for` but I wasn't entirely sure.

PR #2575, opened 2022-05-17
Additional Dependents PDF

bytheway875:
Screen Shot 2022-05-17 at 2 03 40 PM


app/lib/additional_dependents_pdf.rb
```diff
@@ -0,0 +1,37 @@
...
+  private
+
+  def dependents_info(dependent_nodes)
+    answers = {}
+    dependent_nodes.each_with_index do |dependent, index|
+      answers["DependentNameRow#{index + 1}"] = [dependent.at("DependentFirstNm").text, dependent.at("DependentLastNm").text].join(" ")
+      answers["TINRow#{index + 1}"] = dependent.at("DependentSSN").text
+      answers["RelationshipRow#{index + 1}"] = dependent.at("DependentRelationshipCd").text
+      answers["CTCRow#{index + 1}"] = xml_check_to_bool(dependent.at("EligibleForChildTaxCreditInd")) ? "Yes" : nil
+      answers["OTCRow#{index + 1}"] = xml_check_to_bool(dependent.at("EligibleForODCInd")) ? "Yes" : nil
```
alaroia-cfa, line 29: This calls it ODC (which makes sense to me, other dependents credit aka credit for other dependents).

The PDF calls it "OTC" but I think it should be "ODC".

image

app/lib/additional_dependents_pdf.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+    answers
+  end
+
+  def xml_check_to_bool(node)
```
alaroia-cfa, line 34: Consider moving to `PdfHelper`; just an idea, not a very strong suggestion.

app/lib/submission_builder/ty2021/return1040.rb
```diff
@@ -20,6 +20,11 @@ def supported_documents
             pdf: Irs1040Pdf,
             include: true
           },
+          {
+            xml: nil,
+            pdf: AdditionalDependentsPdf,
+            include: @submission.qualifying_dependents.count > 4
```
alaroia-cfa, line 26: This `include` thing is neat.

spec/lib/additional_dependents_pdf_spec.rb
```diff
@@ -0,0 +1,150 @@
+require "rails_helper"
+
+describe AdditionalDependentsPdf do
+  include PdfSpecHelper
+  let(:pdf) { described_class.new(submission) }
+  let(:submission) { create :efile_submission }
+  let(:fake_xml_document) do
+    Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+      xml.IRS1040ScheduleLEP do
```
alaroia-cfa, line 9: This is a nice way to build XML in tests that still gives you syntax highlighting.

spec/lib/additional_dependents_pdf_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        22.times do |i|
+          empty.merge!({
+            "DependentNameRow#{i + 1}" => nil,
+            "CTCRow#{i + 1}" => nil,
+            "OTCRow#{i + 1}" => nil,
+            "RelationshipRow#{i + 1}" => nil,
+            "TINRow#{i + 1}" => nil,
+          })
+        end
+        expect(non_preparer_fields(output_file.path)).to eq empty
```
alaroia-cfa, line 103: It's probably better to use `filled_in_values(filepath).keys` rather than `non_preparer_fields`. I think the word `preparer` in `non_preparer_fields` only has meaning in the 13614-C. Not a blocker.

spec/lib/additional_dependents_pdf_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+          xml.RootNode do
+          end
+        end.doc
+      end
+
+      it "returns default values" do
+        output_file = pdf.output_file
+        empty = {}
+        22.times do |i|
```
alaroia-cfa, line 94: Why 22 here? Is it because you just want a big number? If so, why 22 here and 19 below?

alaroia-cfa (approved):
Approved if we can change "OTC" into "ODC" in the PDF. (Where did the PDF come from? Did you make it by hand? Neato, if so.)

Some other comments and questions inline, as well as the following:

Should we worry about even more dependents than fit on this page? I hope not, but gotta ask.

I don't fully understand the tests, specifically I don't understand the significance of nulling out some things when they're just being merged onto the empty hash. Maybe I'm missing the important code there. I'd love to get more understanding of that; maybe Tuple or Slack DM? Not a blocker.

app/lib/additional_dependents_pdf.rb
alaroia-cfa: Huh, interesting -- the PDF calls it `OTC` in the form field but the XML calls it `ODC`.

spec/lib/additional_dependents_pdf_spec.rb
bytheway875: because there are always 22 rows in the PDF 22 - 4 = 19

app/lib/additional_dependents_pdf.rb
bytheway875: good catch!!

PR #2577, opened 2022-05-18
Send intercom messages instead of creating unlinked clients

bytheway875:


app/models/vita_partner.rb
```diff
@@ -50,7 +50,6 @@ class VitaPartner < ApplicationRecord
   accepts_nested_attributes_for :source_parameters, allow_destroy: true, reject_if: ->(attributes) { attributes['code'].blank? }
 
   def self.client_support_org
-    # When a person messages us, but their contact info does not match any Client, link them to this org.
```
bytheway875, line 53: I think we use this organization for other things, so I'm not going to remove it. Does that sound right @alaroia-cfa

app/services/intercom_service.rb
```diff
@@ -3,147 +3,130 @@
...
-      reply_to_existing_intercom_thread(contact_id, portal_message.body)
-    else
-      create_new_intercom_thread(contact_id, portal_message.body)
-      if inform_of_handoff
-        contact_info = client.intake.contact_info_filtered_by_preferences
-        send_handoff_email(client) if contact_info[:email]
-        send_handoff_sms(client) if contact_info[:sms_phone_number]
-      end
-    end
+    create_intercom_message(
```
bytheway875, line 6: We now need to allow for forwarding without a message object. Thus, a lot of the supporting methods needed to either be duplicated (because they expected an object to reference, and we only have attributes) or consolidated, or I needed to create some kind of wrapper object for a non-client message object. 

To me, the logic for this whole thing can be simplified by looking at what each of these things has in common: a message body, contact method, and optionally a client, and passing those into the consuming method as params.

app/models/vita_partner.rb
alaroia-cfa: @bytheway875 I think that this org in the codebase only exists for the purpose of creating unlinked clients. I did a search for the method and that seems to agree. I think it's peaceful to delete `VitaPartner.client_support_org` from the codebase.

50/50 chance it's peaceful to delete the actual database record in prod. Check with Yvonne or client success and/or query this org to see if it's being used (look for login-able clients assigned to this org maybe?). There's a chance that the record is used for other things by Client Success but also a decent chance it's not.

bytheway875: ummm... interesting enough, this record does not exist in production. so i dont know hwo the unlinked clients thing was working.

app/services/intercom_service.rb
```diff
@@ -3,147 +3,130 @@
...
+    existing_contact = contact_from_client(client) || contact_from_email(email_address) || contact_from_sms(phone_number)
 
-    if contact_id_from_sms.present? && most_recent_conversation(contact_id).present?
-      reply_to_existing_intercom_thread(contact_id, body)
+    if existing_contact.present? && most_recent_conversation(existing_contact.id).present?
+      reply_to_existing_intercom_thread(existing_contact.id, body)
     else
-      create_new_intercom_thread(contact_id, body)
-      send_handoff_sms(incoming_sms.client) if inform_of_handoff
+      contact = existing_contact || create_or_update_intercom_contact(client: client, email_address: email_address, phone_number: phone_number)
```
caitbot, line 44: maybe one day we figure out if we can just use `create_or_update_contact` since its mostly doing the same thing as what we do to get existing

app/services/intercom_service.rb
```diff
@@ -3,147 +3,130 @@
...
+
+      create_new_intercom_thread(contact&.id, message_body)
+      if inform_of_handoff && client.present?
+        SendAutomatedMessage.send_messages(
```
caitbot, line 48: `SendAutomatedMessage` 🤩

caitbot (approved):
looks good, reviewed as pair with SB

PR #2578, opened 2022-05-18
When GYR intake is open, send clients to GYR homepage for file full return

bytheway875:


caitbot (approved):
This answers a question I had on the story I have in acceptance! Thank you!

PR #2579, opened 2022-05-18
Add rack-mini-profiler in demo

caitbot:
It would be helpful to see this little tab in demo.

image

### DIFF
--- a/config/initializers/rack_profiler.rb
+++ b/config/initializers/rack_profiler.rb
@@ -1,4 +1,4 @@
-if Rails.env.demo?
+if Rails.env.demo? || Rails.env.heroku? || Rails.env.staging?
   # Use the "Alt+P" keyboard shortcut to toggle visibility
   Rack::MiniProfiler.config.start_hidden = true
 end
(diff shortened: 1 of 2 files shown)


alaroia-cfa:
I like it! My only fear is... will it annoy people like Kelly who use demo for demos? Does it make sense to hide it behind a query param? (But then is it too hard for engineers to find?)

https://github.com/MiniProfiler/rack-mini-profiler says there's a `start_hidden` config option - but then how does one use it? Maybe we should `start_hidden` on demo?

config/initializers/rack_profiler.rb
```diff
@@ -0,0 +1,4 @@
+if Rails.env.demo?
+  # Use the "Alt+P" keyboard shortcut to toggle visibility
```
caitbot, line 2: I tested this in dev and it does indeed work! Much preferable to a URL param, IMO!

alaroia-cfa (approved):
Feel free to add to the `heroku` and `staging` envs. No re-review required if that's the only additional change.

Simple solution and good comment explaining how to use it.

PR #2582, opened 2022-05-18
Dependent destroy logic

bytheway875:
Some of the associations block deletion of clients.

### DIFF
--- a/app/models/outgoing_email.rb
+++ b/app/models/outgoing_email.rb
@@ -36,6 +36,7 @@ class OutgoingEmail < ApplicationRecord
 
   belongs_to :client
   belongs_to :user, optional: true
+  has_many :bulk_client_message_outgoing_emails, dependent: :destroy
   validates_presence_of :to
   validates_presence_of :body
   validates_presence_of :subject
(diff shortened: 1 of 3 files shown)


app/models/outgoing_email.rb
```diff
@@ -36,6 +36,7 @@ class OutgoingEmail < ApplicationRecord
 
   belongs_to :client
   belongs_to :user, optional: true
+  has_many :bulk_client_message_outgoing_emails, dependent: :destroy
```
caitbot, line 39: Ok so this should destroy the `bulk_client_message_outgoing_emails` when an `outgoing_email` is destroyed, right?

I guess I am having a hard time wrapping my head around how this blocks client deletion? What am I missing?

caitbot, line 39: Ok so this should destroy the `bulk_client_message_outgoing_emails` when an `outgoing_email` is destroyed, right?

I guess I am having a hard time wrapping my head around how this blocks client deletion? What am I missing?

bytheway875: the outgoing emails get deleted when a client gets deleted, and then if those are still present in the db on another table, it blocks deletion. This only deletes the entry that's FOR that email_id. It feels a little weird to do this, because we lose the record that that client was messaged (even though they're deleted), but since we use a reference to create the relationship i dont know if theres a way to retain it...

PR #2584, opened 2022-05-19
Consolidate and refactor file type validation logic

jenny-heath:
This started as a fairly innocent refactor and snowballed a bit, but hopefully the finished product is intuitive enough and consolidates some logic around what file types we allow for what models.

I left one comment on something I feel unsure about.

Also, I am now at the end of the day realizing that we didn't remove the file type lists from the validation error like we said we would. (the idea was to only list valid file types in the help text and have a more generic "file type invalid" type of message for the error)

### DIFF
--- a/app/forms/document_type_upload_form.rb
+++ b/app/forms/document_type_upload_form.rb
@@ -1,6 +1,7 @@
 class DocumentTypeUploadForm < QuestionsForm
-  set_attributes_for :intake, :document
-  validate :instantiate_document
+  set_attributes_for :intake, :upload
+  before_validation :instantiate_document
+  validate :validate_document
 
   def initialize(document_type, *args, **kwargs)
     @document_type = document_type
@@ -10,35 +11,26 @@ class DocumentTypeUploadForm < QuestionsForm
   def save
     return false unless valid?
 
-    @doc.save!
+    @document.save!
   end
 
   private
 
   def instantiate_document
-    document_file_upload = attributes_for(:intake)[:document]
-    if document_file_upload.blank?
-      errors.add(:document, I18n.t("validators.file_type", valid_types: FileTypeAllowedValidator.extensions(Document).to_sentence))
-      return
-    end
-
-    # Rewind to avoid IntegrityError
-    document_file_upload.tempfile.rewind
-    doc = @intake.documents.new(
-        document_type: @document_type,
-        client: @intake.client,
-        uploaded_by: @intake.client,
-        upload: {
-          io: document_file_upload.tempfile,
-          # Remove non-utf-8 characters from the original filename
-          filename: document_file_upload.original_filename.encode("UTF-8", invalid: :replace, replace: ""),
-          content_type: document_file_upload.content_type
-        }
+    @upload.tempfile.rewind if @upload.present?
+    @document = @intake.documents.new(
+      document_type: @document_type,
+      client: @intake.client,
+      uploaded_by: @intake.client,
+      upload: @upload.present? ? {
+        io: @upload.tempfile, # Rewind to avoid integrity error
+        filename: @upload.original_filename.encode("UTF-8", invalid: :replace, replace: ""), # Remove non-utf-8 characters from the original filename
+        content_type: @upload.content_type
+      } : nil
     )
-    if doc.valid?
-      @doc = doc
-    else
-      doc.errors.map { |error| error.message }.flatten.each { |msg| errors.add(:document, msg) }
-    end
+  end
+
+  def validate_document
+    errors.copy!(@document.errors) unless @document.valid?
   end
 end
--- a/app/forms/document_type_upload_form.rb
+++ b/app/forms/document_type_upload_form.rb
@@ -1,6 +1,5 @@
 class DocumentTypeUploadForm < QuestionsForm
   set_attributes_for :intake, :document
-  validates :document, file_type_allowed: true
   validate :instantiate_document
 
   def initialize(document_type, *args, **kwargs)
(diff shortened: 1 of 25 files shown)


app/models/verification_attempt.rb
```diff
@@ -13,6 +13,9 @@
 #  index_verification_attempts_on_client_id  (client_id)
 #
 class VerificationAttempt < ApplicationRecord
+  VALID_FILE_TYPES = [".jpg", ".pdf", ".png"].freeze
```
alaroia-cfa, line 16: I remember Shannon saying that it's good to omit PDF. Plus if you want this to match the VALID_MIME_TYPES then it might be good to make sure each of those is represented here.

app/models/verification_attempt.rb
```diff
@@ -13,6 +13,9 @@
...
+  # TODO: check this list?
+  VALID_MIME_TYPES = ["image/jpeg", "image/png", "image/heic", "image/bmp", "text/plain", "image/tiff", "image/gif"].freeze
```
alaroia-cfa, line 18: Seems pretty good to me, except presumably remove text/plain.

app/models/verification_attempt.rb
```diff
@@ -44,8 +47,8 @@ def state_machine
   def attachment_file_types
     [selfie, photo_identification].each do |attachment|
       if attachment.present?
-        unless FileTypeAllowedValidator::VALID_MIME_TYPES.include?(attachment.content_type)
-          errors.add(attachment.name, I18n.t("validators.file_type"))
+        unless VALID_MIME_TYPES.include?(attachment.content_type)
+          errors.add(attachment.name, I18n.t("validators.file_type", valid_types: VALID_FILE_TYPES.to_sentence(locale: I18n.locale)))
```
alaroia-cfa, line 51: Because of `switch_locale` in `ApplicationController`

```
class ApplicationController < ActionController::Base
# ...
  around_action :switch_locale
```

we probably don't need the `locale: I18n.locale` here. My request is that you manually test this by switching to an ES locale on the website with the `locale: I18n.locale` code removed.

app/validators/file_type_allowed_validator.rb
```diff
@@ -2,15 +2,19 @@ class FileTypeAllowedValidator < ActiveModel::EachValidator
...
+  # VALID_FILE_EXTENSIONS = [".jpg", ".jpeg", ".pdf", ".png", ".heic", ".bmp", ".txt", ".tiff", ".gif"]
+  # VALID_MIME_TYPES = ["image/jpeg", "image/png", "application/pdf", "image/heic", "image/bmp", "text/plain", "image/tiff", "image/gif"]
 
   def validate_each(record, attr_name, value)
     return if !value
 
     extension = File.extname(value.path)
-    unless VALID_FILE_EXTENSIONS.include?(extension.downcase)
-      record.errors.add(attr_name, I18n.t("validators.file_type"))
+    valid_types_from_record = record.classify.constantize::VALID_FILE_TYPES
```
alaroia-cfa, line 14: `record.classify.constantize::VALID_FILE_TYPES` is some good Ruby evil genius programming; I do wonder if there's a simpler way to do it. Can you do `record.class::VALID_FILE_TYPES`?

app/validators/file_type_allowed_validator.rb
```diff
@@ -2,15 +2,19 @@ class FileTypeAllowedValidator < ActiveModel::EachValidator
...
+    unless valid_types_from_record.include?(extension.downcase)
+      # TODO: has access to locale?
```
alaroia-cfa, line 16: I think yes; manually test

app/views/ctc/portal/verification_attempts/edit.html.erb
```diff
@@ -39,7 +39,7 @@
             </div>
           <% end %>
         <% end %>
-        <span class="is-tablet-hidden--inline text--small"><%= t("views.layouts.document_upload.accepted_file_types") %></span>
+        <span class="is-tablet-hidden--inline text--small"><%= t("views.layouts.document_upload.accepted_file_types", accepted_types: VerificationAttempt::VALID_FILE_TYPES.to_sentence(locale: locale)) %></span>
```
alaroia-cfa, line 42: Should already have access to locale per above

app/validators/file_type_allowed_validator.rb
alaroia-cfa: The old code used to validate file _extensions_ but this code seems to validate _mime types_. I'm not sure the new approach works.

alaroia-cfa (commented):
Some comments; we'll pair later but figured I'd tell you what I saw!

app/lib/f15080_vita_consent_to_disclose_pdf.rb
```diff
@@ -6,7 +6,7 @@ def source_pdf_name
   end
 
   def output_filename
-    "F15080 - VITA Consent To Disclose"
+    "F15080 - VITA Consent To Disclose.pdf"
```
jenny-heath, line 9: not feeling so super confident about this but without the extension the file type validator fails?

app/forms/requested_document_upload_form.rb
```diff
@@ -1,22 +1,36 @@
...
         client: @documents_request.client,
         upload: document_file_upload,
       )
     end
+
+    if doc.valid?
+      @doc = doc
+    else
+      doc.errors.map { |error| error.message }.flatten.each { |msg| errors.add(:document, msg) }
+    end
```
jenny-heath, line 34: this pattern is copied from `app/forms/document_type_upload_form.rb` above!

app/forms/document_type_upload_form.rb
```diff
@@ -19,7 +18,7 @@ def save
   def instantiate_document
     document_file_upload = attributes_for(:intake)[:document]
     if document_file_upload.blank?
```
bytheway875, line 20: After removing the file_type_allowed validator, I feel like this may only push an error if the document_type is blank... it seems like we potentially would want to do both.

bytheway875:
I pushed a new commit on top of this that refactors the form logic a little bit, including using `.copy!` to write the errors from one object to the other, and moving the instantiate method to a `before_validate` hook so that there's a bit more separation of concerns.

caitbot (approved):

PR #2586, opened 2022-05-19
Save home_location and update 8812 to be based on XML

caitbot:
This is preliminary work for making 1040 + 8812 compliant with PR scheduling.

### DIFF
--- a/app/lib/irs8812_ty2021_pdf.rb
+++ b/app/lib/irs8812_ty2021_pdf.rb
@@ -6,10 +6,7 @@ class Irs8812Ty2021Pdf
   end
 
   def initialize(submission)
-    @submission = submission
-    @qualifying_dependents = submission.qualifying_dependents
-    @ctc_qualifying_dependents =  @qualifying_dependents.select(&:qualifying_ctc?)
-    @benefits = Efile::BenefitsEligibility.new(tax_return: submission.tax_return, dependents: @qualifying_dependents)
+    @xml_document = SubmissionBuilder::Ty2021::Documents::Schedule8812.new(submission).document
   end
 
   def hash_for_pdf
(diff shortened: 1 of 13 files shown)


app/lib/additional_dependents_pdf.rb
```diff
@@ -7,8 +7,8 @@ def source_pdf_name
   end
 
   def initialize(submission, start_node: 4)
-    # For some PDF fields, use values from the database b/c the XML values are truncated or missing.
-    @xml_document = SubmissionBuilder::Ty2021::Documents::Irs1040.new(submission).document
+    # Use base because these fields do not deviate amongst xml versions
+    @xml_document = SubmissionBuilder::Ty2021::Documents::Irs1040::Base.new(submission).document
```
bytheway875, line 11: ohh this is going to be interesting, because if we have sub-files that dependent on specific parent files (for example, divergenct 1040s that both fill out an 8812 but have different values), we're probably going to have to deal with that somehow within these files. 🤔

app/lib/irs8812_ty2021_pdf.rb
```diff
@@ -6,10 +6,7 @@ def source_pdf_name
   end
 
   def initialize(submission)
-    @submission = submission
-    @qualifying_dependents = submission.qualifying_dependents
-    @ctc_qualifying_dependents =  @qualifying_dependents.select(&:qualifying_ctc?)
-    @benefits = Efile::BenefitsEligibility.new(tax_return: submission.tax_return, dependents: @qualifying_dependents)
+    @xml_document = SubmissionBuilder::Ty2021::Documents::Schedule8812::Base.new(submission).document
```
bytheway875, line 9: Like... could we potentially run into that here? Do any of these values in base that are consumed diverge in the PR case?

bytheway875: This always looks at Base, but seems like there is both a Base and PR version of the 8812. Do we need that?

When there is divergence in those docs, maybe we should create a method so that we can get the right method here...

app/lib/additional_dependents_pdf.rb
caitbot: Discussed offline, going to move to a different pattern (updating benefits calculations, most of these changes should really take place upstream).

app/lib/irs8812_ty2021_pdf.rb
caitbot: Discussed offline, going to move to a different pattern (updating benefits calculations, most of these changes should really take place upstream).

bytheway875:
i sometimes go back and forth on whether it makes sense to have an unfilled option for the questions since we hvae them on some others, but then im like nil IS unfilled... so this makes sense haha

bytheway875 (approved):

PR #2588, opened 2022-05-23
Add PDF reader to verifications page when there are PDF uploads.

bytheway875:
Screen Shot 2022-05-20 at 11 56 29 AM

### DIFF
--- a/app/views/hub/verification_attempts/show.html.erb
+++ b/app/views/hub/verification_attempts/show.html.erb
@@ -25,13 +25,21 @@
       <% else %>
         <div class="grid-flex verification-attempts-photo-container">
           <% if @verification_attempt.selfie.present? %>
-            <div class="item-15r">
-              <%= image_tag transient_storage_url(@verification_attempt.selfie.blob), id: "selfie" %>
+            <div class="item item-15r">
+              <% if @verification_attempt.selfie.content_type == "application/pdf" %>
+                <embed src="<%= transient_storage_url(@verification_attempt.selfie.blob) %>" height="500" type="application/pdf" />
+              <% else %>
+                <%= image_tag transient_storage_url(@verification_attempt.selfie.blob), id: "selfie" %>
+              <% end %>
             </div>
           <% end %>
           <% if @verification_attempt.photo_identification.present? %>
-            <div class="item-15r">
-              <%= image_tag transient_storage_url(@verification_attempt.photo_identification.blob), id: "photo_id" %>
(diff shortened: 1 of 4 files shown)


app/views/hub/verification_attempts/show.html.erb
```diff
@@ -25,13 +25,21 @@
       <% else %>
         <div class="grid-flex verification-attempts-photo-container">
           <% if @verification_attempt.selfie.present? %>
-            <div class="item-15r">
-              <%= image_tag transient_storage_url(@verification_attempt.selfie.blob), id: "selfie" %>
+            <div class="item item-15r">
```
caitbot, line 28: we can probably remove the `item-15r` class since we deleted it? maybe not if its actually just defined elsewhere

app/views/hub/verification_attempts/show.html.erb
```diff
@@ -25,13 +25,21 @@
...
+              <% if @verification_attempt.selfie.content_type == "application/pdf" %>
+                <embed src="<%= transient_storage_url(@verification_attempt.selfie.blob) %>" height="500" type="application/pdf" />
```
caitbot, line 30: is it just me or is it weird that people are uploading selfies as pdfs?

app/views/public_pages/home.html.erb
```diff
@@ -26,7 +26,7 @@
         <div class="grid__item">
           <b>
             <%= t('views.public_pages.home.ctc_banner.claim_ctc') %>
-            <%= link_to(t('views.public_pages.home.ctc_banner.click_here'), url_for(host: MultiTenantService.new(:ctc).host, controller: "/public_pages", action: 'home'), class: "text--underline link--subtle") %>
+            <%= link_to(t('views.public_pages.home.ctc_banner.click_here'), url_for(host: MultiTenantService.new(:ctc).host, controller: "/public_pages", action: 'home', params: { source: "gyrhomebanner"}), class: "text--underline link--subtle") %>
```
caitbot, line 29: I am surprised to see this snuck it's way in here. had we just forgotten to change it in the other PR?

caitbot (approved):
looks good, made a suggestion, left a q or two

app/views/hub/verification_attempts/show.html.erb
bytheway875: its defined elsewhere, but extended here... i dont think that particular class should have been the one that was extended though.

PR #2590, opened 2022-05-23
Update benefits calculation and the resulting XML for Puerto Rico clients

caitbot:


app/lib/efile/benefits_eligibility.rb
```diff
@@ -60,6 +60,7 @@ def advance_ctc_amount_received
     # A quick calculation for ODC (Other Dependents Credit) which does not get paid out to our filers,
     # but is needed for the 8812 calculation.
     def odc_amount
+      return nil if intake.home_location_puerto_rico?
```
bytheway875, line 63: should we comment the difference between returning nil or 0? 🤔  maybe at the TOP of benefits eligibility?

PR #2599, opened 2022-05-25
Add new columns to AcceptedTaxReturnAnalytics and backfill rake task

embarnard:


app/state_machines/efile_submission_state_machine.rb
```diff
@@ -120,21 +120,26 @@ class EfileSubmissionStateMachine
     tax_return.transition_to(:file_accepted)
 
     benefits = Efile::BenefitsEligibility.new(tax_return: tax_return, dependents: submission.qualifying_dependents)
-    ctc_amount = benefits.ctc_amount / 2
-    refund_amount = client.intake.claim_owed_stimulus_money_no? ? 0 : (benefits.eip1_amount + benefits.eip2_amount)
-    eip3_amount = benefits.eip3_amount
+    eip1_and_eip2_amount = [benefits.eip1_amount, benefits.eip2_amount].compact.sum
```
bytheway875, line 123: There's a lot of calculations happening here. I wonder if its a good time time to move these calculations into a method on accepted_tax_return_analytics where we can make the analytics object based on the tax return object passed in so that we have a separation of concerns, especially now that we have more cals and data here.

app/state_machines/efile_submission_state_machine.rb
```diff
@@ -120,21 +120,26 @@ class EfileSubmissionStateMachine
...
+    total_refund_amount = [benefits.outstanding_ctc_amount, benefits.outstanding_recovery_rebate_credit].compact.sum
+    eip3_amount_received = tax_return.intake.eip3_amount_received || 0
```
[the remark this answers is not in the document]
bytheway875: Alternatively, we can set a database default for this value to 0 OR create a custom reader on the model that reads from the db value if present or returns 0.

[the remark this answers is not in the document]
bytheway875: i think putting it on benefits may make the most sense, since in the context of the intake there may be some logic that distinguishes between a 0 and nil value somewhere

app/state_machines/efile_submission_state_machine.rb
```diff
@@ -120,21 +120,26 @@ class EfileSubmissionStateMachine
...
+      advance_ctc_amount_cents: benefits.advance_ctc_amount_received * 100,
+      outstanding_ctc_amount_cents: benefits.outstanding_ctc_amount * 100,
+      ctc_amount_cents: benefits.ctc_amount * 100,
+      eip3_amount_received_cents: eip3_amount_received * 100,
+      eip3_amount_cents: benefits.eip3_amount * 100,
+      outstanding_eip3_amount_cents: benefits.outstanding_eip3 * 100,
+      total_refund_amount_cents: total_refund_amount * 100,
     )
 
     send_mixpanel_event(submission, "ctc_efile_return_accepted", data: {
```
bytheway875, line 139: this still seems right to me, but might be worth checking with product if you havent already to see whether there are any other values that they want to receive in mixpanel, or confirm how/if this data is being used.

db/schema.rb
```diff
@@ -19,9 +19,15 @@
...
     t.bigint "advance_ctc_amount_cents"
     t.datetime "created_at", null: false
+    t.bigint "ctc_amount_cents"
+    t.bigint "eip1_and_eip2_amount_cents"
     t.bigint "eip3_amount_cents"
-    t.bigint "refund_amount_cents"
+    t.bigint "eip3_amount_received_cents"
+    t.bigint "outstanding_ctc_amount_cents"
+    t.bigint "outstanding_eip3_amount_cents"
     t.bigint "tax_return_id"
```
bytheway875, line 28: since there is a relationship to tax_return which indicates the year, we may not actually need to explicitly store the tax return year. 🤔  data team should be able to query for only the accepted analytics using a join to the tax return year which already exists.

We also might want to consider adding an index on tax return on year OR on this table for tax_return_year if we keep it, since data team will likely be querying based on year pretty often and an index can help that query's performance.

bytheway875 (commented):
Please consider the feedback and suggested refactors before moving forward! Happy to chat more  async on the PR or over Tuple.

app/models/accepted_tax_return_analytics.rb
```diff
@@ -2,18 +2,38 @@
...
+#  tax_return_id                 :bigint
 #
 # Indexes
 #
 #  index_accepted_tax_return_analytics_on_tax_return_id  (tax_return_id)
 #
 class AcceptedTaxReturnAnalytics < ApplicationRecord
   belongs_to :tax_return
+
+  def calculated_benefits_attrs
```
bytheway875, line 25: nice!

PR #2600, opened 2022-05-25
Refactor standard deduction

caitbot:
The logic for calculating an applied standard deduction lived on `tax_return` and the `StandardDeduction` class was mostly meant to be used as a way to grab `base` standard deduction amounts. Built this way, it would require additional logic be added each year, and with each additional calculation constraint, more logic be added to `tax_return`. 

In a conversation with SB, we dreamed of a world where all the logic for "constructing" a standard deduction was removed from the model and contained in its own service class, especially as our world of standard deduction calculation grows more complex.

These changes introduce that with the `AppliedStandardDeduction` class. 

They also separate out the use case of obtaining a `base` standard deduction amount for displaying in the UI -- now accessed through `StandardDeduction.base_deductions` which returns a hash of deductions for a given year mapped to filing statuses. 

From year to year, we will need to update the yaml when standard deduction amounts are updated.


app/lib/applied_standard_deduction.rb
```diff
@@ -0,0 +1,43 @@
...
+  def initialize(tax_return:)
+    @tax_return = tax_return
+    @tax_year = tax_return.year
+    @filing_status = tax_return.filing_status
+    @intake = tax_return.intake
+  end
+
+  def applied_standard_deduction
+    return nil if @intake.is_ctc? && @intake.home_location_puerto_rico?
+    raise NotImplementedError, "Standard deduction rule not implemented for #{@tax_year}" unless StandardDeductions.base_deductions(tax_year: @tax_year).present?
```
bytheway875, line 13: Nit, but maybe we move the raise before the return nil? That would make the heirarchy of the method return from broad -> specific.

app/lib/applied_standard_deduction.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    StandardDeductions.base_deductions(tax_year: @tax_year)[@filing_status] + additional_blind_standard_deduction + additional_older_than_65_standard_deduction
+  end
+
+  private
+
+  def additional_older_than_65_standard_deduction
+    deductions = StandardDeductions.older_than_65_deductions(tax_year: @tax_year)
```
bytheway875, line 21: this is probably so negligible, but since this only applies to people that are over 65, we could add a return at the top that gets out of here if neither primary or spouse is over the age limit.

Also, I noticed that the method on tax_return does a date comparison every time we call the method. its probably really negligible memory allocation wise, but i guess it does that calculation every time we call the method. if we really wanted to, we could call the method once, set it to a variable, and then use the same computed value for the comparison.

Or maybe I'm totally wrong and the method only does the computation once and there's some kind of caching, but i think thats only for active record queries that rails can do that.

```
primary_senior = @tax_return.primary_age_65_or_older?
spouse_senior = @tax_return.spouse_age_65_or_older?

return unless primary_senior or spouse_senior

case ...
when "single"
  return deductions[:single_filer] if primary_senior
...
```

PR #2602, opened 2022-05-26
Add the flipper gem for feature flagging

caitbot:
Adds the flipper-ui, as documented [here](https://www.flippercloud.io/docs/ui)

Accessed at `/hub/flipper` and requires an admin account.

Flipper also offers a cloud version, but that seemed like too much for our use case. Mostly we need ways to toggle something on or off. You do that by asking `Flipper.enabled?(:flag_name)`

Spec to come...

image

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,6 +77,11 @@ gem 'redcarpet'
 gem 'platform-api'
 gem 'strong_migrations'
 
+# Use Flipper for feature flagging
+gem 'flipper'
+gem 'flipper-active_record'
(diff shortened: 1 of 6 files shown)


bytheway875:
@caitbot This is super cool! Looks like we can also use this to do some A/B testing stuff if we want to.

bytheway875:
@caitbot One more thing! Should we ask if we can/should add a button to this page on Admin Tools? I think it will likely be mostly engineering using it, but could be nice so we don't have to remember the pathname

PR #2643, opened 2022-05-26
Add no dependents with actc offboarding

wschaefer:


app/controllers/ctc/offboarding/actc_without_dependents_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    class ActcWithoutDependentsController < CtcController
+      helper_method :illustration_path, :illustration_folder, :prev_path
+
+      layout "intake"
+
+      def show; end
+
+      def file_with_gyr
+        current_intake.touch(:clicked_file_with_gyr_at)
+        redirect_to url_for(host: MultiTenantService.new(:gyr).host, controller: "/public_pages", action: 'home', params: { source: 'received_ctc' })
```
bytheway875, line 12: this will create a url that looks like 
getyourrrefund.org/en/?source=received_ctc

if you use the recently added root_with_source_url(host: ..., source: "received_ctc"),
it will create the url as getyourrrefund.org/en/received_ctc, which stores the source param then issues a redirect to home so that the source param isn't maintained in the url.

app/controllers/ctc/questions/no_dependents_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Ctc
+  module Questions
+    class NoDependentsController < QuestionsController
+      include AuthenticatedCtcClientConcern
+
+      layout "intake"
+
+      def self.show?(intake)
+        intake.dependents.none?
```
bytheway875, line 9: the ticket doesnt explicitly indicate, but I wonder whether people with no QUALIFYING dependents should also see this flow... to do that, we'd need to run each dependent through the qualifications service.

I'm not super up to date with the context of the original ticket, so not sure whether it should be raw dependents being 0, or no dependents who qualify for benefits. 🤔

bytheway875: yeah, the figma indicates some text about "eligible" dependents, so I guess there is a case where someone could have added 5 dependents, only one was originally qualifying, so we'd want to show this page if they remove the one qualifying dependent.

wschaefer: Yeah @embarnard and I talked about it this afternoon and I'm going to confirm with Yvonne but it seems likely that it should only be if there are eligible dependents.

wschaefer:
The DB on this heroku app got messed up so I have to try to open a fresh PR from this branch

PR #2654, opened 2022-05-27
Validate format of drivers license numbers

bytheway875:


spec/forms/ctc/drivers_license_form_spec.rb
```diff
@@ -23,6 +23,39 @@
       end
     end
 
+    context "license number" do
```
caitbot, line 26: Consider testing this instead in a spec for the `AlphanumericValidator` rather than duplicating it this and the `spouse` forms.

Not blocking but preferrable.

app/validators/alphanumeric_validator.rb
```diff
@@ -0,0 +1,9 @@
+class AlphanumericValidator < ActiveModel::EachValidator
```
caitbot, line 1: nice

caitbot (approved):

bytheway875:
@caitbot i moved the tests into the validator spec and then left one test each in both forms to ensure it was applied in both.

PR #2704, opened 2022-05-31
Update great grandchild relationship 182248770

bytheway875:


wschaefer (approved):
It occurs to me that because we generate the PDF from the XML now that in this case the PDF would say 'Grandchild' even though the client entered 'Great-grandchild' - I wonder if we should check with Yvonne to see if that's what we should really be doing?

PR #2707, opened 2022-05-31
Update webpack-dev-server version

wschaefer:
To get it to pull in a version of eventsource that does not
have a security vulnerability

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "expose-loader": "^2.0.0",
     "jest": "^26.5.2",
     "jest-junit": "^12.0.0",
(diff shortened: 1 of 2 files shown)


caitbot (commented):
Wow wow so many yarn updates. I know when doing `bundle install` to update a dependency, you can do `--conservative` to try to lower the number of updates. I tried to do a search to see if there was something similar we could use here, but couldn't find it. In this case, I'm going to assume these were all just cascading dependency updates? Probably good that we update them then..

Looks good, thanks for keeping us secure!

PR #2708, opened 2022-05-31
Use mimetype instead of extension to validate documents

bytheway875:


caitbot (approved):
huh TIL mime type stands for `Multipurpose Internet Mail Extensions`

this seems like a straight-forward and welcome change to me

PR #2710, opened 2022-06-01
Do not require authentication for local or heroku uses of feature flags

caitbot:
It was kind of hard to use and there's no need to authenticate for local uses or on heroku since dev + product are the only ones with access to those.


embarnard (approved):
Yay! Looks good

PR #2714, opened 2022-06-02
Move USPS address validation into CTC flow

jenny-heath:
I haven't yet removed the existing call to the API (at submission) because I'm not sure whether we want to! If it should be removed, would love input on whether that should be part of this PR or can be a second step/commit

image

image

### DIFF
--- a/app/forms/ctc/mailing_address_form.rb
+++ b/app/forms/ctc/mailing_address_form.rb
@@ -1,14 +1,53 @@
 module Ctc
   class MailingAddressForm < QuestionsForm
     set_attributes_for :intake, :street_address, :street_address2, :state, :city, :zip_code
+    set_attributes_for :misc, :address_not_found
 
     validates_presence_of :street_address
     validates_presence_of :city
     validates_presence_of :state
     validates :zip_code, us_or_puerto_rico_zip_code: true
+    validate :usps_valid_address
 
     def save
+      attrs = {
+        zip_code: address_service.zip_code,
+        street_address: address_service.street_address,
+        street_address2: nil,
+        state: address_service.state,
+        city: address_service.city
+      }
+      @intake.update(attrs)
+    rescue => e
       @intake.update(attributes_for(:intake))
+    end
+
+    private
+
+    def usps_valid_address
+      return unless errors.blank?
+
+      @intake.assign_attributes(attributes_for(:intake))
+      if !address_service.valid?
+        case address_service.error_code
+        when "USPS-2147219400"
+          errors.add(:city, I18n.t("forms.errors.mailing_address.city"))
+        when "USPS-2147219402"
+          errors.add(:state, I18n.t("forms.errors.mailing_address.state"))
+        when "USPS-2147219403"
+          errors.add(:address_not_found, I18n.t("forms.errors.mailing_address.multiple"))
+        when "USPS-2147219401"
+          errors.add(:address_not_found, I18n.t("forms.errors.mailing_address.not_found"))
+        when "USPS-2147219396"
+          errors.add(:address_not_found, I18n.t("forms.errors.mailing_address.invalid"))
+        end
+      end
+    rescue
+      true
+    end
+
+    def address_service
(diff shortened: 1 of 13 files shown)


app/forms/ctc/mailing_address_form.rb
```diff
@@ -1,15 +1,35 @@
...
+      @intake.assign_attributes(attributes_for(:intake))
+      address_service = StandardizeAddressService.new(@intake)
+      if !address_service.valid?
+        case address_service.error_code
+        when "USPS-2147219400"
+          errors.add(:city, I18n.t("forms.errors.mailing_address.city"))
+        when "USPS-2147219402"
+          errors.add(:state, I18n.t("forms.errors.mailing_address.state"))
+        when "USPS-2147219403", "USPS-2147219401", "USPS-2147219396"
+          errors.add(:address_not_found, "Error: Address Not Found")
```
jenny-heath, line 30: thinking these error codes / attached fields+errors should be in their own data structure of some sort but not sure how best to do it. maybe something like:
``` 
[
  {
    code: ""USPS-2147219400"",
    attr: :city,
    error: I18n.t("forms.errors.mailing_address.city")
  }
]
```
or
```
{
  "USPS-2147219400": {
    attr: :city,
    error: I18n.t("forms.errors.mailing_address.city")
  }
}
```
not sure whether the I18n function will work in a constant...

bytheway875: I wonder if you could move all of this logic into the translations themselves, especially given that they're each tied to specific error codes that could be used as keys:

````
forms:
  errors:
     mailing_address:
        USPS-2147219402: "Error: Invalid State"
        USPS-2147219402: "Error: Invalid City"
        USPS-2147219403: &generic_address_error "Error: Invalid Address"
        USPS-2147219401: *generic_address_error
        ... etc
```

This falls down if there is a new or different error code that we get back from the IRS API, but I'm fairly certain you could create a fallback pretty easily.

spec/support/mock_successful_usps_api_call.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.configure do |config|
+  config.before(:each) do |example|
+    if example.metadata[:checks_address_with_usps_api]
```
bytheway875, line 3: i think it may make more sense to change the name of this to stub_usps_address_requests

The current name of "checks_address..." had me thinking that we were actually calling out to the API in the tests when in reality we're intercepting and providing mocked data.

I guess the question is whether these support assertions should be written from the point of view of the behavior of the underlying code (which checks the address) or the tests themselves (which need to stub the request). I think the way we use them currently, writing them from the POV of the test setup behavior would be more consistent.

app/forms/ctc/mailing_address_form.rb
bytheway875: I wonder if you could move all of this logic into the translations themselves, especially given that they're each tied to specific error codes that could be used as keys:

````
forms:
  errors:
     mailing_address:
        USPS-2147219402: "Error: Invalid State"
        USPS-2147219402: "Error: Invalid City"
        USPS-2147219403: &generic_address_error "Error: Invalid Address"
        USPS-2147219401: *generic_address_error
        ... etc
```

This falls down if there is a new or different error code that we get back from the IRS API, but I'm fairly certain you could create a fallback pretty easily.

app/forms/ctc/mailing_address_form.rb
```diff
@@ -1,14 +1,49 @@
...
+
+      @intake.assign_attributes(attributes_for(:intake))
+      if !address_service.valid?
+        case address_service.error_code
+        when "USPS-2147219400"
+          errors.add(:city, I18n.t("forms.errors.mailing_address.city"))
+        when "USPS-2147219402"
+          errors.add(:state, I18n.t("forms.errors.mailing_address.state"))
+        when "USPS-2147219403", "USPS-2147219401", "USPS-2147219396"
+          errors.add(:address_not_found, "Error: Address Not Found")
```
jenny-heath, line 38: replace "address not found" with actual error

PR #2717, opened 2022-06-03
Show different Spanish text if Puerto Rico client and if single vs mfj

embarnard:


embarnard:
This PR aims to provide different Spanish translations for clients residing in Puerto Rico and different text for PR clients that are single versus married filing jointly.

Below is a single PR client
PR client single

Below is a mfj PR client
PR client mfj

Below is a client that does not reside in PR and is single
Not PR client single

config/locales/en.yml
```diff
@@ -2237,11 +2237,18 @@ en:
               p1: 'Income could come from any of the following sources:'
             title: What counts as income?
           puerto_rico_income_qualifications:
-          - earned less than %{standard_deduction} in total income
-          - earned less than $400 in self-employment income
-          - did not receive the Advance Premium Tax Credit
-          - are a bonafide resident of Puerto Rico
-          - are not required to file a full tax return for any other reason
+            one:
```
caitbot, line 2240: ooo i like this use of the `i18n` count parameter, slick

caitbot (approved):
looks good

PR #2721, opened 2022-06-06
Write consented_to_service_at to client objects

bytheway875:
The bug where some clients are able to log in but are not discoverable in the application has been filed again, so I decided to take care of it in the way that we'd discussed in a CGC meeting a while back: writing consented_to_service_at to the client object so that we can grab consented clients without doing a join between intake and client.

In this PR:
- Switch to writing consented_to_service_at onto client, and reading primary_consented_to_service_at from client if it exists
- Write a backfill so that all consented clients will have a value for consented_to_service_at (including archived intakes)
- Update the delete barely started intake code to delete unconsented clients code (we're still not actively USING this code, but I'd like to have us start running this)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_05_26_161533) do
+ActiveRecord::Schema[7.0].define(version: 2022_06_06_183453) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "plpgsql"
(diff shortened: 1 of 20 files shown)


db/schema.rb
```diff
@@ -569,6 +571,20 @@
     t.index ["client_id"], name: "index_consents_on_client_id"
   end
 
+  create_table "contents", force: :cascade do |t|
```
caitbot, line 574: remove this?

caitbot (approved):
reviewed/discussed as pair, looks good, except remove the schema change

db/schema.rb
bytheway875, line 574: oh no, remove

PR #2726, opened 2022-06-07
Add filing_status to tax returns for CTC intakes in seeder

wschaefer:
Otherwise when you try to view the intakes in the hub you get an error


bytheway875 (approved):
Woo! thanks for doing this.

PR #2771, opened 2022-06-15
Update filed-prior-tax-year for Puerto Rico residents

caitbot:
image

### DIFF
--- a/app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
+++ b/app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
@@ -1,0 +1,22 @@
+<% @main_question = t("views.ctc.questions.filed_prior_tax_year.pr.title", prior_tax_year: prior_tax_year) %>
+
+<% content_for :page_title, @main_question %>
+
+<% content_for :form_card do %>
+  <h1 class="h2"><%= @main_question %></h1>
+  <p><%= t("views.ctc.questions.filed_prior_tax_year.help_text", prior_tax_year: prior_tax_year) %></p>
+  <p class="text--small text--bold"><%= t("views.ctc.questions.filed_prior_tax_year.pr.note") %></p>
+  <%= form_with model: @form, url: current_path, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+    <%= f.cfa_radio_set(
+          :filed_prior_tax_year,
+          collection: [
+              { value: :filed_full, label: t("views.ctc.questions.filed_prior_tax_year.pr.filed_full") },
+              { value: :did_not_file, label: t("views.ctc.questions.filed_prior_tax_year.pr.did_not_file", current_tax_year: current_tax_year, prior_tax_year: prior_tax_year) }
+          ]
+      )
+  %>
+    <div class="spacing-above-60">
+      <%= f.submit t("general.continue"), class: "button button--primary button--wide" %>
+    </div>
+  <% end %>
+<% end %>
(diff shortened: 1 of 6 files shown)


app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<% @main_question = t("views.ctc.questions.filed_prior_tax_year.pr.title", prior_tax_year: prior_tax_year) %>
```
caitbot, line 1: there was probably a fancy way to inline these changes, but the copy changed for both the title, and the options, and ultimately I deemed messing with the options inline to be too much of a mess and so went the duplication route.

this way, as opposed to creating a whole new page, at least allows us to leverage the same controller + form for it all

config/locales/en.yml
```diff
@@ -2169,6 +2169,11 @@ en:
           filed_full: Yes, I filed a full return
           filed_non_filer: Yes, I filed using the IRS non-filer portal or GetCTC
           help_text: Your %{prior_tax_year} tax return would have originally been due on May 17, 2021 and would have typically been completed in 2021.
+          pr:
+            did_not_file: No, I didn’t file a 2020 IRS tax return
+            filed_full: Yes, I filed an IRS tax return
+            note: 'Note: federal return with the IRS is separate from a Puerto Rico return you may have filled with Hacienda.'
```
bytheway875, line 2175: should this be filed? I guess filled makes similar sense... 🤔

app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
bytheway875, line 1: interesting. i think i was expecting the approach of creating a separate controller / new question for the different question and then using the `show?` function to decide whether to show the PR one or the US one. Tbh, this is actually simpler to reason about from within the controller.

caitbot: by `tbh this is simpler to reason about`, do you mean this approach or the approach where there are two separate controllers with `show?` logic?

config/locales/en.yml
caitbot: lol good point, I went ahead and asked anu if shes got a preference or meant for one or the other

bytheway875 (approved):
one question about whether the copy has an extra L and some commentary.

app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
bytheway875: this one! but also, where is the controller change for this? dont we need to have logic to render this when its PR?

app/views/ctc/questions/filed_prior_tax_year/edit.html.erb
```diff
@@ -1,22 +1,26 @@
-<% @main_question = t("views.ctc.questions.filed_prior_tax_year.title", prior_tax_year: prior_tax_year) %>
+<% if current_intake.puerto_rico_filing? %>
```
caitbot, line 1: @bytheway875 its right here (: hahaha it isn't in the controller.

I tried putting it in the controller. Having the controller render a different template for edit based on home location would be my preferred method. 

However, there's some complexity in doing that, since the `QuestionController` defines `edit` and `update` and it turns out it isn't even as simple as calling `super` because when a form fails to save `update` calls `render :edit` so really, to have the controller correctly handle the rendering, you need to copy the entire `update` method into the child class and handle the re-rendering of edit based on home location.

This never seems worth it, so I usually just branch in the template, like here.

app/views/ctc/questions/filed_prior_tax_year/_pr_edit.html.erb
caitbot: I left a comment on the pertinent line of code explaining!

PR #2772, opened 2022-06-15
Client filtering based on consented_to_Service_at instead of tax return status

bytheway875:


spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -314,12 +314,12 @@
         let!(:tobias) { create :client, vita_partner: organization, intake: create(:intake, :filled_out, preferred_name: "Tobias", needs_help_2018: "yes", preferred_interview_language: "es", state_of_residence: "TX") }
         let!(:tobias_2019_return) { create :tax_return, :intake_in_progress, client: tobias, year: 2019, assigned_user: assigned_user }
         let!(:tobias_2018_return) { create :tax_return, :intake_in_progress, client: tobias, year: 2018, assigned_user: assigned_user }
-        let!(:lucille) { create :client, vita_partner: organization, intake: create(:intake, preferred_name: "Lucille") }
+        let!(:lucille) { create :client, vita_partner: organization, consented_to_service_at: nil, intake: create(:intake, preferred_name: "Lucille") }
         let!(:lucille_2018_return) { create(:tax_return, :intake_before_consent, client: lucille, year: 2018, assigned_user: assigned_user) }
```
caitbot, line 318: I would recommend removing the tax_return status specification here as well, since it isn't necessary anymore but not blocking

spec/factories/intakes.rb
```diff
@@ -552,7 +552,7 @@ def self.roman_numerals(n)
 
   factory :intake, class: Intake::GyrIntake do
     had_wages { :unfilled }
-    client
+    client { create :client, consented_to_service_at: nil }
```
caitbot, line 555: Its interesting to me that for the default `client` factory we set the `consented_to_service_at` but here we leave it `nil`, why is that? I guess I would expect most intakes or our concept of a "default" intake to have a consented client.

caitbot (approved):
looks good to me, I left a suggestion and a question

spec/factories/intakes.rb
bytheway875: the thought process behind this (to the limited extent there was one) is that we start to treat someone as a "client" in the hub once they consent to service -- they are not viewable until then. So most of the hub stuff relies on the concept of a consented client. Alternatively, in the intake we generally start with an unconsented client and they grow into being consented.

Realistically, though, this solves a problem of all of the created clients as a result of creating an intake and a client at the same time from having TWO consented clients.

With the way our factories are set up, 

`create :client, intake: (create :intake)` does a create on the intake, which creates a client and then creates ANOTHER client. Realistically, it should be a `create :client, intake: (build :intake)` to prevent that behavior, but it's so widespread (and in the case of a similar problem on tax returns, prevents transition of tax returns to other states 😬 ) that this creates a bit of a work around.

spec/controllers/hub/clients_controller_spec.rb
bytheway875: a lot of the tests within the context still look for information about the tax return because these tests `render_views` and look for the presence of the right tax return info

spec/factories/intakes.rb
caitbot: Got it, got it, thats pretty funny. Big oops.

caitbot: I like your justification though, beyond fixing the double client thing.

PR #2834, opened 2022-06-16
Prevent app crash when stage/status are nil for tax return

bytheway875:


alaroia-cfa (approved):
Simple and nice.

This could be a case for a shared example called `it_returns_nil_given_nil`, or it could be a case for writing positive assertions in addition to these negative assertions. Not a blocker at all.

PR #2840, opened 2022-06-16
Puerto Rico landing page for getctc.org

bytheway875:


config/locales/es.yml
```diff
@@ -2718,6 +2718,22 @@ es:
         - Si tiene un problema de privacidad o de uso de datos sin resolver que no hayamos solucionado satisfactoriamente, favor de comunicarse con nuestro proveedor de resolución de disputas con sede en Estados Unidos (sin costo alguno) a <a href="https://feedback-form.truste.com/watchdog/request" target="_blank" rel="noopener noreferrer">https://feedback-form.truste.com/watchdog/request</a>.
         - 'Si tiene preguntas, comentarios, inquietudes o quejas sobre este Sitio, comuníquese con nosotros por correo electrónico a support@getyourrefund.org o por correo postal a: %{email_link}'
         51_do_our_best: Haremos todo lo posible para resolver el problema.
+      puerto_rico:
```
wschaefer, line 2721: Do we have Spanish for the es.yml?

app/views/ctc/ctc_pages/puerto_rico.html.erb
```diff
@@ -0,0 +1,136 @@
+<% content_for :page_title do %>
```
wschaefer, line 1: It looks like a lot of this is the same of the regular home page. To make them easier maintain do you think it would make sense to extract some partials for the parts that are the same?

wschaefer (approved):

PR #2845, opened 2022-06-17
Offboard PR clients if no dependents qualify for CTC

caitbot:


config/locales/en.yml
```diff
@@ -2328,7 +2329,12 @@ en:
             us_territory: Other U.S. territory (such as Guam, American Samoa, U.S. Virgin Islands)
           title: Where was your main home for %{current_tax_year}?
         no_dependents:
+          add_child: Add a child
           add_dependents: Add more dependents
+          pr:
+            subtitle1: You will not receive the Child Tax Credit. You have not added any dependents eligible for the Child Tax Credit, so we can not file a return at this time.
```
bytheway875, line 2335: my general preference when we've got multiple paragraphs would be to do 

```
subtitle:
- something
- something
```

and then loop through in the view. It doesnt really make a difference and im sure there are solid reasons for doing the opposite but idk.

bytheway875: also not a blocker, just sayin

caitbot: sweet, I like that, it does seem nicer.

PR #2846, opened 2022-06-17
One more database connection [#182457065]

alaroia-cfa:
https://www.pivotaltracker.com/n/projects/2409240/stories/182457065

Audit Hour

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,11 +1,11 @@
 local_default: &local_default
   adapter: postgis
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + 1 %>
   timeout: 5000
 
 deploy_default: &deploy_default
   adapter: postgis
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + 1 %>
   timeout: 5000
   url: <%= Rails.application.credentials[:database_url] %>
 


config/database.yml
```diff
@@ -1,11 +1,11 @@
 local_default: &local_default
   adapter: postgis
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + 1 %>
```
caitbot, line 3: I see so the 5 is just a fallback default value and so really the end result will either be `3` (based on 2 being the `RAILS_MAX_THREADS` or `6`)

Why not just bump the `RAILS_MAX_THREADS`?

caitbot (approved):
looks good, I left a q

config/database.yml
alaroia-cfa: `RAILS_MAX_THREADS` defines the number of threads per Rails process. In Aptible, we have one Rails process per `web` container. In Aptible, we have 1 CPU core in each `web` container [1]. A few months ago, I noticed that tail latency [2] was weirdly high. I had recently read this Dan Luu blog post remarking that "Exceeding the reserved core limit is extremely bad" [3], and that made a lot of sense to me based on my experience maintaining/optimizing Linux boxes. For that reason, I thought 5 was way too many Rails threads. Indeed, when we reduced this number, tail latency improved.

If we bump `RAILS_MAX_THREADS` we'll probably have 20%+ higher tail latency, so I'd prefer to bump a more surgical number.

1. https://deploy-docs.aptible.com/docs/cpu-limits unless we have >4GB RAM, we have at most 1 CPU core per container
2. See e.g. https://www.section.io/blog/preventing-long-tail-latency/ for a reference
3. https://danluu.com/cgroup-throttling/

alaroia-cfa:
I now think the theory behind this pull request is wrong, although I still intend to merge it and try it!

I tried to reproduce the bad behavior by setting `RAILS_MAX_THREADS=1` locally, then visiting an admin page that caused an ActionCable subscription and also loading a few other pages. I was **not** able to reproduce the database connection pool problem. I think that disproves my theory.

PR #2847, opened 2022-06-17
Add IRS language preference page

embarnard:


alaroia-cfa (commented):
One blocker and a handful of comments. Good code overall!

app/helpers/irs_languages.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def self.keys
+    hash.keys
+  end
+
+  def self.display_values(locale)
+    languages = hash.values
+    return languages unless locale == :es
+
```
embarnard, line 13: add comment, this is to add spanish as default

app/views/ctc/questions/irs_language_preference/edit.html.erb
```diff
@@ -0,0 +1,11 @@
+<% content_for :page_title, t("views.ctc.questions.irs_language_preference.title") %>
+
+<% content_for :form_card do %>
+  <h1 class="h2"><%= content_for(:page_title) %></h1>
+  <p><%= t("views.ctc.questions.irs_language_preference.subtitle") %></p>
+
+  <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+    <%= f.cfa_select(:irs_language_preference, t("views.ctc.questions.irs_language_preference.select_language"), IrsLanguages.display_values(locale)) %>
```
embarnard, line 8: Add value in select_options to select default

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:intake) { create :ctc_intake, irs_language_preference: irs_language_preference }
+
+  describe "#edit" do
+    before do
+      sign_in intake.client
+    end
+
+    render_views
+
+    context "locale is es and irs_language_preference is nil" do
```
alaroia-cfa, line 14: I suggest creating a:

`context "without a saved language preference" do`

and moving these two tests into it.

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      it "Spanish is selected" do
```
alaroia-cfa, line 15: I suggest `it "selects Spanish by default"` so that it flows like an English sentence.

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        expect(response).to render_template :edit
+
+        html = Nokogiri::HTML.parse(response.body)
+        value = html.at_css('#ctc_irs_language_preference_form_irs_language_preference option[@selected="selected"]')['value']
+        expect(value).to eq "spanish"
+      end
+    end
+
+    context "locale is en and irs_language_preference is nil" do
+      it "English is selected" do
```
alaroia-cfa, line 26: It would be nice to make this flow like an English sentence, too.

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        get :edit, params: { locale: "en" }
+        expect(response).to render_template :edit
+
+        html = Nokogiri::HTML.parse(response.body)
+        value = html.at_css('#ctc_irs_language_preference_form_irs_language_preference option[@selected="selected"]')['value']
+        expect(value).to eq "english"
+      end
+    end
+
+    context "irs_language_preference is russian" do
```
alaroia-cfa, line 36: Maybe this could be:

```
context "with a saved language preference" do
```

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      let(:irs_language_preference) { "russian" }
+
+      it "Russian is selected" do
```
alaroia-cfa, line 39: It would be nice to make this flow like an English sentence, too.

alaroia-cfa (approved):
I approve. I'd prefer to see a handful of changes implemented per follow-up comments; you may merge with or without those changes. I strongly encourage doing them.

This code is now simple! Great work.

Happy to Tuple to implement/discuss the suggestions together.

spec/controllers/ctc/questions/irs_language_preference_controller_spec.rb
alaroia-cfa: Per Tuple conversation, it reads OK when you see the whole RSpec test in context, but we still ended up changing it.

bytheway875:
@alaroia-cfa @embarnard FYI, as implemented this will begin attaching the ScheduleLEP to EVERY return, once merged.

I think we need to update the logic for when this gets attached in the `supported_documents` method in submission_builder/ty2021/return1040.rb before merging this. (The original implementation assumed we would be asking people a second question of whether they wanted to change their preference, and would only save a value if they said yes).


Screen Shot 2022-06-22 at 11 19 38 AM

bytheway875 (changes requested):
I think we need to make sure to only attach this to returns where the preference either a) has changed from previous year (probably impossible without asking another intake question) or b) has a value other than english, instead of right now the logic is "any value at all". See other comment!

alaroia-cfa:
@bytheway875 Oh! I didn't realize it would have that effect. Good catch.

I'm seeing that there's a `.present?` check. @embarnard , what if we allowed "No preference" as an option in the list? I haven't gone and checked design's mocks to see if they have an opinion on how to handle this.

I'm in theory happy to pair but actually it's subject to my calendar. Maybe I'll call you Em to connect later, but feel free to do whatever is reasonable without me too.

PR #2848, opened 2022-06-21
Special eligibility rules for Puerto Rican dependents

bytheway875:


app/lib/efile/dependent_eligibility/qualifying_child.rb
```diff
@@ -53,6 +54,14 @@ def under_qualifying_age_limit?
         !over_qualifying_age_limit?
       end
 
+      def additional_puerto_rico_rules?
```
bytheway875, line 57: because the rules are ||-based, if we broke these into two methods it won't exclude if EITHER the bday is okay OR the ssn is valid for employment

PR #2852, opened 2022-06-22
Update dependent does not qualify page for Puerto Rico residents

caitbot:
image

### DIFF
--- a/app/views/ctc/questions/dependents/does_not_qualify_ctc/edit.html.erb
+++ b/app/views/ctc/questions/dependents/does_not_qualify_ctc/edit.html.erb
@@ -1,4 +1,4 @@
-<% @puerto_rico_filing = current_intake.puerto_rico_filing? %>
+<% puerto_rico_filing = current_intake.puerto_rico_filing? %>
 <% @main_question = @puerto_rico_filing ? t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.title", name:  @dependent.first_name.upcase_first) : t("views.ctc.questions.dependents.does_not_qualify_ctc.title", name:  @dependent.first_name.upcase_first) %>
 
 <% content_for :page_title, @main_question %>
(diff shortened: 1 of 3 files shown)


app/views/ctc/questions/dependents/does_not_qualify_ctc/edit.html.erb
```diff
@@ -1,43 +1,42 @@
-<% @main_question = t("views.ctc.questions.dependents.does_not_qualify_ctc.title", name:  @dependent.first_name.upcase_first) %>
+<% @puerto_rico_filing = current_intake.puerto_rico_filing? %>
```
alaroia-cfa, line 1: If we're setting the variable purely for this template, I'd rather it not be an `@` variable.

Another idea is to adjust the controller to set this variable.

I'd prefer either of those rather than this approach where the `@` variable is created in the `erb`.

Not a blocker.

app/views/ctc/questions/dependents/does_not_qualify_ctc/edit.html.erb
```diff
@@ -1,43 +1,42 @@
...
 <% content_for :form_card do %>
   <h1 class="h2"><%= @main_question %></h1>
 
-  <% if current_intake.puerto_rico_filing? %>
-    <p><%= t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.help_text", name:  @dependent.first_name.capitalize) %></p>
+    <p><%= @puerto_rico_filing ? t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.help_text", name:  @dependent.first_name.upcase_first) : t("views.ctc.questions.dependents.does_not_qualify_ctc.help_text", name:  @dependent.first_name.upcase_first) %></p>
     <%= render('components/molecules/reveal', title: t("views.ctc.questions.dependents.does_not_qualify_ctc.reveal")) do %>
       <ul class="list--bulleted">
-        <% t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.conditions", current_tax_year: current_tax_year, filing_year: current_tax_year + 1).each do |condition| %>
+        <% @conditions = @puerto_rico_filing ? t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.conditions", current_tax_year: current_tax_year, filing_year: current_tax_year + 1) : t("views.ctc.questions.dependents.does_not_qualify_ctc.conditions", current_tax_year: current_tax_year, filing_year: current_tax_year + 1) %>
```
alaroia-cfa, line 12: Same - seems like no `@` would be better, or set it in the controller perhaps.

alaroia-cfa (approved):
This is good, and simple. I left some suggestions; no blockers.

app/views/ctc/questions/dependents/does_not_qualify_ctc/edit.html.erb
```diff
@@ -1,43 +1,42 @@
-<% @main_question = t("views.ctc.questions.dependents.does_not_qualify_ctc.title", name:  @dependent.first_name.upcase_first) %>
+<% puerto_rico_filing = current_intake.puerto_rico_filing? %>
+<% @main_question = @puerto_rico_filing ? t("views.ctc.questions.dependents.does_not_qualify_ctc.pr.title", name:  @dependent.first_name.upcase_first) : t("views.ctc.questions.dependents.does_not_qualify_ctc.title", name:  @dependent.first_name.upcase_first) %>
```
alaroia-cfa, line 2: Same thing here - probably doesn't need `@` and not a blocker and feel free to merge with/without this change.

PR #2855, opened 2022-06-22
Puerto Rico landing page defaults to es locale

caitbot:


app/controllers/ctc/ctc_pages_controller.rb
```diff
@@ -74,7 +74,13 @@ def how_do_i_get_an_itin; end
     def completed; end
 
     def puerto_rico
-      redirect_to action: :home unless Flipper.enabled?(:puerto_rico_home_location)
+      redirect_to action: :home and return unless Flipper.enabled?(:puerto_rico_home_location)
+
+      if params[:locale].nil?
+        I18n.with_locale('es') do
+          render :puerto_rico and return
```
alaroia-cfa, line 81: Nice simple `if`!

For this line, I'd prefer `redirect_to` to change the URL to contain `/es` . I guess it doesn't matter though! So it's not a blocker.

alaroia-cfa (approved):
Works for me. I'm honestly surprised that visiting our landing pages e.g. demo.getyourrefund.org/volunteers doesn't redirect to demo.getyourrefund.org/en/volunteers . Anyway, I have a preference, but your approach also works. No blockers.

app/controllers/ctc/ctc_pages_controller.rb
caitbot: I like that actually, its a bit cleaner. I was avoiding doing a redirect but I really have no good reason for that.

app/controllers/ctc/ctc_pages_controller.rb
```diff
@@ -74,7 +74,9 @@ def how_do_i_get_an_itin; end
     def completed; end
 
     def puerto_rico
-      redirect_to action: :home unless Flipper.enabled?(:puerto_rico_home_location)
+      redirect_to action: :home and return unless Flipper.enabled?(:puerto_rico_home_location)
+
+      redirect_to "/es/puertorico" and return unless params[:locale].present?
```
bytheway875, line 79: can you just set params[:locale] to :es unless params[:locale] is present? instead of doing a redirect? Idk if that would work given our setup, but that's what i saw recommended on stack overflow

app/controllers/ctc/ctc_pages_controller.rb
bytheway875: oh i see... you actually changed it from this.

app/controllers/ctc/ctc_pages_controller.rb
caitbot: that was actually the first thing that I tried, but unfortunately, the way this is called the `params[:locale]` is used in the `switch_locale` method in the `ApplicationController`, that method is an `around_action` which allows you to `yield` to a block, the block that it `yields` to is the `CtcPagesController#puerto_rico` block. So you really need to set the `params[:locale]` BEFORE youre in this block, which means doing a redirect (or forcing it to render with a locale, thereby overwriting whatever the `switch_locale` action did, I changed it to a redirect since that seems more straightforward)

PR #2857, opened 2022-06-23
PR copy changes and add reveals in flow

embarnard:


embarnard:
This PR adds copy changes in the GetCTC mostly for Puerto Rico client but also some copy changes for non-pr clients as well as adding a few reveals.
Overview page:
overview

income page for single non-pr client
income single

income page for mfj non-pr client
income mfj

income page for single pr client
income pr single

income page for mfj pr client
income mfj pr

file-full-return page
file-full-return

use-gyr page
use-gyr

alaroia-cfa (approved):
Approved plus we paired.

PR #2858, opened 2022-06-23
Fix overly ambitious offboarding bug

bytheway875:
We were immediately offboarding any PR applicant with a child that did not qualify, instead of only when ALL children do not qualify and they're done adding.


alaroia-cfa (approved):
Looks good! We also resolved a merge conflict that came up from some work we just did

PR #2865, opened 2022-06-24
Change order of reveals on filing status page

bytheway875:
The reveal order should be swapped, per ticket.


embarnard (approved):
Looks good!

PR #2867, opened 2022-06-24
Ip pin validation 182551982

bytheway875:
limit the length of these inputs to 5 or 6 digits so that they cant be too long.


alaroia-cfa (approved):
This is a great change. Nice to see the new tests covering existing code, and very nice to see `maxlength` in that input box. It makes me wonder if there's a way for that info to bubble down automatically into the input field, but also, forget I said that, because I have no good ideas how to do that.

PR #2874, opened 2022-06-27
Use Gyr page redirects to start of flow if no current_intake

wschaefer:
We need it now to determine whether to show puerto rico specific content


alaroia-cfa (approved):
Makes me wonder if we should use the anonymous intake page concern. Not a blocker; feel free to use that or not, and merge.

PR #2884, opened 2022-06-27
Add a retry for USPS address. Try up to 3 times

bytheway875:


app/models/efile_submission.rb
```diff
@@ -119,10 +119,15 @@ def previously_transmitted_submission
     previous_submission if previous_submission.last_transition_to("transmitted").present?
   end
 
-  def generate_verified_address
+  def generate_verified_address(i = 0)
     return OpenStruct.new(valid?: true) if verified_address.present?
 
-    address_service = StandardizeAddressService.new(intake)
+    address_service = StandardizeAddressService.new(intake, read_timeout: 1500)
```
bytheway875, line 125: a timeout of 1.5 seconds is totally arbitrary. We set a one second timeout in the application call for this and it seemed ok, so decided to go for a bit longer here.

app/services/standardize_address_service.rb
```diff
@@ -4,6 +4,9 @@
 # PLEASE do NOT use this service to standardize addresses in bulk (do NOT use in backfills)
 # as that violates the terms of agreement with USPS and will result in access being revoked.
 class StandardizeAddressService
+  attr_accessor :timeout
+  alias_method :timeout?, :timeout
```
jenny-heath, line 8: ohh never seen this: is this so that you can call `timeout` and `timeout?`?

jenny-heath (approved):
this looks good! seems to cover the case where the API times out (not sure whether we want the story to cover other types of failures? if so that could go in a separate PR though)

app/services/standardize_address_service.rb
bytheway875: yep!

PR #2895, opened 2022-06-28
More button width changes

bytheway875:


alaroia-cfa (approved):
width or widthout

[just having fun; here, that's a synonym for 'I approve']

PR #2897, opened 2022-06-28
Use GetCTC and GetYourRefund logos in emails

bytheway875:
And also change some of the styling per request from design.

Screen Shot 2022-06-28 at 12 32 26 PM

Unfortunately, I couldn't figure out a way to get the logo to set using the different service types in a before action... i had to set it individually in each mailer.

I tried `around_action` and `after_action`, but after_action only invokes after the template is already generated (which means that it cant grab the attachment), and around_action doesn't have access to the instance variables (like before_action) if I try to set before `block.call` and has the same problem as after_action if I put it after.

I could refactor to use a regular image instead of an attachment OR switch to passing in data to the mailer in a different way (I think), but since we don't use that many different mailers and we also have a couple other things that need to be explicitly set up in the other ones too, I thought this would be fine.


bytheway875:
@tgrathwell i think there was some other work we were planning on doing with wanting to make sure the domain settings got set in each thing. maybe ill make a follow up ticket for it.

PR #2899, opened 2022-06-30
Change the routing method for clients who were at capacity but are now getting assigned

bytheway875:
This will prevent a bug in which clients in the intake flow will go through another routing attempt if they were previously marked as an at-capacity routing.

The alternative here would be to update the intake flow so that it only attempts to reroute if the organization is nil AND the routing method was at capacity, but to me changing routing method seemed to be the way to ensure we don't have to maintain multiple pieces of data in our mind.

### DIFF
--- a/app/controllers/questions/at_capacity_controller.rb
+++ b/app/controllers/questions/at_capacity_controller.rb
@@ -8,7 +8,9 @@ module Questions
     end
 
     def edit
-      redirect_to next_path unless current_client.routing_method_at_capacity?
+      if current_client.present? && current_client.routing_method != "at_capacity"
+        redirect_to next_path
+      end
 
       current_intake.update(viewed_at_capacity: true)
     end
--- a/app/controllers/questions/at_capacity_controller.rb
+++ b/app/controllers/questions/at_capacity_controller.rb
@@ -8,6 +8,8 @@ module Questions
     end
 
     def edit
+      redirect_to next_path unless current_client.routing_method_at_capacity?
+
       current_intake.update(viewed_at_capacity: true)
     end
   end
--- a/app/controllers/questions/consent_controller.rb
+++ b/app/controllers/questions/consent_controller.rb
@@ -35,12 +35,7 @@ module Questions
 
       # the vita partner the client was routed to has capacity
       unless current_intake.client.routing_method_at_capacity?
-        tax_returns = []
-        TaxReturn.filing_years.each do |year|
-          tax_returns.push(TaxReturn.find_or_initialize_by(year: year, client: current_intake.client)) if current_intake.send("needs_help_#{year}") == "yes"
-        end
-        current_intake.client.tax_returns.replace(tax_returns)
-        tax_returns.map { |tr| tr.advance_to(:intake_in_progress) }
+        CreateInitialTaxReturnsService.new(intake: current_intake).create!
         GenerateF13614cPdfJob.perform_later(current_intake.id, "Preliminary 13614-C.pdf")
       end
 
@@ -50,7 +45,9 @@ module Questions
         locale: I18n.locale
       )
 
-      unless current_intake.client.routing_method_at_capacity?
+      if current_intake.client.routing_method_at_capacity?
+        session[:intake_id] = nil
+      else
         sign_in current_intake.client
       end
     end
--- a/app/controllers/questions/consent_controller.rb
+++ b/app/controllers/questions/consent_controller.rb
@@ -45,11 +45,7 @@ module Questions
         locale: I18n.locale
       )
 
-      if current_intake.client.routing_method_at_capacity?
-        session[:intake_id] = nil
-      else
-        sign_in current_intake.client
-      end
+      sign_in current_intake.client unless current_intake.client.routing_method_at_capacity?
     end
   end
 end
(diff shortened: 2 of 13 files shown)


wschaefer (approved):
This looks good! 

Asheesh and I went down a rabbit hole of how this code all works since its new to me. In that process we realized that the places we check `routing_method_at_capacity?` can instead check for `vita_partner.nil?`. This works even on the `AtCapacityController` because if you're seeing it, you saw the consent screen where routing happens. That could be another option. This would be a bigger change, but has the advantage of not maintaining the same information twice and having to keep it in sync. One hypothetical is an engineer changing `vita_partner` in `rails c` (and maybe there's others actually in the code).

spec/services/update_client_vita_partner_service_spec.rb
```diff
@@ -14,6 +14,19 @@
     let(:tax_return) { create :tax_return, year: 2019, assigned_user: assigned_user }
     let(:assigned_user) { create :team_member_user, site: current_site }
 
+    context "when a client was previously routed to no one because we were at capacity" do
```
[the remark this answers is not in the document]
alaroia-cfa: I added this assertion to a test below.

bytheway875:
@alaroia-cfa thanks for making these changes!

app/controllers/questions/at_capacity_controller.rb
```diff
@@ -8,6 +8,8 @@ def self.show?(intake)
     end
 
     def edit
+      redirect_to next_path unless current_client.routing_method_at_capacity?
```
bytheway875, line 11: @alaroia-cfa Even though the .show? method says "don't show unless its at capacity", it seems that the show method actually only gets used in ControllerNavigation. Which means that when there's a direct link to the page, the show logic never runs. 

This is how it has always been BUT I think I didn't really internalize that in the original implemention / maybe wasn't completely aware that that's how it worked.

tgrathwell (approved):

tgrathwell:
@alaroia-cfa and I reviewed this and it looks good

needs a rebase on main so it could pass the tests

app/controllers/questions/consent_controller.rb
```diff
@@ -35,20 +35,20 @@ def after_update_success
 
       # the vita partner the client was routed to has capacity
       unless current_intake.client.routing_method_at_capacity?
-        tax_returns = []
-        TaxReturn.filing_years.each do |year|
-          tax_returns.push(TaxReturn.find_or_initialize_by(year: year, client: current_intake.client)) if current_intake.send("needs_help_#{year}") == "yes"
-        end
-        current_intake.client.tax_returns.replace(tax_returns)
-        tax_returns.map { |tr| tr.advance_to(:intake_in_progress) }
+        CreateInitialTaxReturnsService.new(intake: current_intake).create!
```
bytheway875, line 38: nit: `InitialTaxReturnsService` would be fine too since you're calling create! on it.

app/services/update_client_vita_partner_service.rb
```diff
@@ -1,14 +1,21 @@
 class UpdateClientVitaPartnerService < BaseService
-  def initialize(clients:, vita_partner_id:, change_initiated_by: nil)
+  def initialize(clients:, vita_partner_id:, change_initiated_by: nil, first_time: false)
     @clients = clients
     @new_vita_partner = VitaPartner.find_by_id(vita_partner_id)
     @change_initiated_by = change_initiated_by
+    @first_time = first_time
```
bytheway875, line 6: Are we using first_time anywhere in the code at the moment?

bytheway875 (commented):
A couple things to consider, and also a question about @first_time

app/services/update_client_vita_partner_service.rb
alaroia-cfa: Nope. I forgot to clean up a half-implemented idea. It's gone now. :)

app/services/create_initial_tax_returns_service.rb
```diff
@@ -0,0 +1,11 @@
+class CreateInitialTaxReturnsService < BaseService
+  def initialize(intake:)
+    @intake = intake
+  end
+
+  def create!
+    preexisting_tax_years = @intake.client.tax_returns.pluck(:year)
+    create_years = TaxReturn.filing_years.filter { |year| !preexisting_tax_years.include?(year) && @intake.send("needs_help_#{year}") == "yes" }
```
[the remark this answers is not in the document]
alaroia-cfa: Your idea of marking a tax return as created by hub vs. created directly by client is an interesting one. I'm in favor of not implementing that yet, but I hadn't considered it yet, and I always like new ideas.

I'm in favor of the code in the pull request - every once in a while we got godforesaken Sentry errors from someone re-doing consent somehow, and in the process trying to delete a tax return presumably by accident, and then somehow deletion failing to cascade or triggering a foreign key problem in Postgres (fuzzy memory on my part), and then me thinking "yikes" about that `.replace` :D.

Plus Yvonne and I discussed the use case of "A Hub user creates a tax year for the client, and then assigns them," and we landed in favor of retaining the new tax year and letting VITA mark it as not filing.

app/controllers/questions/consent_controller.rb
alaroia-cfa: This is good because the service name is clearly too long.

alaroia-cfa:
There's a test left to fix, and a service rename to do; I expect to do those tomorrow.

bytheway875 (commented):
approved!

PR #2904, opened 2022-07-01
Set consented_to_service_at for clients created in /flows

wschaefer:
This allows them to be visible in the hub

- Add a convenience link to the client's hub page from /flows


alaroia-cfa:
Looks good to me!

PR #2906, opened 2022-07-05
[#179239944] Handle invalid cross origin JS request errors

alaroia-cfa:
I believe this will result in fewer pages and is harmless.

See https://cfa.slack.com/archives/G013YJMPVAA/p1657056544215089

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -402,6 +402,15 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  # Rails's CSRF countermeasures trigger two exceptions. Mostly these are triggered by security scans.
+  # Catch them to avoid being noisy in logs etc.
+  rescue_from 'ActionController::InvalidCrossOriginRequest' do
+    DatadogApi.increment("rails.invalid_cross_origin_request")
+    respond_to do |format|
+      format.any { head 422 }
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -402,6 +402,15 @@ def wrapping_layout
     end
   end
 
+  # Rails's CSRF countermeasures trigger two exceptions. Mostly these are triggered by security scans.
+  # Catch them to avoid being noisy in logs etc.
+  rescue_from 'ActionController::InvalidCrossOriginRequest' do
+    DatadogApi.increment("rails.invalid_cross_origin_request")
+    respond_to do |format|
+      format.any { head 422 }
```
bytheway875, line 410: How do we handle a 422? Does it show the same 400 error page or something different?

bytheway875 (approved):

app/controllers/application_controller.rb
alaroia-cfa: Since it's a `head 422` it responds with an empty body (rather than the visual 400 page). I figure that's best since the requested content-type is not HTML (JS in this case), plus also it's only ever triggered by bots as far as I can tell.

bytheway875: right right -- that makes sense. Our 404 page is a little specific haha, so just thought it might be confusing if it ended up rendering here. glad that it wont!

PR #2907, opened 2022-07-11
Update mailing-address copy for CTC

caitbot:
image


bytheway875:
nice! this seems like a good improvement to the UX.

bytheway875 (approved):

PR #2923, opened 2022-07-13
Remove comma from urbanization in address [#182705742]

alaroia-cfa:
This is required for the address to conform to the IRS 1040 XML schema rules.

### DIFF
--- a/app/lib/irs1040_pdf.rb
+++ b/app/lib/irs1040_pdf.rb
@@ -15,15 +15,12 @@ class Irs1040Pdf
   end
 
   def hash_for_pdf
-    urbanization = @address.urbanization
-    urbanization += "," if urbanization.present?
-
     answers = {
       FilingStatus: @xml_document.at("IndividualReturnFilingStatusCd")&.text,
       PrimaryFirstNm: @intake.primary_middle_initial.present? ? "#{@intake.primary_first_name} #{@intake.primary_middle_initial}" : @intake.primary_first_name,
       PrimaryLastNm: @intake.primary_last_name,
       PrimarySSN: @xml_document.at("PrimarySSN")&.text,
-      AddressLine1Txt: [urbanization, @address.street_address, @address.street_address2].compact.join(" "),
+      AddressLine1Txt: [@address.urbanization, @address.street_address, @address.street_address2].compact.join(" "),
       CityNm: @address.city,
       StateAbbreviationCd: @address.state,
       ZipCd: @address.zip_code,
--- a/app/lib/submission_builder/shared/return_header1040.rb
+++ b/app/lib/submission_builder/shared/return_header1040.rb
@@ -73,7 +73,7 @@ module SubmissionBuilder
               xml.USAddress {
                 # IRS provides no information on how to shorten addresses
                 # but requires that they be < 36 characters long
-                xml.AddressLine1Txt trim([address.urbanization, address.street_address].map(&:presence).compact.join(", "), 35)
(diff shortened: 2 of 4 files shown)


spec/lib/submission_builder/shared/return_header1040_spec.rb
```diff
@@ -287,11 +287,11 @@
           response = described_class.build(submission)
           expect(response).to be_an_instance_of SubmissionBuilder::Response
           xml = Nokogiri::XML::Document.parse(response.document.to_xml)
-          expect(xml.at("AddressLine1Txt").text).to eq "URB PICARD, 23627 HAWKINS CREEK CT"
+          expect(xml.at("AddressLine1Txt").text).to eq "URB PICARD 23627 HAWKINS CREEK CT"
           expect(xml.at("CityNm").text).to eq "SAN JUAN"
           expect(xml.at("StateAbbreviationCd").text).to eq "PR"
           expect(xml.at("ZIPCd").text).to eq "00797"
-          expect(response).not_to be_valid
+          expect(response).to be_valid
```
alaroia-cfa, line 294: 💀

app/lib/irs1040_pdf.rb
```diff
@@ -15,15 +15,12 @@ def initialize(submission)
...
-    urbanization = @address.urbanization
-    urbanization += "," if urbanization.present?
-
     answers = {
       FilingStatus: @xml_document.at("IndividualReturnFilingStatusCd")&.text,
       PrimaryFirstNm: @intake.primary_middle_initial.present? ? "#{@intake.primary_first_name} #{@intake.primary_middle_initial}" : @intake.primary_first_name,
       PrimaryLastNm: @intake.primary_last_name,
       PrimarySSN: @xml_document.at("PrimarySSN")&.text,
-      AddressLine1Txt: [urbanization, @address.street_address, @address.street_address2].compact.join(" "),
+      AddressLine1Txt: [@address.urbanization, @address.street_address, @address.street_address2].compact.join(" "),
```
bytheway875, line 23: nice. im glad this pattern is mostly working out without getting too gnarly working out.

app/lib/submission_builder/shared/return_header1040.rb
```diff
@@ -73,7 +73,7 @@ def document
               xml.USAddress {
                 # IRS provides no information on how to shorten addresses
                 # but requires that they be < 36 characters long
-                xml.AddressLine1Txt trim([address.urbanization, address.street_address].map(&:presence).compact.join(", "), 35)
+                xml.AddressLine1Txt trim([address.urbanization, address.street_address].map(&:presence).compact.join(" "), 35)
```
bytheway875, line 76: i feel unsure why we had to do a presence.compact here, but i remember testing it in the console and realizing it was in fact necessary

app/lib/irs1040_pdf.rb
alaroia-cfa: I do wonder if at some point soon we should change it so the `Address` class computes this. I know the XML truncates at 35, and the PDF doesn't; it would be good to make those behave the same anyway.

alaroia-cfa: I don't really want to do that in this pull request, though!

bytheway875 (approved):

app/lib/submission_builder/shared/return_header1040.rb
alaroia-cfa: `""` vs `nil`

Travis and I worked somewhat hard to avoid storing a `""` in the database for `.urbanization`, trying to instead either store `nil` or an actual value, but we might have gotten it wrong, and no point in extra space characters showing up.

PR #2934, opened 2022-07-19
Eitc first page

jenny-heath:


alaroia-cfa (approved):
yay

alaroia-cfa:
Would love to see the Puerto Rico feature flag go away! https://cfa.slack.com/archives/GT1B45BS5/p1658346922118099

caitbot (approved):

app/controllers/ctc/questions/claim_eitc_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Ctc
+  module Questions
+    class ClaimEitcController < QuestionsController
+      include Ctc::ResetToStartIfIntakeNotPersistedConcern
+
+      layout "yes_no_question"
+
+      def self.show?(intake)
+        Flipper.enabled?(:eitc) && !intake.home_location_puerto_rico?
```
bytheway875, line 9: i think technically it might be better if we used intake.puerto_rico_filing? here, since that also includes the feature flag and could potentially include other things that make something a puerto rico filing in the future... but also maybe not. im not sure which is better. my preference would be to change it to intake.puerto_rico_filing? though.

app/views/ctc/questions/claim_eitc/edit.html.erb
```diff
@@ -0,0 +1,7 @@
+<% content_for :form_question, t("views.ctc.questions.claim_eitc.title") %>
+
+<% content_for :form_help_text do %>
+  <p><%= t("views.ctc.questions.claim_eitc.p1") %></p>
```
bytheway875, line 4: i prefer to see multiple paragraphs of text in the yml as an array, but i dont know if we have specific guidance over whether it must be one way or the other to merge.

spec/controllers/ctc/ctc_pages_controller_spec.rb
```diff
@@ -96,7 +96,7 @@
   describe "#puerto_rico" do
     context "when the puerto rico features has launched" do
       before do
-        allow(Flipper).to receive(:enabled?).with(:puerto_rico_home_location).and_return true
+        Flipper.enable :puerto_rico_home_location
```
bytheway875, line 99: is there an explicit reason or thought process on enabling directly vs stubbing the enabling? Does Flipper docs recommend one way over the other?

spec/features/ctc/puerto_rico_spec.rb
```diff
@@ -26,7 +22,7 @@
 
   context "when the puerto rico intake has launched" do
     before do
-      allow(Flipper).to receive(:enabled?).and_return true
+      Flipper.enable :puerto_rico_home_location
```
bytheway875, line 25: Do you have to explicitly un-enable the feature after the test completes? I'm wondering if this leaves the Flipper context where the feature is enabled as the rest of the tests run...

spec/controllers/ctc/ctc_pages_controller_spec.rb
tgrathwell: `allow .. to receive .. and_return` will turn on/off *every* feature flag currently in effect, which can cause bad results

`allow .. to receive .. with .. and_return` doesn't scale well with multiple features to flip, because it will complain if `enabled?` is called with different arguments in a different order -- "expected to be called with puerto_rico but was called with eitc"

so we could beef up the stubbing even further to compensate for that, but all we're really looking for is an in-memory list of features that are flipped on that gets reset every test, and Flipper provides that with `Flipper.instance = Flipper.new(Flipper::Adapters::Memory.new)`

tgrathwell:
@bytheway875 https://www.flippercloud.io/docs/testing

> For CI/test environments, we recommend the memory adapter. Switching the default instance of Flipper to use the memory adapter is easy. Before each test, simply configure Flipper to use the memory adapter like so...

PR #2945, opened 2022-07-21
Some changes to facilitate latest requirements of signup messaging

bytheway875:
- add an "after" argument to the signups so you can provide a date
- add the message class

### DIFF
--- a/app/models/signup.rb
+++ b/app/models/signup.rb
@@ -2,14 +2,15 @@
 #
 # Table name: signups
 #
-#  id                            :bigint           not null, primary key
-#  ctc_2022_open_message_sent_at :datetime
-#  email_address                 :citext
-#  name                          :string
-#  phone_number                  :string
-#  zip_code                      :string
-#  created_at                    :datetime         not null
-#  updated_at                    :datetime         not null
+#  id                               :bigint           not null, primary key
+#  ctc_2022_open_message_sent_at    :datetime
+#  email_address                    :citext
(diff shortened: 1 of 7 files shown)


app/models/signup.rb
```diff
@@ -19,12 +20,14 @@ class Signup < ApplicationRecord
...
-  def self.send_message(message_name, batch_size=nil)
+  def self.send_message(message_name, batch_size=nil, after: nil)
     message_class = "AutomatedMessage::#{message_name.camelize}".constantize
     message = message_class.new
     sent_at_column = "#{message_name}_sent_at"
 
-    Signup.where("#{message_name}_sent_at" => nil).order(created_at: :asc).limit(batch_size).find_each do |signup|
+    signups = Signup.where("#{message_name}_sent_at" => nil)
+    signups = Signup.where('created_at >= ?', after) if after.present?
+    signups.find_each do |signup|
```
caitbot, line 30: do we want to keep the `order.(created_at: :asc)` here to ensure that we send them sequentially (should it fail in the middle and we need to enter a new `after`)?

caitbot (approved):
This looks good, left a comment about keeping the order of loading but its ultimately your call.

bytheway875:
as we chatted about on slack, i had been getting a warning that order was being ignored either because of something else about the query (either limit or find_each). Since its not a super big deal and wasn't being applied, I just removed it

PR #2947, opened 2022-07-21
Remove usages of puerto_rico_home_location feature flag

tgrathwell:


spec/features/ctc/eitc_spec.rb
```diff
@@ -6,7 +6,6 @@
   before do
     allow_any_instance_of(Routes::CtcDomain).to receive(:matches?).and_return(true)
     Flipper.enable :eitc
-    Flipper.enable :puerto_rico_home_location
```
bytheway875, line 9: Does calling Flipper.enable like this keep the feature flag turned on for later tests? I've been looking for documentation on whether this will automatically disable in a test context afterward...

tgrathwell: https://github.com/codeforamerica/vita-min/blob/main/spec/rails_helper.rb#L138 makes it so we start with a fresh (empty) set of flags every test

bytheway875: nice, good to know. thanks.

bytheway875 (approved):
🥳

PR #2952, opened 2022-07-21
Rails: use 7.0 defaults

tgrathwell:
we were still using the 6.1 defaults but hopefully all the new defaults are fine? let's see


bytheway875:
@tgrathwell This looks good to me. Wondering if this is worth socializing in next CGC. Wondering if we'll run into situations where links could look like they work locally and in demo environment (where host name is the same) and then fail on production and we might not catch it. 😬

app/controllers/ctc/portal/submission_pdfs_controller.rb
```diff
@@ -29,6 +29,6 @@ def show
 
   def error_redirect
     flash[:alert] = "We encountered a problem generating your tax return pdf. For assistance, please reach out to GetCTC client support."
-    redirect_back(fallback_location: request.referer)
+    redirect_back(fallback_location: request.referer, allow_other_host: true)
```
alaroia-cfa, line 32: maybe remove? maybe tests need it?

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -1,117 +0,0 @@
...
-# Rails.application.config.action_view.apply_stylesheet_media_default = false
-
-# Change the digest class for the key generators to `OpenSSL::Digest::SHA256`.
-# Changing this default means invalidate all encrypted messages generated by
-# your application and, all the encrypted cookies. Only change this after you
-# rotated all the messages using the key rotator.
-#
-# See upgrading guide for more information on how to build a rotator.
-# https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html
-# Rails.application.config.active_support.key_generator_hash_digest_class = OpenSSL::Digest::SHA256
```
alaroia-cfa, line 26: let's do sha1 for now again and try harder to rotate session cookies sometime and as part of that sha256 'em

alaroia-cfa: `OpenSSL::Digest::SHA1`

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -1,117 +0,0 @@
...
-# Uncomment each configuration one by one to switch to the new default.
-# Once your application is ready to run with all new defaults, you can remove
-# this file and set the `config.load_defaults` to `7.0`.
-#
-# Read the Guide for Upgrading Ruby on Rails for more info on each option.
-# https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
-
-# `button_to` view helper will render `<button>` element, regardless of whether
-# or not the content is passed as the first argument or as a block.
-# Rails.application.config.action_view.button_to_generates_button_tag = true
```
alaroia-cfa, line 14: hmm `form__feelings-survey` and/or percy

PR #2960, opened 2022-07-26
Add n+1 query gem

bytheway875:
I played around with trying to raise and fix the problems pointed out in the tests, but because of the way a lot of our code is designed the gem will tell you oh this needs to be optimized, then when you add the eager load the next error says you SHOULDN'T eager load there.


tgrathwell:
I don't remember if we actually got anything of value out of `bullet` on GCF... it definitely gives a lot of false positives and you have to fight with its heuristics

I think there's probably more value in getting everybody well versed in Datadog to evaluate performance of pages after they've been deployed and are receiving real traffic

PR #2962, opened 2022-07-27
Flow explorer: capture spanish screenshots in a new tab

tgrathwell:
Previously, `:flow_explorer_screenshot` tests captured spanish
screenshots by clicking 'Español' in the header, taking a screenshot,
then clicking 'English' in the header

This meant we incurred ~3 pageloads for every screenshot we wanted
to take (initial english, spanish, then english again)

This change keeps a dedicated second tab open in the Chrome browser,
so that every time we want to capture the Spanish version of the page
we load the /es/... URL in that tab, capture the screenshot, and switch
back to the original tab

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,8 +16,9 @@ module ApplicationHelper
 
   def link_to_locale(locale, label, additional_attributes = {})
     link_to(label,
-            { :locale=>locale, :params=>request.query_parameters },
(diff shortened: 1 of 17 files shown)


alaroia-cfa (approved):
party time

alaroia-cfa:
We had an idea for a further refactor to remove the `no_screenshot` parameter but decided not to do it.

PR #2963, opened 2022-07-27
Create the EITC pdf and xml

caitbot:


app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -0,0 +1,43 @@
+class Irs1040ScheduleEicPdf
```
bytheway875, line 1: @caitbot Seems like we're using EITC and EIC interchangably which also is what the IRS does. Wondering if it makes sense to use one or the other consistently within the code base, or if reflecting what the IRS calls things at different places makes more sense.

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -0,0 +1,43 @@
...
+    @intake = submission.intake
+  end
+
+  def hash_for_pdf
+    answers = {
+      FullPrimaryName: @intake.primary_full_name,
+      PrimarySSN: @intake.primary_ssn
+    }
+    dependent_nodes = @xml_document.search("QualifyingChildInformation")
+    answers.merge!(dependents_info(dependent_nodes[0..]))
```
bytheway875, line 19: does calling with the range at the end transform the data somehow? Wondering what the distinction between `dependent_nodes` and `dependent_nodes[0..]`? Are they different data types?

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -0,0 +1,43 @@
...
+    answers
```
bytheway875, line 20: Since the return value of merge! is the entire new hash, I think that returning answers is redundant. But also, it is more clear...

Screen Shot 2022-07-28 at 11 55 12 AM

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  private
+
+  def dependents_info(dependent_nodes)
+    answers = {}
```
bytheway875, line 26: I'm not familiar with EITC as a benefit. Can you claim for 4 children maximum?

app/lib/submission_builder/ty2021/documents/schedule_eic.rb
```diff
@@ -0,0 +1,33 @@
...
+                  xml.PersonFirstNm person_name_type(dependent.first_name)
+                  xml.PersonLastNm person_name_type(dependent.last_name)
+                }
+                xml.IdentityProtectionPIN dependent.ip_pin
+                xml.QualifyingChildSSN dependent.ssn
+                xml.ChildBirthYr dependent.birth_date.year
+                xml.ChildIsAStudentUnder24Ind dependent.full_time_student_yes? && dependent.age_during_tax_year < 24
+                xml.ChildPermanentlyDisabledInd dependent.permanently_totally_disabled_yes?
+                xml.ChildRelationshipCd dependent.irs_relationship_enum
+                xml.MonthsChildLivedWithYouCnt '07'
```
bytheway875, line 25: interesting. i remember how we need to have this be at least 7. Is this a placeholder until we add a question that asks for this, or are we using 7 for everyone? I think either way this might be a good use case for a code comment.

app/lib/submission_builder/ty2021/return1040.rb
```diff
@@ -46,16 +46,17 @@ def supported_documents
             pdf: Irs1040ScheduleLepPdf,
             include: @submission.intake.irs_language_preference.present? && @submission.intake.irs_language_preference != "english"
           },
-        ]
-
-        if ENV['W2_SUPPORT'] == 'true'
```
bytheway875, line 51: interesting. are we removing this? was that just for a test?

app/models/efile_submission_dependent.rb
```diff
@@ -23,7 +23,7 @@ class EfileSubmissionDependent < ApplicationRecord
   belongs_to :efile_submission
   has_one :intake, through: :efile_submission
 
-  delegate :first_name, :last_name, :full_name, :ssn, :irs_relationship_enum, :birth_date, :tin_type_ssn?, :tin_type_atin?, :ip_pin, to: :dependent
+  delegate :first_name, :last_name, :full_name, :ssn, :irs_relationship_enum, :birth_date, :tin_type_ssn?, :tin_type_atin?, :ip_pin, :full_time_student_yes?, :permanently_totally_disabled_yes?, to: :dependent
```
bytheway875, line 26: i had to add something to this list the other day too. maybe we should just delegate all undefined methods to dependent 🤔

app/models/intake/ctc_intake.rb
```diff
@@ -440,4 +440,8 @@ def spouse_prior_year_agi_amount_computed
   def puerto_rico_filing?
     home_location_puerto_rico?
   end
+
+  def claiming_eitc?
```
bytheway875, line 444: YAY. I like this pattern. Can you let me know once this is merged so that Maria and I can change the code we wrote in intake to use this method?

spec/lib/irs1040_schedule_eic_pdf_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+             "ChildBirthYr1[3]" => nil,
+             "ChildBirthYr2[0]" => nil,
+             "ChildBirthYr2[1]" => nil,
+             "ChildBirthYr2[2]" => nil,
+             "ChildBirthYr2[3]" => nil,
+             "ChildBirthYr3[0]" => nil,
+             "ChildBirthYr3[1]" => nil,
+             "ChildBirthYr3[2]" => nil,
+             "ChildBirthYr3[3]" => nil,
+             "ChildFirstAndLastName1" => nil,
```
bytheway875, line 70: interesting. im surprised theres not a ChildFirstAndLastName0 ...

bytheway875: or maybe im confused as to what the output onto the PDF looks like. Probably not that important...

bytheway875 (approved):
A couple comments but nothing really blocking. Maybe one place we can add a comment for future devs. Looks awesome.

app/lib/irs1040_schedule_eic_pdf.rb
caitbot: Oy you know in the course of writing this I definitely confused myself a couple times. 

The problem is that the IRS calls it a `Schedule EIC` but we refer to our product as `EITC`. I'm not sure how often they call it that... I'm almost in favor of continuing the inconsistency, as it is at least... accurate in its reflections of the different terms. But open for debate, certainly.

app/lib/irs1040_schedule_eic_pdf.rb
caitbot: Ahh thats a good catch, this is a vestigial bit of what I copied from our additional dependents pdf. Will remove...

app/lib/irs1040_schedule_eic_pdf.rb
caitbot: Mer, I vote to remove it, then people can learn that same bit of info.

app/lib/irs1040_schedule_eic_pdf.rb
caitbot: So once you claim more than 3 dependents, it doesn't change the amount of credit that you get. So they only ever want you to tell them about three dependents max, cause they don't care about the rest of them.

app/lib/submission_builder/ty2021/documents/schedule_eic.rb
caitbot: I'll add a comment! It is just a placeholder!

app/models/efile_submission_dependent.rb
caitbot: It wouldn't be the worst idea - I was a little torn here as well.. it is getting to be a long list.

app/models/intake/ctc_intake.rb
caitbot: Definitely! Yes, I actually removed the other `W2_support` env variable above in favor of using this pattern.

spec/lib/irs1040_schedule_eic_pdf_spec.rb
caitbot: So we named the fields ourselves, and chose to index from 1, because it matched the pdf appearance?

app/lib/submission_builder/ty2021/return1040.rb
caitbot: I actually just changed it over to using our feature flag for eitc since they are essentially the same thing. So now it follows the pattern used by the other documents and does `include: @submission.intake.claiming_eitc?`

app/lib/irs1040_schedule_eic_pdf.rb
caitbot: I left a comment calling out the inconsistency, I want to leave it as is, I'm mildly concerned that if we change it to be internally consistent, depending on where someone is coming from, they will search for that string value and be a bit lost. The comment should help resolve that by increasing findability. Thats the hope at least.

PR #2968, opened 2022-07-28
Remove mini_racer

tgrathwell:
I think the modern solution is to just expect every env to have some version of `node` installed

let's see if we can swing that


alaroia-cfa:
Good harmless fun. Seems we only use ExecJS via autoprefixer-rails which definitely supports using node instead.

PR #2974, opened 2022-08-02
start to convert attr_encrypted to encrypt

bytheway875:


app/models/user.rb
```diff
@@ -47,7 +45,7 @@
 #
 class User < ApplicationRecord
   include PgSearch::Model
-
+  self.ignored_columns = ["encrypted_access_token", "encrypted_access_token_iv"]
```
bytheway875, line 48: This column is not used, im planning on running a migration to remove it in an upcoming commit.

app/models/bank_account.rb
```diff
@@ -27,12 +29,23 @@
 class BankAccount < ApplicationRecord
   belongs_to :intake
   has_one :client, through: :intake
-  attr_encrypted :bank_name, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
-  attr_encrypted :account_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
-  # Enum values are acceptable BankAccountType values to be sent to the IRS (See efileTypes.xsd)
```
tgrathwell, line 32: maybe this comment should stay? I know I was surprised when I dove into the XML and checking/saving is 1/0 instead of a string but maybe this comment isn't the best way of recording that

lib/tasks/backfill_bank_data.rake
```diff
@@ -0,0 +1,13 @@
+namespace :bank_account do
+  desc "Backfill account number to new encrypted column"
+  task backfill: :environment do
+    BankAccount.find_each do |ba|
```
tgrathwell, line 4: maybe make this `BankAccount.where(account_number: nil).or.where(bank_name: nil)` so it runs a little quicker if you have to run it more than once

app/models/user.rb
tgrathwell, line 48: may want to record in a commit message that this is being removed because it's an ancient zendesk thing

tgrathwell:
certainly seems like this could work and is a lot simpler than the way they do it in that blog post. neat!

app/models/bank_account.rb
bytheway875: oh! yeah, this removal of the comment was accidental.

lib/tasks/backfill_bank_data.rake
bytheway875: good call

PR #2980, opened 2022-08-04
Create data migration rake task for encrypted dependents attrs

bytheway875:


lib/tasks/backfill_encrypted_dependent_data.rake
```diff
@@ -0,0 +1,13 @@
+namespace :dependents do
+  desc "Backfill account number to new encrypted column"
```
mquadri23, line 2: sorry, could you remind me what the **account number** in this context was again?

mquadri23 (approved):
Looks good to me!

lib/tasks/backfill_encrypted_dependent_data.rake
bytheway875: ah, the description is not updated. good catch.

mquadri23 (approved):
Updated description great!

PR #2982, opened 2022-08-04
Remove temporary code for dependents attr_encrypted (Step 2)

bytheway875:


app/models/dependent.rb
```diff
@@ -57,12 +53,9 @@
 
 class Dependent < ApplicationRecord
   include SoftDeletable
-
+  self.ignored_columns = [:encrypted_ssn, :encrypted_ssn_iv, :encrypted_ip_pin, :encrypted_ip_pin_iv]
```
mquadri23, line 56: in the notion document it says "self.ignore" in the present tense but here it says "self.ignored". I'm assuming the latter is correct and the notion doc can be updated? Just want to make sure this ignoring works as intended.

bytheway875: yeah, its a typo in the notion document.

mquadri23 (approved):

PR #2984, opened 2022-08-04
Intake encrypted fields (Part 1)

bytheway875:


app/models/intake/ctc_intake.rb
```diff
@@ -287,13 +298,6 @@ class Intake::CtcIntake < Intake
   attribute :primary_prior_year_agi_amount, :money
   attribute :spouse_prior_year_agi_amount, :money
 
-  attr_encrypted :primary_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
```
bytheway875, line 290: i moved these onto the base intake because it makes the data migration simpler. will be removed entirely soon.

db/migrate/20220804193922_add_new_fields_for_encrypted_intake_attributes_to_intake.rb
```diff
@@ -0,0 +1,15 @@
+class AddNewFieldsForEncryptedIntakeAttributesToIntake < ActiveRecord::Migration[7.0]
+  def change
+    add_column :intakes, :primary_last_four_ssn, :text
+    add_column :intakes, :spouse_last_four_ssn, :text
+    add_column :intakes, :primary_ssn, :text
+    add_column :intakes, :spouse_ssn, :text
+    add_column :intakes, :bank_account_number, :text
+    add_column :intakes, :bank_name, :string
```
bytheway875, line 8: string for the ones that will no longer be encrypted, text for the rails encrypts fields

db/schema.rb
```diff
@@ -98,8 +98,8 @@
   end
 
   create_table "admin_roles", force: :cascade do |t|
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at", precision: nil, null: false
```
caitbot, line 101: We should probably remove this? Seems unrelated

db/schema.rb
```diff
@@ -1678,7 +1689,7 @@
            SELECT DISTINCT tax_returns.client_id
              FROM (tax_returns
                JOIN intakes ON ((intakes.client_id = tax_returns.client_id)))
-            WHERE ((tax_returns.current_state)::text <> ALL (ARRAY[('intake_before_consent'::character varying)::text, ('intake_in_progress'::character varying)::text, ('intake_greeter_info_requested'::character varying)::text, ('intake_needs_doc_help'::character varying)::text, ('file_mailed'::character varying)::text, ('file_accepted'::character varying)::text, ('file_not_filing'::character varying)::text, ('file_hold'::character varying)::text, ('file_fraud_hold'::character varying)::text]))
```
caitbot, line 1681: and remove this

app/models/intake.rb
```diff
@@ -333,13 +344,19 @@ def self.searchable_fields
...
+  attr_encrypted :attr_encrypted_bank_routing_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_bank_routing_number'
+  attr_encrypted :attr_encrypted_bank_account_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_bank_account_number'
+  attr_encrypted :attr_encrypted_primary_ip_pin, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_primary_ip_pin'
+  attr_encrypted :attr_encrypted_spouse_ip_pin, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_spouse_ip_pin'
+  attr_encrypted :attr_encrypted_primary_signature_pin, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_primary_signature_pin'
+  attr_encrypted :attr_encrypted_spouse_signature_pin, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_spouse_signature_pin'
+
+  encrypts :primary_last_four_ssn, :spouse_last_four_ssn, :primary_ssn, :spouse_ssn, :bank_account_number, :primary_ip_pin, :primary_signature_pin, :spouse_signature_pin, :spouse_ip_pin
```
mquadri23, line 359: do we also need to include **:bank_routing_number** and **:bank_name** with the rest of these ?

caitbot (approved):
sweet this looks awesome! thank you for the write up (for those reading - find it [here](https://www.notion.so/cfa/How-to-convert-from-attr_encrypted-to-Rails-encrypts-01d7d6df16764e75a92aeea61d3eaf9a)

remove the unrelated schema changes + ship it

app/models/intake.rb
bytheway875: no, because routing number and bank name will no longer be encrypted (like we did on the bank account table). good question!

app/models/intake.rb
```diff
@@ -333,13 +344,19 @@ def self.searchable_fields
...
-  attr_encrypted :spouse_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
-  attr_encrypted :bank_name, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
-  attr_encrypted :bank_routing_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
-  attr_encrypted :bank_account_number, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }
+  attr_encrypted :attr_encrypted_primary_last_four_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_primary_last_four_ssn'
+  attr_encrypted :attr_encrypted_spouse_last_four_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_spouse_last_four_ssn'
+  attr_encrypted :attr_encrypted_primary_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_primary_ssn'
+  attr_encrypted :attr_encrypted_spouse_ssn, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_spouse_ssn'
+  attr_encrypted :attr_encrypted_bank_name, key: ->(_) { EnvironmentCredentials.dig(:db_encryption_key) }, attribute: 'encrypted_bank_name'
```
mquadri23, line 351: I'm confused with this line because I thought **bank name** is not an encrypted attribute

mquadri23 (approved):
I just left one comment on something I was confused about but from whatever I can understand it all looks good to me!

app/models/intake.rb
bytheway875: These attributes are the _existing_ attribtues, getting prefixed with attr_encrypted so that we can do the data transition. We're writing the data onto new fields. For 7/9, they'll be written onto new encrypted fields. For bank_name and bank_routing_number, we're going to forgo encrypting them and just store them raw since they're not particularly sensitive or private on their own.

PR #2985, opened 2022-08-08
Dependent residency question

embarnard:
Ask how many months in home instead of if they have lived there for more than 6 months

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -77,7 +77,7 @@ module ApplicationHelper
   end
 
   def months_in_home_options_for_select
-    {I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.less_than_six") => 6, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.six_to_seven") => 7, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.seven_to_eight") => 8, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.eight_to_nine") => 9, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.nine_to_ten") => 10, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.ten_to_eleven") => 11, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.eleven_to_twelve") => 12}
+    {I18n.t("views.ctc.questions.dependents.child_residence.select_options.less_than_six") => 5, I18n.t("views.ctc.questions.dependents.child_residence.select_options.six_to_seven") => 6, I18n.t("views.ctc.questions.dependents.child_residence.select_options.seven_to_eight") => 7, I18n.t("views.ctc.questions.dependents.child_residence.select_options.eight_to_nine") => 8, I18n.t("views.ctc.questions.dependents.child_residence.select_options.nine_to_ten") => 9, I18n.t("views.ctc.questions.dependents.child_residence.select_options.ten_to_eleven") => 10, I18n.t("views.ctc.questions.dependents.child_residence.select_options.eleven_to_twelve") => 11}
   end
 
   def suffix_options_for_select
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -76,6 +76,10 @@ module ApplicationHelper
     options
   end
 
+  def months_in_home_options_for_select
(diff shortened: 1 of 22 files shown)


embarnard:
Screen Shot 2022-08-08 at 12 51 06 PM
 We save this residency information in `months_in_home` instead of `lived_with_more_than_six_months`

lib/tasks/backfill_dependents_months_in_home.rake
```diff
@@ -0,0 +1,14 @@
+namespace :dependents do
+  desc "Backfill months_in_home with lived_with_more_than_six_months value"
+  task backfill_months_in_home: :environment do
+    Dependent.where(months_in_home: nil).where.not(lived_with_more_than_six_months: "unfilled").find_in_batches(batch_size: 100) do |batch|
```
embarnard, line 4: This task would alter around 50,038 dependents

app/helpers/application_helper.rb
```diff
@@ -76,6 +76,10 @@ def tin_options_for_select(include_ssn_no_employment: false, include_atin: false
     options
   end
 
+  def months_in_home_options_for_select
+    {I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.less_than_six") => 6, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.six_to_seven") => 7, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.seven_to_eight") => 8, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.eight_to_nine") => 9, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.nine_to_ten") => 10, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.ten_to_eleven") => 11, I18n.t("views.ctc.questions.dependents.child_residence.reveal.select_options.eleven_to_twelve") => 12}
```
caitbot, line 80: Is`6` the best option for the value of `less_than_six`? I worry that it will be confusing in the future/at first glance. Maybe we should consider `0` or '1' or even '5'?

Another option -- rather than using an integer, we could keep the slugs as is and make it an enum. Then we get the method `less_than_six_yes?`. But this would require transforming the data into integers when we insert it into the XML.

And nit but regardless of what value we store, can we change this formatting so that each entry is on a new line? That will help with readability.

config/locales/en.yml
```diff
@@ -2113,8 +2113,24 @@ en:
...
-            title: Did %{name} live with you for more than 6 months in %{current_tax_year}?
+            help_text: If they were born, adopted, or passed away in %{current_tax_year}, select 12 months.
+            reveal:
+              body: If your dependent was at a temporary location in 2021, select the number of months that your home was their official address.
+              list:
+              - school
+              - medical facility
+              - a juvenile facility
+              list_title: 'Some examples of temporary locations:'
+              select_options:
```
caitbot, line 2124: These probably shouldn't be nested under the reveal.

caitbot (commented):
In general, this looks good. 

I'm curious if using an integer is the best way to store this info, it feels like while the IRS expects an integer, we are asking about a range, and maybe it makes sense to store the value as an enum and convert it to the IRS appropriate integer upon XML generation. Down to have a conversation about it if you want.

config/locales/en.yml
embarnard: Good catch! You're right

app/helpers/application_helper.rb
embarnard: Yes! I was a little conflicted about this. I know we talked about doing string enums for this but I went with `months_in_home`because that is a column that was already in the dependents' model and was used in this form app/views/dependents/_form.html.erb. 

In that form, they ask "How many months did they live in your home in %{year}?" and give a drop down of numbers from 0-12. So I guess in that case, less than six would be 5. If we wanted to use enums, we could but then we would have to backfill all the months_in_home values into their enum values. Or I guess we could make a new column and backfill values from months_in_home and lived_with_more_than_six_months into it. And then later drop those columns if it's okay with Max. 

What do you think, is it worth it to do a bunch of rake tasks to change it to enums that might be a bit more syntactically clear or to change the logic on the present months_in_home to check for >= 5 months? @caitbot

caitbot: Oooo sweet, thanks for sharing that info, that totally makes sense. Definitely agree is not super clear what is the best -- I suppose if this method is an established pattern then lets just go with it. 

 So my vote is to keep the integers and update the logic to be `>=5`.

caitbot (approved):

PR #2995, opened 2022-08-12
Update system note for identity cards

mquadri23:


alaroia-cfa (approved):
I approve this pull request. It has some code formatting issues that are optional to fix. One tool that works for me is using RubyMine auto-format - see https://www.jetbrains.com/help/ruby/reformat-and-rearrange-code.html#reformat_code .

I would strongly prefer that you fix the code style issues before merging. If it's important to you not to, then it's OK to merge it as-is.

Good changes overall! Feel free to merge with or without the style changes; no need for re-review if that's all you fix.

PR #2997, opened 2022-08-12
Drop old encrypted_column bonanza

bytheway875:


bytheway875:
This drops the columns from the intake table, the archived intake table, the archived dependent table, and the archived bank accounts table after doing the data migrations. This should be clear to merge and then release.

PR #3000, opened 2022-08-15
Dependent residency: save months in home differently for non-EITC clients

embarnard:


embarnard:
This PR attempts to do five things: 
1) Add the value of months_in_home on the Schedule EIC, right now it is just filling in a "07" placeholder value 

2) Map the values of the dependent residency question so that it matches what we see in figma (after talking to Yvonne it seems we have to stick to this format even though it might be confusing, you can see [here](https://www.irs.gov/pub/irs-pdf/f1040sei.pdf) the IRS doc states in the residency question _'If the child lived with you for more than half of 2021 but less than 7 months, enter “7.”'_ Which means all these values mapped in Application helper should be increased by one right now 

3) Make sure that we are storing months_in_home values differently for EITC and non-EITC clients. EITC clients should have values matching application helper mapper while non-EITC clients should 7 months_in_home saved if they resided with the client more than 6 months and 6 if they resided with the client less than 6 months 

4) Add a rake task to fix values since this [release](https://github.com/codeforamerica/vita-min/releases/tag/version-7.61.16) that don't match this mapping 

5) Make a change so that if a dependent was born in the last 6 months of the year then months in home is 12 (According to [this](https://www.irs.gov/pub/irs-pdf/f1040sei.pdf) that states _"If the child was born or died in 2021 and your home was the child’s home for more than half the time he or she was alive during 2021, enter “12.”_)

lib/tasks/backfill_dependents_months_in_home.rake
```diff
@@ -6,9 +6,22 @@ namespace :dependents do
...
     end
   end
+
+
+  desc "Reformat months_in_home to match new mapping"
+  task reformat_months_in_home: :environment do
+    # released at Wed Aug 10 13:27:48 2022 -0700
+    Dependent.where("created_at >= ?", DateTime.parse('Wed Aug 10 13:27:48 2022 -0700')).where(lived_with_more_than_six_months: "unfilled").find_in_batches(batch_size: 100) do |batch|
+      batch.each do |dependent|
+        next if dependent.intake.claim_eitc_yes? || dependent.months_in_home.nil? || dependent.months_in_home.to_i > 11
```
embarnard, line 21: maybe remove the claim_eitc_yes?

app/lib/submission_builder/ty2021/documents/schedule_eic.rb
```diff
@@ -23,7 +23,7 @@ def document
                 xml.ChildIsAStudentUnder24Ind dependent.full_time_student_yes? && dependent.age_during_tax_year < 24
                 xml.ChildPermanentlyDisabledInd dependent.permanently_totally_disabled_yes?
                 xml.ChildRelationshipCd dependent.irs_relationship_enum
-                xml.MonthsChildLivedWithYouCnt '07' # this is a placeholder value until we collect the actual number of months in intake
+                xml.MonthsChildLivedWithYouCnt dependent.months_in_home.to_s.rjust(2, '0')
```
embarnard, line 26: nit but we could move this out to a formatting method for IRS month expectations

PR #3007, opened 2022-08-17
Update the confirm-dependents page to new designs

caitbot:
image


app/models/dependent.rb
```diff
@@ -146,6 +146,10 @@ def qualifying_eip3?(tax_year = TaxReturn.current_tax_year)
     Efile::DependentEligibility::EipThree.new(self, tax_year).qualifies?
   end
 
+  def qualifying_eitc?(tax_year = TaxReturn.current_tax_year)
```
embarnard, line 149: can you still qualify for EITC if you're not claiming it?

config/locales/es.yml
```diff
@@ -1990,10 +1995,8 @@ es:
           title: "¡Excelente! Revise que sus datos bancarios estén correctos."
         confirm_dependents:
           add_a_dependent: Agregar un dependiente
-          atin: Número de Adopción
-          birthday: Fecha de nacimiento
+          birthday: DOB
```
embarnard, line 1998: FDN?

embarnard (approved):
Looks good!

config/locales/es.yml
caitbot: Ooooo good call out. Is that equivalent?

app/models/dependent.rb
caitbot: fair point fair point, I will add that

config/locales/es.yml
caitbot: Based on googling, I believe it is

PR #3012, opened 2022-08-23
Offboard clients if tin type not ssn valid for employment

embarnard:


embarnard:
This PR attempts to show the EITC offboarding page if the client and if at least one of their dependents uses a tin type that is not valid for employment or an ITIN.
Screen Shot 2022-08-17 at 4 58 58 PM

spec/lib/efile/benefits_eligibility_spec.rb
```diff
@@ -432,5 +433,52 @@
...
+      before do
+        intake.update(dependents: dependents)
+      end
+
+      it "returns false" do
+        expect(subject.qualified_for_eitc?).to eq false
+      end
+    end
+
+    context "when primary and at least one dependents' tin type is SSN" do
```
jenny-heath, line 473: how odd...only one dependent needs a valid ssn?

jenny-heath: oh i guess it makes sense if they're getting $ per qualifying dependent. nvm!

jenny-heath (approved):
2 things:
- looking at the story, i'm not 100% sure of the rule where only one dependent needs a valid SSN for employment. this info could have come from somewhere else though so i would just make sure yvonne tests all of the cases in acceptance
- the story specifies "make sure that the EITC column on the dependents review page shows that their dependents did not qualify for the EITC". i'm too lazy to check whether modifying `qualified_for_eitc?` covers that. also check in acceptance?

otherwise looks good!

PR #3016, opened 2022-08-24
Update refund payment

mquadri23:


app/views/hub/clients/show.html.erb
```diff
@@ -124,68 +124,68 @@
...
+                <% unless @client.intake.drop_off? %>
+                  <% tax_return = @client.tax_returns.find_by(year: TaxReturn.current_tax_year) %>
+                  <% if tax_return %>
+                    <% benefits = Efile::BenefitsEligibility.new(tax_return: tax_return, dependents: @client.intake.dependents) %>
                     <div class="field-display">
-                      <span class="form-question">Outstanding CTC / Advance CTC Received:</span>
-                      <span class="label-value"><%= "$#{benefits.outstanding_ctc_amount}" if benefits.outstanding_ctc_amount %></span>
+                      <span class="form-question"> Outstanding RRC: </span>
+                      <span class="label-value">$<%= benefits.outstanding_recovery_rebate_credit || 0 %></span>
+                      <% case @client.intake.claim_owed_stimulus_money.capitalize %>
```
embarnard, line 169: sometimes I like to break this out into a variable if I'm calling a really long chain multiple times like this

embarnard, line 169: I wonder if we can do something like this 
```
<% status_background_color_class = intake.claim_owed_stimulus_money_yes? ? "background--green" : (intake.claim_owed_stimulus_money_no? ? "background--red" : "background--grey") %>
                      <span class="label-value client-profile__status-label client-profile__status-label <%= status_background_color_class %>"><%= @client.intake.claim_owed_stimulus_money.capitalize %></span>
```
instead of this case statement

embarnard (commented):
Great job @mquadri23! Asheesh and I started this review. Some of these comments are more nit-picky and are not a hard requirement for merge but some of them should probably be changed like removing the heigh, width, position stuff in the CSS. But I will call you and we can review it together :)

PR #3017, opened 2022-08-25
Rename Diy::TaxSlayerController to Diy::ContinueToFsaController

alaroia-cfa:
We need to change the URL of the last page of the DIY flow.

While I was in the neighborhood, I made some other changes.

- Add DIY flow to flow explorer.
  - Rename controller actions in DIY flow to match expectations of flow explorer. They used to be `show` and `create`; they are now `edit` and `update`.
- Within Flow Explorer UI, clarify copy to explain that the 'reference object' can affect which controllers are visible.
- When viewing GYR/DIY flow from CTC hostname, redirect. This addresses a problem I ran into where I was at https://ctc.demo.getyourrefund.org/en/flows/gyr somehow and I clicked into the personal info controller and I got a URL that was on the CTC hostname.


config/locales/en.yml
```diff
@@ -74,8 +74,18 @@ en:
         someone_invited_you: "%{their_name} (%{their_email}) has invited %{your_name} to create an account on GetYourRefund to help people with their taxes."
         subject: You've been invited to GetYourRefund
   diy:
-    diy_intakes:
-      new:
+    continue_to_fsa:
+      edit:
+        continue_to_tax_slayer: Continue to TaxSlayer
```
tgrathwell, line 79: is there a separate story somewhere for "remove all instances of TaxSlayer from the copy"? because I thought the high level goal was to get that word off the website, not just out of the URL

alaroia-cfa: That's a great question. I **think** when Yvonne wrote this story I am working on, Yvonne assumed this page doesn't show up at all; probably **this** is the story where we decide if it's OK to say TaxSlayer on this page. I can take that question to Yvonne.

alaroia-cfa: @tgrathwell I checked with Yvonne, and she says on Slack:

> actually, don’t let this be a blocker. we have another story for the copy on that page. feel free to proceed!

https://cfa.slack.com/archives/GT1B45BS5/p1661445229317149

alaroia-cfa:
Travis Slack'd me to say it's cool so I'll merge.

PR #3018, opened 2022-08-25
Include Optimizely in production environment

caitbot:


alaroia-cfa:
Let's find out more about how Optimizely works before merging this. I probably should have made time to do that earlier in the process, but maybe now you and I can find out some stuff together Catie!

tgrathwell:
this got committed as https://github.com/codeforamerica/vita-min/pull/3024

PR #3020, opened 2022-08-25
Restrict GyrCompletionSurvey to clients whose tax returns are ALL in terminal statuses

jenny-heath:


app/models/survey_messages/gyr_completion_survey.rb
```diff
@@ -7,12 +7,10 @@ def self.clients_to_survey(now)
...
         .where(SENT_AT_COLUMN => nil)
         .where(intake: { type: "Intake::GyrIntake" })
+        .where.not(id: Client.has_active_tax_returns.select(:id))
         .where(
           tax_returns: {
-            service_type: "online_intake",
             tax_return_transitions: TaxReturnTransition
-              .where(to_state: RELEVANT_STATES)
-              .where(most_recent: true)
               .where(created_at: (now - 30.days)...(now - 1.day))
```
jenny-heath, line 14: i don't love having to do a whole new query on transitions after already using the new scope, but i don't know how else to pick out the recent ones...

spec/models/survey_messages/gyr_completion_survey_spec.rb
```diff
@@ -2,45 +2,66 @@
 
 RSpec.describe SurveyMessages::GyrCompletionSurvey do
```
jenny-heath, line 3: tried reorganizing these tests to make them more intuitive but the odd thing about them is that the thing being manipulated isn't the data in the setup (date of the tax return transition) but the argument being passed in (`expected_send_time` etc)

would also love another set of eyes on whether these cases are comprehensive enough

PR #3022, opened 2022-08-25
Update the link on the GetCTC button on GYR homepage

mquadri23:


alaroia-cfa (approved):
Good fix.

The old code was creating a "path" not a "URL". See in `rails c`:

```
irb(main):010:0> Rails.application.routes.url_helpers.root_with_source_path(host: MultiTenantService.new(:ctc).host, source: "gyractivechoice")
=> "/gyractivechoice"
```

We needed something more like this:

```
irb(main):011:0> Rails.application.routes.url_helpers.root_with_source_url(host: MultiTenantService.new(:ctc).host, source: "gyractivechoice")
=> "http://ctc.localhost:3000/gyractivechoice"
```

Your fix accomplishes that based on my manual testing on localhost.

Non-blocking: I prefer `root_with_source_url` as the fix here. I encourage you to make that change and merge without further code review.

PR #3030, opened 2022-08-31
Calculate EITC amount and edit /confirm-payments

embarnard:


app/lib/efile/benefits_eligibility.rb
```diff
@@ -89,6 +89,29 @@ def claimed_recovery_rebate_credit
       outstanding_recovery_rebate_credit
     end
 
+    def eitc_amount
```
embarnard, line 92: @wschaefer @jenny-heath this is the function you'll want to use for your branch

embarnard:
This PR adds a calculation for EITC amount, updates the /confirm-payments page to show EITC and total amount, and shows it on the client profile.

Screen Shot 2022-09-01 at 4 31 32 PM

Screen Shot 2022-09-01 at 4 29 03 PM

app/lib/efile/dependent_eligibility/eligibility.rb
```diff
@@ -21,6 +21,7 @@ def benefit_amounts
             eip2: eip_two_eligibility.benefit_amount,
             eip3: eip_three_eligibility.benefit_amount,
             ctc: ctc_eligibility.benefit_amount
+            #add EITC?
```
embarnard, line 24: I guess I should remove this comment, but wanted a read on if this should be added to this list. It seems this list is only used for calculating the total_benefit_amount for a dependent which right now is only used in a test

embarnard: But I guess we don't calculate EITC on a per dependent basis more on the whole so I'm thinking we would leave it out of here

app/controllers/ctc/questions/confirm_payment_controller.rb
```diff
@@ -10,13 +10,16 @@ def edit
         benefits = Efile::BenefitsEligibility.new(tax_return: tax_return, dependents: current_intake.dependents)
         @ctc_amount = benefits.outstanding_ctc_amount
         @third_stimulus_amount = benefits.outstanding_eip3
-        @not_collecting = @ctc_amount.zero? && @third_stimulus_amount.zero?
+        @eitc_amount = benefits.claiming_and_qualified_for_eitc? ? benefits.eitc_amount : nil
+        @not_collecting = @ctc_amount.zero? && @third_stimulus_amount.zero? && @eitc_amount.zero?
```
alaroia-cfa, line 14: Blocking request: I think this crashes when `eitc_amount` is `nil`. I encourage you to compute `@total_amount` first, and then compute `@not_collecting` as `@total_amount == 0`.

Non-blocking request: What do you think about renaming the variable to be positive, like `@collecting` (rather than `@not_collecting`), or perhaps inlining the variable? I think it's only used in the template, which could just as easily check if `@total_amount > 0`. The best reason IMHO to keep the variable is to cover it in unit tests.

L13 sometimes sets `@eitc_amount` to `nil`, `@eitc_amount.zero?` will crash (see `rails c` transcript below). If `@total_amount` already uses `.compact` to handle the nil case, rely on that. I'm assuming the intent is that `@not_collecting` is `true` if `benefits.claiming_and_qualified_for_eitc?` is false and everything else is zero.

```
irb(main):001:0> nil.zero?
Traceback (most recent call last):
        1: from (irb):1
NoMethodError (undefined method `zero?' for nil:NilClass)
```

app/views/ctc/questions/confirm_payment/edit.html.erb
```diff
@@ -16,6 +16,16 @@
         <p class="spacing-below-15"><%= number_to_currency(@third_stimulus_amount) %></p>
       </div>
     <% end %>
+    <% if @eitc_amount %>
```
alaroia-cfa, line 19: Non-blocking: What's your philosophy on checking within the ERB if `Flipper.enabled?(:eitc)`? What you've written is fine since this value is nil if EITC is disabled. I prefer adding `Flipper` checks everywhere we render EITC stuff within the ERB, and very-few-but-just-enough checks in the backend. Feel free not to answer if you're busy.

app/views/hub/clients/show.html.erb
```diff
@@ -178,6 +178,12 @@
                       <span class="label-value"><%= "$#{benefits.advance_ctc_amount_received}" if benefits.advance_ctc_amount_received %></span>
                     </div>
                   <% end %>
+                  <% unless benefits.eitc_amount.nil? %>
+                    <div class="field-display">
+                      <span class="form-question">EITC estimated amount:</span>
+                      <span class="label-value"><%= "$#{benefits.eitc_amount}" %></span>
```
alaroia-cfa, line 184: Non-blocking: This could use the I18n number formatting.

app/lib/efile/benefits_eligibility.rb
```diff
@@ -89,6 +89,28 @@ def claimed_recovery_rebate_credit
...
+      # EITC amount = [phase-in function, plateau amount, phase-out function].min
+      # where phase-in function = earned-income amount * phase-in rate
+      # But b/c of simplified filing rules, those above the phase out threshold cannot use the tool
+      # so we are not including the phase-out function but keep in mind this might change next year
+      return nil unless qualified_for_eitc?
+
+      earned_income = intake.w2s.sum(&:wages_amount).to_f
+
+      case dependents.count { |d| d.qualifying_eitc? && (d.qualifying_child? || d.qualifying_relative?) }
```
alaroia-cfa, line 101: This `case` line is pretty cool.

spec/lib/efile/benefits_eligibility_spec.rb
```diff
@@ -470,4 +470,112 @@
...
+        allow(subject).to receive(:qualified_for_eitc?).and_return true
+        allow_any_instance_of(Dependent).to receive(:qualifying_eitc?).and_return(true)
+      end
+
+      context "when they have 0 EITC-qualifying children" do
+        before do
+          intake.dependents.destroy_all
+        end
+
+        context "their income is $2,724" do
```
alaroia-cfa, line 499: Non-blocking: I don't understand the purpose of the specific income brackets in these tests. They're OK with me, but I wonder if more description would help. Testing arithmetic is always a bit of a pain, and it's definitely non-blocking since people can change it later as people continuously use/refactor the EITC code.

alaroia-cfa (changes requested):
Only one blocking change (a `nil`-related issue). Nice!

app/views/hub/clients/show.html.erb
embarnard: @alaroia-cfa like putting "EITC estimated amount:" in the yml files? I didn't put it in originally b/c there was some other text on the client profile that was left out.

spec/lib/efile/benefits_eligibility_spec.rb
embarnard: I put all these test cases in because they were what were provided to me by Gabriel in here https://docs.google.com/spreadsheets/d/1SAGDWnAmAHo9Ya-f8SItCfSpma9fwbiT6rAC9LSv33k/edit#gid=0

app/controllers/ctc/questions/confirm_payment_controller.rb
embarnard: Yes that makes sense! Going to get rid of non-collecting and just check for @total.zero? in the code

PR #3031, opened 2022-08-31
test bulk tax return inserting so we will know whether it needs to be jobbified

tgrathwell:


tgrathwell:
@alaroia-cfa 

```
Seeder.new.create_many_clients_and_tax_returns(1000)

5.times.map {
  Benchmark.realtime {
    Seeder.new.create_tax_return_selection_slow(Intake.where("primary_first_name LIKE 'TaxReturnImportTester%'").pluck('client_id')); nil
  }
}

=> [1.49, 1.59, 1.57, 1.90, 1.73]

5.times.map {
  Benchmark.realtime {
    Seeder.new.create_tax_return_selection_fast(Intake.where("primary_first_name LIKE 'TaxReturnImportTester%'").pluck('client_id')); nil
  }
}
=> [0.42, 0.19, 0.19, 0.17, 0.16]
```

so the bulk insert is 5-10x faster I guess. hard to say how much faster/slower the prod DB will be though

this is with 1000 records, so if our biggest selection is 50,000 then it would take 50 * 0.2sec = 10sec to do the fast way, or a whole minute to do it the slow way

alaroia-cfa:
10 sec is so borderline. On one hand, totally livable. OTOH people freak out after around 2 sec.

PR #3033, opened 2022-09-01
Update the W2 XML to allow for multiple W2s and use intake values

caitbot:


wschaefer (approved):
just need wages and withholding rounded!

PR #3036, opened 2022-09-01
Remove unneeded eitc dependent conditions

wschaefer:
They are tied to the primary and we check them at the intake level
so we don't need to check them for each dependent


tgrathwell (approved):
Code deleted is code tested.

PR #3042, opened 2022-09-06
Update the ScheduleLEP pdf to only include the first page

mquadri23:


tgrathwell (approved):
Asheesh and I validated that the PDF now (yay) lost the instruction page and (yay) is still clickable.

PR #3049, opened 2022-09-09
Check that dependent IP pin exists before adding it to Schedule EIC

caitbot:
Since we weren't checking, when testing the XML for EITC we were getting the following error:

```
["157:0: ERROR: Element '{http://www.irs.gov/efile}IdentityProtectionPIN': [facet 'length'] The value has a length of '0'; this differs from the allowed length of '6'.", "157:0: ERROR: Element '{http://www.irs.gov/efile}IdentityProtectionPIN': [facet 'pattern'] The value '' is not accepted by the pattern '([!-~£§ÁÉÍÑÓ×ÚÜáéíñóúü] ?)*[!-~£§ÁÉÍÑÓ×ÚÜáéíñóúü]'."] 
```

Checking that it is present fixes the issue.


caitbot (commented):
reviewed this with JH irl

PR #3065, opened 2022-09-15
Use the same W-2 views in portal & questions

alaroia-cfa:
This would allow us to unify portal views with intake question views.


alaroia-cfa:
review'd by jh

PR #3070, opened 2022-09-16
ApplicationCable: rescue and ignore warden UncaughtThrowError

tgrathwell:
We've been silencing this in Sentry for months and not doing anything about it, so this is doing something (?) about it

It seems to have been caused by bumping some gems a couple months ago, but it's not certain which gem or whether that's really true


wschaefer (approved):
thank you

PR #3074, opened 2022-09-20
When adding a W-2, allow client to conveniently log out b/c they will find their W-2 later #183213718

alaroia-cfa:
Since product & program want very easy-to-query data about when this button is clicked, we also store a timestamp of when that occurred.

### DIFF
--- a/app/controllers/ctc/questions/w2s_controller.rb
+++ b/app/controllers/ctc/questions/w2s_controller.rb
@@ -13,10 +13,7 @@ module Ctc
       end
 
       def add_w2_later
-        analytics_journey = AnalyticsJourney.find_or_initialize_by(client: current_intake.client)
-        analytics_journey.update(w2_logout_add_later: Time.now)
-
-        send_mixpanel_event(event_name: "w2_logout_add_later", data: MixpanelService.data_from([current_intake.client, current_intake]))
+        track_click_history(:w2_logout_add_later)
 
         clear_intake_session
         redirect_to root_path
--- a/app/controllers/ctc/questions/w2s_controller.rb
+++ b/app/controllers/ctc/questions/w2s_controller.rb
@@ -18,7 +18,8 @@ module Ctc
 
         send_mixpanel_event(event_name: "w2_logout_add_later", data: MixpanelService.data_from([current_intake.client, current_intake]))
 
-        sign_out current_intake.client
+        clear_intake_session
+        redirect_to root_path
       end
 
       def next_path
--- a/app/controllers/ctc/questions/w2s_controller.rb
+++ b/app/controllers/ctc/questions/w2s_controller.rb
@@ -12,6 +12,15 @@ module Ctc
         benefits_eligibility.claiming_and_qualified_for_eitc_pre_w2s?
       end
 
+      def add_w2_later
+        analytics_journey = AnalyticsJourney.find_or_initialize_by(client: current_intake.client)
+        analytics_journey.update(w2_logout_add_later: Time.now)
+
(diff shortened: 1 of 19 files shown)


app/controllers/ctc/questions/w2s_controller.rb
```diff
@@ -12,6 +12,16 @@ def self.show?(intake)
         benefits_eligibility.claiming_and_qualified_for_eitc?
       end
 
+      def add_w2_later
+        analytics_journey = AnalyticsJourney.find_or_initialize_by(client: current_intake.client)
```
alaroia-cfa, line 16: Safest is to use `create_or_find_by` per [this article](https://sikac.hu/use-create-or-find-by-to-avoid-race-condition-in-rails-6-0-f44fca97d16b).

app/models/analytics_journey.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# Indexes
+#
+#  index_analytics_journeys_on_client_id  (client_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (client_id => clients.id)
+#
+class AnalyticsJourney < ApplicationRecord
```
alaroia-cfa, line 19: I think it might be smart to move this into `module Analytics` and move `efile_acceptance_analytics` there too.

app/models/analytics_journey.rb
```diff
@@ -0,0 +1,21 @@
...
+  belongs_to :client
```
alaroia-cfa, line 20: I want this to be unique on client.

app/models/data_science/click_history.rb
```diff
@@ -0,0 +1,23 @@
+# == Schema Information
+#
+# Table name: ds_click_histories
+#
+#  id                  :bigint           not null, primary key
+#  w2_logout_add_later :datetime
```
tgrathwell, line 6: If we don't make a column for each property, we'd be able to add new properties seamlessly without a migration. For example, a string for event_name.

tgrathwell: or _at

jenny-heath: i think the idea was to have many events on one record (the record as a whole being a "journey"/"history") since it's a `has_one` on the client

app/models/data_science/click_history.rb
```diff
@@ -0,0 +1,23 @@
...
+# Indexes
+#
+#  index_ds_click_histories_on_client_id  (client_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (client_id => clients.id)
+#
+module DataScience
+  class ClickHistory < ApplicationRecord
```
jenny-heath, line 20: ngl kinda hate this new model name. why are we giving data science their own namespace??

jenny-heath:
closing to re-open for db stuff

app/models/data_science/click_history.rb
alaroia-cfa: @jenny-heath I think we should change it; I can take that on.

Although @wschaefer and I came up with the name last week and were fond of it, I talked with a data science team member (Chris Chan) today over Tuple, and he really did not like a namespace for the data science team. Chris said it's better to call this Ruby module 'analytics' or 'program' or something not targeted at the data science team. I think analytics is fine, even though there would be 3 things called analytics (AcceptedTaxReturnAnalytics, this model, and the `analytics.*` SQL VIEW).

There's some enthusiasm for having this table store _events_ (one event per row) rather than one row per client plus one column per event type. @tgrathwell suggested that to me earlier, and so did Chris Chan. For the table/class name, Chris liked the phrase "event stream." I think was Travis's idea, too.

I'll probably go with `Analytics::EventStream` or something like that and do a rename tomorrow so it's renamed before it has prod users.

PR #3077, opened 2022-09-20
ActcWithoutDependents offboarding requires authentication

wschaefer:
Because the view needs a current_intake to be present


wschaefer (commented):
reviewed with travis

PR #3081, opened 2022-09-21
Testing feedback copy changes

mquadri23:


app/controllers/ctc/questions/confirm_payment_controller.rb
```diff
@@ -11,7 +11,8 @@ def edit
         @ctc_amount = benefits.outstanding_ctc_amount
         @third_stimulus_amount = benefits.outstanding_eip3
         @eitc_amount = benefits.claiming_and_qualified_for_eitc? ? benefits.eitc_amount : nil
-        @total_amount = [@ctc_amount, @third_stimulus_amount, @eitc_amount].compact.sum
+        @fed_income_tax_withholding_amount = current_intake.w2s.pluck(:federal_income_tax_withheld).sum
```
embarnard, line 14: We should probably have a test for this in the controller spec where we create multipe w2s for an intake

embarnard:
Make sure you include the pivotal id in the branch name so that we can see the link to this PR in pivotal. So instead of `testing-feedback-copy-changes` it would be `testing-feedback-copy-changes-183063754`

embarnard (changes requested):
Great job @mquadri23!! This PR looks really good -- just have a few small suggestions and one request to add a test for the `fed_income_tax_withholding_amount`

app/views/layouts/ctc.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= extends :application do %>
-  <% content_for :app_name do %>GetCTC<% end %>
+  <% content_for :app_name do %>GetCTC
```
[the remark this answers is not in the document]
mquadri23: oh hmm I don't remember explicitly making this change but I may have ran a 'reformat' on this file and it formatted it like this is my guess. Should I change it back?

app/controllers/ctc/questions/confirm_payment_controller.rb
mquadri23: I paired with Whitney and we decided to update this line to use the existing method for calculating the fed_income_tax_withholding_amount. Since there are tests for the the method already we decided that would be sufficient as far as testing this instance variable here. Let me know if you agree with that!

PR #3098, opened 2022-09-27
Schedule eic checkboxes 4a and 4b #183316475

jenny-heath:


spec/lib/submission_builder/ty2021/documents/schedule_eic_spec.rb
```diff
@@ -48,6 +48,37 @@
     expect(dependent_nodes[1].at("MonthsChildLivedWithYouCnt").text).to eq "08"
   end
 
+  context "checkboxes 4a and 4b" do
```
jenny-heath, line 51: these tests are actually repetitive with the ones above but i like showing that different cases are being tested with the context blocks instead of all of the fields lumped together? idk

spec/models/efile_submission_dependent_spec.rb
```diff
@@ -105,4 +105,151 @@
...
+    before do
+      efile_submission_dependent.efile_submission.tax_return.update(year: DateTime.now.year)
+      efile_submission_dependent.intake.update(primary_birth_date: 50.years.ago)
+      efile_submission_dependent.intake.update(spouse_birth_date: 50.years.ago)
+    end
+
+    context "when the dependent is between 19 and 24 and a full time student" do
+      let(:birth_date) { 22.years.ago }
+      let(:age_during_tax_year) { 22 }
+      let(:full_time_student) { "yes" }
```
jenny-heath, line 122: for some reason i found it impossible to use `EfileSubmissionDependent.create_qualifying_dependent` to create `efile_submission_dependent`—it just refused to create one. as such, we have this awkward duplication of birth date and age during tax year because the method being tested checks A) the dependent's `age_during_tax_year` is between 19 and 24 and B) the dependent is younger than the filer(s) by comparing their birth dates.

we started out comparing `age_during_tax_year` to a calculation of the filers' age during tax year, but i got confused by that because i was worried that the calculation of a filer's age during tax year could be off by one (since one is 2 ages, respectively, during any given tax year). would love to talk this through with someone!

jenny-heath:
@wschaefer it looks like the actual schedule eic pdf (`app/lib/pdfs/tax_documents/f1040sei-TY2021.pdf`) was changed as part of the very first commit (which maybe you paired on?) but i can't tell in what way it was changed. do you know whether it was intentional? if not i'll revert it

app/lib/submission_builder/ty2021/documents/schedule_eic.rb
```diff
@@ -20,8 +20,8 @@ def document
                 xml.IdentityProtectionPIN dependent.ip_pin if dependent.ip_pin.present?
                 xml.QualifyingChildSSN dependent.ssn
                 xml.ChildBirthYr dependent.birth_date.year
-                xml.ChildIsAStudentUnder24Ind dependent.full_time_student_yes? && dependent.age_during_tax_year < 24
-                xml.ChildPermanentlyDisabledInd dependent.permanently_totally_disabled_yes?
+                xml.ChildIsAStudentUnder24Ind dependent.schedule_eic_4a?
+                xml.ChildPermanentlyDisabledInd dependent.schedule_eic_4b? if dependent.schedule_eic_4a? == false
```
mquadri23, line 24: maybe `unless`? not sure

app/models/efile_submission_dependent.rb
```diff
@@ -41,4 +41,13 @@ def self.create_qualifying_dependent(submission, dependent)
       )
     end
   end
+
+  def schedule_eic_4a?
+    full_time_student_yes? && age_during_tax_year > 19 && age_during_tax_year < 24 &&
```
mquadri23, line 46: `>` vs `>=` ?

spec/models/efile_submission_dependent_spec.rb
mquadri23: 💀

spec/models/efile_submission_dependent_spec.rb
```diff
@@ -105,4 +105,151 @@
...
+
+          it "they do not meet the conditions for checkbox 4a" do
+            expect(efile_submission_dependent.schedule_eic_4a?).to eq false
+          end
+        end
+      end
+    end
+
+    context "when the dependent is over 24" do
+      let(:age_during_tax_year) { 25 }
```
mquadri23, line 191: would be nice to pick just 1

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -30,10 +30,12 @@ def dependents_info(dependent_nodes)
         answers["ChildBirthYr#{index + 1}[1]"] = dependent.at("ChildBirthYr").text[1]
         answers["ChildBirthYr#{index + 1}[2]"] = dependent.at("ChildBirthYr").text[2]
         answers["ChildBirthYr#{index + 1}[3]"] = dependent.at("ChildBirthYr").text[3]
-        answers["ChildIsAStudentUnder24IndYes#{index + 1}"] =  xml_check_to_bool(dependent.at("ChildIsAStudentUnder24Ind")) ? "Yes" : nil
-        answers["ChildIsAStudentUnder24IndNo#{index + 1}"] = !xml_check_to_bool(dependent.at("ChildIsAStudentUnder24Ind")) ? "Yes" : nil
-        answers["ChildPermanentlyDisabledIndYes#{index + 1}"] = xml_check_to_bool(dependent.at("ChildPermanentlyDisabledInd")) ? "Yes" : nil
-        answers["ChildPermanentlyDisabledIndNo#{index + 1}"] = !xml_check_to_bool(dependent.at("ChildPermanentlyDisabledInd")) ? "Yes" : nil
+        bool = xml_bool_to_bool(dependent.at("ChildIsAStudentUnder24Ind"))
```
alaroia-cfa, line 33: @jenny-heath @mquadri23 I had to change these lines since it's a `BooleanType` not `CheckboxType` in the XSD.

app/models/efile_submission_dependent.rb
```diff
@@ -41,4 +41,13 @@ def self.create_qualifying_dependent(submission, dependent)
       )
     end
   end
+
+  def schedule_eic_4a?
```
alaroia-cfa, line 45: Would be nice to move these to the `ScheduleEic` class instead.

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -25,17 +25,29 @@ def dependents_info(dependent_nodes)
...
+      answers["ChildBirthYr#{index + 1}[1]"] = dependent.at("ChildBirthYr").text[1]
+      answers["ChildBirthYr#{index + 1}[2]"] = dependent.at("ChildBirthYr").text[2]
+      answers["ChildBirthYr#{index + 1}[3]"] = dependent.at("ChildBirthYr").text[3]
+      case xml_boolean_type_value(dependent.at("ChildIsAStudentUnder24Ind"))
+      when true
+        answers["ChildIsAStudentUnder24IndYes#{index + 1}"] = "1"
+        answers["ChildIsAStudentUnder24IndNo#{index + 1}"] = nil
+      when false
+        answers["ChildIsAStudentUnder24IndYes#{index + 1}"] = nil
+        answers["ChildIsAStudentUnder24IndNo#{index + 1}"] = "2"
```
alaroia-cfa, line 39: The PDF requires that you use the value "2" to indicate a checked "No" checkbox! Yowza.

app/lib/irs1040_schedule_eic_pdf.rb
```diff
@@ -25,17 +25,29 @@ def dependents_info(dependent_nodes)
...
+      end
+      case xml_boolean_type_value(dependent.at("ChildPermanentlyDisabledInd"))
```
alaroia-cfa, line 41: The reason for the `case` stuff is that we need to allow ourselves to let `nil` pass through as neither `true` nor `false`.

PR #3166, opened 2022-10-05
With certain CA-related source params, redirect to /stimulus

jenny-heath:
here's the whole story description:

> WHY:
>
> a major outreach partner is sketched out by getctc.org/stimulus?s=source links and wants to send people to getctc.org/source links instead. If they can't, they might not link to us, and we would miss out on thousands of returns.
> 
> WHAT:
> 
> make getctc.org/source redirect to getctc.org/stimulus?s=source (or, equivalently, redirect to getctc.org/stimulus and store the source) for the following:
> GetCTC.org/cdss
> GetCTC.org/ca
> GetCTC.org/cactc
> GetCTC.org/cagov

### DIFF
--- a/app/controllers/ctc/ctc_pages_controller.rb
+++ b/app/controllers/ctc/ctc_pages_controller.rb
@@ -16,11 +16,11 @@ module Ctc
       redirect_to root_path and return if params[:source].present?
 
       case session[:source]
-      when "cactc", "fed", "child"
+      when "fed", "child"
         redirect_to action: :help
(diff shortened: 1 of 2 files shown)


alaroia-cfa (approved):
LGTM. During acceptance testing, we should validate that the source is sticking through into the intake record. Or maybe you have? Or we could add a feature spec showing that.

Our `source` param stuff is just a little scary to me and I sometimes prefer to see it in action.

PR #3170, opened 2022-10-06
Remove the Child Residence Exceptions page

embarnard:


app/models/archived/dependent_2021.rb
```diff
@@ -170,7 +170,7 @@ def meets_qc_residence_condition_generic?
       # The age check is handled in the year-specific rules; the rest is handled here.
       lived_with_more_than_six_months_yes? ||
         (lived_with_more_than_six_months_no? &&
-          (born_in_2020_yes? || passed_away_2020_yes? || placed_for_adoption_yes? || permanent_residence_with_client_yes?))
```
embarnard, line 173: There are a few fields that have been used to store information about the dependent's residence that have changed overtime to accommodate how we collect this information differently.

alaroia-cfa (approved):
Optionally deploy a commit with `show?` -> `false` first, so that no one lands on the page, so there's no issues with people being on the page then the page being deleted then the client clicks next.

PR #3174, opened 2022-10-06
Fix issue where "Complete CTC form" link would lead to a 404 if the client was last editing a dependent or w2

tgrathwell:
`current_step` was always set to /questions/whatever, when URLs for subobjects can be more complicated like /questions/w2/:id/whatever


app/controllers/questions/questions_controller.rb
```diff
@@ -33,7 +33,7 @@ def update
     end
 
     def current_path(params = {})
-      question_path(self.class.to_param, params)
+      request.path
```
jenny-heath, line 36: AL and i took an idle friday afternoon look at this and are wondering, do we even need `current_path` at all? it seems to mostly be used in `form_with`s in `erb` files to post to ones own path which A) might be the default anyway and B) if not, `request` can be accessed in the view so maybe doesn't need this helper. a little late in the day to be sure but leaving you with this thought in case it's useful.

spec/features/ctc/eitc_spec.rb
```diff
@@ -111,6 +111,36 @@
     expect(page).to have_selector("h1", text: I18n.t("views.ctc.portal.home.title"))
   end
 
+  scenario "a client who is in the middle of the W2 but has to come back to finish later" do
```
jenny-heath, line 114: cool!

spec/features/ctc/eitc_spec.rb
```diff
@@ -42,6 +42,24 @@
     expect(page).not_to have_text 'lumen inc'
   end
 
+  scenario "a client who starts a W-2 can exit the app then continue from the portal" do
+    fill_in_can_use_ctc(filing_status: "single", claim_eitc: true)
+    fill_in_eligibility
+    fill_in_basic_info
+
+    expect(page).to have_selector("h1", text: I18n.t('views.ctc.questions.investment_income.title',current_tax_year: current_tax_year))
```
alaroia-cfa, line 50: trivial: should be `, ` not `,`

alaroia-cfa, line 50: For some reason this line is failing. Maybe the order of quetsions moved around?

app/controllers/questions/questions_controller.rb
alaroia-cfa: @wschaefer What if we remove this hunk from the diff? Once we add the W-2s-specific code, we don't need this change I think.

PR #3176, opened 2022-10-07
Add eitc full launch to session toggles

jenny-heath:


spec/controllers/application_controller_spec.rb
```diff
@@ -872,6 +873,18 @@ def index
       end
     end
 
+    context "during full launch" do
+      around do |example|
+        Timecop.freeze(Rails.configuration.eitc_full_launch + 1.day) do
```
jenny-heath, line 878: Maybe it's nicer to mock the config than to mock the clock. Tis OK tho - AL

jenny-heath (commented):
approve says al

PR #3181, opened 2022-10-11
Enforce limit of 22 chars in both W-2 city fields [#183510308]

alaroia-cfa:
This adds an IRS CityType validator. If you want to see where the regex came from, do a search for `CityType` in RubyMine; you'll find the XSD type definition.

This fixes a live problem where clients are entering data that doesn't validate when it comes to make the IRS bundle, i.e., we get a BUNDLE-FAIL.

I want more custom validators for IRS data types in the codebase! I think it's better to have a Rails custom validator that embodies all the rules for the XML schema type, rather than have a sprinkling of validation rules in the form. One downside of this approach is we have more lines of code in the codebase, but I think the lines are good lines. I also think those custom validators should handle the length validation like this one does.

I'd like to know if people agree, or disagree, or don't care!

## Screenshots

image

image

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1683,6 +1683,7 @@ en:
     file_type: Please upload a valid document type. Accepted types include %{valid_types}.
     file_zero_length: Please upload a valid file. The file you uploaded appears to be empty.
     ip_pin: IP PINs must be a 6 digit number.
+    irs_city: Only letters A-Z, a-z, and space are accepted.
     irs_street_address: Only numbers 0-9, letters A-Z and a-z, hyphen, slash and single spaces are accepted.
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -1683,6 +1683,7 @@ en:
     file_type: Please upload a valid document type. Accepted types include %{valid_types}.
     file_zero_length: Please upload a valid file. The file you uploaded appears to be empty.
     ip_pin: IP PINs must be a 6 digit number.
+    irs_city: Only letters A-Z, a-z, and space are accepted.
```
alaroia-cfa, line 1686: flag to design? space_s_?

PR #3197, opened 2022-10-17
Copy changes to W2 pages and others

tgrathwell:


wschaefer (approved):
It seemed like more of the english one changed than the spanish - was that on purpose?

PR #3200, opened 2022-10-17
W2 stuff: preserve numbers, dash, and ampersand when transforming employer name into EmployerNameControlTxt

tgrathwell:
previously, clients who entered a name like '0' (which is a valid EmployerName, even if it's probably wrong) would see a bundle fail because the EmployerNameControlTxt would become an empty string.

now, the EmployerNameControlTxt will be '0' which will successfully bundle even if it probably won't lead to an ideal outcome


wschaefer (approved):
Looks good! It might be nice if the test also tested the `.first(4)` by having more valid characters in the test string (ie 'a & - 2 banana' -> 'A&-2')

PR #3201, opened 2022-10-18
CTC: on save for phone number, get metadata from Twilio (carrier, type)

wschaefer:


app/jobs/get_phone_metadata_job.rb
```diff
@@ -0,0 +1,19 @@
...
+  queue_as :default
+
+  def perform(intake)
+    return if intake.phone_number.blank?
+
+    metadata = TwilioService.get_metadata(phone_number: intake.phone_number)
+
+    return if metadata.blank?
+
+    if metadata["carrier_name"]
```
alaroia-cfa, line 11: An idea I want to float - does it make sense to have this be a separate model/table, like `PhoneMetadata`? I personally prefer that since it's not data the client directly is providing. I'm creeping on your PR and have no real context from the work you've done on it, so feel free to ignore as needed.

jenny-heath: i feel silly making a new table for just two columns? also i think there are a couple of things on intake the client doesn't provide directly so it wouldn't be without precedent

PR #3203, opened 2022-10-19
Remove multiple whitespaces in box d control num W2

embarnard:


app/forms/ctc/w2s/employer_info_form.rb
```diff
@@ -24,6 +24,7 @@ class EmployerInfoForm < W2Form
 
       before_validation do
         self.employer_ein = employer_ein&.delete('-')
+        self.box_d_control_number = box_d_control_number&.squish
```
alaroia-cfa, line 27: We have `before_validation_squish` above with a few attributes. What do you think of adding this Box D thing into that? Not a blocker.

Should the Box D Control Number get an `irs_xyz_type` custom validator like the Business Name got one? Not a blocker.

alaroia-cfa (approved):
Approved w/ two questions which are not blockers.

PR #3213, opened 2022-10-25
Catch error when 1040 pdf not yet attached for client viewing #183304032

jenny-heath:


alaroia-cfa (approved):
I am sad that any `Document` records don't have a `.upload`! Is there a way to prevent the useless `Document` from getting created in this state?

Not a blocker. LGTM.

PR #3221, opened 2022-10-26
Remove rails-erb-loader JS dependency

alaroia-cfa:
It was last updated 4 years ago, and we do not need it. In the process, change how we store Mixpanel data in the DOM.

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -17,19 +17,9 @@ window.$ = $;
 RailsUJS.start();
 ActiveStorage.start();
 
-// *
-// IMPLEMENTATION NOTE 2/22/21
-// I want us to be able to import honeycrisp JS directly from honeycrisp, but the main JS we need is all included
-// right inside of the cfa_styleguide_main sprockets JS file. I'm planning on making a PR to honeycrisp to change this imminently. - SB
-//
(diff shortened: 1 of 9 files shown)


app/javascript/packs/application.js
```diff
@@ -17,19 +17,9 @@ window.$ = $;
...
-//
-//import "<%#= File.join(Gem.loaded_specs['cfa-styleguide'].full_gem_path, 'app', 'assets', 'javascripts', 'honeycrisp') %>";
-
 import "../lib/honeycrisp";
 
-// *
-<% unless Rails.env.test? %>
-  AjaxMixpanelEvents.init();
-<% end %>
+AjaxMixpanelEvents.init();
```
tgrathwell: i had a desire to keep this behavior (don't do mixpanel ajax stuff in full-stack tests) so I added another commit to try to restore it but didn't actually test to see if it works yet

alaroia-cfa: I agree with the idea of the test environment not sending the Mixpanel AJAX messages. I tested out a different approach to accomplish the same end:

```
pageData.sendMixpanelBeacon
```

I figure it's good for the test suite to _run_ the JS, just not to actually do the XHR.

tgrathwell: fantastic

i have revised my added commit to just be the deletey parts

PR #3225, opened 2022-10-27
Client portal: check EITC eligibility if remove all qualifying dependents

embarnard:


embarnard:
This change will include EITC qualification checks when removing dependents on the client portal and display which credits the client is no longer eligible for on `en/portal/dependents/not-eligible`. Currently this page would only show disqualifications for CTC.

Screen Shot 2022-10-27 at 5 32 45 PM
Screen Shot 2022-10-27 at 5 33 09 PM

app/controllers/ctc/portal/dependents_controller.rb
```diff
@@ -18,12 +18,8 @@ def destroy
...
       client: current_client
     )
-    ctc_eligibility = Efile::DependentEligibility::ChildTaxCredit
-    if current_intake.dependents.filter { |d| ctc_eligibility.new(d, TaxReturn.current_tax_year).qualifies? }.length.zero?
-      redirect_to not_eligible_ctc_portal_dependents_path
-    else
-      redirect_to Ctc::Portal::PortalController.to_path_helper(action: :edit_info)
-    end
+
+    redirect_to not_eligible_ctc_portal_dependents_path
```
embarnard, line 22: maybe this is hasty but only want to do the check in one place if possible

config/locales/es.yml
```diff
@@ -1819,8 +1819,9 @@ es:
           title: Edita la información básica de %{name}
         dependents:
           no_eligible_dependents:
-            subtitle: No recibirá el Crédito Tributario por Hijos porque no tiene dependientes elegibles.
-            title: 'Usted ha eliminado a todos sus dependientes elegibles para CTC. '
+            credits: CTC y EITC
+            subtitle: No recibirá %{ineligible_credits} porque no tiene dependientes elegibles.
```
alaroia-cfa, line 1823: Does this string interpolation make sense in Spanish? I'm concerned it doesn't, and am unsure.

spec/controllers/ctc/portal/pages_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  before do
+    allow(subject).to receive(:open_for_eitc_intake?).and_return true
+    sign_in intake.client
+  end
+
+  context "#no_eligible_dependents" do
+    it "renders no_eligible_dependents template" do
+      get :no_eligible_dependents
+      expect(response).to render_template :no_eligible_dependents
```
alaroia-cfa, line 27: Rendering a template with the same name as the controller action is something we generally trust Rails to do. Given that, I'd say it's better to check that the response is OK in this test. A redirect is a 302/301, not 200 (OK), after all.

config/locales/es.yml
alaroia-cfa: Nice `credits` thing

PR #3228, opened 2022-10-28
Update income question #183631734

mquadri23:


app/views/ctc/questions/income_qualifier/edit.html.erb
```diff
@@ -0,0 +1,33 @@
+<% @filing_count = current_intake.default_tax_return.filing_status_single? ? 1 : 2 %>
+<% standard_deduction = StandardDeductions.base_deductions(tax_year: current_tax_year, puerto_rico_filing: current_intake.puerto_rico_filing?)[current_intake.default_tax_return.filing_status] %>
```
embarnard, line 2: I almost wonder if this should become an instance variable and put in the controller

embarnard (approved):
Looks good! Could make that one change but not a blocker

PR #3232, opened 2022-10-28
ClientSorter references the new `filterable_tax_return_*` columns

tgrathwell:


app/lib/client_sorter.rb
```diff
@@ -32,22 +32,52 @@ def filtered_and_sorted_clients(default_order: nil)
...
+        tax_return_filters_expanded << special_tax_return_filters.merge(assigned_user_id: id)
+      end
+    elsif current_user&.greeter?
+      tax_return_filters_expanded = [tax_return_filters.merge(greetable: true), tax_return_filters.merge(assigned_user_id: current_user.id)]
+    else
+      tax_return_filters_expanded = [tax_return_filters]
+    end
+
+    clients = tax_return_filters_expanded.map do |tax_return_filters|
+      clients.where("filterable_tax_return_properties @> '[#{tax_return_filters.to_json}]'::jsonb")
```
alaroia-cfa, line 75: I think we can use ActiveRecord `?` here safely and effectively. Try instead:

```
clients.where("filterable_tax_return_properties @> [?]::jsonb", tax_return_filters.to_json)
```

I think this works OK based on the following testing:

rails c

```
irb(main):014:0> puts AdminRole.where('primary_first_name @> [?]::jsonb', {hi: "there's"}.to_json).select('
primary_first_name').to_sql
SELECT primary_first_name FROM "admin_roles" WHERE (primary_first_name @> ['{"hi":"there''s"}']::jsonb)
=> nil

```

psql

```
vita-min_development=# select '{"hi":"there''s"}'::jsonb as ok from intakes limit 1;
        ok         
-------------------
 {"hi": "there's"}
(1 row)

```

It looks like `''''` is a single apostrophe in SQL:

```
vita-min_development=# select '''' as ok from intakes limit 1;
 ok 
----
 '
(1 row)

```

It looks like without the `?`, one does run into problems.

```
irb(main):016:0> val = {hi: "there's"} ; AdminRole.where("primary_first_name @> '[#{val.to_json}]'::jsonb")

  AdminRole Load (2.8ms)  SELECT "admin_roles"."id", "admin_roles"."created_at", "admin_roles"."updated_at" FROM "admin_roles" WHERE (primary_first_name @> '[{"hi":"there's"}]'::jsonb) /* loading for inspect */ LIMIT $1  [["LIMIT", 11]]
Traceback (most recent call last):
ActiveRecord::StatementInvalid (PG::SyntaxError: ERROR:  syntax error at or near "s")
LINE 1: ...oles" WHERE (primary_first_name @> '[{"hi":"there's"}]'::jso...
                                                             ^
```

Happy to Tuple or not. Let me know what you think.

alaroia-cfa: Rather

```
clients.where("filterable_tax_return_properties @> '[?]'::jsonb", tax_return_filters.to_json)
```

alaroia-cfa: Hmm I might be wrong about where it's OK to put these `'`s

PR #3233, opened 2022-10-31
[CTC] Update the FAQ footer link

mquadri23:


app/views/shared/_footer.html.erb
```diff
@@ -26,11 +26,12 @@
         </li>
 
         <li>
-          <%= link_to faq_path do %>
-            <%= t("general.faq") %>
-          <% end %>
+          <%= link_to variation == "ctc"? "https://www.getyourrefund.org/en/faq/child_tax_credit" : faq_path do %>
```
embarnard, line 29: There should be a space between the question mark and the previous character so that it looks like variation == "ctc" ? "https://www.getyourrefund.org/en/faq/child_tax_credit" : faq_path

PR #3244, opened 2022-11-08
Add delayed job web ui

wschaefer:
Anyone can see it on dev/heroku but on prod/demo you have to have to be logged in as an admin with an @codeforamerica.org email address


config/initializers/delayed_job.rb
```diff
@@ -19,3 +19,12 @@ class AddExtraJobAttributes < Plugin
 end
 
 Delayed::Worker.plugins << Delayed::Plugins::AddExtraJobAttributes
+
+class CanAccessDelayedJobWeb
+  def self.matches?(request)
+    return true if Rails.env.development? || Rails.env.heroku?
+
+    current_user = request.env['warden'].user
+    current_user.present? && current_user.admin? && current_user.email.include?("@codeforamerica.org")
```
tgrathwell, line 28: we're thinking maybe instead of checking the email there should be an extra bit somewhere in the database specifying who is an engineer. maybe on AdminRole ?

PR #3246, opened 2022-11-09
Add tests for bulk signup messaging for CtcSignup records [#183711250]

alaroia-cfa:


app/controllers/hub/signup_selections_controller.rb
```diff
@@ -28,7 +28,8 @@ def create
           @signup_selection.errors.add :upload, "Invalid CSV"
           render :index and return
         end
-      validated_ids = Signup.where(id: id_array).pluck(:id)
+      signup_model = create_params[:signup_type] == "GYR" ? Signup : CtcSignup
```
tgrathwell, line 31: oh whoops. context fail on my part

alaroia-cfa: I prolly could have left a test supporting this behavior in the code as part of handoff :)

PR #3250, opened 2022-11-09
Change full time student EITC qualifier to less_than_five_months

jenny-heath:
- change copy to ask the right question
- change name of column to reflect actual rule
  - create a new column
  - write to both columns
  - add backfill rake task


jenny-heath:
closing to retry heroku

PR #3254, opened 2022-11-10
Client#filterable_tax_return_properties contains active, stage and greetable

tgrathwell:
it's easier to query on this jsonb field if there's single bits rather than arrays of different possible stages


alaroia-cfa (approved):
yay.json

PR #3272, opened 2022-11-16
Add notification flag for menu items side bar

nishinelson:

### DIFF
--- a/app/views/hub/components/_menu_item.html.erb
+++ b/app/views/hub/components/_menu_item.html.erb
@@ -2,5 +2,8 @@
 <a class="menu-item" href="<%= link %>"
    <% if link_method.present? %>data-method="<%= link_method %>" <% end %>>
   <%= image_tag(icon_path) %>
-  <div><%= item_label %></div>
+  <div>
+    <p><%= item_label %></p>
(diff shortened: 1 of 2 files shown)


app/views/hub/components/_menu_item.html.erb
```diff
@@ -2,5 +2,8 @@
 <a class="menu-item" href="<%= link %>"
    <% if link_method.present? %>data-method="<%= link_method %>" <% end %>>
   <%= image_tag(icon_path) %>
-  <div><%= item_label %></div>
+  <div>
+    <p><%= item_label %></p>
+    <%= render 'shared/unread_icon', flag: true if current_user.notifications.pluck(:read).any?(false) && item_label == "My Updates" %>
```
tgrathwell, line 7: this statement makes an array of `true` or `false` values in memory and then checks it for any `false` ones

we have some users with 7000+ notifications, so it will make an array of `[false, false, false...]` 7000 times

it's not unacceptably slow, but it's 5x faster to let the database do the work

`current_user.notifications.where(read: false).any?`
which SQLs to
```SQL
SELECT 1 AS one FROM "user_notifications"
WHERE "user_notifications"."user_id" = 12345 AND "user_notifications"."read" = false
LIMIT 1
```

PR #3274, opened 2022-11-17
Update pdftk java

alaroia-cfa:
## Approach taken

Create a `vendor/pdftk` directory which contains a specific version of `pdftk` whose version is in the path. This allows us to easily download/experiment with other versions, which sometimes is useful when trying to debug a pdftk problem.

Rely on this for dev & Heroku & Circle & Aptible, rather than using slightly different approaches in each.

Include the test in #3277 that shows the necessity of the upgrade.

## Testing performed

```
rails c
system('pdftk -version')
```

shows 3.3.3 on:

- ✅ local dev
- ✅ heroku review app
- Aptible (testing now)

## Other approaches considered

- Considered using the pdftk 3.3.2 package from Ubuntu. It can't be used on Debian.
- Considered using the pdftk 3.3.2 package from Debian. This probably would work, but would make it non-obvious how to upgrade past that if we want to in the future. Plus when I tested this out, I thought the tests were failing because of the pdftk upgrade but in fact they were failing due to the issue fixed in #3275 .
- Considered writing a pdftk downloader in Ruby/Rake but it wouldn't probably be cached nicely in the Docker build.


Brewfile
```diff
@@ -1,7 +1,7 @@
 brew "gh"
 brew "ghostscript"
 brew "imagemagick"
-brew "pdftk-java"
+brew "java"
```
jenny-heath, line 4: our `gyr_efiler_spec` has previously been suggesting you install Java like this:
`brew tap AdoptOpenJDK/openjdk && brew install adoptopenjdk8`

is `brew "java"` better? worse? the same?

jenny-heath: never mind i see the old one is just for the efiler

jenny-heath (approved):

alaroia-cfa:
@jenny-heath You had asked somewhere about `java` vs `adoptopenjdk8`. I tested pdftk-java this morning with Java 8, and it seems fine.

I'll make a follow-up PR with that change.

PR #3276, opened 2022-11-17
Sort clients by the percentage of required documents

jalcine-cfa:


jalcine-cfa:
The failing test on this PR for the commit https://github.com/codeforamerica/vita-min/commit/6417e771545c6acf307dd1fdd015553fb7360338 looks like to https://github.com/codeforamerica/vita-min/blob/3c5137700c742b52e94206ac90ca135065ac1fff/spec/controllers/hub/clients_controller_spec.rb#L799-L808 around `child_index` and its value. I could play with this locally until it works.

jalcine-cfa:
Never mind [this](https://github.com/codeforamerica/vita-min/pull/3276#issuecomment-1329485038), the actual issue seems like the stubs are not encouraging the expected text values on the screen.

PR #3279, opened 2022-11-18
Flow explorer: show GYR dependents controller at the appropriate spot

tgrathwell:
it's not in the actual navigation order so we have to splice it in

also includes some improvements to deal with controllers that don't just have an #edit (which is just this one, I guess)


tgrathwell:
lgtm say asheesh

PR #3289, opened 2022-12-01
update portal/login copy

mquadri23:


spec/features/portal/client_login_spec.rb
```diff
@@ -156,7 +156,7 @@
       scenario "getting redirected to the page client was trying to access after login" do
         visit portal_tax_return_authorize_signature_path(locale: "es", tax_return_id: tax_return.id)
 
-        expect(page).to have_text "Le mandaremos un código seguro para que pueda ver su progreso."
+        expect(page).to have_text "Te enviaremos un código seguro para iniciar sesión"
```
wschaefer, line 159: I like the way you updated all the other tests to use the I18n key so the tests wouldn't have to change the next time we change this text! Do you want to do this one too?

wschaefer (approved):
looks good!

PR #3306, opened 2022-12-15
Remove 'CTC Capacity' admin tool and associated model

tgrathwell:


alaroia-cfa (approved):
yum yum yum trash monster eating some code

we should confirm with yvonne but I bet she'll be like yum yum yum trash monster

PR #3309, opened 2022-12-19
Update Answers about Visa to Citizenship

jalcine-cfa:
- Changed copy and translation ID of self status
- Changed copy and translation ID of spouse status


app/views/questions/spouse_life_situations/edit.html.erb
```diff
@@ -12,7 +12,7 @@
       <%= f.cfa_checkbox(:spouse_had_disability, t('views.questions.spouse_life_situations.options.had_disability'), options: { checked_value: "yes", unchecked_value: "no" }) %>
       <%= f.cfa_checkbox(:spouse_was_blind, t('views.questions.spouse_life_situations.options.was_blind'), options: { checked_value: "yes", unchecked_value: "no" }) %>
       <%= f.cfa_checkbox(:spouse_was_full_time_student, t('views.questions.spouse_life_situations.options.was_full_time_student'), options: { checked_value: "yes", unchecked_value: "no" }) %>
-      <%= f.cfa_checkbox(:spouse_was_on_visa, t('views.questions.spouse_life_situations.options.was_on_visa'), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:spouse_was_on_visa, t('views.questions.spouse_life_situations.options.was_citizen'), options: { checked_value: "yes", unchecked_value: "no" }) %>
```
jalcine-cfa, line 15: Should we be changing `:spouse_was_on_visa` here to be a different answer? Or is this the needed value (despite the text)?

app/views/questions/life_situations/edit.html.erb
```diff
@@ -12,7 +12,7 @@
       <%= f.cfa_checkbox(:had_disability, t('views.questions.life_situations.options.had_disability'), options: { checked_value: "yes", unchecked_value: "no" }) %>
       <%= f.cfa_checkbox(:was_blind, t('views.questions.life_situations.options.was_blind'), options: { checked_value: "yes", unchecked_value: "no" }) %>
       <%= f.cfa_checkbox(:was_full_time_student, t('views.questions.life_situations.options.was_full_time_student'), options: { checked_value: "yes", unchecked_value: "no" }) %>
-      <%= f.cfa_checkbox(:was_on_visa, t('views.questions.life_situations.options.was_on_visa'), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:was_on_visa, t('views.questions.life_situations.options.was_citizen'), options: { checked_value: "yes", unchecked_value: "no" }) %>
```
jalcine-cfa, line 15: Same here as https://github.com/codeforamerica/vita-min/pull/3309/files#r1052381464

app/views/questions/spouse_life_situations/edit.html.erb
jalcine-cfa: Actually so https://www.pivotaltracker.com/n/projects/2409240 answers that, it seems! But only partially - is the family of `_was_on_visa` meant (for our use) to show this text now?

alaroia-cfa: > Should we be changing :spouse_was_on_visa here to be a different answer? Or is this the needed value (despite the text)?

Good question! We should probably store this in a new column on `intakes`.

Some considerations/clarifications:
* The first param `:spouse_was_on_visa` is used as the HTML FORM field name generated by `f.cfa_checkbox(:spouse_was_on_visa, ...)` and it's good for HTML FORM field name to accurately describe what is being asked.
* In this PR, we're basically reversing the meaning of 'yes' and 'no' for this field. This will be important when reading the data out in order to fill out the Form 13614-C PDF.
* During the tax filing season, when there are production GYR users, we need to be thoughtful to retain any existing data. Right now, there's no prod GYR users, so it'd be peaceful to create a new column called e.g. `spouse_was_citizen`.

For those reasons, I think this PR should include a migration to add a new field called e.g. `spouse_was_citizen`. The same logic applies to the primary filer as well.

Happy to Tuple or otherwise discuss more; wanted to give this a review since I saw it.

alaroia-cfa (dismissed):
Good start! I think we should change the name of the column we store this in, which will also likely mean changes to the Form 13614-C PDF generation.

PR #3311, opened 2022-12-20
Remove RRC and Advance CTC questions

embarnard:


alaroia-cfa (commented):
This PR removes the advance CTC questions from GetCTC, but the story only asks for changes within GYR. Is it better to be more surgical and only remove the questions from GYR? I think so, but I wonder what you think.

Overall I think it is probably smart to delete the GetCTC intake questions soon, but I don't personally want to see that happen piecemeal. I think it makes sense to do that all at once when we do it, and leave them untouched until we do.

embarnard:
> This PR removes the advance CTC questions from GetCTC, but the story only asks for changes within GYR. Is it better to be more surgical and only remove the questions from GYR? I think so, but I wonder what you think.
> 
> Overall I think it is probably smart to delete the GetCTC intake questions soon, but I don't personally want to see that happen piecemeal. I think it makes sense to do that all at once when we do it, and leave them untouched until we do.

Oh I guess I got confused because I found clear referencing of RRC and Advance CTC questions in GetCTC and I haven't been working on 13614-c work so I didn't think about how that is only in GYR. Yeah don't think we should piecemeal it.

embarnard:
> This PR removes the advance CTC questions from GetCTC, but the story only asks for changes within GYR. Is it better to be more surgical and only remove the questions from GYR? I think so, but I wonder what you think.
> 
> Overall I think it is probably smart to delete the GetCTC intake questions soon, but I don't personally want to see that happen piecemeal. I think it makes sense to do that all at once when we do it, and leave them untouched until we do.

So instead we would only remove it from the ArpPayments pages, right?

alaroia-cfa:
> So instead we would only remove it from the ArpPayments pages, right?

Yeah, I think so. That's the only reference I could find based on my own searching plus reading this pull request.

embarnard:
> > So instead we would only remove it from the ArpPayments pages, right?
> 
> Yeah, I think so. That's the only reference I could find based on my own searching plus reading this pull request.

Sorry for the confusion! I think this PR should accomplishment that then: https://github.com/codeforamerica/vita-min/pull/3312

PR #3314, opened 2022-12-22
Loader-utils dependency upgrade

embarnard:

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1723,10 +1723,10 @@
   resolved "https://registry.yarnpkg.com/@xtuc/long/-/long-4.2.2.tgz#d291c6a4e97989b5c61d9acf396ae4fe133a718d"
   integrity sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==
 
-"@yaireo/tagify@^4.9.8":
-  version "4.17.6"
-  resolved "https://registry.yarnpkg.com/@yaireo/tagify/-/tagify-4.17.6.tgz#505e8d5e8fb0421c36870ac29c551c488ca8693b"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -2017,24 +1724,16 @@
   integrity sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==
 
 "@yaireo/tagify@^4.9.8":
-  version "4.9.8"
-  resolved "https://registry.yarnpkg.com/@yaireo/tagify/-/tagify-4.9.8.tgz#67082ecca81fbbec4d75bb95581280f27af22fba"
-  integrity sha512-JqRTN4vb5tx06PJIRIOAlriO/B6u+a3jD/VBb5x6E1MOlC4wrMfaLHmNnrMpiZCHMXox7pVmrtNuyBe4YReJ4w==
+  version "4.17.6"
```
embarnard, line 1727: Prolly good to read their release notes for the versions between 4.9.8 and 4.17.6 to see if we have to change anything in our code (or at least manually test the tagify functionality) (or revert this hunk of the diff?)

embarnard:
1. Ran `yarn upgrade --recursive loader-utils@^2.0.4` to update the loader-utils without adding it to package.json
2. Then looked through the diff to check using `git diff origin/main...HEAD -- yarn.lock ` and then in the diff mode `/^[+][^ ]` to search for lines that started with + (unindented lines) to see just the changes of package versions
3. Look for package version changes that are in package json because we are concerned about breaks in packages we intentionally added -- Open RubyMine to package.json and use `n` to go forward in the pager (`N` (shift and n) to go back) to find lines that correspond to package.json
4. Through this we found tagify's version got changed, and it was simple enough to pin it to the version we had in the past since the new version caused a build problem `yarn upgrade @yaireo/tagify@4.9.8`
5. Through this we found percy's version got changed, and we ran `bin/percy` and things seemed OK
6. Saw there are no others (also gained a mild familiarity with our JS dependencies)

PR #3316, opened 2023-01-03
Update session toggle GYR launch dates

mquadri23:


embarnard (changes requested):
Looks good! Would change that one test

config/environments/heroku.rb
```diff
@@ -1,4 +1,4 @@
-require Rails.root.join('config/environments/demo')
+require_relative "./shared_deployment_config"
```
embarnard, line 1: not sure what this is doing?

PR #3317, opened 2023-01-04
Updated auto reply to archived clients. SMS has not been updated per t…

nishinelson:
Updated auto reply to archived clients. 


embarnard (approved):
Looks good! I would update the commit message before squashing and merging so that is complete and in present tense. Github automatically cuts off commit message after a certain number of characters (140 or so) so sometimes you have to copy it back in

PR #3318, opened 2023-01-04
Add additional fields to 13614c page 2

wschaefer:
This PR:
- adds new fields from page 2 of the 13614c to the db
- adds them to the view and form
- fills them out in the PDF

note: some of the fields we thought were going to be new actually existed so this takes care of adding them to the view too


spec/lib/f13614c_pdf_spec.rb
```diff
@@ -117,6 +129,12 @@
           eip1_amount_received: 500,
           eip2_amount_received: 1500,
           eip3_amount_received: 2500,
+          # paid_misc_expenses: "unfilled",
+          # TODO: uh oh we might already have these
+          # misc_expenses_medical_and_dental: "unfilled",
+          # misc_expenses_mortgage_interest: "unfilled",
+          # misc_expenses_taxes: "unfilled",
+          # misc_expenses_charitable_contributions: "unfilled",
```
jenny-heath, line 137: we added these fields on this branch but i think they might already exist :(
see https://github.com/codeforamerica/vita-min/blob/main/app/lib/f13614c_pdf.rb#L94-L97

PR #3320, opened 2023-01-04
Fix hub create GYR client/drop off page

embarnard:


embarnard:
The drop off/create client page was broken because we are using current year to determine the name of the checkbox for tax returns but we were not setting the @current_year in the create method in the client controller. And also the `create_tax_return_for_year?(year)` method was returning true no matter what if the current year was selected

jenny-heath (approved):

PR #3323, opened 2023-01-09
Archiving: add product_year to intakes for new archiving strategy

wschaefer:
we query using product_year in the ClientSorter and when checking for dupes with the 'accessible_intakes' scope

### DIFF
--- a/app/lib/client_sorter.rb
+++ b/app/lib/client_sorter.rb
@@ -1,7 +1,7 @@
 class ClientSorter
   QUICK_FILTERS = [
-    [{last_contact: "approaching_sla", active_returns: true}, "Approaching SLA"],
-    [{last_contact: "breached_sla", active_returns: true}, "Breached SLA"]
+    [{ last_contact: "approaching_sla", active_returns: true }, "Approaching SLA"],
+    [{ last_contact: "breached_sla", active_returns: true }, "Breached SLA"]
   ]
 
   attr_reader :current_user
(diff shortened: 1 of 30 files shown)


app/lib/client_sorter.rb
```diff
@@ -36,8 +37,10 @@ def filtered_clients
               else
                 @clients.after_consent
               end
-    # Force an inner join to `intakes` to exclude clients from previous years
-    clients = clients.joins(:intake)
+    # Filter on product_year to only show clients who used this-year's product
+    if @use_product_year
+      clients = clients.where(intake: Intake.where(type: "Intake::CtcIntake", product_year: MultiTenantService.new(:ctc).current_product_year).or(Intake.where(type: "Intake::GyrIntake", product_year: MultiTenantService.new(:gyr).current_product_year)))
```
wschaefer, line 42: remove type, just look for product year (should there be a rails.configuration.product_year ?)

wschaefer: also this should use `filterable_product_year` from client which should be the same value as intakes.product_year

db/migrate/20221219230651_add_product_year_to_intakes.rb
```diff
@@ -0,0 +1,7 @@
+class AddProductYearToIntakes < ActiveRecord::Migration[7.0]
+  def change
+    # Safe enough for migrations since in Dec 2022 there aren't a lot of reads & writes on the web app.
+    add_column :intakes, :product_year, :integer, null: false, default: 2022
```
wschaefer, line 4: also need to add filterable_product_year and assure it gets denormalized

and maybe some indices on all these sortable client attributes, so (filterable_product_year, last_contacted_at_or_whatever)

PR #3324, opened 2023-01-10
Archiving: set product year to 2023

tgrathwell:
Co-authored-by: Whitney Schaefer <wschaefer@codeforamerica.org>


tgrathwell:
asheesh is worried that we still need to support login for 2022 GetCTC clients, this change will make it so that is not possible

PR #3325, opened 2023-01-11
Use current tax year in GYR homepage's file myself subtitle

embarnard:


app/views/shared/_service_comparison.html.erb
```diff
@@ -35,7 +35,7 @@
                 </h2>
               </div>
               <div class="service__description">
-                <%= t("views.shared.service_comparison.services.diy.service_description") %>
+                <%= t("views.shared.service_comparison.services.diy.service_description", current_tax_year: MultiTenantService.new(:gyr).current_tax_year) %>
```
alaroia-cfa, line 38: Non-blocking comment: You could shorten this by making the interpolation variable be called `year`, like:

```
<%= t("views.shared.service_comparison.services.diy.service_description", year: MultiTenantService.new(:gyr).current_tax_year) %>
```

I prefer it when things are slightly shorter. It's not a blocker; fine either way.

alaroia-cfa (approved):
Looks good to me! One non-blocking comment. If you make that change and the tests pass, no need for re-review.

PR #3337, opened 2023-01-18
Update 13614c select fields from long to short versions: Y/N/?

nishinelson:


nishinelson:
Decided with Tiffany that this is no longer needed.

PR #3342, opened 2023-01-19
Update taxslayer link. Deleted getctc related taxslayer link and test.

nishinelson:


embarnard (changes requested):
I would only change the link for GYR, and maybe do a check for where else we show this link to make sure we are unintentionally changing it where we don't want to. Also I would remove those schema changes. But otherwise looks good!

embarnard (approved):

PR #3362, opened 2023-01-25
Import new FAQ copy

jenny-heath:
Trying to avoid copying in individual strings from [here](https://docs.google.com/spreadsheets/d/1gQ_VbXlMxuJL1H2tZZ5WS2ow4eNs6wL8vuLT6fgeAMI/edit#gid=1052833239)

Steps to recreate what I have so far:
- download the csv from google
- open a rails console and do this:
```
faq_csv = File.read(<wherever you downloaded the google sheet>)
result = FaqCsvImportJob.perform_now(faq_csv)
File.open('/tmp/test_faq.yml', 'w') {|f| f.write result.to_yaml }
```
- open the temp file and copy the whole thing minues the "---" at the beginning
- replace everything under `en.views.public_pages.faq.question_groups` with what you copied
- run `i18n-tasks normalize`

I think this may be far from done and here's what's left:
- review a little more thoroughly to see that the content changes are correct
- do the same for spanish
- fix broken tests
- figure out if we can avoid any edits to the content that not changing—this mainly means formatting stuff. the way it's written now it will collapse any prettily written html in the yaml to be one long grotesque string. maybe we can find a way to edit the existing yaml instead of generating a whole new block


app/jobs/faq_csv_import_job.rb
```diff
@@ -3,38 +3,35 @@
...
-  def perform(faq_csv_content, write: false)
+class FaqCsvImportJob
+  def self.perform_now(faq_csv_content, write: false)
     [:en, :es].each do |locale|
       filename = "config/locales/#{locale}.yml"
-      puts 'parse_results'
-      # puts self.class.parse(faq_csv_content, locale).to_json
-      new_translations = self.class.updated_translations(
-        YAML.safe_load(File.read("config/locales/#{locale}.yml")).with_indifferent_access,
+      translation_data = YAML.load_file("config/locales/#{locale}.yml")
```
jenny-heath, line 10: maybe we should use the old YAML method? can check

alaroia-cfa: cool

PR #3365, opened 2023-01-25
Add Automated Message to Hub

jalcine-cfa:
Co-authored-by: Nishi Nelson <nnelson@codeforamerica.org>


app/models/survey_messages/gyr_completion_survey.rb
```diff
@@ -1,50 +0,0 @@
-module SurveyMessages
```
jalcine-cfa, line 1: Should this have been deleted? I can't remember

PR #3367, opened 2023-01-25
Add Automated Message to Hub for Newcomers  #184300083

jalcine-cfa:


app/jobs/send_client_in_progress_message_job.rb
```diff
@@ -0,0 +1,8 @@
+class SendClientInProgressMessageJob < ApplicationJob
+  def perform(client)
+    SurveySender.send_survey(
```
tgrathwell, line 3: I guess if we wanted to send the message to both contact methods (email+sms) when relevant this could instead be something more like
```
ClientMessagingService.send_system_message_to_all_opted_in_contact_methods(
  client: client,
  message: AutomatedMessage::InProgress
)
```

which would also quell some of the dissonance of this being a "SurveySender"

makes me wonder if the `SurveySender` should be deprecated in favor of making a `ClientMessagingService.send_system_message_to_best_opted_in_contact_method` or something... but we don't have to do that now...

tgrathwell (approved):

app/jobs/send_client_in_progress_message_job.rb
jalcine-cfa: Good point re: making that method. I think parts of the survey sender could be used (like building that common payload for surveys to pull out their links)

jalcine-cfa: s/could be used/can be kept/

jalcine-cfa:
Going to merge this, if nothing works in demo then I'll roll these changes back.

PR #3387, opened 2023-02-02
Hub: show archived banner for intakes using the new archiving method

jenny-heath:
Would like opinions on how much more test coverage to add if any. There are a few different behaviors for archived intakes but not all of them have a test that covers non-archived intake + new archiving method intake + old archiving method intake.


jenny-heath:
idea: add new-type archived intake to seeds

PR #3399, opened 2023-02-08
Try to replace our OrganizationCapacity view with a with_capacity scope on Organization

tgrathwell:
We're using the activerecord-cte gem, when Rails 7.1 is released it can be removed because the very same '.with' syntax will be in rails

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -31,6 +31,8 @@
 class Organization < VitaPartner
   TYPE = "Organization"
 
+  attribute :active_client_count
+
   belongs_to :coalition, optional: true
   has_one :organization_capacity, foreign_key: "vita_partner_id"
   has_many :child_sites, -> { order(:id) }, class_name: "Site", foreign_key: "parent_organization_id"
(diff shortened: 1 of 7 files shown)


app/models/organization.rb
```diff
@@ -43,6 +45,24 @@ class Organization < VitaPartner
 
   default_scope -> { includes(:child_sites).order(name: :asc) }
   alias_attribute :allows_greeters?, :allows_greeters
+  scope :with_capacity, -> do
+    not_these_states = ['intake_before_consent', 'intake_in_progress', 'intake_greeter_info_requested', 'intake_needs_doc_help', 'file_mailed', 'file_accepted', 'file_not_filing', 'file_hold', 'file_fraud_hold']
```
alaroia-cfa, line 49: Consider using `TaxReturnStateMachine::EXCLUDED_FROM_CAPACITY`

```
irb(main):001:0> TaxReturnStateMachine::EXCLUDED_FROM_CAPACITY
=> [:intake_before_consent, :intake_in_progress, :intake_greeter_info_requested, :intake_needs_doc_help, :file_mailed, :file_accepted, :file_not_filing, :file_hold, :file_fraud_hold]
```

alaroia-cfa (commented):
This is a cool PR. I left a comment.

app/models/organization.rb
tgrathwell: i was hoping something like that existed and i'm glad i could just push it up the dumb way and let market forces sort it out

PR #3400, opened 2023-02-08
Send Mixpanel events for outbound links and form submission

jalcine-cfa:

### DIFF
--- a/app/javascript/lib/mixpanel_event_tracking.js
+++ b/app/javascript/lib/mixpanel_event_tracking.js
@@ -20,9 +20,11 @@ const MixpanelEventTracking = (function () {
 
           for (const key in clickedElement.dataset) {
             const attributePrefix = "trackAttribute";
+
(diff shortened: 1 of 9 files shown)


app/javascript/lib/mixpanel_event_tracking.js
```diff
@@ -0,0 +1,52 @@
+import "../vendor/navigator.sendbeacon.min.js";
+
+
+const MixpanelEventTracking = (function () {
+    const addClickTrackingToOutboundLinks = function () {
+      const links = document.querySelectorAll("a[href]:not([href^='/']):not([href^='#']):not([href^='" + location.protocol + "//" + location.host + "'])");
```
jalcine-cfa, line 6: ✨  praise be to the power of [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) and [negative selection](https://css-tricks.com/almanac/selectors/n/not/) ✨ !

PR #3401, opened 2023-02-09
Twilio sms: retry sending outgoing messages that fail due to queue overflow

jenny-heath:


app/controllers/twilio_webhooks_controller.rb
```diff
@@ -6,7 +6,13 @@ class TwilioWebhooksController < ActionController::Base
   def update_outgoing_text_message
     status = params["MessageStatus"]
     DatadogApi.increment("twilio.outgoing_text_messages.updated.status.#{status}")
-    OutgoingTextMessage.find(params[:id]).update_status_if_further(status)
+    message = OutgoingTextMessage.find(params[:id])
+    message.update_status_if_further(status)
+
+    if params["ErrorCode"] == "30001" && message.number_of_retries < 3
+      message.update(number_of_retries: message.number_of_retries + 1)
+      SendOutgoingTextMessageJob.perform_later(message.id)
```
alaroia-cfa, line 14: We should introduce a backoff + jitter to the job start time to avoid a situation where delaying it repeats the same queued up problem. Something like:

```ruby
jitter_time = SecureRandom.rand(30) # to reduce the chance of multiple jobs running on top of each other
SendOutgoingextMessageJob.set(wait_until: (5.minutes ** message.number_of_retries) + jitter_time).perform_later(message.id)
```

An example of this can be found at https://github.com/codeforamerica/vita-min/blob/f49178e7d930d785edab292ed93ebed8081b301b/app/models/efile_submission.rb#L194-L202

alaroia-cfa (commented):
One blocking comment about adding more delayed time when queuing up a job for outgoing SMS!

co-written by JA

PR #3415, opened 2023-02-17
Drop excess events to Mixpanel (184443398)

jalcine-cfa:
- Add stubbed tests to check constraints
- Add tests to check for security metrics + AWS addresses

### DIFF
--- a/app/services/mixpanel_service.rb
+++ b/app/services/mixpanel_service.rb
@@ -107,7 +107,7 @@ class MixpanelService
                    request: nil,
                    source: nil,
                    path_exclusions: [])
-      return if request.present? and should_event_be_dropped?(request) != :ok
+      return if should_event_be_dropped?(request)
 
       default_data = {}
       default_data[:locale] = I18n.locale.to_s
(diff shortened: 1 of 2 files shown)


spec/services/mixpanel_service_spec.rb
```diff
@@ -919,5 +921,53 @@ def inst_test
...
+
+    context "when dropping events" do
+      before do
+        routes.draw { get "index" => "anonymous#index" }
+      end
+
+      context "for those not allowed" do
+        def self.test_security_metrics_ips
+          security_metrics_subnet = IPAddr.new("192.211.152.0/24") # 255 IP addresses in a Class C subnet
+          ip_addresses = security_metrics_subnet.to_range
```
jalcine-cfa, line 933: This is a bit blunt of an approach because it's testing _every_ IP address but it's a peace of mind that I'm comfortable with. Advise!

app/services/mixpanel_service.rb
```diff
@@ -121,6 +123,22 @@ def send_event(distinct_id:,
       )
     end
 
+    SECURITY_METRICS_SUBNET = IPAddr.new("192.211.152.0/24").freeze
+    EC2_COMPUTE_DOMAIN_REGEX = /^ec2-(.*)\.compute-1\.amazonaws.com$/i.freeze
+
+    def should_event_be_dropped?(request)
+      return :hund if request.query_parameters[:source]
```
tgrathwell, line 130: it might be better for `should_event_be_dropped?` to return true/false instead of a symbol

also this line seems off -- should probably be `return true if request.query_parameters[:source] == 'hund'`

spec/services/mixpanel_service_spec.rb
```diff
@@ -919,5 +920,53 @@ def inst_test
...
+        it "drops events coming from non-public AWS domains" do
+          request.set_header("HTTP_HOST", "ec2-18-204-251-64.compute-1.amazonaws.com")
+
+          get :index
+
+          expect(fake_tracker).not_to have_received(:track)
+        end
+
+        it "drops events coming from status checks" do
+          get :index, params: { source: :hund }
```
tgrathwell, line 957: I think in reality the source would come in as a string rather than a symbol so the test should probably use a string

spec/services/mixpanel_service_spec.rb
tgrathwell: I'd be happier just having one example of an IP in the subnet which gets filtered, and one example of an IP outside of the subnet where the mixpanel event is still sent. Testing every IP in the subnet doesn't tell us anything new, just whether `IpAddr.new ... include?` is working correctly

jalcine-cfa: Gotcha! That makes sense to me

app/services/mixpanel_service.rb
jalcine-cfa: Indeed, not sure how the tests missed that one

PR #3425, opened 2023-02-22
Intercom bad-request bug

embarnard:
IntercomService's create messages has flaky `Intercom::BadRequestError (No conversation for this user)` errors when trying to reply to clients' last message even though we are able to find the last message.

### DIFF
--- a/app/services/intercom_service.rb
+++ b/app/services/intercom_service.rb
@@ -8,14 +8,15 @@ class IntercomService
     return nil if body.blank?
 
     contact = contact_from_client(client) || contact_from_email(email_address) || contact_from_sms(phone_number)
+    most_recent_convo = most_recent_conversation(contact.id)
 
-    if contact.present? && most_recent_conversation(contact.id).present?
+    if contact.present? && most_recent_convo.present?
       # Per https://developers.intercom.com/intercom-api-reference/reference/reply-to-a-conversation
       # type is always user and message_type is always comment.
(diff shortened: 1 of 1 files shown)


app/services/intercom_service.rb
```diff
@@ -8,14 +8,15 @@ def self.create_message(email_address:, phone_number:, body:, client:, has_docum
...
     contact = contact_from_client(client) || contact_from_email(email_address) || contact_from_sms(phone_number)
+    most_recent_convo = most_recent_conversation(contact.id)
 
-    if contact.present? && most_recent_conversation(contact.id).present?
+    if contact.present? && most_recent_convo.present?
       # Per https://developers.intercom.com/intercom-api-reference/reference/reply-to-a-conversation
       # type is always user and message_type is always comment.
       Rails.logger.info("Replying to intercom conversation -- intercom contact id ##{contact.id}")
       intercom_api(:conversations,
-                   :reply,
```
embarnard, line 17: this seems to work in most cases but for some clients were their conversations have been close for a long time it doesn't

app/services/intercom_service.rb
```diff
@@ -8,14 +8,15 @@ def self.create_message(email_address:, phone_number:, body:, client:, has_docum
...
-                   { id: 'last',
+                   :reply_to_last,
```
embarnard, line 18: making these changes seemed to have let one client, who had errors, messages through when I tested it but should probably confirm for more clients

PR #3436, opened 2023-02-27
Limit potential duplicate clients to match the same product year

wschaefer:
We had been getting this behavior for free with the accessible_intakes scope limiting to the current product year for GYR but CTC's current product year is still last year


embarnard (approved):
Looks good! Reviewed by Travis and Em

PR #3438, opened 2023-02-27
Add spec to check against hund.io

jalcine-cfa:


wschaefer (approved):
Looks good! Might be slightly more readable as `['hund', 'hund.io'].include? source` but that's just a thought, not a change request

jalcine-cfa:
> Looks good! Might be slightly more readable as `['hund', 'hund.io'].include? source` but that's just a thought, not a change request

I like this approach (esp because it'll be easier to change in the future)

PR #3442, opened 2023-02-28
Store and retry hub user assignment emails when mailgun errors

wschaefer:
This matches the behavior of our other email types where we store a record in the database, retry when there is a mailgun error, and track the status mailgun sends us

Also: replace rails previews for ad hoc messages with their own admin page. They should be easier to find and look more like our automated messages. Also we'd let the preview ones break without realizing it so they weren't very useful


app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -126,6 +126,7 @@ def update_outgoing_email_status
     email_to_update = (
       OutgoingEmail.find_by(message_id: message_id) ||
         VerificationEmail.find_by(mailgun_id: message_id) ||
+        AssignmentEmail.find_by(message_id: message_id) ||
```
tgrathwell, line 129: I have a personal desire for us to have **less** tables that store a `message_id` / `mailgun_status`

otherwise we'll keep adding to this list for any new email we make, when they should probably be handled more uniformly

Is there a reason we can't try to cram this `assignment_email` into the existing `SendOutgoingEmailJob` / `OutgoingEmail` / `OutgoingEmailMailer` infrastructure?

I imagine the issue is that messages sent through those classes are targeted at clients rather than hub users, but we could maybe make it work? Or if there's a dealbreaker we could build a parallel `OutgoingHubUserEmail` pathway that is more generalizable for when we add new emails targeting hub users

PR #3444, opened 2023-03-01
Fix hub client filter bug and add SearchIndexer to seeder file

nishinelson:


app/lib/client_sorter.rb
```diff
@@ -17,7 +17,7 @@ def initialize(clients, current_user, params, cookie_filters, use_product_year =
     @default_order = { "last_outgoing_communication_at" => "asc" }
     @sort_column = clients_sort_column
     @sort_order = clients_sort_order
-    @filters = filters_from(filter_source)
+    @filters = filters_from(params)
```
tgrathwell, line 20: `filter_source` had a purpose here -- if you look at [the implementation](https://github.com/codeforamerica/vita-min/blob/main/app/lib/client_sorter.rb#L16), you'll see it's drawing the filters from *either* the `params` in the URL *or* the cookie

this allows the filters to 'stick' to whatever was last chosen -- if you choose to filter by 'unassigned' for example, you can navigate away then back to 'all clients' and it will still be filtered to 'unassigned' even though there are no params in the URL

so properly fixing this bug will involve figuring out some other change

PR #3447, opened 2023-03-01
Remove hardcoded years from document types keys

embarnard:


app/lib/document_types/form13614C.rb
```diff
@@ -8,7 +8,7 @@ def relevant_to?(intake)
       end
 
       def key
-        "F13614C 2020"
+        "F13614C"
```
tgrathwell, line 11: we probably need a migration path to ensure existing docs in the database have this `key` eventually

I dunno what's a sane way... If we want to avoid errors we would probably have to support both the new and the old-style `key` for this doctype class while migrating the data

embarnard: @tgrathwell what do you mean by support both new keys? I'm trying to find places in the code where it would be disrupted if we changed the key but I can not but maybe I'm looking for the right thing?

db/data/20230306222313_change_f13614c_document_type_on_documents.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ChangeF13614cDocumentTypeOnDocuments < ActiveRecord::Migration[7.0]
```
embarnard, line 3: When I ran `rake data:migrate` locally it failed to migrate "ArchiveOnVisaColumns" because "PG::UndefinedColumn: ERROR:  column "was_on_visa" does not exist." I had to remove both the archived data migrations to test if this migration worked. Hopefully we won't have this issue on prod and demo b/c it was already run on there?

app/lib/client_pdf_document.rb
```diff
@@ -2,7 +2,13 @@ class ClientPdfDocument
   def self.create_or_update(output_file:, document_type: , client:, filename:, tax_return: nil)
     tempfile = output_file
     tempfile.seek(0)
-    document = client.documents.find_or_initialize_by(document_type: document_type.key)
+    # TODO: after no document_types of "F13614C 2020" change back to => document = client.documents.find_or_initialize_by(document_type: document_type.key)
+    document =
+      if document_type == DocumentTypes::Form13614C && (client.documents.find_by(document_type: document_type.key) || client.documents.find_by(document_type: "F13614C 2020"))
```
tgrathwell, line 7: these `find_by`s could be simplified as such

`client.documents.find_by(document_type: [document_type.key, "F13614C 2020"])`

there might also be a craftier way to avoid the double-find but maybe it doesn't matter much because it's only temporary

tgrathwell (approved):

PR #3460, opened 2023-03-08
Use encrypted cookie for visitor_id #184653844

alaroia-cfa:
Migrate legacy unencrypted cookie when its value seems OK


alaroia-cfa:
Reviewed by TG & WS

PR #3461, opened 2023-03-08
Return clients flow

embarnard:


app/models/intake/gyr_intake.rb
```diff
@@ -441,6 +443,13 @@ class Intake::GyrIntake < Intake
   after_save_commit { SearchIndexer.refresh_filterable_properties([client_id]) }
   after_destroy_commit { SearchIndexer.refresh_filterable_properties([client_id]) }
 
+  def matching_previous_year_intakes
+    attrs = [:primary_birth_date, :hashed_primary_ssn]
+    intakes = DeduplicationService.duplicates(self, *attrs, from_scope: self.class.previous_year_completed_intakes)
+    archived_intakes = DeduplicationService.duplicates(self, *attrs, from_scope: Archived::Intake::GyrIntake2021.completed_intakes)
```
jenny-heath, line 449: would like a second opinion on whether this is kosher. the deduplication service is probably not meant to compare 2 different tables but in this case it works because the archived intakes table has the 2 columns that are being matched. are there any problems this could cause in the future?

app/models/archived/intake/gyr_intake_2021.rb
```diff
@@ -338,6 +338,8 @@ def self.sti_name
   enum widowed: { unfilled: 0, yes: 1, no: 2 }, _prefix: :widowed
   enum wants_to_itemize: { unfilled: 0, yes: 1, no: 2, unsure: 3 }, _prefix: :wants_to_itemize
 
+  scope :completed_intakes, -> { joins(:tax_returns).where(tax_returns: {current_state: TaxReturnStateMachine::INCLUDED_IN_PREVIOUS_YEAR_COMPLETED_INTAKES})}
```
jenny-heath, line 341: would like a second opinion on whether it's ok to add a new scope to this archived model. in my mind they shouldn't really change ever but we did this same thing for the current model so this was a continuation of that pattern.

app/models/intake/gyr_intake.rb
embarnard: What if we renamed this service to something more accurate for how we are using it across the board like `FindMatchingIntakesService` and the duplicates method could be renamed "matches"

PR #3469, opened 2023-03-10
Check if tax return is moving to a known state

jalcine-cfa:


config/locales/en.yml
```diff
@@ -224,6 +224,7 @@ en:
       tax_returns_attributes:
         certification_level: certification level
         is_hsa: is HSA
+      invalid_tax_status: The provided tax status is not valid.
```
jalcine-cfa, line 227: Should be something like "the new tax return state was not recognized" (or something less mechanical).

tgrathwell (approved):

PR #3476, opened 2023-03-15
Create ForcedPasswordResetsController, view, and spec

nishinelson:


app/models/user.rb
```diff
@@ -52,10 +53,12 @@ class User < ApplicationRecord
   ], using: { tsearch: { prefix: true } }
 
   self.per_page = 25
+  # PASSWORD_LENGTH = 10..128
```
tgrathwell, line 56: comment to be deleted

app/validators/password_integrity_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+    #   0: too guessable: risky password. (guesses < 10^3)
+    #   1: very guessable: protection from throttled online attacks. (guesses < 10^6)
+    #   2: somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
+    #   3: safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)
+    #   4: very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)
+    # (from https://github.com/dropbox/zxcvbn#usage)
+
+    # NOTE: Add note of feedback from `password_test.feedback`?
+    puts JSON.pretty_generate(password_test.feedback.suggestions) if Rails.env.development?
+    record.errors.add(attr_name, I18n.t("errors.attributes.password.insecure", feedback: password_test.feedback.suggestions)) if password_test.score <= 2
```
tgrathwell, line 20: when we tried this it just said we need to "increase the complexity" -- seems like we would need some advice about what this means -- longer password, more types of characters, whatever. otherwise the error may frustrate people who don't know how it needs to change

app/views/hub/forced_password_resets/edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="slab">
+  <div class="grid">
+    <div class="grid__item width-one-half shift-one-fourth">
```
tgrathwell, line 3: The login page has very little left whitespace compared to this page, it might be nice if the two matched. Possibly by getting rid of the grid (could try to match devise/sessions/new)

app/controllers/application_controller.rb
```diff
@@ -430,8 +430,12 @@ def all_identifiers
     ].filter { |e| e && !e.to_s.empty? }.uniq
   end
 
-  def after_sign_in_path_for(_user)
-    @after_login_path || hub_assigned_clients_path
+  def after_sign_in_path_for(user)
+    if !user.admin? && user.forced_password_reset_at.nil?
```
tgrathwell, line 434: The text of the story actually says we should only reset passwords for people who currently have weak passwords; we should change this line to exercise the password integrity validation so we only redirected to the reset if the password was weak

app/controllers/application_controller.rb
```diff
@@ -430,8 +430,12 @@ def all_identifiers
     ].filter { |e| e && !e.to_s.empty? }.uniq
   end
 
-  def after_sign_in_path_for(_user)
-    @after_login_path || hub_assigned_clients_path
+  def after_sign_in_path_for(user)
```
tgrathwell, line 433: This hook fires after the person has already been signed in, so they can actually navigate around the app and ignore the force password reset page

If we want to lock people off from doing stuff before their password has been reset, we either need to:
* add a `before_action` to all authenticated pages making sure people get redirected back to the password reset (would probably require keeping an extra bit on info on the user saying they still need to reset their password, because after the initial login we don't have it anymore)
OR,
* find a different way in the code to redirect, where we have access to the password and have confirmed it is correct but the client is not devise-authenticated (not sure if this is possible or not!)

app/controllers/application_controller.rb
jalcine-cfa: Good point - we'll be looking into doing this!

app/controllers/application_controller.rb
jalcine-cfa: Resolved in 3f16554f4

jalcine-cfa: We took the second path of checking the password _after_ sign-in (and updated the `forced_password_reset_at` value to prevent checking it again).

In order to go with that suggestion of doing it _before_ Devise has authenticated the user, it _should_ require some sort of nonce or state value so we could avoid access to the password reset page from a non-privileged user.

app/validators/password_integrity_validator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+class PasswordIntegrityValidator < ActiveModel::EachValidator
+  def self.is_strong_enough?(password, record)
+    # Scoring of the test is a range of 0 - 4
+    #   0: too guessable: risky password. (guesses < 10^3)
+    #   1: very guessable: protection from throttled online attacks. (guesses < 10^6)
+    #   2: somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
+    #   3: safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)
+    #   4: very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)
+    # (from https://github.com/dropbox/zxcvbn#usage)
```
wschaefer, line 10: Instead of the long comment, we recommend:

```
# zxcvbn scores go from 0-4; see docs: https://github.com/dropbox/zxcvbn#usage
```

app/validators/password_integrity_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+    Zxcvbn.test(password, [record.email, record.phone_number, record.name]).score >= 3
+  end
+
+  def validate_each(record, attr_name, value)
+    return if record.admin?
+    return if value.nil? and record.encrypted_password.present?
+
+    record.errors.add(attr_name, I18n.t("errors.attributes.password.insecure")) unless PasswordIntegrityValidator.is_strong_enough?(value, record)
+    record.errors.add(attr_name, I18n.t("errors.attributes.password.incorrect_size", minimum_password_length: User.PASSWORD_LENGTH.begin)) unless User.PASSWORD_LENGTH.member?(value&.length)
```
wschaefer, line 19: No need for `&.` because we already did a nil check

wschaefer, line 19: We strongly prefer `.include?` over `.member?`.

`.include?` is part of arrays as well as ranges; `.member?` is range-only; `.include?` is very common in our codebase but `.member?` appears 0 times

app/validators/password_integrity_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+end
```
wschaefer, line 20: Please fix indenting; Use Cmd-Option-L in RubyMine

config/initializers/devise.rb
```diff
@@ -205,7 +205,9 @@
   # config.rememberable_options = {}
 
   # ==> Configuration for :validatable
-  # Range for password length.
+  # Range for password length. This is kept for use
+  # for admin users. Non-admin users have a higher minimum
+  # password length as defined by User.PASSWORD_LENGTH.
```
wschaefer, line 210: We would strongly prefer relying on setting `config.password_length` to the new minimum everywhere. As far as we can tell, if we increase this (`config.password_length`) and an admin logs in, they won't be disrupted, so we can't really understand the explanation of "kept for use for admin users." In the unlikely event an admin decides on their own volition (or a new invite) to set a new password, it's fine if it has to be >6 chars.

app/validators/password_integrity_validator.rb
wschaefer, line 19: Assuming we adjust the Devise config for password length to prefer the longer (10) value, we can remove this line.

app/views/hub/components/_main_menu.html.erb
```diff
@@ -10,7 +10,7 @@
           <div class="logo-text">The Hub</div>
         </div>
 
-        <% if current_user %>
+        <% if current_user&.forced_password_reset_at? || current_user&.admin? %>
```
wschaefer, line 13: We now have code that redirects the user to the password change page when needed, so there's no need for this.

app/views/hub/components/_main_menu.html.erb
```diff
@@ -42,11 +42,13 @@
       </div>
       <% if current_user %>
         <div class="bottom">
-          <%= render "hub/components/menu_item",
-                     link: hub_user_profile_path,
-                     item_label: current_user&.name,
-                     icon_path: image_path("icons/my-profile.svg")
-          %>
+          <% if current_user&.forced_password_reset_at? || current_user&.admin? %>
```
wschaefer, line 45: Same as above: We now have code that redirects the user to the password change page when needed, so there's no need for this.

app/views/hub/forced_password_resets/edit.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="slab">
+  <div class="grid__item width-one-half shift-one-fourth">
+    <div class="form-card form-card--compact">
+      <div class="form-card__header">
+        <h1 class="form-card__title">Please update your password. </h1>
```
wschaefer, line 5: Non-strong suggestion: Mark this for translation with `I18n.t`. Use Google Translate to get a translation. Optionally use `GOOGLE_TRANSLATE_API_KEY=xyz i18n-tasks translate-missing` to automatically get the translation so you have an easy time doing it.

(the API key is in LastPass)

In the client/intake-facing code, it's very important to mark strings for translation. In the Hub, it's less important. You'll find a handful of Hub pages that don't use translation, and it's OK not to, and this could be a page where we don't bother. However, since we are so consistent in using translations in the client-facing flow, it's surprising as a reviewer to see inline copy like this. Since this is a weak suggestion from us, it's fine if you decide to leave this comment unaddressed. I'm under the impression there a tiny number of partners who are maybe in Puerto Rico who prefer Spanish, but it's not really a use case we've optimized for.

Also minor: Consider removing the unnecessary ` ` before the `</h1>`.

app/views/hub/forced_password_resets/edit.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        <p class="text--help">It’s important that you keep this password secure and do not share it with anyone.
+          Please keep all client information safe and confidential.</p>
+      </div>
+      <%= form_for(@user, builder: VitaMinFormBuilder, url: { action: :update }, html: { method: :put }) do |f| %>
+        <div class="form-card__content">
+          <%= f.cfa_input_field(:password, "New password", type: "password", classes: ['form-width--long', 'password']) %>
```
wschaefer, line 11: `I8n.t('general.views.passwords.edit.new_password')` would do the job here (if you feel like marking strings for translation).

Tip: You can search the codebase for some English copy and see if exists in `en.yml` then right-click in RubyMine and do `Copy reference` and get an I18n key.

app/views/hub/forced_password_resets/edit.html.erb
```diff
@@ -0,0 +1,20 @@
...
+          <%= f.cfa_input_field(:password_confirmation, "Confirm new password", type: "password", classes: ['form-width--long', 'password']) %>
+          <div class="form-card__footer">
+            <%= f.continue "Update" %>
+          </div>
+        </div>
+      <% end %>
+    </div>
+  </div>
+</div>
```
wschaefer, line 20: Totally unimportant: Asheesh likes it when files end in newline characters, but no one else on the team seems to care, so do with this whatever you will. There's a RubyMine/IntelliJ option for this called, "Ensure every saved file ends with a line break".

config/routes.rb
```diff
@@ -184,6 +184,8 @@ def scoped_navigation_routes(context, navigation)
           mount DelayedJobWeb => "/delayed_job"
         end
 
+        resource :forced_password_resets, only: [:edit, :update]
```
wschaefer, line 187: A few lines down there are a number of routes nested within `users`. It would be nicer if this were next to those in this file. When doing that, it's best to use nested namespaces in the Ruby code -- rename the controller to e.g. `Hub::Users::XyzController`.

wschaefer (changes requested):
Good progress so far. We have some minor style recommendations and one big substantial question.

We know there’s a lot of text in this pull request review, so we recommend Tupling with us (Whitney and/or Asheesh) when you’re ready to make sure we understood the code properly and make sure our feedback makes sense to you.

Substantial question: Can we change how we store that the user’s password needs to change? We’d like the database column naming, in the context of the relevant code, to make it crystal clear to a code reader that if your password was a high-quality password, we don’t need you to change the password.

One suggestion for naming:

`high_quality_password_as_of` (current timestamp when you log in if your password is good and you don’t have a value for this column yet _or_ current timestamp after you successfully reset your password) If you’re a non-admin and high_quality_password_as_of is nil, you are in a forced-password-reset situation.

When the password reset happens, we could send a custom Mixpanel event so we can see how many people successfully transitioned from bad passwords to good ones.

There are a lot of style suggestions which will make a huge difference in having this pull request fit in nicely with the rest of our Rails app.

We also noticed that when showing an error message to the user showing them what needs to change, it shows a few different errors (which is fine given the situation) but that one of them is wrong - when we entered a 5 character password, it said the password needs to be at least 6 characters.

Asheesh felt that the copy on the page where people reset their password was scary, implying that our website is hacked and the user urgently needs to change their password. Can you connect with design to make sure we use great copy here and see if there’s any other communication we need to do as part of rolling this out? We would like something that communicates that we are choosing to upgrade our password rules, and their password needs to change accordingly, and make sure it is something that doesn’t include the messaging that people need to ‘keep client data safe’ (surely they already know that; what they need to know is why they’re seeing this page).

We wrote this together, Whitney has to go but Asheesh may follow up with a few more things.

spec/controllers/application_controller_spec.rb
```diff
@@ -1369,4 +1369,45 @@ def hub_path; end
...
+
+      it "redirects to change the password" do
+        get :index
+        expect(response).to redirect_to edit_hub_forced_password_resets_path
+      end
+
+      it "does not redirect if the password was already set" do
+        non_admin_user.update!(forced_password_reset_at: DateTime.now)
+        get :index
+        expect(response).not_to redirect_to edit_hub_forced_password_resets_path
```
alaroia-cfa, line 1391: It's better IMHO to expect the response to `be_ok` if you want to show it's not a redirect. `be_ok` is better than `not_to redirect_to` because presumably you want to show it doesn't redirect at all.

Relatively unimportant.

spec/controllers/application_controller_spec.rb
```diff
@@ -1369,4 +1369,45 @@ def hub_path; end
       end
     end
   end
+
+  describe "#ensure_non_admins_reset_passwords_when_forced" do
+    let(:non_admin_user) { create(:team_member_user) }
+    let(:admin_user) { create(:admin_user) }
+
+    context "when a non-admin is logged in" do
+      before do
```
alaroia-cfa, line 1378: Code style blocker:

We should move the `let` into this block. It's a common pattern in our code to have the `let` _within_ the `context... when`. Let's stick to it unless there's a reason it doesn't fit here; happy to chat.

Unless we need both users to exist in the database at once, I recommend renaming to `let(:user)` within the block.

There's one more way to make these tests more concise (and concise is good). Since the code doesn't refer to the `user` except in the `before`, it's an option not to really have a variable:

```
    context "when a non-admin is logged in" do
      before { sign_in(create(:admin_user, forced_password_reset_at: nil)) }

      it "redirects to change the password" do 
      # ...
      end
    end
```

The only essential change is moving the `let`s into the `context` where they are relevant. I do also suggest, non-blockingly, any other changes like this that can make the tests more concise.

spec/controllers/application_controller_spec.rb
```diff
@@ -1369,4 +1369,45 @@ def hub_path; end
...
+        sign_in admin_user
+      end
+
+      it "does not redirect to change password" do
+        get :index
+        expect(response).not_to redirect_to edit_hub_forced_password_resets_path
+      end
+    end
+
+    context "when no user is logged in" do
```
alaroia-cfa, line 1406: Nice that you test the `nil` case! This test will let you find out if if you can safely remove the `&.` we suggested to remove in the implementation code.

spec/controllers/hub/forced_password_resets_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  describe "#edit" do
+    context "with a logged in user with non-admin hub access" do
+      before { sign_in non_hub_admin_user }
+
+      it "shows the form for updating the password" do
+        get :edit
+        expect(response).to have_rendered(:edit)
+      end
+
+      it "redirects if the password was already reset" do
```
alaroia-cfa, line 19: Change this `it` so it doesn't have the word `if` in it by wrapping it in a context block that sets up the scenario. Generally in our tests, we strongly prefer a `context "when ..."` over an `it` with an `if`. The context block gives us a place to change the test setup.

spec/controllers/hub/forced_password_resets_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        }
+        non_hub_admin_user.reload
+      end
+
+      it "fails to update the user's password" do
+        expect(response.body).to include I18n.t("errors.attributes.password.must_be_different")
+      end
+
+      it "does not update the last forced reset date" do
+        expect(non_hub_admin_user.forced_password_reset_at).to be_nil
```
alaroia-cfa, line 57: In general, we should use `expect ... not_to ... change ...` where we can for consistency with other tests in the codebase.

spec/controllers/hub/forced_password_resets_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+            password: non_hub_admin_user.password + "new",
+            password_confirmation: "another_failed_password"
+          }
+        }
+
+        non_hub_admin_user.reload
+      end
+
+      it "fails to update the user's password" do
+        expect(non_hub_admin_user.valid_password?("one_form_of_pa$$word")).to be false
```
alaroia-cfa, line 78: What is this line doing here? I don't think it relates to the test?

spec/controllers/hub/forced_password_resets_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+    end
+
+  end
+
+  describe "#update" do
+    before { sign_in non_hub_admin_user }
+
+    context "with the same password" do
+      before do
```
alaroia-cfa, line 42: In general, it's uncommon for the `before` to make a request, and usually better to move that into an `it`.

spec/controllers/hub/forced_password_resets_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      it "does not update the last forced reset date" do
+        expect(non_hub_admin_user.forced_password_reset_at).to be_nil
+      end
+
+      it "does not redirect" do
+        expect(response).not_to redirect_to root_path
+      end
+    end
+
+    context "with new password with a mismatched confirmation" do
```
alaroia-cfa, line 65: This is really `user` model validation I think, so it belongs in `user_spec` or some other place.

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -48,6 +48,32 @@
       expect(response).to redirect_to hub_assigned_clients_path
     end
 
+    render_views
```
alaroia-cfa, line 51: Why `render_views`? I don't see any assertions on the content of the page in this test. For me, since I know `render_views` is something to be used sparingly, I check that the code I submit for review only uses it when needed as part of reading the diff before submitting the review.

spec/features/hub/authenticate_spec.rb
```diff
@@ -81,4 +81,22 @@
     # Show our specific custom error message
     expect(page).to have_text("Oops, we're sorry, but something went wrong")
   end
+
+  scenario "non-admin user is forced to reset password" do
+    user.assign_attributes(forced_password_reset_at: nil)
```
alaroia-cfa, line 86: If you're updating one attribute on `user` it's tidier to do `user.update()`. It's also good to put that in a `before`.

I notice the next line calls `validate: false`. Is that needed? I didn't see why.

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -48,6 +48,32 @@
...
+        expect do
+          post :create, params: { user: { email: non_admin_user.email, password: non_admin_user.password } }
+        end.to change(subject, :current_user).from(nil).to(non_admin_user)
+
+        expect(response).not_to redirect_to edit_hub_forced_password_resets_path
+        non_admin_user.reload
+        expect(non_admin_user.forced_password_reset_at).not_to be_nil
+      end
+
+      it "redirects if they have yet to reset their password" do
```
alaroia-cfa, line 66: Replace this `it... if` with an `it` inside a context block.

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+describe PasswordIntegrityValidator do
```
alaroia-cfa, line 4: This spec is great and unit-y!

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+         expect(subject.errors[:password]).to include(I18n.t("errors.attributes.password.incorrect_size"))
+       end
+    end
+
+    context "with an admin" do
+      before do
+        subject.admin = true
+      end
+
+      it "does not run validations" do
```
alaroia-cfa, line 51: Perhaps `it "allows any password"`

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        subject.password = "forceV"
+        expect(subject).to be_valid
+        expect(subject.errors[:password].length).to eq(0)
```
alaroia-cfa, line 54: Delete this line.

Once you know it is is `valid` you don't need to check the `.errors`.

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  describe "#validate_each" do
+    context "with an invalid password" do
+      it "is not valid" do
+        subject.password = "InvalidPha"
+        expect(subject).not_to be_valid
+        expect(subject.errors[:password]).to include(I18n.t("errors.attributes.password.insecure"))
+      end
+
+      it "is not valid if it uses a commonly used password" do
```
alaroia-cfa, line 27: Consider: `it "rejects commonly-used passwords"`

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        subject.password = "password123"
+        expect(subject).not_to be_valid
+        expect(subject.errors[:password]).to include(I18n.t("errors.attributes.password.insecure"))
+      end
+
+      it "is not valid for overly long passwords" do
```
alaroia-cfa, line 33: Consider: `it "rejects overly-long passwords"`

spec/validators/password_integrity_validator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        subject.password = "fake".ljust(129, "fake")
+        expect(subject).not_to be_valid
+        expect(subject.errors[:password]).to include(I18n.t("errors.attributes.password.incorrect_size"))
+      end
+
+       it "is not valid for too short passwords" do
```
alaroia-cfa, line 39: Consider: `it "rejects overly-short passwords"`

alaroia-cfa (changes requested):
Good start on the tests. I suggest a substantial amount of changes to get these tests consistent with the rest of the codebase.

app/controllers/hub/forced_password_resets_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    layout "hub"
+
+    def edit
+      @user = current_user
+    end
+
+    def update
+      @user = current_user
+      if @user.valid_password?(user_params[:password])
+        @user.errors.add(:password, I18n.t("errors.attributes.password.must_be_different"))
```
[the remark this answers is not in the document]
jalcine-cfa: AFAICS, This call wasn't to check if it's complex enough but if it's been used before/the stored password (via https://www.rubydoc.info/github/heartcombo/devise/main/Devise/Models/DatabaseAuthenticatable#valid_password%3F-instance_method). This is a bit of a workaround to doing decryption work of the password and adding it as a 'known' value at https://github.com/codeforamerica/vita-min/blob/d72232c64cbdead0324f24fe0ecace5e34ab66d3/app/validators/password_integrity_validator.rb#L11.

app/controllers/hub/forced_password_resets_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      elsif user_params[:password] != user_params[:password_confirmation]
+        @user.errors.add(:password, I18n.t("errors.attributes.password.not_matching"))
+      else
+        @user.assign_attributes(user_params)
+        @user.forced_password_reset_at = DateTime.now
+
+        if @user.save
+          bypass_sign_in(@user) # Devise signs out after a password change, don't do that
```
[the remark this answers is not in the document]
jalcine-cfa: If you're referring to `bypass_sign_in(@user)`, that was existing! Down to try without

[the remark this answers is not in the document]
alaroia-cfa: https://github.com/codeforamerica/vita-min/search?q=bypass_sign_in

The code in main doesn't mention `bypass_sign_in` as far as I can tell. What do you mean it was existing?

app/controllers/application_controller.rb
```diff
@@ -197,6 +199,14 @@ def set_navigator
     end
   end
 
+  def ensure_non_admins_reset_passwords_when_forced
+    return if current_user.nil? or current_user.admin? or current_user.forced_password_reset_at.present?
+    return if controller_name == Hub::ForcedPasswordResetsController.controller_name
```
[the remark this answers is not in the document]
jalcine-cfa: Will make this change in the moved controller

app/controllers/hub/forced_password_resets_controller.rb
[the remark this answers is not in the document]
jalcine-cfa: During a review with @tgrathwell and @nishinelson, looks like I misunderstood this answer! We've removed it

app/controllers/users/invitations_controller.rb
```diff
@@ -98,4 +98,10 @@ def raw_invitation_token
   def get_user_by_invitation_token
     self.resource = resource_class.find_by_invitation_token(raw_invitation_token, true)
   end
+
+  def accept_resource
+    resource = resource_class.accept_invitation!(update_resource_params)
```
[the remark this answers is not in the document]
nishinelson: accept_invitation! comes from devise_invitable and only has a bang method... this seems to happen after the params have all been validated

app/validators/password_integrity_validator.rb
nishinelson: travis thinks this is fine and cool as is

spec/features/hub/authenticate_spec.rb
jalcine-cfa: This prevents the password strength validation from occurring so we can use a weaker password when signing in (to confirm it'll check it and encourage a redirect)

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -357,8 +357,8 @@
         {
           user: {
             name: "Cher Cherimoya",
-            password: "secret password",
-            password_confirmation: "secret password",
+            password: "Unsteady-Doorframe4-Evade",
```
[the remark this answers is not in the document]
jalcine-cfa: We added Rails config for this

tgrathwell (approved):

PR #3483, opened 2023-03-17
Update file myself flow 184653371

jenny-heath:
image


app/forms/file_yourself_form.rb
```diff
@@ -0,0 +1,20 @@
+class FileYourselfForm < Form
+  include FormAttributes
+  attr_accessor :diy_intake
+  set_attributes_for :diy_intake, :email_address, :preferred_first_name, :received_1099, :filing_frequency, :referrer, :locale, :visitor_id, :source
+
+  validates :email_address, 'valid_email_2/email': true, presence: true
+  validates :preferred_first_name, presence: true
+  validates :received_1099, presence: true
+  # TODO: Validate that this check works i.e. Rails sends empty string
```
jenny-heath, line 9: i think the works but maybe we wanted to add another test and i can't remember what it was

spec/controllers/diy/file_yourself_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        it "updates the intake from the session except source, referrer, etc" do
+          expect do
+            post :update, params: params
+          end.not_to change(DiyIntake, :count)
+
+          existing_diy_intake.reload
+          expect(existing_diy_intake.email_address).to eq "example@example.com"
+          expect(existing_diy_intake.preferred_first_name).to eq "Robot"
+          expect(existing_diy_intake.received_1099).to eq "yes"
+          expect(existing_diy_intake.filing_frequency).to eq "some_years"
```
jenny-heath, line 61: i got rid of the form mocking because it wasn't actually going to call any of the form methods with the double. this is a little redundant with the form spec but wasn't sure how else to check that it updated the right subset of params

app/forms/file_yourself_form.rb
alaroia-cfa: One idea I had was manually testing this.

alaroia-cfa: I manually tested this, and it works great, so I'll delete the TODO.

spec/controllers/diy/file_yourself_controller_spec.rb
alaroia-cfa: Yay

app/views/documents/intro/edit.html.erb
```diff
@@ -8,18 +8,18 @@
     <%= t("views.documents.intro.info") %>
   </p>
 
-  <div class="document-intro-list">
+  <div class="document-list spacing-below-60">
```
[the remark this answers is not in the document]
alaroia-cfa: yay

PR #3488, opened 2023-03-23
Reorder secondary ID dropdown

embarnard:


embarnard:
This is what the drop down looks like before the change:
Screen Shot 2023-03-23 at 1 07 28 PM

This is what the drop down looks after the change: 
Screen Shot 2023-03-23 at 1 00 45 PM

config/locales/en.yml
```diff
@@ -333,15 +333,15 @@ en:
...
         certificate_of_citizenship: Certificate of Citizenship
         irs_transcript: IRS Transcript
         itin: ITIN
-        ssa_1099_for_id: SSA-1099 form
+        ssa_1099_for_id: SSA-1099 Form
         ssa_notice: SSA Notice
-        ssn: Social security card
-        w2: Current year W-2
+        ssn: Social Security Card
+        w2: Current Year W-2
```
jenny-heath, line 344: does the spanish need similar capitalization? idk who to ask about that but maybe Jor would have a quick answer

jenny-heath (approved):

PR #3494, opened 2023-03-28
Add 1099 column in experiment participants table if in DIY High-Low Experiment

embarnard:


app/views/hub/admin/experiments/index.html.erb
```diff
@@ -65,6 +66,7 @@
               <td><%= participant.record_type %>#<%= participant.record_id %></td>
             <% end %>
             <td><%= participant.treatment %></td>
+            <% if @experiment.key == "diy_high_and_low_experiment" %><td><%= participant.record.received_1099 %></td><% end %>
```
alaroia-cfa, line 69: Nice.

Not strictly related to this PR - the dotted access to `participant.record.(something)` means an [N+1 query](https://semaphoreci.com/blog/2017/08/09/faster-rails-eliminating-n-plus-one-queries.html) on this page, I think. That's just as true for the pre-existing `.client_id` as it is for the new `.received_1099` so it's not a blocker for this.PR.

In the long run, it might make this page slow. We can probably avoid it in the controller with `.include(:record)`.

That is presumably evident in the rack mini profiler that we use in dev.

Can you look into that? Happy to help with that too.

As I said, not a blocker, since the page already has this behavior.

alaroia-cfa (approved):
Looks good to me. One performance consideration that isn't a blocker for this PR.

PR #3496, opened 2023-03-30
Include all primary and secondary identity types for count of required documents

embarnard:


embarnard:
client in experiment treatment that requires selfie
Screen Shot 2023-03-30 at 11 35 43 AM

client in experiment treatment that does not require selfie
Screen Shot 2023-03-30 at 11 35 37 AM

I think the code already accounts for this but I added a test anyways for all treatments in the ID_VERIFICATION and RETURNING_CLIENT experiments.

PR #3499, opened 2023-03-30
Fix note styling of tagged users

wschaefer:


alaroia-cfa (commented):
It looks good to me, and I wondered yesterday if we needed the `html_safe`. I'm surprised this slipped through our fingers! Oh well.

I'd rather see us do the `sanitize` call (aka `simple_format`) first -- does it work to do that, then create our tag_data, mark tag_data as html_safe (the way this PR does), then keep the regex substitution in. That way, we know `simple_format` won't mangle the carefully-constructed `tag_data`.

I'll test my theory out locally and report back. If someone is in a huge hurry, this PR is fine as-is.

alaroia-cfa (approved):
lgtm

In local dev, I can't make my proposal work, and this version works great. Cheers. :)

alaroia-cfa:
> yeah i tried keeping the original structure and adding the html_safe to the content_tag first but couldn't get it to work i dunno why - we could play with it more later if that sounds fun

It does sound fun but I have too much fun in my fun queue already :D

PR #3501, opened 2023-04-01
Fix margin-bottom on forms

nishinelson:


app/assets/stylesheets/components/_form-cards.scss
```diff
@@ -4,10 +4,20 @@
   padding: 0;
   margin-left: 0;
   margin-right: 0;
+  margin-bottom: 0;
 
   .notice--warning {
     max-width: 30em;
   }
+
```
nishinelson, line 12: I do not remember adding this I do not know if it's safe to remove...

PR #3506, opened 2023-04-05
Backfill for demographic data bug #184762838

jenny-heath:
jk swapping for a data migration


alaroia-cfa:
@tgrathwell we would love to know if

- you think our data migration is good 'cause we haven't written many and you likely have advice
- do you think this is a good file path to store the specs for these (having a spec was helpful though sure if we delete the migration after it runs we could delete its spec)

PR #3508, opened 2023-04-06
Expose irs_language_preference to metabase view

jenny-heath:
questionable if needed, the story was v old but it's so easy to add things to the view


jenny-heath:
lgtm by al

PR #3509, opened 2023-04-06
Hub take action page styling: fix extra margin next to send button

jenny-heath:
the real thing that fixed it was putting a div around the two buttons because otherwise the `first-of-type` selector doesn't work. but also the closing tags were out of order so that's why there are other changes.

before:
image

after:
image


jenny-heath:
lgtm by al

also wow maybe we can detect these someday with like 'uh is our website valid??' tooling

PR #3516, opened 2023-04-11
Redirect away from outbound-call actions if intake is from past product years

embarnard:


tgrathwell:
The change as-is is mostly harmless, but I think it might be better to hide the 'call' link and redirect away from the OutboundCallsController for prior-product-year clients.

We lock out a lot of features on prior-product-year clients based on `HubClientPresenter#editable?`, like you can't add messages [or notes](https://github.com/codeforamerica/vita-min/blob/main/app/views/hub/notes/index.html.erb#L18). So we should probably make it so you can't do outbound calls either.

tgrathwell (approved):

PR #3518, opened 2023-04-13
Spike on finding minimal state required to efile in some states

tgrathwell:


tgrathwell:
will close this PR in favor of https://github.com/codeforamerica/vita-min/pull/3600 but keeping the branch around for now

PR #3524, opened 2023-04-18
Add Google OAuth2 login for hub admins

embarnard:
Success cases we handled:

* User exists on GYR side with the same email address at Google, case-insensitively

Edge cases we handled:

* User is suspended on GYR side (refuse login)
* User does not exist on GYR side (refuse login)
* User should use login form but does Google login (show error message)
* User should use Google login but uses login form (show error message)

Add more GYR default users to Seeder class so Google login can be used in dev

Disable Google login on Heroku app for now

Do not change invitation flow for now

Disable Google login on production for now so we can deploy to demo & merge to main & release safely before target date that makes it turn-on-able in production


app/models/user.rb
```diff
@@ -19,12 +19,14 @@
...
 #  name                   :string
 #  phone_number           :string
+#  provider               :string
 #  reset_password_sent_at :datetime
 #  reset_password_token   :string
 #  role_type              :string           not null
 #  sign_in_count          :integer          default(0), not null
 #  suspended_at           :datetime
 #  timezone               :string           default("America/New_York"), not null
+#  uid                    :string
```
tgrathwell, line 29: should this be `google_uid` or such? I know we're only using google for oauth but it might make it more obvious who generated this

app/models/user.rb
```diff
@@ -240,6 +244,24 @@ def activate!
...
+    email_host = email.include?('@') ? email.split("@")[-1] : email
+    Devise.omniauth_configs[:google_oauth2].options[:hd].include?(email_host)
+  end
+
+  def self.from_omniauth(auth_hash)
+    return nil unless Rails.configuration.google_login_enabled
+    return nil unless auth_hash['provider'] == "google_oauth2"
+
+    email = auth_hash.info['email']
+    return nil unless google_login_domain?(email) && google_login_domain?(auth_hash.extra.id_info["hd"])
```
tgrathwell, line 258: what extra do we get from verifying this `hd` in the auth_hash ? -- seems to me since Google gave us a callback, they believe `whoever@google-oauth-supported-domain.org` is verified to be a google user so we should be able to just validate the hostname of that if we want

but I could be wrong!!

tgrathwell (approved):

app/lib/seeder.rb
```diff
@@ -109,6 +109,19 @@ def run
       password: "theforcevita")
     additional_user.update(role: OrganizationLeadRole.create(organization: first_org)) if additional_user.role_type != OrganizationLeadRole::TYPE
 
+    if Rails.configuration.google_login_enabled
```
jalcine-cfa, line 112: Nice way of doing a backfill

jalcine-cfa (approved):
Some of the thoughts I have are non-blocking so I'm approving this

app/models/user.rb
alaroia-cfa: Yeah, agreed that uid is confusing. Makes me think the two newly-added columns (`provider` and `uid`) could be prefixed like `oauth_provider` and `oauth_uid` or `external_provider` & `external_uid`. I think `external_*` is clearer so I will try that.

app/models/user.rb
alaroia-cfa: The `hd` check verifies that they're coming in from one of the Google hosted domains (i.e., gsuite domains) in question.

I believe we get nothing or very close to nothing from doing the additional verification here. However, I'm always nervous about login integrations. I figure in the unlikely case we misunderstood the meaning of `email` and how much verification is applied to it, we can at least be sure the only people who can access this codepath are logged into an @codeforamerica or @getyourrefund Google account (i.e. in the relevant hosted domain).

app/models/user.rb
alaroia-cfa: BTW one reason I'm excited to prefix the column names with _something_ is that we used to do Zendesk OAuth2 login, and we used to have a `provider` and `uid` column then, and I found them confusing in 2020.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  private
+
+  def handle_auth(kind)
+    @user = User.from_omniauth(request.env['omniauth.auth'])
+
+    if @user.present?
+      flash[:notice] = I18n.t('devise.omniauth_callbacks.success', kind: kind)
+      sign_in_and_redirect @user, event: :authentication
+    else
+      Rails.logger.warn("Error during google_oauth2 callback: #{failure_message}") if failure_message.present?
```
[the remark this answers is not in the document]
alaroia-cfa: Can you say more about what you mean here? Do you recommend putting this into the `es.yml` as a key, but writing English text in the es.yml?

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+RSpec.describe Users::OmniauthCallbacksController do
+  describe "#google_oauth2" do
+    let(:user) { create(:admin_user) }
+    before do
+      allow(User).to receive(:from_omniauth).and_return(user)
+      @request.env["devise.mapping"] = Devise.mappings[:user]
+    end
+
+    context "when User.from_omniauth returns a user" do
+      it "signs the user in" do
```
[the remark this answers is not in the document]
alaroia-cfa: If I understand right, you're suggesting modifying this test case to validate the flash message, similar to how the next one in the file validates the flash message? I like that idea. Done!

spec/features/hub/authenticate_spec.rb
```diff
@@ -81,4 +81,25 @@
...
+    end
+
+    after do
+      OmniAuth.config.mock_auth[:google_oauth2] = nil
+    end
+
+    it "signs the user in", js: true do
+      visit new_user_session_path
+      click_on I18n.t("general.sign_in_admin")
+      expect(page).to have_text("Successfully authenticated from Google account.")
```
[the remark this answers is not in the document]
alaroia-cfa: I'm now realizing this is redundant with the controller spec validation of the flash message... but regardless I'll leave both in.

The reason I went with the raw string here is that we don't really control the underlying I18n.t string (it comes from a gem), so it's nice to see what we're assuming the gem contains. It's not a big deal, and I'm happy to make this consistent and use a translation string. Done!

app/controllers/users/omniauth_callbacks_controller.rb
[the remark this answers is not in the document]
jalcine-cfa: Yeah! It seems a bit unavoidable since that failure message can't be translated

PR #3525, opened 2023-04-18
Add offseason filing banner to homepage, display after tax deadline until closing

jenny-heath:
2 things I would like checked:
1. did the language-dependent date stuff actually work and do we like it
2. did i correctly understand the dates that each banner should be shown and did it therefore make sense to change the end_of_login date per the [story](https://www.pivotaltracker.com/story/show/184967876)

1 thing i would like an opinion on:
1. should we add a feature spec? the part that feels uncovered is the logic for the document deadline banner and just the fact that we invoke any logic in the view at all (as far as i can tell there are no feature specs that look for these banners), but it is pretty simple so idk if worth it

a note: i repurposed the previous off season partial and controller method because they weren't used anywhere (offseason here meant "closed" so i changed the names to include "offseason filing")

english:
image

spanish:
image

### DIFF
--- a/app/views/public_pages/home.html.erb
+++ b/app/views/public_pages/home.html.erb
@@ -12,7 +12,10 @@
       </div>
     </div>
   <% end %>
-  <% if open_for_gyr_intake? %>
+  <% if show_offseason_filing_banner? %>
+    <%= render 'shared/offseason_filing_banner' %>
+  <% end %>
+  <% if open_for_gyr_intake? && !show_offseason_filing_banner? %>
     <%= render 'shared/document_deadline_warning' %>
   <% end %>
 <% end %>
--- a/app/views/shared/_offseason_filing_banner.html.erb
+++ b/app/views/shared/_offseason_filing_banner.html.erb
@@ -2,12 +2,12 @@
   <div class="grid">
     <div class="grid__item">
       <% end_of_intake = Rails.configuration.end_of_intake %>
-      <% end_of_login = Rails.configuration.end_of_login %>
+      <% end_of_doc_submission = Rails.configuration.end_of_login - 1.week %>
       <%= t("views.shared.environment_warning.off_season_filing",
             end_of_intake_month: I18n.l(end_of_intake, format: "%B", locale: locale),
             end_of_intake_day: locale == :en ? end_of_intake.day.ordinalize : end_of_intake.day,
-            end_of_login_month: I18n.l(end_of_login, format: "%B", locale: locale),
-            end_of_login_day: locale == :en ? end_of_login.day.ordinalize : end_of_login.day,
+            end_of_docs_month: I18n.l(end_of_doc_submission, format: "%B", locale: locale),
+            end_of_docs_day: locale == :en ? end_of_doc_submission.day.ordinalize : end_of_doc_submission.day,
             year: Time.now.year) %>
     </div>
   </div>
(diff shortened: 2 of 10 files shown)


config/application.rb
```diff
@@ -86,8 +86,9 @@ class Application < Rails::Application
     # These defaults can be overridden per-environment if needed
     config.start_of_unique_links_only_intake = Time.find_zone('America/Los_Angeles').parse('2023-01-24 09:59:59')
     config.start_of_open_intake = Time.find_zone('America/Los_Angeles').parse('2023-01-31 09:59:59')
+    config.tax_deadline = Time.find_zone('America/New_York').parse('2023-04-18 23:59:59')
```
jenny-heath, line 89: i know this is today but maybe (maybe??) we can in future just change these dates when we want to update the banners but who knows it might have been a big waste of time to make it this smart and maybe i should have just replaced the dang thing instead of adding more helper methods

spec/controllers/application_controller_spec.rb
jenny-heath, line 1: sadly the diff view here is completely useless and i wish this all just looked like new lines

app/views/shared/_offseason_filing_banner.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  <div class="grid">
+    <div class="grid__item">
+      <% end_of_intake = Rails.configuration.end_of_intake %>
+      <% end_of_login = Rails.configuration.end_of_login %>
+      <%= t("views.shared.environment_warning.off_season_filing",
+            end_of_intake_month: I18n.l(end_of_intake, format: "%B", locale: locale),
+            end_of_intake_day: locale == :en ? end_of_intake.day.ordinalize : end_of_intake.day,
+            end_of_login_month: I18n.l(end_of_login, format: "%B", locale: locale),
+            end_of_login_day: locale == :en ? end_of_login.day.ordinalize : end_of_login.day,
+            year: Time.now.year) %>
```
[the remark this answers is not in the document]
jenny-heath: asheesh says: cool, good luck if we support chinese someday, but it's great for until 1 de octubre 2027 or so

app/views/public_pages/home.html.erb
```diff
@@ -12,7 +12,10 @@
       </div>
     </div>
   <% end %>
-  <% if open_for_gyr_intake? %>
+  <% if show_offseason_filing_banner? %>
```
jenny-heath, line 15: AL says: Super awesome refactor idea: Make a method that returns _which_ banner to show. Then that thing can be unit tested, which avoids the need for a feature spec.

Even if that doesn't happen: Gotta include this on intro page.

config/application.rb
```diff
@@ -86,8 +86,9 @@ class Application < Rails::Application
...
     config.end_of_intake = Time.find_zone('America/New_York').parse('2023-10-01 23:59:59')
-    config.end_of_login = Time.find_zone('America/New_York').parse('2023-10-15 23:59:00')
+    config.end_of_login = Time.find_zone('America/New_York').parse('2023-10-8 23:59:00')
```
jenny-heath, line 91: add another one for end of documents, different from end of login

jenny-heath: 'end of docs' is consistently a week before end-of-login, so could be calc'd if that sounds fun

jenny-heath:
@alaroia-cfa says: Super awesome refactor idea: Make a method that returns which banner to show. Then that thing can be unit tested, which avoids the need for a feature spec.

Maybe do this as a follow-up to this story

jenny-heath:
another adjacent refactor: closed for season banner and document deadline banner could get dates from config instead of hard-coding

nishinelson (approved):
asheesh says lgtm

PR #3529, opened 2023-04-19
Redirect when we have no documents to show in the intro controller

jalcine-cfa:


spec/controllers/documents/intro_controller_spec.rb
```diff
@@ -39,6 +39,17 @@
       end
     end
 
+    context "when no documents are required" do
+      before do
+        allow(subject.class).to receive(:show?).and_return(false)
```
alaroia-cfa, line 44: Not a blocker: it's tidier to do `described_class` rather than `subject.class`.

I encourage you to make this one change, then merge (no need for re-review if you make this change).

spec/controllers/documents/intro_controller_spec.rb
```diff
@@ -13,7 +13,7 @@
       end
     end
 
-    context "when there are no document_types_definitely_needed" do
+    context "when no documents have been provided but are needed" do
```
alaroia-cfa, line 16: This test description doesn't seem right to me. I see there are some documents provided - the `create :document` lines. Can you explain?

alaroia-cfa (commented):
Overall, I like the changes, but before approving, I have a question I'd like to follow up on. Maybe we can Tuple?

spec/controllers/documents/intro_controller_spec.rb
alaroia-cfa: Or if you revert this change, fine to merge.

alaroia-cfa (changes requested):
See comments - if you handle those, happy to see it merged.

alaroia-cfa:
Since I was code-reviewing this and Jacky is out today, in the interest of getting the change out, I figured I'd make the suggested changes.

PR #3531, opened 2023-04-26
When converting HEIC files, wait for conversion before running AnalyzeJob [#184792253]

alaroia-cfa:
I believe this will fix many/all of the crashes involving AnalyzeJob.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -65,18 +65,14 @@ class Document < ApplicationRecord
   scope :archived, ->() { where(archived: true) }
 
   after_save do
-    if upload.present? && upload.blob.present? && upload.blob.filename.extension_without_delimiter.downcase == "heic" && !upload.blob.persisted?
-      # Skip AnalyzeJob when initially creating .heic files, since we will analyze them after JPG conversion
-      upload.blob.analyzed = true
-    end
+    # Skip AnalyzeJob when initially creating .heic files, since we will analyze them after JPG conversion
+    upload.blob.analyzed = true if is_heic? && !upload.blob.persisted?
   end
 
   after_create_commit do
     uploaded_by.is_a?(Client) ? InteractionTrackingService.record_incoming_interaction(client) : InteractionTrackingService.record_internal_interaction(client)
 
-    if upload.filename.extension_without_delimiter.downcase == "heic"
-      HeicToJpgJob.perform_later(id)
-    end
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -64,6 +64,13 @@ class Document < ApplicationRecord
   scope :active, ->() { where(archived: false) }
   scope :archived, ->() { where(archived: true) }
 
+  after_save do
+    if upload.present? && upload.blob.present? && upload.blob.filename.extension_without_delimiter.downcase == "heic" && !upload.blob.persisted?
```
tgrathwell, line 68: maybe `upload.filename.extension_without_delimiter.downcase` can be `Document#is_heic?`

app/models/document.rb
```diff
@@ -64,6 +64,13 @@ class Document < ApplicationRecord
...
+      # Skip AnalyzeJob when initially creating .heic files, since we will analyze them after JPG conversion
+      upload.blob.analyzed = true
```
tgrathwell, line 70: feels sneaky but if it works it works

spec/models/document_spec.rb
```diff
@@ -217,25 +217,34 @@
...
-        allow(HeicToJpgJob).to receive(:perform_later)
+      before do
+        allow(HeicToJpgJob).to receive(:perform_later).and_call_original
+      end
 
+      it "creates a job to convert the file to jpg and delays ActiveStorage::AnalyzeJob until then" do
+        document = build :document, upload_path: Rails.root.join("spec", "fixtures", "files", "IMG_4851.HEIC")
         document.save!
 
+        expect(ActiveJob::Base.queue_adapter.enqueued_jobs.map { |x| x["job_class"] }).not_to include("ActiveStorage::AnalyzeJob")
```
tgrathwell, line 231: could we instead assert here that the `enqueued_jobs` is just a `[HeicToJpgJob]`

tgrathwell (approved):

app/models/document.rb
alaroia-cfa: 👟

spec/models/document_spec.rb
alaroia-cfa: done, yay

app/models/document.rb
alaroia-cfa: yes but with so many ampersands you'll be amazed

PR #3535, opened 2023-05-01
Remove covid banner vita providers index and show pages. Delete covid banner partial

nishinelson:


wschaefer (approved):
Looks great! I moved the cut off part of your PR title from the comments into the title, hope that's ok :)

PR #3540, opened 2023-05-03
Add spouse doc handling copy to doc overview in client portal and intake flow for joint filers

nishinelson:


app/views/shared/_document_overview_list.html.erb
```diff
@@ -67,6 +70,9 @@
             <li>
               <%= document.upload.filename %>
             </li>
+            <% if documents.length <= 1 && current_intake.filing_joint == "yes" %>
+              <p class="text--red-bold"><%= t("views.documents.overview.spouse_doc_needed") %></p>
```
tgrathwell, line 74: This <p> will end up a direct child of the <ul> which is technically not allowed, it should probably be an `<li><p>` or `<li><em>`

app/views/shared/_document_overview_list.html.erb
```diff
@@ -57,16 +57,22 @@
             <li>
               <%= link_to document.upload.filename, portal_document_path(id: document.id), target: "_blank", rel: "noopener noreferrer" %>
             </li>
+            <% if documents.length <= 1 && current_intake.filing_joint == "yes" %>
```
tgrathwell, line 60: we tend to prefer `current_intake.filing_jointly?` instead of `current_intake.filing_joint == "yes"`

PR #3544, opened 2023-05-10
Add Retry Logic for Intercom

jalcine-cfa:
A naive approach of doing some retries against Intercom in the event of an authentication failure.

### DIFF
--- a/app/services/intercom_service.rb
+++ b/app/services/intercom_service.rb
@@ -134,7 +134,14 @@ class IntercomService
       Rails.logger.debug("Calling Intercom: #{collection}.#{verb}(#{params})")
     end
 
-    result = intercom.send(collection).send(verb, params)
+    retry_counts = 0
+
+    begin
+      result = intercom.send(collection).send(verb, params)
+    rescue
+      retry_counts += 1
+      retry if retry_counts < 3
+    end
 
     if Rails.env.development?
       Rails.logger.debug("Intercom provided response: #{result.inspect}")
--- a/app/services/intercom_service.rb
+++ b/app/services/intercom_service.rb
@@ -141,8 +141,9 @@ class IntercomService
     begin
       result = intercom.send(collection).send(verb, params)
     rescue
-      retry_counts += 1
-      retry if retry_counts < MAX_RETRY_COUNT
+      Intercom::AuthenticationError => e
+        retry_counts += 1
+        retry if retry_counts <= MAX_RETRY_COUNT
     end
 
     if Rails.env.development?
--- a/app/services/intercom_service.rb
+++ b/app/services/intercom_service.rb
@@ -133,7 +133,9 @@ class IntercomService
     if Rails.env.development?
       Rails.logger.debug("Calling Intercom: #{collection}.#{verb}(#{params})")
     end
+
     result = intercom.send(collection).send(verb, params)
+
     if Rails.env.development?
       Rails.logger.debug("Intercom provided response: #{result.inspect}")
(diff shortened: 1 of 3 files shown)


abibbs-cfa:
@jalcine-cfa I've added @tgrathwell as a reviewer to this PR. He might have some good feedback on it in its current state.

app/services/intercom_service.rb
```diff
@@ -129,11 +129,22 @@ def self.intercom
...
     end
-    result = intercom.send(collection).send(verb, params)
+
+    retry_counts = 0
+
+    begin
+      result = intercom.send(collection).send(verb, params)
+    rescue
+      retry_counts += 1
+      retry if retry_counts < MAX_RETRY_COUNT
```
abibbs-cfa, line 145: This may only retry twice since we increment retry_counts before checking if it's less than 3.

We could possibly change `< MAX_RETRY_COUNT` to `<= MAX_RETRY_COUNT`, but there may be reasons why we only want a max of 2 retries.

app/services/intercom_service.rb
```diff
@@ -129,11 +129,22 @@ def self.intercom
...
+    end
```
abibbs-cfa, line 146: Is there a graceful way of surfacing this error in the instance where we've reached the `MAX_RETRY_COUNT` (datadog/sentry etc)?

app/services/intercom_service.rb
```diff
@@ -129,11 +129,22 @@ def self.intercom
...
     if Rails.env.development?
       Rails.logger.debug("Calling Intercom: #{collection}.#{verb}(#{params})")
     end
-    result = intercom.send(collection).send(verb, params)
+
+    retry_counts = 0
+
+    begin
+      result = intercom.send(collection).send(verb, params)
+    rescue
```
tgrathwell, line 143: this needs to be limited to just the networky errors we think are worthy of retry, possibly just  `Intercom::AuthenticationError`

otherwise it will retry needlessly 2x for other mistakes we might make in the future on that line that are less likely to be recoverable

app/services/intercom_service.rb
jalcine-cfa: I'm waffling on this. I was considering throwing the error with an additional note in the logs about the attempted retry counts _or_ making a custom error _specific_ for this (so it's easier to alert on).

app/services/intercom_service.rb
tgrathwell, line 145: could consider booping a Datadog counter whenever this retries, like in https://github.com/codeforamerica/vita-min/blob/main/app/services/efile/poll_for_acknowledgments_service.rb#L16

that way we could tell how often we're getting into this rescue at all

app/services/intercom_service.rb
jalcine-cfa: Opted for re-raising the error with a custom message. I think with the other mention of a datadog counter, it'd be clear if it happens here

spec/services/intercom_service_spec.rb
```diff
@@ -233,13 +233,62 @@
...
+
+    context "when there's an upstream authentication failure" do
+      before do
+        allow(fake_intercom.messages).to receive(:create).with(params) {
+          raise Intercom::AuthenticationError.new("fake error")
+        }
+      end
+
+      it "fails after a number of retries" do
+        # FIXME: Confirm that datadog counter was incremented by expected amount (3)
```
tgrathwell, line 265: does this comment still apply?

PR #3566, opened 2023-06-09
Redirect clients from the /efile page if they never efiled

embarnard:


app/views/hub/efile_submissions/show.html.erb
```diff
@@ -2,6 +2,7 @@
   <%= render "hub/clients/client_header" %>
   <%= render "hub/clients/navigation" %>
   <% @tax_returns.each do |tax_return| %>
+    <% next if tax_return.efile_submissions.empty? %>
```
embarnard, line 5: maybe this is overkill

tgrathwell (approved):

app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -22,7 +22,7 @@ def show
       # Eager-load tax returns with submissions and data we may need to render
       @tax_returns = client.tax_returns.includes(:efile_submissions, efile_submissions: :fraud_score).where.not(tax_returns: {efile_submissions: nil}).load
       @fraud_indicators = Fraud::Indicator.unscoped
-      redirect_to hub_client_path(id: @client.id) and return unless @tax_returns.present?
+      redirect_to hub_client_path(id: @client.id) and return if @tax_returns.empty? || !@client.online_ctc?
```
tgrathwell, line 25: maybe instead of checking anything about CTC it should just look for efile submissions

that's the criteria that the navigation partial uses when choosing whether to hide/show the tab anyway
https://github.com/codeforamerica/vita-min/blob/main/app/views/hub/clients/_navigation.html.erb#L6

PR #3574, opened 2023-06-15
Add Job for Blur Detection (#3571)

embarnard:


Dockerfile
```diff
@@ -64,6 +64,9 @@ RUN set -a \
 
 RUN echo "IRB.conf[:USE_AUTOCOMPLETE] = false" > ./.irbrc
 
+# Install the OpenCV headless library
+RUN pip3 install opencv-python-headless --break-system-packages
```
jalcine-cfa, line 68: This is in response to a warning that Debian throws when this is deployed to Aptible. See https://stackoverflow.com/a/75696359 for more information.

bin/setup
```diff
@@ -42,6 +42,9 @@ FileUtils.chdir APP_ROOT do
   puts "== Install JavaScript dependencies =="
   system! "yarn install"
 
+  puts "== Install Python dependencies =="
+  system! "pip3 install --user opencv-python-headless"
```
jalcine-cfa, line 46: Developers should re-run `bin/setup` to make sure. this job can run with its needed deps.

Dockerfile
tgrathwell: it sounds so hardcore!! smash the system!!

PR #3587, opened 2023-06-28
Access control audit

embarnard:


app/controllers/hub/analytics_controller.rb
```diff
@@ -2,6 +2,8 @@ module Hub
   class AnalyticsController < ApplicationController
     include AccessControllable
     before_action :require_sign_in
+    load_and_authorize_resource parent: false
```
tgrathwell, line 5: This line probably doesn't do anything (if we had a `show` action it would try to do an `Analytics.find` but since this controller only has an `index` it does nothing)

embarnard: Oh right I think I added that in before the require_admin thing existed but then I ended up rebasing the branch and using that instead @tgrathwell

PR #3591, opened 2023-07-06
Replace hub profile SVG with filled out version

embarnard:


embarnard:
Before the change: 
Screen Shot 2023-07-06 at 3 30 17 PM

After the change: 
Screen Shot 2023-07-06 at 3 29 27 PM

You can see now the profile icon matches the rest of the hub tool icons.

tgrathwell (approved):

PR #3593, opened 2023-07-06
Restrict and Assert Permissions of the Metabase Database User

jalcine-cfa:


spec/db/data/metabase_user_restrictions_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "rails_helper"
+
+describe "RestrictiveAccessForMetabaseUser" do
+  before(:all) do
+    Rails.application.load_tasks
+  end
+
+  context "when attempting intentional behavior" do
+    it "can select tables from the analytics table" do
+      RSpec::Expectations.configuration.on_potential_false_positives = :nothing
```
tgrathwell, line 10: I think disabling this RSpec behavior is the wrong call because the message RSpec gives is right on the money:

> WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass, message)` risks false positives, since literally any other error would cause the expectation to pass, including those raised by Ruby (e.g. `NoMethodError`, `NameError` and `ArgumentError`), meaning the code you are intending to test may not even get reached. Instead consider using `expect { }.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`.

In this case I would prefer the basic `not_to raise_error`, or to make an assertion about the result of the `execute` statement.

This overly specific `not_to raise_exception` assertion actually hides a number of errors that happen when running the code -- no such table `team_role_members`, `failed_attempts ==` should be a single equals sign, etc.

More importantly, the code being tested no longer seems to give permissions to the analytics table, it will raise this: `PG::InsufficientPrivilege: ERROR:  permission denied for schema analytics`

spec/db/data/metabase_user_restrictions_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+select count(*) from analytics.team_role_members (vita_partner_id, id);
+select * from analytics.users where failed_attempts == 0;
+")
+      }.not_to raise_exception(ActiveRecord::StatementInvalid, /permission denied/)
+
+    end
+  end
+
+  context "when attempting unintentional behavior" do
+    before(:all) do
```
tgrathwell, line 24: I think this `before(:all)` should probably be in the outermost block of the file to ensure it re-creates the user for both the positive and negative tests

spec/db/data/metabase_user_restrictions_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "rails_helper"
+
+describe "RestrictiveAccessForMetabaseUser" do
```
tgrathwell, line 3: This test file currently is in the `spec/db/data` folder right now, which is meant to be the specs for our data migrations

It probably belongs in the existing `spec/tasks/analytics_spec.rb` file with the other tests for this rake task

spec/db/data/metabase_user_restrictions_spec.rb
jalcine-cfa: So I've removed that warning and flipped it to check for the exception - good catch. When I did so to see what output would occur, I got the following locally:

```

expected no ActiveRecord::StatementInvalid with message matching /permission denied/, got #<ActiveRecord::StatementInvalid:"PG::InsufficientPrivilege: ERROR:  permission denied for schema analytics\nLINE 3: select count(*) from analytics.team_role_members;\n                             ^\n"> with backtrace:
  # ./spec/tasks/analytics_spec.rb:24:in `block (5 levels) in <top (required)>'
  # ./spec/tasks/analytics_spec.rb:23:in `block (4 levels) in <top (required)>'
  # ./spec/support/locale.rb:4:in `block (2 levels) in <main>'

  0) creating database views metabase access when attempting intentional behavior can select tables from the analytics table
     Failure/Error:
                 ActiveRecord::Base.connection.execute("
       set role metabase;
       select count(*) from analytics.team_role_members;
       select * from analytics.users where failed_attempts == 0;
       ")

       expected no ActiveRecord::StatementInvalid with message matching /permission denied/, got #<ActiveRecord::StatementInvalid:"PG::InsufficientPrivilege: ERROR:  permission denied for schema analytics\nLINE 3: select count(*) from analytics.team_role_members;\n                             ^\n"> with backtrace:
         # ./spec/tasks/analytics_spec.rb:24:in `block (5 levels) in <top (required)>'
         # ./spec/tasks/analytics_spec.rb:23:in `block (4 levels) in <top (required)>'
         # ./spec/support/locale.rb:4:in `block (2 levels) in <main>'
     # ./spec/tasks/analytics_spec.rb:23:in `block (4 levels) in <top (required)>'
     # ./spec/support/locale.rb:4:in `block (2 levels) in <main>'
1 example, 1 failure, 0 passed
Finished in 0.436265 seconds

Process finished with exit code 1
```
Going to wait to see CI has any issues but it hasn't failed on my machine.

jalcine-cfa: Getting a different error (missing role) in https://github.com/codeforamerica/vita-min/pull/3593/commits/687a5fa1dd5761a383555aa6c14c5b5624eb30ae. Going to try running the changes only once in https://github.com/codeforamerica/vita-min/pull/3593/commits/8453d94fc88f4839021051e72118cebde9f3f378 - think there could be a conflict between the tests in parallel.

jalcine-cfa: That seems to have fixed it.

spec/db/data/metabase_user_restrictions_spec.rb
```diff
@@ -5,27 +5,26 @@
...
 
       expect {
         ActiveRecord::Base.connection.execute("
 set role metabase;
-select count(*) from analytics.team_role_members (vita_partner_id, id);
+select count(*) from analytics.team_role_members;
 select * from analytics.users where failed_attempts == 0;
 ")
-      }.not_to raise_exception(ActiveRecord::StatementInvalid, /permission denied/)
+      }.to raise_exception(ActiveRecord::StatementInvalid, /permission denied/)
```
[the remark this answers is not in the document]
jalcine-cfa: _Oh_, I see. I don't know what I didn't grok that the first time around. I'll correct this.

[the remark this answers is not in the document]
jalcine-cfa: Okay, I've corrected that and added more negative tests to confirm that we can't mess with the schema _or_ peek at the public tables.

tgrathwell (approved):

PR #3599, opened 2023-07-12
Add ID verification to Intercom Requests

jalcine-cfa:


tgrathwell:
@jalcine-cfa this PR sets up a different `secure_mode_secret_key` for every domain but I think we only need two, a production one and a demo one

I found the secret keys we are supposed to use by going to https://app.intercom.com/a/apps/[app id]/settings/identity-verification/web as directed in https://www.intercom.com/help/en/articles/183-set-up-identity-verification-for-web-and-mobile

jalcine-cfa:
> @jalcine-cfa this PR sets up a different `secure_mode_secret_key` for every domain but I think we only need two, a production one and a demo one
> 
> I found the secret keys we are supposed to use by going to [https://app.intercom.com/a/apps/[app](https://app.intercom.com/a/apps/%5Bapp) id]/settings/identity-verification/web as directed in https://www.intercom.com/help/en/articles/183-set-up-identity-verification-for-web-and-mobile

I used the production key for prod and made up different ones for the other environments. I _don't_ think I grabbed the one for demo but I can use for the heroku and demo envs. I did make fake ones for the other envs b/c I didn't think it was worthwhile using the demo token there

jalcine-cfa:
Okay, so it looks like the key for intercom can _only_ be used for production. I don't think that's an issue for other levels (like it doesn't provide an option for a key for GYR [Test]) so I'm going to only roll the one to be matching for prod and leave the stubbed ones for each other env.

Unless we're okay with using it for something like demo as well.

tgrathwell (approved):

PR #3606, opened 2023-07-21
Allow clients with archived intakes to view their tax returns

nishinelson:


app/services/client_login_service.rb
```diff
@@ -1,31 +1,58 @@
...
+    else
+      Client.where(intake: intake_matches).uniq
+    end
   end
 
   def can_login_by_email_verification?(email_address)
-    service_class.accessible_intakes.where(email_address: email_address).or(service_class.accessible_intakes.where(spouse_email_address: email_address)).exists?
+    service_class.accessible_intakes.where(email_address: email_address).or(service_class.accessible_intakes.where(spouse_email_address: email_address)).exists? || backup_service_class.where(primary_consented_to_service: true, email_address: email_address).or(backup_service_class.where(primary_consented_to_service: true, spouse_email_address: email_address)).exists?
+
+    # if backup_service_class
```
jalcine-cfa, line 46: This can be deleted, I believe!

app/services/client_login_service.rb
```diff
@@ -1,31 +1,58 @@
...
+    #   return backup_service_class.where(primary_consented_to_service: true, email_address: email_address).or(backup_service_class.where(primary_consented_to_service: true, spouse_email_address: email_address)).exists?
+    # end
   end
 
   def can_login_by_sms_verification?(sms_phone_number)
-    service_class.accessible_intakes.where(sms_phone_number: sms_phone_number, sms_notification_opt_in: "yes").exists?
+    service_class.accessible_intakes.where(sms_phone_number: sms_phone_number, sms_notification_opt_in: "yes").exists? || backup_service_class.where(primary_consented_to_service: true, sms_phone_number: sms_phone_number).exists?
+
+    # if backup_service_class
```
jalcine-cfa, line 54: Same as above!

jalcine-cfa (commented):
Added some notes on unneeded comments for a first-pass review!

app/controllers/application_controller.rb
```diff
@@ -55,7 +55,7 @@ def hub?
   end
 
   def current_intake
-    current_client&.intake || (Intake.find_by_id(session[:intake_id]) unless session[:intake_id].nil?)
+    current_client&.intake || (Intake.find_by_id(session[:intake_id]) unless session[:intake_id].nil?) || current_client&.current_intake
```
jalcine-cfa, line 58: We can drop `current_client&.intake` since `current_client&.current_intake` also does this check!

app/lib/seeder.rb
```diff
@@ -460,17 +460,21 @@ def run
...
-    find_or_create_intake_and_client(
+    archived_intake = find_or_create_intake_and_client(
       Archived::Intake::GyrIntake2021,
       primary_first_name: "ArchivedGyr2021",
       primary_last_name: "Adams",
       primary_consented_to_service: "yes",
       primary_birth_date: 75.years.ago,
       primary_tin_type: 'ssn',
+      primary_ssn: "323659989",
+      primary_last_four_ssn: "9989",
```
jalcine-cfa, line 471: Do we need to keep this? I think this was for testing, correct?

tgrathwell:
I'm going to close this PR for now because I think we've elected to pause this feature until we've answered some other product questions.

I'm gonna keep the branch not-deleted because if we do choose to revisit it most of the code should be valid

PR #3612, opened 2023-07-31
Fix crash when assigning people to an organization lead

tgrathwell:


app/models/organization_lead_role.rb
```diff
@@ -21,6 +21,10 @@ class OrganizationLeadRole < ApplicationRecord
   belongs_to :organization, foreign_key: "vita_partner_id", class_name: "VitaPartner"
   validate :no_sites
 
+  def sites
+    served_entities
```
embarnard, line 25: I wonder if we can just use served_entities directly in the tax return assignment service, add this sites method causes the edit_role method in the users_controller to get messed up

PR #3614, opened 2023-08-02
Replace org/site dropdowns in Hub

embarnard:
This PR adds single type-to-select drop-downs for the following pages:

**1. Edit client organization page**
Screen Shot 2023-08-08 at 11 59 19 AM

**2. User edit role page for org and coalition leads**
Screen Shot 2023-08-08 at 11 59 00 AM

**3. Send user invitation page for org and coalition leads**
Screen Shot 2023-08-08 at 11 58 18 AM

It is similar to the multi-select tool that is used for team members but the user can only select one value.


jalcine-cfa (approved):
Got a few thoughts but LGTM!

PR #3619, opened 2023-08-09
Add Client ID in Portal Home

tahsinaislam:


config/locales/en.yml
```diff
@@ -1632,6 +1632,7 @@ en:
           add_signature_primary: Please add your final signature to your tax return
           add_signature_spouse: Please add your spouse's final signature to your tax return
           finish_intake: Please answer remaining tax questions to continue.
+        client_ID: 'Client ID: %{id}'
```
embarnard, line 1635: I remove the capitalization from client_ID so it looks like `client_id` but other wise looks good!

PR #3623, opened 2023-08-14
Add job to update a gauge for worker latency

jalcine-cfa:
Lots of inspiration from https://github.com/codeforamerica/gcf-backend/blob/main/app/jobs/worker_heartbeat_job.rb and https://github.com/codeforamerica/gcf-backend/blob/main/lib/tasks/worker_heartbeat.rake.


embarnard (approved):
Looks good to me! Although I can't see though reference links, I must not have access to gcf-backend

tgrathwell (approved):

PR #3627, opened 2023-08-15
Add git-credential-manager instructions and additional IRS schemas for download in set-up README

embarnard:


jalcine-cfa (commented):
Do we want to add `git-credential-manager` to `Brewfile` as well? That'll optimize setup for whomever runs `bin/setup`

embarnard:
Actually leaving it as is because we need to authenticate with the fraud gem before running the setup script

PR #3634, opened 2023-08-18
Protect tagify selects from javascript injections and pull vita partner from filtered vita partners on edit_role update  

embarnard:


app/controllers/hub/clients/organizations_controller.rb
```diff
@@ -36,13 +36,13 @@ def load_vita_partners
...
 
-      def authorize_vita_partner
-        raise CanCan::AccessDenied unless @vita_partners.find_by(id: parsed_vita_partner_id).present?
-      end
-
-      def parsed_vita_partner_id
-        return nil unless client_params[:vita_partners].present?
-        JSON.parse(client_params[:vita_partners]).pluck("id")
+      def load_vita_partner
+        id = JSON.parse(client_params[:vita_partners]).pluck("id").first
```
alaroia-cfa, line 40: Non-blocker: If `JSON.parse()` can raise exceptions, it can be nice to catch them. The best way to handle them would probably be to treat it the same as when the vita_partner can't be found, i.e., the `head :forbidden` section.

app/views/hub/users/edit_role.html.erb
```diff
@@ -38,7 +38,7 @@
         <div class="form-group">
           <%= label_tag(:coalition_id, t("devise.invitations.new.coalition_label"), class: "form-question") %>
           <div class="form-width--long">
-            <% coalition_value = @user.role.is_a?(CoalitionLeadRole) && @user.role.coalition.present? ? [{ value: @user.role.coalition.id, name: @user.role.coalition.name }].to_json.to_s.html_safe : [] %>
+            <% coalition_value = @user.role.is_a?(CoalitionLeadRole) && @user.role.coalition.present? ? taggable_format([{ value: @user.role.coalition.id, name: @user.role.coalition.name }]) : [] %>
```
alaroia-cfa, line 41: Non-blocking:

This ternary returns different data types depending which side is used. Can that be cleaned up?

Specifically, I'm under the impression `taggable_format()` returns a string, but after the `:` we see `[]`, which is an array.

Perhaps it's best to move the ternary into the call to `taggable_format` like

```
taggable_format(
  condition ? [{ value: @user.role.coalition.id, name: @user.role.coalition.name }] : []
)
```

That way we'd get the same data type out no matter what.

app/views/hub/users/edit_role.html.erb
```diff
@@ -49,7 +49,7 @@
         <div class="form-group">
           <%= label_tag(:organization, t("devise.invitations.new.organization_label"), class: "form-question") %>
           <div class="form-width--long">
-            <% org_value = @user.role.is_a?(OrganizationLeadRole) && @user.role.organization.present? ? [{ value: @user.role.organization.id, name: @user.role.organization.name }].to_json.to_s.html_safe : [] %>
+            <% org_value = @user.role.is_a?(OrganizationLeadRole) && @user.role.organization.present? ? taggable_format([{ value: @user.role.organization.id, name: @user.role.organization.name }]) : [] %>
```
alaroia-cfa, line 52: Non-blocking - would it improve cleanliness for us to return the same data type no matter which side of the ternary we're on, like above?

app/views/hub/users/edit_role.html.erb
```diff
@@ -61,7 +61,8 @@
         <div class="form-group">
           <%= label_tag(:sites, t("devise.invitations.new.site_label"), class: "form-question") %>
           <div class="select form-width--long">
-            <%= hidden_field_tag(:sites, (@role&.respond_to?(:sites) ? @role.sites : []).map { |site| { value: site.id, name: site.name }}.to_json.to_s.html_safe, class: "multi-select-vita-partner select__element") %>
+            <% site_value = taggable_format((@role&.respond_to?(:sites) ? @role.sites : []).map { |site| { value: site.id, name: site.name }}) %>
```
alaroia-cfa, line 64: This is a good example of always returning the same data type no matter which side of the ternary executes.

spec/controllers/hub/clients/organizations_controller_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Hub::Clients::OrganizationsController, type: :controller do
+  let!(:other_org_2) { create :organization }
```
alaroia-cfa, line 4: 🗑️

spec/controllers/hub/clients/organizations_controller_spec.rb
```diff
@@ -28,6 +29,7 @@
     let(:vita_partners) { JSON.generate([{ id: site.id, name: site.name, value: site.id }]) }
     let(:params) { { id: client.id, client: { vita_partners: vita_partners } } }
     let(:instance) { instance_double(UpdateClientVitaPartnerService) }
+    let!(:orig_class) { UpdateClientVitaPartnerService }
```
alaroia-cfa, line 32: 🗑️

alaroia-cfa (approved):
I left some non-blocking suggestions. Overall this is a good fix for the issues!

PR #3637, opened 2023-08-22
Add alert when deploy fails

tahsinaislam:


.circleci/config.yml
```diff
@@ -172,6 +175,10 @@ jobs:
       - run: echo $APTIBLE_PUBLIC_KEY >> ~/.ssh/known_hosts
       - run: git fetch --depth=1000000
       - run: git push git@beta.aptible.com:vita-min-demo/vita-min-demo.git $CIRCLE_SHA1:master
+      - slack/notify:
+          event: fail
+          mentions: "@badgers"
+          template: basic_fail_1
```
alaroia-cfa, line 181: Is it going to be clear what exactly failed? It would be nice to share a screenshot of what the alert might look like.

alaroia-cfa (approved):
This is a pretty cool improvement to the developer experience!

I have a few non-blocking questions and one big one which may be blocking. I'm happy to Tuple up if that's a better way to get the questions answered or discuss anything!

## Possible big question

In the story, there's a mention that deploys may fail due to test failures - "Add an alert when a deploy fails because of a test failure in CircleCI". This seems to be referring to a situation where a pull request passes the tests, but when it gets merged to main, it happens to fail the tests, so the Aptible deploy to demo never runs.

Given that the Aptible deploy never runs, will any alerts be fired from the new configuration here? If not, I think this pull request doesn't accomplish the goal, and we should change it so it does.

## Non-blocking Heroku questions

For Heroku deploys, will this update the Heroku deployment GitHub Action "Check" with a status of failed? If not, is it feasible to add that? See screenshot:

image

Also, I see we're repeating the instructions on how to get the Heroku logs. Does it make sense to remove that repetition, since it's also in the previous comment? Just asking, not trying to push things one way or the other; conciseness is a virtue but relevant information is also a virtue. Hard to pick sometimes.

## Non-blocking Aptible questions

Is there a way to rehearse this, e.g, trigger a test failure message? It would be very nice to be able to do that. CircleCI has a painful but adequate way to set that up: Look for "Trigger Pipeline" [in this doc](https://circleci.com/docs/triggers-overview/#run-a-pipeline-from-the-circleci-web-app). I strongly recommend adding this.

Do we know what `basic_fail_1` looks like? Can you share screenshots of that or a link to it? I see it will go to Slack, which seems good.

.circleci/config.yml
jalcine-cfa: This is using a stock message that Slack's orb provides, similar to what we see in https://cfa.slack.com/archives/G0114SW0SAU/p1692816895824119

jalcine-cfa:
I think re: the Heroku pull request bit, if I understand correctly, it _won't_ be possible because that happens via a GitHub action and this is happening from CircleCI's own flow of operations (unless we're considering issuing a call to [update that status](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#update-a-check-run) - although I don't think we can do that for an action via the API, from a quick glance).

Regarding the comment, yeah, that _makes sense_ to remove the bit re: Heroku logs. Before I was in favor of keeping the whole message but it might be worth removing that log with a link to _which_ commit failed (for historical purposes as well).

tahsinaislam:
Updates on comments from Asheesh:

- **Possible big question:** Notifications now run on when tests fails on main as the deploy job does not execute when tests fail
- **Non-blocking Heroku questions**: Heroku notification has been removed from previous commit as it is redundant 
- **Non-blocking Aptible questions**: To rehearse the test failure: config has been updated so the special alerting behavior happens on main and also happens if the branch name is fake-main-for-testing

jalcine-cfa (approved):
:shipit:

PR #3639, opened 2023-08-24
Remove Questions::WelcomeController from gyr_questions_navigation, delete welcome_controller.rb and view, update home Get Started button link to triage_personal_info_questions_path

nishinelson:


tgrathwell:
The somewhat more graceful thing, in case there's any external links to /welcome, is to keep WelcomeController itself but have it redirect to the first page in the GYR flow.

Also per the comment Asheesh made in the story, we happened to look at this story a little bit and found there's going to be a problem with `returning_filer_spec`:

currently when you click "Get Started", if you already have a completed intake in your session it redirects you to the portal

with this change that no longer happens, because the new first page `Questions::TriagePersonalInfoController` overrides `current_intake` to be a fresh new intake instance

so either we change the expectation for that returning client experience, or change `TriagePersonalInfoController` to ensure it can still redirect to the portal if there's a completed intake in the session

app/controllers/questions/personal_info_controller.rb
```diff
@@ -19,6 +19,10 @@ def visitor_record
       Intake.find_by(id: session[:intake_id]) || current_intake
     end
 
+    def prev_path
```
tgrathwell, line 22: It's usually a bummer when a controller has to care about it's neighbors by overriding next_path/prev_path

A better way here would be for WelcomeController to declare that it is never shown:
```
def self.show?(_intake, _current_controller)
  false
end
```

...then the normal implementation of `prev_path` for PersonalInfoController would not see it, and return nil

Another alternative is to remove `WelcomeController` from the navigation order completely, but then you'd have to find some other way to preserve its routes in routes.rb.

If it remains in `GyrNavigationOrder`, you might want to put a comment next to it saying it's deprecated / exists only for redirect / will someday be removed.

app/views/questions/welcome/edit.html.erb
```diff
@@ -0,0 +1,33 @@
+<% content_for :page_title, t("views.questions.welcome.title") %>
```
tgrathwell, line 1: the view (and associated i18n string) doesn't need to exist anymore if the controller is just a redirect

PR #3640, opened 2023-08-25
Store SMS Opt Out Notices to the Hub

jalcine-cfa:


app/models/system_note/sms_opt_out.rb
```diff
@@ -0,0 +1,31 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (client_id => clients.id)
+#  fk_rails_...  (user_id => users.id)
+#
+class SystemNote::SmsOptOut < SystemNote
+  def self.generate!(client:)
+    create!(
+      client: client,
+      body: 'Client replied "STOP" to opt out of text messages'
```
tgrathwell, line 28: For modern `SystemNote` records the preferred pattern is to store `data` about the event and compute the copy from that data in the [_note partial](https://github.com/codeforamerica/vita-min/blob/main/app/views/hub/notes/_note.html.erb) (one example is [CtcPortalAction](https://github.com/codeforamerica/vita-min/blob/main/app/models/system_note/ctc_portal_action.rb))

From the Twilio docs it looks like ["end" "cancel" etc are all viable alternatives to "STOP"](https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-#h_01FBWGDY78H0PYY28VC341JPKE). Capturing the actual message text in the `data` attribute would allow us to build the note text more accurately (not the hardcoded STOP) and the note copy could change without having to update the DB for all old notes

jalcine-cfa: Excellent highlight! I'll add that just now

jalcine-cfa: Reading that post, it looks like `OptOutType` is an _alias_ for the action taken (like if someone sent "END"), the opt-out type value would be "Stop". I got that understanding from https://support.twilio.com/hc/en-us/articles/360034798533-Getting-Started-with-Advanced-Opt-Out-for-Messaging-Services around this part (emphasis mine):

> Twilio will automatically handle replies to the keywords you've configured, but we will still let you know when you receive a message containing these keywords. If you have configured your Messaging Service to send incoming messages to a webhook, Twilio will POST the opt-out, help, or opt-in messages to your webhook to help you track these events. **These will arrive with an extra parameter OptOutType, which will indicate whether the message is a STOP, HELP or START message**. The incoming keyword messages will also appear in your SMS Logs in Twilio Console. Additionally, the OptOutType parameter is not included in the webhook payload unless you have Advanced Opt-Out enabled on the Messaging Service.

tgrathwell: yeah that's my reading too, it seems like it's okay to compare `OptOutType` only against "Stop", but the `Body` of the incoming message could still be STOP, END, CANCEL etc and that's what we could choose to store as the system note's data

jalcine-cfa: Cool! Made that change to store the body in https://github.com/codeforamerica/vita-min/pull/3640/commits/360a3e3a055217bc130135522def4a5bc795f5f4

tgrathwell (approved):

PR #3648, opened 2023-08-31
Add list of counties [#185826466]

alaroia-cfa:
I chose this technical approach because it is fairly similar to how we load `Efile::Relationship` data, see `config/initializers/efile_relationships.rb`.

Additionally, it makes it easy to update the data if it changes next tax year, since it covers tax year 2022, but anything could happen in the future.

I validated that this works in `rails c`. See screenshot.

image

See Pivotal Tracker story for more information about where the data comes from and how it gets used in IT-201.


alaroia-cfa:
Thanks @tgrathwell for the review.

I will be somewhat blindly optimistic despite your smart ideas and merge this as-is.

PR #3649, opened 2023-09-06
NYS 2022 tax math [#185950068]

alaroia-cfa:


app/lib/efile/ny/it201.rb
```diff
@@ -0,0 +1,68 @@
...
+      # The purpose is that this stores the lines already filled out either
+      # by the client or through data imports from the IRS Form 1040.
+      attr_reader :lines
+
+      def initialize(year, filing_status, lines, it227)
+        @year = year
+
+        @filing_status = filing_status # single, mfj, that's all we support for now
+        @lines = ActiveSupport::HashWithIndifferentAccess.new(lines)
+        validate_lines
```
tgrathwell, line 16: not sure whether we should be doing this validation here vs expecting all values have been validated beforehand either by being part of the 1040 or validated before going into our DB

app/lib/efile/ny/it201.rb
```diff
@@ -0,0 +1,68 @@
...
+        @lines["60E"] = it227.calculate()["part2_line1"]
```
tgrathwell, line 17: should this be happening in .calculate instead?

tgrathwell:
seems like the example isn't doing much in this case so it's hard for me to see how this code will work in practice

Maybe we can take a line that actually does some calculations and see how it would look in this format

I'm also most interested in how someone could use this calculator to identify how specific lines got calculated... like will we have methods called "calculate_line_47" that can be linked to etc

PR #3650, opened 2023-09-12
submission pdf button to state file

jenny-heath:


app/controllers/state_file/submission_pdfs_controller.rb
```diff
@@ -0,0 +1,31 @@
+module StateFile
+  class SubmissionPdfsController < ApplicationController
+    def show
+      @submission = EfileSubmission.find_by(id: params[:id])
```
alaroia-cfa, line 4: Let's wrap this in a `current_client.find(...)` or `current_intake.find(...)` or similar, to ensure we're only looking at the current client's data.

I took a look at how to do that, and I got a little confused about the best way to do it. Maybe it does make sense to subclass the questions controller.

alaroia-cfa: I wrote a test locally covering that case, but I didn't push it to GitHub yet because I wanted to start by seeing how we handle other scenarios, and I figured I could turn my ideas into GitHub comments.

spec/controllers/state_file/submission_pdfs_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    context "when the pdf has not been created" do
+      it "creates the pdf and then shows it" do
+        get :show, params: { id: efile_submission.id }
+
+        expect(CreateSubmissionPdfJob).to have_received(:perform_now)
+
+        tempfile = Tempfile.new('output.pdf')
+        tempfile.write(response.body)
+        expect(filled_in_values(tempfile.path)).to match(a_hash_including("TP_first_name" => "Jerry"))
```
alaroia-cfa, line 42: Wow, this means it worked. Cooooooool

alaroia-cfa (commented):
Some comments! Happy to pair on making more progress.

PR #3651, opened 2023-09-12
Render all emails on automated message page in HTML version

embarnard:


embarnard:
This PR adds the HTML rendered version of automated emails 
Screen Shot 2023-09-12 at 11 03 30 AM

via https://gyr-review-app-3651-36ceb2496017.herokuapp.com/en/hub/automated_messages

tgrathwell:
Screen Shot 2023-09-12 at 2 04 15 PM

frustratingly when I visit the review app the auto-resize on most of the iframes isn't working

I know we had this problem when initially starting to do this iframe stuff but I was never able to really figure anything out because once I investigate the issue it starts working again 😢 

It might work better if we enhance our resizey `<script>` to be this:
```
      document.querySelectorAll('iframe').forEach((iframe) => {
          if (iframe.contentDocument.readyState === 'complete') {
              resizeIFrameToFitContent(iframe);
          } else {
              iframe.onload = function () {
                  resizeIFrameToFitContent(iframe);
              }
          }
      });
```

app/controllers/hub/automated_messages_controller.rb
```diff
@@ -52,14 +37,34 @@ def index
...
-        args = message[1]
-        args.present? ? message_class.new(args) : message_class.new
+      automated_messages_and_mailers = automated_messages.map do |m|
+        message = m[0].new
+        replacement_args = {
+          body: message.email_body(**m[1]),
+          client: gyr_client,
+          preparer: User.first,
+          tax_return: gyr_client.tax_returns.first }
+        replaced_body = ReplacementParametersService.new(**replacement_args).process
```
tgrathwell, line 51: Personally I would prefer if the `<<Client.ClientId>>` etc things got left uninterpolated... 

otherwise we have to set up all these test objects and invoke the replacement params service and whatnot, which seems like a lot of extra code just to show fake data

Maybe instead of invoking the real ReplacementParametersService we could just take the body and escape any of the `<<` and `>>` so they become &lt;&lt; or whatever and are still visible within the iframe

embarnard:
> Screen Shot 2023-09-12 at 2 04 15 PM
> 
> frustratingly when I visit the review app the auto-resize on most of the iframes isn't working
> 
> I know we had this problem when initially starting to do this iframe stuff but I was never able to really figure anything out because once I investigate the issue it starts working again 😢
> 
> It might work better if we enhance our resizey `<script>` to be this:
> 
> ```
>       document.querySelectorAll('iframe').forEach((iframe) => {
>           if (iframe.contentDocument.readyState === 'complete') {
>               resizeIFrameToFitContent(iframe);
>           } else {
>               iframe.onload = function () {
>                   resizeIFrameToFitContent(iframe);
>               }
>           }
>       });
> ```

I tried the code you suggested and it didn't help the resizing for me unless I'm missing something so instead I tried putting the email text and sms in their own separate flex container/row and increased the width of the iframe. Seemed to auto-resize after that. 

Screen Shot 2023-09-12 at 9 32 49 PM

app/controllers/hub/automated_messages_controller.rb
embarnard: Yeah I agree the fake data does seem a little annoying to keep up with. Perhaps equally as much annoying extra code to show fake data but more comprehendible to me would be to add some fake replacement service methods like I did in this updated commit. If this is still bad though I change to escape `<<` @tgrathwell

tgrathwell:
It still wasn't doing the resizing for me on the heroku app so I added that extra resizing code to the PR. It seems like it's working more often now, but there's definitely something quirky about iframe sizing that makes this hard to debug.

I also noticed we were still showing the old plain div HTML versions alongside the iframey ones so I deleted those from the template.

Also my point about the replacements was that I don't think we really need to show anything more 'real' than <<Client.ClientId>> -- those tokens should have meaning to anyone in the admin area, they're respected when you send bulk messages and whatnot.[ So I changed the replacement parameters stuff to just escape '<<' and '>>'](https://github.com/codeforamerica/vita-min/pull/3651/commits/fece661beb2a1fd556c51c62c236b1f700a74cdf) so they're visible in iframes and leave it at that, without any specific fake data.

tgrathwell (approved):

embarnard:
> It still wasn't doing the resizing for me on the heroku app so I added that extra resizing code to the PR. It seems like it's working more often now, but there's definitely something quirky about iframe sizing that makes this hard to debug.
> 
> I also noticed we were still showing the old plain div HTML versions alongside the iframey ones so I deleted those from the template.
> 
> Also my point about the replacements was that I don't think we really need to show anything more 'real' than <<Client.ClientId>> -- those tokens should have meaning to anyone in the admin area, they're respected when you send bulk messages and whatnot.[ So I changed the replacement parameters stuff to just escape '<<' and '>>'](https://github.com/codeforamerica/vita-min/pull/3651/commits/fece661beb2a1fd556c51c62c236b1f700a74cdf) so they're visible in iframes and leave it at that, without any specific fake data.

@tgrathwell In the [pivotal story ](https://www.pivotaltracker.com/story/show/184847561) it says "As a bonus it'd be cool to see both html and text versions." which is why I included the text version as well. I like having the variables replaced with actual text because thats how the other mailers were rendered the emails above before this story and also if we also have the text version it make more sense to have this contrast. But ultimately I think the <<>> versus text rendering is just a matter of preference maybe more up to program or who ever is going to be looking at this than anything.

PR #3656, opened 2023-09-15
State file: ny201 - xml and pdf up through header/filing status/claimed as dep

tgrathwell:
added sanity-check tests for ny and az to ensure the output of hash_for_pdf is only fields that actually exist in the pdf


jenny-heath (approved):
al & jh say yeah

PR #3660, opened 2023-09-25
More NYS calculation [#186044880]

alaroia-cfa:


app/lib/efile/ny/it201.rb
```diff
@@ -3,63 +3,70 @@ module Ny
...
 
-      def initialize(year:, filing_status:, claimed_as_dependent:, dependent_count:, input_lines:, it227:)
+      def initialize(year:, filing_status:, claimed_as_dependent:, dependent_count:, input_lines:, it213:, it214:, it215:, it227:)
         @year = year
 
         @filing_status = filing_status # single, married_filing_jointly, that's all we support for now
         @claimed_as_dependent = claimed_as_dependent # true/false
         @dependent_count = dependent_count # number
+
+        @value_access_tracker = ValueAccessTracker.new
```
alaroia-cfa, line 13: I think I'd prefer if the caller is responsible for setting up the `ValueAccessTracker`. However, I haven't fully taken the time to understand the new `set_line` API yet.

app/lib/efile/ny/it201.rb
```diff
@@ -3,63 +3,68 @@ module Ny
...
-        set_line(:AMT_69, calculate_line_69)
-        set_line(:AMT_69A, calculate_line_69a)
-        set_line(:AMT_70, calculate_line_70)
-        set_line(:AMT_73, calculate_line_72)
-        set_line(:AMT_73, calculate_line_73)
-        set_line(:AMT_76, calculate_line_76)
-        set_line(:AMT_77, calculate_line_77)
-        set_line(:AMT_78, calculate_line_78)
-        set_line(:AMT_78B, calculate_line_78b)
-        set_line(:AMT_80, calculate_line_80)
```
alaroia-cfa, line 55: If `set_line` took a block, we wouldn't have to mess with lambdas. Is that a better, more Ruby-ish API?

app/lib/efile/ny/it201.rb
```diff
@@ -3,63 +3,68 @@ module Ny
     class It201
       attr_reader :lines
 
-      def initialize(year:, filing_status:, claimed_as_dependent:, dependent_count:, input_lines:, it227:)
+      def initialize(year:, filing_status:, claimed_as_dependent:, dependent_count:, input_lines:, it213:, it214:, it215:, it227:, value_access_tracker:)
```
alaroia-cfa, line 6: I made a few changes here:

- Rename `@collaborators` to `@inputs` (I think inputs is a more straightforward word in this context)
- Change ValueAccessTracker so it's created earlier in the timeline, rather than created in this initialize (this is a 'worse is better' approach to initialization - make your callers do more work and then they are confident in the implementation)
- Unify all TaxFormLine classes into one, since their behavior changes are limited now to factory methods

alaroia-cfa (commented):
Some comments & some code changes; we don't have to keep my code changes but I do want to know what you think of them.

One big (?) question remains about the `set_line` API - should we change it to take a block?

```set_line(:AMT_17) { calculate_line_35 }```

I think that would be more Ruby-ish than the lambdas. I'm curious what you think.

**Comment by alaroia-cfa** [comment]
I forgot to say -- overall I like this ValueTracker approach.

I'm not blocked, in that I can keep adding NYS tax math within this PR.

File: app/lib/efile/ny/it201.rb (hunk 3 above)
Comment (tgrathwell) [reply app/lib/efile/ny/it201.rb:6]: I think I agree with everything but creating `ValueAccessTracker` outside -- it is **extremely** an implementation detail of the tax calculator / tax form line classes, and the less things that need to be aware of it the better

I understand the light ickyness in assigning a value_access_tracker to already-instantiated lines, but it seems like a good tradeoff to keep references to it isolated

File: app/lib/efile/ny/it201.rb (hunk 2 above)
Comment (tgrathwell) [reply app/lib/efile/ny/it201.rb:55]: blocks would be more natural than procs, but I dunno if we can do a `.source` of a block

though I'm kind of thinking we should just make a `calculate_line_x` for every line even if it's trivial, for consistency's sake --- `method(:calculate_line_4).source` produces something better to look at than sourcing a proc as it turns out

File: app/lib/efile/ny/it201.rb (hunk 3 above)
Comment (alaroia-cfa) [reply app/lib/efile/ny/it201.rb:6]: Interesting, yeah. I'm OK with restoring the `value_access_tracker=` stuff.

It's true that value access tracking is an implementation detail. I wonder if there's yet another option waiting for us which accomplishes even more of our shared goals.

File: app/lib/efile/ny/it201.rb (hunk 2 above)
Comment (alaroia-cfa) [reply app/lib/efile/ny/it201.rb:55]: Internet people say .source works for a block, so I might try that route.

Comment (alaroia-cfa) [reply app/lib/efile/ny/it201.rb:55]: > I'm kind of thinking we should just make a calculate_line_x for every line even if it's trivial, for consistency's sake

I like this approach, and I like it more than the `block.source` hypothetical approach, too.

**Comment by alaroia-cfa** [comment]
General comment for a future pull request:

I'd like to see some clear answers for how we'll do 'explain calculations' for the tax math we perform in the other forms, since those are going to be complicated eventually.

I'd like, if possible, a way for the 'explain calculations' system to integrate with the `table_rows.reverse.find { ... }` system, so we can say 'we picked xyz row in the table due to income abc'.

I don't need that to be part of this pull request; just sharing that it's on my mind; if you have ideas, I definitely want to hear about 'em.

All the various little systems are still in flux so it's harmless not to have it answered yet.

**Comment by tgrathwell** [comment]
@alaroia-cfa same same

I think a good goal of all this line-tracking shenanigans should be that AMT_63 reports that it depends on IT213_AMT_16 (which itself is also in the table and depends on other IT213_AMT values)

so whatever design facilitates that will be good 👍

File: app/lib/efile/ny/it201.rb (hunk 3 above)
Comment (alaroia-cfa) [reply app/lib/efile/ny/it201.rb]: Restored

File: app/lib/efile/ny/it201.rb (hunk 1 above)
Comment (alaroia-cfa) [reply app/lib/efile/ny/it201.rb:12]: I restored `.value_access_tracker=`

File: app/lib/efile/ny/it201.rb [hunk 4]
```diff
@@ -64,7 +64,18 @@ def calculate
...
+        method_name = "calculate_#{line_id.to_s.sub('AMT_', 'line_').downcase}".to_sym
+        method =
+          begin
+            Efile::Ny::It201.instance_method(method_name)
+          rescue NameError
+            nil
+          end
+        source_description = [value_fn.source, method&.source].compact.join("\n\n")
+
+        value = @value_access_tracker.with_tracking { value_fn.call }
```
alaroia-cfa, line 76: FYI @tgrathwell `.with_tracking` replaces the need for `.clear`, or so I claim

PR #3666, opened 2023-09-27
Create rake task to upload dependent/spouse ssn data

embarnard:
Co-authored-by: Tahsina Islam <tislam@codeforamerica.org>

### DIFF
--- a/lib/tasks/data_task_spouse_dependent_ssns.rake
+++ b/lib/tasks/data_task_spouse_dependent_ssns.rake
@@ -1,64 +1,43 @@
 require 'csv'
 
-namespace :socure do
-  desc "create socure CSV file"
-  task create_client_file: :environment do
-    # client_ids = []
-    # file_in = File.open("#{Rails.root}/lib/tasks/client_ids.csv")
-    # in_data = CSV.read(file_in, headers: true)
-    # in_data.each do |record|
-    #   record = record.to_ary
-    #   client_id = record[0][1]&.strip
-    #   client_ids << client_id.to_i
-    # end
+namespace :data_task_spouse_dependent_ssns do
+  desc "create file of unhashed SSNs and respective unhashed spouse and dependent SSNs"
+  task create_file: :environment do
+    file_in = File.open("#{Rails.root}/tmp/ca_ids_for_eng_nov02.csv")
+    in_data = CSV.read(file_in, headers: true)
+    client_ids = []
+    in_data.each do |record|
+      record = record.to_ary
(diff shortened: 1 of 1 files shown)


lib/tasks/data_task_spouse_dependent_ssns.rake
```diff
@@ -0,0 +1,43 @@
...
+    client_ids = []
+    in_data.each do |record|
+      record = record.to_ary
+      client_id = record[0][1]&.strip
+      client_ids << client_id.to_i
+    end
+
+    temp_file = Tempfile.open(%w[temp_file .csv], "tmp/")
+    CSV.open(Rails.root.join(temp_file.path), 'wb') do |csv|
+      csv << %w{client_id ssn type}
```
embarnard, line 17: switch ssn and type

lib/tasks/data_task_spouse_dependent_ssns.rake
```diff
@@ -0,0 +1,43 @@
...
+    blob = ActiveStorage::Blob.create_and_upload!(
+      io: File.open(Rails.root.join(temp_file.path), 'rb'),
+      filename: "spouse_dependents_results.csv",
+      content_type: "text/csv"
+    )
+
+    puts "*******FILE KEY: #{blob.key}"
+    puts blob.inspect
+    puts "***********blob URL:"
+    temp_file.unlink
```
embarnard, line 40: probably not needed

lib/tasks/data_task_spouse_dependent_ssns.rake
```diff
@@ -0,0 +1,43 @@
...
+      record = record.to_ary
+      client_id = record[0][1]&.strip
+      client_ids << client_id.to_i
+    end
+
+    temp_file = Tempfile.open(%w[temp_file .csv], "tmp/")
+    CSV.open(Rails.root.join(temp_file.path), 'wb') do |csv|
+      csv << %w{client_id ssn type}
+
+      client_ids.each do |id|
```
embarnard, line 19: start with `Client.where(id: client_ids).includes(:intake, :dependents)` and each over that, so we don't have to fetch them one-by-one

maybe do .find_in_batches

lib/tasks/data_task_spouse_dependent_ssns.rake
embarnard: https://apidock.com/ruby/Tempfile seeing here it is good practice to explicitly close and unlink and not wait for garbage collection

tgrathwell (approved):

PR #3669, opened 2023-09-27
Move DependentsController into Questions::DependentsController

tgrathwell:
it should be with its other GYR intake friends

DependentsController itself sticks around as a pure redirect to the Questions:: version

keeping the translation ymls in the same place for now; I don't think there's any relative references we need to move to fix, maybe the tests will tell me


alaroia-cfa:
lgtm obvi

PR #3674, opened 2023-10-03
State file: tax calculators: set_line no longer tries to look up source via string messiness

tgrathwell:
now every calculate_x invocation uses set_line with a symbol


app/lib/efile/tax_calculator.rb
```diff
@@ -4,22 +4,19 @@ class TaxCalculator
 
     private
 
-    def set_line(line_id, value_fn)
+    def set_line(line_id, value_fn_or_data_source, field = nil)
       source_description = nil
-      if value_fn.is_a?(Symbol)
-        value_fn = method(value_fn)
+      if field
```
alaroia-cfa, line 9: Since `field` is only available when `value_fn` is not a callable, I wonder if there's a way to avoid the optional argument. Not a blocker.

alaroia-cfa (approved):
Looks good to me. One non-blocking question. Overall, it's a big improvement to pass the direct file data down into each class.

PR #3677, opened 2023-10-04
State File: only generate .source_description for tax lines when on the /explain_calculations route

tgrathwell:
Currently the calculator is used to do XML, PDF, and /explain_calculations

Since method(...).source is prone to crash if not cared for, we should do it only when needed. Hence now the calculator objects can take an `include_source` param.

also fixup the AZ calculator object to no longer receive input_lines, it can find the stuff off direct_file_data itself


alaroia-cfa (approved):
Nice

It adds a little more complexity, but it buys us a little less worrying about `.source`, and that's worth it.

PR #3684, opened 2023-10-09
State File: unify NY calculator classes around :IT123_LINE_1 naming convention

tgrathwell:


jenny-heath (approved):
looks pretty straightforward

PR #3685, opened 2023-10-10
State File: ny IT213 XML

tgrathwell:


app/lib/pdf_filler/ny213_pdf.rb
```diff
@@ -19,9 +19,86 @@ def hash_for_pdf
       answers = {
         'Your last name' => @submission.data_source.primary.full_name,
         'Your SSN' => @submission.data_source.primary.ssn,
-        'Line 6 Dollars' => @calculator.lines[:IT213_LINE_6].value
+        'Spouse \'s last name' => @submission.data_source.primary.ssn,
+        'Spouse \'s SSN' => @submission.data_source.primary.ssn,
```
tahsinaislam, line 23: this should be spouse ssn instead of primary

app/lib/pdf_filler/ny213_pdf.rb
```diff
@@ -19,9 +19,86 @@ def hash_for_pdf
       answers = {
         'Your last name' => @submission.data_source.primary.full_name,
```
embarnard, line 20: also full name or just last name?

embarnard (approved):
I think this looks good! Made some changes to the top so that it calls for last name (although I saw in the [PDF ](https://www.tax.ny.gov/pdf/current_forms/it/it213_fill_in.pdf)it says full name so maybe I should change it back and also changed to call for `spouse.ssn` instead of `primary.ssn` twice

PR #3691, opened 2023-10-18
Start adding state styles #186264946

jenny-heath:
first commit: how it would look if we keep all the new styles in one file mostly

second commit: how it would look to disperse the new styles into our existing files

Screen Shot 2023-10-18 at 10 11 06 AM

Screen Shot 2023-10-18 at 10 12 23 AM


jenny-heath:
update: i reverted the first commit and have decided to keep the styles separate for now

PR #3693, opened 2023-10-19
Edit .pairs file and refactor GYR homepage banner code

embarnard:

PR #3694, opened 2023-10-20
Fix time-related test about offseason filing banner

tgrathwell:


tgrathwell:
Closed in favor of #3693

PR #3713, opened 2023-10-26
NY permanent address fields

jenny-heath:
image

the inputs look like garbage but as discussed this is none of our business and someone will style them later
image


jenny-heath:
closing & opening to see if it fixes the review app

app/assets/stylesheets/_state-file.scss
```diff
@@ -57,6 +55,12 @@
       margin-right: 1rem;
     }
   }
+
+  .text--body {
+    font-size: 1.6rem;
+    font-weight: 700;
+    line-height: 2.4rem;
+  }
```
jenny-heath, line 63: just alphabetizing

app/forms/state_file/ny_permanent_address_form.rb
```diff
@@ -12,6 +12,15 @@ class NyPermanentAddressForm < QuestionsForm
     validates :permanent_city, presence: true, if: -> { confirmed_permanent_address == "no" }
     validates :permanent_zip, presence: true, if: -> { confirmed_permanent_address == "no" }
 
+    def initialize(intake = nil, params = nil)
+      if params[:confirmed_permanent_address] == "yes"
+        [:permanent_apartment, :permanent_street, :permanent_city, :permanent_zip].each do |param|
+          params[param] = ""
+        end
+      end
```
jenny-heath, line 20: i bet we do something like this (conditionally remove/modify params from the form) elsewhere but i couldn't find an example

jenny-heath: either way let me know if this is good, neutral, or janky

app/views/state_file/questions/federal_dependents/edit.html.erb
```diff
@@ -1,4 +1,13 @@
 <% @allow_deletion = true %>
-<% content_for :card do %>
-  <%= render "state_file/questions/federal_dependents/form" %>
+
+<% content_for :main do %>
```
jenny-heath, line 3: my thinking is that the edit and new pages should not have the "go back" button because it will take them to the previous question in the flow (permanent address), NOT the previous page they were on (dependents index)

app/views/state_file/questions/federal_dependents/index.html.erb
```diff
@@ -3,39 +3,31 @@
 
 <% content_for :body_class, "dependents-index" %>
```
jenny-heath, line 5: this looks like a lot of changes but i'm just replacing `content_for :main` with `content_for :card` and removing the redundant wrapping elements

app/views/state_file/questions/ny_permanent_address/edit.html.erb
```diff
@@ -35,6 +35,7 @@
             <%= f.cfa_input_field(:permanent_street, "Street Address") %>
             <%= f.cfa_input_field(:permanent_apartment, "Apartment/Unit Number") %>
             <%= f.cfa_input_field(:permanent_city, "City") %>
+            <%= f.cfa_input_field(:unfillable, "State", options: { value: "New York", disabled: true }) %>
```
jenny-heath, line 38: this is in the mock but we missed it. it's not editable

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -27,8 +27,17 @@
...
+      choose "Yes"
+      click_on "Continue"
+
+      expect(page).to have_text "The page that shows your dependents"
+      click_on "Go back"
+
       expect(page).to have_text "Was this your permanent home address on December 31, 2023?"
       choose "No"
+      # if they previously confirmed their address from DF, don't show it filled in on the form for a new permanent address
+      expect(find_field("state_file_ny_permanent_address_form[permanent_street]").value).to eq ""
```
jenny-heath, line 40: discussed this with carmen and this is the pattern she wanted

app/forms/state_file/ny_permanent_address_form.rb
```diff
@@ -1,17 +1,36 @@
...
+                       :confirmed_permanent_address,
+                       :permanent_apartment,
+                       :permanent_street,
+                       :permanent_city,
+                       :permanent_zip
+
+    validates :confirmed_permanent_address, presence: true
+    validates :permanent_street, presence: true, if: -> { confirmed_permanent_address == "no" }
+    validates :permanent_city, presence: true, if: -> { confirmed_permanent_address == "no" }
+    validates :permanent_zip, presence: true, if: -> { confirmed_permanent_address == "no" }
```
bengolder, line 13: do we have an existing zip code validation?

app/forms/state_file/ny_permanent_address_form.rb
bengolder: this looks good, but it might be helpful to document this with a unit spec

app/views/state_file/questions/federal_dependents/edit.html.erb
bengolder: I think this is fine for now, though I imagine if we stick with this dependents flow, then dependents new & edit should link back to dependents index, and the index should link back to whatever the prior page was.

bengolder (approved):
This is great! Let’s ship it! I think it would be helpful to follow up with one additional unit test for clearing the address fields after a “yes” answer, but I don’t think that warrants blocking the pr. Approving and merging.

PR #3717, opened 2023-10-26
State file family DOB (and months in home) page

embarnard:


app/controllers/state_file/questions/az_dependents_dob_controller.rb
```diff
@@ -0,0 +1,35 @@
+module StateFile
+  module Questions
+    class AzDependentsDobController < QuestionsController
+      layout "state_file/question"
+
+      def edit
+        @intake = current_intake
+        @dependents = @intake.dependents
+        @form = form_class.from_intake(@intake)
```
jenny-heath, line 9: i think you can delete this whole override. you should have access to `current_intake` in the view, which would also give you access to dependents, and the form instantiation here appears to be identical to what's in questions_controller

app/controllers/state_file/questions/az_dependents_dob_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def update
+        @form = form_class.new(current_intake, form_params)
+        if @form.save && @form.valid?
+          redirect_to next_path
+        else
+          render :edit
+        end
```
jenny-heath, line 18: ditto this one: the whole thing can go because the parent implementation does all this

app/controllers/state_file/questions/az_dependents_dob_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      def illustration_path; end
+
+      private
+
+      def form_params
+        params.require(:state_file_az_dependents_dob_form).permit(dependents_attributes: [:id, :dob_month, :dob_day, :dob_year, :months_in_home])
+      end
+
+      def form_class
+        StateFile::AzDependentsDobForm
```
jenny-heath, line 30: this is also the same as the inherited method, can be deleted

app/forms/state_file/az_dependents_dob_form.rb
```diff
@@ -0,0 +1,41 @@
...
+      super
+    end
+
+    def dependents
+      @intake.dependents
+    end
+
+    def save
+      @intake.update!(state_file_dependents_attributes: formatted_dependents_attributes)
+      @intake.save!
```
jenny-heath, line 18: i think save is redundant with update so maybe just do the update

app/models/state_file_az_intake.rb
```diff
@@ -58,4 +60,8 @@ def qualifying_parents_and_grandparents
   def sentenced_for_60_days
     # TODO
   end
+
+  def most_recent_filing_year
+    2023
+  end
```
jenny-heath, line 66: maybe this should live in a config variable instead idk

jenny-heath:
trying to get the heroku app working

app/views/state_file/questions/dob/edit.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+<% content_for :card do %>
+  <p><%= t('.title1') %></p>
+  <% if @form.ask_primary_dob? %>
+    <p><%= t('.title2_you_and_household') %></p>
+  <% else %>
+    <p><%= t('.title2_family') %></p>
+  <% end %>
+
+  <%= form_with model: @form, url: { action: :update }, method: :put, local: true, builder: VitaMinFormBuilder, class: 'form-card' do |f| %>
```
embarnard, line 11: I found a bug, if there are no dependents New York does not show primary DOB

PR #3725, opened 2023-10-27
Encrypt direct file data 186318125

loumoore:
The point of this change was to add rails model encryption to the raw data for NY and AZ

I thought we should encrypt the banking fields too and I went down a rabbit hole adding them to the NY xml and pdf while I was at it. It was helpful for me to learn at least a little bit of what is going on there, but the second commit for that is more complicated, so if there are issues with it I could revert it and just submit the encryption stuff.


app/controllers/state_file/questions/federal_info_controller.rb
```diff
@@ -66,7 +66,7 @@ def create_sample_intake
             amount_owed_pay_electronically: "yes",
             refund_choice: "paper",
             account_type: "personal_checking",
-            routing_number: "34567878",
+            routing_number: "013456789",
```
loumoore, line 69: make our default here a valid routing number

app/forms/state_file/ny201_form.rb
```diff
@@ -27,6 +27,8 @@ class Ny201Form < QuestionsForm
                        :primary_signature,
                        :spouse_signature
 
+    validates :routing_number, length: { is: 9 }, allow_blank: true, routing_number: true
```
loumoore, line 30: add validation so we don't get bunk routing numbers, but still allow it to be blank

app/lib/pdf_filler/ny201_pdf.rb
```diff
@@ -93,13 +93,16 @@ def hash_for_pdf
...
       )
+      unless @xml_document.at('ACCT_TYPE_CD').nil?
+        answers.merge!(
+          Line83a_account: xml_value_to_pdf_checkbox("Line83a_account", 'ACCT_TYPE_CD'),
+          Line83b_routing: claimed_attr_value('ABA_NMBR'),
+          Line83c_account_num: claimed_attr_value('BANK_ACCT_NMBR'),
+          Line84_withdrawal_Date: claimed_attr_value('ELC_AUTH_EFCTV_DT'),
+          Line84_withdrawal_amount: claimed_attr_value('PYMT_AMT'),
+        )
+      end
```
loumoore, line 105: added the account type checkbox, but only attempt to write all this to pdf if it is present

app/lib/seeder.rb
```diff
@@ -148,7 +148,7 @@ def run
       sms_notification_opt_in: :yes,
       email_notification_opt_in: :yes,
       bank_name: "Self-help United",
-      bank_routing_number: "12345678",
+      bank_routing_number: "011234567",
```
loumoore, line 151: valid routing number

app/lib/submission_builder/ty2022/states/ny/documents/rtn_header.rb
```diff
@@ -0,0 +1,85 @@
+module SubmissionBuilder
```
loumoore, line 1: hopefully i interpreted the xml spec right. it seems like there's a rtnHeader element allowed under the composition>forms section of the NY state data that's separate from the top level return header. this is the start of an implementation for it with just the required fields and the banking info if present

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -109,6 +109,11 @@ def supported_documents
             calculated_fields = tax_calculator.calculate
             receiving_214_credit = calculated_fields[:IT214_LINE_33] > 0
             supported_docs = [
+              {
+                xml: SubmissionBuilder::Ty2022::States::Ny::Documents::RtnHeader,
+                pdf: PdfFiller::Ny201Pdf,
+                include: true
+              },
```
loumoore, line 116: add the new rtnHeader element for NY

app/models/state_file_az_intake.rb
```diff
@@ -24,7 +24,7 @@
 #  visitor_id             :string
 #
 class StateFileAzIntake < StateFileBaseIntake
-
+  encrypts :bank_account_number, :bank_routing_number, :raw_direct_file_data
```
loumoore, line 27: here's the actual encryption change for AZ

app/models/state_file_ny_intake.rb
```diff
@@ -60,6 +60,7 @@
 #  visitor_id                     :string
 #
 class StateFileNyIntake < StateFileBaseIntake
+  encrypts :account_number, :routing_number, :raw_direct_file_data
```
loumoore, line 63: here's the actual encryption change for NY

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -47,6 +47,8 @@
       expect(page).to have_link "Download PDF"
       click_on "Show XML"
       expect(page.body).to include('efile:ReturnState')
+      expect(page.body).to include('<ABA_NMBR claimed="013456789"/>')
+      expect(page.body).to include('<BANK_ACCT_NMBR claimed="456789008765"/>')
```
loumoore, line 51: not sure this is really necessary since rails encryption is presumably tested at the framework level, and even the xml snippet "efile:ReturnState" we're already checking for is being encrypted now. but routing and account number are encrypted too outside of the data blob so i threw a check in there.

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -82,6 +84,7 @@
       expect(page).to have_text "You have successfully submitted your taxes"
       click_on "Show XML"
       expect(page.body).to include('efile:ReturnState')
+      expect(page.body).to include('<FirstName>Testy</FirstName>')
```
loumoore, line 87: again maybe not needed but validating we're still getting user data back with encryption

spec/lib/submission_builder/ty2022/states/ny/individual_return_spec.rb
```diff
@@ -34,5 +34,18 @@
         expect(xml.at("IT215 E_FED_EITC_IND").attribute('claimed').value).to eq("1")
       end
     end
+
+    context "when getting a refund to a personal checking account" do
+      let(:intake) { create(:state_file_ny_intake, filing_status: filing_status, routing_number: "011234567", account_number: "123456789", account_type: 1) }
+      let(:filing_status) { 'single' }
+
+      it 'generates XML from the database models' do
+        xml = Nokogiri::XML::Document.parse(described_class.build(submission).document.to_xml)
```
tgrathwell, line 43: for what it's worth `described_class.build(submission).document` is already a Nokogiri::XML::Document so we shouldn't need to `.to_xml` it and parse again

but it looks like this pattern is in use plenty of places and I copied it into this file myself so 🤷

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -109,6 +109,11 @@ def supported_documents
             calculated_fields = tax_calculator.calculate
             receiving_214_credit = calculated_fields[:IT214_LINE_33] > 0
             supported_docs = [
+              {
+                xml: SubmissionBuilder::Ty2022::States::Ny::Documents::RtnHeader,
+                pdf: PdfFiller::Ny201Pdf,
```
tgrathwell, line 114: pdf needs to be nil for the RtnHeader document, if we set `pdf` on two documents it will be included in the combined pdf twice :)

loumoore: right on, fixed. thank you.

spec/lib/submission_builder/ty2022/states/ny/individual_return_spec.rb
loumoore: fixed all the instances in this file

PR #3726, opened 2023-10-27
Add NY date of birth collection page for primary, spouse and dependents

embarnard:


spec/factories/state_file_dependents.rb
```diff
@@ -0,0 +1,31 @@
...
+#  suffix         :string
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  intake_id      :bigint           not null
+#
+# Indexes
+#
+#  index_state_file_dependents_on_intake  (intake_type,intake_id)
+#
+FactoryBot.define do
```
embarnard, line 23: if the other DOB story from AZ get merged in before this we should do a rebase bc this is duplicated there

app/views/state_file/questions/ny_dob/edit.html.erb
```diff
@@ -0,0 +1,47 @@
...
+      <div class="date-select spacing-below-5">
+        <%= f.cfa_date_select(:primary_birth_date,
+                              "Date of birth for #{f&.object&.intake&.primary_first_name&.capitalize}",
+                              options: {
+                                start_year: Time.now.year,
+                                end_year: Time.now.year - 100,
+                              }) %>
+      </div>
+    </div>
+    <% if f&.object&.intake&.spouse_first_name&.present? %>
```
embarnard, line 19: is this the best way to check for presence of a spouse?

embarnard: aka not sure if I should use filing status or something else instead

tgrathwell:
Elected to try to cram this functionality in https://github.com/codeforamerica/vita-min/pull/3717 instead and have one unified controller

hopefully this doesn't make story acceptance on these two too weird

PR #3731, opened 2023-10-30
WIP: Database fields for state file login

bengolder:
This adds updates the state file schema
- adds source & referrer columns (for tracking how people arrive at state file). This is data we want to save on the first page where we create an intake
- updates claimed_as_dep to have a default so we can create state intakes before we have an answer


embarnard:
failing annotation test might just need you to run db:migrate to update the schemas

bengolder:
moved to a different branch

PR #3736, opened 2023-10-31
Ny out of state purchases page

embarnard:


spec/i18n_spec.rb
```diff
@@ -43,6 +45,7 @@
 
         (@i18n.locales - [@i18n.base_locale]).each do |other_locale|
           node = @i18n.data[other_locale].first.children[key]
+          next if node.nil?
```
embarnard, line 48: was failing here previously bc would still check for `state_file.questions.ny_sales_use_tax.edit.use_tax_method_manual_html`, I'm a little iffy if this logic is right but it did seem to fix the test

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -40,7 +40,7 @@
       choose "Yes"
       click_on "Continue"
 
-      expect(page).to have_text "The page with all the info from the 201"
+      expect(page).to have_text I18n.t('state_file.questions.ny_sales_use_tax.edit.title', year: "2023")
```
bengolder, line 43: Optionally, this could use `MultiTenantService.statefile.current_tax_year` (very recently added to main), but it's not clear if this project would continue to next year. I don't think there's any good reason to change this, but I just wanted to let you know in case you'd prefer it.

bengolder: Maybe it makes more sense to hardcode the year in the specs anyways

spec/i18n_spec.rb
bengolder: I ran into a similar issue and added some [minimal html ](https://github.com/codeforamerica/vita-min/blob/main/config/locales/es.yml#L1805)to the translation string. Once we have spanish translations, I imagine that we'll want to make sure we have a spanish translation for every english string. That feels like a different test to me than this one which checks whether the html matches. I feel comfortable with this edit.

spec/models/state_file_ny_intake_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+    end
+
+    context "when sales_use_tax_calculation_method changes to automated" do
+      it "calculates sales use tax" do
+        expect {
+          intake.update(sales_use_tax_calculation_method: "automated")
+        }.to change(intake, :sales_use_tax).to(29)
+      end
+    end
```
bengolder, line 91: This is great! I love that you put this on the model in a `before_save` hook. I think that makes a ton of sense and makes me feel confident that the data will always maintain integrity.

spec/models/state_file_ny_intake_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    context "when the federal agi is 700,000" do
+      let(:household_fed_agi) { 700_000 }
+      it "returns 125" do
+        expect(intake.calculate_sales_use_tax).to eq 125
+      end
+    end
+  end
+end
```
bengolder, line 126: This and the method in the model are super clear to me as a reviewer. This is very readable!

bengolder: This could maybe use a test for the escape clause (no fed agi) but that doesn't feel critical to me.

app/forms/state_file/ny_sales_use_tax_form.rb
```diff
@@ -0,0 +1,32 @@
...
+        params[:sales_use_tax] = intake.calculate_sales_use_tax
+      end
+      super(intake, params)
+    end
+
+    def save
+      attributes = attributes_for(:intake)
+      attributes[:untaxed_out_of_state_purchases] = "unfilled" unless attributes[:untaxed_out_of_state_purchases].present?
+      attributes[:sales_use_tax_calculation_method] = "unfilled" unless attributes[:sales_use_tax_calculation_method].present?
+      attributes[:sales_use_tax] = intake.calculate_sales_use_tax if attributes[:sales_use_tax_calculation_method] == :automated
```
bengolder, line 27: I think you already covered this functionality in the model, and your form_spec passes if these are deleted. Maybe they're unnecessary?

bengolder, line 27: I thought that perhaps this logic is handled in the model's before_save hook. The form_spec still passes if these are deleted, but it doesn't look like there are tests that would catch if this or the model didn't adjust other attributes based on the saved value.

app/forms/state_file/ny_sales_use_tax_form.rb
```diff
@@ -0,0 +1,32 @@
...
+    def initialize(intake = nil, params = nil)
+      if params[:untaxed_out_of_state_purchases] == "no"
+        params[:sales_use_tax_calculation_method] = "unfilled"
+        params[:sales_use_tax] = nil
+      end
+      if params[:sales_use_tax_calculation_method] == "automated"
+        params[:sales_use_tax] = intake.calculate_sales_use_tax
+      end
+      super(intake, params)
+    end
```
bengolder, line 21: I think the before_save hook on the model will prevent bad data from going, but I appreciate that this will update the form to reflect the user's choices if they have a validation error and it rerenders the page. Very thoughtful of you to include this.

app/forms/state_file/ny_sales_use_tax_form.rb
bengolder: I feel confident that the model's before_save hook should take care of this. the first two are default values, so if they aren't present when you save the model, they'll be set to unfilled. If I understand the feature correctly, we only need to calculate sales use tax when this gets changed or initial set as "automated". I think this will calculate the sales use tax twice (in the form.save, and in the model.before_save) every time.

bengolder: Also, the validation should prevent anyone from saving with `untaxed_out_of_state_purchases = nil`

app/views/state_file/questions/ny_sales_use_tax/edit.html.erb
```diff
@@ -0,0 +1,39 @@
+<% content_for :page_title, "Out-of-state purchases" %>
```
bengolder, line 1: I think this should use `t('.title', year "2023)`

bengolder (approved):
Looks great! Super readable and clear code, thank you!

I think there are some lines in the form you can safely delete if you want, but they also won't harm anything if you leave them in.

Also, just a heads up, I noticed that this also creates a new file I that recently created too and is now on main (`state_file_ny_intake_spec.rb`), so you might get some merge conflicts. Happy to help walk through those with you if it would be helpful

app/forms/state_file/ny_sales_use_tax_form.rb
embarnard: Oh you're right! thanks for catching this, I think I added this in before adding the model validations and forgot to take it out

spec/features/state_file/complete_intake_spec.rb
embarnard: Ooo yes good idea!

PR #3740, opened 2023-11-01
new questions for AZ 65+ dependents

loumoore:


app/forms/state_file/az_senior_dependents_form.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def dependents
+      @intake.dependents.select(&:ask_senior_questions?)
+    end
+
+    def save
+      @intake.update!({ dependents_attributes: dependents_attributes.to_h })
+    end
+
+    def valid?
```
loumoore, line 21: not sure i'm doing this right. seemed to be something magical/not explicit happening in other similar forms i referenced with dependent attributes

app/controllers/state_file/questions/az_senior_dependents_controller.rb
```diff
@@ -0,0 +1,17 @@
+module StateFile
+  module Questions
+    class AzSeniorDependentsController < QuestionsController
+      def self.show?(intake)
+        !intake.nil? && intake.dependents.select(&:ask_senior_questions?).size.positive?
```
tgrathwell, line 5: Not a huge deal but looking through the existing `show?` implementations in the app, none of them seem to guard against the presence of `intake` -- I think typically they're invoked with a real one or they're not invoked at all

The state file flow is still immature, the other flows typically have a `before_action` installed on them after login that redirects out of the flow if the sessioned intake is nil. Maybe that has something to do with it?

would be curious to know what starts blowing up if the nil-guard is removed

app/forms/state_file/az_senior_dependents_form.rb
tgrathwell: I think the cleanest implementation right now is on DobForm https://github.com/codeforamerica/vita-min/blob/main/app/forms/state_file/dob_form.rb#L34

The form object's `valid?` should return the combination of the form's own validations plus the validations of any dependents. Since each dependent is just an ActiveRecord `StateFileDependent` the dependent validations are set up on the model, though they can be conditional to ensure they're not required every place we create/edit a dependent.

So in this case `StateFileDependent` should probably have something like a `validates_inclusion_of :passed_away, in: ['yes', 'no'], on: :senior_dob_form` and within `AzSeniorDependentsForm` we should call `dependent.valid?(:senior_dob_form)`

It might be more elegant if there was a `StateFileDependentSeniorQuestionsForm` wrapping each StateFileDependent and the form classes were better able to merge validation errors between the layers, but we haven't built anything like that for our existing use cases.

app/controllers/state_file/questions/az_senior_dependents_controller.rb
loumoore: there is a flow_controller_spec or something that blew up without this guard. it was outside the bin/test state_file target so it blew up after I pushed my original change and then I added this. 

I didn't look real closely at whether the test could be changed instead, could look into that

tgrathwell: I think one of Ben's to-be-merged PRs has some changes that might make the flow controller spec thing work better; anyway it's fine to keep in the guard for now maybe it's something we'll notice and remove later.

app/views/state_file/questions/az_senior_dependents/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+<% content_for :card do %>
+  <p><%= t('.title1') %></p>
+
+  <% if @form.errors.any? %>
+    <div class="notice--error">
+      <% @form.errors.full_messages.each do |msg| %>
+        <div><%= msg %></div>
+      <% end %>
+    </div>
+  <% end %>
```
bengolder, line 12: Ideally this should be handled the same way on all pages, but it's helpful to double check and affirm that we need to see the errors.

app/views/state_file/questions/az_senior_dependents/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+          </div>
+        </div>
+        <div class="highlight-box highlight-box--blue spacing-below-15">
+          <div class="spacing-below-5">
+            <%= ff.cfa_radio_set(
+                  :passed_away,
+                  label_text: t(".passed_away_label", name: ff&.object&.first_name&.capitalize),
+                  collection: [
+                    { value: "yes", label: t(".affirmative") },
+                    { value: "no", label: t(".negative") },
```
bengolder, line 35: we can use the general.yes or general.no strings for these

app/views/state_file/questions/az_senior_dependents/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                  ]
+                ) %>
+          </div>
+        </div>
+        <div class="highlight-box highlight-box--blue spacing-below-15">
+          <div class="spacing-below-5">
+            <%= ff.cfa_radio_set(
+                  :passed_away,
+                  label_text: t(".passed_away_label", name: ff&.object&.first_name&.capitalize),
```
bengolder, line 32: I assume we will never have an empty object, so I don't believe the & accessor is needed to protect against an missing record.

app/models/state_file_dependent.rb
```diff
@@ -32,4 +37,9 @@ def full_name
     parts << suffix if suffix.present?
     parts.compact.join(' ')
   end
+
+  def ask_senior_questions?
+    return false if dob.nil?
+    dob <= DateTime.new(2023, 12, 31).years_ago(65) && months_in_home == 12 && (relationship == 'PARENT' || relationship == 'GRANDPARENT')
```
bengolder, line 43: It might be helpful to pull out the end of tax year value into some global thing

spec/forms/state_file/az_senior_dependents_form_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+              needed_assistance: "unfilled",
+              passed_away: "unfilled"
+            }
+          }
+        }
+      end
+
+      it "returns false" do
+        form = described_class.new(intake, invalid_params)
+        expect(form).not_to be_valid
```
bengolder, line 24: add an expectation that the specific errors are set on the correct on object

app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -52,7 +52,7 @@ def documents_wrapper
                 xml.VisionExemp calculated_fields.fetch(:AZ140_LINE_9)
                 xml.DependentsUnder17 calculated_fields.fetch(:AZ140_LINE_10A)
                 xml.Dependents17AndOlder calculated_fields.fetch(:AZ140_LINE_10B)
-                xml.QualifyingParentsAncestors calculated_fields.fetch(:AZ140_LINE_11A)
+                # xml.QualifyingParentsAncestors calculated_fields.fetch(:AZ140_LINE_11A)
```
bengolder, line 55: can probably delete

PR #3744, opened 2023-11-02
Az last name 186317988

loumoore:
New AZ question page for collecting different last names the person has filed under in past years

### DIFF
--- a/app/assets/stylesheets/components/_form-elements.scss
+++ b/app/assets/stylesheets/components/_form-elements.scss
@@ -92,6 +92,10 @@
   &:before {
     display: none;
   }
+
+  &:after {
+    display: none;
+  }
 }
 
 .textarea {
--- a/app/forms/state_file/az_prior_last_names_form.rb
+++ b/app/forms/state_file/az_prior_last_names_form.rb
@@ -1,0 +1,16 @@
+module StateFile
+  class AzPriorLastNamesForm < QuestionsForm
+    attr_accessor :has_prior_last_names
+    set_attributes_for :intake, :prior_last_names
+
+    validates :prior_last_names, presence: true, allow_blank: false, if: -> { has_prior_last_names == "yes" }
+
+    def save
+      if has_prior_last_names == "no"
+        @intake.update(prior_last_names: nil)
+      else
+        @intake.update(attributes_for(:intake))
+      end
+    end
+  end
+end
--- a/app/forms/state_file/az_prior_last_names_form.rb
+++ b/app/forms/state_file/az_prior_last_names_form.rb
@@ -1,13 +1,12 @@
 module StateFile
   class AzPriorLastNamesForm < QuestionsForm
-    attr_accessor :has_prior_last_names
-    set_attributes_for :intake, :prior_last_names
+    set_attributes_for :intake, :prior_last_names, :has_prior_last_names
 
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/components/_form-elements.scss
```diff
@@ -92,6 +92,10 @@
   &:before {
     display: none;
   }
+
+  &:after {
+    display: none;
+  }
 }
```
loumoore, line 99: removes a little pointer UI doohickey. didn't seem to be needed for any other usages i could find but it is a global change.

app/forms/state_file/az_prior_last_names_form.rb
```diff
@@ -0,0 +1,15 @@
+module StateFile
+  class AzPriorLastNamesForm < QuestionsForm
+    set_attributes_for :intake, :prior_last_names, :has_prior_last_names
+
+    validates :prior_last_names, presence: true, allow_blank: false, if: -> { has_prior_last_names == "yes" }
+
+    def save
+      if has_prior_last_names == "no"
+        @intake.update(has_prior_last_names: "no", prior_last_names: nil)
+      else
```
loumoore, line 10: this works but it feels a little dirty?

db/migrate/20231102062958_add_has_prior_last_names_to_state_file_az_intake.rb
```diff
@@ -0,0 +1,5 @@
+class AddHasPriorLastNamesToStateFileAzIntake < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_az_intakes, :has_prior_last_names, :integer, default: 0, null: false
```
loumoore, line 3: could technically live without this being in the database since we have the prior_last_names field already but it makes it a lot easier to have the right state if someone goes back in the flow to edit this page

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -120,6 +120,11 @@
       choose "state_file_az_senior_dependents_form_dependents_attributes_0_passed_away_no"
       click_on "Continue"
 
+      expect(page).to have_text "Did you file with a different last name in the last four years?"
+      choose "state_file_az_prior_last_names_form_has_prior_last_names_yes"
+      fill_in "state_file_az_prior_last_names_form_prior_last_names", with: "Jordan, Pippen, Rodman"
```
loumoore, line 125: a list of comma-separated values is allowed per the mock, and i thought about trying to validate that the input is a comma-separated list, but it seems like it'd be easy to mess up and deny names that have different characters in them. if we want to make it harder to mess this up we should probably consider have an "add more" button or something so each prior last name has its own field instead of being comma-separated in a single field.

db/migrate/20231102062958_add_has_prior_last_names_to_state_file_az_intake.rb
tgrathwell: agreed, I tried to go without the extra enum column in another area and regretted it

spec/features/state_file/complete_intake_spec.rb
tgrathwell: I was gonna say we should validate based on whatever LNPriorYrs accepts in the XML, but it seems to have no restrictions on characters or length?? Shocking for a MeF schema item.

tgrathwell (approved):

PR #3747, opened 2023-11-02
Az last name 186317988

loumoore:


app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -87,7 +87,7 @@ def documents_wrapper
                     end
                     xml.RelationShip dependent.relationship
                     xml.NumMonthsLived dependent.months_in_home
-                    if dependent.dob <= 65.years.ago # TODO: needs to be based on a specific tax year date, also assumes we will have dob at all
+                    if dependent.dob <= MultiTenantService.statefile.end_of_current_tax_year.years_ago(65)
```
loumoore, line 90: unrelated to the prior last names change, but this was incorrectly using today's date instead of the end of the tax year

app/lib/pdf_filler/az140_pdf.rb
```diff
@@ -81,24 +82,24 @@ def hash_for_pdf
...
+        "56" => @xml_document.at('IncrExciseTaxCr')&.text,
+        "59" => @xml_document.at('TotalPayments')&.text,
         "60" => @xml_document.at('TaxDueOrOverpayment')&.text,
         "61" => @xml_document.at('OverPaymentOfTax')&.text,
-        "63"=> @xml_document.at('OverPaymentBalance')&.text,
+        "63" => @xml_document.at('OverPaymentBalance')&.text,
         "79" => @xml_document.at('RefundAmt')&.text,
         "80" => @xml_document.at('AmtOwed')&.text,
-                     })
+      })
```
loumoore, line 102: this is all just formatting

app/lib/efile/az/az140.rb
```diff
@@ -59,6 +59,7 @@ def calculate
           set_line(:AZ140_LINE_80, :calculate_line_80)
         end
         set_line(:AZ140_LINE_79, :calculate_line_79)
+        set_line(:AZ140_LINE_97, @intake, :prior_last_names)
```
tgrathwell, line 62: my conception of this layer (the so-called "calculator classes") is that we only need to put numbers and booleans in them, things like names and birthdates only need to be added to the xml/pdf layers as they're needed

loumoore: gotcha, removing

PR #3751, opened 2023-11-03
Az state credits 186326079

loumoore:


app/forms/state_file/az_state_credits_form.rb
```diff
@@ -0,0 +1,15 @@
+module StateFile
+  class AzStateCreditsForm < QuestionsForm
+    set_attributes_for :intake, :tribal_member, :tribal_wages, :armed_forces_member, :armed_forces_wages
+
+    validates :tribal_wages, presence: true, allow_blank: false, if: -> { tribal_member == "yes" }
```
tgrathwell, line 5: enums when configured with a `prefix` create methods like `tribal_member_yes?` which we generally prefer to `tribal_member == "yes"`

loumoore: i don't think it's an enum yet here though? it doesn't seem to recognize those methods on validation

app/lib/pdf_filler/az140_pdf.rb
```diff
@@ -83,6 +83,9 @@ def hash_for_pdf
         "12" => @xml_document.at('FedAdjGrossIncome')&.text,
         "14" => @xml_document.at('ModFedAdjGrossInc')&.text,
         "30" => @xml_document.at('USSSRailRoadBnft')&.text,
+        "31" => @xml_document.at('WageAmIndian')&.text,
+        "32" => @xml_document.at('CompNtnlGrdArmdFrcs')&.text,
+        "35" => @xml_document.at('TotalSubtractions')&.text,
```
loumoore, line 88: this was already being calculated but never added to xml or pdf. it's a calculation based on the new fields for this ticket so i went ahead and added it

loumoore: line 35 that is

PR #3758, opened 2023-11-07
Update name & date of birth form

bengolder:
- Renames DobController to NameDobController
- Adds primary, spouse, and dependents name inputs
- ask spouse name if MFS or MFJ in NY, but only if MFJ in AZ (per conversation with Courtney)
- removes names from the federal info controller
- fills out more of the XML for spouse info, but couldn't figure out where/how to add spouse dob
[[#186385135]](https://www.pivotaltracker.com/story/show/186385135)

**For the XML:**
SP_DOB_DT was already being added in the individual NY return, but it raised an error
TISpouseBO does not have SP_DOB_DT in the xsd

We added a test to more thoroughly fill TISpouseBO with SP_SSN_NMBR, but did not include SFX_NAME, DCSD_DT, or SP_EMP_DESC

We couldn't figure out where to put the spouse birthdate. 
We see SP_DOB_DT in RtnHeaderBO, but including it there causes an error


app/lib/submission_builder/formatting_methods.rb
```diff
@@ -18,6 +18,12 @@ def date_type(date)
       date.strftime("%F")
     end
 
+    def date_no_dash(date)
```
tgrathwell, line 21: unused?

bengolder: This was because I was trying to add SP_DOB_DT and the [[Working Doc] State File Data by State](https://docs.google.com/spreadsheets/d/1gTwfhe84ycTAXSuagk0UlT7sq6_9dmdrzE0j9EHiras/edit#gid=90796644) says the format should be "mmddyyyy". But I couldn't get SP_DOB_DT to work. Happy to remove this, not attached

PR #3759, opened 2023-11-08
AZ Review page and fixed some calculations for the AZ 140 form

loumoore:


app/forms/state_file/az_state_credits_form.rb
```diff
@@ -10,6 +10,14 @@ def save
...
       attributes_to_save[:armed_forces_wages] = nil if armed_forces_member == "no"
       @intake.update(attributes_to_save)
+
+      # Create submission now so PDF link button can be shown on the next page
+      efile_submission = EfileSubmission.create!(
+        data_source: @intake,
+        )
+      if Rails.env.development? || Rails.env.test?
+        efile_submission.transition_to(:preparing)
+      end
```
loumoore, line 20: Didn't seem possible to construct a link to the PDF on the next screen unless I create the submission first. So I took out the "Submit" page and moved the creation code in here. Looks like it's ok to create it early and not advance it in the state machine flow, but i could be missing something.

app/lib/efile/az/az140.rb
```diff
@@ -55,12 +55,8 @@ def calculate
           set_line(:AZ140_LINE_62, -> { 0 })
           set_line(:AZ140_LINE_63, :calculate_line_63)
         end
-        if (line_or_zero(:AZ140_LINE_63) - line_or_zero(:AZ140_LINE_78)) >= 0
-          set_line(:AZ140_LINE_79, :calculate_line_79)
-        else
-          set_line(:AZ140_LINE_80, :calculate_line_80)
-        end
         set_line(:AZ140_LINE_79, :calculate_line_79)
+        set_line(:AZ140_LINE_80, :calculate_line_80)
```
loumoore, line 59: line 79 was always being set and line 80 never was because stuff is always at least 0

i tracked this down because the review page needed to display the refund or owed amount and it was always 0

app/lib/efile/az/az140.rb
```diff
@@ -142,11 +138,9 @@ def calculate_line_44C
...
 
       def calculate_line_45
-        subtractions = 0
-        (33..44).each do |line_num|
-          subtractions += line_or_zero("AZ140_LINE_#{line_num}").to_i
-        end
-        [line_or_zero(:AZ140_LINE_42) - subtractions, 0].max
+        taxable_income = line_or_zero(:AZ140_LINE_42)
+        deductions_and_charity = line_or_zero(:AZ140_LINE_43) + line_or_zero(:AZ140_LINE_44)
+        [taxable_income - deductions_and_charity, 0].max
```
loumoore, line 143: this was the wrong math for this field according to the instructions in the pdf

app/lib/navigation/state_file_az_question_navigation.rb
```diff
@@ -16,7 +16,7 @@ class StateFileAzQuestionNavigation
       StateFile::Questions::AzPriorLastNamesController,
       StateFile::Questions::UnemploymentController,
       StateFile::Questions::AzStateCreditsController,
-      StateFile::Questions::SubmitReturnController,
+      StateFile::Questions::AzReviewController,
```
loumoore, line 19: removed the submit return controlled since i'm now creating the submission after AzStateCreditsController

app/lib/pdf_filler/az140_pdf.rb
```diff
@@ -86,11 +86,13 @@ def hash_for_pdf
...
         "35" => @xml_document.at('TotalSubtractions')&.text,
+        "37" => @xml_document.at('Form140/Subtotal')&.text,
+        "42" => @xml_document.at('AZAdjGrossIncome')&.text,
         "43" => @xml_document.at('AZDeductions')&.text,
-        "44" => @xml_document.at('ClaimCharitableDed')&.text,
+        "44" => @xml_document.at('TotalIncStdDeduction')&.text,
         "45" => @xml_document.at('AZTaxableInc')&.text,
         "46" => @xml_document.at('ComputedTax')&.text,
-        "48" => @xml_document.at('SubTotal')&.text,
+        "48" => @xml_document.at('Form140/DeductionAmt/SubTotal')&.text,
```
loumoore, line 95: Lots of SubTotal nodes!

app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -156,7 +156,7 @@ def documents_wrapper
                   end
                 end
               end
-              if calculated_fields[:AZ140_LINE_79]
+              if calculated_fields[:AZ140_LINE_79] > 0
```
loumoore, line 159: we were always writing RefundAmt and never AmtOwed before

app/forms/state_file/az_state_credits_form.rb
loumoore: maybe it would be better to at least move this into AzReviewContoller::edit in case the page order changes

tgrathwell: I'm a little wary of creating it here or on the review page's `#edit`

In CTC-land, we created it on the "confirm legal" page when people did their final sign-off https://github.com/codeforamerica/vita-min/blob/main/app/forms/ctc/confirm_legal_form.rb#L31

From the figma, they haven't actually filled out their Direct Deposit details or signed that PDF yet so it's technically incomplete. I'm wary of including a link to download a PDF at this point in the flow, but if it is meant to be somewhat incomplete, maybe we could find a way to generate it that doesn't require a persisted EfileSubmission?

Might require a lot of architectural change, maybe worth putting off till later.

app/controllers/state_file/questions/az_review_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  module Questions
+    class AzReviewController < QuestionsController
+      def edit
+        calculated_fields = current_intake.tax_calculator.calculate
+        if calculated_fields[:AZ140_LINE_79] > 0
+          @refund_or_owed_label = I18n.t("state_file.questions.az_review.edit.your_refund")
+          @refund_or_owed_amount = calculated_fields.fetch(:AZ140_LINE_79)
+        else
+          @refund_or_owed_label = I18n.t("state_file.questions.az_review.edit.your_tax_owed")
+          @refund_or_owed_amount = calculated_fields.fetch(:AZ140_LINE_80)
```
bengolder, line 11: it would be great to have a simple unit test for this conditional

app/forms/state_file/az_state_credits_form.rb
bengolder: I agree with Travis. I think we should remove the link for now, or just use a fake one, and perhaps have a follow up story for generating a PDF preview. We definitely don't want to send off the submission at this stage, which I assume `EfileSubmission.create!` would do.

bengolder (commented):
Blocking request for changes: I think we should remove the PDF link and do a follow up story about generating a PDF preview, because we want to separate that link from the e-file submission, which would happen later.

Non-blocking request for changes: I would also love to see a little bit of unit tests for the conditionals in the controller and the pdf/xml stuff. While we don't have many existing tests and I don't think we need tests for everything in there, the amount owed/due calculations and conditionals feel significant enough that I think we should document our expectations for that with some unit tests. It's hard to feel confident in these changes as a reviewer without tests.

app/forms/state_file/az_state_credits_form.rb
loumoore: i can look into what it would take to show the pdf without creating a submission first.

pretty sure the point of adding the pdf here was to make this simpler for engineering :) the previous design was displaying a bunch of federal return data instead, which wouldn't actually be that hard, so i can talk to design about that option too.

loumoore:
> Blocking request for changes: I think we should remove the PDF link and do a follow up story about generating a PDF preview, because we want to separate that link from the e-file submission, which would happen later.
> 
> Non-blocking request for changes: I would also love to see a little bit of unit tests for the conditionals in the controller and the pdf/xml stuff. While we don't have many existing tests and I don't think we need tests for everything in there, the amount owed/due calculations and conditionals feel significant enough that I think we should document our expectations for that with some unit tests. It's hard to feel confident in these changes as a reviewer without tests.

ok re: blocking change

The non-blocking thing was a conscious decision because there is so much untested in here. and lots that is probably still broken. so it felt like a massive rabbit hole to go down, but i guess i could have just covered the things i fixed. either way we'll need a story to get better test coverage on that thing.

app/forms/state_file/az_state_credits_form.rb
loumoore: lauren agreed to move the pdf viewing later in the flow, and we just replaced that button with a short message telling the user they'll be able to review the completed pdf later. easy peasy.

PR #3762, opened 2023-11-08
Add an acme challenge route so that letsencrypt will sign our AZ demo.fileyourstatetaxes.org cert

mrotondo:


mrotondo:
Not that it really matters in terms of how it gets validated by DF, but if they're going to be using staging to link back to in the end-to-end test should this be on the staging url?

config/routes.rb
```diff
@@ -372,6 +372,7 @@ def scoped_navigation_routes(context, navigation)
 
     mount ActionCable.server => '/cable'
     get '/.well-known/pki-validation/:id', to: 'public_pages#pki_validation'
+    get '/.well-known/acme-challenge/:id', to: 'public_pages#acme_challenge'
```
tgrathwell, line 375: i think if you want this to be served on the FYST domain you would need it to be in the third and final constraints block in the routes file instead

mrotondo: Good catch! Done

mrotondo:
results verified locally:
`➜  vita-min git:(acme_challenge) curl http://statefile.localhost:3000/.well-known/acme-challenge/nAiwr5D4js5JKPcdlY7oN2Dxim4UUq-N6kNOvHew18o`
`nAiwr5D4js5JKPcdlY7oN2Dxim4UUq-N6kNOvHew18o.cwZpB_0Yw5Z-XZHRGYZPtCj4fUivbvN1yK4zBuNBVBk%`

PR #3763, opened 2023-11-08
Add Data Review Screen

rickreyhsig:
Adding the Data review screen according to https://www.pivotaltracker.com/n/projects/2650644/stories/186317906


app/controllers/state_file/questions/data_review_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      def form_class
+        NullForm
+      end
+
+      def prev_path
+        nil
+      end
+    end
+  end
+end
```
rickreyhsig, line 18:
```suggestion
end

```

PR #3765, opened 2023-11-09
State File: add first eligibility screener page for AZ and NY

jenny-heath:
image

image


bengolder (commented):
This looks good! There a few possible non-blocking refactors that I thought might be helpful, but aren't essential.

For the "ineligible reason" logic, it seems like it might be nice to have a unit test for it. I think putting the logic into the intake model and having a method that returns a key could be a nice way to reduce the conditional branching and eliminate the need for a helper method. It would also allow us to include the ineligibility reason in analytics and might make it easier to add more reasons for ineligibility in one spot if we expand the number of screening pages. For example, we could use it as a key to lookup translations for each reason.

```
<p><%= t(".ineligible_reason.#{current_intake.ineligible_reason}", prev_link: prev_path) %></p>
```

For the feature specs, I think it would be nice to reduce our usage of them. For feature specs that test a certain pair of pages, perhaps we can start a feature spec at a particular page before a specific thing we want to test rather than testing every page from the beginning. I'm also wondering if we can trust the ".show?" unit tests sufficiently so maybe we don't need a feature spec for testing those flows? In general I'm trying to look for ways to reduce reliance & size of feature specs so our test suite doesn't get too slow.

rickreyhsig (approved):

bengolder (approved):

PR #3768, opened 2023-11-09
Add esign declaration page

embarnard:


embarnard:
Screenshot 2023-11-09 at 5 23 20 PM
Screenshot 2023-11-09 at 5 23 35 PM

embarnard:
Screenshot 2023-11-13 at 4 46 17 PM

app/models/state_file_az_intake.rb
```diff
@@ -49,6 +53,8 @@ class StateFileAzIntake < StateFileBaseIntake
   enum charitable_contributions: { unfilled: 0, yes: 1, no: 2 }, _prefix: :charitable_contributions
   enum eligibility_married_filing_separately: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_married_filing_separately
   enum eligibility_529_for_non_qual_expense: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_529_for_non_qual_expense
+  enum primary_esigned: { unfilled: 0, yes: 1, no: 2 }, _prefix: :primary_esigned
+  enum spouse_esigned: { unfilled: 0, yes: 1, no: 2 }, _prefix: :spouse_esigned
```
bengolder, line 57: Even though this is stored as an enum, it doesn't look like there is any way to input a "no" option. How have we done other signatures? Are they booleans? Just the date? I think if we were trying to store only the necessary data, a primary/spouse_esigned_at field would be sufficient, though we might have to adjust the form & template to set the timestamp based on the checkbox value.

db/migrate/20231113211138_add_esign_to_state_intakes.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    add_column :state_file_ny_intakes, :primary_esigned, :integer, default: 0, null: false
+    add_column :state_file_ny_intakes, :primary_esigned_at, :datetime
+    add_column :state_file_az_intakes, :primary_esigned, :integer, default: 0, null: false
+    add_column :state_file_az_intakes, :primary_esigned_at, :datetime
+
+    add_column :state_file_ny_intakes, :spouse_esigned, :integer, default: 0, null: false
+    add_column :state_file_ny_intakes, :spouse_esigned_at, :datetime
+    add_column :state_file_az_intakes, :spouse_esigned, :integer, default: 0, null: false
+    add_column :state_file_az_intakes, :spouse_esigned_at, :datetime
```
bengolder, line 11: I think we can do this with datetimes only, but I'd want to double check against prior signature implementations.

app/forms/state_file/esign_declaration_form.rb
```diff
@@ -0,0 +1,26 @@
...
+      @intake.touch(:primary_esigned_at) if @intake.primary_esigned_yes?
+      @intake.touch(:spouse_esigned_at) if @intake.spouse_esigned_yes? && @intake.filing_status_mfj?
+
+      # Submits return
+      efile_submission = EfileSubmission.create!(
+        data_source: @intake,
+      )
+      if Rails.env.development? || Rails.env.test?
+        efile_submission.transition_to(:preparing)
+      end
```
bengolder, line 22: Why is this scoped to the test & dev environments? Is that to fake the behavior of getting a response back from the IRS?

app/views/state_file/questions/esign_declaration/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<% state_name = States.name_for_key(params[:us_state].upcase) %>
+<% title = t(".title", state_name: state_name) %>
+<% content_for :page_title, title %>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= title %></h1>
+  <%= params[:us_state] == "ny" ? t(".ny.body_html") : t(".az.body_html") %>
```
bengolder, line 7: Just an idea/observation: I feel like we could use a helper method for checking the state param.

app/views/state_file/questions/ny201/edit.html.erb
bengolder, line 1: I don't have much context on this page or why the sales use tax field was deleted (I can guess why the signature fields were deleted). Would you be willing to explain a little about what this page is for and why you deleted the use tax field, just for my own learning?

app/lib/pdf_filler/ny214_pdf.rb
```diff
@@ -67,7 +67,6 @@ def hash_for_pdf
         '31 dollars14' => claimed_attr_value('R_TX_AVL_CR_AMT'),
         '32 dollars14' => claimed_attr_value('R_CR_LMT_AMT'),
         '33 dollars14' => claimed_attr_value('R_RL_PROP_CR_AMT'),
-        # TODO signature
```
bengolder, line 70: Is this being removed because we won't need to add signatures to this pdf?

config/locales/es.yml
```diff
@@ -1802,6 +1802,23 @@ es:
...
+          ny:
+            body_html: |
+              <p>By checking the box shown below, I declare, under penalty of perjury, that I have examined the information on my 2023 New York State electronic personal income tax return, including any accompanying schedules, attachments, and statements, and certify that my electronic return is true, correct, and complete.</p>
+              <p>If I am paying my New York State personal income taxes owed by electronic funds withdrawal, I authorize the New York State Tax Department and its designated financial agents to initiate an electronic funds withdrawal from the financial institution account indicated on my 2023 electronic return, and I authorize my financial institution to debit the entry to my account. I understand and agree that I may revoke this authorization for payment only by contacting the Tax Department no later than five (5) business days prior to the payment date.</p>
```
bengolder, line 1821: You added values to es.yml. I've been leaving the es.yml keys out for translations that we don't have yet. should I be adding placeholder values like this?

bengolder (approved):

bengolder (approved):
Overall looks good, I had some questions and I'd prefer that we use something other than a yes/no enum for the signature checkbox, but no blocking concerns and it looks like it will get the job done.

bengolder (approved):
Overall looks good, I had some questions and I'd prefer that we use something other than a yes/no enum for the signature checkbox, but no blocking concerns and it looks like it will get the job done.

app/models/state_file_az_intake.rb
embarnard: @bengolder I based this off the `consented_to_legal` field on the CTC intake which also did a unfilled/yes/no pattern and a matching consented_to_legal_at date field. I agree that I think a boolean would also make sense or just use the date field but I like keeping in line with what we've done in the past more for consistency.

app/forms/state_file/esign_declaration_form.rb
embarnard: I just moved this from the previous controller. It just looks like new code because the controller name has changed but it has already been code reviewed and I don't have context on it @bengolder

config/locales/es.yml
embarnard: @bengolder I only added it because these are _html fields and they wouldn't pass the missing tags test, I don't think you have to worry about adding es.yml if the tests are passing

PR #3771, opened 2023-11-10
Statefile: Add second eligibility page

bengolder:
Screenshot 2023-11-10 at 1 14 23 PM
Screenshot 2023-11-10 at 1 14 10 PM

### DIFF
--- a/app/controllers/concerns/state_file/eligibility_offboarding_concern.rb
+++ b/app/controllers/concerns/state_file/eligibility_offboarding_concern.rb
@@ -1,0 +1,19 @@
+module StateFile
+  module EligibilityOffboardingConcern
+    # This concern can be used by any controller that needs to redirect
+    # to the eligibility offboarding page if given a disqualifying answer
+    extend ActiveSupport::Concern
+
+    private
+
+    def offboarding_path
+      StateFile::Questions::EligibilityOffboardingController.to_path_helper(action: :edit, us_state: params[:us_state])
+    end
+
+    def next_path
+      return offboarding_path if current_intake.has_disqualifying_eligibility_answer?
+
+      super
+    end
+  end
+end
--- a/app/controllers/state_file/questions/eligibility_offboarding_controller.rb
+++ b/app/controllers/state_file/questions/eligibility_offboarding_controller.rb
@@ -4,21 +4,12 @@ module StateFile
       helper_method :ineligible_reason
 
       def ineligible_reason
-        self.class.ineligible_reason(current_intake)
+        key = current_intake.disqualifying_eligibility_answer
+        I18n.t("state_file.questions.eligibility_offboarding.edit.ineligible_reason.#{key}")
       end
 
       def self.show?(intake)
-        ineligible_reason(intake).present?
-      end
-
-      def self.ineligible_reason(intake)
-        if intake.eligibility_lived_in_state_no?
-          I18n.t("state_file.questions.eligibility_offboarding.edit.not_full_year_resident")
-        elsif intake.class == StateFileNyIntake && intake.eligibility_yonkers_yes?
-          I18n.t("state_file.questions.eligibility_offboarding.edit.yonkers")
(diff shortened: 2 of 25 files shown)


app/controllers/concerns/state_file/eligibility_offboarding_concern.rb
bengolder, line 1: On three different controllers, we will need to redirect to the offboarding page if they submit a disqualifying answer to an eligibility question.
This allows us to use the same code for all three controllers.

app/controllers/state_file/questions/eligibility_offboarding_controller.rb
bengolder, line 1: I refactored most of this into the two state file intake models and provided unit tests at the model layer to check for disqualifying eligibility answers and provide the specific reason.

app/models/state_file_az_intake.rb
```diff
@@ -45,8 +47,8 @@ class StateFileAzIntake < StateFileBaseIntake
   enum tribal_member: { unfilled: 0, yes: 1, no: 2 }, _prefix: :tribal_member
   enum armed_forces_member: { unfilled: 0, yes: 1, no: 2 }, _prefix: :armed_forces_member
   enum charitable_contributions: { unfilled: 0, yes: 1, no: 2 }, _prefix: :charitable_contributions
-  enum eligibility_lived_in_state: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_lived_in_state
```
bengolder, line 48: I moved this into the base intake because it exists in both states.

app/models/state_file_az_intake.rb
```diff
@@ -97,4 +99,13 @@ def ask_spouse_name?
...
   end
+  
+  def disqualifying_eligibility_answers
+    {
+      eligibility_lived_in_state: "no",
+      eligibility_married_filing_separately: "yes",
+      eligibility_out_of_state_income: "yes",
+      eligibility_529_for_non_qual_expense: "yes",
+    }
+  end
```
bengolder, line 110: These will be used by the base intake to check for any disqualifying answers and to provide the reason.

app/models/state_file_base_intake.rb
```diff
@@ -72,4 +75,16 @@ def full_name
...
     end
   end
+
+  def disqualifying_eligibility_answer
+    disqualifying_eligibility_answers.each do |col, value|
+      return col if self.public_send(col) == value
+    end
+
+    nil
+  end
```
bengolder, line 85: This method will return the symbol for the column that had a disqualifying answer (for example `:eligibility_out_of_state_income`). We can use the symbol to look up a relevant message and translation.

db/migrate/20231110205748_add_more_e11y_fields_to_state_intakes.rb
```diff
@@ -0,0 +1,10 @@
+class AddMoreE11yFieldsToStateIntakes < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_ny_intakes, :eligibility_out_of_state_income, :integer, default: 0, null: false
+    add_column :state_file_ny_intakes, :eligibility_part_year_nyc_resident, :integer, default: 0, null: false
+    add_column :state_file_ny_intakes, :eligibility_withdrew_529, :integer, default: 0, null: false
```
bengolder, line 5: I added the `:eligibility_withdrew_529` column which I don't use in this PR, but will be useful for [the next eligibility story](https://www.pivotaltracker.com/story/show/186439181)

rickreyhsig: It seems as the specs have some kind of dependency?
https://github.com/codeforamerica/vita-min/blob/ee8b6f1f5fd9ea53d06e486702dbfa9e69063205/spec/support/helpers/state_file_intake_helper.rb#L18

rickreyhsig (approved):
Looks great! There's room to refactor the form class and name into a separate concern so we don't repeat ourselves in the future

PR #3776, opened 2023-11-14
Add AZ and NY demo keys to the development and demo credentials, and …

mrotondo:
…read them and use them in the IRS API Service


mrotondo:
I tested this by executing:
`IRS_STATE_ACCOUNT_ID=12345 rails runner fake_api_server/test_api_request.rb 012345` and 
`IRS_STATE_ACCOUNT_ID=012345 rails runner fake_api_server/test_api_request.rb 012345`, 
and verified that the right keys and certs got decoded and written from the credentials into the already-used paths, and getting an expected API error from the IRS

PR #3777, opened 2023-11-14
add a review flow for AZ and NY 

loumoore:
… where the user can go back to certain pages and then return directly to the review page at the end of the flow

I thought maybe there'd be an existing preferred pattern for doing something like this but I couldn't find one. Happy to redo it with a different approach if there's a better way. I'm just passing a parameter around.


spec/support/helpers/state_file_intake_helper.rb
```diff
@@ -84,4 +84,64 @@ def step_through_df_data_transfer
...
+    expect(page).to have_text I18n.t("state_file.questions.ny_review.edit.title1")
+
+    # Out of state purchases/use tax
+    within "#use-tax" do
+      click_on I18n.t("general.edit")
+    end
+    expect(page).to have_text I18n.t('state_file.questions.ny_sales_use_tax.edit.title', year: MultiTenantService.statefile.current_tax_year)
+    click_on I18n.t("general.continue")
+    expect(page).to have_text I18n.t("state_file.questions.ny_review.edit.title1")
+  end
```
bengolder, line 146: Non-important preference: This will get the job done and is quite thorough! I've been trying to use unit tests over feature specs wherever possible because feature specs are so much slower to run. I think this could probably be well covered by unit tests and maybe one shorter feature spec that goes back, does an edit, and returns. This is definitely not critical, but would help prevent the test suite from getting impractically slow.

bengolder (approved):
Approved! I think the overall approach of using a query param and a hidden field is a great solution. I was confused about why the hidden input field would go inside of the submit button and I would prefer to use a mixin and generally avoid overloading shared functions with additional options if we could keep that new option separate. Nothing critical or blocking though! Thank you!

app/views/state_file/questions/az_state_credits/edit.html.erb
```diff
@@ -60,8 +60,6 @@
       </div>
     </div>
 
-    <button class="button button--primary" type="submit">
-      <%= t("general.continue") %>
-    </button>
+    <%= f.submit(t("general.continue"), review: params[:review], class: "button button--primary button--wide") %>
```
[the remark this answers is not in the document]
loumoore: yeah, i was just looking for ways to add less repeated code to each erb, but couldn't figure out a way not to. what i ended up with isn't really any easier than your suggestion, and it's less explicit. so i agree, i'll make this change. thanks!

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -26,6 +26,14 @@ def question_navigator
...
       def next_path
         next_step = form_navigation.next
+        unless params[:review].nil?
+          case params[:us_state]
+          when 'az'
+            next_step = StateFile::Questions::AzReviewController
+          when 'ny'
+            next_step = StateFile::Questions::NyReviewController
+          end
+        end
```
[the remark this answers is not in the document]
loumoore: ok i will learn what a concern mixin is and see about doing this :) i did think it was kinda nice not to have to edit every controller that might be used in the review flow though.

app/helpers/vita_min_form_builder.rb
```diff
@@ -330,7 +330,7 @@ def vita_min_date_text_fields(method, label_text, help_text: nil, options: {}, o
   def submit(value, options = {})
     options[:data] ||= {}
     options[:data][:disable_with] = value
-    super(value, **options)
+    options.delete(:review) ? @template.hidden_field_tag(:review, 'y') + super(value, **options) : super(value, **options)
```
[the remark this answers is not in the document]
loumoore: the generated html isn't inside a submit button, it's prepended, sorta like how the form tag generates the csrf token hidden fields. when i was doing it, i thought it made sense to say "i want to submit and go back to review if review mode is on" all together. 

but i agree, there's no big benefit of extra magic here and we can just use a hidden input field explicitly. will change.

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -24,10 +24,14 @@ def question_navigator
...
+        form_navigation.next
+      end
+
       def next_path
-        next_step = form_navigation.next
-        options = { us_state: params[:us_state], action: next_step.navigation_actions.first }
-        if next_step.resource_name.present? && next_step.resource_name == self.class.resource_name
+        step_for_next_path = next_step
+        options = { us_state: params[:us_state], action: step_for_next_path.navigation_actions.first }
+        if step_for_next_path.resource_name.present? && step_for_next_path.resource_name == self.class.resource_name
```
loumoore, line 34: i refactored this because it seems like we might want the resource_name logic here to run. all i really want to override is the next step, not necessarily the rest of the logic to construct the path. i think.

app/views/state_file/questions/az_charitable_contributions/edit.html.erb
```diff
@@ -26,8 +26,9 @@
       </div>
     </div>
 
-    <button class="button button--primary" type="submit">
-      <%= t("general.continue") %>
-    </button>
+    <% unless params[:return_to_review].nil? %>
+      <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
+    <% end %>
+    <%= f.continue %>
```
loumoore, line 32: these are a little more verbose now because we don't want to include the tag at all if the return_to_review param isn't set

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -90,38 +90,48 @@
...
+
+      # From the review page, the user can go back to certain screens to edit and then should return directly to the
+      # review page. This is well-covered by unit tests, but let's test just one of those screens here
+      expect(page).to have_text I18n.t("state_file.questions.ny_review.edit.title1")
+      within "#county" do
+        click_on I18n.t("general.edit")
+      end
+      expect(page).to have_text I18n.t("state_file.questions.ny_county.edit.title", filing_year: MultiTenantService.statefile.current_tax_year)
+      click_on I18n.t("general.continue")
+      expect(page).to have_text I18n.t("state_file.questions.ny_review.edit.title1")
```
loumoore, line 122: i just kept one screen in each state flow, now that everything is covered in controller unit tests too

spec/controllers/state_file/questions/ny_sales_use_tax_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  let(:intake) { create :state_file_ny_intake }
+  before do
+    session[:state_file_intake] = intake.to_global_id
+  end
+
+  describe "#update" do
+    # use the return_to_review_concern shared example if the page
+    # should skip to the review page when the return_to_review param is present
+    # requires form_params to be set with any other required params
+    it_behaves_like :return_to_review_concern do
```
loumoore, line 13: in the other similar example we have, an intake factory is passed in. but at least some of these require additional fields to be set on the intake, so i'm creating it and setting it in the session in here instead

spec/support/shared_examples/return_to_review_concern.rb
```diff
@@ -0,0 +1,34 @@
...
+      it "navigates to the next page in the flow" do
+        post :update, params: form_params
+        controllers = []
+        case form_params[:us_state]
+        when "az"
+          controllers = Navigation::StateFileAzQuestionNavigation::FLOW.to_a
+        when "ny"
+          controllers = Navigation::StateFileNyQuestionNavigation::FLOW.to_a
+        end
+        next_controller = controllers[controllers.index(described_class) + 1]
```
loumoore, line 29: this obvs has some edge cases where it could break, but i think it's safe to assume that we should be testing controllers that have a valid next controller in this context.

app/controllers/concerns/state_file/return_to_review_concern.rb
```diff
@@ -0,0 +1,22 @@
...
+      when 'az'
+        StateFile::Questions::AzReviewController
+      when 'ny'
+        StateFile::Questions::NyReviewController
+      end
+    end
+
+    def next_step
+      params[:return_to_review].nil? ? super : review_step
+    end
```
bengolder, line 20: You could alternatively overwrite `next_path` and that might make for an simpler integration with questions controller. I'm not saying this as a suggestion so much as pointing out another option in case you'd like it better.

app/views/state_file/questions/ny201/edit.html.erb
```diff
@@ -37,6 +37,9 @@
     <%= f.cfa_date_select :date_electronic_withdrawal, "Electronic funds withdrawal Date" %>
     <%= f.cfa_input_field :amount_electronic_withdrawal, "Electronic funds amount" %>
 
+    <% unless params[:return_to_review].nil? %>
+      <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
+    <% end %>
```
bengolder, line 42: This is totally good, but if you'd prefer avoiding a double negative you could do:
```erb
<% if params[:return_to_review].present? %>
      <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
<% end %>
```

spec/support/shared_examples/return_to_review_concern.rb
bengolder, line 1: This is excellent!

spec/controllers/state_file/questions/ny_201_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+          state_file_ny201_form: {
+            account_type: "personal_checking",
+            nyc_full_year_resident: "yes",
+            refund_choice: "direct_deposit"
+          }
+        }
+      end
+    end
+  end
+end
```
bengolder, line 26: I really appreciate the unit tests and shared concern, thank you!

bengolder (approved):
I really appreciate these changes, thank you!
I noted a few things that aren't even suggestions but just additional options I can see in case you would have a preference.

I do want to acknowledge that, as you noted, adding the concern and unit tests to each controller can be a bit repetitive.

app/controllers/concerns/state_file/return_to_review_concern.rb
loumoore: i think i'm going to just push this as is, but it's not because i have a strong stylistic preference -- i was concerned that overriding all of next_path would always work properly because there's logic in there beyond selecting the next controller that wouldn't run. specifically this bit:

`        
if step_for_next_path.resource_name.present? && step_for_next_path.resource_name == self.class.resource_name
    options[:id] = current_resource.id
end
`

i'm not sure what that's for, but whatever it is, it seemed to make sense to be able to override just the next step, as opposed to all the logic around determining the next path.

app/views/state_file/questions/ny201/edit.html.erb
loumoore: haha duh, yeah that's better will change

PR #3779, opened 2023-11-14
Collect and transmit IP address and device id

embarnard:


app/controllers/flows_controller.rb
```diff
@@ -571,16 +571,19 @@ def self.common_attributes
         primary_esigned_at: 1.minute.ago,
         spouse_esigned: "yes",
         spouse_esigned_at: 1.minute.ago,
+        raw_direct_file_data: File.read(Rails.root.join('app', 'controllers', 'state_file', 'questions', 'df_return_sample.xml')),
```
embarnard, line 574: just moved some duplicate attributes between NY and AZ into common atrributes

app/controllers/flows_controller.rb
```diff
@@ -622,22 +622,31 @@ def self.az_attributes(first_name: 'Testuser', last_name: 'Testuser')
...
         prior_last_names: "Jordan, Pippen, Rodman",
-        raw_direct_file_data: File.read(Rails.root.join('app', 'controllers', 'state_file', 'questions', 'df_return_sample.xml')),
         tribal_member: "yes",
         tribal_wages: 100,
-        payment_or_deposit_type: "mail",
-        account_type: 'unfilled'
+      )
+    end
+
+    def generate_efile_device_info(intake)
```
embarnard, line 638: this makes sure that we create the efile device infos when we generate an intake for the flow. Otherwise someone would have to go to the data review page and esign page in order to get the XML to work

app/controllers/state_file/questions/data_review_controller.rb
```diff
@@ -1,12 +1,27 @@
...
+          ip_address: ip_for_irs,
+          intake: current_intake,
+        )
+      end
 
-      def form_class
-        NullForm
+      def update
+        if form_params["device_id"].blank?
+          flash[:alert] = I18n.t("general.enable_javascript")
```
embarnard, line 15: if the client does not have javascript enabled we won't be able to grab the device id; in CTC we also displayed this "please enable javascript message" if there was no javascript. Although might be worth checking if we want to be so harsh as to not let them continue without a device id. We do need it inorder for the XML to pass but to what end will IRS need one to accept the return.

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -7,6 +7,14 @@ class QuestionsController < ::Questions::QuestionsController
       # default layout for all state file questions
       layout "state_file/question"
 
+      def ip_for_irs
+        if Rails.env.test?
+          "72.34.67.178"
```
embarnard, line 12: for passing the feature spec

app/controllers/state_file/questions/submission_confirmation_controller.rb
```diff
@@ -1,7 +1,9 @@
 module StateFile
   module Questions
     class SubmissionConfirmationController < QuestionsController
-      def edit; end
+      def edit
+        raise ActiveRecord::RecordNotFound unless EfileSubmission.where(data_source: current_intake).present?
```
embarnard, line 5: noticed we would receive a weird error on this page if we generated a intake from the flow service and then immediately went to the submission page. This is because we need a submission on this page and we can only get one from submitting on the esign_declaration form. Hopefully this error message makes it more obvious in those cases.

app/lib/submission_builder/ty2022/states/authentication_header.rb
```diff
@@ -53,25 +53,28 @@ def document
...
-                  xml.IPv4AddressTxt '1.2.3.4'
+                  xml.IPv4AddressTxt device_info&.ip_address if device_info.ip_address.ipv4?
+                  xml.IPv6AddressTxt device_info&.ip_address if device_info.ip_address.ipv6?
                 end
-                xml.IPTs '2023-10-10T12:00:00-05:00'
-                xml.DeviceId 'AB' * 20
-                xml.DeviceTypeCd 'Desktop'
-                xml.IPPortNum '1234'
+                xml.IPTs datetime_type(device_info&.created_at)
+                xml.DeviceId device_info&.device_id || 'AB' * 20
```
embarnard, line 63: noticed we did this `|| 'AB' * 20` in CTC land presumably to allow clients to submit a return even without a device id

app/lib/submission_builder/ty2022/states/authentication_header.rb
```diff
@@ -53,25 +53,28 @@ def document
...
-                xml.DeviceId 'AB' * 20
-                xml.DeviceTypeCd 'Desktop'
-                xml.FinalIPPortNumberSubmit '1234'
+                xml.IPTs datetime_type(device_info.created_at)
+                xml.DeviceId device_info&.device_id || 'AB' * 20
+                xml.DeviceTypeCd 'Browser-based'
               end
-              xml.TotActiveTimePrepSubmissionTs '30'
-              xml.TotalPreparationSubmissionTs '5'
+              xml.TotActiveTimePrepSubmissionTs '30' # total_active_preparation_minutes -- Total Active Time Preparation Submission Time Span
```
embarnard, line 76: this is not part of this story technically but I should make sure Carmen is aware of these fields

app/controllers/state_file/questions/data_review_controller.rb
embarnard: Just to note, asked Alisa on the DF team and they said they would ask their team what they are doing

app/controllers/flows_controller.rb
```diff
@@ -622,22 +622,31 @@ def self.az_attributes(first_name: 'Testuser', last_name: 'Testuser')
...
+      StateFileEfileDeviceInfo.find_or_create_by!(
+        event_type: "initial_creation",
+        ip_address: "72.34.67.178",
```
rickreyhsig, line 641: I may have a better understanding later in this PR review, but why are we hardcoding the device ID and IPs here?

spec/lib/submission_builder/ty2022/states/az/individual_return_spec.rb
```diff
@@ -4,6 +4,8 @@
   describe '.build' do
     let(:intake) { create(:state_file_az_intake) }
     let(:submission) { create(:efile_submission, data_source: intake) }
+    let!(:initial_efile_device_info) { create :state_file_efile_device_info, :initial_creation, :filled, intake: intake }
```
rickreyhsig, line 7: Are these objects used somewhere outside of this spec?

app/controllers/flows_controller.rb
rickreyhsig: Initial - DataReview step
Submission - Esign step

spec/controllers/state_file/questions/data_review_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        post :update, params: { us_state: "az", device_id: "" }
+        expect(flash[:alert]).to eq(I18n.t("general.enable_javascript"))
+      end
+    end
+
+    context "with device id" do
+      it "updates device id" do
+        post :update, params: { us_state: "az", state_file_data_review_form: { device_id: "ABC123" } }
+        expect(efile_device_info.reload.device_id).to eq "ABC123"
+      end
```
rickreyhsig, line 38: May be a good idea to assert the efile_device_infoevent_type == 'initial_creation'

spec/controllers/state_file/questions/esign_declaration_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      it "flashes an alert and does re-renders edit" do
+        post :update, params: params
+        expect(flash[:alert]).to eq(I18n.t("general.enable_javascript"))
+      end
+    end
+
+    context "with device id and signatures" do
+      it "updates device id and signatures" do
+        post :update, params: params
+        expect(efile_device_info.reload.device_id).to eq "ABC123"
```
rickreyhsig, line 49: Ditto. May be a good idea to assert the efile_device_infoevent_type == 'submission'

db/schema.rb
```diff
@@ -1637,6 +1637,17 @@
...
 
+  create_table "state_file_efile_device_infos", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.string "device_id"
+    t.string "event_type"
+    t.bigint "intake_id", null: false
+    t.string "intake_type", null: false
+    t.inet "ip_address"
+    t.datetime "updated_at", null: false
+    t.index ["intake_type", "intake_id"], name: "index_state_file_efile_device_infos_on_intake"
```
rickreyhsig, line 1648: Do you foresee use querying these fields a bunch, hence the indexes?

app/models/state_file_base_intake.rb
```diff
@@ -98,6 +99,16 @@ def has_disqualifying_eligibility_answer?
     disqualifying_eligibility_answer.present?
   end
 
+  def initial_efile_device_info
+    with_device_id = efile_device_infos.where(event_type: "initial_creation").where.not(device_id: nil).first
```
rickreyhsig, line 103: I assume this is for the case when the user had JS disabled?

app/controllers/state_file/questions/questions_controller.rb
rickreyhsig: It seems questions controller is a child class of Application controller which already contains this method.  Can we add a clause for `elsif Rails.env.test?` there?

app/controllers/flows_controller.rb
rickreyhsig: So it seems we are using `rails request.remote_ip` to get the IP, from the request and this is the thing we use for setting up https://demo.fileyourstatetaxes.org/en/flows, so you can disregard my initial concern.

app/lib/submission_builder/ty2022/states/authentication_header.rb
```diff
@@ -53,25 +53,28 @@ def document
...
                 xml.IPAddress do
-                  xml.IPv4AddressTxt '1.2.3.4'
+                  xml.IPv4AddressTxt device_info&.ip_address if device_info&.ip_address&.ipv4?
+                  xml.IPv6AddressTxt device_info&.ip_address if device_info&.ip_address&.ipv6?
                 end
-                xml.IPTs '2023-10-10T12:00:00-05:00'
-                xml.DeviceId 'AB' * 20
-                xml.DeviceTypeCd 'Desktop'
-                xml.IPPortNum '1234'
+                xml.IPTs datetime_type(device_info&.created_at)
```
rickreyhsig, line 62: I assume this is the timestamp we are passing to the final XML? If so, I think updated_at would make more sense in case it is ever updated.

PR #3790, opened 2023-11-15
State file: move intake creation to landing page

jenny-heath:


app/controllers/state_file/questions/landing_page_controller.rb
```diff
@@ -1,6 +1,12 @@
 module StateFile
   module Questions
     class LandingPageController < QuestionsController
+      include StartIntakeConcern
+
+      def update
+        current_intake.save
+        super
```
jenny-heath, line 8: i can't say i feel particularly confident about this approach? the issue is that when `StartIntakeConcern` was on a page that actually saved data to intake, the intake would be persisted by the time it got to `after_update_success`, where it tries to call `current_intake.to_global_id` (which doesn't work if the intake isn't persisted). open to doing this another way, i can think of several

bengolder (approved):
Looks good!

PR #3795, opened 2023-11-16
Claimed as dependent fix

embarnard:
Previously we had the `claimed_as_dep` field as a placeholder but now we are replacing it with the `TotalExemptPrimaryAndSpouseCnt` value that we are grabbing from the DF XML. Where if they are claimed as a dependent this value should be `0`

### DIFF
--- a/app/lib/efile/ny/it201.rb
+++ b/app/lib/efile/ny/it201.rb
@@ -444,7 +444,7 @@ module Efile
       end
 
       def calculate_line_69a
-        return 0 unless @nyc_full_year_resident && @direct_file_data.claimed_as_dependent? == false
+        return 0 unless @nyc_full_year_resident && !@direct_file_data.claimed_as_dependent?
 
 
         nyc_taxable_income = line_or_zero(:IT201_LINE_47)
--- a/app/lib/efile/ny/it201.rb
+++ b/app/lib/efile/ny/it201.rb
@@ -3,11 +3,10 @@ module Efile
     class It201 < ::Efile::TaxCalculator
       attr_reader :lines
 
-      def initialize(year:, filing_status:, claimed_as_dependent:, intake:, direct_file_data:, nyc_full_year_resident:, dependent_count:, include_source: false)
+      def initialize(year:, filing_status:, intake:, direct_file_data:, nyc_full_year_resident:, dependent_count:, include_source: false)
         @year = year
 
         @filing_status = filing_status # single, married_filing_jointly, that's all we support for now
-        @claimed_as_dependent = claimed_as_dependent # true/false
         @intake = intake
         @direct_file_data = direct_file_data
         @nyc_full_year_resident = nyc_full_year_resident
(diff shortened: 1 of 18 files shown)


app/models/state_file_base_intake.rb
```diff
@@ -95,4 +93,13 @@ def disqualifying_eligibility_answer
   def has_disqualifying_eligibility_answer?
     disqualifying_eligibility_answer.present?
   end
+
+  def save_nil_enums_with_unfilled
```
embarnard, line 97: Added this so that any enum that has a `unfilled` type in the base intakes will replace a nil save with unfilled. I noticed that this was causing problems with the flow explorer because we would set initial value account type value to mail and then the value_changed?(to: "mail") statement in the before_save would not catch it in-order change the nil to unfilled. Seen this issue pop up for other stories as well, so hopefully this will prevent database issues

app/lib/efile/ny/it214.rb
```diff
@@ -24,8 +23,9 @@ def calculate
           offboard
           return
         end
-        set_line(:IT214_LINE_4, -> { @claimed_as_dependent })
-        if @claimed_as_dependent
+        # Dependent on federal return?
```
embarnard, line 26: this field actually doesn't have to do with claimed_as_dependent and is actually an indicator for if a dependent is on the federal return, the 214 XML has a lot of errors which you can see when you force an inclusion in the supported docs. We probably want to make a ticket to go through and debug these

app/lib/efile/ny/it201.rb
```diff
@@ -446,7 +444,7 @@ def calculate_line_69
       end
 
       def calculate_line_69a
-        return 0 unless @nyc_full_year_resident && @claimed_as_dependent == false
+        return 0 unless @nyc_full_year_resident && @direct_file_data.claimed_as_dependent? == false
```
mpidcock, line 447: Just a suggestion

```suggestion
        return 0 unless @nyc_full_year_resident && !@direct_file_data.claimed_as_dependent?
```

app/models/direct_file_data.rb
```diff
@@ -391,7 +404,8 @@ def attributes
       :fed_taxable_ssb,
       :fed_adjustments_claimed,
       :fed_total_adjustments,
-      :total_state_tax_withheld
+      :total_state_tax_withheld,
```
mpidcock, line 407: Delete this line

```suggestion
```

app/views/state_file/questions/federal_info/edit.html.erb
```diff
@@ -76,8 +77,8 @@
     <div style="background: #eaacea; padding: 15px; margin-bottom: 15px;">
       <h3>TODO ZONE: where do these things go?</h3>
 
-      <%= f.cfa_checkbox(:claimed_as_dep, "claimed as dependent?", options: { checked_value: "yes", unchecked_value: "no" }) %>
-
+      <%= f.state_file_qa_input_field :phone_daytime, "daytime phone number" %>
+      <%= f.state_file_qa_input_field :phone_daytime_area_code, "daytime phone area code" %>
```
mpidcock, line 81: Merge conflict

```suggestion
```

mpidcock (approved):
A few minor deletion/clean up comments, then lgtm

PR #3796, opened 2023-11-16
Add federal adjustments to ny 201 #186483743

mrotondo:
https://www.pivotaltracker.com/story/show/186483743


app/models/direct_file_data.rb
```diff
@@ -186,12 +189,23 @@ def fed_non_taxable_ssb
...
+        xml_label: "Educator Expenses"
+      },
+      fed_student_loan_interest: {
+        pdf_label: "stud loan ded",
+        xml_label: "Student Loan Interest Deduction"
+      }
+    }
+    adjustments.keys.each { |k| adjustments[k][:amount] = df_xml_value(k)&.to_i }
+    adjustments.select { |k, info| info[:amount].present? && info[:amount] > 0 }
   end
```
bengolder, line 205: Non-blocking suggestion:

It surprised me to see pdf labels in this file, because it looks like most of the existing code keeps the pdf keys in the ny201_pdf.rb


I think I would have expected those strings to be in the ny201_pdf.rb file but not here. I think of DirectFileData as a wrapper around the direct file return XML that does not need to know anything about the downstream PDFs, and that the Ny201Pdf object is responsible for translating data from direct file or elsewhere into the structure needed for the PDF. I'd prefer to move any pdf-specific logic or labels into the pdf code instead of keeping them in the direct file data.

Similarly, I would expect the It201 do define the XML labels it needs, even if they happen to be the same as the keys we use to pull from the fed return XML.

spec/models/direct_file_data_spec.rb
```diff
@@ -41,4 +41,56 @@
       end
     end
   end
+
+  describe '#fed_adjustments_claimed' do
```
bengolder, line 45: Praise: I really appreciate seeing this spec and the different cases it checks!

app/lib/efile/ny/it215.rb
```diff
@@ -62,7 +62,7 @@ def calculate_line_16
 
       def calculate_wk_c_line_2
         rates = [
-          [0, 5000, nil, nil, 0.30],
+          [-99999999999999, 5000, nil, nil, 0.30],
```
bengolder, line 65: Question: What did this line do and why was it deleted?

bengolder (approved):
Approved with some non-blocking suggestions and questions. I appreciate the unit test for direct file data

app/lib/efile/ny/it215.rb
mpidcock: I think that Mike discovered that the deduction math can cause the agi to go negative, so this change means a rate can still be calculated in that scenario.

app/models/direct_file_data.rb
mpidcock: I see what you are saying here, and the appeal was to make it easy to add more types of adjustments in the future (since it is my understanding that there are many). This location may not be the most ideal, but splitting these labels up by xml/pdf will make it harder to track in the future.

PR #3799, opened 2023-11-17
NY TR-579-IT Form

tahsinaislam:


loumoore (approved):
looks good! 

non-blocking: you could add a spec for the pdf filler that validates it's using field names from the pdf like we have for most of the others.

i am also curious where you got the esigned_at > 3 years ago requirement, cuz i didn't notice that on the arizona form and hoping i didn't miss anything!

PR #3800, opened 2023-11-17
Az 8879 e-file authorization form #186264976

loumoore:


app/controllers/state_file/questions/az_review_controller.rb
```diff
@@ -2,7 +2,7 @@ module StateFile
   module Questions
     class AzReviewController < QuestionsController
       def edit
-        @refund_or_owed_amount = current_intake.calculated_refund_or_owed_amount
+        @refund_or_owed_amount = current_intake.calculated_refund_or_owed_amount.abs
```
loumoore, line 5: fixed bug: was displaying negative if amount owed

app/controllers/state_file/questions/ny_review_controller.rb
```diff
@@ -2,7 +2,7 @@ module StateFile
   module Questions
     class NyReviewController < QuestionsController
       def edit
-        @refund_or_owed_amount = current_intake.calculated_refund_or_owed_amount
+        @refund_or_owed_amount = current_intake.calculated_refund_or_owed_amount.abs
```
loumoore, line 5: fixed same issue in NY review screen with negative amount owed

app/views/state_file/questions/az_review/edit.html.erb
```diff
@@ -5,7 +5,7 @@
   <div class="highlight-box highlight-box--blue spacing-below-15 padding-20">
     <div class="spacing-below-5">
       <p class="text--bold spacing-below-5"><%=@refund_or_owed_label %></p>
-      <p class="spacing-below-0"><%=number_to_currency(@refund_or_owed_amount) %></p>
+      <p class="spacing-below-0"><%=number_to_currency(@refund_or_owed_amount.abs) %></p>
```
loumoore, line 8: was printing out negative if taxes owed

app/views/state_file/questions/ny_review/edit.html.erb
```diff
@@ -5,7 +5,7 @@
   <div class="highlight-box highlight-box--blue spacing-below-15 padding-20">
     <div class="spacing-below-5">
       <p class="text--bold spacing-below-5"><%=@refund_or_owed_label %></p>
-      <p class="spacing-below-0"><%=number_to_currency(@refund_or_owed_amount) %></p>
+      <p class="spacing-below-0"><%=number_to_currency(@refund_or_owed_amount.abs) %></p>
```
loumoore, line 8: was printing out negative if taxes owed

app/lib/efile/az/az140.rb
```diff
@@ -68,7 +68,7 @@ def calculate
       end
 
       def refund_or_owed_amount
-        calculate_line_79
+        calculate_line_79 - calculate_line_80
```
loumoore, line 71: line 79 is just the refund, so i subtracted the amount owed on line 80 from it, which i believe is the intent of this method. @rickreyhsig could you double check in case I misunderstood? if this is correct, I think we might need a similar change on the NY side cuz I think it also has refund and amount owed on separate lines iirc

rickreyhsig: @loumoore the way I interpreted this initially, is that if `line63 - line78` would get written to line79 regardless if positive or negative, but coming back a second time it seems indeed to make more sense that only one would box filled would make more sense since there is either a refund or an amount owed.  Thanks for the change.  It seems the 201 form math may capture the intent of method which is returning (+) if refund and (-) if owed 

```
      def refund_or_owed_amount
        line_or_zero(:IT201_LINE_76) - line_or_zero(:IT201_LINE_62)
      end
```

app/lib/pdf_filler/az8879_pdf.rb
```diff
@@ -0,0 +1,76 @@
...
+      if @submission.data_source.spouse_esigned_yes?
+        answers["Spouse Signature"] = @submission.data_source.spouse.full_name
+        answers["Spouse Date Signed"] = @submission.data_source.spouse_esigned_at.to_date
+      end
+      answers
+    end
+
+    private
+
+    def calculated_fields
```
rickreyhsig, line 72: Is this being used anywhere?

loumoore: nope! good catch, i'll remove it

rickreyhsig (approved):
LGTM!

PR #3802, opened 2023-11-18
Ny 2023 pdfs 186486144

loumoore:
this is kind of a beast because of all the changes to IT-213. IT-214 and IT-215 were pretty simple. updated all of them to the new pdfs, and tried to handle as much of the IT-213 instructions as possible. couple TODOs still in there but I think it's overall a good step forward. 

this also pulls in the updated NY schema that matches the new pdfs, which i cherry-picked from travis' branch


app/lib/efile/ny/it201.rb
```diff
@@ -3,12 +3,13 @@ module Ny
...
       attr_reader :lines
 
-      def initialize(year:, filing_status:, intake:, direct_file_data:, dependent_count:, include_source: false)
+      def initialize(year:, filing_status:, intake:, direct_file_data:, eligibility_lived_in_state:, dependent_count:, include_source: false)
         @year = year
 
         @filing_status = filing_status # single, married_filing_jointly, that's all we support for now
         @intake = intake
         @direct_file_data = direct_file_data
+        @eligibility_lived_in_state = eligibility_lived_in_state
```
mpidcock, line 12: Could this be replaced with `@intake.eligibility_lived_in_state`, since `intake` should be an instance of `StateFileNyIntake`? Could simplify the initialize parameters.

app/lib/efile/ny/it201.rb
```diff
@@ -17,8 +18,8 @@ def initialize(year:, filing_status:, intake:, direct_file_data:, dependent_coun
           lines: @lines,
           filing_status: filing_status.to_sym,
           direct_file_data: direct_file_data,
-          federal_dependent_child_count: @intake.dependents.length,
-          federal_dependent_child_count_between_4_and_17: @intake.dependents.length, # TODO
+          eligibility_lived_in_state: eligibility_lived_in_state,
```
mpidcock, line 21: Same as other comment

app/lib/efile/ny/it213.rb
```diff
@@ -3,154 +3,267 @@ module Ny
...
-        set_line(:IT213_LINE_3, :calculate_line_3)
-        set_line(:IT213_LINE_4, -> { @federal_dependent_child_count})
-        set_line(:IT213_LINE_5, -> { @federal_dependent_child_count_between_4_and_17})
+        set_line(:IT213_LINE_1, -> { @eligibility_lived_in_state })
+        # If line 1 is false, stop, you do not qualify for the credit
+        if @eligibility_lived_in_state
+          set_line(:IT213_LINE_2, -> { @direct_file_data.fed_ctc_claimed })
+          set_line(:IT213_LINE_3, :calculate_line_3)
+          # If lines 2 and 3 are false, stop, you do not qualify for this credit
+          if @lines[:IT213_LINE_2].value == true || @lines[:IT213_LINE_3].value == true
```
mpidcock, line 22: I'm not as familiar with ruby, but I've seen `?.value` used pretty often for getting values from lines, and also seen `?` instead of `== true`. I'm curious if there are semantic differences or if it is a style preference. Would this also work?

```suggestion
          if @lines[:IT213_LINE_2]?.value || @lines[:IT213_LINE_3]?.value
```

mpidcock (commented):
Leaving some initial comments, will return later after doing some testing

app/lib/efile/ny/it201.rb
loumoore: yeah that's fine, i'll update it. i was following the pattern that was being used before but it look like it's been changed since i started working on this.

app/lib/efile/ny/it213.rb
loumoore: i don't think ?.value would work here, and i don't see that being used anywhere else myself. but you're right, the == true part in my code here is redundant, i can just check for thing.value instead of thing.value == true. so i'll go ahead and clean that up.

FWIW "thing_that_might_be_truthy_or_falsey?" with a question mark is a common naming convention in ruby , e.g. present? or nil?  or the *_yes? methods our enums produce, but you can't just call "?" on one of the TaxFormLine objects in the @lines array like that afaik.

mpidcock: thanks, that helped clear things up for me 🙏

app/lib/efile/ny/it201.rb
loumoore: actually, filing_status and direct_file_data are also coming from the intake object. not sure what the reason was for them being passed in separately instead of just passing intake before?

loumoore: gonna see about refactoring it but i wonder if this was helpful for setting up tests or something

mpidcock (approved):
I appreciated the comments in the it213 tax calculator. I was able to follow the steps and validate that the code matched the comments. Tracking/validating the xml and pdf logic is much harder, but I don't see any issues in the code. Only minor nits.

There are a few TODOs, would be nice if those can be captured in pivotal or resolved before merging.

app/lib/efile/ny/it201.rb
loumoore: i'm just testing this now, seems like it's gonna work out

app/lib/efile/ny/it213.rb
```diff
@@ -3,154 +3,292 @@ module Ny
...
           set_line(:IT213_WORKSHEET_A_LINE_10, :calculate_worksheet_a_line_10)
-          set_line(:IT213_LINE_6, -> { @lines[:IT213_WORKSHEET_A_LINE_10].value })
-          set_line(:IT213_LINE_7, -> { 0 }) # TODO revisit if worksheet C
+          set_line(:IT213_WORKSHEET_A_LINE_11, :calculate_worksheet_a_line_11)
+          set_line(:IT213_WORKSHEET_A_LINE_12, :calculate_worksheet_a_line_12)
+          # Is the amount on line 8 of this worksheet more than the amount on line 12?
+          # If No: Stop here. Enter the amount from line 8 here and on Form IT-213, line 6; and enter 0 on Form IT-213, line 7.
+          # If Yes: Enter the amount from line 12 here and on Form IT-213, line 6; and complete Worksheet B: Additional child tax credit amount.
+          set_line(:IT213_WORKSHEET_A_LINE_13, :calculate_worksheet_a_line_13)
+          if @lines[:IT213_WORKSHEET_A_LINE_8].value > @lines[:IT213_WORKSHEET_A_LINE_12].value
```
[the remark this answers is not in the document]
loumoore: ++ agreed

app/lib/efile/ny/it213.rb
```diff
@@ -3,154 +3,292 @@ module Ny
...
+            set_line(:IT213_LINE_15, -> { 0 })
+            set_line(:IT213_LINE_16, -> { 0 })
+          else
+            offboard
+            return
+          end
+        else
+          offboard
+        end
+      end
```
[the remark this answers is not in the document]
loumoore: hmm yeah i wanted to mimic the flow of the instructions as much as possible but i don't think the way i did it is quite right. i think i'll actually switch the two checks that might result in offboarding from "if" to "unless" so the first branch is offboard and otherwise you continue. that matches the instructions closest.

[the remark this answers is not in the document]
loumoore: nevermind, tried it and it just got messier. gonna go with your suggestion plus a comment 👍

app/lib/efile/ny/it201.rb
loumoore: done and pushed

PR #3809, opened 2023-11-22
State file importing federal data: return df sample if auth code is our fake one

jenny-heath:


embarnard (approved):
LGTM!

app/services/irs_api_service.rb
```diff
@@ -10,7 +10,7 @@ def self.df_return_sample
   end
 
   def self.import_federal_data(authorization_code, state_code)
-    unless server_url
+    if authorization_code == "abcdefg"
```
jenny-heath, line 13: leaving this comment for posterity: this is the main reason for this PR. demo was breaking because direct_file_data was not present on intakes, and i suspect it's because this method doesn't take into account whether someone clicked the fake data transfer button or the real one, and the reason for the breakage is that we changed the env vars on demo while testing yesterday

PR #3811, opened 2023-11-22
IT-201 PDF Fix

tahsinaislam:


app/lib/efile/ny/it201.rb
```diff
@@ -102,6 +102,7 @@ def refund_or_owed_amount
 
       private
 
+
```
mpidcock, line 105: Stray newline got added here
```suggestion
```

app/lib/submission_builder/ty2022/states/ny/documents/it201.rb
```diff
@@ -96,6 +101,8 @@ def document
             def calculated_fields
               @it201_fields ||= @submission.data_source.tax_calculator.calculate
             end
+
+
```
mpidcock, line 105: Two stray newlines got added

```suggestion
```

mpidcock (approved):
Minor cleanup if you feel like it

PR #3816, opened 2023-11-29
State file: add pending federal return page

jenny-heath:
image

image


app/assets/images/state-file-big-logo.svg
jenny-heath, line 1: this looks very similar to our existing illustration but it's not *exactly* the same

app/controllers/state_file/questions/pending_federal_return_controller.rb
```diff
@@ -0,0 +1,19 @@
+module StateFile
+  module Questions
+    class PendingFederalReturnController < QuestionsController
+      def self.show?(_intake)
+        false
```
embarnard, line 5: Is this false because this page will be directly linked on the IRS side?

config/locales/en.yml
```diff
@@ -2022,6 +2022,12 @@ en:
       ny_spouse_state_id:
         edit:
           title_html: Please provide information for <strong>your spouse’s</strong> state issued ID
+      pending_federal_return:
+        edit:
+          title_html: |
+            We’re sorry to keep you waiting. Your federal tax return must be accepted before you can file your state tax return.
```
embarnard, line 2028: why not just put in <p> tags instead of having two </br>

app/controllers/state_file/questions/pending_federal_return_controller.rb
jenny-heath: yes! i considered not putting it in the flow at all but in the end it was nice that the routes just come with the navigation and also it sort of demonstrates "where" clients would encounter this page

config/locales/en.yml
jenny-heath: good question, no reason

jenny-heath: actually i guess the reason is that it's already in an h1 because it's the only text on the page

bengolder (approved):
Looks great, thank you! Approved!

PR #3821, opened 2023-11-29
Add cache control to hub

embarnard:
This work comes out of [this BugCrowd ticket](https://tracker.bugcrowd.com/code-for-america-mbb/submissions/7b1c21b0-c571-4582-afe5-d3741735687f?duplicate[]=false&sort[]=severity-asc&substate[]=unresolved). Basically in the hub, a user could logout and then click the back button and return to the page that they were previously on. Adding this `set_cache_headers` on the hub basecontroller should clear the cache and prevent this from happening. 

References to this work came from [Arin's PR in GCF](https://github.com/codeforamerica/gcf-backend/pull/8866) and this [stackoverflow post ](https://stackoverflow.com/questions/711418/how-to-prevent-browser-page-caching-in-rails/748646#748646).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -480,6 +480,13 @@ class ApplicationController < ActionController::Base
     "application"
   end
 
+  def set_cache_headers
+    # prevents browser caching on pages set with before action
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = "Mon, 01 Jan 1990 00:00:00 GMT"
+  end
+
   rescue_from CanCan::AccessDenied do |exception|
     respond_to do |format|
       format.html do
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -480,7 +480,7 @@ class ApplicationController < ActionController::Base
     "application"
   end
 
-  def set_cache_headers
+  def set_no_cache_headers
     # prevents browser caching on pages set with before action
     response.headers["Cache-Control"] = "no-cache, no-store"
     response.headers["Pragma"] = "no-cache"
--- a/app/controllers/hub/admin/experiment_participants_controller.rb
+++ b/app/controllers/hub/admin/experiment_participants_controller.rb
@@ -1,8 +1,6 @@
 module Hub
   module Admin
-    class ExperimentParticipantsController < ApplicationController
-      include AccessControllable
-      before_action :require_sign_in
+    class ExperimentParticipantsController < Hub::BaseController
       load_and_authorize_resource
       layout "hub"
       before_action :redirect_on_prod, only: [:update]
--- a/app/controllers/hub/base_controller.rb
+++ b/app/controllers/hub/base_controller.rb
@@ -1,0 +1,7 @@
+module Hub
+  class BaseController < ApplicationController
+    include AccessControllable
+    before_action :set_cache_headers, :require_sign_in
+
+  end
+end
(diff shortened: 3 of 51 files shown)


app/controllers/application_controller.rb
```diff
@@ -480,6 +480,13 @@ def wrapping_layout
     "application"
   end
 
+  def set_cache_headers
+    # prevents browser caching on pages set with before action
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = "Mon, 01 Jan 1990 00:00:00 GMT"
+  end
```
bengolder, line 488: Not important, might be nice: I feel like there could be a clearer name for this method, like `set_no_cache_headers` `no_cache_headers` or `prevent_response_caching` or `do_not_cache_response`.

app/controllers/hub/admin/experiment_participants_controller.rb
```diff
@@ -1,8 +1,6 @@
 module Hub
   module Admin
-    class ExperimentParticipantsController < ApplicationController
-      include AccessControllable
-      before_action :require_sign_in
+    class ExperimentParticipantsController < Hub::BaseController
```
bengolder, line 3: Great refactor! I love this

app/controllers/hub/base_controller.rb
bengolder, line 1: YES! This is such a great move! Love it!

app/controllers/hub/bulk_actions/base_bulk_actions_controller.rb
```diff
@@ -1,14 +1,10 @@
...
-      include AccessControllable
-
-      layout "hub"
-
-      before_action :require_sign_in
+    class BaseBulkActionsController < Hub::BaseController
       load_and_authorize_resource :tax_return_selection
       before_action :load_clients, :load_template_variables
       before_action :load_edit_form, only: :edit
+      layout "hub"
```
bengolder, line 7: Would it make sense to also move `layout "hub"` into the `Hub::BaseController` or would that cause other problems?

spec/features/hub/cache_control_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  context "user on profile page, logs out and clicks back button", js: true do
+    it "cache is cleared and redirects to the login page" do
+      visit hub_user_profile_path
+      click_on 'Sign out'
+      page.evaluate_script('window.history.back()')
+      expect(page).to have_text "Sign in"
+    end
+  end
```
bengolder, line 17: Question for my own learning: Were you able to recreate the reported bug with this test before adding the no-cache headers?

app/controllers/hub/users/strong_passwords_controller.rb
bengolder, line 1: Question for my own learning: How did you figure out which controllers could inherit from the new hub base vs. which needed to be handled separately? Was it a matter of looking through all the controllers in the hub folder?

bengolder (approved):
This PR made my day, great refactor! I love the new base controller and the net negative number of lines! This is fantastic work!!

Approved with some questions for my own learning, and an unimportant "might be nice" suggestion.

app/controllers/hub/bulk_actions/base_bulk_actions_controller.rb
embarnard: Yeah I originally tried that but there were a few controllers that didn't include `layout "hub"` in the first place so I decided against it

spec/features/hub/cache_control_spec.rb
embarnard: Yes! Good question, the tests failed when I commented out the set_cache_headers method

app/controllers/hub/bulk_actions/base_bulk_actions_controller.rb
bengolder: I wonder if we should move it in, and change the few to have `layout "application"`

PR #3823, opened 2023-11-29
Add recipient address XML for 1099G

tahsinaislam:


app/models/state_file1099_g.rb
```diff
@@ -46,10 +46,10 @@ class StateFile1099G < ApplicationRecord
...
   def update_conditional_attributes
     if address_confirmation_yes?
-      self.recipient_city = nil
-      self.recipient_state = nil
-      self.recipient_street_address = nil
-      self.recipient_zip = nil
+      self.recipient_city = intake.direct_file_data.mailing_city
+      self.recipient_state = intake.direct_file_data.mailing_state
+      self.recipient_street_address = intake.direct_file_data.mailing_street
+      self.recipient_zip = intake.direct_file_data.mailing_zip
```
tahsinaislam, line 52: Changed recipient address information to be set to the default address that is being confirmed instead of NIL so that the value can be passed into the XML.

app/lib/submission_builder/ty2022/states/ny/documents/state1099_g.rb
```diff
@@ -29,6 +29,12 @@ def document
                 end
                 xml.RecipientSSN recipient.ssn
                 xml.RecipientName recipient.full_name
+                xml.RecipientUSAddress do
+                  xml.AddressLine1Txt form1099g.recipient_street_address
```
embarnard, line 33: is there a AddressLine2? just curious

app/views/state_file/questions/unemployment/_form.html.erb
```diff
@@ -48,8 +48,8 @@
             <% content_for :address_text do %>
               <div class="text--bold">Address</div>
               <div><%= current_intake.direct_file_data.mailing_street %></div>
-              <div><%= current_intake.direct_file_data.mailing_city %>, New
-                York <%= current_intake.direct_file_data.mailing_zip %></div>
+              <div><%= current_intake.direct_file_data.mailing_city %>,<%= state_full_name %>
```
embarnard, line 51: should there be a space after the comma? 
```suggestion
              <div><%= current_intake.direct_file_data.mailing_city %>, <%= state_full_name %>
```

embarnard (approved):
LGTM! some minor comments

app/lib/submission_builder/ty2022/states/ny/documents/state1099_g.rb
tahsinaislam: Yes for the Apt. But we don't have a separate field in the database for apt. I think the user would put it in the same line as the street address

PR #3829, opened 2023-11-30
State Tax Withholdings

tahsinaislam:


app/lib/efile/ny/it201.rb
```diff
@@ -472,11 +472,16 @@ def calculate_line_69a
       end
 
       def calculate_line_72
-        0 # TODO: Computed from W-2 and 1099G forms and their NYS wrapper, IT-2
+        total_state_taxes_withheld = @direct_file_data.total_state_tax_withheld
+        state_file_1099gs = @intake.state_file1099_gs
+        state_file_1099gs.each do |state_file_1099g|
+          total_state_taxes_withheld += state_file_1099g.state_income_tax_withheld
+        end
+        total_state_taxes_withheld
```
loumoore, line 480: am i correct in thinking that total_state_tax_withheld already has the sum of all W2 withholdings, and now you're adding the sum of all freelance/contracting work by summing up the 1099 withholdings too? if so then this makes sense, just wanted to double check that we're not double-counting anything.

PR #3830, opened 2023-11-30
tie off it-214 UI, tax calculations, and pdf generation

loumoore:
- deleted the IT-214 question page from the UI, which was a placeholder anyway
- commented out the tax calculations and pdf generation and related tests rather than deleting, in case there's any chance we'd want to bring support for this back later in the tax season

### DIFF
--- a/app/lib/efile/ny/it201.rb
+++ b/app/lib/efile/ny/it201.rb
@@ -19,11 +19,12 @@ module Efile
           lines: @lines,
           intake: @intake
         )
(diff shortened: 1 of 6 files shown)


bengolder (approved):
Approved! I'm assuming much of this is commented out because there is some chance that we may need to include IT-214 later. I don't have enough context to know whether commenting or deleting is the best choice for this and I trust that you'll say something when/if you think deletion would make sense. Thank you!

PR #3832, opened 2023-12-01
add household info to NY and AZ review pages 

loumoore:
and refactor the shared sections into a partial view


bengolder (approved):
Approved! This is great! I appreciate the refactor! Thanks Lou!

PR #3833, opened 2023-12-01
allow an xml node to be inserted after a specific sibling

loumoore:
which enables spouse ssn to be set without breaking xml validation

### DIFF
--- a/app/helpers/df_xml_crud_methods.rb
+++ b/app/helpers/df_xml_crud_methods.rb
@@ -3,7 +3,7 @@ module DfXmlCrudMethods
     node.at(selectors[key])&.text
   end
 
-  def create_or_destroy_df_xml_node(key, value)
+  def create_or_destroy_df_xml_node(key, value, after = nil)
     # Given a key path Root1 Root2 Leaf,
     #   if 'value' is truthy, creates the entire path <Root1><Root2><Leaf> if it doesn't exist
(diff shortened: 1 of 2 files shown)


bengolder (approved):
Nice! Great addition. Thank you!

app/helpers/df_xml_crud_methods.rb
```diff
@@ -15,7 +15,13 @@ def create_or_destroy_df_xml_node(key, value)
         next_node = current_node.at(node_name)
         # TODO: this just adds it as the next thing on the parent but if we want it to be schema-compliant it needs to be in the right order
```
mrotondo, line 16: Should we get rid of this comment as part of this change?

app/helpers/df_xml_crud_methods.rb
```diff
@@ -15,7 +15,13 @@ def create_or_destroy_df_xml_node(key, value)
...
         # (we might not care for this to be schema-compliant)
-        current_node = next_node || current_node.add_child("<#{node_name}/>").first
+        if next_node
+          current_node = next_node
+        elsif after && current_node.at(after)
```
mrotondo, line 20: It feels very unlikely, but this could incorrectly trigger in the case that we have a selector which includes one or more nodes that need to be created (i.e. the third branch of this `if` block will trigger) but where a node at the last level where the current XML structure matches the key also contains a node that matches the `after` name. Should we explicitly only apply the `after` logic for the final node (the leaf) of the selector?

PR #3835, opened 2023-12-02
Update NY Submitted page copy

bengolder:
Implements a small copy change: https://www.pivotaltracker.com/story/show/186527695

### DIFF
--- a/app/views/state_file/questions/submission_confirmation/edit.html.erb
+++ b/app/views/state_file/questions/submission_confirmation/edit.html.erb
@@ -23,9 +23,9 @@
   </p>
 
   <p>
-    <!-- TODO - replace with real link -->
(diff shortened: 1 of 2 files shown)


app/views/state_file/questions/submission_confirmation/edit.html.erb
```diff
@@ -23,9 +23,9 @@
   </p>
 
   <p>
-    <!-- TODO - replace with real link -->
-    <%= link_to t(".apply_to_state_benefits", state_name: States.name_for_key(params[:us_state].upcase)), "" %>
```
bengolder, line 27: It's not clear if we should keep this empty link or not, but it wasn't in the mock. I asked and can restore it if we need to keep it.

loumoore (approved):
looks good ship it!

PR #3841, opened 2023-12-04
Update AZ state ID info link and XML

mpidcock:


app/views/state_file/questions/ny_primary_state_id/_state_id.html.erb
```diff
@@ -33,9 +33,11 @@
         <% if options[:include_first_three_doc_num] == true %>
           <%= f.cfa_input_field(:first_three_doc_num, t("state_file.questions.primary_state_id.state_id.id_details.first_three_doc_num"), classes: ["form-width--short"]) %>
         <% end %>
-        <div class="spacing-above-25">
-          <a href="https://dmv.ny.gov/id-card/sample-photo-documents" target="_blank" rel="nofollow noopener"><%= t('state_file.questions.primary_state_id.state_id.info_link') %></a>
-        </div>
+        <% if options[:info_link] %>
```
jenny-heath, line 36: since they both have the info link, we probably don't *need* this—i'm assuming the purpose is to make it more flexible but we only have the 2 cases so i've been assuming no need to make it flexible. that said, it really doesn't matter whether we keep this condition or not

jenny-heath (approved):
looks good!

PR #3842, opened 2023-12-04
fix incorrect assignments in IT213 calculations and add test coverage

loumoore:


spec/lib/efile/ny/it201_spec.rb
loumoore, line 1: there are still a few more logic branches to cover in the IT213 calculations, which i'll make sure we have tickets filed for. these new tests exercise the branches that triggered this bug and weren't previously covered.

spec/lib/efile/ny/it201_spec.rb
```diff
@@ -39,20 +39,127 @@
...
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_4B].value).to eq(0)
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_5].value).to eq(18000)
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_6].value).to eq(2700)
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_7].value).to eq(0)
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_8].value).to eq(2700)
+        expect(instance.lines[:IT213_WORKSHEET_B_LINE_9].value).to eq(2700)
+        expect(instance.lines[:IT213_LINE_7].value).to eq(2700)
+        expect(instance.lines[:IT213_LINE_14].value).to eq(891)
+      end
+    end
```
tahsinaislam, line 162: nonblocking question out of curiosity: should this be in a separate it213 spec file or should this test suite be labeled more generally for ny calculations?

tahsinaislam (approved):
Accepted!

spec/lib/efile/ny/it201_spec.rb
loumoore: yeah good question! tl;dr is i don't feel super strongly about whether it'd be better to rename or do some extra work to initialize the it213/214/215 tests in a different place, but they always require it201 to be calculated first so this was the easiest path to adding test coverage across all the NY forms. 

back when i first tried writing tests for it213/214/215 i started them in their own files but in practice they are always calculated together starting with it201 and many use cases require it201 to be calculated first. i'm sure we could do the work of initializing it201 everywhere but the easiest route was to just test all of the NY calculations together in one place. we could rename it too, although i wasn't sure if breaking the standard naming convention was worth it. since what's going on here is sorta it201 delegating some of its calculations to other forms, i thought it wasn't too much of a stretch to just consider it all it201 testing 🤷

PR #3843, opened 2023-12-04
About page

rickreyhsig:
https://www.pivotaltracker.com/story/show/186317094

https://statefile.pr-3843.getyourrefund-testing.org/

http://statefile.localhost:3000/

AboutPage


app/controllers/state_file/questions/pending_federal_return_controller.rb
```diff
@@ -1,6 +1,6 @@
 module StateFile
   module Questions
-    class PendingFederalReturnController < QuestionsController
+    class PendingFederalReturnController < LandingPageController
```
rickreyhsig, line 3: May need a bigger refactor to make this PR and a few other pages `namespaced` entirely to LandingPages instead of to Questions.

bengolder (changes requested):
The content is good!

I also realized the ticket is not specific enough about the desired outcome. To rephrase the outcome of this story, when a user goes to the root path, `"/"`, on any fileyourstatetaxes domain (`statefile.localhost:3000/`, `demo.fileyourstatetaxes.org`, `staging.fileyourstatetaxes.org` or `www.fileyourstatetaxes.org`) they should see this page.

You are correct that we would not want this to be a questions controller. Currently, the root path for these domains will go to `StateFilePagesController#home`, which is the page that loads when you open `http://statefile.localhost:3000/` on your computer. The team had to do a similar thing when making the [CTC home page](https://www.getctc.org/en). Notice the [route constraint](https://github.com/codeforamerica/vita-min/blob/main/config/routes.rb#L377) and [home page routing](https://github.com/codeforamerica/vita-min/blob/main/config/routes.rb#L410-L414) for CTC.

rickreyhsig:
Had a flaky test here: https://app.circleci.com/pipelines/github/codeforamerica/vita-min/18898/workflows/ede0017f-8e97-4957-9ec2-b4dda8d2e718/jobs/48360/tests

But passes locally;  merging....

PR #3846, opened 2023-12-04
Decline terms and conditions page

tahsinaislam:


app/controllers/state_file/questions/declined_terms_and_conditions_controller.rb
```diff
@@ -0,0 +1,12 @@
+module StateFile
+  module Questions
+    class DeclinedTermsAndConditionsController < QuestionsController
+
+      def self.show?(_intake)
+        false
+      end
```
bengolder, line 7: I think we should be storing their answer and whether they consented ([similar to GYR](https://github.com/codeforamerica/vita-min/blob/main/app/models/intake.rb#L380)) and that this should be shown if they have any answer other than "yes"

app/views/state_file/questions/terms_and_conditions/edit.html.erb
```diff
@@ -9,10 +9,17 @@
...
+        <%= f.button :submit, name: "#{f.object_name}[consented_to_terms_and_conditions]", value: 'yes', class: 'button button--primary' do %>
+          <%= t(".accept") %>
+        <% end %>
+        <%= f.button :submit, name: "#{f.object_name}[consented_to_terms_and_conditions]", value: 'no', class: 'button' do %>
+          <%= t(".decline") %>
+        <% end %>
+      </div>
+    </div>
+  <%end %>
+<% end %>
```
tahsinaislam, line 25: I couldn't find simpler versions of the yes no buttons. It looks like every example in the codebase is similar to this format if I'm not mistaken.

bengolder: You're right! I can't find better examples either. This seems like a good opportunity for a refactor at some point!

bengolder (approved):
Approved! Great work! Thank you Tahsina!

PR #3849, opened 2023-12-05
Add devise to state file intake tables, require sign in past data transfer

jenny-heath:
temporarily redirect to root path


app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -0,0 +1,16 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :require_state_file_intake_login
+  end
+
+  private
+
+  def require_state_file_intake_login
+    if current_state_file_az_intake.blank? && current_state_file_ny_intake.blank?
```
bengolder, line 11: I wonder if we can make one clear method that we can use, but I'm not sure exactly what it would be.

bengolder (approved):

bengolder (approved):
Great! This is super helpful, thank you for getting this done so quickly! Approved! The only non-blocking thing I would request is to add comments to the navigation files so that it's easier to understand where sign in occurs.

PR #3850, opened 2023-12-06
Add other filing options offboarding

embarnard:
**`questions/unemployment/new` page**
Before changes:
Screenshot 2023-12-06 at 12 27 15 PM

With changes:
Screenshot 2023-12-05 at 7 34 56 PM

**`/questions/other_filing_options?us_state=az` page (new)**
Screenshot 2023-12-06 at 12 27 55 PM

### DIFF
--- a/app/controllers/state_file/questions/other_filing_options_controller.rb
+++ b/app/controllers/state_file/questions/other_filing_options_controller.rb
@@ -1,0 +1,6 @@
+module StateFile
+  module Questions
+    class OtherFilingOptionsController < QuestionsController
+    end
+  end
+end
--- a/app/controllers/state_file/questions/other_filing_options_controller.rb
+++ b/app/controllers/state_file/questions/other_filing_options_controller.rb
@@ -1,6 +1,7 @@
 module StateFile
   module Questions
     class OtherFilingOptionsController < QuestionsController
+      def edit; end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/views/state_file/questions/other_filing_options/edit.html.erb
```diff
@@ -0,0 +1,23 @@
+<% title = t(".title") %>
+
+<% content_for :page_title, title %>
+<% content_for :card do %>
+  <h1 class="h2"><%= title %></h1>
+  <p><%= t(".subtitle", state: (params[:us_state] == "az" ? "Arizona" : "New York")) %></p>
```
rickreyhsig, line 6: I would suggest using the `States.name_for_key` helper here:

```ruby
irb(main):042> States.name_for_key('ny'.upcase)
=> "New York"
irb(main):043> States.name_for_key('az'.upcase)
=> "Arizona"

```

app/controllers/state_file/questions/other_filing_options_controller.rb
```diff
@@ -0,0 +1,6 @@
+module StateFile
+  module Questions
+    class OtherFilingOptionsController < QuestionsController
+    end
```
rickreyhsig, line 4: I would suggest adding and empty edit action here to prevent confusion since on the config and the template file you reference it.

PR #3852, opened 2023-12-06
update NY county and school district screens to have a blank option to start

loumoore:
also updated styles on those pages

### DIFF
--- a/app/controllers/state_file/questions/ny_county_controller.rb
+++ b/app/controllers/state_file/questions/ny_county_controller.rb
@@ -4,23 +4,8 @@ module StateFile
   module Questions
     class NyCountyController < AuthenticatedQuestionsController
       include ReturnToReviewConcern
-      
-      def edit
-        @counties = counties_from_csv
-        super
-      end
 
       private
-
-      def counties_from_csv
-        csv_file_path = Rails.root.join('docs', 'ny_school_districts.csv')
-        csv_content = File.read(csv_file_path)
-
-        io = StringIO.new(csv_content)
-        CSV.parse(io, headers: true).map do |row|
-          row['County']
-        end.uniq
-      end
 
       def illustration_path; end
     end
--- a/app/helpers/vita_min_form_builder.rb
+++ b/app/helpers/vita_min_form_builder.rb
@@ -45,6 +45,44 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
     label_text.html_safe
   end
 
+  def vita_min_state_file_select(
+    method,
+    label_text,
+    collection,
+    options = {},
+    &block
+  )
+
+    html_options = {
+      class: "select__element",
+    }
+
+    label_class = options[:label_class] || ""
+    label_class += options[:hide_label] ? " sr-only" : ""
+    formatted_label = label(
+      method,
+      label_contents(
+        label_text,
+        options[:help_text],
+        options[:optional],
+        ),
+      class: label_class,
+      )
+    html_options_with_errors = html_options.merge(error_attributes(method: method))
+
+    html_output = <<~HTML
+          <div class="form-group#{error_state(object, method)}">
+            #{formatted_label}
+            <div class="select">
+              #{select(method, collection, options, html_options_with_errors, &block)}
+            </div>
+            #{errors_for(object, method)}
+          </div>
+        HTML
+
+    html_output.html_safe
+  end
+
   def vita_min_select(
     method,
     label_text,
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -510,6 +510,7 @@ en:
     search: Search
     security_html: Your information is secure and the information you provide will be handled in accordance with our %{privacy_link}.
     select_file: Select file
+    select_prompt: "-Select-"
     send: Send
     send_message: Send message
     separated: Separated
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -512,6 +512,7 @@ es:
     search: Búsqueda
     security_html: Su información está segura y la información que proporcione será tratada de acuerdo con nuestro %{privacy_link}.
     select_file: Seleccionar archivo
+    select_prompt: "-Seleccionar-"
     send: Enviar
     send_message: Enviar un mensaje
     separated: Separado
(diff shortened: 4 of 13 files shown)


app/controllers/state_file/questions/ny_county_controller.rb
```diff
@@ -4,24 +4,9 @@ module StateFile
...
-      def counties_from_csv
-        csv_file_path = Rails.root.join('docs', 'ny_school_districts.csv')
-        csv_content = File.read(csv_file_path)
-
-        io = StringIO.new(csv_content)
-        CSV.parse(io, headers: true).map do |row|
-          row['County']
-        end.uniq
-      end
-
```
loumoore, line 24: moved to application_helper

app/helpers/vita_min_form_builder.rb
```diff
@@ -45,6 +45,44 @@ def h1_label_contents(label_text, help_text, optional = false)
...
+            <div class="select">
+              #{select(method, collection, options, html_options_with_errors, &block)}
+            </div>
+            #{errors_for(object, method)}
+          </div>
+        HTML
+
+    html_output.html_safe
+  end
+
```
loumoore, line 85: did all this just to add the label_class option, because the label on these select inputs should be bold. and looking across the state file designs, i think whether it will be bold might depend on whether there's a single question or if it's part of a bigger form group, in which case it should be an option. and the vita_min_select function is already taken, but it implements the labels as h1s, which i don't think we want either. so i'm not super in love with this but i think we need something different than what we've got 🤷

config/locales/es.yml
```diff
@@ -512,6 +512,7 @@ es:
     search: Búsqueda
     security_html: Su información está segura y la información que proporcione será tratada de acuerdo con nuestro %{privacy_link}.
     select_file: Seleccionar archivo
+    select_prompt: "-Seleccionar-"
```
loumoore, line 515: maybe i should have put this under the state_file namespace to avoid i18n-tasks from complaining instead? but i think the correct translation is pretty clear?

spec/controllers/state_file/questions/ny_county_controller_spec.rb
```diff
@@ -7,17 +7,6 @@
...
-    it "assigns the correct data structure to @counties" do
-      get :edit, params: { us_state: "ny" }
-
-      counties = assigns(:counties)
-      expect(counties).to include('Montgomery')
-      expect(counties).to include('Nassau')
-      expect(counties).to eq counties.uniq
-    end
-  end
-
```
loumoore, line 20: moved to application_helper_spec

spec/controllers/state_file/questions/ny_school_district_controller_spec.rb
```diff
@@ -7,17 +7,6 @@
...
-    it "assigns the correct data structure to @school_districts including combined district names" do
-      get :edit, params: { us_state: "ny" }
-
-      school_districts = assigns(:school_districts)
-      expect(school_districts).to include(['Bellmore-Merrick CHS North Bellmore', 'Bellmore-Merrick CHS North Bellmore'])
-      expect(school_districts).to include(['Carle Place', 'Carle Place'])
-      expect(school_districts).to eq school_districts.uniq
-    end
-  end
-
```
loumoore, line 20: moved to application_helper_spec

app/controllers/state_file/questions/ny_county_controller.rb
bengolder: Non-blocking preference: I appreciate the effort to refactor, my instinct is that this would be clearer in the controller, because the data is specific to this particular page. I wouldn't put this in application_helper unless I expected it to be used by multiple controllers, ideally only if used by many of them, because it makes the method available to all controllers in the app. Here's another option, for your consideration:

```ruby
class NyCountyController < AuthenticatedQuestionsController
  include ReturnToReviewConcern
  helper_method :counties_from_csv
  # ...
  def counties_from_csv
    # ...
  end
end
```

I'd also recommend checking out `config/initializers/nys_data.rb` and `app/lib/efile/ny/constants.rb`. The `counties_from_csv` method, as currently implemented, will load the same data from the file system with every new request. If instead we load the options from the filesystem during the preload process, they can hang out in memory and won't need to be reloaded.

app/helpers/vita_min_form_builder.rb
bengolder: That sounds frustrating to have to do. I believe it's possible to update the styles with css only, so I don't think this is necessary. We also have [a story](https://www.pivotaltracker.com/story/show/186624747) for updating select elements across all of state file.

config/locales/en.yml
```diff
@@ -510,6 +510,7 @@ en:
     search: Search
     security_html: Your information is secure and the information you provide will be handled in accordance with our %{privacy_link}.
     select_file: Select file
+    select_prompt: "-Select-"
```
bengolder, line 513: For a generic default, I'd remove the dashes, maybe say "Select option". My first instinct would have been to use "Select county" and "Select school district" but I'm not sure whether that would look okay visually.

spec/helpers/application_helper_spec.rb
```diff
@@ -44,4 +44,22 @@
...
+  end
+
+  describe "ny_school_district_options_for_select" do
+    it "loads the correct data structure for NY School Districts" do
+      school_districts = ny_school_district_options_for_select('Nassau')
+      expect(school_districts).to include(['Bellmore-Merrick CHS North Bellmore', 'Bellmore-Merrick CHS North Bellmore'])
+      expect(school_districts).to include(['Carle Place', 'Carle Place'])
+      expect(school_districts).to eq school_districts.uniq
+    end
+  end
```
bengolder, line 64: I appreciate that you added unit tests for this!

app/controllers/state_file/questions/ny_county_controller.rb
loumoore: i did this because it doesn't work in the controller, the @counties variable is lost when validation fails. i assume that's solvable somehow but it wasn't immediately obvious to me, and when i started to poke around i noticed that we frequently use this pattern to load options from application_helper instead of in controllers. 

i didn't implement the csv loading code, but the caching sounds useful, i'll look into that before i merge.

app/helpers/vita_min_form_builder.rb
loumoore: re-checked the designs and confirmed, some "form-question" labels are bold and others are not, and there's nothing i can see to differentiate them in the code currently without adding a class. so i'm gonna leave this for now, maybe later designs will change or someone will be smarter and we can get rid of it.

PR #3853, opened 2023-12-06
customized 404 and 500 error pages for state file namespace

loumoore:


loumoore:
i looked into some other ways to do this, but could only solve more elegantly for the 404 page, which can be handled in routes. that doesn't seem possible for the 500 page though, since rails is unaware of namespaces at the time the exeption app handles them (as far as i can tell).

so the approach i used is similar to how we were already detecting that a page was part of the hub: I added an application_controller method that looks at the namespace of the current class as well as the request domain to determine whether the page is part of state_file. that method is then used in the application layout to decorate pages with the right header and footer, and in the 404 and 500 pages to set the right content.

embarnard (approved):
Looks good!

PR #3855, opened 2023-12-06
Add deduction amount for MFS filing status and only show spouse name if present in 201 and return header

embarnard:


embarnard:
Tiffany was seeing these XML errors when filing MFS and adding spouse SSN for a New York client 
Screenshot 2023-12-06 at 10 26 52 AM

loumoore (approved):

PR #3859, opened 2023-12-08
NY IT2 PDF

tahsinaislam:


app/lib/pdf_filler/ny_it2_pdf.rb
```diff
@@ -0,0 +1,60 @@
...
+      w2 = @kwargs[:w2]
+      answers = {
+        'Box_a' => w2.employee_ssn,
+        'Box_b' => w2.employer_ein,
+        'Box_c1' => w2.employer_name,
+        'Box_c2' => w2.employer_street_address,
+        'Box_c3_city' => w2.employer_city,
+        'Box_c3_state' => w2.employer_state,
+        'Box_c3_zip' => w2.employer_zip_code,
+        'Box_c3_county' => 'USA', #TODO check if other countries allowed
```
rickreyhsig, line 24: This is county, not country

app/models/w2.rb
```diff
@@ -65,7 +67,7 @@ class W2 < ApplicationRecord
   enum box13_retirement_plan: { unfilled: 0, yes: 1, no: 2 }, _prefix: :box13_retirement_plan
   enum box13_third_party_sick_pay: { unfilled: 0, yes: 1, no: 2 }, _prefix: :box13_third_party_sick_pay
 
-  delegate :first_name, :middle_initial, :suffix, :last_name, :ssn, to: :employee_obj, allow_nil: true, prefix: :employee
+  delegate :first_name, :middle_initial, :suffix, :last_name, to: :employee_obj, allow_nil: true, prefix: :employee
```
rickreyhsig, line 70: Is this because we are saving the SSN on the W2 record?

tahsinaislam: yes!

app/lib/pdf_filler/ny_it2_pdf.rb
tahsinaislam: the pdf field name is county when i dump the data fields. But the label on the pdf outside the box is Country

spec/controllers/ctc/portal/w2s/misc_info_controller_spec.rb
```diff
@@ -14,6 +14,7 @@
   end
 
   before do
+    skip
```
tahsinaislam, line 17: paired with Em on this. Skipping only CTC related tests that are failing from W2 model changing from has_one w2_box14 to has_many: w2_box14

app/models/w2.rb
bengolder, line 1: I don't think we should be modifying this model for state file work. If we truly need a model, then we should create an independent StateFileW2 model, similar to the way we handled dependents. But generally we should only need a model for this information if we are allowing someone to edit the information.

app/models/w2.rb
```diff
@@ -27,6 +27,7 @@
 #  creation_token                     :string
 #  employee                           :integer          default("unfilled"), not null
 #  employee_city                      :string
+#  employee_ssn                       :string
```
bengolder, line 30: We should never store SSN information without ensuring that it is encrypted

app/lib/pdf_filler/ny_it2_pdf.rb
```diff
@@ -14,39 +14,40 @@ def initialize(submission, kwargs)
...
+      #   'Box_17a' => w2.w2_state_fields_group.box17_state_income_tax,
+      #   'Box_18a' => w2.w2_state_fields_group.box18_local_wages,
+      #   'Box_19a' => w2.w2_state_fields_group.box19_local_income_tax,
+      #   'Box_20a' => w2.w2_state_fields_group.box20_locality_name,
+      # }
+      # w2.w2_box14.each_with_index do |box14, index|
+      #   break if index >= 4
+      #   answers["14#{('a'..'d').to_a[index]}_description"] = box14.other_description
+      #   answers["Box_14#{('a'..'d').to_a[index]}"] = box14.other_amount
+      # end
```
mpidcock: Wondering if this can be removed? Or if it is likely to be useful to reference later

tahsinaislam: yes it was removed!

bengolder (approved):
Excellent job!!

bengolder (approved):
Excellent job!!

PR #3866, opened 2023-12-09
update test xml for multiple dependents with unique SSNs, and add test coverage for this case

loumoore:


bengolder (approved):
Thank you! I probably entered the same ssn for all of them on Direct File. I'll have to add this tip to the DF Testing guidance

PR #3869, opened 2023-12-13
negative-value-check-#186665230

tofarr:
Manual sales tax entry on state forms is now limited to between 7 and 125


bengolder (approved):
This is great! Thanks Tim! Assuming you'll update the 3 to a 7, feel free to merge!

PR #3870, opened 2023-12-13
[temp branch just for reference] Require auth #186527604

jenny-heath:


jenny-heath:
story is done, closing

PR #3884, opened 2023-12-14
Added flag for state_file to admin role and new endpoint

tofarr:
At Em's convincing, we introduced a new flag on the admin role rather than a full new role for this. I think it makes sense given that some users will need both `admin` and `state_admin` privileges.

The hub/admin/tools view has a State File section if the flag is present
image

Currently there is nothing in this view, but it is protected by the new flag:
image

Specs and Seeds were updated with this new flag.

### DIFF
--- a/spec/controllers/hub/state_file_controller_spec.rb
+++ b/spec/controllers/hub/state_file_controller_spec.rb
@@ -23,9 +23,9 @@ describe Hub::StateFileController do
         sign_in user
       end
 
-      it "allows them to see the page" do
+      it "is forbidden" do
         get :index
-        expect(response).not_to be_ok
+        expect(response).to be_forbidden
       end
     end
   end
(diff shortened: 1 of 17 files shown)


embarnard (approved):

spec/controllers/hub/state_file_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      let(:user) { create(:admin_user) }
+
+      before do
+        sign_in user
+      end
+
+      it "allows them to see the page" do
+        get :index
+        expect(response).not_to be_ok
+      end
```
bengolder, line 29: Would expect this to be `it "is forbidden"`

```
expect(response).to be_forbidden
```

bengolder (approved):
Thank you so much! Approved!

PR #3887, opened 2023-12-15
Style updates as requested

tofarr:
- Focus color is now green (Rather than yellow - this also applies to links)
- Checkbox checked is now black (And thicker)
- Radio checked is now black
- Buttons now have hover style mimicing focus style
- Slight transition on focus state
image
image
image
image
image

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -365,6 +365,68 @@
     font-size: 2.5rem;
     font-weight: normal;
   }
+
+  .text-input,
+  a,
+  .button,
+  .radio-button,
+  .radio-button input[type='radio'],
+  .select__element,
+  .checkbox,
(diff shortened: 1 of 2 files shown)


loumoore (approved):
Looks good Tim! Two non-blocking observations, take em or leave em:
- There are a couple almost-but-not-quite-the-same rgb colors. You could use a variable there instead for consistency. Color variables are in _variables.scss.
- I think "accent-color" might be an option instead of the hacks to make the radio buttons and checkboxes black, but i'm not sure about browser compatibility. Your approach might be better!

tofarr:
@loumoore  - Thank you for your feedback - I really appreciate it. (You rock!) - I used a variable for the color as suggested.

I tried the "accent-color" before my current implementation but unfortunately it did not seem to work as expected in either chrome or safari, so hence the current implementation using `:after` and `background-image`

PR #3890, opened 2023-12-15
Framework for State e-file submission status pages

rickreyhsig:


app/controllers/hub/efile_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Hub
+  class EfileController < Hub::BaseController
+    include FilesConcern
+    authorize_resource
```
embarnard, line 4: we should always load the resource first and then authorize it because otherwise the authorize will have nothing to check 

see more here:https://docs.google.com/document/d/10WNXTBOHogniT1YWcNR2q7EdGmcDOKznAQXRQZnoLHQ/edit or here https://github.com/codeforamerica/vita-min/pull/3481/files

app/controllers/hub/efile_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    load_resource except: [:index, :show]
```
embarnard, line 5: but actually for lines 4 and 5 we can just replace it with a single `load_and_authorize` call and we will have to specify the name of the resource/model we are using since it doesn't match the name of controller "Efile" so I would recommend replacing these two lines with: 

```
load_and_authorize_resource :efile_submission, through: :intake, only: [:index, :show]
```
might have to play around with the through intake part to make sure its correct, it would be good to add a test case if not now then eventually that will check that a user can access the correct selection of efile submissions on this page

tofarr (approved):

app/controllers/hub/efile_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    layout "hub"
+    def index
+      @efile_submissions = EfileSubmission.includes(:efile_submission_transitions, tax_return: [:client, :intake]).most_recent_by_current_year_tax_return.page(params[:page])
+      @efile_submissions = @efile_submissions.in_state(params[:status]) if params[:status].present?
+    end
+
+    def show
```
embarnard, line 12: our State-file intakes do not have clients or tax returns unlike CTC where we basing this page off of  -- not sure if the plan is to add tax returns or to change the model so they `has_one :intake` not `through: :tax_return`. But I'm guessing in any case that this show page should be using the StateFile intake models instead of the client

app/controllers/hub/efile_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      client = Client.find(params[:id])
+      authorize! :read, client
+      @client = Hub::ClientsController::HubClientPresenter.new(client)
+      # Eager-load tax returns with submissions and data we may need to render
+      @tax_returns = client.tax_returns.includes(:efile_submissions, efile_submissions: :fraud_score).where.not(tax_returns: {efile_submissions: nil}).load
```
embarnard, line 17: we are also not calculating this fraud score for State file

embarnard (changes requested):
General scaffolding looks good! But I think we will have to convert these pages so that they just take in Statefile intakes and efile submissions and do away with the tax return and client calls that are leftover from CTC

app/controllers/hub/efile_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Hub
+  class EfileController < Hub::BaseController
```
jenny-heath, line 2: I would name this something more specific to state file. i think it's a little confusing that the current pages for viewing efile-related things are in the `EfileSubmissionsController` while this is just `EfileController`. Consider either namespacing under a `StateFile` module or including state file in the name of the controller

app/controllers/hub/efile_controller.rb
rickreyhsig: I don't have access to the Google docs, can you grant me access @embarnard?

app/models/efile_submission.rb
```diff
@@ -48,6 +48,11 @@ class EfileSubmission < ApplicationRecord
     where(data_source_type: STATE_INTAKE_CLASS_NAMES)
   }
 
+  scope :not_for_state_filing, lambda {
+    where.not(data_source_type: STATE_INTAKE_CLASS_NAMES)
+         .or(where(data_source_type: nil))
+  }
```
rickreyhsig, line 54: Can probably remove this file

rickreyhsig: Scope**

app/controllers/hub/efile_controller.rb
rickreyhsig: Done

app/controllers/hub/efile_controller.rb
rickreyhsig: Done

app/controllers/hub/efile_controller.rb
rickreyhsig: Done

PR #3892, opened 2023-12-18
State file: make RequestVerificationCodeForLoginJob work for sms when there is a matching intake

jenny-heath:


app/jobs/request_verification_code_for_login_job.rb
```diff
@@ -31,12 +31,20 @@ def perform(email_address: nil, phone_number: nil, locale:, visitor_id:, service
...
+        body = case service_type
+               when :ctc
+                 I18n.t("verification_code_sms.no_match_ctc", url: url, locale: locale)
+               when :gyr
+                 I18n.t("verification_code_sms.no_match_gyr", url: url, locale: locale)
+               when :statefile_az
+                 I18n.t("state_file.intake_logins.no_match_sms", url: "#{url}/az/questions/landing-page", locale: locale)
+               when :statefile_ny
+                 I18n.t("state_file.intake_logins.no_match_sms", url: "#{url}/ny/questions/landing-page", locale: locale)
+               end
```
bengolder, line 47: Non-blocking:
- How much can/should we pull into methods on multi_tenant_service? 
- Can we use a better method for determining the URL?

bengolder: I think it makes sense to use this landing page as the .url value for state file. I would also find it helpful to have a comment saying that this is a "start page" or to rename the method to give more context about how it is expected to be used.

app/services/client_login_service.rb
```diff
@@ -28,6 +28,10 @@ def can_login_by_email_verification?(email_address)
   end
 
   def can_login_by_sms_verification?(sms_phone_number)
-    service_class.accessible_intakes.where(sms_phone_number: sms_phone_number, sms_notification_opt_in: "yes").exists?
+    if service_class == Intake::CtcIntake || service_class == Intake::GyrIntake
+      return service_class.accessible_intakes.where(sms_phone_number: sms_phone_number, sms_notification_opt_in: "yes").exists?
+    elsif service_class == StateFileAzIntake || service_class == StateFileNyIntake
+      return service_class.accessible_intakes.where(phone_number: sms_phone_number).exists?
```
bengolder, line 34: I think it would make sense to use `can_be_authenticated` here in place of `accessible_intakes`

bengolder (approved):
Approved with non-blocking suggestions for improvement, thank you!

PR #3894, opened 2023-12-18
Schema updates from the IRS

tofarr:
This is not actually an update to the schema for the 1040, which was my original concern - but one that is required anyway.


embarnard (approved):
Looks good! But wonder how to fix the failing test -- forgot if we've had to add the zip to test environments in the past

PR #3897, opened 2023-12-19
Create dynamic headlines for individual taxpayers vs MFJ for state-specific questions

embarnard:


app/views/state_file/questions/az_incarcerated/edit.html.erb
```diff
@@ -1,17 +1,15 @@
...
+  <h1 class="h2"><%= title %></h1>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
     <div class="blue-group">
-      <%= f.cfa_radio_set(:was_incarcerated, collection: [
+      <%= f.cfa_radio_set(:was_incarcerated, label_text: title, collection: [
         { value: "yes", label: t("general.affirmative") },
         { value: "no", label: t("general.negative") },
-      ]) %>
+      ], legend_class: "sr-only") %>
```
mpidcock, line 12: Is this some general consistency clean-up?

config/locales/en.yml
```diff
@@ -1857,16 +1857,21 @@ en:
             <li>money spent on fundraisers including raffles, bingo; or</li>
             <li>items purchased for a fundraiser.</li></ul>
           learn_more_title: What does this mean?
-          title: Did you make charitable contributions (cash and/or non-cash) to a qualifying organization in 2023?
+          title:
+            one: Did you make charitable contributions (cash and/or non-cash) to a qualifying organization in %{tax_year}?
+            other: Did you or your spouse make charitable contributions (cash and/or non-cash) to a qualifying organization in %{tax_year}?
```
mpidcock, line 1862: Maybe consider calling these `not_mfj`/`mfj` to be more descriptive, but it's not a big deal

mpidcock (commented):
There's one more bit of text that needs to be fixed and a minor suggestion. Appreciate taking the opportunity to make some spots a bit cleaner too, like the variable tax year.

config/locales/en.yml
embarnard: @mpidcock this count feature is something that is built into to [I18nt](https://www.i18next.com/translation-function/plurals) so I wouldn't be able to change it into  `not_mfj`/`mfj` but I could change it into something like `not_mfj_one` and `mfj_other` although I prefer to keep it `one`/`other` since that has been the pattern in GYR

app/views/state_file/questions/az_incarcerated/edit.html.erb
embarnard: @mpidcock good question! yeah this was something I noticed on some pages included `id="main-question"` in the main header. I wasn't sure what this was referencing so I called Ben about it. But basically we concluded that these `main-question` ids were not referencing anything without an `aria-describedby` being used anywhere. Since there wasn't a param field where we could feed an `aria-describedby` into the `cfa_radio_set` item we decided to add the label text to the radio set with "screen reader only" class that would allow the screen reader to associate it to the radio options. Ideally we would do an accessibility check and check the screenreader could read the whole app in the intended way eventually but not sure if this will happen in the near future so just opting to change them as I come across the pages for now.

config/locales/en.yml
```diff
@@ -1857,16 +1857,21 @@ en:
...
+            other: Were you or your spouse incarcerated in a county, state, or federal prison for at least 60 days in %{tax_year}?
       az_prior_last_names:
         edit:
           prior_last_names_label: 'Enter all last names used in the last four years. (Ex: Smith, Jones, Brown)'
-          title1: Did you file with a different last name in the last four years?
-          title2: This includes tax years 2019-2022.
+          subtitle: This includes tax years 2019-2022.
+          title:
+            one: Did you file with a different last name in the last four years?
+            other: Did you or your spouse file with a different last name in the last four years?
```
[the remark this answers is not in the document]
embarnard: @mpidcock good catch! I'll make the change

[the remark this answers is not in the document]
embarnard: Screenshot 2023-12-21 at 11 46 51 AM
done!

config/locales/en.yml
mpidcock: ah, I was worried it might be something like that but couldn't figure out where the names come from. thanks, and I agree that keeping it is good to match GYR

PR #3899, opened 2023-12-19
style disabled text inputs

loumoore:


app/views/state_file/questions/unemployment/_form.html.erb
```diff
@@ -64,7 +64,7 @@
             <div class="blue-group">
               <p class="form-question spacing-below-25"><%= t('state_file.questions.unemployment.edit.confirm_address_html') %></p>
               <div class="form-question spacing-below-15">
-                <div class="text--bold">Address</div>
+                <div class="text--bold"><%= t('general.address') %></div>
```
loumoore, line 67: noticed this un-externalized string while i was in this file

app/views/state_file/questions/unemployment/_form.html.erb
```diff
@@ -100,7 +100,6 @@
           <div class="form-group-tight">
             <%= f.cfa_input_field(:unemployment_compensation, t('state_file.questions.unemployment.edit.unemployment_compensation_html'), classes: ["form-width--long"]) %>
             <%= f.cfa_input_field(:federal_income_tax_withheld, t('state_file.questions.unemployment.edit.federal_income_tax_withheld_html'), classes: ["form-width--long"]) %>
-            <%= f.cfa_input_field(:unfillable, t('state_file.questions.unemployment.edit.box_10a_html'), classes: ["form-width--long"], options: { value: state_full_name, disabled: true }) %>
```
loumoore, line 103: this field is no longer in the figma designs, so i figured the right move was to delete it rather than keep it with the disabled styles

tofarr (approved):
Looks great - good catch on the uninternationalized string

PR #3900, opened 2023-12-19
add copy to the bottom of every state specific page for Ny and AZ

loumoore:
if there's some way to do this without touching so many files that'd be cool, but i'm not sure there is. implemented as a concern in a fashion that would allow other pages to show a custom message below the button too if desired.


tofarr (commented):
Looks good - I wonder would it be possible to include the concern in the authenticated_queations_controller so we don't have to touch all the other files?

bengolder:
I agree with Tim's suggestion. We intend the AuthenticatedQuestionsController to be the base controller for any page that someone would be able to return to and see their previously entered questions. So it's a great fit for this specific piece of copy.

bengolder (approved):
Approved! It looks like there is no super easy 1:1 match for the copy. I think you're approach here is the best we could expect in terms of concision. Thank you!

PR #3901, opened 2023-12-19
Reverse ordering on date picker for license expiration

tofarr:
So the first options people see are close to now, rather than 2073
image


bengolder (approved):
Approved!

PR #3904, opened 2023-12-20
Xml updates from Javier testing #186684699

tofarr:


app/lib/efile/ny/it201.rb
```diff
@@ -75,7 +75,7 @@ def calculate
         @it213.calculate
         set_line(:IT201_LINE_63, -> { @lines[:IT213_LINE_14].value })
         @it215.calculate
-        set_line(:IT201_LINE_65, -> { @lines[:IT215_LINE_16].value })
+        set_line(:IT201_LINE_65, :calculate_line_65)
```
mpidcock, line 78: I see that there is a calculate method for it215, which also conditionally sets line 1 instead of exiting if line 1 is false. do you think that is worth updating? it would replace the need for :calculate_line_65

mpidcock: it works as is, just something to consider

mpidcock (approved):
Approved with small suggestions

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -130,8 +130,10 @@ def included_documents
           def supported_documents
             tax_calculator = @submission.data_source.tax_calculator
             calculated_fields = tax_calculator.calculate
+            # I think the below should be calculated fields
```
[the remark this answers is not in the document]
tofarr: I removed that comment - I was originally thinking to merge the logic, but then I figured they were actually calculating different things that only looked the same and it was better to keep them apart

app/lib/efile/ny/it201.rb
tofarr: With the `calculate_line_65` method we get some level of accountability in the explain calculations section - which is useful
image

mpidcock: nice, that's convenient. seems good the way it is then

PR #3908, opened 2023-12-20
Polling acknowledgements

tofarr:


bengolder (commented):
Before adding 'Denied by IRS' to the set of responses that could put the submission into a "rejected" state, I'd want to be sure that this response is in the same spirit as others. My current understanding is that we can receive an explicit "Rejected" status that is distinct from this denial, and the rejection typically includes errors about why it was rejected. This looks like a different kind of response.

app/services/efile/poll_for_acknowledgments_service.rb
```diff
@@ -89,10 +89,17 @@ def self._handle_submission_status_response(response)
...
       doc.css('StatusRecordGrp').each do |status_record_group|
-        if status_record_group.css('SubmissionStatusTxt').text == 'Acknowledgement Received from State'
-          status_updates += 1
-          submission = EfileSubmission.find_by(irs_submission_id: status_record_group.css('SubmissionId').text)
+        status = status_record_group.css('SubmissionStatusTxt').text
+        status_updates += 1
+        submission = EfileSubmission.find_by(irs_submission_id: status_record_group.css('SubmissionId').text)
+        if status == 'Received'
           submission.transition_to(:ready_for_ack, raw_response: status_record_group.to_xml)
+        elsif status == 'Denied by IRS'
```
rickreyhsig, line 97: We should keep track of the error status responses we get in a constant somewhere.

PR #3910, opened 2023-12-21
Page should not error out if xml element is missing

tofarr:
We have a case where the XML page simply fails because the XML element `IRS1040Schedule1 UnemploymentCompAmt` is optional and missing. This change adds logic that we only set the value if it has actually changed


embarnard:
is this the show XML page? I guess I'm concerned if the page doesn't error out we won't know what to fix

embarnard:
> is this the show XML page? I guess I'm concerned if the page doesn't error out we won't know what to fix

oh never mind I was thinking about another page

embarnard (approved):

rickreyhsig:
For a little more context on this PR: https://cfastaff.slack.com/archives/C0544ERAFQV/p1703284923076919

PR #3912, opened 2023-12-21
Style updates for the taxes-owed page

tofarr:
Making the view look like the mocks in figma - we now have a blue background. (Also there was a copy update)
image


rickreyhsig (approved):
LGTM

PR #3916, opened 2023-12-21
Arizona now has a flat tax rate of 2.5%

tofarr:
The change is quite simple (Line 46):
image
I also added additional specs to cover the fact that the tax rate is now a flat 2.5%


bengolder (approved):
Approved! I really appreciate the new unit test as well!

PR #3919, opened 2023-12-21
Update AZ standard deduction values

mpidcock:
I looked into adding AZ standard deductions to the existing `standard_deductions.yml`, but since that is currently attached only to the model for federal returns, I decided it was out of scope for a minor bug fix.


tofarr (approved):

tofarr:
Seems to check out
image

PR #3920, opened 2023-12-22
Implement age #186716147

tofarr:
Implemented the dependents, added specs, and tested with the Johnny Rose case:
image

### DIFF
--- a/app/models/state_file_dependent.rb
+++ b/app/models/state_file_dependent.rb
@@ -60,8 +60,8 @@ class StateFileDependent < ApplicationRecord
 
   def age
     # Below is age as of today
-    #((DateTime.current.beginning_of_day - dob.to_time) / 1.year.seconds).floor
+    ((Time.now.beginning_of_day - dob.to_time) / 1.year.seconds).floor
     # Below is age as of the end of the year
-    DateTime.current.year - dob.time.year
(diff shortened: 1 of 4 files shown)


app/models/state_file_dependent.rb
```diff
@@ -57,4 +57,11 @@ def ask_senior_questions?
   def self.senior_cutoff_date
     MultiTenantService.statefile.end_of_current_tax_year.years_ago(65)
   end
+
+  def age
+    # Below is age as of today
+    ((Time.now.beginning_of_day - dob.to_time) / 1.year.seconds).floor
+    # Below is age as of the end of the year
+    # DateTime.current.year - dob.time.year
```
mpidcock, line 65: Based on [this slack thread](https://cfastaff.slack.com/archives/C0544ERAFQV/p1703198025859729) I was expecting this math to use `end_of_current_tax_year` instead of the current day.

tofarr: You're right - Updated

mpidcock: Nice! The test cases should be updated too - looks like there are some failures

mpidcock (commented):
Looks like the tests should also be updated

PR #3923, opened 2023-12-22
Now filtering unqualified parent ancestors out of dependents to prevent duplicates

tofarr:
This prevents people from section 11 from also appearing in section 10
image
Note: Sections 10 and 11 continue on the next page of the PDF

### DIFF
--- a/app/lib/submission_builder/ty2022/states/az/individual_return.rb
+++ b/app/lib/submission_builder/ty2022/states/az/individual_return.rb
@@ -56,7 +56,7 @@ module SubmissionBuilder
               end # TODO fix after we figure out dependent information
               xml.SupplementPageAttached 'X' # TODO Check box if theres not enough space on the first page for dependents
               xml.Dependents do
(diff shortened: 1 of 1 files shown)


app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -95,6 +95,7 @@ def documents_wrapper
                     end
                   end
                 end
+                # ERROR HERE
```
mpidcock, line 98: What does this comment mean?

mpidcock (approved):
Looks good, but consider removing or explaining the comment more

app/lib/submission_builder/ty2022/states/az/individual_return.rb
tofarr: Oops - I didn't mean to add that - I scribbled it as I was tracing through the code to find the source of the bug - removed

PR #3927, opened 2023-12-22
When county changes, we make sure the school district is valid

tofarr:
Suppose I have the following settings...
image

And I click edit on Edit for County, and choose Manhatten:
image

When I click continue, I am then taken to the school district selection page, and the districts shown are those from Manhatten
image

After I click continue, I am then taken back to data review as previously.
image

If I don't actually change the County, I am taken directly back to the data review page.


spec/controllers/state_file/questions/ny_county_controller_spec.rb
```diff
@@ -31,6 +31,30 @@
...
+        it "navigates to the state review screen" do
+          post :update, params: form_params.merge({return_to_review: "y"})
+          expect(response).to redirect_to(controller: "ny_school_district", action: :edit, us_state: "ny", return_to_review: 'y')
+        end
+      end
+
+      context "without return_to_review_param set" do
+        it "navigates to the next page in the flow" do
+          post :update, params: form_params
+          controllers = Navigation::StateFileNyQuestionNavigation::FLOW.to_a
```
rickreyhsig, line 45: Do we really need a .to_a here? Seems like  `Navigation::StateFileNyQuestionNavigation::FLOW` already returns an array.

tofarr: Good catch - removed

rickreyhsig (approved):
Nice work! The comment about return_to_review param might be helpful to still be there

PR #3931, opened 2023-12-22
Consume AdditionalChildTaxCreditAmt from the correct xml location

mpidcock:
The xml path for `AdditionalChildTaxCreditAmt` was incorrect, leading to incorrect calculations for the Empire State Child Credit. I updated it to match the location in all of the federal xml examples we have.


rickreyhsig (approved):
LGTM! https://cfastaff.slack.com/archives/C0544ERAFQV/p1703285413010989 has a little more context

PR #3932, opened 2023-12-26
Offboarding users when they have an out of state W2

tofarr:
When a user has a W2 from outside the current state in their direct file data, they get the following page:
image

To get tests working again after adding this additional constraint, I had to add a new test dataset that mimics an existing one, but has the state as AZ instead of NY

### DIFF
--- a/app/models/state_file_az_intake.rb
+++ b/app/models/state_file_az_intake.rb
@@ -135,10 +135,13 @@ class StateFileAzIntake < StateFileBaseIntake
   end
 
   def disqualifying_df_data_reason
-    :married_filing_separately if direct_file_data.filing_status == 3
-    direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd').each do |state|
-      return :has_out_of_state_w2 if state.text != 'AZ'
+    return :married_filing_separately if direct_file_data.filing_status == 3
+
+    w2_states = direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd')
+    :has_out_of_state_w2 if w2_states.any? do |state|
+      (state.text || '').upcase != 'AZ'
     end
+
   end
 
   def disqualifying_eligibility_rules
--- a/app/models/state_file_az_intake.rb
+++ b/app/models/state_file_az_intake.rb
@@ -136,6 +136,9 @@ class StateFileAzIntake < StateFileBaseIntake
 
   def disqualifying_df_data_reason
     :married_filing_separately if direct_file_data.filing_status == 3
+    direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd').each do |state|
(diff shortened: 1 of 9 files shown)


app/models/state_file_az_intake.rb
```diff
@@ -135,7 +135,13 @@ def ask_spouse_dob?
   end
 
   def disqualifying_df_data_reason
-    :married_filing_separately if direct_file_data.filing_status == 3
+    return :married_filing_separately if direct_file_data.filing_status == 3
+
+    w2_states = direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd')
+    :has_out_of_state_w2 if w2_states.any? do |state|
+      (state.text || '').upcase != 'AZ'
```
rickreyhsig, line 142:
```suggestion
      (state.text || '').upcase != state_code.upcase
```

app/models/state_file_ny_intake.rb
```diff
@@ -196,6 +196,11 @@ def ach_debit_transaction?
   IRC_125_CODES = ['IRC125S', 'IRS125']
   YONKERS_CODES = ['YK', 'YON', 'YNK', 'CITYOFYK', 'CTYOFYKR', 'CITYOF YK', 'CTY OF YK']
   def disqualifying_df_data_reason
+    w2_states = direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd')
+    return :has_out_of_state_w2 if w2_states.any? do |state|
+      (state.text || '').upcase != 'NY'
```
rickreyhsig, line 201:
```suggestion
      (state.text || '').upcase != state_code.upcase
```

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -165,7 +165,7 @@
       expect(page).to have_text I18n.t('state_file.questions.terms_and_conditions.edit.title')
       click_on I18n.t("state_file.questions.terms_and_conditions.edit.accept")
 
-      step_through_df_data_transfer
+      step_through_df_data_transfer("Transfer AZ Old sample")
```
rickreyhsig, line 168: Are you using the string for label purposes?

app/models/state_file_az_intake.rb
```diff
@@ -135,7 +135,13 @@ def ask_spouse_dob?
...
+    end
+
```
rickreyhsig, line 144: Nit: extra line

rickreyhsig (approved):
LGTM! Left a few suggestions.

spec/features/state_file/complete_intake_spec.rb
tofarr: The string is actually used to select the button to click in the test. This was hardcoded to 
```click_on "Transfer my 2023 federal tax return to FileYourStateTaxes"```
 so I added an optional parameter to allow clicking on one of the other elements instead. (See state_file_intake_helper.rb Line 70

app/models/state_file_az_intake.rb
tofarr: Removed

PR #3937, opened 2023-12-28
AZ bug: Dependents over 17 not being counted in AZ-140 10b

rickreyhsig:
https://www.pivotaltracker.com/story/show/186716200

### XML fix
Screen Shot 2023-12-28 at 1 31 26 PM

### PDF fix
Screen Shot 2023-12-28 at 1 31 34 PM

### DIFF
--- a/app/models/state_file_az_intake.rb
+++ b/app/models/state_file_az_intake.rb
@@ -111,11 +111,11 @@ class StateFileAzIntake < StateFileBaseIntake
   end
 
   def federal_dependent_count_over_17
-    self.dependents.select{ |dependent| dependent.age >= 17 }.length
+    self.dependents.select{ |dependent| dependent.age >= 17 && !dependent.ask_senior_questions? }.length
   end
 
   def qualifying_parents_and_grandparents
-    dependents.select(&:ask_senior_questions?).length
+    self.dependents.select(&:ask_senior_questions?).length
   end
 
   def ask_months_in_home?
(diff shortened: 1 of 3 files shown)


app/models/state_file_az_intake.rb
```diff
@@ -111,11 +111,11 @@ def federal_dependent_count_under_17
   end
 
   def federal_dependent_count_over_17
-    self.dependents.select{ |dependent| dependent.age >= 17 }.length
+    self.dependents.select{ |dependent| dependent.age >= 17 && !dependent.ask_senior_questions? }.length
   end
 
   def qualifying_parents_and_grandparents
-    dependents.select(&:ask_senior_questions?).length
+    self.dependents.select(&:ask_senior_questions?).length
```
rickreyhsig, line 118: Adding the self to keep consistent with other two

app/models/state_file_az_intake.rb
```diff
@@ -111,11 +111,11 @@ def federal_dependent_count_under_17
   end
 
   def federal_dependent_count_over_17
```
tofarr, line 113: I wonder should we rename this method / create a new, given that it now described a situation where the dependent is over 17 but not a senior?

rickreyhsig: Yep, good catch. Will rename!

rickreyhsig: Done

tofarr (approved):

PR #3939, opened 2023-12-29
Add mobile desktop menu #186639306

tofarr:
A whole bunch of CSS changes / fixes went in here. The menu links are not fully right yet, but match the mocks in figma.
The mobile menu for state file now has a button:
image
Menu Items are vertically centered in state file (They were 4px off due to different toolbar sizes for gyr):
image
There is a mobile menu with colors that match figma for both AZ and NY. It is now positioned as in the mock - also, the X is the correct color and does not overhang the menu:
image

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -5,13 +5,7 @@
 
 .state-file-html-bg--az {
   background-color: $color-az-grey;
-}
 
-.state-file-html-bg--ny {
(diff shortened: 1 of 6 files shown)


rickreyhsig (approved):
Needs adding some i18n keys, but otherwise LGTM.

https://app.circleci.com/pipelines/github/codeforamerica/vita-min/19313/workflows/6569ce3d-12d6-4d92-82be-a9aa56e25eeb/jobs/49663

PR #3942, opened 2023-12-29
Added 15 minute session timeout

tofarr:
After 15 minutes, you get this error when you try to proceed to a next step:
image


app/models/state_file_az_intake.rb
```diff
@@ -77,6 +77,8 @@ class StateFileAzIntake < StateFileBaseIntake
   enum eligibility_married_filing_separately: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_married_filing_separately
   enum eligibility_529_for_non_qual_expense: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_529_for_non_qual_expense
 
+  devise :timeoutable, :timeout_in => 15.minutes
```
rickreyhsig, line 80: Could we set this config on the base intake instead?

tofarr: Good call

rickreyhsig (approved):

PR #3943, opened 2024-01-02
Work around for not hiding field

tofarr:
Honeycrisp seems to have a bug when hiding fields nested fields - for example, the "Sales Use Tax" field. e.g.:

1. Click the "I will calculate my use tax manually" radio button.
 => the "Enter calculated use tax" field appears as expected.
image

2. Click the "All of your purchases were less than $1,000 each." radio button
 => The "Enter calculated use tax" field does not disappear.
image

The reason is honeycrisp.js line 111:
image

The jquery selector `$(self).find('> .question-with-follow-up__follow-up')` does not find the element to hide.

Funnily enough, the selector `$(self).find('>').find('.question-with-follow-up__follow-up')` does actually work, so I can only conclude that this is a bug in jquery (Given that `self.parentElement.querySelector('.question-with-follow-up__follow-up')` also works, but `>` is not a valid css selector outside jquery):

image


bengolder (approved):
Approved! I'm okay with a workaround for now, but I think we should implement this as a new component.

PR #3945, opened 2024-01-03
Payment voucher download on Accepted/taxes owed screen in AZ + NY

rickreyhsig:
https://www.pivotaltracker.com/story/show/186713503

nyVoucher


app/controllers/state_file/questions/return_status_controller.rb
```diff
@@ -44,6 +48,55 @@ def refund_url
         end
       end
 
+      def tax_payment_url
```
embarnard, line 51: Sometimes we give different urls based on locale, worried about if we add spanish later if having this pattern will make it harder. Not sure if that will be the case for these links this time so this comment is non-blocking but just something to keep in mind

config/locales/en.yml
```diff
@@ -2230,19 +2230,24 @@ en:
...
             '
+          download_voucher: Download payment voucher form
+          include_payment: You'll need to include the payment voucher form.
+          mail_voucher_and_payment: 'Mail the voucher form and payment to:'
+          pay_by_mail_or_moneyorder: 'If you are paying by mail via check or money order:'
           register_to_vote_html: '<a target="_blank" rel="noopener nofollow" href="https://vote.gov/">Register to vote.</a>
 
             '
           title: Your 2023 %{state_name} state tax return is accepted
+          track_your_payment_html: 'Track your payment <a target="_blank" rel="noopener nofollow" href="">here.</a>
```
embarnard, line 2248: is this link's href supposed to be blank?

embarnard (approved):
Looks good! Left non-blocking comments

app/controllers/state_file/questions/return_status_controller.rb
rickreyhsig: Interesting point.  Do you have an example of handling this based on locale?

config/locales/en.yml
rickreyhsig: We don't have the url yet.

rickreyhsig: https://www.pivotaltracker.com/story/show/186713503/comments/239721160

PR #3948, opened 2024-01-03
Fix county parsing in csv

mpidcock:
I am thinking that adding a test purely for csv parsing is too much, but lmk if someone disagrees. This commit has two fixes - one to remove a specific typo, and one to make the code more durable if the typo gets introduced again. I tested both manually individually and the error was resolved by both.


bengolder (approved):
Looks great! Thank you, approved!

PR #3952, opened 2024-01-03
Transmitted and Ready for Ack are now treated separately

tofarr:


app/services/efile/poll_for_acknowledgments_service.rb
```diff
@@ -36,6 +36,12 @@ def self.poll(submission_ids, endpoint)
       DatadogApi.increment("efile.poll_for_#{dd_tag}")
     end
 
+    def self.ready_for_ack_submission_ids
+      transmitted_submissions = EfileSubmission.in_state(:ready_for_ack)
+      transmitted_submissions.touch_all(:last_checked_for_ack_at)
```
embarnard, line 41: nice adding this here! I think we should remove it from line 56 then, I think I misinterpreted acknowledgments vs transmitted statuses earlier and added it there a few weeks ago

embarnard (approved):
Looks good! one non-blocking comment

PR #3954, opened 2024-01-04
Fix for issue where data offboarding takes you back to login

tofarr:
Initiate the NY flow:
image
Upload an AZ return:
image
Before the fix, you are taken to the login page.
This is because data_transfer_offboarding_controller  extended AuthenticatedQuestionsController and require_state_file_intake_login was called before the action - and since we are in offboarding both current_state_file_az_intake and current_state_file_ny_intake are blank.
image


bengolder (approved):
Good catch! Approved!

PR #3955, opened 2024-01-04
Filtering incarcerated question from review page #186728319

tofarr:
Using the existing rules (Moved to model from controller) to determine whether to show / hide the incarcerated section in review (Shown below hidden):
image


bengolder (approved):
Great job! I appreciate the specs and refactoring!

PR #3957, opened 2024-01-04
Only Displaying Months in Home if Present #186728244

tofarr:
Months in home is now only displayed in the review page if it is present in the record (Shown filtered below). This view is shared between AZ and NY, and checking for presence means we don't need to change that:
image


bengolder (approved):
Looks good! Thanks!

PR #3958, opened 2024-01-05
Move efile submissions that get "Denied by IRS" status into "read_for_ack" status

embarnard:


embarnard:
I noticed we were actually receiving acknowledgments for those "Denied by IRS" status efile submissions. So we should we move those into a "ready_for_ack" state and then it will poll for ack.

tofarr (approved):

PR #3960, opened 2024-01-05
Remove name from address fields in the Ivy Hoh NY XML

rickreyhsig:
The name `co Amanda Jones` is erroneusly added on the XML.  Tiffany has requested that I remove it.


bengolder (approved):
Approved!

PR #3964, opened 2024-01-08
Polling updates

tofarr:
This resolves a few things:
* The status for returned submissions was not being set correctly (Prematurely in :ready_for_ack)
* Multiple status are being returned for each sing single submission - we now process only the first (Most recent) one. e.g.: (Check out the spec that includes the XML sample - this is actually something we got back in demo)

### DIFF
--- a/app/services/efile/poll_for_acknowledgments_service.rb
+++ b/app/services/efile/poll_for_acknowledgments_service.rb
@@ -128,9 +128,6 @@ module Efile
         # no action required - the IRS are still working on it
         :transmitted
       elsif READY_FOR_ACK_STATUSES.include?(status)
-        unless status == "Acknowledgement Received from State"
-          Sentry.capture_message("Retrieved status for submission #{submission.id} that should already be in ready_for_ack state")
-        end
         :ready_for_ack
       else
         :failed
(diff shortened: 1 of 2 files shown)


app/services/efile/poll_for_acknowledgments_service.rb
```diff
@@ -93,32 +105,39 @@ def self._handle_ack_response(response)
...
-          unless status == "Acknowledgement Received from State"
-            Sentry.capture_message("Retrieved status for submission #{submission.id} that should already be in ready_for_ack state")
-          end
-          submission.transition_to(:ready_for_ack, raw_response: raw_response)
-        else
-          submission.transition_to(:failed, raw_response: raw_response)
+      status_updates
+    end
+
+    def self.group_status_records_by_submission_id(doc)
```
embarnard, line 121: nice!

spec/services/efile/poll_for_acknowledgments_service_spec.rb
```diff
@@ -279,5 +279,54 @@
         expect(described_class.transmitted_state_submission_ids).to match_array([irs_submission_id3, irs_submission_id4])
       end
     end
+
+    context "getting status from state" do
```
embarnard, line 283: thank you for adding this test!

spec/services/efile/poll_for_acknowledgments_service_spec.rb
```diff
@@ -279,5 +279,54 @@
...
+      it "interprets ready_for_ack successfully" do
+        ["Acknowledgement Received from State", "Acknowledgement Retrieved", "Notified"].each do |status|
```
embarnard, line 285: should add `Denied by IRS` in here too

embarnard (approved):
Looks good!

PR #3973, opened 2024-01-09
Added spec verifying timeout #186582387

tofarr:


spec/models/state_file_base_intake_spec.rb
```diff
@@ -52,5 +52,12 @@
...
       end
     end
+
+    context "when timed out" do
+      let!(:intake) { create :state_file_az_intake }
+      it "actually times out" do
+        expect(intake.timedout?(15.minutes.ago)).to eq true
+      end
+    end
   end
```
bengolder, line 62: I'm confused. What this test seems to assert is that if we make a fresh intake, it starts off as having timed out 15 minutes ago. This may be true and may be useful information, but it doesn't seem to cover the most common/expected scenario: how an authenticated intake gets timed out.

PR #3974, opened 2024-01-09
Efile submission transitions logs reformatting in hub

embarnard:
Consolidating the display of the transmitted states so that its easier to read in the hub
Screenshot 2024-01-09 at 12 38 05 PM


embarnard:
got review from tim

PR #3976, opened 2024-01-10
Data audit: IT-201 Section M, add tests

jenny-heath:


spec/lib/efile/ny/it201_spec.rb
```diff
@@ -240,6 +240,25 @@
...
   end
 
+  describe "#calculate_line_61" do
+    before do
+      intake.update(sales_use_tax: 50)
+    end
+
+    it "adds up some of the prior lines" do
+      instance.calculate
+      sum = instance.lines[:IT201_LINE_46].value + instance.lines[:IT201_LINE_58].value + instance.lines[:IT201_LINE_59].value
```
jenny-heath, line 250: an alternative to this is to stub the methods `calculate_line_46` and `calculate_line_58`. not sure which is better

bengolder (approved):
Approved!

PR #3981, opened 2024-01-10
Remove efile: prefix, add show xml to efile show page, parse acknowledgement error code and add "R" rejected status

embarnard:


embarnard:
Add show XML button to efile submission page so we can more easily debug and parse out the error message from acknowledgments and add "R" status to list of rejected codes
Screenshot 2024-01-10 at 3 13 32 PM

jenny-heath (approved):
looks good!

PR #3986, opened 2024-01-10
add production certs and private keys to prod credentials

mrotondo:


mpidcock (approved):
I was able to compare to demo and looks good to me, thanks Mike!

PR #3987, opened 2024-01-10
Add w2 to submission xml #186819099

tofarr:
Adding a W2 to the XML for the state file submission - it basically just copies the W2 from the StateFileAzIntake / StateFileNyIntake


app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -189,23 +189,25 @@ def supported_documents
...
-                xml: nil,
-                pdf: PdfFiller::NyIt2Pdf,
+                xml: SubmissionBuilder::Ty2022::States::Ny::Documents::State1099G,
+                pdf: nil,
                 include: true,
-                kwargs: { w2: w2 }
+                kwargs: { form1099g: form1099g }
               }
             end
+
```
rickreyhsig, line 210: We could add an expectation on the spec/lib/submission_builder/ty2022/states/ny/individual_return_spec.rb and the arizona one that these are present.

tahsinaislam: just a question out of curiosity: why was the order of the 1099 xml and it2 pdf switched for ny?

tofarr: This is not strictly speaking necessary - I wanted to maintain the same order in the XML as is defined in the XSD (FormsBO.xsd). It makes the data easier to correlate.

rickreyhsig (approved):
LGTM

PR #3989, opened 2024-01-11
Transform relationship labels

bengolder:
- use a standard gender neutral label for all pdf filling
- for NY XML, use the direct file key without spaces
- for AZ XML, use a specific lookup table to translate

### DIFF
--- a/app/lib/pdf_filler/az140_pdf.rb
+++ b/app/lib/pdf_filler/az140_pdf.rb
@@ -42,18 +42,17 @@ module PdfFiller
       end
 
       answers["10a_10b check box"] = 'Yes' if @xml_document.css('DependentsDetail').length > 3
-
-      @xml_document.css('DependentDetails').each_with_index do |dependents_node, index|
+      @submission.data_source.dependents.reject(&:is_qualifying_parent_or_grandparent?).each_with_index do |dependent, index|
         # PDF fields seem to be named consistently (10c ... 10p) whether they are on Page 1 or Page 4
         prefix = "10#{('c'..'p').to_a[index]}"
         answers.merge!(
-          "#{prefix} First" => dependents_node.at("FirstName")&.text,
-          "#{prefix} Last" => dependents_node.at("LastName")&.text,
-          "#{prefix} SSN" => dependents_node.at("DependentSSN")&.text,
-          "#{prefix} Relationship" => dependents_node.at("RelationShip")&.text,
-          "#{prefix} Mo in Home" => dependents_node.at("NumMonthsLived")&.text,
-          "#{prefix}_10a check box" => dependents_node.at("DepUnder17")&.text,
-          "#{prefix}_10b check box" => dependents_node.at("Dep17AndOlder")&.text,
+          "#{prefix} First" => dependent.first_name,
+          "#{prefix} Last" => dependent.last_name,
+          "#{prefix} SSN" => dependent.ssn.delete('-'),
+          "#{prefix} Relationship" => dependent.relationship_label,
+          "#{prefix} Mo in Home" => dependent.months_in_home,
+          "#{prefix}_10a check box" => dependent.under_17? ? "X" : nil,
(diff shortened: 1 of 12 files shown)


app/lib/pdf_filler/az140_pdf.rb
bengolder, line 1: For the PDF, I needed to pull dependents from the database instead of the XML so that I could use the `relationship_label` method. For selecting which dependents to iterate over in each section, I copied the logic from the XML for consistency.

app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -4,12 +4,15 @@ module Ty2022
     module States
       module Az
         class IndividualReturn < SubmissionBuilder::Document
-          FILING_STATUSES = {
-            single: 'Single',
-            married_filing_jointly: 'MarriedJoint',
-            married_filing_separately: 'MarriedFilingSeparateReturn',
-            head_of_household: 'HeadHousehold',
-          }.freeze
```
bengolder, line 12: This was unused, so I deleted it

app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -4,12 +4,15 @@ module Ty2022
...
+          include DependentRelationshipTable
```
bengolder, line 7: I put the look up hash into a separate module so that this file wouldn't get so cluttered.

app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -4,12 +4,15 @@ module Ty2022
...
+
+          FILING_STATUS_OPTIONS = {
+            :married_filing_jointly => 'MarriedJoint',
+            :head_of_household => 'HeadHousehold',
+            :married_filing_separately => 'MarriedFilingSeparateReturn',
+            :single => "Single"
+          }
```
bengolder, line 14: This was already in the file, much lower down. I moved it up next to the other constants.

app/models/state_file_dependent.rb
bengolder, line 1: The labels in here are used for all PDFs. I figured that we may want to use these in places other than the PDFs so I included it with the model.

app/models/state_file_dependent.rb
```diff
@@ -47,18 +64,26 @@ def full_name
...
+  end
+
+  def senior?
+    age >= 65
   end
 
   def age
-    ((MultiTenantService.statefile.end_of_current_tax_year.to_time - dob.to_time) / 1.year.seconds).floor
+    MultiTenantService.statefile.current_tax_year - dob.year
+  end
```
bengolder, line 84: I had to change this because it was calculating slightly different ages on servers vs locally. Calculating age using seconds can cause some bugs because years are not the same quantity of seconds. Some years have more seconds than others. Additionally, calling `to_time` can add a timezone to a date that can adjust it to be earlier or later than the original date.

app/views/state_file/questions/ny_permanent_address/edit.html.erb
```diff
@@ -15,7 +15,7 @@
           <div class="form-question spacing-below-15">
             <div class="text--bold"><%=t(".address_header") %></div>
             <div><%= current_intake.direct_file_data.mailing_street %></div>
-            <div><%= current_intake.direct_file_data.mailing_city %>,<%= current_intake.direct_file_data.mailing_state %> <%= current_intake.direct_file_data.mailing_zip %></div>
+            <div><%= current_intake.direct_file_data.mailing_city %>, <%= current_intake.direct_file_data.mailing_state %> <%= current_intake.direct_file_data.mailing_zip %></div>
```
bengolder, line 18: This is a tiny unrelated fix for how we display addresses

tofarr (approved):

PR #3990, opened 2024-01-11
Forms xsd audit

tofarr:
I moved the action that copies the address from the intake from `before_save` to `before_validation` - this means that the address will always be present and can be validated. I added validations for the address

### DIFF
--- a/app/lib/submission_builder/ty2022/states/az/documents/state1099_g.rb
+++ b/app/lib/submission_builder/ty2022/states/az/documents/state1099_g.rb
@@ -30,7 +30,7 @@ module SubmissionBuilder
                 xml.RecipientSSN recipient.ssn
                 xml.RecipientName recipient.full_name
                 xml.RecipientUSAddress do
-                  xml.AddressLine1Txt form1099g.recipient_street_address
+                  xml.AddressLine1Txt form1099g.form1099g.recipient_address_line1
                   xml.CityNm form1099g.recipient_city
                   xml.StateAbbreviationCd "AZ"
                   xml.ZIPCd form1099g.recipient_zip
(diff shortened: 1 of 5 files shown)


app/models/state_file1099_g.rb
```diff
@@ -30,7 +30,7 @@
 #
 class StateFile1099G < ApplicationRecord
   belongs_to :intake, polymorphic: true
-  before_save :update_conditional_attributes
+  before_validation :update_conditional_attributes
```
rickreyhsig, line 33: Was the intent here so that we it will pre-validate presence of the 3 attributes before updating trying to conditional_attributes? Do we need to validate mailing_apartment?

tofarr: The intent here is that the update to the conditional attributes happens before validation instead of after it but before save. This way, the conditional attributes get set, and then validation happens, so the values are valid. (Instead of validation happening and then the conditional attributes getting set, resulting in an invalid model)

rickreyhsig (approved):

app/lib/submission_builder/ty2022/states/az/documents/state1099_g.rb
```diff
@@ -30,7 +30,7 @@ def document
                 xml.RecipientSSN recipient.ssn
                 xml.RecipientName recipient.full_name
                 xml.RecipientUSAddress do
-                  xml.AddressLine1Txt form1099g.recipient_street_address
+                  xml.AddressLine1Txt form1099g.recipient_address_line1
```
bengolder, line 33: @tofarr Should the apartment go in a `AddressLine2Txt` element instead of being concatenated into line 1?

PR #3991, opened 2024-01-11
Update county and school district handling

bengolder:
As part of https://www.pivotaltracker.com/story/show/186685604
Also based off of this spreadsheet: https://docs.google.com/spreadsheets/d/11265aXVY_SdtxL-Jvq5Mbz8wXpnn_dk8NRnj4nZIZWs/edit#gid=56211979

**Goals of this PR**
- Use the correct short county name when writing to the PDF or XML (for example "New York" instead of "Manhattan (see New York)"
- Add the correct four-letter county code in the PDF & XML
- Refactor where possible to improve the code

**Key changes**
- Updated the ny/school_districts.csv so that we could have a unique id for each district along with clear correct county names.
- Added a column to the intake for the school district id
- Refactored the corresponding NySchoolDistricts class for easier faster lookups, and to ensure it's lazy loaded only once into memory for a given process.
- Refactored the associated controllers, forms, and their specs for county & district
- Updates the XML & PDF code

**Possible followup: remove school_district column?**
Now that it's saving `school_district_id` on the intake, we arguably don't need the `school_district` column. On demo, all the old intakes will have `school_district` but not `school_district_id`. To be minimally disruptive, I could migrate the data over on demo, and then remove the `school_district` column in a small followup PR.


tofarr (approved):
LGTM

PR #3993, opened 2024-01-11
Fix FAQ bug where showing same page for same slug and different states

embarnard:
There is a problem right now where the New York "[Who can use FileYourStateTaxes?](https://demo.fileyourstatetaxes.org/en/ny/faq/who_can_use_this_tool)" page is showing Arizona text, this PR fixes this bug

before: 
Screenshot 2024-01-11 at 8 19 52 PM

after:
Screenshot 2024-01-11 at 8 20 39 PM


app/models/faq_category.rb
```diff
@@ -41,4 +41,13 @@ def description(locale)
       description_es.present? ? description_es : description_en
     end
   end
+
+  def self.state_to_product_type(state)
```
tofarr, line 45: We should probably add a spec for this

tofarr (approved):

PR #3994, opened 2024-01-11
State file: update 2023 federal schema to version 5.0

jenny-heath:


bengolder (approved):
Thanks!

jenny-heath:
@bengolder this wasn't quite ready when you approved it! (still had some failing tests and hadn't uploaded the new schema to S3). i hope it's ready now, please give it another look if you have time!

bengolder (approved):
Approved again!

bengolder:
@jenny-heath I noticed that the fed return sample schemas were updated. Should we be concerned about existing intakes on demo that used the old schema? Would it break any of those to merge this? Is direct file now sending us XML that matches the new schema?

bengolder:
In regards to my previous questions:
> Should we be concerned about existing intakes on demo that used the old schema? Would it break any of those to merge this? Is direct file now sending us XML that matches the new schema?

@embarnard answered in Slack that we should not be concerned about breaking existing intakes that were based on the old schema. This is likely safe to merge. It remains unclear if direct file is yet sending us XML that matches the new schema.

Merging

PR #3995, opened 2024-01-11
Fix section 3 check boxes

mpidcock:
There were a few boxes not showing up that are there now.
image
image


bengolder (approved):
Excellent! I appreciate the clarifying comments, thank you!

PR #3997, opened 2024-01-12
Form updates

tofarr:


app/forms/state_file/ny_primary_state_id_form.rb
```diff
@@ -16,7 +16,7 @@ class NyPrimaryStateIdForm < StateIdForm
                        :expiration_date_month,
                        :expiration_date_year
 
-    validates :first_three_doc_num, alphanumeric: true, length: {is: 3}, unless: -> { id_type == "no_id" }
+    validates :first_three_doc_num, alphanumeric: true, length: {is: 3}, if: -> { state == "NY" && id_type != "no_id" }
```
tofarr, line 19: Only New York drivers licenses are fully validated

app/views/state_file/questions/ny_primary_state_id/_state_id.html.erb
```diff
@@ -40,10 +40,12 @@
           <span id="checkbox">
             <%= f.cfa_checkbox(:non_expiring, t("state_file.questions.primary_state_id.state_id.id_details.no_expiration_date")) %>
           </span>
-          <%= f.cfa_select(:state, t("state_file.questions.primary_state_id.state_id.id_details.issue_state"), States.name_value_pairs, include_blank: true, classes: ["form-width--short, spacing-above-0"]) %>
-          <% if options[:include_first_three_doc_num] == true %>
+          <div onchange="$('#first_3_doc_num_container')[event.target.value == 'NY' ? 'show' : 'hide']()">
```
tofarr, line 43: The requirement is that the 3 doc number only shows if the license selection is New York. We don't currently have a pre-built component for showing / hiding a section based on a select, so some custom JS was needed.

app/views/state_file/questions/ny_primary_state_id/_state_id.html.erb
bengolder, line 1: If I enter my answer as NY with the 3 doc number and then change it to out of state, will it clear my 3 doc number or keep it?

bengolder (approved):
Non-blocking suggestion: seems like the different ID forms could be refactored to use more shared code. At a glance they seem to have a bunch of overlap.

Thanks!

PR #3998, opened 2024-01-12
barcodes in NY form PDFs

mrotondo:


app/helpers/pdf_helper.rb
```diff
@@ -47,7 +50,65 @@ def output_file
...
+
+  def nys_10digit_barcode_value(form_type, page_num)
+    vendor_source_code = "1963"
+    last_two_digits_of_tax_year = tax_year.to_s[-2..-1]
+    "#{form_type}#{page_num}#{last_two_digits_of_tax_year}#{vendor_source_code}"
+  end
+
+  def nys_12digit_barcode_value(form_type, page_num)
+    three_digit_page_num = "%03d" % page_num
+    vendor_source_code = "1963"
```
mpidcock, line 104: Nit: this is also declared on line 97, consider moving to `Ny201VPdf` (or `MultiTenantService` if it has broader reusability)

app/lib/pdf_filler/ny201_v_pdf.rb
```diff
@@ -0,0 +1,26 @@
...
+      "it201v_1223"
+    end
+
+    def nys_form_type
+      "040"
+    end
+
+    def tax_year
+      2023
+    end
```
mpidcock, line 15: Nit: wondering if we can avoid having this declared here? I know there's also `MultiTenantService.statefile.current_tax_year`, which is used as a parameter to initialize the It201. However, I don't want to get stuck in a rabbit hole about this, and it's perfectly fine to update this later, so if it makes the most sense right now to keep this here I won't question it.

public/pdfs/it201v_1223.pdf
mpidcock, line 1: Could there still be value in keeping this pdf for reference?

app/lib/pdf_filler/ny214_pdf.rb
mpidcock, line 1: Thanks for including this for consistency

mpidcock (approved):
LGTM and I'm happy to let my nits be left behind for some future refactor

app/helpers/pdf_helper.rb
mrotondo: Moved to Rails config

app/lib/pdf_filler/ny201_v_pdf.rb
mrotondo: Forgot all about that - switched over to using it instead

public/pdfs/it201v_1223.pdf
mrotondo: It's still in the repo, just moved out from assets to app/lib since we serve it dynamically now instead of statically

PR #3999, opened 2024-01-12
Fill out more IT-215 dependent details

bengolder:
Addresses https://www.pivotaltracker.com/story/show/186733283

Update IT-215 pdf dependent info to include
- eic disability info
- months lived in home


embarnard (approved):
Looks good!

PR #4001, opened 2024-01-12
Data transfer errors out on first attempt

rickreyhsig:
https://www.pivotaltracker.com/n/projects/2650644/stories/186828012

Screen Shot 2024-01-12 at 11 34 49 AM

### Description
This code fix prevents intakes from previous sessions to time out the devise's timeoutable feature though warden.  However, it will still error if you two simultaneous sessions from either state running along side each other.

Thanks @bengolder & @tofarr for your input!

### DIFF
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -101,12 +101,6 @@ module Questions
     end
 
     def redirect_if_no_intake
-      # session.delete(:state_file_intake)
-      # current_intake.timedout?(14.minutes.ago)
-      # session[:state_file_intake]
-      # current_intake
-      # binding.pry
-
       if current_intake.present?
         # Assign the global id of the state_file_intake of the session to be that of the current_intake
         # This attempts to prevent the weird timeout issues we have been experiencing in the flow.
--- a/app/controllers/questions/questions_controller.rb
+++ b/app/controllers/questions/questions_controller.rb
@@ -2,7 +2,6 @@ module Questions
   class QuestionsController < ApplicationController
     before_action :redirect_in_offseason
     before_action :redirect_if_completed_intake_present
-    before_action :redirect_if_no_intake
     before_action :set_current_step, :set_in_intake_flow
     delegate :form_name, to: :class
     delegate :form_class, to: :class
@@ -97,26 +96,6 @@ module Questions
     def redirect_if_completed_intake_present
       if current_intake && current_intake.completed_at.present?
         redirect_to portal_root_path
-      end
-    end
-
-    def redirect_if_no_intake
-      if current_intake.present?
-        # Assign the global id of the state_file_intake of the session to be that of the current_intake
-        # This attempts to prevent the weird timeout issues we have been experiencing in the flow.
-        # Where the session's intake does match the true current one and ends up timing out erroneously.
-        session[:state_file_intake] = "gid://vita-min/#{current_intake.class}/#{current_intake.id}"
-        # Sign out from previous intake if they differ from current_intake
-        # TODO
-      else
-        begin
-          visitor_id = cookies['visitor_id']
-          raise "The session for visitor with id:#{visitor_id} has expired"
-        rescue => e
-          Sentry.capture_exception(e)
-        end
-        flash[:notice] = 'Your session expired. Please sign in again to continue.'
-        redirect_to root_path
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/controllers/questions/questions_controller.rb
```diff
@@ -99,6 +100,26 @@ def redirect_if_completed_intake_present
...
     end
 
+    def redirect_if_no_intake
+      if current_intake.present?
+        # Assign the global id of the state_file_intake of the session to be that of the current_intake
+        # This attempts to prevent the weird timeout issues we have been experiencing in the flow.
+        # Where the session's intake does match the true current one and ends up timing out erroneously.
+        session[:state_file_intake] = "gid://vita-min/#{current_intake.class}/#{current_intake.id}"
+        # Sign out from previous intake if they differ from current_intake
+        # TODO
```
bengolder, line 110: I don't believe this would change anything and I'm concerned about including it. `current_intake` essentially checks if the value in the session matches the value on line 108. Line 108 sets the session value to the same value that has just been confirmed by line 105.

bengolder (changes requested):
Left a few comments and I'd want to be sure they were addressed before merging.

app/controllers/questions/questions_controller.rb
rickreyhsig: This logic makes sure that the session intake matches the current_intake, because as you suggested the intake on the session may be a different one that is bound to time out or already has.

app/controllers/questions/questions_controller.rb
bengolder, line 1: We should move these edits to the state file questions controller instead of this questions controller (which is used by GYR & CTC) in order to make sure it only applies to state file.

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -33,6 +34,26 @@ def question_navigator
...
+        if current_intake.present?
+          # Assign the global id of the state_file_intake of the session to be that of the current_intake
+          # This attempts to prevent the weird timeout issues we have been experiencing in the flow.
+          # Where the session's intake does match the true current one and ends up timing out erroneously.
+          session[:state_file_intake] = "gid://vita-min/#{current_intake.class}/#{current_intake.id}"
+          # Sign out from previous intake if they differ from current_intake
+          # TODO
+        else
+          begin
+            visitor_id = cookies['visitor_id']
```
rickreyhsig, line 47: We can't retrieve from cookies because this is written encrypted in cookies...
https://github.com/codeforamerica/vita-min/blob/main/app/controllers/concerns/state_file/start_intake_concern.rb#L13
```
    def current_intake
      @intake ||= question_navigator.intake_class.new(
        visitor_id: cookies.encrypted[:visitor_id],
        source: session[:source],
        referrer: session[:referrer]
      )
    end
```

app/controllers/questions/questions_controller.rb
```diff
@@ -99,6 +100,19 @@ def redirect_if_completed_intake_present
       end
     end
 
+    def redirect_if_no_intake
+      unless current_intake.present?
+        begin
+          visitor_id = question_navigator.intake_class.last.visitor_id
+          raise "The session for visitor with id:#{visitor_id} has expired"
```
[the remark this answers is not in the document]
rickreyhsig: No longer relevant

app/controllers/questions/questions_controller.rb
rickreyhsig: This code has been deleted

app/controllers/state_file/questions/questions_controller.rb
rickreyhsig: I nuked this code

app/controllers/state_file/questions/landing_page_controller.rb
```diff
@@ -2,6 +2,11 @@ module StateFile
   module Questions
     class LandingPageController < QuestionsController
       include StartIntakeConcern
+      before_action :before_update, only: :update
+
+      def before_update
+        sign_out current_intake if current_intake
+      end
```
bengolder, line 9:
```ruby

    class LandingPageController < QuestionsController
      include StartIntakeConcern
      before_action :sign_out_old_intakes, only: :update

      def sign_out_old_intakes
        sign_out current_state_file_ny_intake if current_state_file_ny_intake.present?
        sign_out current_state_file_az_intake if current_state_file_az_intake.present?
      end
      
    end
```

bengolder: also for sign in and clearing session:
https://github.com/codeforamerica/vita-min/blob/0fb5189d03a1e0c5ec144173057a1a8656661d31/app/controllers/application_controller.rb#L70-L79

PR #4008, opened 2024-01-16
NY State Child Credit

tahsinaislam:


tofarr (approved):

app/lib/efile/ny/it213.rb
```diff
@@ -131,8 +133,21 @@ def calculate_line_3
...
+      def calculate_line_5
+        count = 0
+        @intake.dependents.each do |dependent|
+          if dependent.relationship.present?
+            is_valid_relationship = %w[daughter stepchild foster_child grandchild sister nephew half_sister stepbrother son brother niece half_brother stepsister].include?(dependent.relationship.downcase)
+          end
+          if dependent.odc_qualifying == true && dependent.under_17? && is_valid_relationship
+            count += 1
+          end
+        end
```
mpidcock, line 145: Looks like this is very similar logic as `eligible_for_child_tax_credit`, can that method be used here (might need some adjustments)?

mpidcock (approved):
Looks good! I only noticed one minor improvement.

app/lib/efile/ny/it213.rb
tahsinaislam: yes! i can reuse this method here.

PR #4009, opened 2024-01-16
Calculate NYS taxes

bengolder:
Addresses Section J: https://www.pivotaltracker.com/story/show/186692297
Implements https://www.tax.ny.gov/forms/html-instructions/2023/it/it201i-2023.htm#nys-tax-table

The existing calculations had several bugs. I refactored the code to be more easily unit-tested if we want more tests.
The new unit tests I added were given by Courtney.

### DIFF
--- a/app/lib/efile/ny/nys_tax_rate_schedule.rb
+++ b/app/lib/efile/ny/nys_tax_rate_schedule.rb
@@ -1,6 +1,8 @@
 module Efile
   module Ny
     class NysTaxRateSchedule
+      # Implements the NYS Tax Rate Schedule, as documented in the IT-201 instructions
+      # https://www.tax.ny.gov/forms/html-instructions/2023/it/it201i-2023.htm#nys-tax-rate-schedule
       class << self
         def calculate(taxable_income, filing_status)
           rows = case filing_status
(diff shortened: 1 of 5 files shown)


app/lib/efile/ny/nys_tax_rate_schedule.rb
```diff
@@ -0,0 +1,85 @@
...
+        def initialize(floor, ceiling, cumulative, rate)
+          @floor = floor
+          @ceiling = ceiling
+          @cumulative = cumulative
+          @rate = rate
+        end
+
+        def compute(amount)
+          if amount < 0
+            raise "Negative income input"
```
mpidcock, line 72: Can negative income happen legitimately when claiming deductions? I remember seeing this in some testing scenarios with student loan interest/teaching supplies deductions, but I don't know if they were realistic. (That's when the "negative infinity" range was introduced.)

mpidcock (approved):
I looked through the instructions and the rate schedules and could not find any errors! Thanks for simplifying this code. I also looked for ways to show the instructions in comments, but couldn't find a good answer so future coders will probably still have to pull up the NY documents.

app/lib/efile/ny/nys_tax_rate_schedule.rb
bengolder: I don't think so, because I assume it should just be zero instead of negative. [The instructions](https://www.tax.ny.gov/pit/file/tax-tables/it201i-2023.htm) imply that $0 is the lowest possible value. I can check with Courtney.

PR #4025, opened 2024-01-17
Javier submission #186646846

tofarr:
Updates to the XML. The story has a bunch of updates to the XML - mostly omitting elements when the value is 0, but also adding some values (soft_vndr_id, form_type), and removing some altogether (country_name)


app/lib/submission_builder/ty2022/states/ny/documents/it201.rb
```diff
@@ -59,27 +59,27 @@ def document
                 xml.TOT_NRFNDCR_AMT claimed: calculated_fields.fetch(:IT201_LINE_43)
                 xml.TX_AFT_NRFNDCR_AMT claimed: calculated_fields.fetch(:IT201_LINE_44)
                 xml.TOT_TX_AMT claimed: calculated_fields.fetch(:IT201_LINE_46)
-                xml.NYC_TXBL_INC_AMT claimed: calculated_fields.fetch(:IT201_LINE_47)
```
bengolder, line 62: Totally minor observation: If you wanted, you could also move the call to calculated_fields into the new method and just pass in the line symbol as the second argument

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -64,13 +66,15 @@ def documents_wrapper
                 xml.SCHOOL_CD @submission.data_source.school_district_number
                 xml.SCHOOL_NAME @submission.data_source.school_district&.truncate(30)
                 xml.PR_EMP_DESC @submission.data_source.direct_file_data.primary_occupation
-                xml.COUNTRY_NAME @submission.data_source.mailing_country
+                #xml.COUNTRY_NAME @submission.data_source.mailing_country
```
bengolder, line 69: A comment to explain why country is not needed might help future maintainers

bengolder (approved):
Looks great! I appreciate having a new method for appropriately handling blank or zero values.

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
tofarr: Good call!

app/lib/submission_builder/ty2022/states/ny/documents/it201.rb
tofarr: Excellent Suggestion - Implemented!

PR #4029, opened 2024-01-18
NY 21 item review

bengolder:
- Addresses https://www.pivotaltracker.com/story/show/186864973
- Adds 21 new fields in a collapsible group to the NY data review page
- I had to make `line_or_zero` into a public method on the tax calculator in order to easily access all the various numbers

Kapture 2024-01-17 at 18 00 30

### DIFF
--- a/app/models/state_file_base_intake.rb
+++ b/app/models/state_file_base_intake.rb
@@ -46,9 +46,15 @@ class StateFileBaseIntake < ApplicationRecord
     end
   end
 
+  def calculator
+    unless @calculator.present?
+      @calculator = tax_calculator
+      @calculator.calculate
+    end
+    @calculator
+  end
+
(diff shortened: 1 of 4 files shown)


app/models/state_file_base_intake.rb
```diff
@@ -46,9 +46,15 @@ def synchronize_df_dependents_to_database
     end
   end
 
+  def calculator
+    unless @calculator.present?
+      @calculator = tax_calculator
+      @calculator.calculate
+    end
+    @calculator
+  end
```
bengolder, line 55: This should cache the calculations in memory so that we don't accidentally rerun the calculations on the same web request.

mpidcock (approved):
LGTM! thanks for the dividers and good catch with caching the calculate

PR #4031, opened 2024-01-18
Calculate NYC EIC

bengolder:
Part of https://www.pivotaltracker.com/story/show/186814386

- I implemented the NYC EIC Rate Worksheet using a new TaxCalculator subclass, because I wanted us to be able to pull explanations of the calculations.
- I added the NYC EIC Rate Table as a separate class, so I could test it independently
- This only updates calculations. I still need to confirm if the XML and PDF are filled correctly.


tofarr (approved):
🍰

PR #4034, opened 2024-01-18
Small fixes for NYC EIC

bengolder:
Finishing touches for https://www.pivotaltracker.com/story/show/186814386
- Ensures that NYC EIC is only calculated for NYC full-year residents
- Moves NYC EIC calculation into method so that the calculator can extract an explanation of how line 27 was calculated.
- Removes the conditional on the XML, allowing the value of zero to be passed through.


embarnard (approved):
Looks good!

PR #4040, opened 2024-01-18
Add columns for new NYC residency questions

bengolder:
First step for: https://www.pivotaltracker.com/story/show/186865012

Adds two new columns for the new NYC residency questions. These are semi-redundant with some existing columns (`nyc_full_year_resident`, `eligibility_part_year_nyc_resident`), but will store a more accurate residency status than the others. Depending on some possible changes to the UI, we may be able to remove the other columns at a later date.

### DIFF
--- a/app/models/state_file_ny_intake.rb
+++ b/app/models/state_file_ny_intake.rb
@@ -44,6 +44,8 @@
 #  ny_mailing_street                  :string
 #  ny_mailing_zip                     :string
 #  nyc_full_year_resident             :integer          default("unfilled"), not null
+#  nyc_maintained_home                :integer          default("unfilled"), not null
+#  nyc_residency                      :integer          default("unfilled"), not null
 #  occupied_residence                 :integer          default("unfilled"), not null
 #  payment_or_deposit_type            :integer          default("unfilled"), not null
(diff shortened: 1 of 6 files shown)


app/models/state_file_ny_intake.rb
```diff
@@ -99,6 +101,8 @@
 class StateFileNyIntake < StateFileBaseIntake
   encrypts :account_number, :routing_number, :raw_direct_file_data
 
+  enum nyc_residency: { unfilled: 0, full_year: 1, part_year: 2, none: 3 }, _prefix: :nyc_residency
+  enum nyc_maintained_home: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nyc_maintained_home
   enum nyc_full_year_resident: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nyc_full_year_resident
```
mpidcock, line 106: Is there a good convention for marking `nyc_full_year_resident` and `eligibility_part_year_nyc_resident` as deprecated/candidates for future deletion?

mpidcock (approved):
Ideally the old fields should be commented or something, but it is fine how it is

PR #4041, opened 2024-01-18
 Numbers fix for IT201 line 48 #186692312

tofarr:
I think this is the result of copy pasta - the first numbers were slightly wrong:

Source: https://www.tax.ny.gov/forms/current-forms/it/it201i.htm

image


rickreyhsig (approved):
LGTM

PR #4043, opened 2024-01-18
Signature Updates #186866900

tofarr:
Signatures are not to be included in the AZ8879 PDF: (I guess because the instructions explicitly say this is to be Pen and Ink):
image


mpidcock (approved):
🍰

PR #4047, opened 2024-01-19
Explicitly signing out az and ny intakes. Renamed method for clarity

tofarr:


app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -5,13 +5,26 @@ module AuthenticatedStateFileIntakeConcern
...
+      @intake = self.send(current_intake_name)
+    end
+    @intake
+  end
+
   private
 
   def card_postscript
     I18n.t("state_file.state_file_pages.card_postscript.responses_saved_html").html_safe
   end
```
bengolder, line 23: Your idea about having an explicit authenticated intake is valuable and could even be used to improve and refactor this file a little. It was something I suggested when we did the earlier auth work, but was seen a "nice to have" refactor that didn't get done.

```suggestion
def current_authenticated_intake
    case params[:us_state]
    when "ny" then current_state_file_ny_intake
    when "az" then current_state_file_az_intake
end
```

app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -5,13 +5,26 @@ module AuthenticatedStateFileIntakeConcern
...
 
   def require_state_file_intake_login
+    # NOTE: Most places in the code do not seem to use the authenticated version of
+    # current_state_file_XX_intake - they use current_intake instead which is
+    # unauthenticated.
     if current_state_file_az_intake.blank? && current_state_file_ny_intake.blank?
```
bengolder, line 28: This is a good spot to use the current_authenticated_intake. We should be explicit here. and bein authenticated for a different state should not qualify as authenticated for this state.
```suggestion
    if current_authenticated_intake.blank?
```

app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -5,13 +5,26 @@ module AuthenticatedStateFileIntakeConcern
...
   end
 
+  def current_intake
+    # NOTE: This intake is different from the unauthenticated value, stored in the regular session
+    # (Which the unauthenticated sessions controller uses)
+    unless @intake
+      current_intake_name = "current_#{question_navigator.intake_class.name.underscore}"
+      @intake = self.send(current_intake_name)
+    end
+    @intake
```
bengolder, line 15: I'm uncertain about the memoization. I like the tiny performance gain but I am unsure if it introduces risks. Either way, I think the explict authenticated intake method would be helpful here.

```suggestion
    @_authenticated_intake ||= current_authenticated_intake
```

tofarr:
The updates from this were handled by a different PR

PR #4052, opened 2024-01-19
Verification email: Update subject, copy and logo

rickreyhsig:


app/mailers/verification_code_mailer.rb
```diff
@@ -11,6 +11,18 @@ def with_code
...
+    @service_name = service.service_name
+    @service_type = service.service_type
+    @locale = params[:locale]
+    @subject = I18n.t("messages.verification_code_subject_with_service_name", service_name: @service_name, locale: @locale)
+    @verification_code = params[:verification_code]
+    attachments.inline['logo.png'] = service.email_logo
+
+    mail(to: params[:to], subject: @subject, from: service.noreply_email, delivery_method_options: service.delivery_method_options)
+  end
+
```
bengolder, line 25: `with_code_fyst` looks like the exact same logic as `with_code` above. The templates appear nearly identical too. I think it would make sense to just use or modify the original template and `with_code` method. I don't think we should add a new method that does the same thing. Am I missing something that can't already be handled by the existing method and template?

bengolder (changes requested):
I could be missing something, but this looks like it copies an existing method and renames it. I don't understand why the duplication is needed and would suggest using the existing method or making any minor updates needed.

app/mailers/verification_code_mailer.rb
rickreyhsig: The main problem here is that the erb template is dictated by the name of this method and I thought a semi repetitive method was a lesser evil than having a conditional in the view.  If there was is way to specify which view is triggered fir the body of the email here, I could get rid of this method altogether.

app/views/verification_code_mailer/with_code.html.erb
```diff
@@ -1 +1,5 @@
-<%= t("verification_code_mailer.with_code.body_html", service_name: @service_name, verification_code: @verification_code, locale: @locale) %>
+<% if @service_type == :statefile %>
+  <%= t("verification_code_mailer.with_code_fyst.body_html", service_name: @service_name, verification_code: @verification_code, locale: @locale) %>
+<% else %>
+  <%= t("verification_code_mailer.with_code.body_html", service_name: @service_name, verification_code: @verification_code, locale: @locale) %>
+<% end %>
```
rickreyhsig, line 5:
```suggestion
<% end %>

```

app/mailers/verification_code_mailer.rb
rickreyhsig: I've tried also using the render template method here but that brings about another can of worms and Tiffany has mentioned that we don't intertwine the statefile template with ctc & gyr.

@mpidcock

app/views/verification_code_mailer/with_code.html.erb
```diff
@@ -1 +1,2 @@
 <%= t("verification_code_mailer.with_code.body_html", service_name: @service_name, verification_code: @verification_code, locale: @locale) %>
+
```
rickreyhsig, line 2:
```suggestion
```

mpidcock (approved):
thanks for the updates!

PR #4054, opened 2024-01-20
Deprecate nyc_full_year_resident

bengolder:
The penultimate PR for https://www.pivotaltracker.com/story/show/186865012
The only remaining step after this PR is to remove the deprecated database column

- updates all code that depended on the now-deprecated `nyc_full_year_resident` column to instead use the newer `nyc_residency` column
- stops the NY County page from setting `nyc_full_year_resident` or any other NYC residency data
- updates the NY review page section about NYC residency to link to NycResidency instead of NyCounty
- updates the NycResidencyController so that it could appropriately handle going back to the review page or redirecting to offboarding depending on whether the answer was disqualifying or the page was accessed from the review page.
- ensures that values relating to nyc residency are set appropriately in the XML and PDF
- includes a minor readability refactor in `submission_builder/ty2022/states/ny/documents/it201.rb` to replace all the calls to `@submission.data_source` with calls to an `intake` method.

### DIFF
--- a/app/lib/pdf_filler/ny201_pdf.rb
+++ b/app/lib/pdf_filler/ny201_pdf.rb
@@ -54,7 +54,8 @@ module PdfFiller
         Foreign_account: xml_value_to_pdf_checkbox('Foreign_account', 'FORGN_ACCT_IND'),
         yonkers_freeze_credit: xml_value_to_pdf_checkbox('yonkers_freeze_credit', 'YNK_LVNG_QTR_IND'),
         D4: 'no',
-        E1: xml_value_to_pdf_checkbox('E1', 'NYC_LVNG_QTR_IND'),
+        E1: claimed_attr_value('NYC_LVNG_QTR_IND') == '2' ? 'no' : 'Off',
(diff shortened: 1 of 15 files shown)


app/lib/pdf_filler/ny201_pdf.rb
```diff
@@ -54,7 +54,8 @@ def hash_for_pdf
         Foreign_account: xml_value_to_pdf_checkbox('Foreign_account', 'FORGN_ACCT_IND'),
         yonkers_freeze_credit: xml_value_to_pdf_checkbox('yonkers_freeze_credit', 'YNK_LVNG_QTR_IND'),
         D4: 'no',
-        E1: xml_value_to_pdf_checkbox('E1', 'NYC_LVNG_QTR_IND'),
+        E1: claimed_attr_value('NYC_LVNG_QTR_IND') == '2' ? 'no' : 'Off',
```
mpidcock, line 57: please clarify with a comment

mpidcock (approved):

PR #4059, opened 2024-01-22
Omitting zeros from XML where possible #186880203

tofarr:


app/lib/submission_builder/ty2022/states/ny/documents/it214.rb
```diff
@@ -9,40 +9,40 @@ class It214 < SubmissionBuilder::Document
...
+                  add_non_zero_claimed_value(xml, :R_ASMT_AMT, :IT214_LINE_24)
+                  add_non_zero_claimed_value(xml, :R_B4_EXMPT_AMT, :IT214_LINE_25)
+                  add_non_zero_claimed_value(xml, :R_HOME_RPTX_AMT, :IT214_LINE_27)
                 end
-                xml.R_GRSS_AVL_AMT claimed: calculated_fields.fetch('IT214_LINE_29')
-                xml.R_TNTV_RL_CR_AMT claimed: calculated_fields.fetch('IT214_LINE_30')
-                xml.R_TX_AVL_CR_AMT claimed: calculated_fields.fetch('IT214_LINE_31')
-                xml.R_CR_LMT_AMT claimed: calculated_fields.fetch('IT214_LINE_32')
-                xml.R_RL_PROP_CR_AMT claimed: calculated_fields.fetch('IT214_LINE_33')
+                xml.R_RL_PROP_TX_AMT claimed: calculated_fields.fetch('IT214_LINE_28') || 0
```
tofarr, line 40: Before this change, this line was inside the if block, but I noticed that it has a "minOccurs" of "1" in the XSD - Indicating that it is required - so I moved it outside the if block and added a default value of 0.

tofarr, line 40: L40 - Before this change, this line was inside the if block, but I noticed that it has a "minOccurs" of "1" in the XSD - Indicating that it is required - so I moved it outside the if block and added a default value of 0.

tahsinaislam (approved):

PR #4061, opened 2024-01-22
Add banking information to AZ 140 pdf and Financial Transaction XML

mpidcock:


app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -29,6 +29,9 @@ def document
             attached_documents.each do |attached|
               document.at('ReturnDataState').add_child(document_fragment(attached))
             end
+            if !@submission.data_source.routing_number.nil? && !@submission.data_source.account_number.nil?
```
jenny-heath, line 32: check that we clear this out if someone switches to mail from direct deposit

jenny-heath (approved):

PR #4068, opened 2024-01-23
Updated additional dependents PDF #186781037

tofarr:
The IT201 form has a supplimental additional dependents PDF:
image


app/lib/pdf_filler/additional_dependents_pdf.rb
```diff
@@ -9,12 +9,7 @@ def source_pdf_name
 
     def initialize(submission, start_node: 4)
       # For some PDF fields, use values from the database b/c the XML values are truncated or missing.
-      if submission.tax_return.present?
-        @xml_document = SubmissionBuilder::Ty2021::Documents::Irs1040.new(submission).document
-      else
-        # State file does not have a submission.tax_return - so we grab the raw data.
-        @xml_document = Nokogiri::XML(submission.data_source.raw_direct_file_data)
-      end
+      @xml_document = SubmissionBuilder::Ty2021::Documents::Irs1040.new(submission).document
```
tofarr, line 12: This is actually a reversion - since statefile does not use this class anymore, I am putting it back to its original state.

tahsinaislam (approved):

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -213,9 +213,8 @@ def supported_documents
 
             supported_docs << {
               xml: nil,
-              pdf: PdfFiller::AdditionalDependentsPdf,
+              pdf: SubmissionBuilder::Ty2022::States::Ny::Documents::It201AdditionalDependents,
```
rickreyhsig, line 216: Was the AdditionalDependentsPdf not working at all?

app/lib/submission_builder/ty2022/states/ny/documents/it201_additional_dependents.rb
```diff
@@ -0,0 +1,41 @@
+module SubmissionBuilder
```
jenny-heath, line 1: i think this might be in the wrong place...? this folder, i *think*, is for xml builders but this is a pdf builder and should go in `app/lib/pdf_filler/`

jenny-heath: also, a unit spec for this would probably be good

spec/lib/submission_builder/ty2022/states/ny/individual_return_spec.rb
```diff
@@ -107,7 +107,7 @@
       it "creates an additional dependents pdf" do
         submission_builder = SubmissionBuilder::Ty2022::States::Ny::IndividualReturn.new(submission)
         additional_dependents = submission_builder.pdf_documents.select do |d|
-          d.pdf == PdfFiller::AdditionalDependentsPdf
+          d.pdf == SubmissionBuilder::Ty2022::States::Ny::Documents::It201AdditionalDependents
```
jenny-heath, line 110: should we check that this doesn't get added when there are <= 6 dependents? maybe overkill since pdfs are less important than xml, so up to you

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
tofarr: It has different fields, accepts different input, and is all in all just a different case. So I decided to change what was used.

spec/lib/submission_builder/ty2022/states/ny/individual_return_spec.rb
tofarr: Good call.

app/lib/pdf_filler/it201_additional_dependents_pdf.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def initialize(submission, dependent_offset: 7, source_pdf_name: "it201_additional_dependents")
+      @intake = submission.data_source
+      @dependent_offset = dependent_offset
+      @source_pdf_name = source_pdf_name
+    end
+
+    def hash_for_pdf
+      answers = {}
+      answers.merge!(dependents_info) if @intake.dependents.length > @dependent_offset
```
jenny-heath, line 14: in general i like it when we fill the pdf's from the xml, but i was having trouble with that myself for the dependent stuff so idk how much of a stickler i want to be about it...

spec/lib/pdf_filler/it201_additional_dependents_pdf_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:intake) { create(:state_file_zeus_intake) }
+
+  let(:submission) { create :efile_submission, tax_return: nil, data_source: create(:state_file_az_intake) }
+  let(:pdf) { described_class.new(submission) }
+
+  describe '#hash_for_pdf' do
+    it 'uses field names that exist in the pdf' do
+      pdf_fields = filled_in_values(submission.generate_filing_pdf.path)
+      missing_fields = pdf.hash_for_pdf.keys.map(&:to_s) - pdf_fields.keys
+      expect(missing_fields).to eq([])
```
jenny-heath, line 14: we have a bunch of tests that look like this and they are all deeply confusing to me

PR #4083, opened 2024-01-24
State file: default numeric values to 0 when reading from direct file data

jenny-heath:


app/models/state_file_ny_intake.rb
```diff
@@ -163,7 +163,6 @@ def tax_calculator(include_source: false)
 
   def calculate_sales_use_tax
     fed_agi = direct_file_data&.fed_agi
-    return unless fed_agi
```
mpidcock, line 166: is there a default value for fed_agi?

mpidcock (approved):

PR #4086, opened 2024-01-24
It 213 att #186828458

jenny-heath:


app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -181,6 +181,12 @@ def supported_documents
                 pdf: PdfFiller::Ny213Pdf,
                 include: receiving_213_credit
               },
+              {
+                xml: SubmissionBuilder::Ty2022::States::Ny::Documents::It213,
+                pdf: PdfFiller::Ny213AttPdf,
+                include: @submission.data_source.dependents.select(&:eligible_for_child_tax_credit).length > 6,
```
mrotondo, line 187: the 6 could be a public constant here (or somewhere else sensible), used in both places in this file and then either passed to or accessed by ny213_att_pdf.rb (and I think the param default should be removed there in either case).

file.txt
mrotondo, line 1: delet this

mrotondo (approved):
looks good but you committed a file accidentally i think!

app/lib/pdf_filler/ny213_att_pdf.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def nys_form_type
+      "239"
+    end
+
+    def barcode_overlay_rect
+      [[0, 26], 125, 29]
+    end
+
```
tahsinaislam, line 18: Are the methods nys_form_type and barcode_overlay_rect being used?

PR #4088, opened 2024-01-25
add prod directfile api url to prod and staging credentials, and copy prod certs and private keys into staging credentials

mrotondo:


mrotondo:
this can be reviewed by looking at prod and staging credentials on this branch and verifying that they have the same df api url and az & ny certs and keys

tofarr (approved):

PR #4090, opened 2024-01-25
fix old migration that no longer ran - :references may have worked in the past but no longer does

mrotondo:


mrotondo:
you can verify this (if you're ok with wiping your local DB) by running `bundle exec rake db:migrate:reset` without and then with this change

PR #4092, opened 2024-01-25
More testing scenarios for dep dropping bug

rickreyhsig:
https://www.pivotaltracker.com/story/show/186900971

NY wants to to have distinction of false from unfilled/nil values for eic_disability & eic_student fields.

Screen Shot 2024-01-26 at 3 32 55 PM

Screen Shot 2024-01-26 at 3 33 15 PM


db/migrate/20240125194731_update_state_file_dependent_eic_disablity_and_student.rb
```diff
@@ -0,0 +1,29 @@
+class UpdateStateFileDependentEicDisablityAndStudent < ActiveRecord::Migration[7.1]
+  def up
```
rickreyhsig, line 2: The reason for the two migrations here is that since we have strong migrations configured, this was the 'easiest' way to actually rename and the column while providing some kind of backfill.  Please let me know if there is a better way of doing this.

app/lib/pdf_filler/ny215_pdf.rb
```diff
@@ -65,8 +65,8 @@ def hash_for_pdf
                          "ln34suf#{index}" => dependent.suffix,
                          "ln34real#{index}" => dependent.relationship_label,
                          "month#{index}" => dependent.months_in_home,
-                         "ln34disability#{index}" =>  dependent.eic_disability ? "Yes" : "Off",
-                         "ln34student#{index}" =>  dependent.eic_student ? "Yes" : "Off",
+                         "ln34disability#{index}" => dependent.eic_disability_yes? ? "Yes" : "Off",
```
embarnard, line 68: nice 💯

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
```diff
@@ -104,10 +104,14 @@ def documents_wrapper
                 xml.dependent do
                   xml.DEP_SSN_NMBR dependent.ssn if dependent.ssn.present?
                   xml.DEP_SEQ_NMBR index+1
-                  xml.DEP_DISAB_IND dependent.eic_disability == true ? 1 : 2
+                  unless dependent.eic_disability_unfilled?
+                    xml.DEP_DISAB_IND dependent.eic_disability_yes? ? 1 : 2
```
embarnard, line 108: why do some of them we fill them `2` when its `unfilled` and some of them we just leave it out?

app/models/direct_file_data.rb
```diff
@@ -616,10 +616,27 @@ def dependents
         unless self.mailing_state == 'AZ'
           dependent.months_in_home = eitc_dependent_node.at('MonthsChildLivedWithYouCnt')&.text.to_i
         end
-        dependent.eic_student = eitc_dependent_node.at('ChildIsAStudentUnder24Ind')&.text == "true"
-        dependent.eic_disability = eitc_dependent_node.at('ChildPermanentlyDisabledInd')&.text == "true"
+        dependent.eic_student = case eitc_dependent_node.at('ChildIsAStudentUnder24Ind')&.text
+                                when 'true'
```
embarnard, line 620: nice! Wonder if we can break this out to a separate method. But also fine to keep it here

db/migrate/20240125194731_update_state_file_dependent_eic_disablity_and_student.rb
embarnard: Do we need to do a backfill if we are just have demo data right now? These migrations look good to me though, good work!

embarnard: Usually we would do the back fills through creating a data migration under `db/data` instead of `db/migrate` and run it through a job on `http://localhost:3000/hub/data_migrations/` but I think for demo data this is fine

embarnard (approved):
Looks good! I had a few qs but nothing blocking

db/migrate/20240125194731_update_state_file_dependent_eic_disablity_and_student.rb
rickreyhsig: I wasn't sure if this was really necessary but figured it might be good to have some semi-realistic data in demo.

rickreyhsig: I'll keep this in mind for next time.

app/lib/submission_builder/ty2022/states/ny/individual_return.rb
rickreyhsig: That is what NY is requesting.  If the DF data comes out with no associated tag we will store this in the DB as unfilled and not show it on the state XML.  Whenever it is present as true or false though we will show it with the appropriate code 1 or false respectively.

app/models/direct_file_data.rb
rickreyhsig: Good idea, I'll break it up!

PR #4096, opened 2024-01-25
It215: don't fill in negative EIC value, use 0

jenny-heath:


app/lib/submission_builder/document.rb
```diff
@@ -63,5 +63,16 @@ def add_non_zero_claimed_value(xml, elem_name, claimed)
...
     end
+
+    def add_positive_claimed_value_or_zero(xml, elem_name, claimed)
+      claimed_value = calculated_fields.fetch(claimed)
+      if claimed_value.present?
+        if claimed_value.to_i < 0
+          xml.send(elem_name, claimed: 0)
+        else
+          xml.send(elem_name, claimed: claimed_value)
+        end
```
mpidcock, line 74: I think you could simplify this a bit
```suggestion
        xml.send(elem_name, claimed: Math.max(claimed_value, 0))
```

mpidcock (approved):

PR #4100, opened 2024-01-26
ny xml header fixes

tahsinaislam:


app/lib/submission_builder/ty2022/states/authentication_header.rb
```diff
@@ -63,19 +62,17 @@ def xml_type_for_state_id(state_id)
...
           xml_type = xml_type_for_state_id(state_id)
           if xml_type
-            xml_builder.send("#{xml_type}Num", state_id.id_number)
-            xml_builder.send("#{xml_type}StCd", state_id.state)
+            xml_builder.send("#{xml_type}Num", state_id.id_number) if state_id.id_number.present?
+            xml_builder.send("#{xml_type}StCd", state_id.state) if state_id.state.present?
             xml_builder.send("#{xml_type}ExprDt") do
               if state_id.non_expiring?
-                xml_builder.NonExpr
+                xml_builder.NonExpr "X"
```
tofarr, line 69: I looked at the XSD for this:
```<xsd:element name="NonExpr" fixed="Non-Expiring"/>```
It looks to me that this should have the value "Non-Expiring" rather than "X"?

app/lib/submission_builder/ty2022/states/authentication_header.rb
tofarr, line 1: General comment - to handle this in IT201, we added a function `add_non_zero_claimed_value` to `SubmissionBuilder::Document` - it may make some of the checks less verbose. Also - if the intent is to omit values where they are `0` or `"0"`, then `present?` wont work - both these values are `present?`

tofarr: NVM - I see the values you check for are not numeric

tofarr (approved):

PR #4102, opened 2024-01-26
Refactor data transfer buttons on initiate-data-transfer and canceled-data-transfer pages

embarnard:


embarnard:
Now production and staging environments will look like this:
Screenshot 2024-01-25 at 6 11 19 PM

And other environments will look like this:
Screenshot 2024-01-25 at 6 10 22 PM

app/views/state_file/questions/initiate_data_transfer/_data_transfer_buttons.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if @fake_data_transfer_link.present? %>
+  <%= link_to @fake_data_transfer_link.to_s, class: "button button--primary button--icon button--full-width", id: "firstCta" do %>
+    <%= image_tag("icons/link-white.svg", alt: "") %>
+    <%= t(".from_fake_df_page") %>
+  <% end %>
+<% end %>
+
+<% if @irs_df_transfer_link.present? %>
+  <% if Rails.env.production? || Rails.env.staging? %>
```
rickreyhsig, line 9: Is there a benefit from doing check here instead of at the concern level?

rickreyhsig (approved):
LGTM, left a non-blocking question/comment.

app/views/state_file/questions/initiate_data_transfer/_data_transfer_buttons.html.erb
embarnard: Oh yes I suppose we could find a way to put it in the concern but since its just a front-end change I thought it made more sense to keep it in the view. They are getting the link from the same credential but we just want to add a note about not clicking unless asked for demo but make sure not to show in prod

PR #4114, opened 2024-01-29
Add qualifying person name when filing as Head Of Household or Qualifying Surviving Spouse

mpidcock:
https://www.pivotaltracker.com/story/show/186870783


spec/fixtures/files/fed_return_qualifyingperson_ny.xml
tahsinaislam, line 1: maybe the file should be named az now

spec/models/state_file_az_intake_spec.rb
```diff
@@ -229,6 +229,55 @@
...
+        expect(intake.hoh_qualifying_person_name[:first_name]).to eq "Nonparent"
+        expect(intake.hoh_qualifying_person_name[:last_name]).to eq "Younger"
+      end
+
+      it 'returns the oldest parent if there are no non-Parents in home more than 6 months' do
+        intake = build(:state_file_az_intake, filing_status: "head_of_household")
+        create :az_hoh_qualifying_person_parent,
+               first_name: "OlderParent", dob: StateFileDependent.senior_cutoff_date + 2.years, intake: intake
+        create :az_hoh_qualifying_person_parent,
+               first_name: "OlderParent", intake: intake
```
tahsinaislam, line 268: Should be named YoungerParent here

app/models/direct_file_data.rb
```diff
@@ -502,6 +503,15 @@ def primary_claim_as_dependent=(value)
     write_df_xml_value(__method__, value)
   end
 
+  def hoh_qualifying_person_name
+    df_xml_value(__method__)
+  end
+
+  def hoh_qualifying_person_name=(value)
+    create_or_destroy_df_xml_node(__method__, true, 'SpouseNm')
```
embarnard, line 511: Im definitely missing context but why are we destroying spouse name when writing hoh_qualifying_person_name?

app/models/state_file_az_intake.rb
```diff
@@ -175,4 +175,39 @@ def filing_status
     return :head_of_household if direct_file_data&.filing_status == 5 # Treat qualifying_widow as hoh
     super
   end
+
+  def requires_hoh_qualifying_person_name?
+    filing_status == :head_of_household
```
embarnard, line 180: Also QSS? Not sure what QSS is but saw it in the story 
```
populate this line on XML and PDF for all AZ HOH filers (including those who were federal QSS filers). 
```

app/models/direct_file_data.rb
mpidcock: I'll say I'm not the biggest fan of this method since it is not very clear, but it will create or destroy based on the boolean parameter, in this case `true` for create. This setter method is actually only called during test setup and not during any real code, it is intended to make sure all required parent fields are created before adding a new one, and that it is inserted sequentially after the expected previous field (in this case `SpouseNm`) @mrotondo showed me all of this, and you can see other setters in this file follow a similar pattern.

app/models/state_file_az_intake.rb
mpidcock: QSS is Qualifying Surviving Spouse, which is a recent rename from Qualifying Widow. (We still use "Widow" throughout most of our non-user facing code.) AZ defaults QSS to HOH, which was implemented as a previous story. The `StateFileAzIntake` overloads the `filing_status` to return HOH if filing as QSS. I can add a test case for this though so that if this logic changes then it won't break.

embarnard (approved):

PR #4116, opened 2024-01-29
MAIL_ZIP_5_ADR can only have 5 digits

rickreyhsig:
https://www.pivotaltracker.com/story/show/186923698


app/lib/pdf_filler/ny214_pdf.rb
```diff
@@ -26,22 +26,22 @@ def initialize(submission)
     end
 
     def hash_for_pdf
-      answers = {
+      {
```
rickreyhsig, line 29: answers was not used here, so I nuked it.

tahsinaislam (approved):

PR #4120, opened 2024-01-29
Fix bug on Worksheet A line 12 on IT-213 calculation

rickreyhsig:
https://www.pivotaltracker.com/story/show/186918929


mpidcock (approved):

mpidcock:
Adding this code doesn't hurt, but I think this alert should be fixed by code that was merged shortly after the last time it went off: https://github.com/codeforamerica/vita-min/pull/4083/files

PR #4132, opened 2024-01-30
test to check DF missing nodes

tahsinaislam:


app/helpers/df_xml_crud_methods.rb
```diff
@@ -1,6 +1,11 @@
 module DfXmlCrudMethods
   def df_xml_value(key)
-    node.at(selectors[key])&.text
+    return nil unless node.present?
+
+    selector = selectors[key]
+    return nil unless selector.present?
+
+    node.at(selector)&.text
   end
```
tahsinaislam, line 9: basically prevents any errors if a node is missing from fed xml

jenny-heath (commented):
can we add some tests for the cases where the nodes are missing? (add a new fixture with missing nodes)

PR #4137, opened 2024-01-31
statefile intercom integration

tahsinaislam:


app/controllers/application_controller.rb
```diff
@@ -304,8 +304,10 @@ def set_sentry_context
         is_bot: user_agent.bot?,
         request_id: request.request_id,
         user_id: current_user&.id,
-        client_id: current_client&.id
+        client_id: (current_client&.id || current_intake&.id)
```
embarnard, line 307: I'm guess we can just revert this back to what it was since we aren't trying to use current_client in the state file intercom partial anymore

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -115,6 +115,22 @@
   }
 }
 
+.button--blue {
```
embarnard, line 118: could we call this `.button--intercom-blue` or `.button--statefile-intercom`

app/views/layouts/application.html.erb
```diff
@@ -107,6 +107,10 @@
     <% end %>
     <%= yield(:script) if content_for?(:script) %>
     <!-- hiding intercom on hub again until we tackle styling it into the sticky action footer -->
-    <%= render("shared/intercom", isHub: hub?) unless (hub? || Routes::CtcDomain.new.matches?(request) || Routes::StateFileDomain.new.matches?(request)) %>
+    <% if state_file? %>
+      <%= render("shared/state_file_intercom", current_intake: current_intake, state: params[:us_state] )%>
```
embarnard, line 111: I'm curious does this work if we don't pass in `current_intake` and `params[:us_state]`?

app/views/shared/_state_file_intercom.html.erb
```diff
@@ -0,0 +1,26 @@
+<button type="button" class="open-intercom button--blue button--intercom ">
+  <%= image_tag("intercom-chat.svg", "aria-hidden": "true", class: "") %>
+  <%= I18n.t("general.intercom_cta") %>
+</button>
+
+<script>
+    const isClient = <%= current_intake.present? %>;
```
embarnard, line 7: change `isClient` to `hasIntake`?

config/credentials/production.yml.enc
embarnard, line 1: ran this `bin/gyr credentials_diff --environment production --base main` and looks like the non-statefile `intercom_access_token` and `secure_mode_secret_key` also changed?

config/credentials/staging.yml.enc
embarnard, line 1: `statefile_intercom_access_token` and `statefile_secure_mode_secret_key` are the same here as in the production file. the production config should use the other workspace and all non-prod environments should use the test workspace credentials

config/credentials/heroku.yml.enc
embarnard, line 1: looks like the `intercom_access_token` also changed here

config/credentials/circleci.yml.enc
embarnard, line 1: did we end up needing this? If not lets remove it

embarnard (changes requested):
Looks great! Really great job. I left some non-blocking suggestions but also some blocking suggestions. Let's double check we have the right access_tokens and secure_mode_secret_keys for all the environments and that they are configured for the right workspace. I wasn't able to check in the actual intercom space because I still don't have access to the statefile one

config/credentials/circleci.yml.enc
tahsinaislam: yeah we did for some reason. it was the only way circleci tests passed. Not sure if there's another way to do this

app/views/layouts/application.html.erb
tahsinaislam: you're right i don't need to pass it in!

config/credentials/circleci.yml.enc
embarnard: Ooo okay interesting, can you add the key to lastpass in the shared folder? 🙏

config/credentials/heroku.yml.enc
embarnard, line 1: since we don't have the access token in this one I'm guessing you'll have to merge to demo for product acceptance

embarnard (approved):
Amazinnnn!! 🎆🎇🎆🎇🎆🎇🎆🎇 LGTM ✅

PR #4141, opened 2024-02-01
Login no Transfer #186956656

tofarr:
Can now login to intakes that do not have a transfer

### DIFF
--- a/app/controllers/state_file/intake_logins_controller.rb
+++ b/app/controllers/state_file/intake_logins_controller.rb
@@ -19,14 +19,18 @@ module StateFile
 
     def edit
       @form = IntakeLoginForm.new(possible_intakes: @records)
+      if @records.all? { |intake| intake.hashed_ssn.nil? }
+        session_sign_in
+        parts = request.path.split('/')
+        redirect_to StateFile::Questions::TermsAndConditionsController.to_path_helper(action: :edit, locale: parts[1], us_state: parts[2])
+      end
     end
 
     def update
       @form = IntakeLoginForm.new(intake_login_params)
       if @form.valid?
-        sign_in @form.intake
-        session[:state_file_intake] = @form.intake.to_global_id
-        redirect_to session.delete(:after_state_file_intake_login_path) || StateFile::Questions::DataReviewController.to_path_helper(us_state: params[:us_state])
+        session_sign_in
+        redirect_to StateFile::Questions::DataReviewController.to_path_helper(us_state: params[:us_state])
       else
         @records.each(&:increment_failed_attempts)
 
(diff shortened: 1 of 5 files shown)


app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -75,7 +76,21 @@ def service_type
...
+          redirect_to StateFile::Questions::TermsAndConditionsController.to_path_helper(us_state: params[:us_state])
+        end
+      end
+    end
+
+    def session_sign_in
+      intake = @records.first
+      sign_in intake
+      session[:state_file_intake] = intake.to_global_id
+      session.delete(:after_state_file_intake_login_path)
```
jenny-heath, line 93: i assume we're trying to redirect to after_state_file_intake_login_path here. does returning the value do that?

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -75,7 +76,25 @@ def service_type
...
+        if intake.hashed_ssn.present?
+          redirect_to StateFile::Questions::DataReviewController.to_path_helper(us_state: params[:us_state])
+        else
+          redirect_to StateFile::Questions::TermsAndConditionsController.to_path_helper(us_state: params[:us_state])
+        end
+      end
+    end
+
+    def sign_in_and_redirect(controller)
+      intake = @records.first
```
rickreyhsig, line 90: If this relation if filtering based on email or phone, we could do .take that will skip the order by in the SQL for a slight improvement.
```suggestion
      intake = @records.take
```

jenny-heath: what's the reason for picking the first one? i assume we just think they're all the same?

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -75,7 +76,25 @@ def service_type
     end
 
     def redirect_to_data_review_if_intake_authenticated
```
jenny-heath, line 78: should we rename this to reflect the fact that they can be redirected to different places now?

spec/controllers/state_file/intake_logins_controller_spec.rb
```diff
@@ -335,6 +335,16 @@
             expect(response).to redirect_to account_locked_portal_client_logins_path
           end
         end
+
+        context "when the client account does not have an ssn" do
```
jenny-heath, line 339: nit: the context block below says "when the intake does not have an ssn" so it would be nice to have consistent descriptions

jenny-heath: actually now that i've read the context block below these look...the same?

jenny-heath: hm i think i am responsible for this inconsistent naming but it would be great if the description were "as an unauthenticated intake" instead of "as an unauthenticated client"

jenny-heath (approved):

PR #4144, opened 2024-02-01
Set up staging and prod to send email from production domains, using new sending API keys provisioned on mailgun

mrotondo:


rickreyhsig (approved):
:shipit:

PR #4147, opened 2024-02-01
Remove coming-soon redirect on FAQ pages and add acts_like_production method for staging & production

embarnard:


app/controllers/state_file/state_file_pages_controller.rb
```diff
@@ -8,7 +8,7 @@ def redirect_locale_home
     end
 
     def fake_direct_file_transfer_page
-      return render "public_pages/page_not_found", status: 404 if Rails.env.production? || Rails.env.staging?
+      return render "public_pages/page_not_found", status: 404 if Rails.env.production?
```
tofarr, line 11: Should this be `acts_like_production?`

app/services/irs_api_service.rb
```diff
@@ -11,7 +11,7 @@ def self.df_return_sample
   end
 
   def self.import_federal_data(authorization_code, state_code)
-    unless Rails.env.production? || Rails.env.staging?
+    unless Rails.env.production?
```
tofarr, line 14: Again - should it be `acts_like_production?`

app/views/state_file/questions/initiate_data_transfer/_data_transfer_buttons.html.erb
```diff
@@ -6,7 +6,7 @@
 <% end %>
 
 <% if @irs_df_transfer_link.present? %>
-  <% if Rails.env.production? || Rails.env.staging? %>
+  <% if Rails.env.production? %>
```
tofarr, line 9: Should this be `acts_like_production?`

app/controllers/state_file/state_file_pages_controller.rb
embarnard: I thought so too at first but they will not have access to the IRS testing site so will have to go through fake-df flow more context here: https://cfastaff.slack.com/archives/C0544ERAFQV/p1706825778050309

app/services/irs_api_service.rb
embarnard: also same as above

app/views/state_file/questions/initiate_data_transfer/_data_transfer_buttons.html.erb
embarnard: also same as above

tofarr (approved):

PR #4149, opened 2024-02-01
Restoring step on login #186950035

tofarr:
When testing this I manually went through each step for NY and made sure that logins took users back to the correct step. I added a couple of specs to verify this too


embarnard (approved):
Nice! This looks great

PR #4151, opened 2024-02-01
Revert: barcodes in NY form PDFs (#3998)

jenny-heath:


app/views/state_file/questions/return_status/_accepted.html.erb
```diff
@@ -31,9 +31,9 @@
     (<%= @download_form_name %>)
   </p>
 
-  <p>
-    <%= link_to t('.download_voucher'), StateFile::PaymentVoucherController.to_path_helper(action: :show, us_state: params[:us_state]) %>
-  </p>
+  <div>
+    <%= link_to t('.download_voucher'), ActionController::Base.helpers.asset_path(@voucher_path) %>
+  </div>
```
jenny-heath, line 36: had a merge conflict here and not 100% sure why. i think this is the right path to the voucher pdf?

public/pdfs/it201v_1223.pdf
jenny-heath, line 1: i believe the story of this file is:
- the git revert restored this file
- my next commit:
  -  adds the new pdf from the story (#186948749) to public/pdfs/, overwriting the one that was restored in the previous step
  - removes the new pdf that i'd added to app/lib because i just blindly put all the new documents there without checking that we use them all

PR #4156, opened 2024-02-02
add a real submission id lookup to xml sample return service, and add a submission id for rudy_v2_ny

mrotondo:


mpidcock (approved):
Looks beautiful, appreciate you

PR #4172, opened 2024-02-05
Update general Homepage

rickreyhsig:
https://www.pivotaltracker.com/story/show/186856017

Screen Shot 2024-02-06 at 1 57 26 PM

### DIFF
--- a/app/views/state_file/state_file_pages/about_page.html.erb
+++ b/app/views/state_file/state_file_pages/about_page.html.erb
@@ -8,7 +8,7 @@
 </div>
 
 <div>
-  <%= t(".subheader_html") %>
+  <%= t(".subheader_html", login_path: login_options_path(us_state: "az")) %>
 </div><br>
 <!--
 <div>
--- a/app/views/state_file/state_file_pages/about_page.html.erb
+++ b/app/views/state_file/state_file_pages/about_page.html.erb
@@ -1,6 +1,12 @@
 <% title = t(".header") %>
-<h1 class="h2 strong" style="font-weight: bold"><%= title %></h1>
-<%= image_tag 'questions/welcome.svg' %>
+
+<div class="fyst-home-container">
+  <div class="fyst-home-content">
+    <h1 class="h1 strong" style="font-weight: bold"><%= title %></h1>
+  </div>
+  <%= image_tag 'questions/welcome.svg', class: 'fyst-home-image' %>
+</div>
+
 <div>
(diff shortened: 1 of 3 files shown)


app/views/state_file/state_file_pages/about_page.html.erb
```diff
@@ -1,18 +1,18 @@
...
+<div class="fyst-home-container">
+  <div class="fyst-home-content">
+    <h1 class="h1 strong" style="font-weight: bold"><%= title %></h1>
+  </div>
+  <%= image_tag 'questions/welcome.svg', class: 'fyst-home-image' %>
+</div>
+
 <div>
-  <%= link_to t(".action1_text"), "" %>
+  <%= t(".subheader_html", login_path: login_options_path(us_state: "az")) %>
```
rickreyhsig, line 11: The login pages currently are tied to a state so we should probably route based on the user's state since we wouldn't necessarily have it at this point.

app/views/state_file/state_file_pages/about_page.html.erb
```diff
@@ -1,18 +1,18 @@
...
 </div><br>
-<div>
-  <%= t(".action2") %>
-  <%= link_to t(".action2_text"), "" %>
-</div>
 
 <div class="blue-group spacing-below-15 spacing-above-15">
-  <%= t(".section1_html", direct_file_url: '') %>
+  <%= t(".section1_html", faq_link: state_faq_path(us_state: 'az')) %>
```
rickreyhsig, line 15: Ditto here or we make an intermediate page to request the state.

app/views/state_file/state_file_pages/about_page.html.erb
rickreyhsig: We could deduce the state after login and route appropriately.

tofarr: I have a PR at the moment that makes "us" an appropriate option for login - when passed in we then determine which state is required as part of login

app/views/state_file/state_file_pages/about_page.html.erb
tofarr: There is no "us" faq page - but instead we use `root_path(anchor: "common_questions")`

rickreyhsig: Done

tofarr (approved):

PR #4176, opened 2024-02-06
Can Login without first specifying State #186974885

tofarr:
We currently have a login to a single state (either az/ny) via a URL like:
http://statefile.localhost:3000/en/ny/login-options

I relaxed the constraints so that a more general login is possible using `us`. e.g.:
http://statefile.localhost:3000/en/us/login-options

I originally wanted to refactor the `ClientLoginService` (I had a branch for this) to break out StateFileLogin from the rest of logins, but this proved to be too tricky - so I settled on refactoring the service to accommodate multiple intake classes.

Routes were also refactored to allow 'us' for login options - though the FAQ does not link to any specific state. Once login occurs, we switch over to the appropriate path (az/ny)


app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -81,6 +81,7 @@ def service_type
       case params[:us_state]
       when "az" then :statefile_az
       when "ny" then :statefile_ny
+      when "us" then :statefile
```
rickreyhsig, line 84: 🥇

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -91,7 +92,7 @@ def redirect_to_data_review_if_intake_authenticated
         controller = intake.controller_for_current_step
         to_path = controller.to_path_helper(
           action: controller.navigation_actions.first,
-          us_state: params[:us_state]
+          us_state: intake.state_code
```
tofarr, line 95: Pulling the state code from the intake rather than the params, so redirecting into the correct state if param is `us`

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -106,7 +107,7 @@ def sign_in_and_redirect
...
```
tofarr, line 110: Pulling the state code from the intake rather than the params, so redirecting into the correct state if param is `us`

spec/services/client_login_service_spec.rb
```diff
@@ -83,6 +83,27 @@
...
         end
       end
+
+
+      context "with US intakes" do
+        subject { described_class.new(:statefile) }
+
+        it "returns the az intake with matching phone number" do
+          create :text_message_access_token, token: "hashed_token", sms_phone_number: "+16505551212"
+          matching_intake = create :state_file_ny_intake_after_transfer, phone_number: "+16505551212"
```
rickreyhsig, line 93: Did you mean to create a NY one here? or does it not affect the result?

config/routes.rb
```diff
@@ -558,14 +558,17 @@ def scoped_navigation_routes(context, navigation)
         resources :submission_pdfs, only: [:show], module: 'state_file/questions', path: 'questions/submission_pdfs'
         resources :federal_dependents, only: [:index, :new, :create, :edit, :update, :destroy], module: 'state_file/questions', path: 'questions/federal_dependents'
         resources :unemployment, only: [:index, :new, :create, :edit, :update, :destroy], module: 'state_file/questions', path: 'questions/unemployment'
+        get "/faq", to: "state_file/faq#index", as: :state_faq
+        get "/faq/:section_key", to: "state_file/faq#show", as: :state_faq_section
+        get "/data-import-failed", to: "state_file/state_file_pages#data_import_failed"
+      end
+
+      scope ':us_state', constraints: { us_state: /az|ny|us/ } do
```
rickreyhsig, line 566: This will make the param case insensitive which may reduce the amount of 404's if for w/e reason the state param isn't the correct case
```suggestion
      scope ':us_state', constraints: { us_state: /az|ny|us/i } do
```

spec/services/client_login_service_spec.rb
tofarr: Good catch

PR #4183, opened 2024-02-07
Session refactor #186742760

tofarr:
So basically, we are making sure that all session storage of intakes is now through devise / warden, rather than a custom session variable. This simplifies our session code as we now have a single method of storage rather than 2 - a lot of the hacks of deleting session variables and then re-signing in simply go away now.


app/channels/application_cable/connection.rb
```diff
@@ -15,7 +15,17 @@ def current_user
     end
 
     def current_state_file_intake
-      GlobalID.find(@request.session[:state_file_intake])
+      return @current_state_file_intake if @current_state_file_intake
```
tofarr, line 18: So I think this is one reason why previous developers were reluctant to use devise / warden for everything - there is no `current_state_file_xx_intake` method in the action cable connection, and getting this is kind of tricky.

We don't check the last_accessed session variable, so I suppose in theory somebody could keep a websocket open longer than 15 minutes, but I am not sure how this would be a usable exploit.

app/controllers/flows_controller.rb
```diff
@@ -39,7 +39,8 @@ def generate
       if intake.respond_to?(:client)
         sign_in(intake.client)
       elsif [:state_file_az, :state_file_ny].include?(type)
-        session[:state_file_intake] = intake.to_global_id
+        sign_in intake
+        # session[:state_file_intake] = intake.to_global_id
```
rickreyhsig, line 43: Do you mean the leave this comment @tofarr ?

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -100,7 +100,6 @@ def redirect_to_data_review_if_intake_authenticated
     def sign_in_and_redirect
       intake = @records.take
       sign_in intake
-      session[:state_file_intake] = intake.to_global_id
```
rickreyhsig, line 103: Nice

app/channels/application_cable/connection.rb
rickreyhsig: This is nice work Tim!! I would add a unit spec for this method to spec/channels/application_cable/connection_spec.rb to really be confident that we are doing the right thing when manipualting the warden values in the session.

app/controllers/flows_controller.rb
```diff
@@ -70,7 +71,7 @@ def show
 
   def current_intake
     if %w[state_file_az state_file_ny].include?(params[:id] || params[:type])
-      GlobalID.find(session[:state_file_intake])
+      current_state_file_az_intake || current_state_file_ny_intake
```
jenny-heath, line 74: this makes me wonder: can we configure devise to only allow one of these to be signed in at the same time? (my assumption right now is that both can be, which i assume we don't want?)

spec/controllers/state_file/intake_logins_controller_spec.rb
```diff
@@ -421,7 +421,7 @@
 
             expect(subject.current_state_file_az_intake).to eq(intake)
             expect(response).to redirect_to az_questions_data_review_path(us_state: "az")
-            expect(GlobalID.find(session[:state_file_intake])).to eq intake
+            expect(session["warden.user.state_file_az_intake.key"].first.first).to eq intake.id
```
jenny-heath, line 424: i suppose there isn't any guidance for how to access these things in tests? tried a cursory search and couldn't find anything

jenny-heath (commented):
assuming you did a global search for session[:state_file_intake] to replace all of them?

app/controllers/state_file/questions/data_review_controller.rb
```diff
@@ -16,7 +16,6 @@ def edit
           ip_address: ip_for_irs,
           intake: current_intake,
         )
-        sign_in current_intake
```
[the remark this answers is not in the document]
jenny-heath: i think we're nuking it because this used to be the first place where we sign them in but now it happens when the intake is created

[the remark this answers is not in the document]
tofarr: We are already logged in - we no longer need to re-login in the data review controller

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -19,9 +19,8 @@ def ip_for_irs
       private
 
       def current_intake
-        intake = GlobalID.find(session[:state_file_intake])
-        return nil if intake && !intake.is_a?(question_navigator.intake_class)
-        intake
+        state_code = question_navigator.intake_class.new.state_code
```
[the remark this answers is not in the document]
tofarr: I like it - I was thinking of something similar

app/controllers/flows_controller.rb
tofarr: Done!

app/channels/application_cable/connection.rb
jenny-heath: found [this](https://stackoverflow.com/questions/49866041/actioncable-authentication-with-multiple-devise-models) and tested it out locally. seems like you can use `env['warden'].user(:state_file_az_intake)` or `env['warden'].user(:state_file_ny_intake)`

tofarr: Done

spec/controllers/state_file/intake_logins_controller_spec.rb
tofarr: This is lower level than I would like - but it does seem to be the easiest way of validating the session state

tofarr:
> assuming you did a global search for session[:state_file_intake] to replace all of them?

Yup.

jenny-heath (approved):
looks good!

PR #4197, opened 2024-02-08
More homepage updates

rickreyhsig:
https://www.pivotaltracker.com/story/show/186856017

Incorporating a bit more feedback from Lauren

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -329,12 +329,10 @@ ol.with-bullets {
 }
 .fyst-home-image {
   flex-shrink: 0;
-  width: 25%;
+  width: auto;
   height: auto;
   padding: 15px;
   padding-top: 45px;
-  max-width: 300px;
-  max-height: 280px;
   align-self: flex-start;
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -325,14 +325,22 @@ ol.with-bullets {
...
+  width: auto;
   height: auto;
   padding: 15px;
-  max-width: 300px;
-  max-height: 280px;
+  padding-top: 45px;
   align-self: flex-start;
 }
 
+/* Media query for screens smaller than 768px (typical mobile devices) */
```
tofarr, line 339: We have a media query above with "tablet-up" - should we use that instead?

tofarr (approved):

app/assets/stylesheets/application.scss
rickreyhsig: Good catch, thanks!

rickreyhsig: $tablet-up is configured to hide whenever the screen is 600px and lower so it looks off in the 6 hundred some range screen sizes

Screen Shot 2024-02-08 at 4 53 50 PM

PR #4200, opened 2024-02-08
Ops script for deleting pre launch submissions #187003357

tofarr:


lib/tasks/ops_240208.rake
```diff
@@ -0,0 +1,11 @@
+namespace :ops_240208 do
+  desc "Ops scripts for 240208"
+  task "delete_invalid_intakes" => :environment do
+    # EFile accidentally pointed at our production servers before launch, so there is some trash test
+    # data in here. Burn it!
```
embarnard, line 5: 🔥

embarnard: Is this cascading aka will it also delete things that are associated to these Intakes?

embarnard (approved):
LGTM!

lib/tasks/ops_240208.rake
tofarr: This really depends on whether the models have this set up - Destroy runs any before/around/after delete hooks on the model, as well as any cascade operations on the foreign key relation - I haven't seen that we have defined any of these though

PR #4212, opened 2024-02-12
Number Validations #187003599

tofarr:
The gyr_numericality validator does not work as expected - it filters out non number characters and retains the rest instead of rejecting the input.


app/forms/state_file/az_state_credits_form.rb
```diff
@@ -2,8 +2,8 @@ module StateFile
   class AzStateCreditsForm < QuestionsForm
     set_attributes_for :intake, :tribal_member, :tribal_wages, :armed_forces_member, :armed_forces_wages
 
-    validates :tribal_wages, presence: true, allow_blank: false, gyr_numericality: { only_integer: true, greater_than_or_equal_to: 1 }, if: -> { tribal_member == "yes" }
-    validates :armed_forces_wages, presence: true, allow_blank: false, gyr_numericality: { only_integer: true, greater_than_or_equal_to: 1 }, if: -> { armed_forces_member == "yes" }
+    validates :tribal_wages, presence: true, allow_blank: false, numericality: { only_integer: true, greater_than_or_equal_to: 1 }, if: -> { tribal_member == "yes" }
+    validates :armed_forces_wages, presence: true, allow_blank: false, numericality: { only_integer: true, greater_than_or_equal_to: 1 }, if: -> { armed_forces_member == "yes" }
```
rickreyhsig, line 6: Would there be a tiny spec we can add to outline the difference between the two validators @tofarr

tofarr: Good question - TBH I think the issue is that the gyr_numericality validator simply does not do what we want it to. For example:
| Input | gyr_numericality | numericality |
| ----- | ---------------- | ------------ |
| Z100Z | 100 | Error: Invalid input |

tofarr: Yeah - you're right we should check our validations

rickreyhsig (approved):

PR #4218, opened 2024-02-13
Header and footer changes

tahsinaislam:


app/views/state_file/state_file_pages/_footer.html.erb
```diff
@@ -15,14 +15,20 @@
...
+          </li>
 
           <li><%= link_to t(".faq"), state_faq_path(us_state: 'us') %></li>
 
-          <li><%= link_to t(".terms_of_service"), "" %></li>
+          <li>
+            <% if I18n.locale != :es %>
+              <%= link_to_spanish %>
+            <% end %>
+          </li>
```
embarnard, line 31: should we add English option if in Spanish locale like we do in the header?
```suggestion
           <% if I18n.locale != :en %>
              <%= link_to_english %>
            <% end %>
            </li>
```

embarnard (approved):
Looks good! Just one small comment

app/views/state_file/state_file_pages/_footer.html.erb
tahsinaislam: such a good catch! fixed it

PR #4220, opened 2024-02-13
Implement mailgun opt outs

jenny-heath:
what:
- add column to track unsubscribes
- add validation on notification email model that intake cannot be unsubscribed
- add unsubscribe and subscribe endpoints
- add unsubscribe link to notification mailer template

### DIFF
--- a/app/controllers/state_file/notifications_settings_controller.rb
+++ b/app/controllers/state_file/notifications_settings_controller.rb
@@ -1,0 +1,40 @@
+module StateFile
+  class NotificationsSettingsController < ApplicationController
+    layout "state_file"
+
+    def unsubscribe_email
+      matching_intakes = matching_intakes(params[:email_address])
+
+      if matching_intakes.present?
+        matching_intakes.each do |intake|
+          intake.update(unsubscribed_from_email: true)
+        end
+      else
+        flash[:alert] = "No record found"
+      end
+    end
+
+    def subscribe_email
+      matching_intakes = matching_intakes(params[:email_address])
+
+      if matching_intakes.present?
+        matching_intakes.each do |intake|
+          intake.update(unsubscribed_from_email: false)
+        end
+
+        flash[:notice] = I18n.t("state_file.notifications_settings.subscribe_email.flash")
+        render :unsubscribe_email
+      else
+        flash[:alert] = "No record found"
+      end
+    end
+
+    private
+
+    def matching_intakes(email_address)
+      return if email_address.blank?
+
+      StateFileAzIntake.where(email_address: email_address) + StateFileNyIntake.where(email_address: email_address)
+    end
+  end
+end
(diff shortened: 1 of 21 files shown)


app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -1,6 +1,7 @@
 class MailgunWebhooksController < ActionController::Base
   skip_before_action :verify_authenticity_token
   before_action :authenticate_mailgun_request
+  skip_before_action :authenticate_mailgun_request, only: :unsubscribe
```
jenny-heath, line 4: this line can be removed when the action is re-homed. see comment below

app/models/state_file_notification_email.rb
```diff
@@ -16,13 +16,22 @@ class StateFileNotificationEmail < ApplicationRecord
   validates_presence_of :to
   validates_presence_of :body
   validates_presence_of :subject
+  validate :not_unsubscribed
```
jenny-heath, line 19: one of the big challenges of this was thinking of a thorough way to intercept outgoing messages. i landed on having a validation because as long as all other outgoing emails for state file are of this type/model, it should work without having to add more unsubscribe-related code.

app/models/state_file_notification_email.rb
```diff
@@ -16,13 +16,22 @@ class StateFileNotificationEmail < ApplicationRecord
...
 
   after_create_commit :deliver
 
   private
 
+  def not_unsubscribed
+    matching_intakes = StateFileAzIntake.where(email_address: to, unsubscribed_from_email: true) +
+      StateFileNyIntake.where(email_address: to, unsubscribed_from_email: true)
+    if matching_intakes.present?
+      DatadogApi.increment("mailgun.state_file_notification_emails.not_sent_because_unsubscribed")
```
jenny-heath, line 29: thought this might be interesting to capture. afaict, we shouldn't see any sentry errors from this failure to validate but someone tell me if i'm wrong...

jenny-heath:
converting back to draft because this is in a seriously messed up state

embarnard (approved):
Let some questions, if you feel like its fine feel free to merge

app/controllers/state_file/notifications_settings_controller.rb
```diff
@@ -0,0 +1,40 @@
+module StateFile
+  class NotificationsSettingsController < ApplicationController
+    layout "state_file"
+
+    def unsubscribe_email
+      matching_intakes = matching_intakes(params[:email_address])
+
+      if matching_intakes.present?
+        matching_intakes.each do |intake|
+          intake.update(unsubscribed_from_email: true)
```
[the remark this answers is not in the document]
jenny-heath: what do you mean by opt them out? i don't think we have the same "opted in" columns on state file that we did on gyr/ctc unless i totally missed it and can't find it...or do you mean opted out via mailgun?

app/views/state_file/notification_mailer/user_message.html.erb
```diff
@@ -5,3 +5,4 @@
     <br />
   <% end %>
 <% end %>
+<%= auto_link(t(".unsubscribe", unsubscribe_link: @unsubscribe_link)) %>
```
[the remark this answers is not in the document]
jenny-heath: oh good point, i'll add one

PR #4221, opened 2024-02-13
Added a logging statement to see WHY a connection to an external service failed

tofarr:


mpidcock (approved):
I'm just going to gripe a little bit about ruby that it seems too magical to me that you can just log the whole `err` object and it just works. I suppose I should be happy about it, but instead I feel a discomfort about not being sure how it works without diving into some documentation about it.

but yeah, thanks for adding this!

tofarr:
By convention, errors override the `to_s` method to give a reasonable description of the error - the base implementation in `exception.rb` prints the class name or a message if one was defined. (The logger checks if the item is a string and if it is not, calls `to_s` before printing).

This is similar to python's `__repr__` funcitonality (Though I prefer that python explicitly calls it out with the dunder syntax)

PR #4223, opened 2024-02-14
1099g validation and flow updates #187002000

tofarr:
Now, when you go to the `unemployment` step, the default action is `index`:
image

... rather than `new`:
image

If there are no 1099G's defined, we redirect into new.

Also, we have loosened the validation on Zip Codes, because our database was missing some (Notably 10572 used in the Javier case). This was causing some weird behavior, as the zip code for the Payee was validated even if it was not specified in the form (Causing saves to fail)
image

### DIFF
--- a/app/controllers/state_file/questions/unemployment_controller.rb
+++ b/app/controllers/state_file/questions/unemployment_controller.rb
@@ -10,14 +10,18 @@ module StateFile
       end
 
       def self.navigation_actions
-        [:new, :index]
+        [:index, :new]
       end
 
       def index
         @state_file1099_gs = current_intake.state_file1099_gs
+        unless @state_file1099_gs.present?
+          redirect_to action: :new
+        end
       end
 
       def new
+        # redirect_to action: :index and return if current_intake.state_file1099_gs
         @state_file1099_g = current_intake.state_file1099_gs.build
       end
 
--- a/app/controllers/state_file/questions/unemployment_controller.rb
+++ b/app/controllers/state_file/questions/unemployment_controller.rb
@@ -21,7 +21,6 @@ module StateFile
       end
 
       def new
-        # redirect_to action: :index and return if current_intake.state_file1099_gs
         @state_file1099_g = current_intake.state_file1099_gs.build
       end
 
--- a/app/validators/zip_code_validator.rb
+++ b/app/validators/zip_code_validator.rb
@@ -2,7 +2,7 @@ class ZipCodeValidator < ActiveModel::EachValidator
   ZIP_CODE_REGEX = /\A\d{5}\z/
 
   def validate_each(record, attr_name, value)
-    unless value =~ ZIP_CODE_REGEX # && ZipCodes.has_key?(value)
+    unless value =~ ZIP_CODE_REGEX
       record.errors.add(attr_name, I18n.t("validators.zip"))
     end
   end
--- a/app/validators/zip_code_validator.rb
+++ b/app/validators/zip_code_validator.rb
@@ -2,7 +2,7 @@ class ZipCodeValidator < ActiveModel::EachValidator
   ZIP_CODE_REGEX = /\A\d{5}\z/
 
   def validate_each(record, attr_name, value)
-    unless value =~ ZIP_CODE_REGEX && ZipCodes.has_key?(value)
+    unless value =~ ZIP_CODE_REGEX # && ZipCodes.has_key?(value)
       record.errors.add(attr_name, I18n.t("validators.zip"))
     end
   end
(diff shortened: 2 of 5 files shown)


app/controllers/state_file/questions/unemployment_controller.rb
```diff
@@ -10,14 +10,18 @@ def self.show?(intake)
...
       def self.navigation_actions
-        [:new, :index]
+        [:index, :new]
       end
 
       def index
         @state_file1099_gs = current_intake.state_file1099_gs
+        unless @state_file1099_gs.present?
+          redirect_to action: :new
+        end
```
tofarr, line 20: If there is no 1099G's present, we show the form that asks users whether they want to create one

spec/validators/zip_code_validator_spec.rb
```diff
@@ -17,10 +17,15 @@
...
-  specify do
+  # For the moment we have excluded this spec because we not checking against known zip codes
+  xspecify do
     assert_invalid("99999")
   end
 
+  specify do
+    assert_valid("10752")
+  end
+
```
tofarr, line 28: This zip code is explicitly missing - there may be others

app/controllers/state_file/questions/unemployment_controller.rb
```diff
@@ -10,14 +10,18 @@ def self.show?(intake)
       end
 
       def self.navigation_actions
-        [:new, :index]
+        [:index, :new]
```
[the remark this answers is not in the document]
jenny-heath: i find this part of the code hard to wrap my head around but i think the option to have multiple actions is for those "loops" within the flow where we create n number of records. 

one way it's used is in the routes file: we have this line `next if controller_class.navigation_actions.length > 1` in the scoped navigation routes loop and we separately define routes for all the crud actions for things like 1099g's and dependents

app/validators/zip_code_validator.rb
```diff
@@ -2,7 +2,7 @@ class ZipCodeValidator < ActiveModel::EachValidator
   ZIP_CODE_REGEX = /\A\d{5}\z/
 
   def validate_each(record, attr_name, value)
-    unless value =~ ZIP_CODE_REGEX && ZipCodes.has_key?(value)
+    unless value =~ ZIP_CODE_REGEX # && ZipCodes.has_key?(value)
```
jenny-heath, line 5: i think it's okay to just remove this instead of commenting it out. this zip codes yml file dates all the way back to 2020 when we built an honest-to-god in-person vita site matcher where you could enter your zip code and we would find the nearest site based on its coordinates (the reason we have postgis not just postgres). probably was never a great idea to use as a validator long term

jenny-heath (commented):
oops forgot to leave this comment on approval:

the one thing i might like to see is a feature spec that covers common cases (client has no 1099g's & sees the new view, client has 1099g's & sees the index, backs up from a subsequent step and doesn't run into the new view again)? not sure it's totally necessary and the only harm i think it would do is slow down our test suite but i was thinking about how i'd like to see thorough acceptance testing on this and that made me think that the test-tests should also be thorough. happy to hear thoughts and not a blocking thing

tofarr:
> oops forgot to leave this comment on approval:
> 
> the one thing i might like to see is a feature spec that covers common cases (client has no 1099g's & sees the new view, client has 1099g's & sees the index, backs up from a subsequent step and doesn't run into the new view again)? not sure it's totally necessary and the only harm i think it would do is slow down our test suite but i was thinking about how i'd like to see thorough acceptance testing on this and that made me think that the test-tests should also be thorough. happy to hear thoughts and not a blocking thing

That is what the specs in spec/controllers/state_file/questions/unemployment_controller_spec.rb are - checking that we redirect if there are no 1099Gs, and that the navigation defaults to `index` rather than `new`

jenny-heath:
> > oops forgot to leave this comment on approval:
> > the one thing i might like to see is a feature spec that covers common cases (client has no 1099g's & sees the new view, client has 1099g's & sees the index, backs up from a subsequent step and doesn't run into the new view again)? not sure it's totally necessary and the only harm i think it would do is slow down our test suite but i was thinking about how i'd like to see thorough acceptance testing on this and that made me think that the test-tests should also be thorough. happy to hear thoughts and not a blocking thing
> 
> That is what the specs in spec/controllers/state_file/questions/unemployment_controller_spec.rb are - checking that we redirect if there are no 1099Gs, and that the navigation defaults to `index` rather than `new`

true, i guess it is a bit redundant. i think the thing i was having trouble feeling confidence in is that the first action *is* the default action and that it works the way we think it does. but as long as there's separate test of the multiple actions concept i think it's pretty thoroughly covered (and i think there is a feature spec with a dependents loop)

PR #4224, opened 2024-02-14
Session Timeout Flow #187040813

tofarr:
When your session times out, we redirect to the login page rather than the landing page:
image

I also updated code to use `StateFileBaseIntake::StateCodes` rather than enumerating out states individually in each case - so this will lower the bar on adding new States

### DIFF
--- a/app/controllers/state_file/questions/questions_controller.rb
+++ b/app/controllers/state_file/questions/questions_controller.rb
@@ -24,22 +24,21 @@ module StateFile
       end
 
       def question_navigator
-        case params[:us_state]
-        when 'az'
-          Navigation::StateFileAzQuestionNavigation
-        when 'ny'
-          Navigation::StateFileNyQuestionNavigation
+        @navigator ||= "Navigation::StateFile#{state_code.titleize}QuestionNavigation".constantize
+      end
+
+      def state_code
+        state_code_ = params['us_state']
+        unless StateFileBaseIntake::STATE_CODES.include?(state_code_)
+          raise StandardError state_code_
         end
+        state_code_
       end
 
       def redirect_if_no_intake
         unless current_intake.present?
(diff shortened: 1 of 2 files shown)


app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -24,22 +24,21 @@ def current_intake
       end
 
       def question_navigator
-        case params[:us_state]
-        when 'az'
-          Navigation::StateFileAzQuestionNavigation
-        when 'ny'
-          Navigation::StateFileNyQuestionNavigation
+        @navigator ||= "Navigation::StateFile#{state_code.titleize}QuestionNavigation".constantize
```
tofarr, line 27: Using this is easier than enumerating out states individually - especially as we add more states

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -24,22 +24,21 @@ def current_intake
...
+      end
+
+      def state_code
+        state_code_ = params['us_state']
+        unless StateFileBaseIntake::STATE_CODES.include?(state_code_)
+          raise StandardError state_code_
         end
+        state_code_
```
tofarr, line 35: This method takes in and validates the state code parameter (So we can constantize / send without fear of code injection)

rickreyhsig: Nice

rickreyhsig: Do you think just passing in the state code is enough? Maybe raising something like `raise StandardError, "State code `#{state_code_}` not supported!"` might make it easier to find the error when we see it in Sentry.

rickreyhsig (approved):
LGTM, left a non-blocking suggestion.

PR #4226, opened 2024-02-14
Log In Messages #187042699

tofarr:
Added some links to the landing pages.
image

Also - if you try to log into an account that does not exist you get a message in the UI (I think this is okay because we lock accounts):
image

One more thing: If you hit the login page and you are already logged in, you now still see the login page (Rather than being taken to the last step of your logged in intake). I figure this makes sense as visiting a login page implies intent to login, and allows us to support use cases where a user wants to switch profiles (For example, if older folks share a family computer and don't have different profiles set up because they don't know what that is.)

### DIFF
--- a/app/controllers/state_file/intake_logins_controller.rb
+++ b/app/controllers/state_file/intake_logins_controller.rb
@@ -1,7 +1,6 @@
 module StateFile
   class IntakeLoginsController < Portal::ClientLoginsController
     helper_method :prev_path, :illustration_path
-    before_action :redirect_to_data_review_if_intake_authenticated
     layout "state_file/question"
 
     def new
@@ -93,19 +92,6 @@ module StateFile
       when "az" then :statefile_az
       when "ny" then :statefile_ny
       when "us" then :statefile
-      end
-    end
-
-    def redirect_to_data_review_if_intake_authenticated
-      intake = current_state_file_az_intake || current_state_file_ny_intake
-      if intake.present?
-        # Redirect to last step
-        controller = intake.controller_for_current_step
-        to_path = controller.to_path_helper(
-          action: controller.navigation_actions.first,
-          us_state: intake.state_code
-        )
-        redirect_to to_path
       end
     end
 
--- a/app/controllers/state_file/intake_logins_controller.rb
+++ b/app/controllers/state_file/intake_logins_controller.rb
@@ -14,6 +14,13 @@ module StateFile
 
     def create
       @contact_method = params[:contact_method] unless @contact_method.present?
+      value = params[:state_file_request_intake_login_form][@contact_method]
+      @records = client_login_service.intake_classes.map { |intake_class| intake_class.where({@contact_method => value}) }.flatten
+      if @records.blank?
+        flash[:alert] = I18n.t("state_file.intake_logins.new.#{@contact_method}.not_found")
+        redirect_to action: :new, contact_method: @contact_method
+        return
+      end
       super
     end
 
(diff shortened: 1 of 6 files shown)


app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -1,7 +1,6 @@
 module StateFile
   class IntakeLoginsController < Portal::ClientLoginsController
     helper_method :prev_path, :illustration_path
-    before_action :redirect_to_data_review_if_intake_authenticated
```
tofarr, line 4: We remove this because it means the sign in page is skipped if a user is already logged in (By for example visiting the landing page in the last 15 minutes)

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -14,6 +13,15 @@ def new
...
       @contact_method = params[:contact_method] unless @contact_method.present?
+      @form = request_login_form_class.new(request_client_login_params)
+      if @form.valid?
+        intake_classes = client_login_service.intake_classes
+        @records = intake_classes.map { |intake_class| @form.filter_records(intake_class) }.flatten
+        if @records.blank?
+          flash[:alert] = I18n.t("state_file.intake_logins.new.#{@contact_method}.not_found")
+          render :new and return
+        end
+      end
```
tofarr, line 24: If the form is valid, we check that there is an existing intake. I figured it made sense to do DB lookups in a controller rather than a validator - please LMK if you disagree.

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -87,19 +95,6 @@ def service_type
...
-        # Redirect to last step
-        controller = intake.controller_for_current_step
-        to_path = controller.to_path_helper(
-          action: controller.navigation_actions.first,
-          us_state: intake.state_code
-        )
-        redirect_to to_path
-      end
-    end
-
```
tofarr, line 102: I love when code goes away
image

app/views/state_file/questions/landing_page/edit.html.erb
```diff
@@ -23,5 +23,10 @@
     <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
       <%= f.submit t("general.get_started"), class: "button button--primary button--wide", id: "firstCta" %>
     <% end %>
+
+    <h2 class="h2">
+      <%= t(".already_started_html", sign_in_url: StateFile::StateFilePagesController.to_path_helper(action: :login_options, us_state: params[:us_state])) %>
+    </h2>
+
```
tofarr, line 30: image

spec/controllers/state_file/intake_logins_controller_spec.rb
```diff
@@ -5,10 +5,11 @@
     create(
       :state_file_az_intake,
       email_address: "client@example.com",
-      sms_phone_number: "+15105551234",
+      phone_number: "+15105551234",
```
tofarr, line 8: This was a typo!

spec/controllers/state_file/intake_logins_controller_spec.rb
```diff
@@ -5,10 +5,11 @@
...
       hashed_ssn: "hashed_ssn"
     )
   end
+  before { intake }
```
tofarr, line 12: This means the intake actually gets created in the db

mpidcock (approved):

PR #4228, opened 2024-02-15
Send message after state-file efile transition to accepted

embarnard:


app/services/state_file/messaging_service.rb
```diff
@@ -1,26 +1,69 @@
...
+        sent_message = StateFileNotificationEmail.create!(
+          data_source: intake,
+          to: intake.email_address,
+          body: @message_instance.email_body(locale: locale, **email_args),
+          subject: @message_instance.email_subject(locale: @locale, **email_args),
         )
+        sent_messages << sent_message if sent_message.present?
+      end
     end
+
```
rickreyhsig, line 46: Do you mean to leave this comment?

app/services/state_file/after_transition_messaging_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      messaging_service = StateFile::MessagingService.new(
+        intake: @intake,
+        message: message,
+        body_args: body_args)
+
+      messaging_service.send_message
+    end
+
```
rickreyhsig, line 26: Ditto

embarnard: yes @rickreyhsig was gonna add this in for the subsequent story

tofarr (approved):

app/jobs/after_transition_tasks_for_rejected_return_job.rb
```diff
@@ -2,7 +2,7 @@ class AfterTransitionTasksForRejectedReturnJob < ApplicationJob
   def perform(submission, transition)
     transition ||= submission.last_transition
 
-    # submission.tax_return&.transition_to(:file_rejected) #ctc only
+    submission.tax_return&.transition_to(:file_rejected) if submission.is_for_federal_filing?
```
mpidcock, line 5: Is this going to also affect gyr?

embarnard: no, GYR doesn't do efiling -- this would only impact CTC. I had commented this out a few weeks ago but decided to add this ctc check instead. We aren't using this code rn but unclear if we should keep it all in tack

app/jobs/after_transition_tasks_for_rejected_return_job.rb
```diff
@@ -13,20 +13,24 @@ def perform(submission, transition)
...
-      #     message: message_class,
-      #     locale: submission.client.intake.locale
-      #   )
-      # end
-    end
 
-    EfileSubmissionStateMachine.send_mixpanel_event(submission, "state_file_efile_return_rejected")
+      message_class = message_class_for_state(submission.current_state)
+
+      if submission.is_for_federal_filing?
```
mpidcock, line 22: same here, `is_for_state_filing?` ?

embarnard: @mpidcock oh oops saw Tim's approval so thought it was good to merge, but yes same case here

app/services/state_file/messaging_service.rb
```diff
@@ -1,26 +1,69 @@
...
+    
+    # def send_sms
+    #   return unless Flipper.enabled?(:sms_notifications)
+    # end
 
-    def self.send_notification(intake:, message:, body_args: {})
-      # TODO: Figure out locale from intake somehow (though we don't currently capture preferred language anywhere)
-      locale = nil
+    def base_args
+      first_name = intake.primary_first_name ? intake.primary_first_name.split(/ |\_/).map(&:capitalize).join(" ") : ""
```
mpidcock, line 52: @embarnard I don't understand this part, isn't primary_first_name already a first name that is provided during intake?

app/state_machines/efile_submission_state_machine.rb
```diff
@@ -150,18 +155,19 @@ class EfileSubmissionStateMachine
         recovery_rebate_credit: [benefits.eip1_amount, benefits.eip2_amount].compact.sum,
         third_stimulus_amount: benefits.eip3_amount,
       })
+    elsif submission.is_for_state_filing?
+      StateFile::AfterTransitionMessagingService.new(submission.data_source).send_efile_submission_accepted_message
+      send_mixpanel_event(submission, "fyst_efile_return_accepted")
```
mpidcock, line 160: @embarnard Minor note, but I see that EfileSubmissionStateMachine. qualifier is missing here compared to line 134. Also the mixpanel events are inconsistent about using fyst vs state_file

app/services/state_file/messaging_service.rb
embarnard: this is just for formatting what the client will see in their message. Many of them start with "Hi Name-of-client" but if we save primary_first_name as "monA lISa" we want it to display in the message as "Mona Lisa"

app/state_machines/efile_submission_state_machine.rb
rickreyhsig: I think we should keep the mixpanel event names consistent for sure.  Now, fyst and statefile makes sense, but this is adding a bit of complexity for posterity.  We should probably keep using statefile as that is what the codebase does.  Unless we want to make a separate PR to update all statefile references to fyst.

PR #4235, opened 2024-02-16
Fix Debugging Info for Gabriel

tofarr:
This dropped off when we swapped over to using devise for all session management:
image

### DIFF
--- a/app/controllers/state_file/state_file_pages_controller.rb
+++ b/app/controllers/state_file/state_file_pages_controller.rb
@@ -41,5 +41,9 @@ module StateFile
       uri.query = { authorizationCode: key }.to_query
       uri.to_s
     end
+
+    def current_intake
+      @current_intake ||= current_state_file_az_intake || current_state_file_ny_intake
+    end
   end
 end
--- a/app/controllers/state_file/state_file_pages_controller.rb
+++ b/app/controllers/state_file/state_file_pages_controller.rb
@@ -43,7 +43,12 @@ module StateFile
     end
 
     def current_intake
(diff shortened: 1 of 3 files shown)


app/controllers/state_file/state_file_pages_controller.rb
```diff
@@ -41,5 +41,9 @@ def transfer_url(key, redirect_url)
       uri.query = { authorizationCode: key }.to_query
       uri.to_s
     end
+
+    def current_intake
+      @current_intake ||= current_state_file_az_intake || current_state_file_ny_intake
+    end
```
tofarr, line 47: A current intake method already exists here from the application controller - I figured it should be customized for statefile

app/views/state_file/state_file_pages/about_page.html.erb
```diff
@@ -19,9 +19,9 @@
 <section>
   <%= link_to "Start Test AZ", Navigation::StateFileAzQuestionNavigation::FLOW.first.to_path_helper(us_state: "az"), class: "button" %>
   <%= link_to "Start Test NY", Navigation::StateFileNyQuestionNavigation::FLOW.first.to_path_helper(us_state: "ny"), class: "button" %>
-  <% if session[:state_file_intake] %>
+  <% if current_intake %>
     <div class="with-padding-med">
-      Your session has this intake: <b><%= session[:state_file_intake] %></b>
+      Your session has this intake: <b><%=current_intake.state_code %> <%= current_intake.id %></b>
```
tofarr, line 24: I figured the state code was useful here too - to make sure the intake is the one we expect

mpidcock (approved):
🍰

PR #4241, opened 2024-02-16
Job that sends notification messages to clients

embarnard:


app/jobs/send_accepted_rejected_notification_job.rb
```diff
@@ -0,0 +1,32 @@
+class SendAcceptedRejectedNotificationJob < ApplicationJob
```
embarnard, line 1: this job is really just to send accepted messages to 6ish clients on demo. Since its only a few people I was just planning to run it in the prod console with each individual id. 

With our messaging tracking logic, even if we run this job for someone that received this message it won't run. 

I plan on removing this job once we run it as not to junk up the jobs folder

app/jobs/send_accepted_rejected_notification_job.rb
```diff
@@ -0,0 +1,32 @@
...
+  def perform(submission_ids)
+    submission_ids.each do |id|
+      send_notification(id)
+    end
+  end
+
+  def send_notification(efile_submission_id)
+    submission = EfileSubmission.for_state_filing.find(efile_submission_id)
+    if submission.nil?
+      puts "*****Error: No efile submission found with id #{efile_submission_id}"
```
tofarr, line 11: Should we use the rails logger here rather than puts ?

tofarr (approved):

app/jobs/send_accepted_rejected_notification_job.rb
embarnard: @tofarr ooo yes something like `Rails.logger.info "*****Error: No efile submission found with id #{efile_submission_id}"`?

PR #4250, opened 2024-02-20
URLs fix for production issues #187080649

tofarr:
This PR should:
* Fix the issue where an intake can be broken by hitting the wrong url.
* Fix the broken `/pending-federal-return` page
* Make the `/pending-federal-return` and `/canceled-data-transfer` state specific matching what we actually sent to directfile:
* Add a temporary redirect so that if direct file don't update their URL, users still get the correct message
image

You can test these by going to:
* `/en/ny/questions/pending-federal-return`
* `/en/ny/questions/canceled-data-transfer`

### DIFF
--- a/app/controllers/state_file/questions/canceled_data_transfer_controller.rb
+++ b/app/controllers/state_file/questions/canceled_data_transfer_controller.rb
@@ -3,6 +3,7 @@ module StateFile
     class CanceledDataTransferController < QuestionsController
       include IrsDataTransferLinksConcern
       before_action :redirect_if_no_us_state_in_params
+      skip_before_action :set_current_step
 
       def self.show?(_intake)
         false
--- a/app/controllers/state_file/questions/canceled_data_transfer_controller.rb
+++ b/app/controllers/state_file/questions/canceled_data_transfer_controller.rb
@@ -25,11 +25,6 @@ module StateFile
 
       private
 
-      # this page sometimes doesn't have us_state in the params so can't use the default current_intake
-      def current_intake
-        GlobalID.find(session[:state_file_intake])
-      end
-
       # in order to give the IRS team one cancel link, this controller will redirect to the path with the us_state
       # based on the intake in the session; part of the reason for this is so we can leverage the default implementation
       # of
--- a/app/controllers/state_file/questions/pending_federal_return_controller.rb
+++ b/app/controllers/state_file/questions/pending_federal_return_controller.rb
@@ -1,6 +1,8 @@
 module StateFile
   module Questions
     class PendingFederalReturnController < QuestionsController
+      skip_before_action :set_current_step
+
       def prev_path
         nil
       end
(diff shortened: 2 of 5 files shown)


app/controllers/state_file/questions/canceled_data_transfer_controller.rb
```diff
@@ -25,11 +25,6 @@ def illustration_path
 
       private
 
-      # this page sometimes doesn't have us_state in the params so can't use the default current_intake
-      def current_intake
-        GlobalID.find(session[:state_file_intake])
-      end
-
```
tofarr, line 32: This was wrong - intakes are are NEVER stored without devise now

app/controllers/state_file/questions/pending_federal_return_controller.rb
```diff
@@ -4,13 +4,6 @@ class PendingFederalReturnController < QuestionsController
       def prev_path
         nil
       end
-
-      private
-
-      # edit page has no us_state in the params so default implementation of current_intake doesn't work
-      def current_intake
-        GlobalID.find(session[:state_file_intake])
-      end
```
tofarr, line 13: Same as above

app/controllers/state_file/questions/return_status_controller.rb
```diff
@@ -117,6 +119,21 @@ def reject_description
 
       def card_postscript; end
 
+      def redirect_if_no_submission
+        if current_intake.efile_submissions.empty?
+          redirect_to StateFile::Questions::InitiateDataTransferController.to_path_helper(us_state: state_code)
+        end
+      end
```
tofarr, line 126: Without this, an end use can break their whole intake by simply visiting the /return-status URL before it is ready.

app/views/state_file/state_file_pages/fake_direct_file_transfer_page.html.erb
```diff
@@ -36,15 +36,15 @@
   <%=
     link_to(
       "Cancel",
-      state_file_questions_canceled_data_transfer_path,
+      StateFile::Questions::CanceledDataTransferController.to_path_helper(us_state: current_intake.state_code),
```
tofarr, line 39: I had to change the format of this because these controllers are mapped differently

config/routes.rb
```diff
@@ -549,8 +549,6 @@ def scoped_navigation_routes(context, navigation)
         namespace :questions do
           get "show_xml", to: "confirmation#show_xml"
           get "explain_calculations", to: "confirmation#explain_calculations"
-          get "pending_federal_return", to: "pending_federal_return#edit"
-          get "canceled_data_transfer", to: "canceled_data_transfer#edit"
```
mpidcock, line 553: Why are these removed?

mpidcock (approved):

config/routes.rb
mpidcock: A: because this section is not state specific, the addition below is to "replace" these lines

PR #4255, opened 2024-02-21
Remove Detectify file

embarnard:


public/a942d4f1c235fde42bdf1751cff43ca3.txt
```diff
@@ -1 +0,0 @@
-detectify
```
mpidcock, line 1: What does this do? I would approve, but I don't understand

embarnard: it was just used to validate that owned the domain for detectify @mpidcock, now that it is validated on detectify we can remove it

PR #4256, opened 2024-02-22
AZ state credit page logic

tahsinaislam:


app/forms/state_file/az_state_credits_form.rb
```diff
@@ -5,11 +5,26 @@ class AzStateCreditsForm < QuestionsForm
...
       attributes_to_save = attributes_for(:intake)
       attributes_to_save[:tribal_wages] = nil if tribal_member == "no"
       attributes_to_save[:armed_forces_wages] = nil if armed_forces_member == "no"
       @intake.update(attributes_to_save)
     end
+
+    private
+
+    def exceeds_1040_amount
+      amount_limit = @intake.direct_file_data.fed_wages_salaries_tips.to_i
```
rickreyhsig, line 22: I think this field might already be an integer, so I'm not sure `.to_i` is necessary, but definitely doesn't break anything.

rickreyhsig (approved):

app/forms/state_file/az_state_credits_form.rb
tahsinaislam: That's a good catch I'll fix this

PR #4257, opened 2024-02-22
AZ Excise Credit question updates

rickreyhsig:
https://www.pivotaltracker.com/story/show/187005390


tahsinaislam (approved):

mpidcock:
Looks like this is waiting on acceptance testing but needs to be updated in pivotal to reflect that

PR #4261, opened 2024-02-23
Add hub errors pages

embarnard:


app/jobs/update_efile_errors_job.rb
```diff
@@ -0,0 +1,15 @@
+class UpdateEfileErrorsJob < ApplicationJob
+  def perform
+    ctc_errors = EfileError.where('created_at < ?', Date.new(2024))
+    state_file_errors = EfileError.where('created_at > ?', Date.new(2024))
+
+    ctc_errors.update_all(product_type: :ctc)
+    state_file_errors.update_all(expose: false)
```
embarnard, line 7: I think changing to expose false should actually happen before we deploy, there are only currently 4 state_file errors could do this manually even in the hub

app/lib/ability.rb
```diff
@@ -20,6 +20,9 @@ def initialize(user)
         cannot :manage, StateFileDependent
         cannot :manage, StateId
         cannot :manage, EfileSubmission, id: EfileSubmission.for_state_filing.pluck(:id)
+        cannot :manage, EfileError do |error|
```
embarnard, line 23: restrictions on who can see and make edits to these errors in the abilities file since it can impact what the client sees on the return-status page

app/controllers/hub/efile_errors_controller.rb
```diff
@@ -1,10 +1,11 @@
 module Hub
   class EfileErrorsController < Hub::BaseController
+    before_action :require_admin
     load_and_authorize_resource
     layout "hub"
 
     def index
-      @efile_errors = @efile_errors.order(:source, :code)
+      @efile_errors = EfileError.where(product_type: :ctc).order(:source, :code)
```
embarnard, line 8: for some reason because of changes in the abilities file it doesn't load anything into @efile_errors but @efile_error does; added require_admin at the top of here just for some added security could even do require state_file_admin if we want to be extra sure (this is a deprecated action anyways)

app/controllers/state_file/questions/return_status_controller.rb
```diff
@@ -23,26 +19,6 @@ def prev_path
 
       private
 
-      def title
```
embarnard, line 26: moved some of these into the view since we should try to keep text there if we can

app/jobs/update_efile_errors_job.rb
```diff
@@ -0,0 +1,17 @@
+class UpdateEfileErrorsJob < ApplicationJob
+  def perform
+    ctc_errors = EfileError.where('created_at < ?', Date.new(2024))
+    state_file_errors = EfileError.where('created_at > ?', Date.new(2024))
+
+    ctc_errors.update_all(product_type: :ctc)
+
+    state_file_errors.update_all(product_type: :state_file)
+    # state_file_errors.update_all(expose: false)
```
embarnard, line 9: we should change expose to false before deploy

db/migrate/20240222202500_add_product_type_column_to_efile_errors.rb
```diff
@@ -0,0 +1,5 @@
+class AddProductTypeColumnToEfileErrors < ActiveRecord::Migration[7.1]
+  def change
+    add_column :efile_errors, :product_type, :integer, default: 0, null: false
```
rickreyhsig, line 3: It seems you are using product_type for ctc/statefile where other places in our codebase we refer to this as service_type.  I would propose to maintain consistency if we can, unless you are going for something different here.

app/jobs/update_efile_errors_job.rb
rickreyhsig: What's the idea of the expose attribute here? Not sure I understand.  Expose on the hub as opposed to keeping it in the DB?

app/views/state_file/questions/return_status/_pending.html.erb
```diff
@@ -5,7 +5,7 @@
 </div>
 
 <h1 class="h2">
-  <%= @title %>
+  <%= t("state_file.questions.return_status.pending.title", state_name: States.name_for_key(params[:us_state].upcase)) %>
```
rickreyhsig, line 8: I feel like we should be able to use the state_name method from the StateFileControllerConcern or some other place so it's a little less verbose than `States.name_for_key(params[:us_state].upcase`, but this is your call.

app/jobs/update_efile_errors_job.rb
embarnard: @rickreyhsig Expose on the efile error means that we will expose that error to the client. In state-file's case if expose is true then we are exposing the efile error description and resolution (what's next text) to the client on the return-status page.

app/views/state_file/questions/return_status/_pending.html.erb
embarnard: I guess I'm wary to do this since `States.name_for_key` is already a part of a the helper `states.rb`, but I agree it is kind of verbose...

rickreyhsig (approved):

app/jobs/update_efile_errors_job.rb
embarnard: this is done

PR #4262, opened 2024-02-23
Updated truncation #187105238

tofarr:
So it turns out that truncation in ruby when using [x..y] is by inclusive (Unlike python / java /javascript substring / slice methods which are exclusive on the end index). Funnily enough, I saw the other parts of the code using `truncate` when writing this and thought about matching for a moment before saying to myself "Nah - this is shorter".

I guess this is a lesson for me in sticking to what is already there.


mpidcock (approved):
nice

rickreyhsig (approved):

PR #4264, opened 2024-02-23
Progress bar #186624823

tofarr:
Progress Bar:
image


app/assets/stylesheets/_state-file.scss
```diff
@@ -80,6 +80,30 @@
...
+      background: $color-state-file-primary;
+    }
+  }
+
+  p.progress-text {
+    color: $color-black;
+    font-size: 1.3rem;
+    margin-bottom: 2rem;
+  }
+
```
tofarr, line 106: Styling for progress steps

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -26,6 +26,7 @@ def current_intake
       def question_navigator
         @navigator ||= "Navigation::StateFile#{state_code.titleize}QuestionNavigation".constantize
       end
+      helper_method :question_navigator
```
tofarr, line 29: I now need this in the views (As it does a bit more than before)

app/lib/navigation/state_file_base_question_navigation_mixin.rb
```diff
@@ -0,0 +1,22 @@
+module Navigation
+  module StateFileBaseQuestionNavigationMixin
```
tofarr, line 2: This mixin is used to add static methods to the navigation classes

app/lib/navigation/state_file_base_question_navigation_mixin.rb
```diff
@@ -0,0 +1,22 @@
...
+      def sections
+        const_get(:SECTIONS)
+      end
+
+      def get_section(controller)
+        sections.detect { |section| section.controllers.select { |c| c == controller }}
+      end
+
+      def get_progress(controller)
+        sections.lazy.map { |s| s.get_progress(controller) }.detect(&:present?)
```
tofarr, line 18: Using lazy here means we stop when we detect a value

PR #4266, opened 2024-02-23
Hide show xml button from intake flow in demo env; add download pdf button to hub view of state file efile submissions

jenny-heath:


mpidcock (approved):
I think it's a little bit odd that we are hiding the pdf also, but that's what the story says. Staging is also "public" - should we hide it there?

jenny-heath:
retrying heroku deploy

app/controllers/application_controller.rb
```diff
@@ -421,6 +421,11 @@ def acts_like_production?
   end
   helper_method :acts_like_production?
 
+  def show_xml?
+    !(Rails.env.demo? || Rails.env.staging?)
```
mpidcock, line 425: Maybe make this an allow list, like only heroku and local. We're now treating it like sensitive data so we should be specific about what envs it should be in. Production should also be excluded anyway.

mpidcock (changes requested):
Replacing `acts_like_production?` with `show_xml?` means the xml button will be introduced to prod, which I don't think we want

app/controllers/application_controller.rb
jenny-heath: oh true whitelist is def better

PR #4267, opened 2024-02-26
Rejected notifications hub #187117719

tofarr:
Notifications of rejection are now manually initialized from the hub.
image

When a notification is in the "rejected" state, a "Notify of Rejection" button is present
image

When clicked, the filer is notified of the rejection

I also added in a link so admins can see the S3 key for submissions, so users with access can grab the full bundle if needed
image

### DIFF
--- a/app/state_machines/efile_submission_state_machine.rb
+++ b/app/state_machines/efile_submission_state_machine.rb
@@ -139,7 +139,7 @@ class EfileSubmissionStateMachine
     end
   end
 
-  before_transition(to: :notified_of_rejection) do |submission, transition|
+  before_transition(to: :notified_of_rejection) do |submission|
     if submission.is_for_state_filing?
       StateFile::AfterTransitionMessagingService.new(submission).send_efile_submission_rejected_message
     end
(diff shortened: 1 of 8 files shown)


app/state_machines/efile_submission_state_machine.rb
```diff
@@ -133,11 +135,16 @@ class EfileSubmissionStateMachine
   after_transition(to: :rejected, after_commit: true) do |submission, transition|
     AfterTransitionTasksForRejectedReturnJob.perform_later(submission, transition)
     if submission.is_for_state_filing?
-      StateFile::AfterTransitionMessagingService.new(submission).send_efile_submission_rejected_message
       EfileSubmissionStateMachine.send_mixpanel_event(submission, "state_file_efile_return_rejected")
     end
   end
 
+  before_transition(to: :notified_of_rejection) do |submission, transition|
```
embarnard, line 142: since transition is never used in this block
```suggestion
  before_transition(to: :notified_of_rejection) do |submission|
```

embarnard (approved):
Nice, looks great! I love that its part of the filter so we can see exactly who needs to be notified

PR #4268, opened 2024-02-26
Add logic for MFJ and deceased spouse case

embarnard:
When married filing jointing and the spouse is deceased we:
* remove the spouse signature option from the esign page
* record date-of-death in the XML
* add date-of-death and 'filing as surviving spouse' in PDF


app/lib/pdfs/az140-TY2023.pdf
embarnard, line 1: Had to update the form with one that has an editable field where the spouse signature is

app/models/state_file_base_intake.rb
```diff
@@ -122,6 +122,14 @@ def ask_spouse_name?
     filing_status_mfj?
   end
 
+  def ask_spouse_esign?
+    filing_status_mfj? && !spouse_deceased?
+  end
+
+  def spouse_deceased?
+    direct_file_data.surviving_spouse?
+  end
```
tofarr, line 131: Not a blocker:
Should we use unified terminology? (ie: `surviving_spouse?` vs `spouse_deceased?`)
Personally, I find `spouse_deceased?` to be clearer.

tofarr (approved):

PR #4269, opened 2024-02-26
State wildcard showing on eligibility page

rickreyhsig:
https://www.pivotaltracker.com/story/show/187101341

Screen Shot 2024-02-26 at 5 48 56 PM

### DIFF
--- a/app/views/state_file/questions/eligible/edit.html.erb
+++ b/app/views/state_file/questions/eligible/edit.html.erb
@@ -33,7 +33,7 @@
       <% end %>
       <div class="with-top-separator">
         <p class="text--bold"><%=t('.want_to_claim_heading') %></p>
-        <p><%=t('.want_to_claim_vita') %></p>
+        <p><%=t('.want_to_claim_vita', state_name: States.name_for_key(params[:us_state].upcase)) %></p>
         <p><%=t('.want_to_claim_learn_more_html', link: @faq_other_options_link) %></p>
       </div>
     </div>


app/views/state_file/questions/eligible/edit.html.erb
```diff
@@ -33,7 +33,7 @@
       <% end %>
       <div class="with-top-separator">
         <p class="text--bold"><%=t('.want_to_claim_heading') %></p>
-        <p><%=t('.want_to_claim_vita') %></p>
+        <p><%=t('.want_to_claim_vita', state_name: States.name_for_key(params[:us_state].upcase)) %></p>
```
rickreyhsig, line 36: The translation file already the right thing:

https://github.com/codeforamerica/vita-min/blob/1d2d3ac42b0d4eecfc7b5d16290df1a3242a471d/config/locales/en.yml#L2178

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -35,6 +35,14 @@ def state_code
         state_code_
       end
 
+      def state_name
```
mpidcock, line 38: Are there other possible applications of this method now that it's here?

mpidcock (approved):

app/controllers/state_file/questions/questions_controller.rb
mpidcock: Factoring this out into a method doesn't seem worth it to me if it is only used in one place

rickreyhsig: Yeah, the params are used everywhere in the views.  I wanted to update all those places using this.

rickreyhsig: I will add another PR making those updates.

rickreyhsig: https://github.com/codeforamerica/vita-min/pull/4272

PR #4272, opened 2024-02-27
Use state_name method from questions controller

rickreyhsig:
https://www.pivotaltracker.com/story/show/187101341


mpidcock (approved):
Nice thanks

PR #4279, opened 2024-02-28
Add fields to state file analytics table

tahsinaislam:


jenny-heath (approved):
2 small things:
1. if we're going to get these values from the calculator (which seems reasonable enough, idk how to get them off the submission bundle easily lol), i see why you had to add some public methods, BUT there's something that makes me nervous about adding these generic-seeming methods to `az140.rb` and `it201.rb` that will actually only be read from the analytics model. since they're not used anywhere else it just makes them feel out of place and potentially confusing later on. i wonder if it would be better to add just one public method to each calculator called something like `analytics_attrs` that returns the hash of attributes you're currently constructing in `calculated_attrs` on the analytics model. thoughts?
2. i think/assume the values being tested in `spec/models/state_file_analytics_spec.rb` came from the default direct file data from the factory(?) but since transcribing these lines from one place to another is a little confusing i wonder if we could get a program person to provide us with a good test example to use here? i.e. they could just pick one of the DF fixtures we already have in the codebase and tell you what the expected analytics values are.

jenny-heath (approved):
wúnderbar

PR #4294, opened 2024-03-01
W2 page index update #187136591

jenny-heath:
continue disabled on index until all valid
image

edit view
image
image

continue enabled when all valid
image


app/models/intake.rb
```diff
@@ -306,8 +306,6 @@
 #
 #  fk_rails_...  (client_id => clients.id)
 #  fk_rails_...  (matching_previous_year_intake_id => intakes.id)
-#  fk_rails_...  (primary_drivers_license_id => drivers_licenses.id)
-#  fk_rails_...  (spouse_drivers_license_id => drivers_licenses.id)
```
mpidcock, line 310: Is this related?

PR #4300, opened 2024-03-04
Filter logic fix #187136591

tofarr:
Fix for issue found by Gabriel. The NY_W2 index page should only show suspicious W2s. It was leaving in one that was fine, and filtering out one that was sus:
image
This is because we were incorrectly matching by index.

I fixed the issue and added a spec to verify

### DIFF
--- a/app/models/state_file_w2.rb
+++ b/app/models/state_file_w2.rb
@@ -50,10 +50,10 @@ class StateFileW2 < ApplicationRecord
     if local_income_tax_amt.present? && local_income_tax_amt.positive? && local_wages_and_tips_amt <= 0
       errors.add(:local_income_tax_amt, I18n.t("state_file.questions.ny_w2.edit.local_wages_and_tips_amt_error"))
     end
-    if state_wages_amt.present? && state_income_tax_amt > state_wages_amt
+    if state_income_tax_amt > state_wages_amt
       errors.add(:state_income_tax_amt, I18n.t("state_file.questions.ny_w2.edit.state_income_tax_amt_error"))
     end
-    if local_wages_and_tips_amt.present? && local_income_tax_amt > local_wages_and_tips_amt
+    if local_income_tax_amt > local_wages_and_tips_amt
       errors.add(:local_income_tax_amt, I18n.t("state_file.questions.ny_w2.edit.local_income_tax_amt_error"))
     end
   end
--- a/app/models/state_file_w2.rb
+++ b/app/models/state_file_w2.rb
@@ -26,19 +26,35 @@ class StateFileW2 < ApplicationRecord
 
   validates :employer_state_id_num, format: { with: /\A(\d{0,17})\z/ }
   validates :state_wages_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { state_wages_amt.present? }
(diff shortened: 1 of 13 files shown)


jenny-heath (approved):

app/models/state_file_w2.rb
```diff
@@ -25,13 +25,43 @@ class StateFileW2 < ApplicationRecord
...
+  validates :state_wages_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { state_wages_amt.present? }
+  validates :state_income_tax_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { state_income_tax_amt.present? }
+  validates :local_wages_and_tips_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { local_wages_and_tips_amt.present? }
+  validates :local_income_tax_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { local_income_tax_amt.present? }
+  validates :locality_nm, presence: { message: -> (_object, _data) { I18n.t('state_file.questions.ny_w2.edit.locality_nm_missing_error') } }, if: -> { local_wages_and_tips_amt.present? && local_wages_and_tips_amt.positive? }
   validates :employer_state_id_num, presence: true, if: -> { state_wages_amt.present? && state_wages_amt.positive? }
+  validates :locality_nm, format: { with: /\A[a-zA-Z]{1}([A-Za-z\-\s']{0,137})\z/ }, if: -> { locality_nm.present? }
```
jenny-heath, line 34: just to make sure i understand this regex: it's checking that the first character is a letter and then the rest can be either letters or whitespace up to a length of 137?

config/locales/en.yml
```diff
@@ -2349,8 +2349,15 @@ en:
...
+          local_income_tax_amt_error: Cannot be greater than local wages and tips.
+          local_wages_and_tips_amt_error: Please enter local wages and tips
+          locality_nm_error: Please type in “NYC” if the locality name on your W-2 is New York City or any of the five boroughs.
+          locality_nm_missing_error: Please select locality name
+          state_income_tax_amt_error: Cannot be greater than State wages and tips.
+          state_wages_amt_error: Please enter State wages and tips
+          w2s_error: Please update all of the W-2s listed above to continue.
         index:
-          title: We need a little more information about the state and local tax information on your W-2s
+          title: Please provide state tax information for the following W-2s. Do not worry if you have multiple W-2s and not all of them are listed below.
```
jenny-heath, line 2360: should we update the spanish too?

app/models/state_file_w2.rb
```diff
@@ -25,13 +25,43 @@ class StateFileW2 < ApplicationRecord
...
+  validate :validate_tax_amts
+  validate :locality_nm_validation
```
jenny-heath, line 36: add a spec for this? or maybe there is one and i'm missing it...

jenny-heath (approved):

app/models/state_file_w2.rb
tofarr: Yes

app/models/state_file_w2.rb
tofarr: Added spec for locality_nm

PR #4309, opened 2024-03-05
More validation fixes #187136591

tofarr:


jenny-heath (approved):
we should add tests later but looks good!

PR #4311, opened 2024-03-06
GYR email opt out

tahsinaislam:


app/controllers/hub/outgoing_emails_controller.rb
```diff
@@ -14,10 +14,43 @@ def create
       redirect_to hub_client_messages_path(client_id: @client, anchor: "last-item")
     end
 
+    def unsubscribe_email
```
embarnard, line 17: Sorry I know we did do that hub route work around but looking at it more I do think it makes more sense to put it in a controller thats not the hub and make it more like the state-file `NotificationsSettingsController`

embarnard (approved):
Looks good! Just that one nitpicky thing about file placement but excellent work 🎆

app/views/hub/outgoing_emails/unsubscribe_email.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :page_title, t(".title") %>
```
embarnard, line 1: Nitpicky but also I think the naming of this page confuses me, can we call it "unsubscribe_from_emails_page` or something. I think when I see `unsubscribe_email` I think its an mailer

PR #4312, opened 2024-03-06
Case fix

tofarr:


jenny-heath (approved):
again would love to see tests later but this is good for now!

PR #4315, opened 2024-03-06
Fixing to_state #187136341

tofarr:
Fix for this:
image
We need a from_state so I can filter out duplicates here in datadog - when state goes from `transmitted` to `transmitted`.


jenny-heath (approved):
looks ok to me! funny that we were sending an id as "from_status" but a state as "to_status"...

PR #4318, opened 2024-03-06
Add third party designee

tahsinaislam:


app/models/direct_file_data.rb
```diff
@@ -612,6 +616,39 @@ def build_new_qualifying_child_information_node
     parsed_xml.css('QualifyingChildInformation').last.add_next_sibling(dd.to_s)
   end
 
+  def third_party_designee
+    df_xml_value(__method__)
+  end
+
+  def third_party_designee=(value)
+    create_or_destroy_df_xml_node(__method__, true, 'ThirdPartyDesigneeInd')
```
mpidcock, line 624: I think technically if `value` is `nil` then you want to pass `false` instead of `true` here, but it seems unlikely that this setter will ever be used to delete the node, so it's not important to fix this.

app/models/direct_file_data.rb
```diff
@@ -56,6 +56,10 @@ class DirectFileData
     primary_claim_as_dependent: 'IRS1040 PrimaryClaimAsDependentInd',
     hoh_qualifying_person_name: 'IRS1040 QualifyingHOHNm',
     surviving_spouse: 'IRS1040 SurvivingSpouseInd',
+    third_party_designee: 'IRS1040 ThirdPartyDesigneeInd',
```
mpidcock, line 59: it might be nice to keep the "ind" / "indicator" in the name to make it more clear that this is the boolean value not the name (out of context it can be hard to keep it straight). suggesting `third_party_designee_ind` or `third_party_designee_indicator`

app/controllers/flows_controller.rb
```diff
@@ -596,6 +596,7 @@ def self.common_attributes
     def self.ny_attributes(first_name: 'Testuser', last_name: 'Testuser')
       common_attributes.merge(
         confirmed_permanent_address: "no",
+        confirm_third_party_designee: "unfilled",
```
mpidcock, line 599: Minor suggestion to rename to `confirmed_third_party_designee` to be consistent with `confirmed_permanent_address` - it is past tense because once it has been set to yes that means the confirmation has already happened.

app/controllers/state_file/questions/ny_third_party_designee_controller.rb
```diff
@@ -0,0 +1,9 @@
+module StateFile
+  module Questions
+    class NyThirdPartyDesigneeController < AuthenticatedQuestionsController
```
mpidcock, line 3: Since this data is on the review page, you should add `include ReturnToReviewConcern` so that if someone edits this from the review page they go back to review after submitting.

config/locales/en.yml
```diff
@@ -2337,6 +2338,13 @@ en:
           helper_heading: How do I find my school district?
           school_district_name: School District Name
           title: Select the school district where you lived on December 31, %{filing_year}
+      ny_third_party_designee:
+        edit:
+          reveal:
+            body_html: "By selecting a third-party designee, you are authorizing NY Department of Taxation and Finance to discuss with the designee any questions related to this return. You are also authorizing the designee to give and receive confidential taxpayer information relating to: \n<br> \n<b><ul class=\"list--bulleted\"> \n  <li>this return, including missing information \n  <li>any notices or bills arising from this filing that you share with the designee (they will not be sent to the designee) \n  <li>any payments and collection activity arising from this filing, and \n  <li>the status of your return or refund. \n</ul></b>'\n"
```
mpidcock, line 2344: There's a trailing apostrophe here, and I think the newlines are probably being ignored and could be removed entirely.
image

mpidcock (approved):
Good code, just some minor fixes to look at

config/locales/en.yml
tahsinaislam: yeah super weird that i18n normalize added all those new lines

app/models/direct_file_data.rb
tahsinaislam: all other setters are set to true. so keeping it consistent here. this is only used in test scenarios

mpidcock: yeah, that's fine. I don't think our tests ever delete direct file nodes, so keeping it always true doesn't ever cause a problem. But if we did want to delete these fields/nodes, we would have to update this code to be conditional instead of hard coded. It's definitely not a problem right now though.

PR #4319, opened 2024-03-07
Truncate mailing city and permanent city on NY XML

mpidcock:
https://www.pivotaltracker.com/n/projects/2650644/stories/187187752

The city appears twice and was not truncated when passed through from Direct File. These values are not used in the AZ xml, and do not have a character limit on the PDF.


rickreyhsig (approved):
Lgtm

tofarr (approved):

PR #4325, opened 2024-03-07
W2 story follow ups #187175908

tofarr:
* Does not show or link to index page if there is a single invalid W2
* Has additional validation checks on W2s
* Has additional show checks on W2s
* Changes the validation flow on the index page
* Adds specs for all these

### DIFF
--- a/app/controllers/state_file/questions/ny_w2_controller.rb
+++ b/app/controllers/state_file/questions/ny_w2_controller.rb
@@ -9,14 +9,18 @@ module StateFile
       end
 
       def index
-        @w2s_with_metadata = @w2s.map do |w2|
-          dfw2 = w2.state_file_intake.direct_file_data.w2s[w2.w2_index]
-          {
-            w2: w2,
-            employer_name: dfw2.EmployerName,
-            wages_amount: dfw2.WagesAmt,
-          }
+        get_w2s_with_metadata
+      end
+
+      def create
+        # The below line triggers validation on all w2s...
+        @errors_present = @w2s.select { |w2| !w2.valid? }.present?
+        if @errors_present
+          get_w2s_with_metadata
+          render :index
+          return
         end
+        redirect_to next_path
       end
 
       def edit
@@ -41,6 +45,17 @@ module StateFile
         params.require(StateFileW2.name.underscore)
               .except(:state_file_intake_id, :state_file_intake_type)
               .permit(*StateFileW2.attribute_names)
+      end
+
+      def get_w2s_with_metadata
+        @w2s_with_metadata ||= @w2s.map do |w2|
+          dfw2 = w2.state_file_intake.direct_file_data.w2s[w2.w2_index]
+          {
+            w2: w2,
+            employer_name: dfw2.EmployerName,
+            wages_amount: dfw2.WagesAmt,
+          }
+        end
       end
 
       def load_w2s
--- a/app/controllers/state_file/questions/ny_w2_controller.rb
+++ b/app/controllers/state_file/questions/ny_w2_controller.rb
@@ -9,6 +9,9 @@ module StateFile
       end
 
       def index
+        if @w2s.length == 1
+          redirect_to action: :edit, id: @w2s[0].w2_index
+        end
         get_w2s_with_metadata
       end
 
@@ -31,6 +34,7 @@ module StateFile
 
         if @w2.valid?
           @w2.save
+          redirect_to next_path and return if @w2s.length == 1
           redirect_to action: :index
         else
           render :edit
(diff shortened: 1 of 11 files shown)


app/controllers/state_file/questions/ny_w2_controller.rb
```diff
@@ -9,14 +9,21 @@ def self.show?(intake)
...
       def index
-        @w2s_with_metadata = @w2s.map do |w2|
-          dfw2 = w2.state_file_intake.direct_file_data.w2s[w2.w2_index]
-          {
-            w2: w2,
-            employer_name: dfw2.EmployerName,
-            wages_amount: dfw2.WagesAmt,
-          }
+        if @w2s.length == 1
+          redirect_to action: :edit, id: @w2s[0].w2_index
```
rickreyhsig, line 13: If there is only one W2, we don't perform any validation?

spec/controllers/state_file/questions/ny_w2_controller_spec.rb
```diff
@@ -164,6 +164,42 @@
...
+        let(:direct_file_xml) do
+          xml = super()
+          xml.at("LocalIncomeTaxAmt").content = "9000"
+          xml
+        end
+
+        it "returns true" do
+          expect(described_class.show?(intake)).to eq true
+        end
+      end
```
rickreyhsig, line 202: Could be good to have a spec where #show returns false also.

db/create_analytics_views.sql
```diff
@@ -297,3 +297,7 @@ CREATE view analytics.state_file_az_intakes AS
 CREATE view analytics.state_file_ny_intakes AS
     SELECT id, updated_at, consented_to_terms_and_conditions, contact_preference, created_at, current_step, eligibility_lived_in_state, eligibility_out_of_state_income, eligibility_part_year_nyc_resident, eligibility_withdrew_529, eligibility_yonkers, hashed_ssn, nyc_maintained_home, nyc_residency, primary_birth_date, primary_esigned_at, primary_state_id_id, raw_direct_file_data, routing_number, sign_in_count, spouse_state_id_id, untaxed_out_of_state_purchases
     FROM public.state_file_ny_intakes;
+
+CREATE view analytics.state_file_w2s AS
+SELECT id, employer_state_id_num, local_income_tax_amt, local_wages_and_tips_amt, locality_nm, state_file_intake_type, state_income_tax_amt, state_wages_amt, w2_index, created_at, updated_at, state_file_intake_id
+FROM public.state_file_w2s;
```
rickreyhsig, line 303: Is this view for the metabase analytics?

app/controllers/state_file/questions/ny_w2_controller.rb
```diff
@@ -9,14 +9,21 @@ def self.show?(intake)
...
 
       def index
-        @w2s_with_metadata = @w2s.map do |w2|
-          dfw2 = w2.state_file_intake.direct_file_data.w2s[w2.w2_index]
-          {
-            w2: w2,
-            employer_name: dfw2.EmployerName,
-            wages_amount: dfw2.WagesAmt,
-          }
+        if @w2s.length == 1
```
rickreyhsig, line 12: So this renders the one W2 if there's only one and and index of the W2s if there are more?

rickreyhsig (commented):
Overall, the PR makes sense. Just added a few questions for my own sanity prior to approving.

app/controllers/state_file/questions/ny_w2_controller.rb
tofarr: Correct!

app/controllers/state_file/questions/ny_w2_controller.rb
tofarr: If there is only 1 W2, we bypass the index page and take users directly into the edit for that W2

config/routes.rb
```diff
@@ -582,7 +582,7 @@ def scoped_navigation_routes(context, navigation)
 
       scope ':us_state', as: 'ny', constraints: { us_state: :ny } do
         scoped_navigation_routes(:questions, Navigation::StateFileNyQuestionNavigation)
-        resources :ny_w2, only: [:index, :edit, :update], module: 'state_file/questions', path: 'questions/ny_w2'
+        resources :ny_w2, only: [:index, :edit, :update, :create], module: 'state_file/questions', path: 'questions/ny_w2'
```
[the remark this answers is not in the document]
tofarr: Only New York has been experiencing this issue - we leave the possiblilty of doing this for other states (Though I would probably rename the controller then from NyW2Controller to simply W2Controller

db/create_analytics_views.sql
tofarr: Yes - that was one of the requirements of the ticket

spec/controllers/state_file/questions/ny_w2_controller_spec.rb
tofarr: Good call - adding it now

tofarr: We have some already on L17, L50 & L112,

rickreyhsig (approved):

PR #4330, opened 2024-03-08
WIP: Refactor seeder file

embarnard:


app/lib/seeder.rb
```diff
@@ -9,6 +10,24 @@ class Seeder; end
 class Seeder
   include ActiveSupport::Testing::TimeHelpers
 
+  def run
+    return if Rails.env.production?
+
+    self.class.load_fraud_indicators
+
+    Flipper.enable(:eitc)
```
embarnard, line 18: any other flippers we should enable or should we delete them from different environments?

PR #4334, opened 2024-03-09
Process taking longer notification #187147601

jenny-heath:
acceptance process:
- [x] add code that prevents heroku submissions from failing on bundle
- [x] make submissions on heroku app; eng transitions manually to failed/rejected/etc
- [x] see that jobs are queued for 24 hours later
- [x] see texts go through as expected
- [x] remove code that prevents heroku submissions from failing

message previews here: https://gyr-review-app-4334-218707a922a7.herokuapp.com/en/hub/state-file/automated_messages

### DIFF
--- a/app/jobs/build_submission_bundle_job.rb
+++ b/app/jobs/build_submission_bundle_job.rb
@@ -5,7 +5,7 @@ class BuildSubmissionBundleJob < ApplicationJob
     begin
       submission.generate_irs_submission_id!
     rescue StandardError => e
-      submission.transition_to!(:failed, error_code: 'IRS-ID-FAIL', raw_response: e.inspect) unless Rails.env.heroku?
+      submission.transition_to!(:failed, error_code: 'IRS-ID-FAIL', raw_response: e.inspect)
     end
 
     address_creation = submission.generate_verified_address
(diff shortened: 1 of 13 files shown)


tofarr (approved):

app/jobs/build_submission_bundle_job.rb
```diff
@@ -5,7 +5,7 @@ def perform(submission_id)
     begin
       submission.generate_irs_submission_id!
     rescue StandardError => e
-      submission.transition_to!(:failed, error_code: 'IRS-ID-FAIL', raw_response: e.inspect)
+      submission.transition_to!(:failed, error_code: 'IRS-ID-FAIL', raw_response: e.inspect) unless Rails.env.heroku?
```
embarnard, line 8: Oh nice! I was also thinking about how to prevent this issue and opted to add transition to reject to the failed options and add a guard to only allow on non-prod environments but I guess this works too!

```
guard_transition(from: :failed, to: :rejected) do |_submission|
    !Rails.env.production?
  end
  ```

embarnard (approved):
Looks great!

app/jobs/build_submission_bundle_job.rb
jenny-heath: oh i was just gonna delete this but we could also add something more permanent, is that what you meant?

PR #4335, opened 2024-03-11
Add whole number error message

tahsinaislam:


app/models/state_file_w2.rb
```diff
@@ -36,14 +36,19 @@ class StateFileW2 < ApplicationRecord
...
-  validates :state_income_tax_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { state_income_tax_amt.present? }
-  validates :local_wages_and_tips_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { local_wages_and_tips_amt.present? }
-  validates :local_income_tax_amt, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, if: -> { local_income_tax_amt.present? }
-  validates :locality_nm, presence: { message: -> (_object, _data) { I18n.t('state_file.questions.ny_w2.edit.locality_nm_missing_error') } }, if: -> { local_wages_and_tips_amt.present? && local_wages_and_tips_amt.positive? }
+  validates_numericality_of :state_wages_amt, only_integer: true, message: I18n.t('errors.messages.whole_number'), if: -> { state_wages_amt.present? }
+  validates :state_wages_amt, numericality: { greater_than_or_equal_to: 0 }, if: -> { state_wages_amt.present? }
+  validates_numericality_of :state_income_tax_amt, only_integer: true, message: I18n.t('errors.messages.whole_number'), if: -> { state_income_tax_amt.present? }
```
rickreyhsig, line 45: I see you have a pattern of two validations per field.  Is it possible to have one that achieves the desired result? It seems we are looking for these things:

- that the number be positive
- that the number be an integer 
- to display the error message otherwise

Have you tried something like:

`validates :your_number_attribute, numericality: { greater_than_or_equal_to: 0, only_integer: true, message: I18n.t('errors.messages.whole_number') } , if: -> { your_number_attribute.present? }`

tahsinaislam: i tried doing this but it would show up as an error for both greater_than_or_equal_to: 0 and only_integer: true

rickreyhsig: So the validation only fails if both errors are present this way?

tahsinaislam: yeah the message becomes the message for both validations

tahsinaislam: Like using the code you gave would not give the right message if someone put 0 for greater than or equal to 0
Screenshot 2024-03-11 at 4 07 26 PM

rickreyhsig: Makes sense! Tnx for clarifying

rickreyhsig (approved):
Looks awesome!!

PR #4340, opened 2024-03-11
Full node delete #187175908

tofarr:
Courtney found a situation where the W2StateTaxGrp node was not deleted even if all content was blank. 
image

This PR fixes that.

### DIFF
--- a/app/lib/submission_builder/shared/return_w2.rb
+++ b/app/lib/submission_builder/shared/return_w2.rb
@@ -6,7 +6,13 @@ module SubmissionBuilder
         intake_w2 = @kwargs[:intake_w2]
         xml_node = Nokogiri::XML(w2.node.to_xml)
         if intake_w2.present?
-          xml_node.at(:W2StateLocalTaxGrp).inner_html = intake_w2.state_tax_group_xml_node
+          state_local_tax_grp_node = xml_node.at(:W2StateLocalTaxGrp)
+          state_tax_group_xml = intake_w2.state_tax_group_xml_node
+          if state_tax_group_xml.present?
+            state_local_tax_grp_node.inner_html = state_tax_group_xml
+          else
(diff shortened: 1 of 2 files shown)


app/models/state_file_w2.rb
```diff
@@ -96,9 +96,8 @@ def state_tax_group_xml_node
     xml_template.at(:LocalWagesAndTipsAmt).content = local_wages_and_tips_amt
     xml_template.at(:LocalIncomeTaxAmt).content = local_income_tax_amt
     xml_template.at(:LocalityNm).content = locality_nm
-
+    delete_blank_nodes(xml_template)
     result = xml_template.at(:W2StateTaxGrp)
-    delete_blank_nodes(result)
-    result.to_xml
+    result.nil? ? "" : result.to_xml
   end
```
tofarr, line 102: We now remove nodes from the overall document object rather than just the root node. This means that we can remove the root node

app/lib/submission_builder/shared/return_w2.rb
```diff
@@ -6,7 +6,13 @@ def document
...
         if intake_w2.present?
-          xml_node.at(:W2StateLocalTaxGrp).inner_html = intake_w2.state_tax_group_xml_node
+          state_local_tax_grp_node = xml_node.at(:W2StateLocalTaxGrp)
+          state_tax_group_xml = intake_w2.state_tax_group_xml_node
+          if state_tax_group_xml.present?
+            state_local_tax_grp_node.inner_html = state_tax_group_xml
+          else
+            state_local_tax_grp_node.remove
+          end
         end
```
tofarr, line 16: We accommodate the situation where the root node was removed

rickreyhsig (approved):
Would be nice to add a spec.  Approving since this is high priority.

PR #4351, opened 2024-03-12
Small refactor: move email-not-sent-because-unsubscribed test to messaging service, delete old validation code

jenny-heath:


embarnard (approved):
Looks good!

PR #4355, opened 2024-03-13
Az w2 page #187183720

tofarr:


app/models/state_file_base_intake.rb
```diff
@@ -132,8 +132,22 @@ def spouse_deceased?
     direct_file_data.spouse_deceased?
   end
 
-  def self.locality_nm_valid?(locality)
-    true
+  def validate_w2(w2)
```
jenny-heath, line 135: does not having this defined on the base/az intake result in an error?

app/models/state_file_ny_intake.rb
```diff
@@ -260,10 +260,23 @@ def disqualifying_eligibility_rules
...
-  def self.locality_nm_valid?(locality)
-    (LOCALITIES.detect do |loc|
-      locality.starts_with?(loc)
-    end).present?
+  def self.locality_nm_valid?(locality_nm)
+    locality_nm = locality_nm.upcase
+    (LOCALITIES.detect { |loc| locality_nm.starts_with?(loc) }).present?
+  end
+
+  def validate_w2(w2)
```
jenny-heath, line 268: this feels like a bit of a misnomer because it's not really telling you if the whole w2 is valid, just the locality name

config/routes.rb
```diff
@@ -587,7 +588,8 @@ def scoped_navigation_routes(context, navigation)
 
       scope ':us_state', as: 'ny', constraints: { us_state: :ny } do
         scoped_navigation_routes(:questions, Navigation::StateFileNyQuestionNavigation)
-        resources :ny_w2, only: [:index, :edit, :update, :create], module: 'state_file/questions', path: 'questions/ny_w2'
+        # TODO: This can be deleted once no intake has w2 as the current step
```
jenny-heath, line 591: does "this" refer to the path being `ny_w2`?

db/schema.rb
```diff
@@ -11,6 +11,8 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.1].define(version: 2024_03_07_232948) do
+  create_schema "analytics"
```
jenny-heath, line 14: this seems unrelated / out of scope?

spec/models/state_file_az_intake_spec.rb
```diff
@@ -311,4 +311,20 @@
...
   end
+
+  describe "invalid_df_w2?" do
+    let(:intake) { create :state_file_az_intake }
+
+    it 'does not validate locality names against known lists' do
+      df_w2 = intake.direct_file_data.w2s[0]
+      df_w2.LocalityNm = "JUNO"
+      expect(intake.invalid_df_w2?(df_w2)).to eq false
+    end
```
jenny-heath, line 322: i can never decide whether it's superfluous to test that something is *not* happening, but my instinct here is that in future, without the context that ny w2s came first, this will look odd

app/models/state_file_base_intake.rb
```diff
@@ -132,8 +132,22 @@ def spouse_deceased?
...
+  end
+
+  def invalid_df_w2?(df_w2)
```
jenny-heath, line 138: sorry if this is a bit tedious but i think that if all this code moved here, the tests should also be moved to the model spec. seems strange to leave it in the controller specs.

app/models/state_file_ny_intake.rb
jenny-heath: oh ok i get it now. maybe consider naming it `validate_state_specific_w2_requirements` or something like that? only suggesting it because it's generally pretty confusing that we have 2 sets of validations for DF W2s and our model W2s and within our model W2s we have validations that live on that model and also these ones that live on the intake model. so anything we can do to add more context than just "validate" and "w2" is maybe good

jenny-heath (approved):
ok, i don't *think* i see anything blocking (except that `state_file_az_question_navigation_spec` needs to be fixed), just a few suggestions around naming and where specs should live

config/routes.rb
tofarr: Fixed.

app/models/state_file_ny_intake.rb
tofarr: Good call - done.

db/schema.rb
tofarr: For sure - reverted

app/models/state_file_base_intake.rb
tofarr: I was on the fence about this - but I agree. Done

PR #4356, opened 2024-03-13
Hide intercom from various pages #187214392

jenny-heath:
i think i would have to test this with a feature spec and that didn't seem worth it because we can just visually inspect for the presence of the intercom widget, but open to feedback if anyone thinks that's the wrong call


rickreyhsig (approved):
LGTM

PR #4367, opened 2024-03-15
Prevent continue as long as there are unsaved W2s #187248240

tofarr:
We now do not allow users to continue and warn users unless they have updated all suspicious W2s: (Previously if a W2 was valid but suspicious, we would allow them not to update it.

image


spec/controllers/state_file/questions/w2_controller_spec.rb
```diff
@@ -255,4 +255,28 @@
       end
     end
   end
+
+  describe "#create" do
+    context "with a sus but valid w2" do
```
embarnard, line 260: lol susssss

spec/controllers/state_file/questions/w2_controller_spec.rb
```diff
@@ -255,4 +255,28 @@
...
+      let(:direct_file_xml) do
+        xml = super()
```
embarnard, line 262: woah so cool I didn't know you could do this

embarnard (approved):
Looks good! 🎆

PR #4371, opened 2024-03-15
Add page routing functionality to auto wait #187091623

tofarr:
The efile_errors table now has a "correction_path" string column.

This may be populated using the error page in the hub:

image

And is used to populate the "Edit Your State Return" link:

image

The dropdown list of pages is auto populated from navigations controllers which include the ReturnToReviewConcern.

This concern has been updated to implement prev_step as well as next step.

Also, a bunch of controllers now include the ReturnToReviewConcern, and a few (W2, Unemployment) have had their flows updates to support return to support this.


app/models/efile_error.rb
```diff
@@ -47,4 +48,19 @@ def resolution(locale)
       resolution_es.present? ? resolution_es : resolution_en
     end
   end
+
+  def data_source
```
embarnard, line 52: hmm this does seem very complex, are we only using this `default_correction_path` method? If so, I would advocate to just leave the default in the efile errors controller and the return-status rejected page

app/views/state_file/questions/return_status/_rejected.html.erb
```diff
@@ -49,6 +49,6 @@
     <%= t('.contact_us') %>
   </section>
 
-  <% edit_return_path = params[:us_state] == 'az' ? az_questions_name_dob_path : ny_questions_name_dob_path %>
-  <%= link_to t('.edit_return'), edit_return_path, class: "button button--primary spacing-above-60" %>
+  <% edit_return_path = @error.correction_path || @error.default_correction_path %>
+  <%= link_to t('.edit_return'), "#{edit_return_path}?return_to_review=y", class: "button button--primary spacing-above-60" %>
```
embarnard, line 53: what does `return_to_review=y` do?

embarnard: is the `y` not significant because its just checking the presence of it at all?

app/controllers/hub/state_file/efile_errors_controller.rb
```diff
@@ -7,7 +7,18 @@ def index
       @efile_errors = @efile_errors.where.not(service_type: "ctc").order(:source, :code)
     end
 
-    def edit; end
+    def edit
```
embarnard, line 10: I think we have this in an upcoming ticket but it might be more efficient to do the work to make EfileErrors state specific right now. So either adding a `state` column or replacing the `state_file` service type with `state_file_az` and `state_file_ny`. We would collect the state in the error parser when we create the record but this would also require us to create a job to update the existing ones which might be annoying.

embarnard: That way we wouldn't have to do all this complicated logic to find the `data_source`. I think its also confusing to be calling the data source in this way because one `EfileError` can have multiple `data_source`s attached to it and usually does. What if one has a locale of english and another doesn't. Then would we alway be sending them back based on the chosen intake's locale?

If we had the state info on the `EfileError` record we could just pull that to grab the navigation options and then put the locale in when we actually create the link on the `return-status` page for that specific client

embarnard (changes requested):
I think this could work but might be overcomplicating some things especially considering some requested changes to the EfileError model that are coming up in future tickets. See my comments in the efile error controller below. Feel free to call me if you want to talk about this more. Also seeing some failing tests.

app/views/state_file/questions/return_status/_rejected.html.erb
tofarr: return_to_review is only checking presents - it means the prev / next step is data review rather than what is before / after the controller in the navigation

app/controllers/concerns/state_file/return_to_review_concern.rb
```diff
@@ -4,19 +4,18 @@ module ReturnToReviewConcern
...
-    private
-
     def review_step
-      case params[:us_state]
-      when 'az'
-        StateFile::Questions::AzReviewController
-      when 'ny'
-        StateFile::Questions::NyReviewController
-      end
+      "StateFile::Questions::#{current_intake.state_code.titleize}ReviewController".constantize
```
tofarr, line 8: Doing it this way means it is more likely to just work with future states

PR #4372, opened 2024-03-15
Add validation to NY 1099-G Payer's TIN #187222366

tofarr:
There are 2 separate error messages. This is for both Ny and Az (If the number is not a 9 digit number):
image

This only occurs for NY if the previous message is not shown (NY has 2 known Ids):
image


tahsinaislam (approved):

app/models/state_file_base_intake.rb
```diff
@@ -150,6 +150,9 @@ def invalid_df_w2?(df_w2)
     false
   end
 
+  def validate_state_specific_1099_g_requirements(state_file1099_g)
+  end
```
tahsinaislam, line 154: is this method supposed to have a body?

config/locales/en.yml
```diff
@@ -179,6 +179,7 @@ en:
         invalid: 'Only letters, numbers, parentheses, hyphen, apostrophe, and # are accepted.'
       payer_tin:
         invalid: EIN must be a 9-digit number. Do not include a dash.
+      payer_tin_ny_invalid: "EIN must be one of: %{valid_tins}"
       phone_number:
```
tahsinaislam, line 183: The story said for the error to be the following: The number entered is not an accepted TIN by New York State. Was this a change made later?

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -99,14 +99,15 @@
...
+      fill_in I18n.t('state_file.questions.unemployment.edit.payer_tin'), with: "270293117"
       choose I18n.t('state_file.questions.unemployment.edit.confirm_address_yes')
       fill_in 'state_file1099_g_unemployment_compensation', with: "123"
       fill_in 'state_file1099_g_federal_income_tax_withheld', with: "456"
       fill_in 'state_file1099_g_state_identification_number', with: "123456789"
       fill_in 'state_file1099_g_state_income_tax_withheld', with: "789"
       click_on I18n.t("general.continue")
 
+      binding.pry
       expect(page).to have_text(I18n.t('state_file.questions.unemployment.index.1099_label', name: StateFileNyIntake.last.primary.full_name))
```
tahsinaislam, line 111: need to take out this binding.pry?

spec/models/state_file_ny_intake_spec.rb
```diff
@@ -579,4 +579,19 @@
...
+  describe ".validate_state_specific_1099_g_requirements" do
+    let(:intake) { create :state_file_ny_intake, untaxed_out_of_state_purchases: "yes", sales_use_tax_calculation_method: "manual", sales_use_tax: "350" }
+    let(:state_file_1099) do
+      build(
+        :state_file1099_g,
+        intake: create(:state_file_ny_intake),
+      )
+    end
+
+    it "validates that the PayerTIN is a known value" do
```
tahsinaislam, line 592: this is validating that the TIN is not a known value because there is an error. should the title be changed?

spec/features/state_file/complete_intake_spec.rb
tofarr: Good catch...

spec/models/state_file_ny_intake_spec.rb
tofarr: Updated the description

config/locales/en.yml
tofarr: I've changed this

PR #4381, opened 2024-03-19
Truncate dependent names

tahsinaislam:


app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -55,7 +55,7 @@ def documents_wrapper
               xml.FilingStatus filing_status
               if @submission.data_source.hoh_qualifying_person_name.present?
                 xml.QualChildDependentName do
-                  xml.FirstName @submission.data_source.hoh_qualifying_person_name[:first_name]&.strip&.gsub(/\s+/, ' ')
+                  xml.FirstName @submission.data_source.hoh_qualifying_person_name[:first_name]&.gsub(/\s+/, ' ')&.slice(0, 16).strip
```
rickreyhsig, line 58: I see a slight alteration in how you are truncating.  The two versions are:

1. gsub(/\s+/, ' ')&.slice(0, 16).strip
2. strip.gsub(/\s+/, ' ')&.slice(0, 16).strip

May be better to combine into like one method/module for example:

```ruby
module DependentString
  def self.truncate(input)
    truncated_string = input.gsub(/\s+/, ' ')
    truncated_string = truncated_string.slice(0, 16) if truncated_string
    truncated_string = truncated_string.strip if truncated_string
    
    return truncated_string
  end
end
```

```ruby
include DependentString
input_string = "   This    is   a   test    string   "
manipulated_result = DependentString.truncate(input_string)
```

app/lib/submission_builder/formatting_methods.rb
```diff
@@ -6,6 +6,10 @@ def trim(string, length)
       string.squish.first(length)&.strip
     end
 
+    def truncate(string, length)
+      string&.gsub(/\s+/, ' ')&.slice(0, length)&.strip
+    end
```
rickreyhsig, line 11: Huge improvement! Thanks for working on this :shipit:

rickreyhsig (approved):

PR #4389, opened 2024-03-19
WIP throttling login requests: install rack-attack, limit based on IP

jenny-heath:


spec/features/hub/rack_attack_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+describe Rack::Attack, type: :request do
+  let(:limit) { 5 }
+  let(:ip) { "1.2.3.4" }
+
+  context "on a post to a login page" do
```
mpidcock, line 7: note: consider adding test case to see that this works on state file too

PR #4390, opened 2024-03-20
W2 page error copy updates #187269981

tofarr:
image
image
image

### DIFF
--- a/app/models/state_file_ny_intake.rb
+++ b/app/models/state_file_ny_intake.rb
@@ -270,8 +270,10 @@ class StateFileNyIntake < StateFileBaseIntake
   end
 
   def validate_state_specific_w2_requirements(w2)
-    if w2.locality_nm.present? && !self.class.locality_nm_valid?(w2.locality_nm)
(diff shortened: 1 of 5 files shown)


app/models/state_file_ny_intake.rb
```diff
@@ -270,8 +270,10 @@ def self.locality_nm_valid?(locality_nm)
   end
 
   def validate_state_specific_w2_requirements(w2)
-    if w2.locality_nm.present? && !self.class.locality_nm_valid?(w2.locality_nm)
-      w2.errors.add(:locality_nm, I18n.t("state_file.questions.ny_w2.edit.locality_nm_error"))
+    unless w2.errors[:locality_nm].present?
+      if w2.locality_nm.present? && !self.class.locality_nm_valid?(w2.locality_nm)
+        w2.errors.add(:locality_nm, I18n.t("state_file.questions.ny_w2.edit.locality_nm_error"))
+      end
     end
```
tofarr, line 277: We only display the second error message on the locality name field if there is not already one present

PR #4394, opened 2024-03-20
Fix i18n validation messages #187278122

tofarr:
The lookup for message Internationalization should be deferred until the validation is being performed so it can use the locale for the current request. Otherwise you can get things like this:
image

Or this:
image

There are 3 types that the message parameter accepts:
* string - not internationalized and simply displayed. **DO NOT DO THIS!!!**
* A symbol - looked up using Rails rules for i18n. e.g.: `validates :charitable_contributions, inclusion: { in: %w[yes no], message: :blank }`
* A function that takes the model and data as parameters `message: ->(_object, _data) { I18n.t('deferred.lookup') }`


jenny-heath (approved):

rickreyhsig (approved):

rickreyhsig:
Looks awesome @tofarr, thanks for fixing this!

PR #4395, opened 2024-03-20
Job for sending survey notifications to clients who did not receive one

rickreyhsig:
https://www.pivotaltracker.com/story/show/187175366

### Rake task
```
rake survey_notifications:send 
```

### Manual testing
```
require 'factory_bot'
FactoryBot.find_definitions
include FactoryBot::Syntax::Methods

intake = create(:state_file_ny_intake)
intake.update(email_address: 'rkreyhsig@codeforamerica.org', email_address_verified_at: Time.now)
sub = create(:efile_submission, :for_state, :accepted, data_source: intake)

intake = create(:state_file_az_intake)
intake.update(email_address: 'rkreyhsig@codeforamerica.org', email_address_verified_at: Time.now)
sub = create(:efile_submission, :for_state, :accepted, data_source: intake)

rake survey_notifications:send 
```


app/services/efile/send_survey_notifications_service.rb
```diff
@@ -0,0 +1,24 @@
+module Efile
+  class SendSurveyNotificationsService
+    include StateFile::SurveyLinksConcern
+    BATCH_SIZE = 5
+
+    def run
+      accepted_submissions = EfileSubmission.joins(:efile_submission_transitions)
```
embarnard, line 7: we should add an index for efile submission transitions `to_state`

lib/tasks/efile.rake
```diff
@@ -23,4 +23,8 @@ namespace :efile do
     puts("End: Found #{EfileSubmission.in_state(:failed).count} failed submissions")
     exit
   end
+
+  task send_survey_notifications: :environment do
```
embarnard, line 27: Hmm I don't _love_ that we are putting this task in this file. I think I would prefer that this file is just for efile tasks that we use regularly. For example, `poll_and_get_acknowledgments` we run in cron every 10 minutes and I assume `retry_failed_submissions` was used semi regularly with CTC failed submissions. I think I would prefer it was in a separate rake task and that all the logic in the service is just placed in that rake task instead of getting called from a service. And then eventually one day we just remove that rake task.

app/services/efile/send_survey_notifications_service.rb
embarnard: actually since we are just running this once maybe depending on the number of clients this isn't necessary, I would just check the number of clients demo before using. Also I think the batch_size can go up.

embarnard (changes requested):
I think this will work but I would prefer everything was just in a separate rake task and we didn't have that service. Also ideally there would be a test for this. And make sure you test on demo before prod.

app/services/efile/send_survey_notifications_service.rb
```diff
@@ -0,0 +1,24 @@
...
+                                            .where("efile_submission_transitions.to_state = 'accepted'")
+                                            .where.not("message_tracker #> '{messages.state_file.survey_notification}' IS NOT NULL")
```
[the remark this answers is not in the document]
rickreyhsig: Yep it works! I filters only based on whether or not there have been a survey_notification message sent. For instance, this one had and accepted_refund message sent:

Screen Shot 2024-03-21 at 2 18 40 PM

lib/tasks/efile.rake
rickreyhsig: I've moved all the code into `lib/tasks/send_survey_notifications.rake`

embarnard (approved):
Perfect!! Looks great 🎆

PR #4406, opened 2024-03-22
Check required intake data #187270706

tofarr:
# TLDR
We have a situation where a user can start a new intake in a different tab, go back to their original tab, and have that data put into the new intake. 

# Details
1. Go to the landing page - answer the screener questions and enter an email and proceed as far as the direct file transfer / fake direct file transfer, but do not come back into our app. Keep this tab open
2. Open a new tab - go to the landing page and start a new intake - Don't go as far as entering your email address
3. In the IRS direct file app / the fake direct file app - go back into our app
4. You now have an intake that has direct file data, but has not consented to the terms, and may not even have an email address.

# Approach
The fix for this is inspired by the Rails XSRF protection. When we load a form - we now add a hidden element into it containing the intake id. We then check this before the resulting action for the form is invoked. If these are both present and do not match, we redirect to login with a message "Please edit one intake at a time!"
image

### DIFF
--- a/app/controllers/state_file/questions/questions_controller.rb
+++ b/app/controllers/state_file/questions/questions_controller.rb
@@ -3,7 +3,8 @@ module StateFile
     class QuestionsController < ::Questions::QuestionsController
       include StateFile::StateFileControllerConcern
       before_action :redirect_if_no_intake
-      before_action :redirect_if_intake_mismatch
+      # Noopping this until Monday
+      # before_action :redirect_if_intake_mismatch
       helper_method :card_postscript
 
       # default layout for all state file questions
--- a/app/controllers/state_file/questions/questions_controller.rb
+++ b/app/controllers/state_file/questions/questions_controller.rb
@@ -3,8 +3,6 @@ module StateFile
     class QuestionsController < ::Questions::QuestionsController
       include StateFile::StateFileControllerConcern
       before_action :redirect_if_no_intake
-      # Noopping this until Monday
-      # before_action :redirect_if_intake_mismatch
       helper_method :card_postscript
 
       # default layout for all state file questions
@@ -50,15 +48,6 @@ module StateFile
         unless current_intake.present?
           flash[:notice] = 'Your session expired. Please sign in again to continue.'
           redirect_to StateFile::StateFilePagesController.to_path_helper(action: :login_options, us_state: state_code)
-        end
-      end
-
-      def redirect_if_intake_mismatch
-        if params[:current_intake].present?
-          if current_intake&.persisted? && current_intake.to_global_id.to_s != params[:current_intake]
-            flash[:alert] = I18n.t("general.one_intake_at_a_time")
-            redirect_to StateFile::StateFilePagesController.to_path_helper(action: :login_options, us_state: current_intake.state_code)
-          end
         end
       end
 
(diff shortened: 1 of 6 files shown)


app/controllers/state_file/questions/waiting_to_load_data_controller.rb
```diff
@@ -7,6 +7,7 @@ class WaitingToLoadDataController < QuestionsController
       def edit
         raise ActionController::RoutingError, 'Not Found' unless params[:authorizationCode]
         return redirect_to next_path if current_intake.raw_direct_file_data.present?
+        return redirect_to IntakeLoginsController.to_path_helper(action: :new, us_state: "us") unless current_intake.consented_to_terms_and_conditions
```
tofarr, line 10: There is no protection on the return from IRS so we check that the intake consented!

jnf:
This PR seems like it's addressing 2 distinct concerns. The additional guard/redirect in `app/controllers/state_file/questions/waiting_to_load_data_controller.rb` should solve the issue of folks progressing without recorded consent. That feels very safe and easy to validate.

The secondary issue of concurrent intakes is thornier, an edge of an edge case. I think we should split these, aiming to get the first part (guard/redirect) validated and shipped before talking through the second piece.

app/helpers/vita_min_form_builder.rb
```diff
@@ -368,10 +368,19 @@ def vita_min_date_text_fields(method, label_text, help_text: nil, options: {}, o
...
   def continue(value = I18n.t("general.continue"))
     submit(value, class: "button button--primary button--wide")
   end
+
+  private
+
+  def current_intake_input
+    # A hidden element containing the identified for the current_intake
+    return "" unless @object.respond_to?(:intake) && @object.intake && @object.intake.id
+    @template.hidden_field_tag(:current_intake, @object.intake.to_global_id)
```
jnf, line 385:
```suggestion
    @template.hidden_field_tag(:current_intake, @object.intake.to_signed_global_id)
```
Can we sign these since they're exposed in the client ui?

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -51,6 +52,15 @@ def redirect_if_no_intake
         end
       end
 
+      def redirect_if_intake_mismatch
+        if params[:current_intake].present?
+          if current_intake&.persisted? && current_intake.to_global_id.to_s != params[:current_intake]
```
jnf, line 58:
```suggestion
          if current_intake&.persisted? && current_intake.to_signed_global_id.to_s != params[:current_intake]
```
Can we sign these since they're exposed in the client ui?

app/controllers/state_file/questions/waiting_to_load_data_controller.rb
```diff
@@ -7,6 +7,10 @@ class WaitingToLoadDataController < QuestionsController
       def edit
         raise ActionController::RoutingError, 'Not Found' unless params[:authorizationCode]
         return redirect_to next_path if current_intake.raw_direct_file_data.present?
+        if [nil, "unfilled"].include?(current_intake.consented_to_terms_and_conditions)
+          flash[:alert] = I18n.t("general.one_intake_at_a_time")
+          redirect_to StateFile::StateFilePagesController.to_path_helper(action: :login_options, us_state: current_intake.state_code)
+        end
```
jnf, line 13: this should ship separately (and first, ideally). can you extract this into its own pr?

jnf (changes requested):
i left a suggestion to use signed global ids, but open to hearing any issues/concerns you have with that. for me the bigger concern is decoupling these two issues into separate PRs.

jnf (approved):
ty! looks good; let's be sure to follow up on the other part early next week.

PR #4424, opened 2024-03-28
Add opt out messages indicator in message tab GYR hub

embarnard:


app/views/hub/messages/index.html.erb
```diff
@@ -32,9 +32,9 @@
       <% end %>
 
       <hr/>
-      <div class="communication-preferences">
-        <% unless @client&.intake.email_address.present? || @client&.intake.sms_phone_number.present? %>
-          <%= t(".no_way_to_communicate") %>
+      <div class="communication-preferences spacing-below-25">
+        <% unless (@client&.intake.email_address.present? && @client&.intake.email_notification_opt_in_yes?) || (@client&.intake.sms_phone_number.present? && @client&.intake.sms_notification_opt_in_yes?) %>
+          <em><%= t(".no_way_to_communicate") %></em>
```
jenny-heath, line 37: so self-centered tsk tsk

config/locales/en.yml
```diff
@@ -991,11 +991,15 @@ en:
         email_form:
           label: Send an email
           submit: Send
+        email_opted_out: Client has opted out of emails.
         no_email_address: Client has not provided an email address.
+        no_email_and_opt_out: Client has not provided an email address and opted out of emails.
+        no_sms_and_opt_out: Client has not provided an SMS phone number and opted out of text messages.
         no_sms_number: Client has not provided an SMS phone number.
-        no_way_to_communicate: Client has not provided contact info for digital communication.
+        no_way_to_communicate: Client has not provided contact info for digital communication and/or has opted-out.
```
jenny-heath, line 999: maybe i'm misremembering how gyr works but isn't it maybe more accurate to say that they never opted in? or is there a way to opt out that i'm forgetting?

config/locales/en.yml
```diff
@@ -991,11 +991,15 @@ en:
...
         preference_intro: "%{preferred_name} has opted in to contact by:"
         response_not_needed: Mark as "Response not needed"
+        sms_opted_out: Client has opted out of text messages.
```
jenny-heath, line 1002: likewise with this one i feel like it's more that they didn't check the box that said that was how they want to be contacted. my additional nitpick with this copy is that there's also the twilio-level opt out which i don't think we actually capture and this could look like it covers that when it doesn't.

jenny-heath (commented):
looks pretty good! however, since there's technically some logic there i think you should add a controller spec. you could test the code as it is by rendering views in the spec or you could move the logic to the controller and have it set the i18n key to an instance variable or something (that's probably what i would do)

PR #4429, opened 2024-03-29
Start again #187334156

tofarr:
People visiting the landing page is messing up their existing sessions and causing confusion.

If they are not already logged in, there should be no change to the landing page:
image

If they are already logged in, the landing page should encourage them to log in again. (matching the new mock in Figma):
image

I made some small changes to the landing_page_controller:

* Removed StartIntakeConcern as the landing_page_controller was the only file using it, and it an include on a file that is only ~30 lines long reduces clarity
* Inlined some of the `before_action` / `after_action` code  in the landing_page_controller for clarity


app/controllers/state_file/questions/landing_page_controller.rb
```diff
@@ -1,14 +1,36 @@
...
 
-      def before_update
+      def edit
+        @state_name = StateFileBaseIntake::STATE_CODE_AND_NAMES[params[:us_state]]
+        if current_intake.present?
+          if current_intake.primary_first_name.present?
+            @user_name = "#{current_intake.primary_first_name} #{current_intake.primary_middle_initial} #{current_intake.primary_last_name}"
+          elsif current_intake.raw_direct_file_data.present?
+            begin
+              @user_name = current_intake.direct_file_data.node.at("Filer NameLine1Txt").text.gsub("&lt;", " ")
```
jenny-heath, line 13: this will be all caps in a lot of cases, yes? just confirming it's intentional

tofarr: This is a fallback for cases where they have not completed the NameDOB step - hopefully that is a really small subset

app/views/state_file/questions/landing_page/edit.html.erb
```diff
@@ -19,17 +21,30 @@
...
+      <h2 class="h2"><%= t(".continue", state_name: @state_name) %></h2>
+      <%= link_to StateFile::StateFilePagesController.to_path_helper(action: :login_options, us_state: current_intake.state_code), class: "button button--primary button--wide", id: "firstCta" do %>
+        <%= t("general.sign_in") %>
+      <% end %>
+      <%= form_with model: @form, url: { action: :update }, local: true, method: :put, builder: VitaMinFormBuilder, id: "start-again-form" do |f| %>
+        <h2 class="h2"><%= t(".not_you_html",
+          new_state_return_url: "javascript:document.querySelector('#start-again-form').submit()",
+          state_name: @state_name,
+          user_name: @user_name
+        )%></h2>
```
jenny-heath, line 36: hm it would be nice if we could avoid having a submit button in an h2, is there a reason this can't be a button with h2 style?

jenny-heath (approved):

app/views/state_file/questions/landing_page/edit.html.erb
tofarr: The reason I used a link was that this matches the mock - if you use a button then the inline styling becomes more complicated (Getting it to appear inline rather than as a block cross browser is tricky)

PR #4431, opened 2024-04-01
Nyc county warning #187296817

tofarr:
The order of the nyc-residency section was updated, a note was added, and copy was updated:
image

In the ny-county page, you get a warning message if the county you selected does not match your NYC residency status. This does not prevent proceeding to the next step:
image

## Implementation
* WarningForSelectComponent.js implements a general pattern of looking for elements with a `dataset-warning-for-select` attribute on page load, and then adding a listener that checks its current value against a list of pre-approved values
* `warning_for_select` added to vita_min_form_builder.rb generates html for a warning that will be picked up by the component.


app/views/state_file/questions/ny_county/edit.html.erb
```diff
@@ -14,4 +20,37 @@
     <% end %>
     <%= f.continue %>
   <% end %>
+
+  <script type="text/javascript">
+    window.addEventListener("load", () => {
+      const nycCounties = [
+          "Bronx",
+          "Kings (Brooklyn)",
+          "Manhattan (see New York)",
```
rickreyhsig, line 29: I know this is verbatim from the ticket but, I think we can just use Manhattan (New York) since it is the Manhattan Borough, New York County.

tofarr: The problem here is that the list of counties is taken from here:
https://github.com/codeforamerica/vita-min/blob/main/app/lib/efile/ny/school_districts.csv

I'm not sure if we should update this list. Maybe product management can give more info.

app/views/state_file/questions/ny_county/edit.html.erb
```diff
@@ -1,11 +1,52 @@
...
+              "Richmond (Staten Isand)"
+          ]
+          const nycResidency = <%= @nyc_residency == "full_year" %>
+
+          function displayWarnings(){
+              const value = document.querySelector("#state_file_ny_county_form_residence_county").value
+              const inNyc = nycCounties.includes(value)
+              const fullYearWarning = document.querySelector("#nyc_residency_full_year_warning")
+              const noneWarning = document.querySelector("#nyc_residency_none_warning")
+              if (nycResidency === inNyc) {
```
jenny-heath, line 24: i think i kind of see how this could work but would it not be more accurate to say `nycResidency && inNyc`? using `===` is confusing imo

also, very nitpicky: i think `selectedNycCounty` is a somewhat more intuitive name than `inNyc` and `nycResidency` could be clearer as `nycResidencyFullYear` (because there's a part year option)

app/views/state_file/questions/ny_county/edit.html.erb
```diff
@@ -1,11 +1,52 @@
...
+              "Bronx",
+              "Kings (Brooklyn)",
+              "Manhattan (see New York)",
+              "New York (Manhattan)",
+              "Queens",
+              "Richmond (Staten Isand)"
+          ]
+          const nycResidency = <%= @nyc_residency == "full_year" %>
+
+          function displayWarnings(){
```
jenny-heath, line 19: so....i have not written javascript in a long time but should there be some semicolons in here...?

app/views/state_file/questions/ny_county/edit.html.erb
```diff
@@ -1,11 +1,52 @@
...
+              const value = document.querySelector("#state_file_ny_county_form_residence_county").value
+              const inNyc = nycCounties.includes(value)
+              const fullYearWarning = document.querySelector("#nyc_residency_full_year_warning")
+              const noneWarning = document.querySelector("#nyc_residency_none_warning")
```
jenny-heath, line 23: so....i have not written javascript in a long time but should there be some semicolons in here...? or does being in a script tag change things?

app/views/state_file/questions/ny_county/edit.html.erb
jenny-heath, line 1: i think it would be good to write a little feature spec for this

app/views/state_file/questions/ny_county/edit.html.erb
tofarr: For booleans a and b, `a === b` != `a && b` (We don't want to show the warning when both a and b are false) - I suppose we could do `!(a ^ b)`, but I don't consider this clearer.

I've updated the variable names as suggested.

PR #4434, opened 2024-04-02
fix regex for daily pipeline tests

tahsinaislam:


jnf (approved):
hah. it's always regex. nice find.

PR #4435, opened 2024-04-02
Retry sending events to mixpanel #187354762

tofarr:
It looks like every so often, the connection to mixpanel is interrupted and events do not send. This interruption seems to last for less than a minute at a time:

https://app.datadoghq.com/logs?query=service%3A%28vita-min-prod-worker%20OR%20vita-min-prod-web%29%20status%3Aerror%20consume%20tracking%20event%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1711572390854&to_ts=1712177190854&live=true

image

This PR introduces a strategy of retrying 3 times with a 30 second wait in between  when this happens. I substituted out the `Concurrent::Future` for a `Concurrent::ScheduledTask` in order to accomplish this. (It could have been done with activejob but I think that is probably overkill)

I also added specs to verify the 3 retries are occurring, and that that regular requests do not retry.


rickreyhsig (approved):
LGTM

PR #4436, opened 2024-04-02
Fix for landing page redirect loop #187357814

tofarr:
* Fix issue where landing page is considered a step
* Fix issue for folks stuck in landing page loop
* Fix for issue where hitting back and trying to log into the same intake again deletes the intake


tofarr:
Co-authored-by: mpidcock <mpidcock@codeforamerica.org>

PR #4442, opened 2024-04-04
Submission search #187278919

tofarr:
Add search to submissions page
image

### I do a rather weird join...

* I build a union of all the columns I want from state intake tables, and then join the submissions to this.
* I can then query on this join
* I load data from my join table rather than relations (Not doing `submission.data_source.email_address` means we don't need to load data sources individually)
* This means we have 1 query only (2 queries if you count the existing `includes`

###

### DIFF
--- a/app/controllers/hub/state_file/efile_submissions_controller.rb
+++ b/app/controllers/hub/state_file/efile_submissions_controller.rb
@@ -5,38 +5,22 @@ module Hub
       before_action :load_efile_submissions, only: [:index]
 
       def index
-        EfileSubmission.joins(<<~SQL
+        @efile_submissions = EfileSubmission.joins(<<~SQL
           INNER JOIN (
-            SELECT state_file_az_intakes.id, state_file_az_intakes.email_address FROM state_file_az_intakes WHERE efile_submissions.data_source_type = 'StateFileAzIntake' AND state_file_az_intakes.id = efile_submissions.data_source_id
+            SELECT state_file_az_intakes.id as intake_id, 'StateFileAzIntake' as ds_type, state_file_az_intakes.email_address FROM state_file_az_intakes
             UNION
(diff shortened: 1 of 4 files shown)


app/views/hub/state_file/efile_submissions/index.html.erb
```diff
@@ -73,8 +79,8 @@
           <td class="index-table__cell">
             <%= submission.irs_submission_id.present? ? link_to(submission.irs_submission_id, hub_state_file_efile_submission_path(id: submission.id)) : "" %>
           </td>
-          <td class="index-table__cell"><%= submission.data_source.state_name %></td>
-          <td class="index-table__cell"><%= submission.data_source.email_address %></td>
+          <td class="index-table__cell"><%= StateFileBaseIntake::STATE_CODE_AND_NAMES[submission.data_source_state_code] %></td>
+          <td class="index-table__cell"><%= submission.email_address %></td>
```
tofarr, line 83: Not using `submission.data_source.email_address` here means that we don't load the intake individually for each submission - n+1 queries becomes a single query.

app/controllers/hub/state_file/efile_submissions_controller.rb
```diff
@@ -5,8 +5,24 @@ class EfileSubmissionsController < Hub::StateFile::BaseController
...
+        search = params[:search]
+        if search.present?
+          @efile_submissions = @efile_submissions.where("email_address LIKE ? OR irs_submission_id LIKE ?", "%#{search}%", "%#{search}%")
+          if search.to_i.to_s == search
+            @efile_submissions = @efile_submissions.where("id LIKE ? OR intake_id LIKE ?", search, search)
+          end
+        end
+
         @efile_submissions = @efile_submissions.includes(:efile_submission_transitions).reorder(created_at: :desc).paginate(page: params[:page], per_page: 30)
         @efile_submissions = @efile_submissions.in_state(params[:status]) if params[:status].present?
```
embarnard, line 24: should this be cleared if they are making another search?

embarnard (approved):

PR #4443, opened 2024-04-04
Refactor IncomingEmail webhook controller

rickreyhsig:
https://www.pivotaltracker.com/story/show/187187633

### DIFF
--- a/app/controllers/mailgun_webhooks_controller.rb
+++ b/app/controllers/mailgun_webhooks_controller.rb
@@ -13,7 +13,9 @@ class MailgunWebhooksController < ActionController::Base
     sender_email_address = parse_valid_email_address(from: params["from"], sender: params["sender"])
     clients = Client.joins(:intake).where(intakes: { email_address: sender_email_address })
     client_count = clients.count
-    if client_count.zero?
+    if sender_email_address.is_FYST?
+      # Create incoming EMail for FYST
+    elsif client_count.zero?
       archived_intake = most_recent_intake(sender_email_address)
       if archived_intake.present?
         locale = archived_intake.locale || "en"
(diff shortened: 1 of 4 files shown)


app/models/application_record.rb
```diff
@@ -1,6 +1,7 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
+  STATE_INTAKE_CLASS_NAMES = %w[StateFileAzIntake StateFileNyIntake].map(&:to_s).freeze
```
rickreyhsig, line 4: Moving this here to make it available to all our models instead of just `EfileSubmission`

spec/models/efile_submission_spec.rb
```diff
@@ -706,4 +706,17 @@
       end
     end
   end
+
```
rickreyhsig, line 709: Adding this spec since I moved `STATE_INTAKE_CLASS_NAMES` form the `efile_submission` model, and I wanted to make sure I didn't break anything.

app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -158,6 +159,16 @@ def update_outgoing_email_status
 
   private
 
+  def re_optin_when_client_replies
+    sender_email_address = parse_valid_email_address(from: params["from"], sender: params["sender"])
+    opted_out_state_intakes = StateFileBaseIntake.opted_out_state_file_intakes(sender_email_address)
+    unless opted_out_state_intakes.empty?
+      opted_out_state_intakes.each do |intake|
```
rickreyhsig, line 166: Not sure there's better logic here for updating only one intake, but since we don't have a unique validation constraint on the intake, there could be multiple intakes with the same email. I kept the pattern of creating multiple incoming_emails established in the `MailgunWebhooksController`

tofarr: I think updating all of them rather than just the last one is absolutely the right call

tofarr (approved):

PR #4447, opened 2024-04-05
Change logic for Reminder notification

rickreyhsig:
https://www.pivotaltracker.com/story/show/187364441


lib/tasks/efile.rake
```diff
@@ -23,8 +23,4 @@ namespace :efile do
     puts("End: Found #{EfileSubmission.in_state(:failed).count} failed submissions")
     exit
   end
-
```
rickreyhsig, line 26: Forgot to remove this in the last PR when it moved from efile to state_file
https://github.com/codeforamerica/vita-min/blob/dd6d1ed7f155503f2c0dc218b58916c9b360f445/lib/tasks/state_file.rake

app/services/state_file/reminder_to_finish_state_return_service.rb
```diff
@@ -3,15 +3,13 @@ class ReminderToFinishStateReturnService
     def self.run
       cutoff_time_ago = 12.hours.ago
       batch_size = 10
-      intakes_with_no_submission = StateFileAzIntake.where("df_data_imported_at < ?", cutoff_time_ago)
-                                                    .left_joins(:efile_submissions)
-                                                    .where(efile_submissions: { id: nil })
-                                                    .where.not("state_file_az_intakes.message_tracker #> '{messages.state_file.finish_return}' IS NOT NULL")
+      intakes_with_no_submission = StateFileAzIntake.where('state_file_az_intakes.created_at < ?', cutoff_time_ago)
+                                                    .where.not(email_address: nil).where.not(email_address_verified_at: nil)
```
mpidcock, line 7: do we also need to check `.where(unsubscribed_from_email: false)` ? 
asking since I saw it on the other PR #4475

app/services/state_file/reminder_to_finish_state_return_service.rb
```diff
@@ -3,15 +3,13 @@ class ReminderToFinishStateReturnService
...
+                                                    .where("state_file_az_intakes.message_tracker #> '{messages.state_file.finish_return}' IS NULL")
 
-      intakes_with_no_submission += StateFileNyIntake.where("df_data_imported_at < ?", cutoff_time_ago)
-                                                     .left_joins(:efile_submissions)
-                                                     .where(efile_submissions: { id: nil })
-                                                     .where.not("state_file_ny_intakes.message_tracker #> '{messages.state_file.finish_return}' IS NOT NULL")
+      intakes_with_no_submission += StateFileNyIntake.where('state_file_ny_intakes.created_at < ?', cutoff_time_ago)
+                                                     .where.not(email_address: nil).where.not(email_address_verified_at: nil)
```
mpidcock, line 11: same as above

app/services/state_file/reminder_to_finish_state_return_service.rb
```diff
@@ -3,15 +3,13 @@ class ReminderToFinishStateReturnService
     def self.run
       cutoff_time_ago = 12.hours.ago
       batch_size = 10
-      intakes_with_no_submission = StateFileAzIntake.where("df_data_imported_at < ?", cutoff_time_ago)
-                                                    .left_joins(:efile_submissions)
-                                                    .where(efile_submissions: { id: nil })
-                                                    .where.not("state_file_az_intakes.message_tracker #> '{messages.state_file.finish_return}' IS NOT NULL")
+      intakes_with_no_submission = StateFileAzIntake.where('state_file_az_intakes.created_at < ?', cutoff_time_ago)
```
mpidcock, line 6: Should consider doing the same iteration over the state intake class names as in #4475 and having `#{base_class.underscore.pluralize}` instead of separate steps for AZ & NY. Would be more consistent and any changes would be applied in one place rather than two.

mpidcock (approved):
Some changes to consider

app/services/state_file/reminder_to_finish_state_return_service.rb
rickreyhsig: Good call, I will add.

app/services/state_file/reminder_to_finish_state_return_service.rb
rickreyhsig: Thanks for raising, this was the first one that was done, so the latter ones have improvements.  I'll update!

PR #4462, opened 2024-04-09
truncate 3rd party designee name

tahsinaislam:


jenny-heath (approved):
maybe in future we can add a test for this case! (approving since we're short on time now)

PR #4468, opened 2024-04-10
Fix submission and skip flaky tests #187278919

tofarr:
* Fix blank submissions page (The join was incorrect and just happened to seem to work in local / staging - there was a test that caught this but circleci thought it was flaky: `efile_submissions.id -> efile_submissions.data_source_id`)
* Skip flaky tests


rickreyhsig (approved):
lgtm

PR #4469, opened 2024-04-10
Pre-Deadline notification

rickreyhsig:
https://www.pivotaltracker.com/story/show/187222780

### DIFF
--- a/app/services/state_file/send_pre_deadline_reminder_service.rb
+++ b/app/services/state_file/send_pre_deadline_reminder_service.rb
@@ -3,7 +3,7 @@ module StateFile
     BATCH_SIZE = 10
     HOURS_AGO = 24
 
-    def run
+    def self.run
       cutoff_time_ago = HOURS_AGO.hours.ago
       intakes_to_notify = []
 
--- a/app/services/state_file/send_pre_deadline_reminder_service.rb
+++ b/app/services/state_file/send_pre_deadline_reminder_service.rb
@@ -6,7 +6,7 @@ module StateFile
     def run
       cutoff_time_ago = HOURS_AGO.hours.ago
       intakes_to_notify = []
-      ApplicationRecord::STATE_INTAKE_CLASS_NAMES.map do |base_class|
+      ApplicationRecord::STATE_INTAKE_CLASS_NAMES.each do |base_class|
         class_object = base_class.constantize
         intakes_to_notify += class_object.left_joins(:efile_submissions)
                                          .where(efile_submissions: { id: nil })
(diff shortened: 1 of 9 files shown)


app/services/state_file/send_pre_deadline_reminder_service.rb
```diff
@@ -0,0 +1,46 @@
+module StateFile
+  class SendPreDeadlineReminderService
+    BATCH_SIZE = 10
+    HOURS_AGO = 24
+
+    def run
+      cutoff_time_ago = HOURS_AGO.hours.ago
+      intakes_to_notify = []
+      ApplicationRecord::STATE_INTAKE_CLASS_NAMES.map do |base_class|
```
tofarr, line 9: Probably could use `each` rather than `map` here. Unless you want to do...
```
intakes_to_notify = ApplicationRecord::STATE_INTAKE_CLASS_NAMES.map do |base_class|
  base_class.constantize.left_joins( ...
end.flatten
```

app/services/state_file/send_pre_deadline_reminder_service.rb
```diff
@@ -0,0 +1,46 @@
...
+        class_object = base_class.constantize
+        intakes_to_notify += class_object.left_joins(:efile_submissions)
+                                         .where(efile_submissions: { id: nil })
+                                         .where("#{base_class.underscore.pluralize}.created_at < ?", cutoff_time_ago)
+                                         .where.not(email_address: nil).where.not(email_address_verified_at: nil)
+                                         .where(unsubscribed_from_email: false)
+                                         .where.not("#{base_class.underscore.pluralize}.message_tracker #> '{messages.state_file.pre_deadline_reminder}' IS NOT NULL")
+      end
```
tofarr, line 17: Not a blocker, but you could do a db filter (Like you suggested below). I think the post-deadline branch I created has an example

tofarr (approved):

app/services/state_file/send_pre_deadline_reminder_service.rb
rickreyhsig: Good catch, done.

app/services/state_file/send_pre_deadline_reminder_service.rb
rickreyhsig: I think what makes the part difficult is that the messages.state_file.pre_deadline_reminder value part of the hash is stored as a string and not a date.  It needs to be casted.

rickreyhsig: Done

PR #4476, opened 2024-04-11
Add js component: dollar amount rounding on focusout

jenny-heath:
requirements here: https://www.pivotaltracker.com/story/show/187398408

### DIFF
--- a/app/javascript/components/DollarAmountRoundingComponent.js
+++ b/app/javascript/components/DollarAmountRoundingComponent.js
@@ -1,17 +1,14 @@
 export default function DollarAmountRoundingComponent() {
-    const dollarInputs = document.querySelectorAll(".round-dollar-amount");
+    function replaceValue() {
+        const value = this.value;
 
-    dollarInputs.forEach((inputElement) => {
-        function render() {
-            const value = inputElement.value;
-
-            if (!isNaN(value) && !isNaN(parseFloat(value))) {
-                const rounded_value = Math.round(value);
-                inputElement.value = rounded_value;
-            }
+        if (!isNaN(value) && !isNaN(parseFloat(value))) {
+            const rounded_value = Math.round(value);
+            this.value = rounded_value;
         }
+    }
 
-        inputElement.addEventListener("focusout", render);
-        render();
+    document.querySelectorAll(".round-dollar-amount").forEach((inputElement) => {
+        inputElement.addEventListener("focusout", replaceValue);
     });
 }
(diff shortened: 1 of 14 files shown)


app/javascript/components/DollarAmountRoundingComponent.js
```diff
@@ -0,0 +1,17 @@
+export default function DollarAmountRoundingComponent() {
+    const dollarInputs = document.querySelectorAll(".round-dollar-amount");
+
+    dollarInputs.forEach((inputElement) => {
+        function render() {
+            const value = inputElement.value;
```
jnf, line 6: seems like it's mixing scopes here a bit. you're making a new `render` in the local scope of each iteration, but then referencing `inputElement` here, necessitating a closure. i am curious if you can accomplish the same with `this.value` instead of the closure ([reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_value_of_this_within_the_handler)).

another path would be to reference `event` in the callback (i.e., `functions render (event) { ...`).  Then you'd have access to `event.currentTarget`, which would be the input in question.

either path you wouldn't need the closure and could have a single `render` function expression that works for every input. it'd be just a wee bit less memory and a wee bit less scope complexity.

app/javascript/components/DollarAmountRoundingComponent.js
```diff
@@ -0,0 +1,17 @@
...
+
+            if (!isNaN(value) && !isNaN(parseFloat(value))) {
+                const rounded_value = Math.round(value);
+                inputElement.value = rounded_value;
+            }
+        }
+
+        inputElement.addEventListener("focusout", render);
+        render();
```
jnf, line 15: why do you need to invoke render here? is there some concern that clients could input numbers before the handler was attached?

jnf (commented):
question -- is a corresponding backend rounding hook/validation coming in this PR or separately?

app/javascript/components/DollarAmountRoundingComponent.js
jenny-heath: also copy-pasta, i could have sworn i tried to remove it and it stopped working but that's probably...not what happened

jenny-heath:
> question -- is a corresponding backend rounding hook/validation coming in this PR or separately?

@jnf since the previous state of these fields was that we were requiring clients to enter integers themselves, the validations for that are already in place! (...except for one, which i just added, because i realized i shouldn't post this comment without double checking them all first. does that beg the question "have we been validating that field correctly?"? yes it does.)

jenny-heath:
not gonna do this before season closes

PR #4477, opened 2024-04-11
[AFTER TAX SEASON] updates lograge to ignore the apitble healthcheck request

jnf:
we already ignore this for mixpanel, but we've been sending thousands and thousands of healthcheck events to datadog.

### DIFF
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -11,6 +11,9 @@ Rails.application.configure do
   # Log to the same place as Rails logs
   config.lograge.logger = Rails.logger
 
+  # Let's not log the healthcheck (and maybe other things, eventually)
+  config.lograge.ignore_actions = ['PublicPagesController#healthcheck']
+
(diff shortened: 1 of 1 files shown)


config/initializers/lograge.rb
```diff
@@ -11,6 +11,9 @@
   # Log to the same place as Rails logs
   config.lograge.logger = Rails.logger
 
+  # Let's not log the healthcheck (and maybe other things, eventually)
+  config.lograge.ignore_actions = ['PublicPagesController#healthcheck']
```
jamesiarmes, line 15: These events can be valuable, especially when investigating networking issues between the load balancer and the application.

Could we add an environment variable that we can use to enable logging if needed? Something like:

```suggestion
  config.lograge.ignore_actions = ['PublicPagesController#healthcheck'] unless ENV.fetch('ENABLE_HEALTHCHECK_LOGGING', false)
```

jnf:
closing for now; will handle in fyst-specific repo

PR #4478, opened 2024-04-11
WIP: monetary values error messages

tahsinaislam:


spec/models/state_file1099_g_spec.rb
```diff
@@ -85,13 +85,13 @@
     end
     it "validates federal_income_tax_withheld" do
       state_file_1099.federal_income_tax_withheld = nil
-      expect(state_file_1099.save).to eq false
+      expect(state_file_1099.save).to eq true
```
rickreyhsig, line 88: This is a good spec for verifying that we accept blanks. I would add at least one spec where we assert on the scenario for `must be greater than or equal to 1` that you have updated.

rickreyhsig (approved):

PR #4480, opened 2024-04-12
Close FYST 2 #187355990

tofarr:
More copy updates.

Removed this paragraph:
image

And added copy to the rejected page:
image


embarnard (approved):
Looks good!

PR #4489, opened 2024-04-13
Add env variable DO_NOT_SEND_APOLOGY_EMAIL

mpidcock:


jenny-heath:
we subsequently realized while testing this on demo that it throws an [error](https://codeforamerica.sentry.io/issues/5190882467) but we're working on a saturday so we've decided that since it still effectively stops the next line from running we're gonna leave it as is for now

PR #4490, opened 2024-04-14
try using gyr email for verification emails

jenny-heath:


jenny-heath:
thankfully we never had to do this

PR #4493, opened 2024-04-15
Replace the error icon #187092299

tofarr:
Error Page for statefile:
image

Error page for GYR is unchanged:
image


tahsinaislam (approved):

jnf (commented):
let's hold off on adding any conditional switching between gyr/fyst until we have a picture of if/how/when we're separating the products.

jnf (changes requested):
hold for now

PR #4504, opened 2024-04-17
Prevent accepted resubmit #187434120

tofarr:
* Do not resubmit if there is already an accepted return with the same SSN.
* Added a feature flag so we can turn this off if it is interfering with testing

### DIFF
--- a/app/forms/state_file/esign_declaration_form.rb
+++ b/app/forms/state_file/esign_declaration_form.rb
@@ -48,6 +48,7 @@ module StateFile
         .joins(:efile_submission_transitions)
         .where(efile_submission_transitions: { to_state: :accepted })
         .where(table_name => { hashed_ssn: intake.hashed_ssn })
+        .where.not(efile_submissions: {id: intake.id})
     end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/forms/state_file/esign_declaration_form_spec.rb
```diff
@@ -122,6 +122,40 @@
...
+        expect(form).to be_valid
+        expect {
+          form.save
+        }.to change(intake.efile_submissions, :count).by(0)
+        expect(intake.reload.efile_submissions.last.current_state).to eq("accepted")
+      end
+    end
+
+    context "when there is already a non-accepted submission in a different account with the same SSN" do
+      #let!(:efile_submission) { create :efile_submission, :accepted, :for_state, data_source: intake }
```
rickreyhsig, line 144: Can we nuke this comment?

app/forms/state_file/esign_declaration_form.rb
```diff
@@ -17,6 +17,13 @@ def save
       efile_info = StateFileEfileDeviceInfo.find_by(event_type: "submission", intake: @intake)
       efile_info&.update!(attributes_for(:state_file_efile_device_info))
 
+      unless Flipper.enabled?(:allow_duplicate_submissions)
+        if accepted_submissions_with_same_ssn(@intake).any?
+          Rails.logger.warn "#{@intake.state_code}#{@intake.id} was not submitted because there is already an accepted submission for that ssn"
```
jnf, line 22: is this sufficient notification? would it be worth elevating this to a slack message or sentry warning or similar?

PR #4505, opened 2024-04-17
On state file login, if multiple intakes match email+ssn, log in the one with an accepted return (if there is one)

jenny-heath:


app/forms/state_file/intake_login_form.rb
```diff
@@ -11,10 +11,19 @@ def intake
       @intake
     end
 
+    def intake_to_log_in(matching_intakes)
```
jenny-heath, line 14: i made this a method because there are at least 2 places where we presume that the intake in question is the first of a group of them that match certain criteria. i don't feel 100% great about the way this looks—it doesn't seem that clear or easy to find, not sure if the form is the best place for it—but there's a lot going on in login-land so i couldn't think of a more elegant solution

app/forms/state_file/intake_login_form.rb
```diff
@@ -11,10 +11,19 @@ def intake
...
+      with_accepted_submissions = matching_intakes.filter { |intake| intake.latest_submission&.in_state?(:accepted) }
```
rickreyhsig, line 15: Would we need to look at all submissions for this? If for w/e reason the client submits and erroneous submission after acceptance or submits more than one and the one accepted is not the chronological last submission, this logic could return an undesired intake or not at all.

rickreyhsig:
Would we need to use `intake_to_log_in()` logic in `submission_to_show` method of the `app/controllers/state_file/questions/return_status_controller.rb`?

rickreyhsig:
I've been able to reproduce the issue locally by running the following steps:

```ruby
require 'factory_bot'
FactoryBot.find_definitions
include FactoryBot::Syntax::Methods

intake = create(:state_file_az_intake)
intake.update(email_address: 'test_email@codeforamerica.org', email_address_verified_at: Time.now, hashed_ssn: SsnHashingService.hash(555002222))

sub = create(:efile_submission, :transmitted)
sub.update(data_source_id: intake.id, data_source_type: "StateFileAzIntake")

intake_two = create(:state_file_az_intake)
intake_two.update(email_address: 'rkreyhsig+02@codeforamerica.org', email_address_verified_at: Time.now, hashed_ssn: SsnHashingService.hash(555002222))

sub = create(:efile_submission, :transmitted)
sub.update(data_source_id: intake_two.id, data_source_type: "StateFileAzIntake")

sub = create(:efile_submission, :accepted)
sub.update(data_source_id: intake_two.id, data_source_type: "StateFileAzIntake")

# Login at:
http://statefile.localhost:3000/en/us/login-options
# Verify you get the first (WRONG) intake on main
# Verify you STILL get the accepted return on this branch
```

mpidcock (approved):

PR #4506, opened 2024-04-17
Lockout failing for AZ #187437777

tofarr:
There was a logic error in the code on Line 77 - the loop redefines the `@records` variable on each iteration.
The first iteration is AZ, then NY - so it will never include records for AZ if used later.
 was redefined on each iteration,was called for AZ and then NY to get these records. The  and would be used at the end. This means that AZ intakes would never lock out, because the `@records` collection would never contain intakes for AZ


embarnard (approved):
Looks good!

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -73,10 +73,10 @@ def intake_login_params
     def increment_failed_attempts_on_login_records
       contact_info = params[:portal_verification_code_form][:contact_info]
       intake_classes = client_login_service.intake_classes
-      intake_classes.each do |intake_class|
-        @records = intake_class.where(email_address: contact_info).or(intake_class.where(phone_number: contact_info))
-        @records.map(&:increment_failed_attempts)
-      end
+      @records = intake_classes.map do |intake_class|
```
jnf, line 76: good candidate for `flat_map`

tofarr: Good call - done

PR #4516, opened 2024-04-23
Follow up to re-opting in opt-outs

rickreyhsig:
https://www.pivotaltracker.com/story/show/187407168


app/controllers/mailgun_webhooks_controller.rb
```diff
@@ -163,11 +163,19 @@ def update_outgoing_email_status
...
     opted_out_state_intakes = StateFileBaseIntake.opted_out_state_file_intakes(sender_email_address)
+    opted_out_gyr_intakes = Intake::GyrIntake.opted_out_gyr_intakes(sender_email_address)
+
     unless opted_out_state_intakes.empty?
       opted_out_state_intakes.each do |intake|
         intake.update(unsubscribed_from_email: false)
       end
     end
+
+    unless opted_out_gyr_intakes.empty?
```
rickreyhsig, line 174: I think .each already takes care if the list is empty

rickreyhsig: Done

PR #4517, opened 2024-04-24
Allow CTC clients from 2022 product year to login and download tax return

embarnard:
- [x] Link to pivotal/JIRA issue: https://www.pivotaltracker.com/story/show/187338000
- [x] What was done?
1. allow CTC clients from the 2022 tax year to login and download their tax return
2. Remove the review box from the portal page

There was another implementation that I was working on here: https://github.com/codeforamerica/vita-min/pull/4470/files that aimed at changing the login system and portal so that intakes from the Archived Intakes model could login and download their returns. But after talking to Tiffany, it seems like there might be some security concerns when matching clients that have filed for multiple years, so we decided to abandon it in favor of just serving 2022 clients. The approach in this PR is much simpler and falls in line with our existing architecture so there weren't many tests needed. 

- [x] How to test?
    I tested it using manual tests from the seeder. I had to specify the 2022 product year in the seeder file so that it would show up as an accessible intake for the client login service. If you would like to test yourself you can login with the user: yfong+eitcundertwentyfourqc@codeforamerica.org, verification code: 000000, client id: 9

 - [x] Screenshots (for visual change)
Before:
Screenshot 2024-04-24 at 5 23 57 PM

After: 
Screenshot 2024-04-24 at 5 24 20 PM


mpidcock (approved):
Reviewed w/Mike

PR #4529, opened 2024-05-13
experiment with sending email via the sendgrid api

mrotondo:


tofarr:
Martha and I are wondering if there a ticket associated with this? How can we track where this came from and how it is proceeding?

mrotondo:
ah sorry, this was purely a spike and can get deleted. i didn't actually mean to make it a PR, just a branch - must have autopiloted

PR #4532, opened 2024-05-16
redirect back to hub client page after successfully saving any 13614-C page

mrotondo:
## What was done?
    - After saving on any of the three 13614-C pages, we redirect to the client page (where previously we redirected back to the same 13614-C page)
    - As I noted in the JIRA ticket, navigating between the 13614-C pages trashes your changes unless you hit Save first. With this change in place, a user who wants to edit multiple 13614-C pages will now have to click from the client into the 13614-C form, edit one page, hit Save, then re-enter the form, navigate to another page and edit that one. Thus, this change removes a step for all users (by navigating them back to the client page after their last save) but adds 1-2 steps for anyone who wants to edit multiple 13614-C pages. 
    - We could also add a "Return to Client" button to the top of all 13614-C pages, if we think this change will be disruptive to enough Hub users.
    - I also added "page #" and "pagina #" text to the titles for page 2 and 3 of the 13614-C form, to match page 1.
## How to test?
        - I manually tested the change by editing 13614-C pages on my local Hub, and verified that I was redirected to the client page only if a save is successful (from any 13614-C page).
        - I am currently working on modifying the existing integration test so that it passes with the new functionality.
    - This change should be able to be acceptance tested on Heroku
    - Risk Assessment
        - I don't believe there are any risks from this change other than the UX impact I described above.
## Other notes
    - I also noticed that the dropdowns on pages 2 and 3 of the 13614-C are poorly laid out, so that you can't see the selected values on some of them and the value is visually truncated on others. Shouldn't be part of this change, but we could make another story for that.

### DIFF
--- a/app/controllers/hub/clients_controller.rb
+++ b/app/controllers/hub/clients_controller.rb
@@ -131,7 +131,7 @@ module Hub
         SystemNote::ClientChange.generate!(initiated_by: current_user, intake: @client.intake)
         GenerateF13614cPdfJob.perform_later(@client.intake.id, "Hub Edited 13614-C.pdf")
         flash[:notice] = "Changes saved"
-        redirect_to edit_13614c_form_page1_hub_client_path(@client)
+        redirect_to hub_client_path(id: @client.id)
       else
         flash[:alert] = I18n.t("forms.errors.general")
         render :edit_13614c_form_page1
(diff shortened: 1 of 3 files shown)


tofarr (approved):

tofarr:
LGTM! 🍰

PR #4534, opened 2024-05-16
Added active flag to Unique Links

tofarr:
## What was done?
 * Unique Links in /en/hub/organizations now has a toggle indicating whether it is active
 * I added a migration to add an active flag to the `SourceParemeter` model (boolean default: true)
 * The model name is an oddity - `SourceParameter` instead of `UniqueLink`.
 * The attribute for link is `code` which is weird. It is not validated to be a URL.
## How to test?
 * Test on Heroku: https://gyr-review-app-4534-60ec07194217.herokuapp.com/
 * Log in to the hub and go to /en/hub/organizations
 * Enable / Disable links, reloading the page between. (Demonstrating that values are persisted)
 * Create / Delete links to verify this still works.
## Risk Assessment
 * I can't see any associated increased risk of heart disease or diabetes.
### Before
image
### Hub Page
image
### When you visit a URL with an active source:
e.g.: https://gyr-review-app-4534-60ec07194217.herokuapp.com/en/foo
image

### When you visit a URL with an inactive source:
image

[GYR1-426]: https://codeforamerica.atlassian.net/browse/GYR1-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/hub/source_params_controller.rb
+++ b/app/controllers/hub/source_params_controller.rb
@@ -16,6 +16,14 @@ module Hub
       end
     end
 
+    def update
+      @source_param = SourceParameter.find(params[:id])
+      @source_param.update(active: params[:source_parameter][:active] == "1")
+      respond_to do |format|
+        format.js
+      end
+    end
+
     def destroy
       @source_param = SourceParameter.find(params[:id])
 
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -15,10 +15,14 @@ class PublicPagesController < ApplicationController
 
     # redirect to home hiding original source param
     if params[:source].present?
-      vita_partner = SourceParameter.find_vita_partner_by_code(params[:source])
-      if vita_partner.present?
-        flash[:notice] = I18n.t("controllers.public_pages.partner_welcome_notice", partner_name: vita_partner.name)
-        cookies[:used_unique_link] = { value: "yes", expiration: 1.year.from_now.utc }
+      source_parameter = SourceParameter.find_by(code: params[:source]&.downcase)
+      if source_parameter.present?
+        if source_parameter.active?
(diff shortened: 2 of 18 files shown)


spec/models/source_parameter_spec.rb
```diff
@@ -24,6 +25,7 @@
   it { should validate_presence_of(:code) }
 
   describe ".find_vita_partner_by_code" do
+    # make sure this tests both inactive and active cases
```
mpidcock, line 28: `# todo` ?

mpidcock (commented):
there's a few minor things to fix before merging

app/controllers/hub/source_params_controller.rb
```diff
@@ -16,6 +16,14 @@ def create
       end
     end
 
+    def update
+      @source_param = SourceParameter.find(params[:id])
+      @source_param.update(active: params[:source_parameter][:active] == "1")
```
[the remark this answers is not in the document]
tofarr: As mentioned earlier, I added some specs to cover this case - `.active?` does not work.

app/controllers/public_pages_controller.rb
```diff
@@ -17,7 +17,7 @@ def home
     if params[:source].present?
       source_parameter = SourceParameter.find_by(code: params[:source]&.downcase)
       if source_parameter.present?
-        if source_parameter.active?
+        if source_parameter.active? && source_parameter.vita_partner.present?
```
mpidcock, line 20: I meant to indicate I this check should go one layer up since the `else` statement also uses `source_parameter.vita_partner.name`. I think that's appropriate, because if there's no partner then it means we will just redirect

mpidcock (approved):
One more thing to consider changing but looks good

app/controllers/public_pages_controller.rb
tofarr: Fixed in second PR

PR #4538, opened 2024-05-17
GYR1-136 disable duplicate clicks

tofarr:
## What was done?
* Using a rails FormBuilder.button adds code that disables the button after it is clicked, thereby preventing duplicate clicks on slow systems. *We should use this by default in all cases instead of the `<button />` tag, which does not prevent duplicate clicks.
 * Alternatively Ricardo created a [more far reaching branch](https://github.com/codeforamerica/vita-min/pull/4524/files) that locks records - I am unsure of the full ramifications of this (And what it will do about duplicates), so I did not merge it.
## How to test?
    - ???
    - ???


mrotondo (approved):
lgtm

PR #4548, opened 2024-05-28
GYR1-468 - Last Contact column updates even if message is undelivered 

tofarr:
## What was done?
 * `last_outgoing_communication_at` is now updated on save rather than create, and only when the message is `completed` / `delivered`
 * Added specs for undelivered interactions
 * Updated specs for delivered interactions
## How to test?
* Test on staging (Because emails do not send on heroku)
* Basically we want to show that the `last contact at` is updated only when messages are delivered.
* Go to a client that has opted in to contact by email, and one that has opted in to contact via sms. The screenshot shows one with a valid email address that has opted in to email image
* You can see the welcome email 
* Go to the messages tab image
* Fill in the form to send an email image
* The last contact at should not update until the message is successfully delivered 
image 
image

[GYR1-468]: https://codeforamerica.atlassian.net/browse/GYR1-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/models/outbound_call_spec.rb
```diff
@@ -25,10 +25,14 @@
 
 describe OutboundCall do
   it_behaves_like "a user-initiated outgoing interaction" do
-    let(:subject) { build(:outbound_call) }
+    let(:subject) { build(:outbound_call, twilio_status: "delivered") }
   end
 
   it_behaves_like "an outgoing interaction" do
-    let(:subject) { build :outgoing_text_message }
```
tofarr, line 32: I think this was a copy pasta error

spec/models/outgoing_email_spec.rb
```diff
@@ -39,7 +39,11 @@
     end
 
     it_behaves_like "an outgoing interaction" do
-      let(:subject) { build :outgoing_text_message }
```
tofarr, line 42: Again - looks like more copypasta

config/environments/staging.rb
```diff
@@ -3,12 +3,12 @@
 Rails.application.configure do
   config.active_storage.service = :s3_staging
 
-  config.ctc_url = "https://ctc.staging.getyourrefund.org"
-  config.statefile_url = "https://staging.fileyourstatetaxes.org"
-  config.gyr_url = "https://staging.getyourrefund.org"
-  ctc_email_from_domain = "getctc.org"
-  gyr_email_from_domain = "getyourrefund.org"
-  statefile_email_from_domain = "fileyourstatetaxes.org"
+  config.ctc_url = "https://ctc.demo.getyourrefund.org"
```
mrotondo, line 6: Did we get rid of staging already? I might just be out of the loop, but what is the basis for these changes?

mrotondo (approved):
Looks good, just would like a bit of context on the staging->demo changes as they don't seem to have any relation to the context described in this PR or attached story. Is there another story where those changes were requested?

PR #4560, opened 2024-06-03
GYR1-138 Searchable FAQ

tofarr:
## A search box was added to the FAQ page
image

* I added a search text box to the FAQ page and leverage postgresql full text search to scan the db for matches.
* I copy the existing full text search implementation from [Intakes](https://github.com/codeforamerica/vita-min/blob/2cf27b00dfbdcc01b1adb90e7eb7e5898a20bf5a/app/models/intake.rb#L319) There are a few small caveats:
* I was hoping to do this in pure SQL using postgres with generated columns, but could not because we want to search the plain text of a rich text field, and we do not store the plain text in the database. (It is actually stored in a different table)
* I had to create 2 separate search vectors for English and Spanish.
* I update the searchable vectors by executing an SQL function an after_save - this will make saves slightly slower but was necessary.
* I ignore the searchable vectors when loading models (There is no need to load this heavy data - only the database uses it)
* 
## How to test?
 * Test on Heroku: https://gyr-review-app-4560-22f87077ed1a.herokuapp.com/
 * Enter the hub and add some FAQ articles: 
image
 * Go to the FAQ page, and search for these (By question and answer - Spanish and English searches are independent)
image

[GYR1-138]: https://codeforamerica.atlassian.net/browse/GYR1-138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


embarnard (approved):
LGTM!

PR #4562, opened 2024-06-03
remove some unused messaging code and add a test

mrotondo:
## What was done?
- The messaging code removed in this PR is redundant with the implementation of the more-correct message sending logic in `ReminderToFinishStateReturnService`
- A test was added to make sure the service doesn't send messages to intakes with associated `efile_submissions`


mpidcock (approved):
Thanks for doing this 🙏

PR #4566, opened 2024-06-03
dynamically generate list of message classes for FYST Hub AutomatedMessages page

mrotondo:
WIP - PR description TBD
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-64
## What was done?
- I changed the contents of the two AutomatedMessages pages in the Hub (one for GYR/CTC, one for FYST), which show previews of all the email and SMS messages we send. They previously had been missing some messages. I changed their implementations so that most messages will now be included automatically if they are implemented as subclasses of our two main message classes, `AutomatedMessage::AutomatedMessage` or `StateFile::AutomatedMessage::BaseAutomatedMessage`. I also added another type of message to the page which had been missed previously.
- Alternatives considered: 
  - I did not consolidate all of our message sending to use one message class and mailer, although I do think that we should do that - both to make these pages fully automatic and to make our codebase more consistent and concise. 
  - I also did not take the smaller step of making all the AutomatedMessage-style subclasses share a common interface of class methods, although I also think that would be a good minor improvement.
  - I also did not add the templates that are available for various actions that can be taken by Hub users on client pages. If we would like to see those templates here, I can add them.
## How to test?
- I did not add any automated tests, but all of our tests still pass.
- To verify that the messages have been added, compare the AutomatedMessages pages on Demo
  - https://demo.getyourrefund.org/en/hub/automated_messages
  - https://demo.getyourrefund.org/en/hub/state-file/automated_messages
- with those on the acceptance deployment
  - https://gyr-review-app-4566-7c8a6daaf351.herokuapp.com/en/hub/automated_messages
  - https://gyr-review-app-4566-7c8a6daaf351.herokuapp.com/en/hub/state-file/automated_messages

### DIFF
--- a/app/controllers/hub/automated_messages_controller.rb
+++ b/app/controllers/hub/automated_messages_controller.rb
@@ -10,39 +10,24 @@ module Hub
     private
 
     def messages_preview
-      automated_messages = [
-        [AutomatedMessage::SuccessfulSubmissionDropOff, {}],
-        [AutomatedMessage::SuccessfulSubmissionOnlineIntake, {}],
-        [SurveyMessages::GyrCompletionSurvey, { survey_link: "https://fakecodeforamerica.co1.qualtrics.com" }],
-        [SurveyMessages::CtcExperienceSurvey, { survey_link: "https://fakecodeforamerica.co1.qualtrics.com" }],
-        [AutomatedMessage::DocumentsReminderLink, { body_args: { doc_type: "ID" } }],
-        [AutomatedMessage::EfileAcceptance, {}],
-        [AutomatedMessage::EfilePreparing, {}],
-        [AutomatedMessage::EfileRejected, {}],
-        [AutomatedMessage::EfileRejectedAndCancelled, {}],
-        [AutomatedMessage::EfileFailed, {}],
-        [AutomatedMessage::CtcGettingStarted, {}],
-        [AutomatedMessage::ClosingSoon, {}],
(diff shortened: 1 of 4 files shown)


app/controllers/hub/automated_messages_controller.rb
```diff
@@ -10,46 +10,23 @@ def index
...
-        [AutomatedMessage::NewPhotosRequested, {}],
-        [AutomatedMessage::VerificationAttemptDenied, {}],
-        [AutomatedMessage::Ctc2022OpenMessage, {}],
-        [AutomatedMessage::PuertoRicoOpenMessage, {}],
-        [AutomatedMessage::IntercomForwarding, {}],
-        [AutomatedMessage::UnmonitoredReplies, { support_email: Rails.configuration.email_from[:support][:gyr] }],
-        [AutomatedMessage::InProgress, {}],
-        [AutomatedMessage::Welcome, {}],
-      ]
+      Rails.application.eager_load!
```
tofarr, line 13: Will this make it likely that there is a second or so delay when this is run for the first time?

tofarr (approved):
🍰

app/controllers/hub/automated_messages_controller.rb
mrotondo: `config.eager_load` is set to true for all environments except for `development` in `config/environments/shared_deployment_config.rb` (and is explicitly set to false in `development.rb`), so it shouldn't - but nonetheless, this is probably not a pattern we want to use on user-facing pages.

PR #4570, opened 2024-06-04
TBE-23 schemas for heroku

tofarr:
## What was done?

We needed to be able to accommodate for the schemas being pulled from S3 on the fly (instead o being in github). In heroku, this means we check whether the schema exists before referencing it, and pull it to the ephemeral file system if it does not exist.

* Introduced a SchemaFileLoader rather than just using the file system directly. This will load the schemas from different locations in different environments.
* Made sure all places use the SchemaFileLoader rather than the file system directly.
* Explicitly excluded gyr-efiler from this functionality.
* Added specs to cover the schema load (The rake task had no specs).

The obvious alternative to this is to place the schemas in git, but we are not permitted to do this.

## How to test?
 * Test on Heroku! https://gyr-review-app-4570-3a542aa5043c.herokuapp.com/
 * You should be able to get to the return status step in FYST: 
image
 * Submissions will get to the "queued" state:
image
* Show XML / Show Direct File MXL works:
image
 

[TBE-23]: https://codeforamerica.atlassian.net/browse/TBE-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/tasks/setup.rake
```diff
@@ -3,80 +3,19 @@ require 'zip'
 namespace :setup do
```
tofarr, line 3: A Lot of this goes away because it was moved over to the SchemaFileLoader

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,107 @@
+
```
tofarr, line 1: A lot of this was copied from setup.rake

spec/lib/schema_file_loader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        stub_const("ENV", {})
+        expect(Rails.application.credentials).to receive(:dig).with(:aws, :access_key_id).and_return "mock-aws-access-key-id"
+        expect(Rails.application.credentials).to receive(:dig).with(:aws, :secret_access_key).and_return "mock-aws-secret-access-key"
+        credentials = SchemaFileLoader.s3_credentials
+        expect(credentials.access_key_id).to eq "mock-aws-access-key-id"
+      end
+    end
+  end
+
+  context "prepare_directories" do
```
jenny-heath, line 43: really minor thing: i think the convention for a block testing a method looks like `describe ".prepare_directories"` (or `describe "#prepare_directories"` if it were an instance method)

spec/lib/schema_file_loader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    ]
+  end
+
+  context "s3_credentials" do
+
+    context "AWS_ACCESS_KEY_ID in ENV" do
+      it "uses the environment variables" do
+        stub_const("ENV", {
+          "AWS_ACCESS_KEY_ID" => "mock-aws-access-key-id",
+          "AWS_SECRET_ACCESS_KEY" => "mock-aws-secret-access-key"
```
jenny-heath, line 25: very small syntax thing: i think we tend to favor `{AWS_ACCESS_KEY_ID: "mock-aws-access-key-id"}` over `{"AWS_ACCESS_KEY_ID" => "mock-aws-access-key-id"}`

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,109 @@
...
+
+class SchemaFileLoader
+
+  BUCKET = "vita-min-irs-e-file-schema-prod"
+  REGION = "us-east-1"
```
jenny-heath, line 6: should we `.freeze` these too?

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      Aws::Credentials.new(
+        Rails.application.credentials.dig(:aws, :access_key_id),
+        Rails.application.credentials.dig(:aws, :secret_access_key),
+        )
+    end
+
+    def prepare_directories(dest_dir)
+      [File.join(dest_dir, 'irs', 'unpacked'), File.join(dest_dir, 'us_states', 'unpacked')].each do |unpack_path|
```
jenny-heath, line 66: i wonder if this could get a flattened list of directory names from `EFILE_SCHEMAS_FILENAMES` so we wouldn't have to update it whenever a new directory was added to the list

spec/lib/schema_file_loader_spec.rb
```diff
@@ -0,0 +1,83 @@
+require "rails_helper"
+
+# TODO: .new is only needed if we are testing instance access instead of table access.
+#   Consider removing .new in specs below unless necessary
```
jenny-heath, line 4: i don't see any `.new`s in the spec?

jenny-heath (approved):
i think i mostly understand this and like it! two things:
1. i'm not that familiar with how to set heroku environment variables (except as a one-off) but from a cursory look around it seems like we set up the ones we want for all review apps in `heroku-pull-request.yml`. should we go ahead and add `AWS_ACCESS_KEY_ID` etc as part of this PR?
2. i think a second reviewer is a great idea since this is a fairly important change

spec/lib/schema_file_loader_spec.rb
tofarr: I found the code in the AWS client actually looks for strings here rather than symbols, so the : syntax did not work.

spec/lib/schema_file_loader_spec.rb
tofarr: Done - for all methods being tested by this spec

app/lib/schema_file_loader.rb
tofarr: Done!

app/lib/schema_file_loader.rb
tofarr: Done!

spec/lib/schema_file_loader_spec.rb
tofarr: Removed auto generated comment.

mrotondo:
Out of curiosity, as I'm reviewing - will this allow us to download the schemas from s3 on development machines as well, replacing the google drive download process when setting up a new dev environment?

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,111 @@
...
+      download_schemas_from_s3(dest_dir)
+      unzip_schemas(dest_dir)
+    end
+
+    def download_schemas_from_s3(dest_dir)
+      s3_client = Aws::S3::Client.new(region: REGION, credentials: s3_credentials)
+      SchemaFileLoader::EFILE_SCHEMAS_FILENAMES.each do |(filename, download_folder)|
+        download_path = File.join(dest_dir, download_folder, filename)
+
+        next if File.exist?(download_path) # If the file already exists, do not re-download.
```
mrotondo, line 40: I know you just copied this over, but this `each-with-conditional-next` seems to duplicate the `filter` call in `get_missing_downloads`. Could we call that method here & enumerate over the result?

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,111 @@
...
+          key: filename,
+          )
+      end
+    end
+
+    def s3_credentials
+      # On Circle CI, get AWS credentials from environment.
+      # In staging, demo, heroku, and prod environment, get credentials from Rails credentials.
+      #
+      # In development, download the file manually from S3. This allows us to avoid storing any AWS credentials in the development secrets.
```
mrotondo, line 54: Alright, seems like this answers the question I asked on the PR conversation - though I don't really see how having AWS creds in our dev credentials file is a higher risk than leaving our browsers signed into AWS, Google, etc.

mrotondo: Do you know why we don't use the credentials file for this on Circle CI?

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      Aws::Credentials.new(
+        Rails.application.credentials.dig(:aws, :access_key_id),
+        Rails.application.credentials.dig(:aws, :secret_access_key),
+        )
+    end
+
+    def prepare_directories(dest_dir)
+      download_folders = EFILE_SCHEMAS_FILENAMES.map { |n| n[1] }.to_set
+      download_folders.each do |download_folder|
```
mrotondo, line 67: No change necessary, but it seems a little strange that we don't re-download zips but _do_ re-extract them.

app/lib/schema_file_loader.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    def unzip_schemas(dest_dir)
+      EFILE_SCHEMAS_FILENAMES.each do |(filename, download_folder)|
+        download_path = File.join(dest_dir, download_folder, filename)
+        next unless File.exist?(download_path)
+        Zip::File.open_buffer(File.open(download_path, "rb")) do |zip_file|
+          # A zip file like AZIndividual2022v1.1.zip will either contain files like AZIndividual2022v1.1/AZIndividual/etc
+          # *or* just AZIndividual. Here we normalize by always trying to unzip in such a way that results in a unique
+          # folder for every unzip
```
mrotondo, line 90: Can you change this to something like "unique folder for every schema version"?

lib/tasks/setup.rake
```diff
@@ -3,80 +3,19 @@ require 'zip'
...
 
-  EFILE_SCHEMAS_FILENAMES = [
-    ["efile1040x_2020v5.1.zip", "irs"],
-    ["efile1040x_2021v5.2.zip", "irs"],
-    ["efile1040x_2022v5.3.zip", "irs"],
-    ["efile1040x_2023v5.0.zip", "irs"],
-    ["NYSIndividual2023V4.0.zip", "us_states"],
-    ["AZIndividual2023v1.0.zip", "us_states"],
-  ].freeze
+  VENDOR = "vendor"
```
mrotondo, line 6: Could you change this to VENDOR_DIR or VENDOR_PATH?

mrotondo (approved):

tofarr:
> Out of curiosity, as I'm reviewing - will this allow us to download the schemas from s3 on development machines as well, replacing the google drive download process when setting up a new dev environment?

Yes. (Assuming the dev has the dev keys in their env.)

lib/tasks/setup.rake
tofarr: Done

app/lib/schema_file_loader.rb
tofarr: Done!

app/lib/schema_file_loader.rb
tofarr: I agree - this was copied from the original setup.rake. I think the idea was to not contact S3 unless required.

app/lib/schema_file_loader.rb
tofarr: I don't know why - I just tried to maintain the status quo - does circleci actually use the circleci.yml.enc ?

app/lib/schema_file_loader.rb
tofarr: Done!

PR #4574, opened 2024-06-05
GYR1-477 Fix for Deployment Failed in Heroku

tofarr:
## What was done?
- This is a fix for the "Deployment Failed" message when deploying to Heroku. The `experiments.rb` is loaded during the build before the database is actually set up. We now catch an additional exception type: `ActiveRecord::ConnectionNotEstablished`

## Before
image

## After
image

## Testing
 - Take a look at the deployed app - there are now no errors.

[GYR1-477]: https://codeforamerica.atlassian.net/browse/GYR1-477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/initializers/experiments.rb
+++ b/config/initializers/experiments.rb
@@ -3,6 +3,9 @@ Rails.application.reloader.to_prepare do
     if ActiveRecord::Base.connection.table_exists?(:experiments)
       ExperimentService.ensure_experiments_exist_in_database
     end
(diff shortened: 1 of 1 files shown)


jnf (approved):
nice catch!

PR #4577, opened 2024-06-07
Fix flaky tests and warnings and add TimeCop testing

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/GYR1-466?atlOrigin=eyJpIjoiOWI0OWM3NzYxYWRjNDk0N2I1ZGZiY2EzZDJhNTM2NTciLCJwIjoiaiJ9
## What was done?
- Added some tests for missing time based methods
- Changed the system time to be 1/17/2025 and ran all the tests and fixed the ones that did
- Fixed some flaky tests that would break when running bin/test
- fix some things that was causing warnings to print when running bin/test


spec/controllers/vita_providers_controller_spec.rb
```diff
@@ -227,7 +227,7 @@
           get :show, params: { id: provider.id.to_s, zip: "94609" }
 
           expect(assigns(:zip)).to eq "94609"
-          expect(assigns(:distance)).to eq 179.95539713817547
+          expect(assigns(:distance)).to be_within(0.000000001).of(179.95539713817547)
```
tofarr, line 230: Nice fix - floating point rounding errors suck!

config/database.yml
```diff
@@ -18,7 +18,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *local_default
-  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] %>
+  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] == '1' ? '' : ENV['TEST_ENV_NUMBER'] %>
```
tofarr, line 21: Can you explain this? I am not sure what is going on here.

tofarr: Em informs me this is a revert to a previous change. It looks like this affects local dev testing only.

tofarr (approved):
🍰

PR #4584, opened 2024-06-08
TBE-44 Restructure the fyst test directories to clearly denote tax year and state

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-44
## PM Acceptance Required?
- No, this can be merged after code review acceptance
## What was done?
- Moved the existing federal return XMLs into spec/fixtures/state_file/fed_return_xmls/{az, ny}
- Renamed them to remove `fed_return_`, `_ny` and `_az`
- Changed the implementation of the `XMLReturnSampleService` to load XMLs from the new directory structure into a hash structure that mirrors it
- **Internal user-facing:** Changed the implementation of `StateFilePagesController` (the list of fake returns to "transfer") so that it only shows returns from the current tax year and the selected state for the current intake
- Alternatives considered
## How to test?
- Along with seeing automated tests pass, I manually verified the change to the fake returns page, and confirmed that they are still loaded correctly and that our stored federal submission IDs are loaded correctly when applicable
- All changes to the fake data transfer page can be viewed and tested on this PR's heroku instance
- Risk Assessment
  - We don't expose these returns externally, and aside from our automated tests, they are only used in the internal-only fake transfer page, so there should be no risk to this change
- Before (showing the page for an AZ intake as an example, but NY would show both NY & AZ returns as well)
image

- After (showing the page for an AZ intake as an example, but NY would show only NY returns)
image


embarnard (approved):
LGTM!

PR #4586, opened 2024-06-10
FYST-53 Now specifying bank account type only if there is a bank account

tofarr:
## What was done?
- When a user specifies that they do not want to use direct deposit, we should not specify a bank account type. (Previously we were specifying "personal" here)
## How to test?
 - Test on staging
 - The XML produced should still have `BNK_ACCT_ACH_IND` if the payment type is direct file, but it should be missing otherwise.
### Before
image

### After
image

[FYST-53]: https://codeforamerica.atlassian.net/browse/FYST-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tofarr:
Co-authored-by: Martha Pidcock <mpidcock@codeforamerica.org>

PR #4588, opened 2024-06-10
Tbe 53 flakey tests

embarnard:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-53
## What was done?
- Everything we found and fixed re: flaky tests is described in the [Flaky Tests investigation doc](https://www.notion.so/cfa/Flaky-Tests-4f3c50c65fac4a2d8115cd7d894f0dbc)
- We also resolved a few warnings being thrown by tests
## How to test?
- CircleCI has been how we've detected flakiness, so the first check should be whether all tests consistently pass on CircleCI. 
- We should also continue to monitor CircleCI to see whether any more/remaining flakiness is reported, and prioritize fixing that as followup
- Risk Assessment
  - We now disable our rate limiter (`Rack::Attack`) in tests, but this shouldn't affect any live traffic environment

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -19,6 +19,7 @@ development:
 test:
   <<: *local_default
   database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] %>
+  pool: 10
 
 production:
   <<: *deploy_default
--- a/config/database.yml
+++ b/config/database.yml
@@ -18,8 +18,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *local_default
-  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] == '1' ? '' : ENV['TEST_ENV_NUMBER'] %>
-  pool: 10
+  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] %>
 
 production:
   <<: *deploy_default
--- a/spec/controllers/ctc/portal/dependents_controller_spec.rb
+++ b/spec/controllers/ctc/portal/dependents_controller_spec.rb
@@ -47,9 +47,9 @@ describe Ctc::Portal::DependentsController do
 
   context "#update" do
     around do |example|
-      Timecop.freeze(DateTime.new(2021, 3, 4, 5, 10))
-      example.run
-      Timecop.return
+      Timecop.freeze(DateTime.new(2021, 3, 4, 5, 10)) do
+        example.run
+      end
     end
 
     it "does not make a system note if nothing significant changed" do
--- a/spec/features/web_intake/returning_filer_spec.rb
+++ b/spec/features/web_intake/returning_filer_spec.rb
@@ -37,7 +37,7 @@ RSpec.feature "Web Intake Returning Filer", :flow_explorer_screenshot do
     click_on I18n.t('general.continue')
 
     # backtaxes
-    check Rails.configuration.gyr_current_tax_year - 2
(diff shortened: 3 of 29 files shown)


spec/tasks/cleanup_orphaned_users_spec.rb
```diff
@@ -22,7 +18,7 @@
     it "it finds users without roles" do
       expect {
         task.invoke
-      }.to output(/These users have no roles: \[#{admin_user.id}, #{team_member_user.id}]/).to_stdout
+      }.to output(/These users have no roles: \[#{admin_user.id}, #{team_member_user.id}\]/).to_stdout
```
mrotondo, line 21: fixed a warning: this regex was missing an escaping \ before it's ], which is invalid syntax but was still able to be parsed

spec/features/web_intake/returning_filer_spec.rb
```diff
@@ -37,7 +37,7 @@
     click_on I18n.t('general.continue')
 
     # backtaxes
-    check Rails.configuration.gyr_current_tax_year - 2
+    check (Rails.configuration.gyr_current_tax_year - 2).to_s
```
mrotondo, line 40: This and the below identical changes are because Capybara was throwing a warning that we shouldn't use an int as an argument to `check`, and that anything other than a string or symbol would throw an error in a future version

config/database.yml
```diff
@@ -18,7 +18,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *local_default
-  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] == '1' ? '' : ENV['TEST_ENV_NUMBER'] %>
+  database: vita-min_test<%= ENV['TEST_ENV_NUMBER'] %>
```
embarnard, line 21: Why did this change?

spec/controllers/ctc/portal/dependents_controller_spec.rb
```diff
@@ -47,9 +47,9 @@
 
   context "#update" do
     around do |example|
-      Timecop.freeze(DateTime.new(2021, 3, 4, 5, 10))
-      example.run
-      Timecop.return
+      Timecop.freeze(DateTime.new(2021, 3, 4, 5, 10)) do
```
embarnard, line 50: am I correct in understanding that this was not causing a failure before because we were returning Timecop but this change is just a stylistic preference? Or was this `Timecop.return` failing to actually exit the freeze block before?

config/database.yml
mrotondo: We can revert this - it just seems unnecessary based on the documentation [here](https://github.com/grosser/parallel_tests?tab=readme-ov-file#add-to-configdatabaseyml). Seems like TEST_ENV_NUMBER is already the empty string for the first process (and I confirmed this in my testing), so we still get a database named vita-min_test when we do `rake parallel:create`. Maybe this isn't how it worked previously? In any case, I think it's safe to get rid of the conditional.

spec/controllers/ctc/portal/dependents_controller_spec.rb
mrotondo: Yes - all Timecop changes except for the test in `rack_attack_spec.rb` were style changes to get us consistent in using the block sytax (`Timecop.freeze do ... end`) instead of needing to remember to call Timecop.return.

PR #4591, opened 2024-06-11
TBE-26 Automated messages now displays different Locales, States & Intakes

tofarr:
## What was done?
- Made it easier to test StateFile Automated Messages
- Can now select state
- Can now pick an existing intake (Defaults to "Cornelius Talmadge")

## How to test?
- Test on heroku: https://gyr-review-app-4591-ab71020335fe.herokuapp.com/en/hub/state-file/automated_messages

### Before
image

### After
image

 * Use "Arizona" and "New York" button to swap between US States
 * You can enter an Id for an intake in your current US State 
image
 * Both English and Spanish are now displayed
image
 * Layout was changed slightly

[TBE-26]: https://codeforamerica.atlassian.net/browse/TBE-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


embarnard:
Screenshot 2024-06-21 at 11 52 30 AM
Why isn't this one interpolating the client id here?

app/controllers/hub/state_file/automated_messages_controller.rb
```diff
@@ -1,26 +1,71 @@
...
-        [klass, message]
+    def get_intake
+      return @intake if @intake.present?
+      state_class = "StateFile#{@us_state.titleize}Intake".constantize
+      if params[:intake_id].present?
+        @intake = state_class.find_by_id(params[:intake_id])
+        return @intake if @intake.present?
+        flash[:alert] = "Unknown Intake"
       end
+      @intake = state_class.new(
```
embarnard, line 25: so if there is not an intake id set then it'll use this, is there no id because its not actually creating it?

app/controllers/hub/state_file/automated_messages_controller.rb
```diff
@@ -1,26 +1,71 @@
...
+      )
+    end
+
+    def message_params
+      intake = get_intake
+      state_code = intake.state_code
+      locale = intake.locale || "en"
+      submitted_key = intake.efile_submissions.count > 1 ? "resubmitted" : "submitted"
+      {
+        primary_first_name: intake.primary_first_name,
```
embarnard, line 37: so every time there is a new message we'll have to add message params here? We have a method called `base_args` in the `app/services/state_file/messaging_service.rb` and as you used here the survey links are in `StateFile::SurveyLinksConcern`. I wonder if we could everything in one place so we won't have to remember to put it in different places every time

embarnard (approved):
Looks good! Left some comments

app/controllers/hub/state_file/automated_messages_controller.rb
```diff
@@ -1,26 +1,71 @@
...
-                                               subject: klass.new.email_subject)
-        message = StateFile::NotificationMailer.user_message(notification_email: email)
-        ActionMailer::Base.preview_interceptors.each do |interceptor|
-          interceptor.previewing_email(message)
-        end
-        [klass, message]
+    def get_intake
+      return @intake if @intake.present?
+      state_class = "StateFile#{@us_state.titleize}Intake".constantize
+      if params[:intake_id].present?
```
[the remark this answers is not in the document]
tofarr: Intake is is getting set optionally by the form in index.html.erb. It allows us to optionally specify an existing intake by it's id for testing. I want to confirm this behavior with Tiffany today.

app/controllers/hub/state_file/automated_messages_controller.rb
tofarr: Yup

app/controllers/hub/state_file/automated_messages_controller.rb
tofarr: Good call - I'm gonna do this and we can re-review

tofarr: So I added additional specs here to test the automated-messages endpoint. It loads all descendants of the `BaseAutomatedMessage`, so if you add one that has additional parameters it should cause the spec to fail until they are added to this hash.

PR #4592, opened 2024-06-12
Fyst 128 state configuration

jenny-heath:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-128
## What was done?
- the goal of this story is to:
  - establish a source of truth for static information about states (name, abbreviation, in future: links etc)
  - consolidate any references to state names as strings ("Arizona") so they all reference this source of truth
- as a side effect, i ended up deleting `StateFileControllerConcern` (a lot of it wasn't being used as far as i could tell) and moving the remaining helper methods to the state file questions controller
- another somewhat-side-effect is that i removed the reference to the state abbreviation and the `STATE_CODE_AND_NAMES` stuff from the intake models and moved it to the service. i say side effect because the story said to just do names.
- another thing i'm thinking about: should this source of truth be a service in the style of the multitenant service or something more config-y? happy for it to be this for the time being and maybe later we move parts of it into the rails config or a yml file or something
- another thing for reviewers to look out for: did i remove anything from the concern's helper methods that are actually used? i would love to believe that the tests would catch all of it but of course it's hard to say 😅
## How to test?
- Everything should work the same so hopefully the tests are the tests??


app/controllers/concerns/state_file/state_file_controller_concern.rb
```diff
@@ -1,52 +1,27 @@
...
-
-    def az?
-      state_param == "az"
-    end
-
-    def service_type
-      case state_param
-      when "az" then :statefile_az
-      when "ny" then :statefile_ny
+    # TODO idea: state_code could prioritize getting the code from the current intake and if there isn't one, default to the params
```
jenny-heath, line 10: i _think_ this idea is out of scope for this story but i wanted to leave this comment because the idea of a controller helper method called something as generic as `state_code` that only looks at the params feels incomplete to me

app/controllers/concerns/state_file/state_file_controller_concern.rb
```diff
@@ -1,52 +1,27 @@
...
+      StateFile::StateInformationService.state_name(state_code)
     end
 
     def current_tax_year
-      tenant_service.current_tax_year
-    end
-
-    def filer_count
-      current_intake&.filer_count
+      MultiTenantService.new(:statefile).current_tax_year
```
jenny-heath, line 24: i tried deleting this initially but a test failed. i would like for this to look different in the future but fixing it felt out of scope

app/services/state_file/state_information_service.rb
```diff
@@ -0,0 +1,43 @@
+module StateFile
+  class StateInformationService
```
jenny-heath, line 2: flexible on this name

app/services/state_file/state_information_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    private
+
+    STATE_INFO = {
+      az: {
+        intake_class: StateFileAzIntake,
+      },
+      ny: {
+        intake_class: StateFileNyIntake,
+      }
+    }
```
jenny-heath, line 41: envisioning that we'll just keep all of the state config-type info in this hash for now and the methods on this service just help extract what is needed

app/controllers/state_file/faq_controller.rb
```diff
@@ -4,11 +4,9 @@ class StateFile::FaqController < ApplicationController
   def index
     @state_code_names = case params[:us_state]
                         when 'us'
```
mrotondo, line 6: Maybe this would make more sense as an index/show pair of methods, and we drop the 'us' param when showing the whole page? I.e. there's a new route with no us_state param that goes to index, and index just returns the state_code_to_name_map, and a route with a us_state_param that goes to show, and there we do the slice?

mrotondo: ^ this is out of scope for this change, just havin' a ponder

app/services/state_file/state_information_service.rb
```diff
@@ -0,0 +1,43 @@
+module StateFile
+  class StateInformationService
+    class << self
+      def active_states # TODO thinking this could also be called `active_state_codes` or just `state_codes`?
```
mrotondo, line 4: I like `active_state_codes`

app/services/state_file/state_information_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      def state_name(state_code)
+        unless active_states.include?(state_code)
+          raise StandardError, state_code
+        end
+
+        States.name_for_key(state_code.upcase)
+      end
+
+      def state_code_to_name_map
```
mrotondo, line 16: I think this would be simpler as `active_states.to_h { |state_code, _| [state_code, state_name(state_code) }` but I haven't tested that

app/services/state_file/state_information_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      def active_states # TODO thinking this could also be called `active_state_codes` or just `state_codes`?
+        STATE_INFO.keys.map(&:to_s)
+      end
+
+      def state_name(state_code)
+        unless active_states.include?(state_code)
+          raise StandardError, state_code
+        end
+
+        States.name_for_key(state_code.upcase)
```
mrotondo, line 13: Feeling less & less convinced that keeping the States class around is a good idea. I think I'd prefer that everyone come to this class for a display name, and that it be stored in the STATE_INFO hash.

app/controllers/concerns/state_file/state_file_controller_concern.rb
```diff
@@ -1,52 +1,27 @@
+# TODO i left these things in here because this concern exists but the thing is the concern is only included once (in the state file questions controller) so should we just move all this over to the controller itself?
```
tofarr, line 1: Please do this! Simpler is better!

app/controllers/state_file/faq_controller.rb
tofarr: Maybe ditch the case parameter here too in favor of an if else?

app/controllers/state_file/questions/landing_page_controller.rb
```diff
@@ -6,7 +6,6 @@ class LandingPageController < QuestionsController
       skip_before_action :redirect_if_in_progress_intakes_ended
 
       def edit
-        @state_name = StateFileBaseIntake::STATE_CODE_AND_NAMES[params[:us_state]]
```
tofarr, line 9: I think this is actually used in landing_page/edit.html.erb ?

app/services/state_file/state_information_service.rb
tofarr: I agree - the problem is that `state` is such an overloaded term I think we need to be specific. (Is a state the current situation within the program [e.g. Statesman], or a geographical region)

app/views/state_file/questions/eligibility_residence/edit.html.erb
```diff
@@ -1,5 +1,4 @@
 <% title = t(".title") %>
-<% state_full_name = params[:us_state] == "ny" ? "New York" : "Arizona" %>
```
tofarr, line 2: Great to see things like this getting cleaned up: 🤮

app/controllers/state_file/questions/landing_page_controller.rb
jenny-heath: oh good catch! forgot to change the view to just call the helper `state_name` instead of the instance variable

app/services/state_file/state_information_service.rb
jenny-heath: well, the States class is unfortunately used in a bunch of places so idk about getting rid of it but i do agree that it would make sense for this service to have its own source of truth about state names

app/services/state_file/state_information_service.rb
jenny-heath: done!

mpidcock (approved):
🎉

PR #4593, opened 2024-06-12
fix flaky notes test

mrotondo:
## What was done?
- A test in `spec/features/hub/clients/notes_spec.rb` was flaky, in that would fail if it was run by itself, or if ever happened to be the case that the user and client it created had the same database ID.
- This was happening often enough that someone noticed it failing and changed it to have the incorrect expectation. It kept failing, but in the opposite scenario: when the client and user had different IDs, and was lost once more in a cloud of CircleCI flakiness.
- The actual problem is that the query the test uses wasn't differentiating properly between documents uploaded by clients and users, probably because the "client" object it receives as an argument is actually a `HubClientPresenter` that is delegating methods to a `Client` object it has a reference to. This seemed to be breaking ActiveRecord's normal handling of `where` clauses that can handle an object and use its ID and type to query a polymorphic column (in this case, `uploaded_by`).
- To fix it, we just had to constrain the `where` clause to only return documents uploaded by a Client.
## How to test?
- The unit test now passes regardless of whether another test has been run that created a User or Client, i.e. regardless of whether the User and Client created by this test have the same database ID. The CircleCI test run should be green. 
- Risk Assessment
  - Zero risk from this change, as long as it is correct per code review.
### Before
  - By itself (User & Client have same ID)
image

  - After another test (User & Client have different IDs)
image

### After
  - By itself (User & Client have same ID)
image

  - After another test (User & Client have different IDs)
image

### DIFF
--- a/app/lib/synthetic_note.rb
+++ b/app/lib/synthetic_note.rb
@@ -26,7 +26,12 @@ class SyntheticNote
   end
 
   def self.from_client_documents(client)
-    grouped = client.documents.order(created_at: :asc).where(uploaded_by: client).group_by { |doc| doc.created_at.beginning_of_day }
+    # The over-specific where clause is necessary because the actual "client" object we receive here can be a
+    # HubClientPresenter, which breaks the usual ActiveRecord polymorphic lookup, which would allow for
+    # where(uploaded_by: client/user) to automatically enforce that the polymorphic uploaded_by_type must have the same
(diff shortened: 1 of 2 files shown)


mrotondo:
Reviewed by @jenny-heath over Tuple

tofarr (commented):
Great job finding why it was sometimes 5!

PR #4594, opened 2024-06-13
Add fed_refund_amt and zip_code to state_file_analytics

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-134
## What was done?
- Used the federal data to populate zip code and federal refund amount in our state file analytics table to be used in metabase
## How to test?
- Updated existing unit tests
- Will check with Max when it is in demo
- Risk Assessment
  - Does not affect business logic
  - Note: I added a getter for `fed_refund_amt` but not a setter since it is only read by the analytics data

### DIFF
--- a/db/migrate/20240612220721_add_direct_file_data_fields_to_state_file_analytics.rb
+++ b/db/migrate/20240612220721_add_direct_file_data_fields_to_state_file_analytics.rb
@@ -15,13 +15,17 @@ class AddDirectFileDataFieldsToStateFileAnalytics < ActiveRecord::Migration[7.1]
       #    use the logic from the model at the time of the db update.
       #  * The data in raw_direct_file_data is encrypted, so we need rails to decrypt it
       #
-      StateFileAnalytics.find_each do |analytics|
-        next unless analytics.record.raw_direct_file_data.present?
-        direct_file_data = analytics.record.direct_file_data
-        analytics.update(
-          fed_refund_amt: direct_file_data.fed_refund_amt,
-          zip_code: direct_file_data.mailing_zip
-        )
+      StateFileAnalytics.find_in_batches(batch_size: 100) do |batch|
+        ActiveRecord::Base.transaction do
+          batch.each do |analytics|
+            next unless analytics.record.raw_direct_file_data.present?
+            direct_file_data = analytics.record.direct_file_data
+            analytics.update(
+              fed_refund_amt: direct_file_data.fed_refund_amt,
+              zip_code: direct_file_data.mailing_zip
+            )
+          end
+        end
       end
     rescue Exception => e
       puts e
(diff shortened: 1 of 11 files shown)


spec/models/state_file_az_intake_spec.rb
```diff
@@ -77,7 +77,6 @@
 #
 # Indexes
 #
-#  index_state_file_az_intakes_on_email_address        (email_address)
```
tofarr, line 80: Weird - I think this may be related to a weird annotation glitch Mike was working on the other week.

tofarr (changes requested):
(Aside from the annotations) - If I understand this PR - the intent is to add the federal refund amount and the zip code to the StateFileAnalytics model and to synchronize these when the IntakeModel is created or updated.

Concerns I have are:
* I don't think we should remove the indexes on the email_address fields - I think this is related to a glitch in the local db which Mike fixed the other week.
* Do we need to sync this data for the existing intakes?

db/migrate/20240612220721_add_direct_file_data_fields_to_state_file_analytics.rb
```diff
@@ -0,0 +1,48 @@
...
+      #
+      #  * In this case we are pulling data from an XML field with custom calculations - we would need
+      #    to reproduce this logic
+      #  * This migration is only used to back populate data to analytics - items created after this will
+      #    use the logic from the model at the time of the db update.
+      #  * The data in raw_direct_file_data is encrypted, so we need rails to decrypt it
+      #
+      StateFileAnalytics.find_in_batches(batch_size: 100) do |batch|
+        ActiveRecord::Base.transaction do
+          batch.each do |analytics|
```
mpidcock, line 20: suggest rename to state_file_analytics for clarity

db/migrate/20240612220721_add_direct_file_data_fields_to_state_file_analytics.rb
```diff
@@ -0,0 +1,48 @@
...
+  #  ActiveRecord::Base.connection.execute(
+  #    <<~SQL
+  #      UPDATE state_file_analytics SET
+  #        fed_refund_amt=xpath('./IRS1040/RefundAmt/text()', state_file_ny_intakes.raw_direct_file_data::xml)::text::integer,
+  #        zip_code=xpath('./ReturnHeader/Filer/USAddress/ZIPCd/text()', state_file_ny_intakes.raw_direct_file_data::xml)
+  #      FROM state_file_ny_intakes
+  #      WHERE state_file_analytics.record_type='StateFileNyIntake' AND state_file_analytics.record_id = state_file_ny_intakes.id
+  #  SQL
+  #  )
+  # end
```
mpidcock, line 47: Could you put this in the PR description instead of commented out?

db/schema.rb
```diff
@@ -10,9 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_05_31_164754) do
-  create_schema "analytics"
-
```
mpidcock, line 15: how did this happen?

mpidcock (commented):
a few small comments

db/schema.rb
tofarr: I think it was accidentally added by me (Since I had run the analytics.rake locally.)

PR #4595, opened 2024-06-13
GYR1-485 Cookie Overflow Fix

tofarr:
### The Problem...
Every week or so, we get an error message in sentry that somebody has had a cookie overflow in the hub. It looks like a reason for this is that we store filter parameters in a session cookie, and the total size of all cookies cannot exceed 4096 bytes. Looking at the cookie, I found that we store an array of vita partners used for filtering in a format like this:
```
[
  {"id": 1234, "name": "Some long partner name", "parent_name": "Some long parent name", "value": 1234}
  {"id": 1235, "name": "Some other long partner name", "parent_name": "Some other long parent name", "value": 1235}
  ...
]
```

The overflowing instance I saw had 10 of these embedded in the cookie! Further, it seems that the client side library we use ([Tagify](https://github.com/yairEO/tagify)) requires that the data be available in this format.

image
image
image

#### Use a different client side tagging library
We could swap out tagify for some other more flexible client side library. Although this could reduce the code complexity, there would be overhead in researching and replacing something we have used extensively for so long. Also, by nature, most solutions are quite javascript intensive, which is something we veer away from philosophically.

#### Use local storage instead of session cookies for storing filters
This will allow us to store more data, but we would need to have an additional page reload when entering to redirect users to have the correct parameters. I worry that this additional redirect could be jarring.

### What Was Done...

*We massively reduced the amount of data submitted to the server and stored in the session cookie.*

* Removed the 'vita_partners' name from the DOM input used by tagify.
* Create a hidden element with `vita_partners` as its name.
* Add an onChange listener to the tagify input to sync and translate it's value to the new hidden element. The new hidden element will submit values in the format `[1234, 1235]`
* On the server side, accommodate filters in both the original format and new format using `instance_of?(Hash)`. This should prevent errors from users with the page open while we upgrade the server.
* Translate values applied to the old input to those acceptable to "Tagify"

## Concerns about this approach

* It is more complex than what was there before.
* In theory I think you could still overflow your cookies (With a long search string for example)
* The visitor_id cookie still seems way longer than it needs to be to me (Like, a 256 bit number encoded as base should be 44 characters - this is waaay longer than that - I should look into why.)

[GYR1-485]: https://codeforamerica.atlassian.net/browse/GYR1-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/concerns/client_sortable.rb
```diff
@@ -24,4 +28,22 @@ def cookie_filters
 
     cookies[filter_cookie_name] ? HashWithIndifferentAccess.new(JSON.parse(cookies[filter_cookie_name])) : {}
   end
+
+  def vita_partners_for_tagify
```
mrotondo, line 32: this looks right to me, but it feels ripe for some unit testing. can you add some to `client_sortable_spec.rb`?

app/controllers/concerns/client_sortable.rb
```diff
@@ -24,4 +28,22 @@ def cookie_filters
...
+    # This returns the vita partners in a format that the tagify js library can understand
+    vita_partners = @client_sorter.active_filters[:vita_partners]
+    return if vita_partners.blank?
+    vita_partners = JSON.parse(vita_partners)
+    result = vita_partners.map do |id|
+      id = id[:id] if id.instance_of?(Hash)
```
mrotondo, line 38: Will `id` ever be something other than a hash, and will the final `to_json` call fail if there are any nils in the `result` list (as i believe there could be currently)? I.e. should we call `compact` on result before we call `to_json`? Or handle nils some other way? 

I probably wouldn't have even thought to ask if you hadn't handled the possible `method_missing`, fwiw :)

app/javascript/lib/tagging/index.js
```diff
@@ -61,6 +61,14 @@ export function initMultiSelectVitaPartner() {
             },
         }
     });
+
+    input.addEventListener("change", (event) => {
+        let { value } = event.target
+        if (value) {
+            value = JSON.stringify(JSON.parse(event.target.value).map(v => v.id))
+        }
+        document.querySelector("#vita_partners").value = value
```
mrotondo, line 70: Sorry if I'm not understanding this correctly, but will this overwrite the current value of the #vita_partners hidden field, rather than append to it?

app/lib/client_sorter.rb
```diff
@@ -96,7 +96,11 @@ def filtered_clients
     end
 
     if @filters[:vita_partners].present?
-      ids = JSON.parse(@filters[:vita_partners]).map { |vita_partner| vita_partner["id"] }
+      # For backwards compatibility for users in the middle of a search while this is deployed,
```
mrotondo, line 99: do you want to make another ticket, or a subtask of this one, to remove the hash-handling code a few days after this is pushed?

app/javascript/lib/tagging/index.js
mrotondo: Nevermind - it looks like because this is a multiselect, the value includes all currently selected partners. Is that right?

tofarr: Yes - the value in the input is overwritten and includes all currently selected partners.

app/controllers/concerns/client_sortable.rb
tofarr: Moving forward, id will always be an integer. The only case where it will be a hash is for folks that have had a cookie defined with the old code. In this case, the array of vita_partners is in the format `{"id": 1234, "name": "Some long partner name", "parent_name": "Some long parent name", "value": 1234}`.

After a while, this check will be able to go away. I noted this in the PR description, but I think I'll add a comment to the code too.

app/lib/client_sorter.rb
tofarr: I think it's okay to leave it as a comment in the code for now with the capability of handling both. We can remove it at our leisure.

mrotondo (approved):

PR #4598, opened 2024-06-17
Fyst 133 state code from intake first

jenny-heath:
[Link to pivotal/JIRA issue](https://codeforamerica.atlassian.net/jira/software/c/projects/GYR1/boards/11?selectedIssue=FYST-133)
## Is PM acceptance required?
- No
## What was done?
- added method to state file questions controller `current_state_code` which first tries to get US state from `current_intake` and, if not there, checks params
- replace all uses of `params[:us_state]` that A) should be checking the intake first and B) have access to the helper method
- replace some lines where display name of state is derived from `States` or whatever and instead use `state_name` helper from questions controller
- rename some local variables so they don't show up in search for `state_code`
## How to test?
- One added test for the helper method
- The goal is for it to work the same so if all the tests pass we should (should) be good

### DIFF
--- a/app/controllers/state_file/questions/questions_controller.rb
+++ b/app/controllers/state_file/questions/questions_controller.rb
@@ -20,7 +20,7 @@ module StateFile
         if current_intake
           current_intake.state_code
         else
-          state_code_from_params
+          params[:us_state]
         end
       end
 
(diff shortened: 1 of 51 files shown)


app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -35,19 +35,29 @@ def current_tax_year
       private
 
       def current_intake
-        state_code = question_navigator.intake_class.state_code
-        send("current_state_file_#{state_code}_intake")
+        StateFile::StateInformationService.active_state_codes
+                                          .lazy
+                                          .map{ |c| send("current_state_file_#{c}_intake".to_sym) }
+                                          .find(&:itself)
```
jenny-heath, line 41: this is fairly significant. would like to discuss

mpidcock (approved):

PR #4601, opened 2024-06-20
Fyst 125 extract state specific static data to state config

jenny-heath:
[Link to pivotal/JIRA issue](https://codeforamerica.atlassian.net/browse/FYST-125?atlOrigin=eyJpIjoiY2QxNmNkM2QzY2U4NDM0Y2I2MDRlNTgzZmRiZmYyN2QiLCJwIjoiaiJ9)
## Is PM acceptance required?
## What was done?
- moves some pieces of static data into `StateInformationService`:
  - refund_url
  - tax_payment_url
  - download_form_name
  - mail_voucher_address
  - pay_mail_online_link
  - vita_link
  - voucher_path
  - survey_link
- refactors `StateInformationService` so it creates getter methods for these fields dynamically
- replaces some but not all instances where intake class (e.g. `StateFileAzIntake`) is referenced directly (gets it from `StateInformationService` instead) 
- there are more things to do for this story / in this category but i didn't want to bloat this PR so will do it in follow-up branches
## How to test?
- Again, this is meant to maintain and not add functionality so we should be able to just rely on the existing tests passing, but since test coverage for state file is generally spotty I did end up writing some more tests


app/controllers/state_file/notifications_settings_controller.rb
```diff
@@ -34,7 +34,7 @@ def subscribe_email
     def matching_intakes(email_address)
       return if email_address.blank?
 
-      StateFileAzIntake.where(email_address: email_address) + StateFileNyIntake.where(email_address: email_address)
+      StateFile::StateInformationService.state_intake_classes.map { |klass| klass.where(email_address: email_address) }.inject([], :+)
```
jenny-heath, line 37: i guess i could also do a reduce here, open to suggestions

app/jobs/send_survey_notification_job.rb
jenny-heath, line 1: moved this to state_file namespace

app/jobs/state_file/send_survey_notification_job.rb
jenny-heath, line 1: looks entirely new because i moved it to the state_file namespace but the only new part is line 8, you can look at the deleted file if you want to double check

app/services/state_file/state_information_service.rb
```diff
@@ -1,16 +1,37 @@
...
-        end
-
-        STATES_INFO[state_code.to_sym][:name]
+      def state_intake_classes
+        STATES_INFO.map { |_, attrs| attrs[:intake_class] }
+      end
+      
+      def state_intake_class_names
+        state_intake_classes.map(&:to_s).freeze
       end
```
jenny-heath, line 35: was on the fence about this because it's so similar to `state_intake_classes` but i didn't like having to `.map(&:to_s).freeze` in all of the places where it needs them as strings (there are kind of a lot). if you have other ideas for this tho lmk

app/services/state_file/state_information_service.rb
```diff
@@ -1,16 +1,37 @@
...
+        :survey_link,
+        :tax_payment_url,
+        :vita_link,
+        :voucher_path,
+      ].each do |attribute|
+        define_method(attribute) do |state_code|
+          raise StandardError, "No state code '#{state_code}'" if !active_state_codes.include?(state_code)
+
+          value = STATES_INFO[state_code.to_sym][attribute]
+          raise StandardError, "State '#{state_code}' does not have '#{attribute}'" if !value
```
jenny-heath, line 20: this is me kind of breaking my own rule, the rule being "don't add code that isn't used just because it seems like a nice idea". there aren't currently any attributes in this list that are missing from a state so this line should never be called. now that i've said this out loud maybe i will delete it...

app/services/state_file/state_information_service.rb
```diff
@@ -30,11 +51,32 @@ def state_code_from_intake_class(klass)
...
+        refund_url: "https://aztaxes.gov/home/checkrefund",
+        tax_payment_url: "AZTaxes.gov",
+        download_form_name: "Form AZ-140V",
+        mail_voucher_address: "Arizona Department of Revenue<br/>"\
+          "PO Box 29085<br/>"\
+          "Phoenix, AZ 85038-9085".html_safe,
+        pay_mail_online_link: "https://azdor.gov/making-payments-late-payments-and-filing-extensions",
+        vita_link: "https://airtable.com/appnKuyQXMMCPSvVw/pag0hcyC6juDxamHo/form",
+        voucher_path: "/pdfs/AZ-140V.pdf",
+        survey_link: "https://codeforamerica.co1.qualtrics.com/jfe/form/SV_7UTycCvS3UEokey",
```
jenny-heath, line 62: should i `.freeze` all these strings...?

spec/controllers/state_file/questions/return_status_controller_spec.rb
```diff
@@ -2,6 +2,24 @@
...
+        expect(assigns(:refund_url)).to eq "https://aztaxes.gov/home/checkrefund"
+        expect(assigns(:tax_payment_url)).to eq "AZTaxes.gov"
+        expect(assigns(:download_form_name)).to eq "Form AZ-140V"
+        expect(assigns(:mail_voucher_address)).to eq "Arizona Department of Revenue<br/>"\
+          "PO Box 29085<br/>"\
+          "Phoenix, AZ 85038-9085"
+        expect(assigns(:voucher_path)).to eq "/pdfs/AZ-140V.pdf"
+        expect(assigns(:survey_link)).to eq "https://codeforamerica.co1.qualtrics.com/jfe/form/SV_7UTycCvS3UEokey"
+      end
+    end
```
jenny-heath, line 21: 2 things:
1. i wonder if this test is overkill? actually what might be better is to have a feature spec that looks for these attributes in the view but adding that as part of this PR seemed a little extra
2. i don't love having all these strings in here but it would be annoying to mock the StateInformationService for every single one...

spec/jobs/state_file/send_survey_notification_job_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe StateFile::SendSurveyNotificationJob, type: :job do
```
jenny-heath, line 3: it's a simple job but still seemed worth having a spec

spec/services/state_file/messaging_service_spec.rb
```diff
@@ -28,6 +28,24 @@
         expect(DatadogApi).to have_received(:increment).with("mailgun.state_file_notification_emails.not_sent_because_unsubscribed")
       end
     end
+
+    context "email verification" do
```
jenny-heath, line 32: none of the email verification checks in this service were tested

app/jobs/state_file/send_survey_notification_job.rb
jenny-heath, line 1: looks entirely new because i moved it to the state_file namespace but the only new part is line 8, you can look at the deleted file if you want to double check

app/controllers/state_file/notifications_settings_controller.rb
tofarr: I think this is actually more concise and clearer than a reduce would be:

`.inject([]) { |result, klass| result.append(klass.where(email_address: email_address)) }`

spec/controllers/state_file/questions/return_status_controller_spec.rb
tofarr: I agree with extracting attributes from the view, but I do actually like that the strings are explicitly stated in the text and would not mock that - having them here is simple and explicit - exactly what we want from unit tests.

spec/lib/navigation/state_file_ny_question_navigation_spec.rb
```diff
@@ -1,7 +1,6 @@
 require "rails_helper"
 
 RSpec.describe Navigation::StateFileNyQuestionNavigation do
-
```
tofarr, line 4: Are you using an automated formatter? I noticed a lot of lines like this being removed. If so, which one? (Maybe we should incorporate it in our process)

app/services/state_file/state_information_service.rb
tofarr: That feels verbose and like something that will come up repeatedly. I wonder if we should implement some kind of `deep_freeze` method for cases like this? (I'm surprised ruby doesn't have one built in).

tofarr (approved):

app/services/state_file/state_information_service.rb
jenny-heath: good point about it being repetitive. we actually do have a gem we're using in one place called `ice_nine` that has a `deep_freeze`

spec/lib/navigation/state_file_ny_question_navigation_spec.rb
jenny-heath: i really should get an automated formatter but no i'm doing these manually when i notice them 🙈

mpidcock (approved):

PR #4603, opened 2024-06-24
FYST-129 First batch of refactors

tofarr:
## Is PM acceptance required?
## What was done?

I'm gonna try keeping refactors for this ticket to 10 files or less per PR - so there will be more!

1. I think we should avoid using constantize where possible, but use the StateFile::StateInformationService instead.
   * Constantize can in theory be used to perform hacks by loading something weird and unexpected. These are difficult to pull off, but it is an extra thing you have to think about.
   * It seems to me that this is exactly the problem that the StateInformationService is designed to solve.
2. I added Navigation to the StateInformationService. It seems to fit here.
3. I removed the `STATE_INTAKE_CLASS_NAMES` variable - this should be in the StateInformationService

[FYST-129]: https://codeforamerica.atlassian.net/browse/FYST-129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/ability.rb
```diff
@@ -14,10 +14,13 @@ def initialize(user)
...
       unless user.state_file_admin?
-        cannot :manage, StateFileAzIntake
-        cannot :manage, StateFileNyIntake
+        StateFile::StateInformationService.intake_classes.each do |intake_class|
+          cannot :manage, intake_class
+        end
+        # Enumerate classes here...
         cannot :manage, StateFile1099G
         cannot :manage, StateFileDependent
+        cannot :manage, StateFileW2
```
tofarr, line 23: I added this because I noticed that we had missed it.

app/services/state_file/state_information_service.rb
```diff
@@ -27,11 +27,19 @@ def active_state_codes
...
       end
 
+      def state_name(state_code)
+        unless active_state_codes.include?(state_code)
+          raise StandardError, state_code
+        end
+
+        STATES_INFO[state_code][:state_name]
+      end
+
```
jenny-heath, line 42: so this one is already defined above on lines 4-20! anything that's just a getter for one of the attributes should go in that list

app/services/state_file/state_information_service.rb
```diff
@@ -75,7 +105,8 @@ def state_code_from_intake_class(klass)
         tax_refund_url: "https://www.tax.ny.gov/pit/file/refund.htm",
         vita_link: "https://airtable.com/appQS3abRZGjT8wII/pagtpLaX0wokBqnuA/form",
         voucher_path: "/pdfs/it201v_1223.pdf",
+        navigation: Navigation::StateFileNyQuestionNavigation,
       }
-    })
+    }).with_indifferent_access
```
jenny-heath, line 110: nice!

app/services/state_file/state_information_service.rb
```diff
@@ -42,6 +50,27 @@ def state_code_from_intake_class(klass)
...
+
+      def intake_class_from_state_code(state_code)
+        STATES_INFO[state_code][:intake_class]
+      end
+
+      def navigation_from_state_code(state_code)
+        STATES_INFO[state_code][:navigation]
+      end
+
+      def intake_classes
```
jenny-heath, line 62: this is redundant with `state_intake_classes` defined on line 27. if you like this name better that's fine but we should just pick one

app/services/state_file/state_information_service.rb
```diff
@@ -42,6 +50,27 @@ def state_code_from_intake_class(klass)
         end
         state_code.to_s
       end
+
+      def intake_class_from_state_code(state_code)
+        STATES_INFO[state_code][:intake_class]
+      end
```
jenny-heath, line 56: this is also defined above as one of the simple getter methods. so accessing it would look like `StateInformationService.intake_class(state_code)`

app/controllers/flows_controller.rb
```diff
@@ -111,20 +111,17 @@ def self.for(type, controller)
...
-          form: SampleStateFileIntakeGenerator.new('ny').form
-        )
+      else
+        if type.to_s.starts_with?('state_file_')
+          state_code = type.to_s.split('_').last
+          intake_class = StateFile::StateInformationService.intake_class_from_state_code(state_code)
+          FlowParams.new(
+            controller: controller,
+            reference_object: controller.current_intake&.is_a?(intake_class) ? controller.current_intake : nil,
+            controller_list: StateFile::StateInformationService.navigation_from_state_code(state_code)::FLOW,
```
mpidcock, line 121: Since the only difference between each state is literally the state code, I ask that we not to store this in the `StateInformationService`, and instead derive it on the spot. I don't think it's a good idea to add too much to the service if it can otherwise be derived.

```suggestion
            controller_list: "Navigation::StateFile#{state_code.capitalize}QuestionNavigation".constantize::FLOW,
```

I also like the added benefit for enforcing a naming convention for `StateFileXxQuestionNavigation` classes.

app/services/state_file/state_information_service.rb
```diff
@@ -42,6 +50,27 @@ def state_code_from_intake_class(klass)
...
+
+      def navigation_from_state_code(state_code)
+        STATES_INFO[state_code][:navigation]
+      end
```
jenny-heath, line 60: this can be added to the dynamically defined methods too i think, just add `:navigation` to the list

app/services/state_file/state_information_service.rb
```diff
@@ -42,6 +50,27 @@ def state_code_from_intake_class(klass)
...
+
+      def intake_classes
+        STATES_INFO.map { |_, state_info| state_info[:intake_class] }
+      end
+
+      def intake_class_names
+        intake_classes.map(&:name)
+      end
```
jenny-heath, line 68: same, also already defined with a different name

mpidcock (changes requested):

app/lib/ability.rb
embarnard: ooo good catch

app/services/state_file/state_information_service.rb
```diff
@@ -42,6 +50,27 @@ def state_code_from_intake_class(klass)
...
+
+      def submission_builder_from_intake_class(klass)
+        state_info = STATES_INFO.values.find { |s| s[:intake_class] == klass }
+        state_info[:submission_builder] if state_info.present?
+      end
```
jenny-heath, line 73: i wonder if we should define a method every time we want to get something via not-the-state-code. this for example could be obtained this way:
```
state_code = StateInformationService.state_code_from_intake_class(intake_class)
StateInformationService.submission_builder(state_code)
```
which is more verbose but doesn't result in a custom method. we could dynamically define a way to get any piece of information for a state given any other piece of information, but probably that's overkill...

also, i just realized i don't see this method used anywhere nor do i see a `submission_builder` defined in the STATES_INFO. is part of this PR missing?

app/controllers/flows_controller.rb
jenny-heath: fine by me, and does seem good to enforce a naming convention

jenny-heath: per the slack thread, i forgot we had talked about constantize earlier. is there a way to make sure we don’t end up in a situation where injected parameters are a concern? if state code is often coming in via the url that seems like enough reason to just avoid doing it but i def don't feel like an expert here

app/services/state_file/state_information_service.rb
tofarr: Good catch - this was the result of a merge.

app/services/state_file/state_information_service.rb
tofarr: After discussing with Martha, we are moving navigation out of this class

app/controllers/flows_controller.rb
tofarr: I added a static method that checks that the code is in the whitelist. Using this everywhere is preferable to peppering the code with constantize methods

app/controllers/state_file/questions/canceled_data_transfer_controller.rb
```diff
@@ -36,11 +36,8 @@ def redirect_if_no_us_state_in_params
         return if params[:us_state]
 
         if session[:state_file_intake].present?
-          if current_intake.class == StateFileAzIntake
-            redirect_link = az_questions_canceled_data_transfer_path(us_state: "az")
-          elsif current_intake.class == StateFileNyIntake
-            redirect_link = ny_questions_canceled_data_transfer_path(us_state: "ny")
-          end
+          state_code = StateFile::StateInformationService.state_code_from_intake_class(current_intake)
+          redirect_link = send("#{state_code}_questions_canceled_data_transfer_path".to_sym, us_state: state_code)
```
mpidcock, line 40: I haven't tested this, but something like this should be possible and slightly more readable?

```suggestion
          redirect_link = StateFile::Questions::CanceledDataTransferController.to_path_helper(action: :cancel_data_transfer, us_state: state_code
```

app/lib/navigation/state_file_base_question_navigation.rb
```diff
@@ -1,11 +1,14 @@
...
-  module StateFileBaseQuestionNavigationMixin
+  class StateFileBaseQuestionNavigation
+    class << self
 
-    def self.included(base)
-      base.extend(ClassMethods)
-    end
+      def navigation_class_from_state_code(state_code)
+        unless StateFile::StateInformationService::STATES_INFO.key?(state_code)
+          raise StandardError state_code
```
mpidcock, line 7: This might not be in scope for this story, but it bothers me a bit that the message here is just the state_code, and it occurs to me that it would make more sense to create an `InvalidStateCodeError` class that can be used where needed with a clearer error.

mpidcock (commented):
Nice, I really like a lot of the changes here. There's a little more here though that I'd like to be addressed

app/controllers/state_file/questions/canceled_data_transfer_controller.rb
```diff
@@ -36,11 +36,8 @@ def redirect_if_no_us_state_in_params
         return if params[:us_state]
 
         if session[:state_file_intake].present?
-          if current_intake.class == StateFileAzIntake
-            redirect_link = az_questions_canceled_data_transfer_path(us_state: "az")
-          elsif current_intake.class == StateFileNyIntake
-            redirect_link = ny_questions_canceled_data_transfer_path(us_state: "ny")
-          end
+          state_code = StateFile::StateInformationService.state_code_from_intake_class(current_intake)
```
mpidcock, line 39: `StateFileBaseIntake` does still implement `self.state_code`, so I think it would be better to use `current_intake.state_code` and inline it in the line below

mpidcock: Perhaps `state_code_from_intake_class` could be made private to enforce using `current_intake.state_code` instead

app/controllers/state_file/questions/canceled_data_transfer_controller.rb
tofarr: I started implementing your change, but realized this code is actually:
1. Not covered by any test
2. Unreachable and dead (The code tests what to do if there is no us_state param - but all routes require one:
`(/:locale)/:us_state/questions/canceled-data-transfer(.:format)`)

So I removed it.

mpidcock (approved):

PR #4607, opened 2024-06-25
FYST-129 Part 3: controller names

tofarr:
## Is PM acceptance required?
## What was done?
- We are specifically looking to reduce / remove code in controllers that looks like "if az do ... elsif ny do ..."
## How to test?
  - Test data used
- Risk Assessment
- Before
- After


app/services/state_file/state_information_service.rb
jenny-heath, line 1: could you add a test for this new method to the service spec?

spec/spec_helper.rb
```diff
@@ -56,7 +56,11 @@
   config.filter_run_when_matching :focus
 
   config.before(:each) do
-    Rack::Attack.enabled = false
+    begin
+      Rack::Attack.enabled = false
+    rescue
+      # RackAttack not loaded - probably running a single test
```
jenny-heath, line 62: did you mean to leave this in here? if so is it a normal pattern to have a rescue block with just a comment?

jenny-heath (commented):
looks pretty good but i left a couple of semi-blocking comments

a note for next time: when there's shuffling around of erb code, could you include screenshots of the pages? makes it easier to see that nothing changed

spec/spec_helper.rb
tofarr: When running `rspec spec/i18n_spec.rb` I was getting:
```
Failure/Error: Rack::Attack.enabled = false
     
     NameError:
       uninitialized constant Rack
       ...
```

Although swallowing exceptions is frowned upon, we don't need to do anything if it RackAttack not been required - it just means we are running a single spec. I have updated this to print a warning message just in case.

app/services/state_file/state_information_service.rb
tofarr: Done!

jenny-heath (approved):

PR #4609, opened 2024-06-25
Fix bug: on taxes owed page, state_name should be current_state_name

jenny-heath:
[Sentry](https://codeforamerica.sentry.io/issues/5536292629/?project=1880321&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=90d&stream_index=0)

## Is PM acceptance required?
- No - seems fine to do an eng review
## What was done?
- when the state_name helper was renamed to current_state_name we missed a spot and there's no test that caught it
## How to test?
- fill out a state file intake and make sure you get the taxes owed page. see that it doesn't crash.

### DIFF
--- a/app/views/state_file/questions/taxes_owed/edit.html.erb
+++ b/app/views/state_file/questions/taxes_owed/edit.html.erb
@@ -1,7 +1,7 @@
 <% content_for :page_title, t(".page_title") %>
 <% content_for :card do %>
   <h1 class="h2">
-    <%= t(".title_html", owed_amount: taxes_owed, state_name: state_name) %>
+    <%= t(".title_html", owed_amount: taxes_owed, state_name: current_state_name) %>
   </h1>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>


mrotondo:
Looks good - do you have a sense of how hard it would be to create a test that exercises this view, maybe by parameterizing an existing journey test so that it sees both the owes & refund pages?

jenny-heath:
@mrotondo how hard? not hard but a little tedious and maybe a little slow for our test suite? martha and i had the idea of changing the complete_intake_spec to have the ny one owe taxes but then realized it will be deleted so idk i wonder if part of your personas epic might touch this page? seems good to have some personas who owe taxes

mrotondo:
> i wonder if part of your personas epic might touch this page? seems good to have some personas who owe taxes

some of the personas do owe - that's how i found the bug - but the way i'm currently generating the output in the tests is by creating the intakes with factorybot, rather than by running through the flow with capybara, so they wouldn't load this view. we can chat about whether that's the right approach or whether capybara would make more sense tomorrow at computer gardening, or on the TBE-45 PR

PR #4611, opened 2024-06-25
FYST-50 : Prevent accepted resubmit

tofarr:
## Is PM acceptance required?
- [x] Yes - don't merge until JIRA issue is accepted!
## What was done?
- I feature flagged this functionality change so we can quickly disable it if required.
- There is no new UI state here - we use the following view:
image
- We no longer simply check that the current intake is submitted and accepted - we also check intakes with the same SSN in an accepted state.
## How to test?
- Create a new AZ Intake and make sure the submission is created as expected
## Risk Assessment
  - This will change the flow subtly - the feature flag allows us to rapidly disable this in the unlikely event that a user needs to create a submission and they already have an accepted one.
  - This may make testing more annoying as repeated tests with the same ssn will fail - this was my primary reason for feature flagging this - we can turn it off on staging / demo.

[FYST-50]: https://codeforamerica.atlassian.net/browse/FYST-50?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrotondo (approved):
paired & reviewed w/ tim

PR #4614, opened 2024-06-27
GYR1-503 [12] Hub quick filters always navigate to All Clients page

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-503
## Is PM acceptance required?
- [x] Yes - don't merge until JIRA issue is accepted!
## What was done?
- Change the quick-filter buttons (`Approaching SLA`, `Breached SLA`) on the All Clients and Assigned Clients pages link to the current page instead of always going to the All Clients page.
- Added tests that verify that these links are correct for both pages.
## How to test?
- I added unit tests that confirm the link contents are correct. I verified that the tests fail before the change, and pass after it.
- Manual testing (on **this PR's heroku deployment**) should be done to confirm that:
  - filters can be applied correctly for both pages - filtering a page keeps you on that page
  - filters can be cleared correctly for both pages
  - if not cleared, filters "stick" for both pages: if you navigate away and back, the filter is still applied even if the URL parameters are not present (this is the same as the old behavior, as far as I can tell)
- Risk Assessment
  - I can't think of any risks with this change, but maybe you can!

### DIFF
--- a/app/lib/client_sorter.rb
+++ b/app/lib/client_sorter.rb
@@ -113,8 +113,8 @@ class ClientSorter
     overlapping_keys.any?
   end
 
-  def filtering_only_by?(filter_values)
-    @filters.select { |_k, v| v.present? } == filter_values.transform_values { |v| v.to_s }
+  def filtering_only_by?(filter_values, ignore: [])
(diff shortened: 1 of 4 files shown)


tofarr (approved):
🍰 - Good that you added specs for both Assigned Clients and All Clients

PR #4615, opened 2024-06-27
TEMP/WIP New state experiment

jenny-heath:


app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -12,7 +12,7 @@ def card_postscript
   end
 
   def require_state_file_intake_login
-    if current_state_file_az_intake.blank? && current_state_file_ny_intake.blank?
+    if current_state_file_az_intake.blank? && current_state_file_ny_intake.blank? && current_state_file_wa_intake.blank?
```
mpidcock, line 15: Just noting that this isn't required due to refactors since this PR was opened

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -87,10 +87,12 @@ def request_client_login_params
       params.require(:state_file_request_intake_login_form).permit(:email_address, :sms_phone_number)
     end
 
+    # TODO: refactor multitenant service
```
mpidcock, line 90: this was done

db/migrate/20240626222401_create_state_file_wa_intakes.rb
```diff
@@ -0,0 +1,53 @@
+class CreateStateFileWaIntakes < ActiveRecord::Migration[7.1]
+  def change
+    create_table :state_file_wa_intakes do |t|
+      # misc
+      t.integer :tax_return_year
+      t.integer :filing_status
```
jenny-heath, line 6: this is wrong!! the intake should not have a `filing_status` column. it only looks like it does in places because the base intake class implements [a method that pulls it from the direct file data](https://github.com/codeforamerica/vita-min/blob/main/app/models/state_file_base_intake.rb#L97-L105)

db/migrate/20240626222401_create_state_file_wa_intakes.rb
```diff
@@ -0,0 +1,53 @@
+class CreateStateFileWaIntakes < ActiveRecord::Migration[7.1]
+  def change
+    create_table :state_file_wa_intakes do |t|
+      # misc
+      t.integer :tax_return_year
```
jenny-heath, line 5: this is also wrong!! the base intake class delegates this to `direct_file_data`

PR #4618, opened 2024-06-28
FYST Refactor: stop getting state code from bundle class, pull from intake instead

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- this one removes the `state_abbreviation` method from `IndividualReturn`s because we don't need yet another place where we get define state abbrevations
## How to test?
- any issues should be caught by tests


mrotondo (approved):

tofarr (approved):
🍰

PR #4619, opened 2024-06-28
FYST Refactor: IndividualReturn#return_type

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- this one removes the `return_type` method from `IndividualReturn`s and adds this info to the StateInformationService
- also renames primary_tax_form_name on the information service because those aren't actually the primary tax form names
## How to test?
- any issues should be caught by tests

### DIFF
--- a/app/lib/submission_builder/ty2022/states/az/individual_return.rb
+++ b/app/lib/submission_builder/ty2022/states/az/individual_return.rb
@@ -39,10 +39,6 @@ module SubmissionBuilder
             "AZ"
           end
 
-          def self.return_type
-            "Form140"
-          end
-
           def pdf_documents
             included_documents.map { |item| item if item.pdf }.compact
           end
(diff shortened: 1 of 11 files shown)


app/lib/submission_builder/ty2022/states/az/individual_return.rb
```diff
@@ -39,10 +39,6 @@ def self.state_abbreviation
             "AZ"
```
mrotondo, line 39: Out of curiosity, is this being refactored out in a separate PR? It seems like it would follow a similar pattern to this change.

mrotondo (approved):
Looks good!

app/lib/submission_builder/ty2022/states/az/individual_return.rb
mrotondo: Nevermind, I see that this is the next PR you sent!

PR #4620, opened 2024-06-28
alternative approach to minimal controller tests

mrotondo:
This is an alternative to PR 4616, where we obtain more coverage with less code at the cost of some clarity and redundancy.

Instead of creating a separate test file for each controller, this PR creates a "minimal controller test file" for each question navigation flow (GYR, NY, AZ) that just hits the edit action and renders the view for each controller in those flows. It has do some special casing for several of those actions to not redirect, which leads to a bit of unintuitive complexity in the test - there may be a clearer way to define those special cases in the code, but I just wanted to get it working for now and we can chat about it in computer gardening.

Also, this approach does not differentiate between controllers which already have tests and those that don't, so there may be some overlap between these expectations and ones in controller-specific tests, where we've created them.


mrotondo:
Team & I decided against this approach

PR #4621, opened 2024-07-01
FYST Refactor: dynamically define state-specific style rules

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- this one moves a couple of style rules into the loop where we already dynamically define state-specific styles
## How to test?
- styles still look good:
image
image


mrotondo (approved):

tofarr (approved):
Nice - I didn't know you could do that in SCSS!

PR #4622, opened 2024-07-01
FYST Refactor: move some methods from IndividualReturn classes to parent class SubmissionBuilder::Document

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- this PR moves some methods into the `SubmissionBuilder::Document` class
- this one i actually feel somewhat unsure about because these methods are exact duplicates across the `IndividualReturn` classes for each state but they _only_ apply to those classes. other classes that inherit from `SubmissionBuilder::Document` either:
  - don't reference them at all, in the case of state-specific supplementary documents like `SubmissionBuilder::Ty2022::States::Ny::Documents::It215` or 
  - have them defined separately, in the case of federal submission builders like `SubmissionBuilder::Ty2021::Documents::Irs1040`
- since we will be keeping the federal submission building code for another year or so this feels a little more complicated to me. please share your thoughts!
## How to test?
- any issues should be caught by the individual return tests


mrotondo (commented):
Would creating an intermediate class in the hierarchy called something like `StateFile::StateDocument < SubmissionBuilder::Document` which had these 5 methods on it, then making all the FYST documents inherit from that, feel better than moving them into the top-level document class where not all the subclasses will use them?

mrotondo:
> Would creating an intermediate class in the hierarchy called something like `StateFile::StateDocument < SubmissionBuilder::Document` which had these 5 methods on it, then making all the FYST documents inherit from that, feel better than moving them into the top-level document class where not all the subclasses will use them?

Re-reading your comment, this seems like it would have the same conceptual issues since there are other state file documents that don't use these methods. Maybe a mixin that the state individual returns include?

I'm failing to find where/how these methods are defined separately for the federal submission builders - can you point me at those definitions?

jenny-heath:
@mrotondo re: separately defined methods for federal, i guess i should have said _some_ of them are separately defined in _some_ places and as it turns out i didn't look that closely because the first example i could find for you is on an intermediate class (similar to what you suggested!) not on the child class: https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/shared/return1040.rb#L25-L27

mrotondo:
> @mrotondo re: separately defined methods for federal, i guess i should have said _some_ of them are separately defined in _some_ places and as it turns out i didn't look that closely because the first example i could find for you is on an intermediate class (similar to what you suggested!) not on the child class: https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/shared/return1040.rb#L25-L27

That's helpful, thanks. I think that following this pattern with a new intermediate subclass of document that's only inherited by FYST state individual return documents, and which includes these methods, seems like it'd be good for this PR.

mrotondo (approved):
This looks good structurally, and I think the names are fine but we're about to talk about that in CGC!

PR #4625, opened 2024-07-02
FYST Refactor: remove references to specific states logged in intakes in AuthenticatedStateFileIntakeConcern

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- this one removes the reference to current_state_file_{state}_intake in AuthenticatedStateFileIntakeConcern and includes the StateFileControllerConcern so we can reuse the code there that looks for any logged in state file intakes
## How to test?
- hopefully any issues should be caught by the tests

### DIFF
--- a/app/controllers/concerns/authenticated_state_file_intake_concern.rb
+++ b/app/controllers/concerns/authenticated_state_file_intake_concern.rb
@@ -1,5 +1,6 @@
 module AuthenticatedStateFileIntakeConcern
   extend ActiveSupport::Concern
+  include StateFile::StateFileControllerConcern
 
   included do
     before_action :require_state_file_intake_login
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/authenticated_state_file_intake_concern.rb
```diff
@@ -12,7 +13,7 @@ def card_postscript
   end
 
   def require_state_file_intake_login
-    if current_state_file_az_intake.blank? && current_state_file_ny_intake.blank?
+    if current_intake.blank?
```
tofarr, line 16: Good job - this will subtly change the behavior - previously we would not redirect to login in Ny if there was an Az intake present, and vice versa. Now we will

tofarr (approved):

PR #4630, opened 2024-07-05
GYR1-512 When filtering on an org to which the user does not have access, clear the invalid filter

tofarr:
## Is PM acceptance required?
## What was done?
- The filter panel stores filter orgs in a cookie. The current user may not have access to these orgs if...
  - They log out and log back in as a different user.
  - Somebody deletes the org
  - The user's access to the org is revoked.
- Before this change, this would trigger an error and effectively lock the user out of the clients page.
- If for any reason an org cannot be found, we now remove that filter rather than throwing an error.
## How to test?
- Test on Heroku
- Log in as an [admin@example.com](mailto:admin@example.com) / theforcevita
- Add [GetCTC.org](http://getctc.org/) and GYR National Organization as filter sites: 
image
- Log out.
- Log in as [lemon@example.com](mailto:lemon@example.com) / vitavitavitavita. Your cookie still includes the filter sites, but you don’t have access to them so you get an error.
## Risk Assessment
  - Minimal risk - I added a spec to verify the behavior
- Before
- After

### DIFF
--- a/app/controllers/concerns/client_sortable.rb
+++ b/app/controllers/concerns/client_sortable.rb
@@ -41,6 +41,7 @@ module ClientSortable
       id = id["id"] if id.instance_of?(Hash)
 
       vita_partner = @vita_partners.find { |p| p.id == id }
+      next unless vita_partner.present?
       {
         id: vita_partner.id,
         name: vita_partner.name,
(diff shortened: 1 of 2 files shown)


jenny-heath (approved):
looks good!

PR #4636, opened 2024-07-09
TBE-28 Allow transitioning intakes to all possible states when not in production

tofarr:
## Is PM acceptance required?
## What was done?
- Refactored all the transition code into a single location
- When not in production, state file admin users are permitted to transition an intake into any permitted next state.
- No change in production behaviour
- Offered alternative to PM where any notification could be triggered for any intake, but PM preferred this approach.
## How to test?
- Test on heroku.
- View a StateFileSubmission - you should see a new `Transition to State` section. (There is some overlap here between this and the existing buttons)
- Clicking on a state transitions the intake to that state.
- Risk Assessment
  - Change is isolated to non production environments, so risk is low.
### Non Production:
image

### Production
image

[TBE-28]: https://codeforamerica.atlassian.net/browse/TBE-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/hub/efile_submissions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    def transition_to
+      to_state = params[:to_state]
+      if %w[failed rejected].include?(to_state) && acts_like_production?
+        flash[:error] = "Transition to #{to_state} failed"
+        redirect_after_action
+        return
+      end
+
+      authorize! :update, @efile_submission
+      if @efile_submission.transition_to!(to_state, { initiated_by_id: current_user.id })
```
tofarr, line 34: lets attach an error code here if possible too

embarnard (changes requested):
This all looks great! I think like we were talking before I would prefer there to be some sort of artificial efile error attached to rejected and failed transitions since it will be what matches what happens in the poll for acknowledgment service and since last season it was critical that we had these errors attached for testing.

in the old code we did something like this
`@efile_submission.transition_to!(:rejected, error_code: EfileError.where(service_type: :state_file).last.code)`

app/controllers/hub/state_file/efile_submissions_controller.rb
```diff
@@ -59,6 +63,27 @@ def state_counts
...
+        to_state = params[:to_state]
+        if %w[failed rejected].include?(to_state) && acts_like_production?
+          flash[:error] = "Transition to #{to_state} failed"
+          redirect_to hub_state_file_efile_submission_path(id: @efile_submission.id)
+          return
+        end
+
+        authorize! :update, @efile_submission
+        metadata = { initiated_by_id: current_user.id }
+        if to_state == "rejected"
```
embarnard, line 76: Looks great! thanks for making that change

embarnard (approved):

PR #4640, opened 2024-07-10
FYST Refactor: multitenant service

jenny-heath:
## Link to pivotal/JIRA issue
## Is PM acceptance required?
## What was done?
- now that we have the StateInformationService we should stop relying on the MultiTenantService for any state-specific things. this is my attempt to make the MultiTenantService care only about _products_ (statefile vs gyr vs ctc)
## How to test?
- tests caught a few issues already so i think as long as they're passing it'll be ok?

### DIFF
--- a/app/controllers/portal/client_logins_controller.rb
+++ b/app/controllers/portal/client_logins_controller.rb
@@ -20,7 +20,8 @@ module Portal
           phone_number: @form.sms_phone_number,
           visitor_id: visitor_id,
           locale: I18n.locale,
-          service_type: service_type
+          service_type: service_type,
+          state_code: state_code
         )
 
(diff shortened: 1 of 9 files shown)


app/controllers/portal/client_logins_controller.rb
```diff
@@ -20,7 +20,8 @@ def create
           phone_number: @form.sms_phone_number,
           visitor_id: visitor_id,
           locale: I18n.locale,
-          service_type: service_type
+          service_type: service_type,
+          state_code: state_code
```
jenny-heath, line 24: intake_logins_controller inherits from this so when dealing with gyr clients it'll have a nil state code and when statefile it will pass in a state code. this is used to generate the url that goes in the message

tofarr (approved):

PR #4642, opened 2024-07-11
Nj/kd/minor setup changes

powersurge360:
## Link to pivotal/JIRA issue
No Jira ticket

## Is PM acceptance required?
## What was done?

The extraction of efiles would fail on multiple runs of `bin/setup`
due to the files already existing. This change allows for the files to
just simply overwrite the existing files, which I think is what we want.

The other thing was I saw that as of December we no longer had adoptopenjdk8 in the Brewfile. I think that because these are just jvm bytecode we can run them on a more modern JVM version as byte code is supposed to be forward compatible. Indeed, I was able to test it partially by running `java -cp vendor/gyr_efiler/gyr-efiler-classes-7bf94e422032527b78cb169fc8674cfe9fd14554.zip org.codeforamerica.gyr.efiler.App vendor/gyr_efiler/gyr-efiler-config-7bf94e422032527b78cb169fc8674cfe9fd14554.zip ` (which is the correct hashes for my system) and get java output.

### efile extraction change
- run `bin/setup` on a project that is already set up
- Note that it completes without errors

### adoptopenjdk
- Remove adoptopenjdk with `brew uninstall adoptopenjdk`
- Run test command `bin/rspec spec/lib/pdf_filler/az140_pdf_spec.rb`
- Note failure
- Run `asdf install`
- Run test command `bin/rspec spec/lib/pdf_filler/az140_pdf_spec.rb`
- Note success

- Risk Assessment
There may be a reason why the Brewfile was changed that maybe it will cause problems somewhere. I think the brew file should only be used on Macs so unless we are deploying to Macs we're probably ok.

There's a chance I'm wrong about how the bytecode works. I'm pretty confident and this change should only impact development anyways. I think we should go for it. I've left a note in the README and if it breaks for folks there's no harm in going right back.


Brewfile
```diff
@@ -2,12 +2,12 @@ brew "gh"
 brew "ghostscript"
 brew "imagemagick"
 # tap "AdoptOpenJDK/openjdk"
-# cask "adoptopenjdk8"
+# cask "adoptopenjdk"
```
mpidcock, line 5: do you think we should remove these lines if we are pointing people to use asdf instead?

mpidcock (approved):
could merge but one minor comment

Brewfile
powersurge360: Yea, could do. I left it in case we didn't want to keep it but it's all in version history anyways so might as well. I'll amend and remerge main.

tofarr (approved):

PR #4650, opened 2024-07-17
FYST-195 No longer skipping data review step

tofarr:
## Is PM acceptance required?
## What was done?
- We had been skipping the data review step. As discussed with Tiffany at standup we no longer are
## How to test?
- Test on Heroku. After the data transfer step, you should see this
image

[FYST-195]: https://codeforamerica.atlassian.net/browse/FYST-195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/assets/stylesheets/_state-file.scss
```diff
@@ -748,6 +748,13 @@ $state-colors: (
   }
 }
 
+.federal-info-outer {
+  .dev-button {
+    margin: 0.5rem auto;
+    display: block;
+  }
+}
+
```
tofarr, line 757: I added this style here so that the "Continue Skipping schema validation" button would not get it's text truncated:
image

spec/controllers/state_file/questions/data_review_controller_spec.rb
```diff
@@ -18,7 +18,7 @@
         expect(efile_info.ip_address.to_s).to eq "72.34.67.178"
         expect(efile_info.device_id).to eq nil
         expect(efile_info.intake).to eq intake
-        expect(response).to redirect_to("/en/az/questions/name-dob")
+        expect(response).to be_ok
```
mpidcock, line 21: I would prefer the more specific condition `render_template :edit` instead of `be_ok`

spec/features/state_file/editing_df_xml_spec.rb
```diff
@@ -27,7 +27,7 @@
 
     xml_before = StateFileNyIntake.last.raw_direct_file_data.strip
 
-    expect(page).to have_text I18n.t('state_file.questions.name_dob.edit.title1')
+    expect(page).to have_text I18n.t('state_file.questions.data_review.edit.title')
```
mpidcock, line 30: Since this test is also ensure we can edit the DF xml, this test should bring back the check for the federal info controller similar to how it is here: https://github.com/codeforamerica/vita-min/pull/4155/files#diff-904908460ac030efc6510fd69668d5a2c9fa2e5438f5d885872601e28f420815

I don't know how much coverage we should have on the federal info form, but it did break silently in the past, so we should consider adding increased coverage this year

mpidcock (commented):
Looks good, just some testing comments

spec/controllers/state_file/questions/data_review_controller_spec.rb
tofarr: Done

spec/features/state_file/editing_df_xml_spec.rb
tofarr: Done - I'm not really doing much in the federal info controller (Since it is still skipped over in non dev contexts), but we make sure it does load and redirect into the NameDobController

mpidcock (approved):
Thank you!

PR #4653, opened 2024-07-17
Gyr1 532/correct unsafe send

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-532

## Is PM acceptance required?

## What was done?
*Note* this is an excerpt from my work notes. See the ticket to see the background unabridged.
- ~The ideal thing would be to create a viewmodel that has the validation but it's perhaps too heavyweight of a solution for this minor change.~
- ~We probably should be careful about precedent though. If much more validation happens then we'll need to adjust it regardless.~
- ~How about a simple set of =require= checks and then run a =delete= on the hash if the parameter is not in the whitelist? Similar solution but prevents accidentally removing the check without noticing.~
- Looks like it didn't like that. Instead, drop the params and just go through with a switch statement.

## How to test?
Unsure. Will poke around and update. Unit tests are green though.


jnf (approved):
works for me

embarnard (approved):
Looks good!

PR #4655, opened 2024-07-18
Fyst 126 handle non matching state code in url and intake

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-126
## Is PM acceptance required?
## What was done?
- The routes (URLs) for all state file pages had a state code (az/ny) in them, even though there should always be an intake in the session for the current state. Therefore, we chose to remove the state code from the routes for all pages except the landing pages and the faq pages.
- We also now delete _all_ intakes from the session when creating a new one at the beginning of the flow, not just ones for the selected state.
## How to test?
- I made sure all existing tests passed once the state code param was removed
- I added a bunch of new controller tests to make sure that the state file questions views load successfully after the change
- I imported the faqs locally and checked that those pages still load manually
- I partially went through the flows manually to make sure that they still work
- Risk Assessment
  - Almost any link in the app could be broken by this change - I am 95% sure that they aren't but not 100% so we should click on everything, especially in the header and footer.
  - We should also confirm that pages still lay out the same and that the state colors still show up correctly everywhere.


app/controllers/concerns/state_file/eligibility_offboarding_concern.rb
```diff
@@ -20,7 +20,7 @@ def next_path
     end
 
     def from_path_params
-      [:us_state, :return_to_review].each_with_object({}) do |key, path_params|
+      [:return_to_review].each_with_object({}) do |key, path_params|
         path_params[key] = params[key] if params[key].present?
       end
     end
```
jenny-heath, line 26: minor thing but this doesn't seem like it needs to be its own method anymore? maybe?

app/controllers/concerns/state_file/state_file_controller_concern.rb
```diff
@@ -3,26 +3,7 @@ module StateFileControllerConcern
...
-      if current_intake
-        current_intake.state_code
-      else
-        params[:us_state]
-      end
-    end
-
-    def current_state_name
-      StateFile::StateInformationService.state_name(current_state_code)
+      helper_method :current_tax_year
```
jenny-heath, line 6: should this just be rolled in to StateFileIntakeConcern? having trouble thinking through whether that makes sense

app/controllers/flows_controller.rb
```diff
@@ -212,11 +212,6 @@ def controller_url
           if respond_to?(:resource_name) && resource_name.present?
             url_params[:id] = "fake-#{resource_name}-id"
           end
-          if @current_controller.params[:id] == 'state_file_az'
-            url_params[:us_state] = 'az'
-          elsif @current_controller.params[:id] == 'state_file_ny'
-            url_params[:us_state] = 'ny'
-          end
```
jenny-heath, line 219: reminder to myself to ask about this

app/jobs/request_verification_code_for_login_job.rb
```diff
@@ -34,17 +34,7 @@ def perform(email_address: nil, phone_number: nil, locale:, visitor_id:, service
...
-                  us_state: state_code,
-                  full_url: true,
-                  host: URI(Rails.configuration.statefile_url).hostname,
-                  locale: locale,
-                  protocol: "https"
-                )
-              else
-                multi_tenant_service.url(locale: locale)
-              end
+        url = multi_tenant_service.url(locale: locale)
```
jenny-heath, line 37: there does seem to be a slight difference in behavior here? `multi_tenant_service.url` returns the root path whereas this used to return the first page in the flow? which admittedly is a weird difference to have between the products but just wanted to make sure we talk about it

app/controllers/portal/client_logins_controller.rb
```diff
@@ -49,7 +48,7 @@ def check_verification_code
         return if redirect_locked_clients # check if any records are already locked
         if @records.present? # we have at least one match and none are locked
           DatadogApi.increment("#{self.controller_name}.verification_codes.right_code")
-          redirect_to self.class.to_path_helper(action: :edit, id: hashed_verification_code, **extra_path_params)
```
jenny-heath, line 52: i made this controller and i just want to say thanks i hate it

jenny-heath: *er, i mean i made the `state_file/intake_logins_controller.rb` along with these changes to the client logins controller

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -91,35 +85,30 @@ def service_type
...
+      unfinished_logged_in_intake_ids = StateFile::StateInformationService.active_state_codes.map do |state_code|
+        session.dig("warden.user.state_file_#{state_code}_intake.key", 0)
+      end.compact
 
       sign_in intake
 
-      if unfinished_logged_in_intake_id.present? && unfinished_logged_in_intake_id[0] != intake.id
-        intake.update(unfinished_intake_ids: intake.unfinished_intake_ids + unfinished_logged_in_intake_id)
+      unfinished_logged_in_intake_ids.each do |unfinished_logged_in_intake_id|
+        if unfinished_logged_in_intake_id[0] != intake.id
```
jenny-heath, line 102: is this meant to be `if unfinished_logged_in_intake_id != intake.id`?

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -91,35 +85,30 @@ def service_type
...
+          intake.update(unfinished_intake_ids: intake.unfinished_intake_ids + unfinished_logged_in_intake_id)
+        end
```
jenny-heath, line 104: is `unfinished_logged_in_intake_ids` always an array of arrays with a single element? if so could we flatten it so we don't have to [0] each time?

app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -91,35 +85,30 @@ def service_type
...
       end
 
       to_path = session.delete(:after_state_file_intake_login_path)
       unless to_path
         controller = intake.controller_for_current_step
-        # This can be removed once we are sure nobody has current_step set to landing_page
```
jenny-heath, line 115: why i don't understand :(

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -73,8 +73,6 @@ def path_for_step(step)
       # by default, most state file questions have no illustration
       def illustration_path; end
 
-      def card_postscript; end
```
jenny-heath, line 76: should we go around removing all of these since the view does first check `respond_to?`?

spec/jobs/request_verification_code_for_login_job_spec.rb
```diff
@@ -235,7 +233,7 @@
             expect(TextMessageVerificationCodeService).not_to have_received(:request_code)
             expect(TwilioService).to have_received(:send_text_message)
                                        .with(a_hash_including(
-                                               body: I18n.t("state_file.intake_logins.no_match_sms", url: "https://statefile.test.localhost/en/az/questions/landing-page", locale: :en),
```
jenny-heath, line 238: see comment on main file

jenny-heath (approved):
oh my god

PR #4656, opened 2024-07-19
get rid of checkboxes in pr template

mrotondo:
they were adding this "1/2 tasks done" UI to all our PRs, and that was annoying


jnf (approved):
love it

PR #4657, opened 2024-07-22
GYR1-508 Preventing access to bank details for greeters

tofarr:
## What was done?
- Added a check to return a 404 if a greeter accesses the BankAccountsController
- Added a check in the clients show.html.erb to hide the bank details section if the user is a greeter.
- Added specs to cover cases where a user can view bank details and when they can't.
## How to test?
- Test on heroku.
- Create a client that has bank details.
- Try to view the client as a non greeter, and you should see the bank details.
- Try to view the client as a greeter and you should not see the bank details
## Risk Assessment
  - Low risk - preventing access to something that was accessible before. Specs mitigate inadvertent access removal

### Greeter does not see bank details
image

### Others still see bank details
image

[GYR1-508]: https://codeforamerica.atlassian.net/browse/GYR1-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -87,7 +87,7 @@ class Ability
     cannot :manage, StateFileW2
     cannot :manage, StateId
 
-    if user.role_type == CoalitionLeadRole::TYPE
+    if user.coalition_lead?
       can :read, Coalition, id: user.role.coalition_id
 
       # Coalition leads can view and edit users who are coalition leads, organization leads, site coordinators, and team members in their coalition
@@ -100,7 +100,7 @@ class Ability
       can :manage, TeamMemberRole, sites: { parent_organization: { coalition: user.role.coalition } }
     end
 
-    if user.role_type == OrganizationLeadRole::TYPE
+    if user.org_lead?
 
       # Organization leads can view and edit users who are organization leads, site coordinators, and team members in their coalition
       can :manage, User, id: user.accessible_users.pluck(:id)
@@ -111,7 +111,7 @@ class Ability
       can :manage, TeamMemberRole, sites: { parent_organization: user.role.organization }
     end
 
-    if user.role_type == SiteCoordinatorRole::TYPE
+    if user.site_coordinator?
       # Site coordinators can create site coordinators and team members in their site
       can :manage, SiteCoordinatorRole do |role|
         user.role.sites.map.any? { |site| role.sites.map(&:id).include? site.id }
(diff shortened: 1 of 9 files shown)


app/lib/ability.rb
```diff
@@ -87,7 +87,7 @@ def initialize(user)
     cannot :manage, StateFileW2
     cannot :manage, StateId
 
-    if user.role_type == CoalitionLeadRole::TYPE
+    if user.coalition_lead?
```
tofarr, line 90: Using shorter syntax here

spec/controllers/hub/clients/bank_accounts_controller_spec.rb
```diff
@@ -59,6 +59,15 @@
           expect(response.body).to include(archived_bank_account.routing_number)
         end
       end
+
+      context "for a greeter" do
+        let(:user) { create(:greeter_user) }
+
+        it "prevents access" do
+          get :show, params: params, format: :js, xhr: true
+          expect(response).to be_forbidden
```
jnf, line 68: nice

jnf (approved):
looks good!

powersurge360 (commented):
Why not use permissions to protect the given properties and then check `can? :read, :bank_properties`? Would allow a more robust check for banking details and allow stuff defined in the `ability.rb` to propagate. Are we moving away from CanCanCan?

app/views/hub/clients/_hidden_bank_account_info.html.erb
```diff
@@ -2,7 +2,7 @@
 <div>
   <div class="field-display">
     <span class="form-question"><%= t("hub.bank_accounts.bank_name") %>:</span>
-    <span class="label-value" aria-hidden="true"><%= bank_account.bank_name %></span>
+    <span class="label-value" aria-hidden="true">●●●●●●●●●●●</span>
```
tofarr, line 5: Redacting this to avoid leaking data.

app/views/hub/clients/show.html.erb
```diff
@@ -96,7 +96,7 @@
             </div>
             <div class="field-display">
               <span class="form-question"><%= t(".filing_status") %>:</span>
-              <div class="label-indented"><%= filing_status(@client) %></div>
+              <span class="label-value"><%= filing_status(@client) %></span>
```
tofarr, line 99: Fix janky layout

PR #4658, opened 2024-07-22
TBE-87 : Switching to buffered consumer

tofarr:
## What was done?
- Switched to using mixpanel's buffered consumer - now sending events to mixpanel in batches of 50 instead of 1 at a time (So we will be sending 1/50 the amount of requests and will not be subject to throttling unless our usage increases drastically!)
- Added a resetting delayed task that flushes the buffer after 5 seconds of inactivity
- Added a mutex to prevent duplication due to tasks tripping over each other (Initial draft duplicated events in mixpanel due to thread safety issues)
### No measurable change in average response time detected in tests
image

### Below is a small sample - I refreshed a page 5 times.
10 entries appeared in Datadog. Events seem to be duplicated in datadog ONLY in the staging environment. (This is not related to this ticket - I put main back in there and the same behavior occurs - I think it is related to the configuration in aptible - I created this ticket for it https://codeforamerica.atlassian.net/browse/TBE-88)
https://app.datadoghq.com/logs?query=service%3Avita-min-staging-web%20FaqController%20&agg_m=count&agg_m_source=base&agg_q=%40controller&agg_q_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&top_n=100&top_o=top&viz=stream&x_missing=true&from_ts=1721945938835&to_ts=1721946838835&live=true
image

5 entries appear in mixpanel:
image

### I ran the jmeter load test for 60 seconds against staging, and all events appeared in mixpanel without losses!
image
image

[TBE-87]: https://codeforamerica.atlassian.net/browse/TBE-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/mixpanel_service.rb
+++ b/app/services/mixpanel_service.rb
@@ -16,8 +16,7 @@ class MixpanelService
     mixpanel_key = Rails.application.credentials.dig(:mixpanel_token)
     return if mixpanel_key.nil?
 
-    #@consumer = Mixpanel::BufferedConsumer.new
-    @consumer = Mixpanel::Consumer.new
+    @consumer = Mixpanel::BufferedConsumer.new(nil, nil, nil, 10)
     @tracker = Mixpanel::Tracker.new(mixpanel_key) do |type, message|
       send_event_to_mixpanel(type, message)
     end
@@ -54,7 +53,6 @@ class MixpanelService
 
   def send_event_to_mixpanel(type, message, num_attempts = 1, delay = 0)
     task = Concurrent::ScheduledTask.new(delay) do
-      Rails.logger "TRACE:send_event_to_mixpanel #{type} #{message}"
       @consumer.send!(type, message)
     rescue StandardError => err
       if num_attempts >= MAX_ATTEMPTS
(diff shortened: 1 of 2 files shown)


app/services/mixpanel_service.rb
```diff
@@ -37,6 +37,7 @@ def initialize
   # @param [Hash] data: (optional, defaults to {}) data to be sent to mixpanel
   #
   def run(distinct_id:, event_name:, data: {})
+    # ZZZZ
```
jnf, line 40: ?

tofarr: Removed

powersurge360 (approved):
I don't see why it shouldn't work. It does make me a little squeamish to start having mutexes and things. I guess in the context of the queue taking a long time to fill up, how long are we talking? I imagine if we're hitting a throttling limit then we're sending pretty frequently. Could it be reasonable to only flush manually when the program is exiting and otherwise let it flush as it hits the end of its queue? Or alternatively, reduce the `max_length` of the buffered consumer?

Approving for now but would love another set of eyes regardless.

powersurge360:
After some discussion in slack, looks like there's no better way to do this, the canonical examples are at least as complex. Feel better about the solution.

PR #4663, opened 2024-07-24
TBE 85 Bugcrowd: Add verification for emails for unsubscribe/subscribe 

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/TBE-85

## What was done?
A bugrowd issue was brought up where any bad actor can pass in an email address to the unsubscribe page [en/unsubscribe_from_emails?email_address=] and can unsubsribe someone else. We added verification so that one can only unsubscribe an email address if they click the link from an email it is in, otherwise it the page will say that it is "invalid link"

## How to test?
Check on being a bad actor: Check to see on heroku using the following https://gyr-review-app-4663-534256ffcb66.herokuapp.com/en/unsubscribe_from_emails?email_address=<"email string"> and passing in a email address or any string. It should say invalid unsubscribtion link. 

You can also check that the origin flow and app work as before. This can be done by completing and intake for FYST or GYR on the heroku link, getting a welcome email, clicking the unsubscribe link and checking on the client in the hub to see that they are unsubscribed.

### DIFF
--- a/app/controllers/concerns/email_subscription_updater.rb
+++ b/app/controllers/concerns/email_subscription_updater.rb
@@ -1,0 +1,33 @@
+module EmailSubscriptionUpdater
+  extend ActiveSupport::Concern
+
+  def update_email_subscription(direction:, column_name:, show_flash_and_render: false)
+    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
+
+    if params[:email_address].blank?
+      flash[:alert] = "No record found"
+      return
+    end
+
+    begin
+      email_address = verifier.verify(params[:email_address])
+      matching_intakes = matching_intakes(email_address)
+
+      if matching_intakes.present?
+        matching_intakes.each do |intake|
+          intake.update(column_name => direction)
+        end
+
+        if show_flash_and_render
+          flash[:notice] = I18n.t("notifications_settings.subscribe_to_emails.flash")
+          render :unsubscribe_from_emails
+        end
+      else
+        flash[:alert] = I18n.t("notifications_settings.no_record")
+      end
+    rescue ActiveSupport::MessageVerifier::InvalidSignature
+      flash[:alert] = I18n.t("notifications_settings.invalid_link")
+    end
+  end
+end
+
--- a/app/controllers/notifications_settings_controller.rb
+++ b/app/controllers/notifications_settings_controller.rb
@@ -3,6 +3,11 @@ class NotificationsSettingsController < ApplicationController
   def unsubscribe_from_emails
     verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
 
+    if params[:email_address].blank?
+      flash[:alert] = "No record found"
+      return
+    end
+
     begin
       email_address = verifier.verify(params[:email_address])
       matching_intakes = matching_intakes(email_address)
@@ -21,6 +26,11 @@ class NotificationsSettingsController < ApplicationController
 
   def subscribe_to_emails
     verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
+
+    if params[:email_address].blank?
+      flash[:alert] = "No record found"
+      return
+    end
 
     begin
       email_address = verifier.verify(params[:email_address])
--- a/app/controllers/notifications_settings_controller.rb
+++ b/app/controllers/notifications_settings_controller.rb
@@ -1,29 +1,43 @@
 class NotificationsSettingsController < ApplicationController
 
   def unsubscribe_from_emails
-    matching_intakes = matching_intakes(params[:email_address])
+    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
 
-    if matching_intakes.present?
-      matching_intakes.each do |intake|
-        intake.update(email_notification_opt_in: "no")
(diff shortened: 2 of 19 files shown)


app/controllers/notifications_settings_controller.rb
jenny-heath, line 1: this file (and the state file one) is pretty repetitive and this seems like a good opportunity to refactor it to reuse more (since the edits you're making have to be made in 4 different places). happy to pair on that if you're interested!

app/controllers/notifications_settings_controller.rb
```diff
@@ -1,29 +1,53 @@
...
-    matching_intakes = matching_intakes(params[:email_address])
+    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
 
-    if matching_intakes.present?
-      matching_intakes.each do |intake|
-        intake.update(email_notification_opt_in: "no")
-      end
-    else
+    if params[:email_address].blank?
       flash[:alert] = "No record found"
```
jenny-heath, line 7: oh we should add this and "Invalid unsubscribe link" to translations

app/controllers/notifications_settings_controller.rb
jenny-heath: sketched it out as an experiment:
```Ruby
  def unsubscribe_from_emails
    update_email_subscription(direction: "no", column_name: :email_notification_opt_in)
  end

  def subscribe_to_emails
    update_email_subscription(direction: "yes", column_name: :email_notification_opt_in, show_flash_and_render: true)
  end

  def update_email_subscription(direction: "no", column_name:, show_flash_and_render: false)
    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)

    if params[:email_address].blank?
      flash[:alert] = "No record found"
      return
    end

    begin
      email_address = verifier.verify(params[:email_address])
      matching_intakes = matching_intakes(email_address)

      if matching_intakes.present?
        matching_intakes.each do |intake|
          intake.update("#{column_name}": direction)
        end

        if show_flash_and_render
          flash[:notice] = I18n.t("notifications_settings.subscribe_to_emails.flash")
          render :unsubscribe_from_emails
        end
      else
        flash[:alert] = "No record found"
      end
    rescue ActiveSupport::MessageVerifier::InvalidSignature
      flash[:alert] = "Invalid unsubscribe link"
    end
  end
```

and then the corresponding code for the state file one would just be:
```Ruby
    def unsubscribe_from_emails
      update_email_subscription(direction: true, column_name: :unsubscribed_from_email)
    end

    def subscribe_to_emails
      update_email_subscription(direction: false, column_name: :unsubscribed_from_email, show_flash_and_render: true)
    end
```
with `StateFile::NotificationsSettingsController` inheriting from this NotificationsSettingsController

tahsinaislam: wow i love this thank you!

config/routes.rb
```diff
@@ -610,8 +610,8 @@ def scoped_navigation_routes(context, navigation)
         get "/coming-soon", to: "state_file_pages#coming_soon"
         post "/clear_session", to: 'state_file_pages#clear_session'
         get "/privacy-policy", to: "state_file_pages#privacy_policy"
-        get "/unsubscribe_email", to: "notifications_settings#unsubscribe_email", as: :unsubscribe_email
-        post "/subscribe_email", to: "notifications_settings#subscribe_email", as: :subscribe_email
```
jenny-heath, line 614: re: pair review, maybe set up temporary redirect for existing links that have been emailed out

app/controllers/concerns/email_subscription_updater.rb
```diff
@@ -0,0 +1,33 @@
+module EmailSubscriptionUpdater
+  extend ActiveSupport::Concern
+
+  def update_email_subscription(direction:, column_name:, show_flash_and_render: false)
+    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
+
+    if params[:email_address].blank?
+      flash[:alert] = "No record found"
```
jenny-heath, line 8: also put i18n key here

app/controllers/concerns/email_subscription_updater.rb
```diff
@@ -0,0 +1,33 @@
+module EmailSubscriptionUpdater
```
jenny-heath, line 1: i think all concerns have "Concern" at the end of the name, consider renaming

app/mailers/diy_intake_email_mailer.rb
```diff
@@ -4,14 +4,18 @@ def high_support_message(diy_intake:)
     @diy_intake = diy_intake
     service = MultiTenantService.new(:gyr)
     attachments.inline['logo.png'] = service.email_logo
+
+    verifier = ActiveSupport::MessageVerifier.new(Rails.application.secret_key_base)
+    signed_email = verifier.generate(diy_intake.email_address)
```
jenny-heath, line 9: consider adding shared example for testing that the unsubscribe link works (has a signed email) i.e. `shared_example :has_a_working_unsubscribe_link` for all mailers

jenny-heath (approved):
2 blocking things discussed:
- one translation string not referencing i18n
- discussion with PM about changing the unsubscribe url

PR #4664, opened 2024-07-24
GYR1-506  - Limit Greeter access to clients

powersurge360:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/GYR1-506

## What was done?
The Greeter role had it's `:manage` permission for clients taken away and more granular `:edit` and `:read` permissions were given instead. Now the permissions are granted conditional to the properties of the client the greeter is trying to view.

## How to test?
**NOTE:** There are many permissions and many custom pages for many roles. While the scope of this is only to reduce permissions for one role, correcting is most secure by building up permissions rather than tearing down permissions. The unit tests check for manage for all roles, so existing roles should be unscathed but custom actions may be broken for greeters. Check thoroughly. See the self-review for more details.

### Prerequisites

Ideally, there will be at least three clients.

1. A client attached to an organization **with** `allow_greeters: true` **with** a tax return in one of the whitelisted states (for example `intake_ready`)
2. A client attached to an organization **with** `allow_greeters: true` **without** a tax return in a whitelisted state
3. A client attached to an organization **without** `allow_greeters: true`.

### Testing clients with allowed_greeters orgs

1. Visit `/en/hub/clients` as a greeter
2. Verify that the filter drop down has only the four whitelisted states:
     1. Ready for Review
     2. Greeter - info requested
     3. Needs doc help
     4. Not filing
3. Verify that the clients shown are only ones that are in a whitelisted state
4. Verify that attempting to visit a clients page (`/en/hub/clients/:id`) directly (by looking up the id) when they are not in the correct state 403s.
5. Verify that visiting a clients page when they are in the correct state, works
6. Verify that changing the state via `edit_take_action` works.

### Testing clients without allowed_greeters orgs

No clients from these groups should show up on the index page or any of the detail pages (where it will give a 403).

- Before
Screenshot 2024-07-26 at 11 28 34 AM

- After
Screenshot 2024-07-26 at 11 27 00 AM


app/lib/ability.rb
```diff
@@ -7,6 +7,15 @@ def initialize(user)
...
 
+    # Custom actions
+    alias_action :flag, :toggle_field, :edit_take_action, :update_take_action,
+                 :unlock, :edit_13614c_form_page1, :edit_13614c_form_page2,
+                 :edit_13614c_form_page3, :save_and_maybe_exit,
+                 :update_13614c_form_page1, :update_13614c_form_page2,
+                 :update_13614c_form_page3, :cancel_13614c,
+                 :resource_to_client_redirect,
+                 to: :hub_client_management
+
```
powersurge360, line 18: I think there are two kinds of actions in here. Some, like `flag`, `unlock`, and `toggle_field` probably make since to be real permissions. Others, like `edit_13614c_form_page1` feel like they should probably be their own sub resource of client (even if it's just a PORO with `ActiveModel` included). In this way, you can tidy the controller and the permissions and use the authorize parent resource functionality of CanCanCan.

I didn't go that far because it's a significant refactor and I felt like this was a good first step anyways.

app/lib/ability.rb
```diff
@@ -58,8 +67,32 @@ def initialize(user)
...
 
-    # Anyone can manage clients and client data in the groups they can access
-    can :manage, Client, vita_partner: accessible_groups
+    # This was overly permissive. We should work out what the permissions should
+    # be for each role and reduce this check. As we need to modify this, please
+    # break out the role and specify permissions more granularly
+    client_role_whitelist = [
+      :client_success, :admin, :org_lead, :site_coordinator,
+      :coalition_lead, :state_file_admin, :team_member
+    ].freeze
```
powersurge360, line 76: These are every role I could find in the codebase to restore existing functionality. Moving from a 'everyone has permissions unless taken away' to a 'no one has permissions unless given' is secure by default while creating a little more friction. Moving permissions around might be painful in the short term but in the long term will save us from having to put role-specific things in views or controllers and will overall make the application easier to reason about.

With as many roles as we have, it may even make sense to break out into sub ability files for each role down the line.

app/lib/client_sorter.rb
```diff
@@ -27,7 +27,6 @@ def active_filters
 
   def filtered_and_sorted_clients(default_order: nil)
     @default_order = default_order
-    setup_sortable_client unless @filters.present?
```
powersurge360, line 30: This line appears to never be possible to hit, and if it ever were hit it would throw an exception. I removed it while I was in there.

app/lib/ability.rb
```diff
@@ -58,8 +67,32 @@ def initialize(user)
     can :read, Organization, id: accessible_groups.pluck(:id)
     can :read, Site, id: accessible_groups.pluck(:id)
 
-    # Anyone can manage clients and client data in the groups they can access
-    can :manage, Client, vita_partner: accessible_groups
+    # This was overly permissive. We should work out what the permissions should
+    # be for each role and reduce this check. As we need to modify this, please
+    # break out the role and specify permissions more granularly
```
tofarr, line 72: I don't think this comment is applicable anymore!

app/lib/ability.rb
tofarr: Great work. I totally agree abilities should almost always be whitelists.

tofarr (approved):

app/lib/ability.rb
powersurge360: The comment was guidance for folks who have to come in and manipulate this part of the file in the future. If you feel it's still unnecessary with the additional context I can cut it out. The idea that I was hoping for is that as people come in and touch it more of these can be broken out into their own role specific permission whitelists.

PR #4671, opened 2024-07-29
FYST 185 Add pages for public school contributions

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-185

## Is PM acceptance required?

## What was done?
Pages have been added to arizona flow for public school contributions. It follows the format of unemployment controller where there is an index of all contribution where the client can edit and add more contributions
 
## How to test?
Go on the heroku link and fill out the flow for arizona. Make sure to change the session toggles for heroku to be during intake time. 

Screenshot 2024-08-05 at 12 27 44 PM
Screenshot 2024-08-05 at 12 32 11 PM
Screenshot 2024-08-05 at 12 35 40 PM


tofarr (approved):
🍰

tofarr (approved):

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+        @az322_contributions = current_intake.az322_contributions
+        if @az322_contribution.made_contribution_no?
+          return redirect_to next_path
+        end
+
+        if @az322_contribution.valid?
+          @az322_contribution.save
+          redirect_to action: :index, return_to_review: params[:return_to_review]
+        else
+          render :new
```
embarnard, line 35: when it fails to validate will it render new again with the failed form fields?

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      def edit
+        @az322_contribution = current_intake.az322_contributions.find(params[:id])
+        @filing_year = Rails.configuration.statefile_current_tax_year
+      end
+
+      def update
+        @az322_contribution = current_intake.az322_contributions.find(params[:id])
+        @filing_year = Rails.configuration.statefile_current_tax_year
```
embarnard, line 46: since this is used for several actions i wonder if we can just add it to the top instead in something like `before_action :load_filing_year, only: [:new, :create, :edit, :update]` and def load_filing_year; @filing_year = Rails.configuration.statefile_current_tax_year; end;`

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+          :ctds_code,
+          :district_name,
+          :amount,
+          :date_of_contribution_day,
+          :date_of_contribution_month,
+          :date_of_contribution_year
+        )
+      end
+
+      def date_of_contribution(year, month, day)
```
embarnard, line 85: is this actually getting used anywhere?

embarnard: if so can probably remove datehelper at the top too

app/models/az322_contribution.rb
```diff
@@ -17,5 +18,44 @@
...
+  validate :date_of_contribution_is_valid_date, if: -> { made_contribution == "yes" }
+
+  attr_accessor :date_of_contribution_day, :date_of_contribution_month, :date_of_contribution_year
+
+  before_validation :set_date_of_contribution, if: -> { made_contribution == "yes" }
+
+  private
+
+  def set_date_of_contribution
+    if date_of_contribution_year.present? || date_of_contribution_month.present? || date_of_contribution_day.present?
```
embarnard, line 40: I added this check because in the command line I'm assuming we will just try to set date_of_contribution directly instead of breaking it up by month, day, year etc

app/models/az322_contribution.rb
```diff
@@ -17,5 +18,44 @@
...
+  end
+
+  def parse_date_params(year, month, day)
+    Date.new(year.to_i, month.to_i, day.to_i) rescue nil
+  end
+
+  def date_of_contribution_is_valid_date
+    if date_of_contribution_year.present? || date_of_contribution_month.present? || date_of_contribution_day.present?
+      valid_text_date(date_of_contribution_year, date_of_contribution_month, date_of_contribution_day, :date_of_contribution)
+    else
```
embarnard, line 52: added this so that we can also just add a single value for date of contribution in the console instead of day, month, year -- should function similarly though

embarnard (changes requested):
Looks really good! Pushed a small change to the Az322Contribution model file, lmk if it doesn't make sense to you and left a few other comments

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
tahsinaislam: yes to clear any validations errors

embarnard (approved):

PR #4672, opened 2024-07-29
GYR1-539 Better error handling for Twilio attachments

tofarr:
## What was done?

We've had this weird thing happening since July 10th where attachments from Twilio sometimes fail to load. There was no change on our end at this time that I can see. Twilio invokes our webhook with a request that includes a url where we should load the attachment, but from what I can see it simply does not load. Interestingly, I have been able to load these attachments later using the same URL. My working theory is that there is some kind of race condition on Twilio's end, where they invoke the webhook before the data is actually available in S3. But I need more data - hence this PR.

TLDR: Rather than simply having the process fail with no explanation, this PR captures the ArgumentError we have seen in the logs and returns a broken attachment mimicking what happens if the attachment does load but is of an unsupported type. It also prints some additional info to the log.

## How to test?
- Specs still pass
- I also tested this on Staging: (Sent an image from my phone via text)
image

## Risk Assessment
  - This code will only be invoked in cases where the read attachment has failed anyway - and provides some better error handling. Low risk

[GYR1-539]: https://codeforamerica.atlassian.net/browse/GYR1-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/twilio_service.rb
+++ b/app/services/twilio_service.rb
@@ -25,8 +25,8 @@ class TwilioService
     end
 
     def fetch_attachment(url)
-      response = Net::HTTP.get_response(URI(url)) # first we get a redirect from Twilio to S3
       begin
+        response = Net::HTTP.get_response(URI(url)) # first we get a redirect from Twilio to S3
         response = Net::HTTP.get_response(URI(response['location'])) # then we get a redirect from S3 to S3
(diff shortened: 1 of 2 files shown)


embarnard (approved):
Looks good to me!

PR #4688, opened 2024-08-06
Remove ctc intake conditions from outgoing email mailer

jenny-heath:
## Link to pivotal/JIRA issue
- need to make one
## What was done?
- went looking for places where CTC is referenced but will definitely not be used anymore
- i think this shouldn't affect anything currently in production but would love a second opinion
## How to test?
- mainly want to know that nothing breaks


jenny-heath:
meh

PR #4692, opened 2024-08-06
Fyst 299 add nc to codebase

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-299
## Is PM acceptance required?
## What was done?
- Added mostly-empty implementations of all basic files for North Carolina
- NC's question flow does not work at this point - I could alternatively leave out most pages from the question navigation but we'd just be adding them back in later.
## How to test?
- Minimal new testing is added here
- Nothing outside FYST should be affected


app/lib/submission_builder/state_return.rb
```diff
@@ -0,0 +1,96 @@
...
+      supported_documents.map { |item| OpenStruct.new(**item, kwargs: item[:kwargs] || {}) if item[:include] }.compact
+    end
+
+    def combined_w2s
+      @submission.data_source.direct_file_data.w2s.map.with_index do |w2, i|
+        intake = @submission.data_source
+        intake_w2 = intake.state_file_w2s.find { |w2| w2.w2_index == i } if intake.state_file_w2s.present?
+
+        {
+          xml: SubmissionBuilder::Shared::ReturnW2,
```
mpidcock, line 64: Looks like `ReturnW2` is the only class in `Shared` that isn't CTC related. Do you think it would make sense to put `StateReturn` et al into shared, or pull `ReturnW2` out? Either way I'd like them to be in the same directory.

app/services/state_file/state_information_service.rb
```diff
@@ -39,12 +40,16 @@ def state_intake_class_names
...
+        STATES_INFO.map { |_, attrs| attrs[:schema_file_name] }
+      end
+
       def state_code_to_name_map
         active_state_codes.to_h { |state_code, _| [state_code, state_name(state_code)] }
       end
     end
 
-    private
+    # private
```
mpidcock, line 52: should this be removed?

mpidcock (approved):
I have some small comments but lgtm! 🥳

app/lib/efile/nc/d400_calculator.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      def initialize(year:, intake:, include_source: false)
+        @year = year
+        @intake = intake
+        @filing_status = intake.filing_status.to_sym
+        @direct_file_data = intake.direct_file_data
+        # intake.state_file_w2s.each do |w2|
+        #   dest_w2 = @direct_file_data.w2s[w2.w2_index]
+        #   dest_w2.node.at("W2StateLocalTaxGrp").inner_html = w2.state_tax_group_xml_node
+        # end
```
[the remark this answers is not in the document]
mrotondo: actually, it looks like since the warshington PR, every uncommented line in this method got moved into the superclass constructor, so i deleted this whole method

app/lib/submission_builder/state_return.rb
mrotondo: I moved everything in submission_builder/shared/ up one level to submission_builder/

PR #4694, opened 2024-08-07
fix xml return sample service to handle invalid keys gracefully

mrotondo:
The XML Return Sample Service was crashing when asked whether it included a key not in the correct format, which actually happens every time we get an API response on a non-prod server. Whoops!


mrotondo:
I also changed the instance members to not be prefixed with _, since that is unnecessary as they are private by default.

mpidcock (approved):

PR #4700, opened 2024-08-14
Fyst 186/create charitable contributions page

powersurge360:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-186

## Sequence that the PR adds

```mermaid
sequenceDiagram
    AZ321Index ->> AZ321New: User has no contributions
    AZ321New -->> AZ321New: Contribution invalid
    AZ321New -->> AZ321Index: Contribution valid
    AZ321Index -->> AZ321Edit: Edit a contribution
    AZ321Edit -->> AZ321Index: User submits valid contribution
    AZ321Edit -->> AZ321Edit: Contribution invalid
    AZ321Edit -->> NextPath: User marks 'no contribution'
    AZ321Index ->> NextPath: User proceeds in the flow
    AZ321New -->> NextPath: User marks 'no contribution'
```

## What was done?
- Adds an AZ 321 Qualified Organization Charitable Contribution page
- Adds a new date accessible concern which can be used in conjunction with `cga_date_select` to write regular validation rules
  - I'd like special attention paid to this as it is metaprogramming, which can incur a steep cost if left unchecked. I don't think this bit is particularly egregious but still, cleverness should be met with skepticism.

### Automated Tests

- Controller specs to test flow
- Model tests to test the unique form_create context for this model
- Some date specific testing in the model as well
- Date Accessible meta programming test added

### Manual testing

- Progress the intake as usual through the AZ flow, continuing to `/en/questions/az-qualifying-organization-contributions/`
- Page should automatically redirect to `/en/questions/az-qualifying-organization-contributions/new` on first visit
- Click continue
- Note that the page does not proceed, instead showing a validation error
- Check 'no' and click continue again
- Note that the page progresses to the next step
- Click 'go back'
- Click 'yes'
- Note form opens
- Click 'continue'
- Note that all fields are considered invalid
- Fill out text values for all fields and set the date
- Note that the date is only for the current tax year
- Click continue
- Note that the amount fails because the value is non-numeric
- Create a numeric value
- Click 'continue'
- Note that the page is now the index page and your contribution has been saved.
- Click edit
- Note that the form displays the values just entered, the form is fully visible, and 'yes' is checked
- Change a value
- Click continue
- Note that the changes are reflected on the page
- Click 'Delete' and note the confirmation window
- Accept
- Note the flash marking 'contribution deleted'
- Proceed to fill in three more contributions
- Note that the 'add another' button is gone (due to there being a limit of three)

## Screenshots

image

image


jenny-heath:
could you add some screenshots of the pages added?

powersurge360:
🫡

Screenshots added as requested.

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        @contribution = contributions.build
+      end
+
+      def create
+        @contribution = contributions.build(az321_contribution_params)
+
+        return redirect_to next_path if @contribution.made_contributions == "no"
+
+        if @contribution.save(context: :form_create)
+          redirect_to action: :index, return_to_review: params[:return_to_review]
```
jenny-heath, line 37: `redirect_to action: :index` is the same as `redirect_to next_path`, no? also, are we intentionally not carrying over the `:return_to_review` param on line 34? i would think we'd want it passed through either way but maybe i'm missing something

app/models/az321_contribution.rb
```diff
@@ -16,5 +16,20 @@
...
+  attr_accessor :made_contributions
+  date_accessor :date_of_contribution
+
   belongs_to :state_file_az_intake
+
+  # Virtual attribute, not in database. Only checked when created via form interface.
+  validates :made_contributions, presence: true, on: :form_create
+
+  validates :charity_name, presence: true
+  validates :charity_code, presence: true
```
jenny-heath, line 28: ok i think i messed up by making this a string because looking at the list of QCOs it seems like it has to be a 5 digit number https://azdor.gov/sites/default/files/2023-05/CREDITS_2023_qco.pdf. we could either change the db or just add the validation step?

app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
```diff
@@ -0,0 +1,71 @@
...
+            <%= f.cfa_input_field(:amount, t('.amount', filing_year: @filing_year), classes: ["form-width--long"]) %>
+            <div class="date-select">
+              <%= f.cfa_date_select(
+                :date_of_contribution,
+                t('.date_of_contribution', filing_year: @filing_year),
+                options: {
+                  start_year: @filing_year,
+                  end_year: @filing_year
+                }
+              ) %>
```
jenny-heath, line 63: minor thing: should we have it default to the filing year so people don't have to click on the one option?

app/views/state_file/questions/az_qualifying_organization_contributions/index.html.erb
```diff
@@ -0,0 +1,53 @@
...
+              class: "button--small button--inline-action last"
+            %>
+          </div>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
+
+  <%= link_to(next_path, class: "button button--primary button--wide spacing-below-10") do %>
+    <%= 'Continue' %>
```
jenny-heath, line 46:
```suggestion
    <%= t("general.continue") %>
```

db/schema.rb
```diff
@@ -11,6 +11,8 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.1].define(version: 2024_08_07_224704) do
+  create_schema "analytics"
+
```
jenny-heath, line 15: huh, i don't understand where this came from?

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+  end
+
+  describe "edit" do
+    let(:contribution) { create(:az321_contribution, state_file_az_intake: intake)}
+
+    it 'should display a form' do
+      get :edit, params: { id: contribution.id }
+
+      expect(response).to be_ok
+      expect(response.body).to include('<form')
```
jenny-heath, line 54: lol, is this a normal test? never seen anything like this before

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+
+      contribution.reload
+
+      expect(contribution.amount.to_i).to be(55)
+    end
+
+    it 'should not update the contribution when invalid' do
+      post :update, params: { id: contribution.id, az321_contribution: { amount: 0 } }
+
+      expect(response).not_to redirect_to(:az_qualifying_organization_contributions)
```
jenny-heath, line 92: should we test that it renders the view again or is that overkill?

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+               date_of_contribution_month: 6,
+               date_of_contribution_day: 16
+             }
+           }
+
+      expect(response).to redirect_to(:az_qualifying_organization_contributions)
+
+      contribution.reload
+
+      expect(contribution.date_of_contribution.to_s).to eql("#{Rails.configuration.statefile_current_tax_year}-06-16")
```
jenny-heath, line 113: oh wow! just had to google `eq` vs `eql`. would this not work with just `eq`? they're both just strings right?

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+          date_of_contribution_day: 12,
+          date_of_contribution_year: Rails.configuration.statefile_current_tax_year,
+          charity_name: "foo",
+          charity_code: "bar",
+          amount: 0
+        }
+      }
+    end
+
+    it 'should create the contribution when the made_contributions missing' do
```
jenny-heath, line 167: er, should this say "is not missing"?

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -245,6 +245,19 @@
       expect(page).to have_text I18n.t('state_file.questions.az_public_school_contributions.index.lets_review')
       click_on I18n.t("general.continue")
 
+      expect(page).to have_text I18n.t('state_file.questions.az_qualifying_organization_contributions.form.main_heading', filing_year: Rails.configuration.statefile_current_tax_year)
+      choose I18n.t("general.affirmative")
+      fill_in "az321_contribution_charity_name", with: "Center for Ants"
```
jenny-heath, line 250: did you see the factory or did we just happen to come up with the same charity name

spec/models/az321_contribution_spec.rb
jenny-heath, line 1: might be good to add a couple of validation tests for the code/name being present?

app/models/az321_contribution.rb
powersurge360: Sure, could do both.

app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
powersurge360: Very good idea, will do.

db/schema.rb
powersurge360: When I merged main it came with that. I can re-remove it.

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
powersurge360: I started writing these as request specs before I realized that we hadn't really begun that migration yet. Since include checks for a string literal I thought I'd just check for the start of a form tag. It _is_ a little wonky though and I can remove the test if you prefer.

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
powersurge360: Thanks for the catch

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
powersurge360: Will change

spec/models/az321_contribution_spec.rb
powersurge360: I was reluctant to test validators that were so plain because it's effectively code duplication and just opted for the bits. I saw we had Thoughtbot's `shoulda_matchers` but even with a nice syntax it feels weird to me. If you feel strongly I can plop some in though.

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
jenny-heath: ok! i think i don't totally understand why it changes but i believe you. any thoughts on the `:return_to_review` param?

app/models/az321_contribution.rb
jenny-heath: ok! just to clarify: did you mean you'll do that as part of this PR or later? i'd be fine with just the validation tbh and it might even be good to ask tiffany or program if that is the correct validation

db/schema.rb
jenny-heath: hm yeah if it's on main it shouldn't show up as a change so i'd just make sure your schema.rb matches main since you didn't do any db stuff as part of this PR

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
jenny-heath: yeah i think the next block in here (that checks for the contribution data on the page) should cover it

spec/models/az321_contribution_spec.rb
jenny-heath: just want to clarify what you mean by code duplication? is it that we don't want to test the the validators work or that there's already a spec that tests these validations? if the former, in my mind the purpose of the test would be to check that the model includes the right validations, not that the rails built-in validators work as expected. i wouldn't want it to be the case that these presence validations disappear and the tests all still pass. lmk if that makes sense and happy to chat about it on tuple if not!

jenny-heath:
not a requirement for this PR, but i'd love to see how the DateAccessible stuff works with one of our form models so if you'd be interested in doing a follow-up task to refactor one of those, that would be great

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
powersurge360: The way the flow currently works is something like:

```mermaid
sequenceDiagram
    AZ Review Page->>Form Index: User clicks to review
    Form Index->>Form Edit: User clicks 'Add another' or 'Edit'
    Form Edit->>Form Index: User saves
    Form Index->>AZ Review Page: User clicks continue, using next_path underneath
```

app/models/az321_contribution.rb
powersurge360: I was planning on doing this as part of the PR but I can make a note to talk to Tiffany while I work on other changes and if she doesn't get back to me before I'm finished I'll punt and make a new ticket to follow up.

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
powersurge360: I'm not inclined to do so but if you feel strongly I can. Add another comment if ya do otherwise I'll resolve this after I finish the other stuff.

spec/models/az321_contribution_spec.rb
powersurge360: I can add it! The point I was making is that the validation is so simple that you're not really testing the functionality of the validation per se, you're just testing that it's there at all. And that's pretty transparently 1:1 to the method being there.

Buuut the other argument would be if it was worth coding it was worth testing, even if it is dead simple. If we're using shoulda_matchers already it's pretty dead simple so nbd to plop it in.

app/models/az321_contribution.rb
powersurge360: Tiffany got back to me and updated the story

image

Not sure atm if I'm going to do integer w/ inclusion or something with string. If the id had leading zeros it'd be easy to do a string but if it's always between these values it might make sense to migrate the column. Will think on it.

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
jenny-heath: hm, ok, the scenario i think is possible here:
1. get to the index from the review page with review param
2. click to add a new contribution
3. select “no” they didn’t make a contribution and click continue to get back to the index, losing review param (line 34)
4. click continue on index
5. instead of going back to the review page, they will end up on whatever the next intake question is

i might be wrong though. probably best to just test it out manually because thinking about it is confusing me

app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
powersurge360: Set the filing year. This will effectively set the default to `1/1/#{filing_year}` though. I think that's probably fine. I'll let you decide if it satisfies your request and mark as resolved.

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
powersurge360: Yea, that case does seem not to work. Also, it looks like I neglected to add a panel to the AzReviewController, so will work on that together with this fix and will come back for another review.

spec/support/helpers/state_file_intake_helper.rb
```diff
@@ -95,7 +95,7 @@ def assert_flow_explorer_sample_params_includes_everything(us_state)
 
     intake = StateFile::StateInformationService.intake_class(us_state).last
     flow_explorer_params = FlowsController::SampleStateFileIntakeGenerator.send("#{us_state}_attributes")
-    intake_attribute_keys = intake.attributes.select { |_k, v| v }.keys - %w[id primary_state_id_id]
+    intake_attribute_keys = intake.attributes.select { |_k, v| v }.keys - %w[id primary_state_id_id made_az321_contributions]
```
powersurge360, line 98: I don't like this but I'm not sure the alternative. This popped up because I was doing a `nested_attributes_for` for the intake, I think.

powersurge360:
I had a brief meeting with Anu today and she showed me the [Figma mockups](https://www.figma.com/design/p2ZmiVl9TUJGwfAIeH4Lfs/2025-AZ-FYST?node-id=7680-7196&t=FTZxoxvCr6lcl09l-0) in progress, which seem like they may have some large changes to how this form will proceed. To that end, I think maybe we should review in the sense that stuff is working and not worry about polish until the next phase.

PR #4703, opened 2024-08-16
Make state file FAQ page show only the previous active season's states until the start of next season

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-442
## What was done?
- The story suggests doing a feature flag but I think this will be a recurring issue if the product exists for multiple years so I added a property to the state information service to track which years states are/were active. I think this can probably be used elsewhere too.
- The code in the controller index is slightly confusing imo but i couldn't think of a better way to handle the cases. I guess we could assume that the start date will always be on Jan 1 and in that case we'd just use the current tax year every time. Open to thoughts!
## How to test?
- I added a controller spec that tests essentially 2 cases: 1) when it's before the start of the season in the year *of* the season and 2) when it's past the start date. I couldn't think of any other cases that seemed important but I might be missing something!
- Manual testing: go to the FAQ page and see that NC disappears
  - I'd love to be able to test the case where NC is open but the config doesn't have next year's start date so I was only able to test it locally by changing the development config. It's kind of the same test though.
- Before
image
- After
image


app/services/state_file/state_information_service.rb
```diff
@@ -4,13 +4,14 @@ class << self
       [
         :intake_class,
         :calculator_class,
-        :navigation_class,
-        :submission_builder_class,
```
jenny-heath, line 8: alphabetized

app/controllers/state_file/faq_controller.rb
```diff
@@ -2,10 +2,18 @@ class StateFile::FaqController < ApplicationController
...
 
   def index
+    intake_start = Rails.configuration.state_file_start_of_open_intake
+    active_tax_year = if (app_time.year == intake_start.year) && app_time < intake_start
+                        # if the start of intake is set to this year and it is currently before that date, we should still show the previous year's states
+                        Rails.configuration.statefile_current_tax_year - 1
+                      else
+                        Rails.configuration.statefile_current_tax_year
+                      end
+    visible_state_code_names = StateFile::StateInformationService.state_code_to_name_map.filter { |code, name| StateFile::StateInformationService.filing_years(code).include?(active_tax_year) }
```
jnf, line 12: since `name` is unused could you use `active_state_codes` directly instead of `state_code_to_name_map`? Minor change, but one fewer loop/data transformation every time the page is hit.

jnf (approved):
i dig the `filing_years` approach. minimally invasive, easy to reason about. left 1 question about a suuuuper minor optimization, but i wouldn't call it blocking.

app/controllers/state_file/faq_controller.rb
jenny-heath: name is used! the original controller action also uses `state_code_to_name_map` and in the view it displays the state name as part of the title of the FAQ section

PR #4704, opened 2024-08-16
memoize state sugar methods

jnf:
## Link to pivotal/JIRA issue
- none

## What was done?
`STATES_INFO` is the foundation of the `StateInformationService`. We have a handful of sugar/convenience methods in that class that all derive structures from `STATES_INFO`. This PR does two things:

1. We weren't truly freezing the definition of `STATES_INFO`. Calling `#with_indifferent_access` on the hash _after_ it was frozen was returning a new, unfrozen object. This moves that call onto the hash literal prior to freezing.
2. Adds new instance variables to hold a memoized output of calls to a few of the convenience methods. All the changed methods munge data from `STATES_INFO` into different representations without relying on passed parameters. That is to say, that after freezing `STATES_INFO`, calling any of these methods should be idempotent. Memoizing saves a bit a runtime complexity.

## Question to the team (answered)
Given these convenience methods are defined on the class itself, are class variables more appropriate for memoizing? We don't ever instantiate `StateInformationService`, as best I can tell, so it's not really a singleton, but we're kind of treating it like one. So what's the right approach here for memoization -- instance or class variable?

**Answer:** Ruby classes are effectively singletons, especially given how we use them. Instance variable or class variable isn't really that important in this context. 

## How to test?
- No functional change; minor optimizations/refactor. As long as CI is green we should be good.

n/a

### DIFF
--- a/app/services/state_file/state_information_service.rb
+++ b/app/services/state_file/state_information_service.rb
@@ -113,6 +113,6 @@ module StateFile
         voucher_form_name: "Form IT-201-V",
         voucher_path: "/pdfs/it201v_1223.pdf",
       }
-    }).with_indifferent_access
+    }.with_indifferent_access)
   end
 end
--- a/app/services/state_file/state_information_service.rb
+++ b/app/services/state_file/state_information_service.rb
@@ -29,11 +29,11 @@ module StateFile
       end
 
       def active_state_codes
-        @_active_state_codes ||= STATES_INFO.keys.map(&:to_s)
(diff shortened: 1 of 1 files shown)


mrotondo (approved):
looks good, nice catch on the deep_freeze parens. 

If you could add .freeze to the end of the method calls in the convenience methods, I think that would really tie the room together.

mrotondo:
IMO ruby classes effectively are singletons already - if you want, you could implement `StateInformationService#initialize` to raise NotImplementedError

PR #4705, opened 2024-08-19
Fyst 182 Retirement income page

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-182
## Is PM acceptance required?
## What was done?
- Adds a page to ask about retirement income
- Page is shown if fed xml indicates value for total pension amount and presence of 1099R(s)
- Checkboxes reveal inputs for amounts
- Page shows additional checkbox for spouse pensions
## How to test?
- Unit tests on controller check: show logic, spouse checkbox presence, return to review functionality
- Unit tests on form check: validations, saving values
- Section added to feature spec and corresponding fixture
- Manual test cases:
  - With fed return that has retirement income (page shows)
  - Without fed return that has retirement income (page does not show)
  - Mfj filers (spouse checkbox present)
  - Non-mfj filer (spouse checkbox absent)
  - Can't enter non-numerical value
  - Can't check box and not enter value
  - Can't enter values that add up to total over TotalTaxablePensionsAmt
- image

### DIFF
--- a/app/forms/state_file/az_retirement_income_form.rb
+++ b/app/forms/state_file/az_retirement_income_form.rb
@@ -1,0 +1,16 @@
+module StateFile
+  class AzRetirementIncomeForm < QuestionsForm
+    set_attributes_for :intake,
+                       :received_military_retirement_payment,
+                       :received_military_retirement_payment_amount,
+                       :primary_received_pension,
+                       :primary_received_pension_amount,
+                       :spouse_received_pension,
+                       :spouse_received_pension_amount
+
+    validates :received_military_retirement_payment_amount, presence: true, allow_blank: false, if: -> { received_military_retirement_payment == "yes" }
+    validates_numericality_of :received_military_retirement_payment_amount, if: -> { received_military_retirement_payment == "yes" }
+    validates_numericality_of :primary_received_pension_amount, if: -> { primary_received_pension == "yes" }
+    validates_numericality_of :spouse_received_pension_amount, if: -> { spouse_received_pension == "yes" }
+  end
+end
--- a/app/forms/state_file/az_retirement_income_form.rb
+++ b/app/forms/state_file/az_retirement_income_form.rb
@@ -8,9 +8,26 @@ module StateFile
                        :spouse_received_pension,
                        :spouse_received_pension_amount
 
-    validates :received_military_retirement_payment_amount, presence: true, allow_blank: false, if: -> { received_military_retirement_payment == "yes" }
-    validates_numericality_of :received_military_retirement_payment_amount, if: -> { received_military_retirement_payment == "yes" }
-    validates_numericality_of :primary_received_pension_amount, if: -> { primary_received_pension == "yes" }
-    validates_numericality_of :spouse_received_pension_amount, if: -> { spouse_received_pension == "yes" }
+    validates :received_military_retirement_payment_amount, presence: true, numericality: true, allow_blank: false, if: -> { received_military_retirement_payment == "yes" }
+    validates :primary_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { primary_received_pension == "yes" }
+    validates :spouse_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { spouse_received_pension == "yes" }
+    validate :below_1040_amount, if: -> { [received_military_retirement_payment_amount, primary_received_pension_amount, spouse_received_pension_amount].any?(&:present?) }
+
+    def save
+      @intake.update(attributes_for(:intake))
+    end
+
+    private
+
+    def below_1040_amount
+      amount_limit = @intake.direct_file_data.fed_taxable_pensions
+      total = received_military_retirement_payment_amount.to_i + primary_received_pension_amount.to_i + spouse_received_pension_amount.to_i
+      if total > amount_limit
+        errors.add(:base, I18n.t("forms.errors.state_credit.exceeds_limit", limit: amount_limit))
+        errors.add(:received_military_retirement_payment_amount, "")
+        errors.add(:primary_received_pension_amount, "")
+        errors.add(:spouse_received_pension_amount, "")
+      end
+    end
   end
 end
--- a/app/forms/state_file/az_retirement_income_form.rb
+++ b/app/forms/state_file/az_retirement_income_form.rb
@@ -11,7 +11,7 @@ module StateFile
     validates :received_military_retirement_payment_amount, presence: true, numericality: true, allow_blank: false, if: -> { received_military_retirement_payment == "yes" }
     validates :primary_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { primary_received_pension == "yes" }
     validates :spouse_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { spouse_received_pension == "yes" }
-    validate :below_1040_amount, if: -> { [received_military_retirement_payment_amount, primary_received_pension_amount, spouse_received_pension_amount].any?(&:present?) }
+    validate :below_1040_amount
 
     def save
       @intake.update(attributes_for(:intake))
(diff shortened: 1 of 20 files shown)


app/forms/state_file/az_retirement_income_form.rb
```diff
@@ -0,0 +1,33 @@
...
+                       :received_military_retirement_payment_amount,
+                       :primary_received_pension,
+                       :primary_received_pension_amount,
+                       :spouse_received_pension,
+                       :spouse_received_pension_amount
+
+    validates :received_military_retirement_payment_amount, presence: true, numericality: true, allow_blank: false, if: -> { received_military_retirement_payment == "yes" }
+    validates :primary_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { primary_received_pension == "yes" }
+    validates :spouse_received_pension_amount, presence: true, numericality: true, allow_blank: false, if: -> { spouse_received_pension == "yes" }
+    validate :below_1040_amount, if: -> { [received_military_retirement_payment_amount, primary_received_pension_amount, spouse_received_pension_amount].any?(&:present?) }
```
powersurge360, line 14: NIT: Consider having a `total_pension_amount` reader which does

```ruby
def total_pension_amount
  received_military_retirement_payment_amount&.to_i + primary_received_pension_amount&.to_i + spouse_received_pension_amount&.to_i
end
```

and then check `total_pension_amount > 0`. This will make the validation a little easier to read and enables a tidier `if` check on line 25.

app/forms/state_file/az_retirement_income_form.rb
```diff
@@ -0,0 +1,33 @@
+module StateFile
+  class AzRetirementIncomeForm < QuestionsForm
+    set_attributes_for :intake,
+                       :received_military_retirement_payment,
+                       :received_military_retirement_payment_amount,
+                       :primary_received_pension,
+                       :primary_received_pension_amount,
+                       :spouse_received_pension,
+                       :spouse_received_pension_amount
```
powersurge360, line 9: I get that this is the standard practice atm but I don't get why we don't just use `delegate` instead;

```ruby
delegate :intake,
         :received_military_retirement_payment,
         :received_military_retirement_payment_amount,
         :primary_received_pension,
         :primary_received_pension_amount,
         :spouse_received_pension,
         :spouse_received_pension_amount
         to: :intake
```
 I guess we would need to delegate the writers as well as the readers but imo being a little closer to standard rails might be worth it. The code around `set_attributes_for` seems to be doing a lot of work, especially around separating and reconnecting attributes to particular models that I feel like we can shed with this approach. Not that I'm advising a refactor atm, just thinking about what we use `set_attributes_for` out loud.

app/forms/state_file/az_retirement_income_form.rb
```diff
@@ -0,0 +1,33 @@
...
+    validate :below_1040_amount, if: -> { [received_military_retirement_payment_amount, primary_received_pension_amount, spouse_received_pension_amount].any?(&:present?) }
+
+    def save
+      @intake.update(attributes_for(:intake))
+    end
+
+    private
+
+    def below_1040_amount
+      amount_limit = @intake.direct_file_data.fed_taxable_pensions
```
powersurge360, line 23: This is interesting to me, because I feel like this is probably something that doesn't change per year. Is it required to be loaded at validation time?

spec/forms/state_file/az_retirement_income_form_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+      end
+
+      it "is invalid and attaches the correct error" do
+        allow(intake).to receive(:filing_status_mfj?).and_return true
+
+        expect(form).not_to be_valid
+        expect(form.errors[:received_military_retirement_payment_amount]).to include "is not a number"
+        expect(form.errors[:primary_received_pension_amount]).to include "is not a number"
+        expect(form.errors[:spouse_received_pension_amount]).to include "is not a number"
+      end
```
powersurge360, line 30: Consider replacing with [shoulda_matchers](https://github.com/thoughtbot/shoulda-matchers?tab=readme-ov-file#matchers).

```ruby
it { should validate_numericality_of(:received_military_payment_amount) }
it { should validate_numericality_of(:primary_received_pension_amount) }
it { should validate_numericality_of(:spouse_received_pension_amount) }
```

powersurge360 (commented):
Couple of notes, couple of questions, couple of nitpicks, and I wanted to point out the `shoulda_matchers` that are already in the gemfile. All of the specs that are testing plain ActiveRecord validation can be collapsed into shoulda_matchers, if you so desire.

I didn't mark as request changes because I don't want to get in the way as a rookie on the team but I also didn't approve because the stuff around forms is still moderately confusing to me. We can talk about it later today maybe and I can more confidently give an approval.

app/forms/state_file/az_retirement_income_form.rb
jenny-heath: hm yeah could be a computer gardening topic! (you could do a spike and present your findings) `QuestionsForm` was lifted from another cfa project when we spun up the codebase and i haven't honestly given it much thought ever

app/forms/state_file/az_retirement_income_form.rb
jenny-heath: i think it's the amount the client received in pensions that year? so i would expect it to change per year. i'm not sure what you mean by "loaded" at validation time? is there a way to pre-load it?

app/forms/state_file/az_retirement_income_form.rb
jenny-heath: question about that! part of having the `below_1040_amount` method is that it attaches 4 different errors: 3 blank ones so that the red line appears next to the offending inputs, 1 with a message that shows at the bottom because it's about the sum of all 3. how would i assign those errors to the right attributes with a one-line validation?

jenny-heath: actually i've just realized i don't need to have this `if` check on line 14 because if the amounts are nil they'll be converted to 0 and the sum when all are absent will always be less than fed_taxable_pensions (because this page only shows if fed_taxable_pensions > 0)

app/models/direct_file_data.rb
```diff
@@ -253,6 +254,14 @@ def fed_taxable_income=(value)
     write_df_xml_value(__method__, value)
   end
 
+  def fed_taxable_pensions
+    df_xml_value(__method__)&.to_i || 0
```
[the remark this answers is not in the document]
jenny-heath: oh actually i think `nil&.to_i` is `nil` and `nil.to_i` is 0 lol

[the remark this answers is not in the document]
jenny-heath: ^ so could simplify to `df_xml_value(__method__).to_i` BUT i think i'm going to leave it as is for now because that exact line is duplicated 34 times across this file 🤦🏻‍♀️ so when (🤞) i get around to a refactor it'll be easier to find and replace

spec/forms/state_file/az_retirement_income_form_spec.rb
jenny-heath: i've never used them! ty!!

jenny-heath: update on this: turns out using the gem for our form objects isn't that straightforward. i tried it out for a while and slacked kevin about it but will consider it out of scope for this feature

app/forms/state_file/az_retirement_income_form.rb
powersurge360: It currently happens at the moment you call `valid?` or `save?` and looks up the value. If it didn’t happen there it could happen instead on class load. But it’s somewhat moot because of the errors being established in this method. Mostly I was thinking whether this case is special enough to be broken out or if there’s a way we could do it with regular validators. I think setting the error message on each makes it special enough though.

spec/forms/state_file/az_retirement_income_form_spec.rb
powersurge360: Sounds good. Probably something to do with forms being not-quite-models

powersurge360 (approved):
Looks good to me

PR #4706, opened 2024-08-20
NC intake: add intercom svg

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-440
## Is PM acceptance required?
## What was done?
- The landing page for NC was breaking when you set the session toggle to during the tax season because it looks for an intercom svg with the state abbreviation in it and that was missing for NC
- Maybe out of scope for this ticket but flow breaks at the name-dob page because the NC intake is missing spouse birth date
## How to test?
- Manual test:
  - Use session toggles to set time to before intake
  - Click "start intake NC" and see the landing page (not an error)
- Automated tests
  - added a feature spec
image


mrotondo (approved):
I'd love for us to not have to do this for every state. If you are interested, you could 
1. inline the svg (just move the svg from the file into the place where it's currently used in an image tag)
2. (I think) switch out the fill colors to "currentColor" (like so https://echobind.com/post/currentcolor-css-property-with-svg) 
3. set the color of the css class to the current state color

But if you don't want to do all that, then this is fine!

mrotondo (approved):
yay!

PR #4707, opened 2024-08-20
Add 1099r pdf, xml and tax calculations

embarnard:
## Link to pivotal/JIRA issue
- e.g. [link](https://codeforamerica.atlassian.net/browse/FYST-235)
## What was done?
- Added fields on the AZ140 XML submission builder for relevant 1099r fields
- Added calculations
- Added fields for filling PDF on 1099 fields
## How to test?
- Go through the Arizona flow and pick DF test cases Richard 1099 or Alexis HOH in Heroku
- Fill in amounts for receiving pension and military retirement payment 
- Validate that the XML and the PDF look correct upon submission


app/lib/efile/az/az140_calculator.rb
```diff
@@ -19,6 +19,8 @@ def calculate
         set_line(:AZ140_LINE_12, @direct_file_data, :fed_agi)
         set_line(:AZ140_LINE_14, :calculate_line_14)
         set_line(:AZ140_LINE_19, :calculate_line_19)
+        set_line(:AZ140_LINE_29A, :calculate_line_29A)
+        set_line(:AZ140_LINE_29B, :calculate_line_29B) # what happens if not present
```
mpidcock, line 23: is this a todo? can you clarify in the comment what is not present?

app/lib/efile/az/az140_calculator.rb
```diff
@@ -208,7 +230,7 @@ def calculate_line_52
       end
 
       def calculate_line_53
-        # zzz
+        # state taxes withheld: sum of W-2, 1099-R, 1099-G
```
mpidcock, line 233: should 1099R be added in this calculation below?

app/lib/efile/az/az140_calculator.rb
```diff
@@ -77,6 +79,25 @@ def calculate_line_19
         line_or_zero(:AZ140_LINE_14)
       end
 
+      def calculate_line_29A
+        # total subtraction amount for pensions up to the maximum of $2,500 each for primary and spouse
+        return 0 unless @intake.primary_received_pension_yes? || @intake.spouse_received_pension_yes?
+
+        primary_pension_amt = @intake.primary_received_pension_yes? && @intake.primary_received_pension_amount.present? ? [@intake.primary_received_pension_amount&.round, 2_500].min : 0
```
mpidcock, line 86: we don't need the ampersand here or on # 91, right? because we just checked to see if it was present, and we can't remove the .present? check because then the .min function might break. 

there's a lot happening on this line so it would be my preference to break it out into an if/else, or have two lines (one to get the pension amt and one to get the subtraction amt). 

a tiny case could also be made to also extract this to a method to be used for both primary and spouse, which normally I wouldn't bother with, but it would mean keeping the 2_500 value in one place. up to you though

spec/lib/submission_builder/ty2022/states/az/az_return_xml_spec.rb
```diff
@@ -141,5 +141,28 @@
...
       end
     end
+
+    context "when 1099-rs are present" do
+      let(:intake) do
+        create(:state_file_az_intake,
+               raw_direct_file_data: StateFile::XmlReturnSampleService.new.read('az_richard_retirement_1099r'),
+               filing_status: "married_filing_jointly",
+               primary_received_pension: "yes",
+               primary_received_pension_amount: 2000.6,
```
mpidcock, line 151: if it's possible to have `primary_received_pension`=yes but `primary_received_pension_amount` be nil, we should have a test that covers that

app/lib/efile/az/az140_calculator.rb
mpidcock: after chatting with tiffany and courtney I learned that there are some more 1099R fields we should add, including the state taxes withheld, so I'm going to add that into this PR ([slack thread here](https://cfastaff.slack.com/archives/C07GD2NS086/p1724789015247049))

app/lib/efile/az/az140_calculator.rb
```diff
@@ -96,7 +115,8 @@ def calculate_line_32
       end
 
       def calculate_line_35
-        subtractions = 0
+        # Subtotal after additions and subtractions
+        subtractions = line_or_zero("AZ140_LINE_29A") + line_or_zero("AZ140_LINE_29B")
```
mrotondo, line 119: no change necessary, but i think these are more often passed around as symbols than as strings

app/lib/efile/az/az140_calculator.rb
```diff
@@ -206,7 +226,7 @@ def calculate_line_50
           wrksht_2_line_5 = 120
         end
 
-        #wrksheet 2
+        # wrksheet 2
```
mrotondo, line 229:
```suggestion
        # worksheet 2
```

app/lib/efile/ny/it201.rb
```diff
@@ -326,7 +326,7 @@ def calculate_line_69a
       end
 
       def calculate_line_72
-        total_state_taxes_withheld = @direct_file_data.total_state_tax_withheld
+        total_state_taxes_withheld = @direct_file_data.total_w2_state_tax_withheld
         state_file_1099gs = @intake.state_file1099_gs
```
mrotondo, line 330: Should this get refactored the same way `Az140Calculator#calculate_line_53` did (to use `@intake.state_file1099_gs.sum(&:state_income_tax_withheld)`? And should we add `@direct_file_data.total_1099r_state_tax_withheld` here too, as was done there? Not sure what our strategy is re: working on NY code at this point.

app/models/direct_file_data.rb
```diff
@@ -317,22 +317,25 @@ def fed_total_adjustments=(value)
...
+    selector = DfW2Accessor.selectors[:StateIncomeTaxAmt]
+    w2_nodes.sum do |w2|
+      w2.at(selector)&.text.to_i
+    end
+  end
+
+  def total_w2_local_tax_withheld
+    selector = DfW2Accessor.selectors[:LocalIncomeTaxAmt]
+    w2_nodes.sum do |w2|
+      w2.at(selector)&.text.to_i
```
mrotondo, line 337: Sorry for not catching this while we were pairing, but if you want to get rid of the direct access on `selectors`, you could replace this loop with (I think) `w2s.sum(&:LocalIncomeTaxAmt)`

mrotondo: And similarly for the other methods just above this one.

mrotondo:
I think it would be good to add the `disqualifying_df_data_reason` changes we discussed to this PR as well

app/lib/efile/az/az140_calculator.rb
mpidcock: Mike and I paired to refactor this line, happy to discuss the changes if you're interested

spec/lib/submission_builder/ty2022/states/az/az_return_xml_spec.rb
mpidcock: it really shouldn't be possible though. Mike and I discussed while pairing, and if such a situation happens, we decided we would prefer a loud error to happen rather than silently handling it

app/lib/efile/ny/it201.rb
mpidcock: I am happy to refactor NY, but I don't feel as comfortable adding 1099R in just this one place without adding it elsewhere too

app/models/direct_file_data.rb
mpidcock: :chefs-kiss:

mpidcock:
Re: disqualifying_df_data_reason, I think we should verify that requirement before putting it in. I've seen some debate about if it is an actual requirement.

app/lib/efile/az/az140_calculator.rb
```diff
@@ -217,17 +237,15 @@ def calculate_line_52
         [line_52_value, 0].max
       end
 
+      # TODO: test combination of multiple forms (w2, 1099G, 1099R) in the spec for this calculator
```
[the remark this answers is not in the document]
mpidcock: oh good catch. it would be great to get it into this PR but my plate is full today and tomorrow

mpidcock (approved):
approving, but it was a joint effort of coding and approving

PR #4708, opened 2024-08-20
GYR1-557 SLA Module count bug fix

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/GYR1-557

## Is PM acceptance required?
- Yes - but this needs to be merged after code review because it will be tested on demo

## What was done?
Before the SLA module was showing clients with inactive returns. The query for the presenter for the dashboard was edited so that it filters for only clients with active returns

## How to test?
- Visit https://demo.getyourrefund.org/en/hub/dashboard/
>     username: demo.hub.test@gmail.com
>     PW: vitavitavitavita
- Select SimplifyCT and observe SLA Overdue number
- Click “View All” 
- Observe client number

### DIFF
--- a/app/presenters/hub/dashboard/service_level_agreements_notifications_presenter.rb
+++ b/app/presenters/hub/dashboard/service_level_agreements_notifications_presenter.rb
@@ -7,7 +7,7 @@ module Hub
       end
 
       def approaching_sla_clients
-        @approaching_sla_clients ||= @clients.select("clients.vita_partner_id, COUNT(clients.id) AS number_of_clients")
+        @approaching_sla_clients ||= @clients.select("clients.vita_partner_id, COUNT(DISTINCT clients.id) AS number_of_clients")
                                              .sla_tracked
                                              .where(vita_partner_id: @selected_orgs_and_sites.map(&:id))
                                              .where(last_outgoing_communication_at: 6.business_days.ago..4.business_days.ago)
@@ -23,7 +23,7 @@ module Hub
       end
 
       def breached_sla_clients
-        @breached_sla_clients ||= @clients.select("clients.vita_partner_id, COUNT(clients.id) AS number_of_clients")
+        @breached_sla_clients ||= @clients.select("clients.vita_partner_id, COUNT(DISTINCT clients.id) AS number_of_clients")
                                              .sla_tracked
                                              .where(vita_partner_id: @selected_orgs_and_sites.map(&:id))
                                              .where("last_outgoing_communication_at < ?", 6.business_days.ago)
--- a/app/presenters/hub/dashboard/service_level_agreements_notifications_presenter.rb
+++ b/app/presenters/hub/dashboard/service_level_agreements_notifications_presenter.rb
@@ -7,11 +7,11 @@ module Hub
       end
 
       def approaching_sla_clients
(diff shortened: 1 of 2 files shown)


app/presenters/hub/dashboard/service_level_agreements_notifications_presenter.rb
```diff
@@ -25,6 +26,7 @@ def breached_sla_clients
         @breached_sla_clients ||= @clients.select("vita_partner_id, COUNT(clients.id) AS number_of_clients")
                                              .where(vita_partner_id: @selected_orgs_and_sites.map(&:id))
                                              .where("last_outgoing_communication_at < ?", 6.business_days.ago)
+                                             .where("filterable_tax_return_properties @> ?::jsonb", [{ active: true }].to_json)
```
powersurge360, line 29: I think this is a cached property. Looking at `self.refresh_filterable_properties` in `search_indexer.rb`

```ruby
  def self.refresh_filterable_properties(client_ids = nil, limit = 1000)
    ActiveRecord::Base.transaction do
      client_ids =
        if client_ids.nil?
          Client.where('needs_to_flush_filterable_properties_set_at < ?', Time.current).limit(limit).pluck(:id)
        else
          Client.where(id: client_ids)
        end

      attributes = Client.where(id: client_ids).includes(:tax_returns, :documents, intake: :dependents).map do |client|
        {
          id: client.id,
          created_at: client.created_at,
          updated_at: client.updated_at,
          filterable_tax_return_properties: client.tax_returns.map do |tr|
            {
              year: tr.year,
              service_type: tr.service_type,
              current_state: tr.current_state,
              assigned_user_id: tr.assigned_user_id,
              stage: TaxReturnStateMachine::STAGES_BY_STATE[tr.current_state],
              active: tr.current_state.present? && !TaxReturnStateMachine::EXCLUDED_FROM_SLA.include?(tr.current_state&.to_sym),
              greetable: TaxReturnStateMachine.available_states_for(role_type: GreeterRole::TYPE).values.flatten.include?(tr.current_state)
            }
          end,
          filterable_product_year: client.intake&.product_year, # we expect clients to always have an intake, it's only in factory-land that this is needed
          filterable_number_of_required_documents_uploaded: client.number_of_required_documents_uploaded,
          filterable_number_of_required_documents: client.number_of_required_documents,
          filterable_percentage_of_required_documents_uploaded: client.number_of_required_documents_uploaded / client.number_of_required_documents.to_f,
          needs_to_flush_filterable_properties_set_at: nil
        }
      end
      return unless attributes.present?

      attributes_to_update = attributes.first.keys - [:id, :created_at, :updated_at]
      Client.upsert_all(attributes, record_timestamps: false, update_only: attributes_to_update)
    end
  end
```

I think you might be able to say `.where.not(current_state: TaxReturnStateMachine::EXCLUDED_FROM_SLA`. Maybe even add it as a scope, like `scope :active, -> { .where.not(current_state: TaxReturnStateMachine::EXCLUDED_FROM_SLA` and then just call `.active` here.

That's my naive take, though. Don't take action on it til someone can confirm that that's the better way to do things.

powersurge360 (approved):
Looks good to me!

PR #4713, opened 2024-08-22
Linux development

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/26
No
## What was done?
Add an optional docker compose setup without disrupting local install
## How to test?
To make sure that local development is not disrupted by this PR, pull down this PR and run local development. Verify that the tests run and the database connection works.

One outstanding question that's marked todo: what should the default username and password be for local install postgres?

To try docker setup, first stop all running servers for this project, as the same ports are used in both. Then, `docker compose up`. Follow the commands added to the readme to run tests in docker.
There should be no visual changes

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,4 +1,4 @@
-# This dockerfile is based on the production dockerfile
+# This dockerfile is based on the production dockerfile, minus GYR Efiler Java deps
 
 ### First build stage used for rails, jobs, and shakapacker containers
 FROM ruby:3.2.2 AS base
@@ -26,15 +26,6 @@ RUN curl -fsSLO "$SUPERCRONIC_URL" \
 
 ADD ./vendor/pdftk /app/vendor/pdftk
 RUN /app/vendor/pdftk/install
-
-# gyr-efiler requires Java 8. Download Java 8 and provide a variable for the Ruby app.
-ENV OPENJDK8_URL=https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u292-b10/OpenJDK8U-jre_x64_linux_hotspot_8u292b10.tar.gz \
-    OPENJDK_SHA1SUM=55848001c21214d30ca1362bace8613ce9733516
-RUN wget -O /tmp/openjdk.tar.gz "$OPENJDK8_URL" \
- && echo "${OPENJDK_SHA1SUM}  /tmp/openjdk.tar.gz" | sha1sum -c - \
- && cd /opt && tar xf /tmp/openjdk.tar.gz \
- && rm -f /tmp/openjdk.tar.gz
-ENV VITA_MIN_JAVA_HOME=/opt/jdk8u292-b10-jre
 
 WORKDIR /app
 ADD package.json yarn.lock /app/
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -17,9 +17,9 @@ RUN apt-get update \
 #  && apt-get -y install imagemagick nodejs \
 #  && rm -rf /var/lib/apt/lists/*
 
-ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.9/supercronic-linux-amd64 \
-    SUPERCRONIC=supercronic-linux-amd64 \
-    SUPERCRONIC_SHA1SUM=5ddf8ea26b56d4a7ff6faecdd8966610d5cb9d85
+# ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.9/supercronic-linux-amd64 \
+#     SUPERCRONIC=supercronic-linux-amd64 \
+#     SUPERCRONIC_SHA1SUM=5ddf8ea26b56d4a7ff6faecdd8966610d5cb9d85
 
 # Is this necessary in dev?
 # RUN curl -fsSLO "$SUPERCRONIC_URL" \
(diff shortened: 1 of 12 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+    image: postgis/postgis:13-3.4-alpine
+    user: postgres
+    ports:
+      - 5432:5432
+    volumes:
+      - database:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: password
+  jobs:
```
mrotondo, line 15: No change necessary, but naming this `worker` would be consistent with our Procfile-based deployment elsewhere (in production on Aptible, and in development via foreman)

docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+    depends_on:
+      rails:
+        condition: service_healthy
+    environment:
+      - DOCKER=true
+      - RAILS_DB_HOST=db
+      - RAILS_DB_USERNAME=postgres
+      - RAILS_DB_PORT=5432
+      - RAILS_DB_PASSWORD=password
+  rails:
```
mrotondo, line 32: Similarly to the above comment, we call the equivalent to this service `web` in our Procfiles - up to you if you feel that it's worth changing here for consistency.

docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+    container_name: rails
+    build:
+      dockerfile: Dockerfile.local
+      target: "app"
+    volumes:
+      - ./:/app
+    ports:
+      - 3000:3000
+      - 9888:9888
```
mrotondo, line 41: Do we need to expose Capybara ports if we always run the test on this service's container?

docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+  shakapacker:
+    container_name: shakapacker
+    build:
+      dockerfile: Dockerfile.local
+      target: "base"
+    volumes:
+      - ./:/app
+    ports:
+      - 3035:3035
+    command: "bin/shakapacker-dev-server --client-web-socket-url auto://0.0.0.0:3035/ws"
```
mrotondo, line 69: Cool to see that this is in and working! Out of curiosity, what is the default url? Does it not work with docker compose?

docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+  pgadmin:
+    container_name: pgadmin
+    profiles: ["pgadmin"]
+    image: chorss/docker-pgadmin4
+    ports:
+      - 5050:5050
+    depends_on:
+      - rails
+volumes:
+  database:
```
mrotondo, line 86: Just curious because of the dangling colon, is this supposed to have child contents?

bin/docker-entrypoint
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+dockerize -wait tcp://db:5432 -timeout 20s -wait-retry-interval 10s bash -c "bin/rails db:create"
```
mrotondo, line 3: I believe that db:create and parallel:create (and maybe even db:seed) might be unnecessary when using db:prepare and parallel:prepare, but I'm not sure. Fine with this as is, but if you wanted to look into it! If that were true that could let us remove some commands from bin/setup as well.

bin/docker-entrypoint
```diff
@@ -0,0 +1,15 @@
...
+bin/rails db:prepare
+bin/rails parallel:create
+bin/rails parallel:prepare
+bin/rails db:seed
+
+bundle exec rake assets:precompile
```
mrotondo, line 9: No change necessary: This might not be needed now that you got shakapacker working?

Dockerfile.local
```diff
@@ -0,0 +1,65 @@
...
+RUN echo "IRB.conf[:USE_AUTOCOMPLETE] = false" > ./.irbrc
+
+# Install the OpenCV headless library
+RUN pip3 install opencv-python-headless --break-system-packages
+
+# https://github.com/jwilder/dockerize
+ENV DOCKERIZE_VERSION=v0.7.0
+RUN wget -O - https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz | tar xzf - -C /usr/local/bin
+
+ADD . /app
```
mrotondo, line 56: Any harm in keeping this up above line 39? And since you've been digging around in here - do you have any hunches as to why lines 40 and 42 add contents from . to app that should have already been added by the (at least in the production Dockerfile) already been added by this line?

Dockerfile.local
```diff
@@ -0,0 +1,65 @@
...
+ && echo "${SUPERCRONIC_SHA1SUM}  ${SUPERCRONIC}" | sha1sum -c - \
+ && chmod +x "$SUPERCRONIC" \
+ && mv "$SUPERCRONIC" "/usr/local/bin/${SUPERCRONIC}" \
+ && ln -s "/usr/local/bin/${SUPERCRONIC}" /usr/local/bin/supercronic
+
+ADD ./vendor/pdftk /app/vendor/pdftk
+RUN /app/vendor/pdftk/install
+
+# gyr-efiler requires Java 8. Download Java 8 and provide a variable for the Ruby app.
+ENV OPENJDK8_URL=https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u292-b10/OpenJDK8U-jre_x64_linux_hotspot_8u292b10.tar.gz \
```
mrotondo, line 31: If we aren't setting up the GYR Efiler on this setup, which is reasonable since we don't expect to exercise efile submission from development environments, we probably don't need to install the JDK and this whole clause can go.

Dockerfile.local
```diff
@@ -0,0 +1,65 @@
...
+RUN NODE_ENV=production yarn install --frozen-lockfile
+ADD .ruby-version Gemfile Gemfile.lock /app/
+
+RUN gem install bundler:$(cat Gemfile.lock | tail -1 | tr -d " ") --no-document \
+&& bundle install
+
+RUN echo "IRB.conf[:USE_AUTOCOMPLETE] = false" > ./.irbrc
+
+# Install the OpenCV headless library
+RUN pip3 install opencv-python-headless --break-system-packages
```
mrotondo, line 50: Glad to see you worked around the warning this spits out. As a heads up, we're working on removing this dependency as it's no longer used, which we found out while looking into this after hearing you'd had an issue.

docker-compose.yml
mmazanec22: No: https://docs.docker.com/reference/compose-file/volumes/#attributes

docker-compose.yml
mmazanec22: Not sure but it wasn't working and I didn't want to change the config and risk disrupting local install, so using this flag seemed the safest bet for now.

config/database.yml
```diff
@@ -2,6 +2,11 @@ local_default: &local_default
   adapter: postgis
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + 1 %>
   timeout: 5000
+  host: <%= ENV.fetch("RAILS_DB_HOST", "localhost") %>
+  port: <%= ENV.fetch("RAILS_DB_PORT", "5432") %>
+  # TODO: what should the default username and password be?
+  username: <%= ENV.fetch("RAILS_DB_USERNAME", "postgres") %>
```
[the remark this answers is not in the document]
mmazanec22: Ty for looking into this!

Dockerfile.local
mmazanec22: Docker would have to reinstall all the deps every time we make a file change. [This article explains it pretty well](https://maximorlov.com/a-beginners-guide-to-building-a-docker-image-of-your-nodejs-application/). Here's the excerpt:

> Docker images are built in layers and each line in a Dockerfile represents a layer. When you build an image, Docker tries to speed up the build time by only rebuilding the layer that has changed, along with the layers on top of it (the ones below in the Dockerfile). If we copy the entire codebase before installing our dependencies, on each change we make during development Docker will have to reinstall all our dependencies, even though most of the time they haven't changed. Whereas now Docker will only run npm install if your package.json or package-lock.json have changed. If not, then it will only copy the latest changes in your codebase. Building an image can take some time so this is a sane optimisation we want to make use of.

docker-compose.yml
mmazanec22: Good point, removed

Dockerfile.local
mmazanec22: This allows us to run `docker compose up --build` after rebasing to make sure any config changes are applied without it taking forever

bin/docker-entrypoint
mmazanec22: From what I could find it seems like that's a maybe, depending on how we're using those two tools. I'm going to leave that judgment call for when I have spent more time working in this repo!

mrotondo (approved):
Looks good, thanks for doing this!

PR #4719, opened 2024-08-26
fix flaky test by removing hardcoded ID expectation (via its presence in coalition name)

mrotondo:
The issue is that if this runs after any other tests that create Coalitions, even if they clean up their database changes, the auto-incrementing ID will give the Coalition this test creates an ID (and thus a name) greater than 1.


mpidcock (approved):
good eyes

PR #4722, opened 2024-08-27
FYST-319 Make AZ flow explorer functional

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-319
## Is PM acceptance required?

## What was done?
- Different filing status were added: single, mfs, mfj, hoh, qss
- Different xmls were sued for the different filing status 
- Attributes were added for hoh, mfs, mfj were added for dependent and spouse

## How to test?
Use heroku link, make sure to adjust session toggles and check the flow explorer 

- Before
- After

### DIFF
--- a/app/controllers/flows_controller.rb
+++ b/app/controllers/flows_controller.rb
@@ -640,8 +640,6 @@ class FlowsController < ApplicationController
         xml_service.read("az_sherlock_mfs")
       when :head_of_household
         xml_service.read("az_alexis_hoh_w2_and_1099")
-      else
-        xml_service.read("az_tycho_loanded") # Default to single if status is not picked?
       end
     end
 
--- a/app/controllers/flows_controller.rb
+++ b/app/controllers/flows_controller.rb
@@ -9,7 +9,7 @@ class FlowsController < ApplicationController
   SAMPLE_GENERATOR_TYPES = {
     ctc: [:single, :married_filing_jointly],
     gyr: [:single, :married_filing_jointly],
-    state_file_az: [:head_of_household],
+    state_file_az: [:single, :married_filing_jointly, :qualifying_surviving_spouse, :married_filing_seperately, :head_of_household],
     state_file_ny: [:head_of_household],
   }.freeze
 
@@ -39,6 +39,7 @@ class FlowsController < ApplicationController
       if intake.respond_to?(:client)
         sign_in(intake.client)
       elsif [:state_file_az, :state_file_ny].include?(type)
+        intake.create_state_file_analytics!
         sign_in intake
       end
     else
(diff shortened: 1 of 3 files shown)


app/controllers/flows_controller.rb
```diff
@@ -626,9 +627,24 @@ def self.ny_attributes(first_name: 'Testuser', last_name: 'Testuser')
...
+      when :married_filing_jointly
+        xml_service.read("az_martha_v2")
+      when :qualifying_widow
+        xml_service.read("az_leslie_qss_v2")
+      when :married_filing_separately
+        xml_service.read("az_sherlock_mfs")
+      when :head_of_household
+        xml_service.read("az_alexis_hoh_w2_and_1099")
+      end
+    end
```
jnf, line 644: Because these are hardcoded, I could see an argument for moving this logic into `XmlReturnSampleService`. Something like `StateFile::XmlReturnSampleService.new.sample_for(state_code, filing_status)`. If we're moving to a world where there's a similar case statement for each active state and a case for each supported filing status, I'd say move that logic into `XmlReturnSampleService` now. If not now, then the very next time we add a state to the flow explorer.

jnf (commented):
@tahsinaislam this looks great. I have one question about where a bit of logic should live. Let's figure out our stance/plan on that, then 🚀

app/controllers/flows_controller.rb
```diff
@@ -680,10 +718,12 @@ def generate_state_file_intake(params)
       _type = params.keys.find { |k| k.start_with?('submit_') }&.sub('submit_', '')&.to_sym
       attributes = self.class.send("#{@us_state}_attributes",
         first_name: params[:flows_controller_sample_intake_form][:first_name],
-        last_name: params[:flows_controller_sample_intake_form][:last_name]
+        last_name: params[:flows_controller_sample_intake_form][:last_name],
+                                   filing_status: _type
       )
       intake_class = StateFile::StateInformationService.intake_class(@us_state)
       intake = intake_class.create(attributes)
+      intake.efile_submissions.create!
```
jenny-heath, line 726: wow this seems pretty significant. it's a little strange to have a submission from the beginning but i guess the flow explorer fills in a lot a the beginning. was it causing an error to not have a submission?

tahsinaislam: @jenny-heath yeah it wasn't creating a submission on its own for some reason but not sure if this is the best way to go about it

jenny-heath (approved):

app/controllers/flows_controller.rb
jenny-heath: maybe the intention was for it to fill out up to the point before hitting submit and you're supposed to click that button manually?

tahsinaislam: @jenny-heath yes! that was the previous way it worked should i keep as that? that makes sense to me

jnf (approved):

PR #4723, opened 2024-08-28
Audit and add validations to XML for AZ-140 (and other documents, including NY stuff because it was easier than not doing it)

mrotondo:
unify our string output filtering and add character limits to match the output XML schema restrictions in as many places as i could find that we didn't have any (even where we're just truncating something we got from Direct File)

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-240
## Is PM acceptance required?
## What was done?
- Renamed `truncate` to `sanitize_for_xml`, made the length param optional and added a few tests
- Replaced all uses of `&.strip&.gsub("\s+", " ")` with calls to `sanitize_for_xml`
- Added truncation where we weren't enforcing a character limit present in the output XSD
## How to test?
- Unit tests, and especially our regression tests, should make sure that we still work on all previously working input
- I will add an additional test that confirms that we no longer pass through last names that won't validate, but you could also confirm that manually by going through the flow with a very long primary last name.
  - _Update: I did not add this test, as it turns out we don't actually currently validate our return headers currently. I'll look to add this in a separate PR, as the change is out of scope._

### DIFF
--- a/app/services/state_file/state_information_service.rb
+++ b/app/services/state_file/state_information_service.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module StateFile
   class StateInformationService
     class << self
--- a/app/lib/submission_builder/document.rb
+++ b/app/lib/submission_builder/document.rb
@@ -99,13 +99,13 @@ module SubmissionBuilder
     def process_mailing_street(xml)
       return unless @submission.data_source.direct_file_data.mailing_street.present?
 
-      mailing_street = @submission.data_source.direct_file_data.mailing_street.strip.gsub(/\s+/, ' ')
+      mailing_street = sanitize_for_xml(@submission.data_source.direct_file_data.mailing_street)
 
       if mailing_street.length > 30
         process_long_mailing_street(xml, mailing_street)
       else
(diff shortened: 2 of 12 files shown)


tahsinaislam (approved):

app/services/state_file/state_information_service.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
```
tahsinaislam, line 1: can probably take out this note

mrotondo: Thanks for the detailed look! This isn't actually a note, but a "magic comment" (sometimes called a "pragma") that Rubymine puts in new ruby files (and which we should probably have more consistently throughout our codebase).

When this comment is at the top of the file, it tells Ruby that the value of the string literals in that file (anything in single or double quotes) can not be changed by other code. In this case, it's not so critical since we use "deep_freeze" on the whole `STATES_INFO` dictionary, but it can matter in some cases both for performance and to make sure that strings are not changed unintentionally while the program is running. Happy to talk more about it if you're interested - it's pretty low-level/nuanced stuff that doesn't usually matter. Here's a document that goes into a bit more detail: https://gist.github.com/fxn/bf4eed2505c76f4fca03ab48c43adc72

PR #4725, opened 2024-08-28
Nj 27 add nj

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/27
- Yes?
## What was done?
Add NJ following NC and Warshington examples.
## How to test?
Added fake NJ directfile import xml, copied from NY examples. You can manually run through the process an import one to see it working.

Also added landing page controller test and return xml test.  Added NJ to schema file loader and state information service tests.


spec/factories/state_file_nj_intakes.rb
```diff
@@ -0,0 +1,111 @@
...
+#
+FactoryBot.define do
+  factory :state_file_nj_intake do
+    transient do
+      filing_status { 'single' }
+    end
+
+    eligibility_lived_in_state { "yes" }
+    eligibility_out_of_state_income { "no" }
+    raw_direct_file_data { File.read(Rails.root.join('app', 'controllers', 'state_file', 'questions', 'df_return_sample.xml')) }
```
mrotondo, line 87: [PEBBLE] Do you want to use one of the NJ DF return xmls you created as the baseline DF XML for this factory?

mrotondo (approved):
Looks good! Left some comments but can be squashed & merged after those have been addressed. Sorry if I'm using the wrong sizes of rocks (pebble, dust etc), I missed the conversation last week and just read the zoom chat.

PR #4737, opened 2024-09-05
Gyr 569/add hold status to greeter access

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-569

## What was done?
- Users that were in the `file_hold` status were not being considered. We modified the ability file to enable them to display if and only if they are assigned to the given greeter for the user.
- **Note** If a user has multiple intakes, and those other intakes are in a status that does not require a user to be assigned (namely `intake_ready`, `intake_greeter_info_requested`, `intake_needs_doc_help`), the user will still be displayed. This is because while we scope on returns we are beginning from the context of the client and not the context of the tax return. To modify this likely would require inverting how that happens

## How to test?
- Unit tests are provided in the PR

### Manual Testing
- Navigate to `/en/hub/clients`
  - Authenticate as a greeter if necessary
- Note that there are only two clients available, each with two intakes.
  - Captain Hook
    - `intake_ready`
    - `intake_ready`
  - Peter Pan
    - `intake_ready`
    - `intake_ready`
- Manipulate intake to be in a state other than those that do not require a user to be assigned. In other words not:
  - `intake_ready`,
  - `intake_greeter_info_requested`,
  - `intake_needs_doc_help`,
- Note that the client is no longer accessible

- Before
image

- After
image

### DIFF
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -95,7 +95,8 @@ class Ability
         Client,
         tax_returns: {
           current_state: [
-            'file_not_filing'
+            'file_not_filing',
+            'file_hold',
           ],
           assigned_user: user,
         },
--- a/app/state_machines/tax_return_state_machine.rb
+++ b/app/state_machines/tax_return_state_machine.rb
@@ -67,7 +67,8 @@ class TaxReturnStateMachine
         'intake_needs_doc_help'
       ],
       'file' => [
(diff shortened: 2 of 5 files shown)


jenny-heath (approved):
did a pair review, looks good

PR #4738, opened 2024-09-05
Add NC screener pages

embarnard:
## Link to pivotal/JIRA issue
## What was done?
- Alternatives considered
## How to test?
  - Test data used
- Risk Assessment
- Before
- After


config/credentials/production.yml.enc
embarnard, line 1: need to remove

powersurge360 (approved):
Looks good aside from the left-in credentials file. Was demo'd for me over pair.

app/forms/state_file/nc_eligibility_residence_form.rb
```diff
@@ -0,0 +1,11 @@
+module StateFile
+  class NcEligibilityResidenceForm < QuestionsForm
```
arinchoi03, line 2: must be some meta-programming thing I don't understand -- when is this NcEligibilityResidenceForm used? When I search for uses for this I only see it be referred to in the spec file 😕 must be something obvious I'm not understanding

config/locales/es.yml
```diff
@@ -2459,6 +2459,16 @@ es:
...
+        edit:
+          out_of_state_income_html: "¿Usted (o su cónyuge, si presentan la declaración juntos) obtuvo algún ingreso en otro estado en %{filing_year}? (Esto incluye los ingresos por desempleo <strong>recibidos</strong> de otro estado)."
+          title: Primero, veamos si puedes usar esta herramienta para presentar tus impuestos.
+          withdraw_529_html: ¿Retiró usted (o su cónyuge, si presentan una declaración juntos) dinero de un <a href="https://www.cfnc.org/save-for-college/" target="_blank" rel="noopener nofollow">Cuenta de ahorros para la universidad 529</a> y usarla para gastos no calificados en %{filing_year}?
+      nc_eligibility_residence:
+        edit:
+          lived_in_state_html: "¿Usted (y su cónyuge, si presentan la declaración juntos) vivieron en %{state} durante <strong>todo</strong> el %{filing_year}?"
+          subtitle: Para la temporada de impuestos %{filing_year}, esta herramienta solo podrá atender a personas con situaciones tributarias específicas. Responda las siguientes preguntas para saber si puede utilizarlo.
+          title: Primero, veamos si puedes usar esta herramienta para presentar tus impuestos.
```
arinchoi03, line 2471: will need a primer on how we translate things here

app/forms/state_file/nc_eligibility_residence_form.rb
powersurge360: Forms and the form controllers are complicated. I am still learning too, but here's a snippet from my notes explaining forms. I need to clean up my notes about form controllers and then I'll share that as well:

Representing a segment of an intake using a convention-laden ruby object hierarchy.

# A Guided Tour

We use forms as the default way of managing our tax questions. They get the majority of their magic by inheriting from a set of active model mixins and supplement with some Code from America specific business logic thereafter. Let&rsquo;s take a guided tour at what forms are and how they&rsquo;re defined.

**Note:** This tour was conducted on <span class="timestamp-wrapper"><span class="timestamp">[2024-09-16 Mon] </span></span> and things may have changed if you are reading this much later than then.

At the start we&rsquo;ll look at the form class definition.

```ruby
    class Form
      include ActiveModel::Model
      include ActiveModel::AttributeAssignment
      include ActiveModel::Validations::Callbacks
      extend ActiveModel::Callbacks
    
      define_model_callbacks :save
    
      def error_summary(exclude_keys: [])
        if errors.present?
          visible_errors = errors.messages.select{ |key, _| exclude_keys.exclude? key }
          concatenated_message_strings = visible_errors.map{ |key, messages| messages.join(" ")}.join(" ")
          "Errors: " + concatenated_message_strings
        end
      end
    
      def self.form_param
        name.delete("::").underscore
      end
    end
```

Looking at the above sample, we can see that we include much from the ActiveModel mixins. I won&rsquo;t cover those in depth in this tour because they are documented externally. Briefly, this enables validation to work as if they were ActiveRecord models, sets up attribute assignment helpers, and sets up callbacks. And notably `define_model_callbacks` sets up a `before_save` method. This is the base root class and is not used directly. Instead, we&rsquo;ll often use the `QuestionsForm`.

```ruby
    class QuestionsForm < Form
      include FormAttributes
    
      attr_accessor :intake
    
      def initialize(intake = nil, params = {})
        @intake = intake
        super(params)
      end
    
      def self.from_intake(intake)
        attribute_keys = Attributes.new(attribute_names).to_sym
        new(intake, existing_attributes(intake).slice(*attribute_keys))
      end
    end
```

Mostly the questions form just includes `initialize` and `from_intake`. This is just to attach an [intake](20240916112237-intake.md) to the form for manipulation by subclassing methods. Remember that intakes are sort of our way of representing a customer, it&rsquo;s the series of questions and answers that will eventually constitute their return. All of the things you may have seen to do with attribute manipulation actually happens in the `FormAttributes` mixin. We&rsquo;ll look at that next.

```ruby
    module FormAttributes
      extend ActiveSupport::Concern
    
      included do
        extend AutoStripAttributes
        class_attribute :attribute_names
      end
    
      class_methods do
        def set_attributes_for(model, *attributes)
          scoped_attributes[model] = attributes
          self.attribute_names = scoped_attributes.values.flatten
          attribute_strings = Attributes.new(attributes).to_s
    
          attr_accessor(*attribute_strings)
          auto_strip_attributes *attribute_strings, virtual: true
        end
    
        def before_validation_squish(*attributes)
          auto_strip_attributes *attributes, virtual: true, squish: true
        end
    
        def scoped_attributes
          @scoped_attributes ||= {}
        end
    
        def existing_attributes(model)
          if model.present?
            HashWithIndifferentAccess.new(model.attributes)
          else
            {}
          end
        end
      end
    
      def attributes_for(model)
        self.class.scoped_attributes[model].reduce({}) do |hash, attribute_name|
          hash[attribute_name] = send(attribute_name)
          hash
        end
      end
    end
```
Skipping over the `included` block for now, we&rsquo;ll instead look at the things in the `class_methods` block. We&rsquo;ll start by looking at the `scoped_attributes` reader:

```ruby
    def scoped_attributes
      @scoped_attributes ||= {}
    end
```

This merely sets up a hash if one does not exist and returns it otherwise. The scoped attributes being a hash is what enables us to specify multiple model&rsquo;s attributes at once and handle validation therein. Skipping ahead a little, you may have seen an API like:

```ruby
    set_attributes_for :intake, :eligibility_lived_in_state, :eligibility_married_filing_separately
```

The first argument specifies the model, the values afterwards specify the actual properties. That method is actually defined here in the `FormAttributes` mixin, so let&rsquo;s see how it works:

```ruby
    def set_attributes_for(model, *attributes)
      scoped_attributes[model] = attributes
      self.attribute_names = scoped_attributes.values.flatten
      attribute_strings = Attributes.new(attributes).to_s
    
      attr_accessor(*attribute_strings)
      auto_strip_attributes *attribute_strings, virtual: true
    end
```

I think this is nearly the least intuitive thing I&rsquo;ve found in our codebase so far. Very dense, very confusing. Let&rsquo;s pull it apart. As we saw before, function accepts one or more arguments, the model followed by any number of attributes. It uses the scoped attributes hash to keep track of each model and its associated attributes. The example invocation would result in a `scoped_attributes` hash that looks something like:

```ruby
    {
      intake: [
        :eligibility_lived_in_state,
        :eligibility_married_filing_separately
      ]
    }
```

Next, it retrieves all attributes listed anywhere by getting values of all top level keys in the hash (in other words a list of lists of attributes). That would look something like:

```ruby
    [
      [
        :eligibility_lived_in_state,
        :eligibility_married_filing_separately
      ]
    ]
```

Finally it flattens this and assigns it to the `attribute_names` class variable. We&rsquo;ll take the next three lines together:

```ruby
    def set_attributes_for(model, *attributes)
      # ... snip
      attribute_strings = Attributes.new(attributes).to_s
    
      attr_accessor(*attribute_strings)
      auto_strip_attributes *attribute_strings, virtual: true
    end
```

In english:

1.  Call into the `Attributes` object to ultimately reserialize all of the keys as strings
2.  Add an `attr_accessor` (that is to say, both an `attr_reader` and an `attr_writer`) for each attribute to the form itself
3.  Use the [auto<sub>strip</sub><sub>attributes</sub>](https://github.com/holli/auto_strip_attributes) gem to strip any whitespace, even if it is a virtual property (that is, a property that is not mapped to a column)

And that&rsquo;s the bulk of the `QuestionsForm`. There are other methods but most of them are more straightforward and not part of the standard path of using the set of objects.

config/locales/es.yml
embarnard: @arinchoi03 we do integrate with transifex! we just put placeholder strings (google translated strings) for now and then eventually someone will translate in transifex and we will pull it into our app https://www.notion.so/cfa/Transifex-manual-sync-e5e04ebb6e1e49fa9bac8288c67cbae9

PR #4740, opened 2024-09-05
Fyst 318 implement data model and add use tax page to nc flow

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-318
## Is PM acceptance required?
## What was done?
- Added a sales & use tax page for NC
- I created this by duplicating the relevant NY pages, but have implemented it so that it should be able to be re-used across other states with sales & use tax for out-of-state purchases.
- I didn't delete or change the NY code - we could refactor it to use this new implementation but I would prefer to not do so and just delete it when we remove NY, accepting that there is some code duplication in the mean time.
- I used a decimal column for the money field, as I believe we are doing for all such fields now
## How to test?
- I added unit tests for the new page's controller & form
- Implementation and testing of use tax calculation and xml/pdf output is left for a future story
image


app/forms/state_file/sales_use_tax_form.rb
```diff
@@ -0,0 +1,35 @@
...
+                       :untaxed_out_of_state_purchases,
+                       :sales_use_tax_calculation_method,
+                       :sales_use_tax
+
+    validates :untaxed_out_of_state_purchases, presence: true
+    validates :sales_use_tax_calculation_method, presence: true, if: -> { untaxed_out_of_state_purchases == "yes" }
+    validates :sales_use_tax,
+      presence: true,
+      numericality: {
+        message: ->(_object, _data) { I18n.t("state_file.questions.sales_use_tax.edit.enter_valid_dollar_amount") },
```
powersurge360, line 13: I'd bet this is the bit that's failing for the 'unused' i18n bit. If it is, the `activerecord.errors.models` [bit of the locale files](https://github.com/codeforamerica/vita-min/blob/ec24141cbeb4da680bbdfadc6490dd6984c93aee/config/locales/en.yml#L12) is for this sort of things, but with us using forms rather than models I don't imagine it would get picked up. But maybe it would since it is handled in an ancestor class by including activemodel? Could be worth a quick smoke test to see if it gets picked up.

powersurge360 (commented):
Seems fine. Soft approval because I haven't set up a page like this before and some i18n book keeping needs to be done to get tests to pass.

app/forms/state_file/sales_use_tax_form.rb
powersurge360: Whoops, noticed this is wip lol. Need to update my filter, sorry

powersurge360 (approved):
Seems good to me.

PR #4741, opened 2024-09-06
Nj 36 add pdf

aloverso:
## Link to pivotal/JIRA issue
- [e.g. https://codeforamerica.atlassian.net/browse/ABCD-123](https://github.com/newjersey/affordability-pm/issues/36)

## What was done?
- Generate PDF
- Fill in header section of PDF with name, spouse, SSNs, address, CountyCode 
- PDF field names are extracted by pdftk (the field name output is in NJ OOI GDrive folder as `pdf fields.txt`

image


spec/factories/state_file_nj_intakes.rb
```diff
@@ -98,6 +98,8 @@
         qualifying_widow: 5,
       }[evaluator.filing_status.to_sym] || evaluator.filing_status
       intake.direct_file_data.filing_status = numeric_status
+      intake.direct_file_data.primary_ssn = evaluator.primary_ssn || intake.direct_file_data.primary_ssn
+      intake.direct_file_data.spouse_ssn = evaluator.spouse_ssn || intake.direct_file_data.spouse_ssn
```
aloverso, line 102: I'm not entirely sure why these lines are necessary, but it's what allows me to pass in an SSN to the factory create and have it actually come back with the SSN I gave it. Without these lines, the SSN will get overridden by whatever is in the raw DF data XML on line 88 here.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+        create :efile_submission, tax_return: nil, data_source: create(
+          :state_file_nj_intake,
+        ) }
+
+      it 'enters values into PDF' do
+        # address values from zeus_one_dep.xml
+        expect(pdf_fields["SpousesCU Partners SSN if filing jointly"]).to eq "391 US-206 B"
+        expect(pdf_fields["CountyMunicipality Code See Table page 50"]).to eq "Hammonton"
+        expect(pdf_fields["State"]).to eq "NJ"
+        expect(pdf_fields["ZIP Code"]).to eq "08037"
```
aloverso, line 197: I couldn't figure out how to pass these into the factory and have them actually be used without being overwritten by the DF raw data in the factory... so at the end of the day this is just referencing the values in that XML instead of setting them in the test context (like the other tests in this file do). I'd love to fix this but I'm not sure how.

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+      spouse_ssn = get_spouse_ssn
+      {
+        # county code
+        "CM4": county_code[1],
+        "CM3": county_code[2],
+        "CM2": county_code[3],
+        "CM1": county_code[4],
+
+        # taxpayer ssn
+        "undefined": taxpayer_ssn[0],
```
jachan, line 29: [DUST] How can we verify that these field names map to the correct fields in the PDF? The `pdf fields.txt` document in the Google drive shows the output of `pdftk 1040.pdf dump_data_fields` but it isn't very helpful for determining which fields each name maps to.

jachan (approved):
Just looked over the commit that adds PDF support and not the one that we're rebasing on top of. Left some questions about strategy. Overall, looks very clear and straightforward, thanks for laying the groundwork for PDF generation that others can build off of!

app/lib/pdfs/nj1040-TY2023.pdf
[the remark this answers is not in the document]
mmazanec22: There's a ticket for this!

spec/factories/state_file_nj_intakes.rb
mmazanec22: I wonder if we can do this for all of the fields rather than individually?

mmazanec22 (approved):

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+      @xml_document = builder.new(submission).document
+    end
+
+    def hash_for_pdf
+      county_code = get_county_code
+      taxpayer_ssn = get_taxpayer_ssn
+      spouse_ssn = get_spouse_ssn
+      {
+        # county code
+        "CM4": county_code[1],
```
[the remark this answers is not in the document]
aloverso: I'm leaning towards no, not worth it
a) This is the 2023 PDF so definitely not worth it for this one
b) In general, this is going to be a "set it and forget it" kind of situation, we're not going to have to be interacting with these field names on a regular basis, I think

tahsinaislam (approved):

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+        # spouse ssn
+        "undefined_3": spouse_ssn ? spouse_ssn[0] : "",
+        "undefined_4": spouse_ssn ? spouse_ssn[1] : "",
+        "undefined_5": spouse_ssn ? spouse_ssn[2] : "",
+        "Text9": spouse_ssn ? spouse_ssn[3] : "",
+        "Text10": spouse_ssn ? spouse_ssn[4] : "",
+        "Text11": spouse_ssn ? spouse_ssn[5] : "",
+        "Text12": spouse_ssn ? spouse_ssn[6] : "",
+        "Text13": spouse_ssn ? spouse_ssn[7] : "",
+        "Text14": spouse_ssn ? spouse_ssn[8] : "",
```
jenny-heath, line 51: [DUST] could also do something like:
```
answers = {
...county code etc...
}

if spouse_ssn
  answers.merge!({ ...spouse ssn stuff })
end
```

we have a similar pattern in other pdf fillers but def not required. interested in any opinions on one approach vs the other!

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        # address
+        "SpousesCU Partners SSN if filing jointly": get_address, # address text field
```
jenny-heath, line 54: [less than dust] this is.....the address field?

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      format_name(first_name, last_name, middle_initial, suffix)
+    end
+
+    def format_name(first_name, last_name, middle_initial, suffix)
+      "#{last_name} #{format_no_last_name(first_name, middle_initial, suffix)}"
+    end
+
+    def format_no_last_name(first_name, middle_initial, suffix)
+      "#{first_name}#{middle_initial ? (" " + middle_initial) : ""}#{suffix ? (" " + suffix) : ""}"
```
jenny-heath, line 104: [DUST] could also do `[first_name, middle_initial, suffix].compact.join(" ")`

jenny-heath: actually for that matter i guess you could just do `[last_name, first_name, middle_initial, suffix].compact.join(" ")` ?

jenny-heath: oh jk i now see the need for a separate no-last-name method

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    def get_taxpayer_ssn
+      @xml_document.at("ReturnHeaderState Filer Primary TaxpayerSSN")&.text
+    end
+
+    def get_address
+      address_line_1 = @xml_document.at("ReturnHeaderState Filer USAddress AddressLine1Txt")&.text
+      address_line_2 = @xml_document.at("ReturnHeaderState Filer USAddress AddressLine2Txt")&.text
+      address_line_1 + (address_line_2.present? ? " #{address_line_2}" : "")
```
jenny-heath, line 74: [DUST] see comment on `format_no_last_name`

app/lib/pdf_filler/nj1040_pdf.rb
jenny-heath: ok last thing sorry, i just realized i'm not totally sure whether the blank ones will come back as `nil` or empty strings but i was assuming `nil`. in the case of empty strings it would be `.reject(&:empty?)` instead of `.compact`

spec/factories/state_file_nj_intakes.rb
jenny-heath: so, this could have something to do with the fact that we usually read ssn like this `intake.primary.ssn` which you can see the inner workings of [here](https://github.com/codeforamerica/vita-min/blob/main/app/models/state_file_base_intake.rb#L199). i just realized the nj intake has a `primary_ssn` column whereas our other state file intake models do not. are you collecting ssn from the client or planning to get it from direct file?

jenny-heath: related but probably out of scope for this PR: this made me realize that the nj intake has a `primary_ssn` column whereas our other state file intake models do not. are you collecting ssn from the client or planning to get it from direct file?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
jenny-heath: see comment on factory!

app/lib/pdf_filler/nj1040_pdf.rb
aloverso: yes 🙃 
I'm usually anti-code-comments but I thought this one warranted a comment rip

spec/factories/state_file_nj_intakes.rb
aloverso: note to self - will circle back to these factory suggestions in a future commit that will remove primary_ssn and spouse_ssn fields altogether from the data model

PR #4744, opened 2024-09-06
[WIP] Added lines 6 and 7

jachan:
# This PR is meant to validate the approach for unit testing before adding even more logic/tests for lines 8-13 of the form

## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/35

## What was done?
- Added Line 6 and 7 logic to calculator and XML submission filer classes

## How to test?
- Unit tests

### DIFF
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
@@ -50,9 +50,15 @@ module SubmissionBuilder
                       raise "Filing status not found"
                     end
                   end
-                  
+
                   xml.Exemptions do
                     xml.YouRegular "X"
+                    xml.SpouseCURegular calculated_fields.fetch(:NJ1040_LINE_6_SPOUSE) ? "X" : ""
+                    xml.DomesticPartnerRegular ""
+                    xml.Line6 calculated_fields.fetch(:NJ1040_LINE_6)
+                    xml.YouOver65 calculated_fields.fetch(:NJ1040_LINE_7_SELF) ? "X" : ""
+                    xml.SpouseCuPartner65OrOver calculated_fields.fetch(:NJ1040_LINE_7_SPOUSE) ? "X" : ""
+                    xml.Line7 calculated_fields.fetch(:NJ1040_LINE_7)
                     xml.NumOfQualiDependChild qualifying_dependents.count(&:qualifying_child?)
                     xml.NumOfOtherDepend qualifying_dependents.count(&:qualifying_relative?)
                   end
@@ -86,7 +92,7 @@ module SubmissionBuilder
             end
 
             def calculated_fields
-              @nj1040_fields ||= intake.tax_calculator.calculate
+              @nj1040_fields ||= @submission.data_source.tax_calculator.calculate
             end
           end
         end
--- a/spec/lib/efile/nj/nj1040_spec.rb
+++ b/spec/lib/efile/nj/nj1040_spec.rb
@@ -46,6 +46,7 @@ describe Efile::Nj::Nj1040 do
           expect(instance.lines[:NJ1040_LINE_7].value).to eq(1000)
         end
       end
+
       context 'when filer is younger than 65' do
         before do
           intake.primary_birth_date = Date.new(2000, 1, 1)
@@ -64,6 +65,7 @@ describe Efile::Nj::Nj1040 do
         end
       end
     end
+
     context 'when filer has a spouse,' do
       let(:intake) { create(:state_file_nj_intake, :married) }
       context 'when filer is older than 65 and spouse is older than 65' do
@@ -83,6 +85,7 @@ describe Efile::Nj::Nj1040 do
           expect(instance.lines[:NJ1040_LINE_7].value).to eq(2000)
         end
       end
+
       context 'when filer is younger than 65 and spouse is older than 65' do
         let(:intake) { create(:state_file_nj_intake, :married_spouse_over_65) }
         before do
@@ -100,6 +103,7 @@ describe Efile::Nj::Nj1040 do
           expect(instance.lines[:NJ1040_LINE_7].value).to eq(1000)
         end
       end
+
       context 'when filer is older than 65 and spouse is younger than 65' do
         let(:intake) { create(:state_file_nj_intake, :married, :self_over_65) }
         before do
@@ -117,6 +121,7 @@ describe Efile::Nj::Nj1040 do
           expect(instance.lines[:NJ1040_LINE_7].value).to eq(1000)
         end
       end
+
       context 'when filer is younger than 65 and spouse is younger than 65' do
         let(:intake) { create(:state_file_nj_intake, :married) }
         before do
--- a/spec/lib/efile/nj/nj1040_spec.rb
+++ b/spec/lib/efile/nj/nj1040_spec.rb
@@ -1,0 +1,139 @@
+require 'rails_helper'
+
+describe Efile::Nj::Nj1040 do
+  let(:intake) { create(:state_file_nj_intake) }
+  let(:instance) do
+    described_class.new(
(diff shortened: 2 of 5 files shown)


spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -13,5 +13,85 @@
...
+      context "when filer is over 65 and spouse is under 65" do
+        let(:intake) { create(:state_file_nj_intake, :self_over_65, :married) }
+        it "populates line 7 XML fields" do
+          expect(xml.at("Exemptions YouOver65").text).to eq("X")
+          expect(xml.at("Exemptions SpouseCuPartner65OrOver").text).to eq("")
+          expect(xml.at("Exemptions Line7").text).to eq("1000")
+        end
+      end
+
+      context "when filer is over 65 and spouse is over 65" do
```
jachan, line 69: Is there a better way to dry up these tests for every possible test-case? Concerned that I'm already adding more tests for just these two lines as the length of the tests in the entire `az140_calculator spec`.

It seems like I'm duplicating a lot of logic between the XML generation and the calculator, also testing the cases where no spouse exists separately from the case where a spouse exists and is under/over 65.

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    context 'when filer has a spouse' do
+      let(:intake) { create(:state_file_nj_intake, :married) }
+      before do
+        instance.calculate
+      end
+      it { expect(instance.lines[:NJ1040_LINE_6].value).to eq(2000) }
+    end
+  end
+
+  describe '#calculate_line_7' do
```
embarnard, line 30: I think instead of trying to separate each of these calculations with the `describe '#calculate_line_7'` I might try to simplify it by just having the larger context blocks of which filing status they have (or other circumstances that come up) and then put the expectations of what the line values should be in there. Additionally, there doesn't necessarily need to be a separate it statement for each one, they could be grouped together and hopefully that will help with making the test file more readable and easier to manage as y'all continue to add more items

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+          expect(instance.lines[:NJ1040_LINE_7_SPOUSE].value).to eq(false)
+        end
+
+        it 'calculates total line 7 exemption amount correctly' do
+          expect(instance.lines[:NJ1040_LINE_7].value).to eq(0)
+        end
+      end
+    end
+  end
+end
```
embarnard, line 144: I think my general comments for this file is that the structure could look like this 

Efile::Nj::Nj1040 do
  context when single
   it does "bla bla what you expect it to do"
      expect line 6 to equal x
      expect line 7 to equal y
      expect line 7seld to equal 
      etc etc
   end 
  end

  context when married
   it does "bla bla what you expect it to do"
      expect line 6 to equal x
      expect line 7 to equal y
      expect line 7seld to equal 
      etc etc
   end 
 context when spouse is older than 65
  end

context when primary older than 65
end

context when primary is younger than 65
end

end

... I might be missing some things but let me know if this doesn't make sense

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
embarnard: Same kind of feedback as the the other spec test. I think simplifying and nesting contexts that are similar into each other ("when filer is over 65 and spouse is over 65" could be when spouse is over 65 in the married context instead) and encompassing "it" statements for multiple lines could make it more concise. We also don't necessarily do tests for every matrix of possibility although it is theoretically good to do that and up to your discretion. For example, IMO I would probably just test for spouse 65 and older and younger and not the combination of filer and spouse both 65 and older or one 65 or older.

embarnard (commented):
Great job! You're picking up things really quickly and I'm impressed. Left some comments let me know if they don't make sense or if you have questions

spec/lib/efile/nj/nj1040_spec.rb
jachan: Thanks! Appreciate the guidance on grouping multiple `expect` blocks together. I was trying to follow https://www.betterspecs.org/#single but agree that it makes the test way too long 😭

spec/lib/efile/nj/nj1040_spec.rb
jachan: Can you clarify how this would test the different married situations? 

In my mind there are four cases we need to test when the filer is married: 1T+2T; 1T+2F; 1F+2T; 1F+2F, where T = older and F = younger for primary/spouse respectively. 

Right now in the test block i'm seeing above, I see coverage for each condition individually but not the test cases where they're together.

context 1T
context 2T
context 1F
context 2F

Might be easier to discuss over a call in-person, happy to do so in our next pairing session.

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,144 @@
+require 'rails_helper'
+
+describe Efile::Nj::Nj1040 do
+  let(:intake) { create(:state_file_nj_intake) }
+  let(:instance) do
+    described_class.new(
+      year: MultiTenantService.statefile.current_tax_year,
```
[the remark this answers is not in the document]
jachan: Yeah, looks like it runs both ways. Keeping it as `.statefile` as I think it's simpler/more legible.

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -86,7 +92,7 @@ def intake
             end
 
             def calculated_fields
-              @nj1040_fields ||= intake.tax_calculator.calculate
+              @nj1040_fields ||= @submission.data_source.tax_calculator.calculate
```
mmazanec22, line 95: [dust, very small particles] why access directly instead of using the method above for consistency?

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -50,9 +50,15 @@ def document
                       raise "Filing status not found"
                     end
                   end
-                  
+
                   xml.Exemptions do
                     xml.YouRegular "X"
+                    xml.SpouseCURegular calculated_fields.fetch(:NJ1040_LINE_6_SPOUSE) ? "X" : ""
+                    xml.DomesticPartnerRegular ""
+                    xml.Line6 calculated_fields.fetch(:NJ1040_LINE_6)
```
mmazanec22, line 58: [boulder] Is Line6 a valid XML element per our schemas? Same question for Line7

jachan: Wow, thanks for catching that. I misread the schema - it is not. It appears that the XML simply has fields for each checkbox. With this in mind, my plan is still to use the calculator to generate calculated values for lines 6-12, store them in `calculated_fields` and then sum them for Line 13 `TotalExemptionAmountA`, which is a valid field.

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
jachan: oversight! didn't realize `intake` referred to the method above doing the same thing!

jachan:
Talked with Melanie about the scope of https://github.com/newjersey/affordability-pm/issues/35 and realized that I was doing more than necessary here. I was supposed to put in dummy values for lines 6-12 and then write the summation logic in calculator, pull the field into the XML, and fill the PDF with the field.

I'll end up cherry picking commits from this code to separate branches and sending out PRs for https://github.com/newjersey/affordability-pm/issues/37 (Line 6) and https://github.com/newjersey/affordability-pm/issues/38 (Line 7) separately.

PR #4745, opened 2024-09-06
Fix schema file loader to skip NJ for now

tahsinaislam:
## Is PM acceptance required?

## What was done?

For the tests for schema file loader for `setup:download_efile_schemas` and `setup:unzip_efile_schemas`, nj schema errors were ignored by adding an optional field for each state information in the StateInformationService. Errors were filtered by checking the optional value for each state.

Tests that were tagged required nj schema were altered to skip over nj files when checking for schema file loading.
Cfa staff would have to run the following command to utilize this tag until we receive the nj schemas: 
`cp .rspec-local.example .rspec-local`

## How to test?
Testing would be accomplished through running circle ci flow

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ bin/setup
 ```
 > ℹ️ **Note:** If `bundler` is not installing, ensure that you have `rbenv` installed and are not using the system Ruby version. Check the `.ruby-version` file in the repository to match the version specified. If necessary, update to the correct Ruby version and modify your `.zprofile` or `.zshrc` to point to the correct path.
 
+> ℹ️ **Note:** You may consider at this point copying or symlinking `.rspec-local.example` to `.rspec-local` to exclude specs that are not core to CfA yet. Copying enables you to set other custom flags for your local environment, symlinking enables the flags to automatically go away when they are removed from the example file. Just be careful not to commit any local changes!
 
 #### Add efile resources locally
 
--- a/app/lib/schema_file_loader.rb
+++ b/app/lib/schema_file_loader.rb
@@ -36,11 +36,15 @@ class SchemaFileLoader
     def download_schemas_from_s3(dest_dir)
       s3_client = Aws::S3::Client.new(region: REGION, credentials: s3_credentials)
       get_missing_downloads(dest_dir).each do |download_path|
-        s3_client.get_object(
-          response_target: download_path,
-          bucket: BUCKET,
-          key: File.basename(download_path),
-        )
+        begin
+          puts "this path #{download_path}"
+          s3_client.get_object(
+            response_target: download_path,
+            bucket: BUCKET,
+            key: File.basename(download_path),
+            )
+        rescue Aws::S3::Errors::NoSuchKey
+        end
       end
     end
 
--- a/app/lib/schema_file_loader.rb
+++ b/app/lib/schema_file_loader.rb
@@ -4,12 +4,14 @@ class SchemaFileLoader
   REGION = "us-east-1".freeze
   EFILE_SCHEMAS_FILENAMES = (
     [
+      # Format is schema name, directory, and whether the file is optional
       ["efile1040x_2020v5.1.zip", "irs", false],
       ["efile1040x_2021v5.2.zip", "irs", false],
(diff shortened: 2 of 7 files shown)


app/lib/schema_file_loader.rb
```diff
@@ -4,13 +4,15 @@ class SchemaFileLoader
...
+      ["efile1040x_2020v5.1.zip", "irs", false],
+      ["efile1040x_2021v5.2.zip", "irs", false],
+      ["efile1040x_2022v5.3.zip", "irs", false],
+      ["efile1040x_2023v5.0.zip", "irs", false]
     ] +
-      StateFile::StateInformationService.state_schema_file_names.map do |schema_file_name|
-        [schema_file_name, "us_states"]
+      StateFile::StateInformationService::STATES_INFO.map do |key, values|
+        # TODO: If adding another member to this array, consider refactoring to a hash instead
+        [values[:schema_file_name], "us_states", values[:optional]]
```
powersurge360, line 15: I felt like this was on the very cusp of needing a refactor. I left it alone since this is fully internal to this file just left a note for the future.

README.md
```diff
@@ -89,6 +89,7 @@ bin/setup
 ```
 > ℹ️ **Note:** If `bundler` is not installing, ensure that you have `rbenv` installed and are not using the system Ruby version. Check the `.ruby-version` file in the repository to match the version specified. If necessary, update to the correct Ruby version and modify your `.zprofile` or `.zshrc` to point to the correct path.
 
+> ℹ️ **Note:** You may consider at this point copying or symlinking `.rspec-local.example` to `.rspec-local` to exclude specs that are not core to CfA yet. Copying enables you to set other custom flags for your local environment, symlinking enables the flags to automatically go away when they are removed from the example file. Just be careful not to commit any local changes!
```
powersurge360, line 92: Was unsure where to put this or how to indicate how to use it. I'd like for it to be automatic but it obviously can't be since different teams will have different needs going forward.

app/lib/schema_file_loader.rb
```diff
@@ -35,12 +37,14 @@ def download_and_unzip_schemas_from_s3(dest_dir)
...
       s3_client = Aws::S3::Client.new(region: REGION, credentials: s3_credentials)
-      get_missing_downloads(dest_dir).each do |download_path|
+      get_missing_downloads(dest_dir).each do |(download_path, optional)|
         s3_client.get_object(
           response_target: download_path,
           bucket: BUCKET,
           key: File.basename(download_path),
         )
+      rescue Aws::S3::Errors::NoSuchKey
+        next if optional
```
mrotondo, line 47: if you `raise if not optional`, then i think you can get rid of this explicit `next` call, since it doesn't skip any subsequent behavior inside the loop

app/services/state_file/state_information_service.rb
```diff
@@ -74,6 +74,7 @@ def state_code_to_name_map
         vita_link: "https://airtable.com/appnKuyQXMMCPSvVw/pag0hcyC6juDxamHo/form",
         voucher_form_name: "Form AZ-140V",
         voucher_path: "/pdfs/AZ-140V.pdf",
+        optional: false
```
mrotondo, line 77: no change necessary, but now that i look at this, i wonder if calling this key `schema_optional` might make it clearer how it's used

app/lib/schema_file_loader.rb
```diff
@@ -35,12 +37,14 @@ def download_and_unzip_schemas_from_s3(dest_dir)
...
     def download_schemas_from_s3(dest_dir)
       s3_client = Aws::S3::Client.new(region: REGION, credentials: s3_credentials)
-      get_missing_downloads(dest_dir).each do |download_path|
+      get_missing_downloads(dest_dir).each do |(download_path, optional)|
         s3_client.get_object(
           response_target: download_path,
           bucket: BUCKET,
           key: File.basename(download_path),
         )
+      rescue Aws::S3::Errors::NoSuchKey
```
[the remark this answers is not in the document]
powersurge360: It is the case. We originally had a begin, in fact, but the linter complained about it!

mrotondo (approved):
looks good!

PR #4748, opened 2024-09-06
Add mTLS certs for AZ, MD, NC, NJ, ID and Demo site

embarnard:
## Link to pivotal/JIRA issue
- e.g. [https://codeforamerica.atlassian.net/browse/FYST-496](https://codeforamerica.atlassian.net/browse/FYST-496)
## What was done?
- Added mTLS certs in credentials file for NC, NJ, ID and demo
## How to test?
- Go into the credentials file for production and demo and make sure its formatted correctly
- can use this command to see the difference between main and this branch's credentials files `bin/gyr credentials_diff --environment production --base main`


config/credentials/production.yml.enc
mpidcock, line 1: Did you also include Maryland? It's not in the PR title/description

embarnard: @mpidcock MD is in a separate PR

arinchoi03 (commented):
The prod file looks good, however, it looks like we have some extra whitespace at the end, could we remove those?

for the demo, I'm also seeing some whitespace & see that we only have `az` and `ny` -- is that what we want there?

mrotondo:
> The prod file looks good, however, it looks like we have some extra whitespace at the end, could we remove those?
> 
> for the demo, I'm also seeing some whitespace & see that we only have `az` and `ny` -- is that what we want there? I'm not certain what this part means in the story:
> 
> > A single demo key/cert pair should be generated (the .pem file that we send to DF can be named cfa.demo.pem) and the **base64-encoded contents should be added to the development, demo, staging and heroku credentials for all the same states as prod.**

Sorry for not being clear - I meant that in the development, demo, staging and heroku credentials, there should be entries for each state (az, id, md, nc, nj, ny) that contain the base64-contents of the `demo` private key & public certificate - i.e. they'll all have identical contents, other than the states which have an existing `account_id`.

arinchoi03:
looks great, just a couple notes:

`development` credentials have whitespace at the very end & also has the `google_oauth2` added -- was that intentional?

arinchoi03 (approved):
🚀

embarnard:
> looks great, just a couple notes:
> 
> `development` credentials have whitespace at the very end & also has the `google_oauth2` `client_id` and `client_secret` added -- was that intentional?

@arinchoi03 Oh oops! Removed whitespace, thanks for catching that. And the `google_oauth2` was already there I just moved it to be above the mtls creds instead of after so it matches the other files

PR #4749, opened 2024-09-07
r1 df api testing

mrotondo:


mrotondo:
DO NOT MERGE

app/models/state_file_nc_intake.rb
```diff
@@ -38,8 +38,12 @@
 #  primary_middle_initial            :string
 #  primary_suffix                    :string
 #  raw_direct_file_data              :text
+#  raw_direct_file_intake_data       :jsonb
+#  raw_direct_file_tax_return_data   :text
 #  referrer                          :string
 #  routing_number                    :integer
+#  sales_use_tax                     :decimal(12, 2)
+#  sales_use_tax_calculation_method  :integer          default(0), not null
```
arinchoi03, line 46: why are these added in this PR? also for below `untaxed_out_of_state_purchases`

mrotondo: they're just there because i didn't roll back the migration from my work on that branch before starting work on this branch. this diff should disappear once that branch is merged

app/models/state_file_az_intake.rb
```diff
@@ -52,6 +52,8 @@
 #  primary_was_incarcerated                    :integer          default("unfilled"), not null
 #  prior_last_names                            :string
 #  raw_direct_file_data                        :text
+#  raw_direct_file_intake_data                 :jsonb
+#  raw_direct_file_tax_return_data             :text
```
arinchoi03, line 56: we probably don't want this since we decided against renaming the column

arinchoi03 (commented):
Probably want another set of eyes other than mine but just noticed something

app/models/state_file_az_intake.rb
mrotondo: good catch - removed it everywhere

arinchoi03 (approved):
looks spiffy to me ✨

PR #4750, opened 2024-09-09
Partial refactor of direct_file_data class

jenny-heath:
## Link to pivotal/JIRA issue
- N/A, did this mostly with some spare minutes
## What was done?
- Chatted recently with Martha and Mike about how it would be nice to apply the DfXmlAccessor dynamic method creation to the df data class because it has so many repetitive xml readers and writers
- Started down a path of defining both readers/writers and quickly realized that our xml writing code is not that robust (esp if tag order matters) so this PR just turns some of the _readers_ into dynamically defined methods
- This required a small refactor of the DfXmlAccessor class so the reader and writer definitions are separate
## How to test?
- I added tests for all of the methods I "deleted". Unclear if they were actually tested anywhere—could be implicitly in the xml builders or something, but having it in the df data spec seems nice to me
- Risk Assessment
  - The main thing this could affect is building the state xml but we've got plenty of time to test that before the season begins

### DIFF
--- a/app/models/direct_file_data.rb
+++ b/app/models/direct_file_data.rb
@@ -16,7 +16,7 @@ class DirectFileData < DfXmlAccessor
     phone_number: 'ReturnHeader Filer PhoneNum',
     cell_phone_number: 'ReturnHeader AdditionalFilerInformation AtSubmissionFilingGrp CellPhoneNum',
     tax_payer_email: 'ReturnHeader AdditionalFilerInformation AtSubmissionFilingGrp EmailAddressTxt',
-    fed_tax: 'IRS1040 TotalTaxBeforeCrAndOthTaxesAmt',
+    fed_tax_amt: 'IRS1040 TotalTaxBeforeCrAndOthTaxesAmt',
     fed_agi: 'IRS1040 AdjustedGrossIncomeAmt',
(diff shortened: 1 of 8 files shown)


app/models/direct_file_data.rb
```diff
@@ -173,72 +135,40 @@ def surviving_spouse=(value)
...
   def mailing_zip=(value)
     write_df_xml_value(__method__, value)
   end
 
-  def fed_tax
-    df_xml_value(__method__)&.to_i || 0
-  end
-
-  def fed_tax=(value)
+  def fed_tax_amt=(value)
```
arinchoi03, line 171: how come `fed_tax_amt` from DirectFileData can be set? Most of the attributes are read-only here, so just trying to understand why some of them have setters/writers and some don't.

jenny-heath: so the setters are for this page that can only be accessed by our team in demo environments: right after the df data transfer there's a page where you can modify some of the data i.e. for acceptance testing purposes. i think which ones are editable is a liiiitle random (some we preemptively made setters for, others were requested by product, etc)

arinchoi03 (commented):
since this is a partial refactor, when are we scheduling in a "full" refactor?

PR #4751, opened 2024-09-10
Nj 37/38 add NJ1040 lines 6 and 7

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/37
- https://github.com/newjersey/affordability-pm/issues/38

## What was done?
- Added logic to the calculator, the XML submission builder, and the PDF filler to populate fields related to lines 6 and 7.
- Rebased on top of @aloverso's PDF and County Code changes, for easier PR review, switch the commits to the latest 3 commits

## How to test?
  - Ran unit tests for the NJ classes
  - Tested with the FYST local app, using both a single filer and the Zeus one dependent persona, verified the following:
    - XML showed correct values 
    - PDF generated correctly

https://github.com/user-attachments/assets/fde41d3a-268b-48ed-912b-3b4658c8d353

### DIFF
--- a/app/lib/efile/nj/nj1040.rb
+++ b/app/lib/efile/nj/nj1040.rb
@@ -3,8 +3,58 @@ module Efile
     class Nj1040 < ::Efile::TaxCalculator
       attr_reader :lines
 
+      def initialize(year:, intake:, include_source: false)
+        super
+      end
+
       def calculate
+        set_line(:NJ1040_LINE_6_SPOUSE, :calculate_line_6_spouse_checkbox)
+        set_line(:NJ1040_LINE_6, :calculate_line_6)
+        set_line(:NJ1040_LINE_7_SELF, :calculate_line_7_self_checkbox)
+        set_line(:NJ1040_LINE_7_SPOUSE, :calculate_line_7_spouse_checkbox)
+        set_line(:NJ1040_LINE_7, :calculate_line_7)
         @lines.transform_values(&:value)
+      end
+
+      def calculate_line_6_spouse_checkbox
+        @intake.filing_status_mfj?
+      end
+
+      def calculate_line_6
+        number_of_line_6_exemptions = 1 # defaults to 1 since self checkbox is always true
+        if calculate_line_6_spouse_checkbox
+          number_of_line_6_exemptions += 1
+        end
+        # DF data does not contain Domestic Partner information, so we are not supporting the Domestic Partner exemption
+        number_of_line_6_exemptions * 1_000
+      end
+
+      def calculate_line_7_self_checkbox
+        over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
+        return true if @intake.primary_birth_date <= Date.new(over_65_birth_year, 12, 31)
+        false
+      end
+
+      def calculate_line_7_spouse_checkbox
+        over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
+        return false unless @intake.spouse_birth_date.present?
+        return true if @intake.spouse_birth_date <= Date.new(over_65_birth_year, 12, 31)
+        false
+      end
+
+      def calculate_line_7
+        number_of_line_7_exemptions = 0
+        if calculate_line_7_self_checkbox
+          number_of_line_7_exemptions += 1
+        end
+        if calculate_line_7_spouse_checkbox
+          number_of_line_7_exemptions += 1
+        end
+        number_of_line_7_exemptions * 1_000
+      end
+
+      def total_exemption_amount
+        0
       end
 
       def refund_or_owed_amount
--- a/app/lib/efile/nj/nj1040.rb
+++ b/app/lib/efile/nj/nj1040.rb
@@ -8,48 +8,35 @@ module Efile
       end
 
       def calculate
-        set_line(:NJ1040_LINE_6_SPOUSE, :calculate_line_6_spouse_checkbox)
+        set_line(:NJ1040_LINE_6_SPOUSE, :line_6_spouse_checkbox)
         set_line(:NJ1040_LINE_6, :calculate_line_6)
-        set_line(:NJ1040_LINE_7_SELF, :calculate_line_7_self_checkbox)
-        set_line(:NJ1040_LINE_7_SPOUSE, :calculate_line_7_spouse_checkbox)
+        set_line(:NJ1040_LINE_7_SELF, :line_7_self_checkbox)
+        set_line(:NJ1040_LINE_7_SPOUSE, :line_7_spouse_checkbox)
         set_line(:NJ1040_LINE_7, :calculate_line_7)
         @lines.transform_values(&:value)
       end
 
-      def calculate_line_6_spouse_checkbox
+      def line_6_spouse_checkbox
         @intake.filing_status_mfj?
       end
 
       def calculate_line_6
-        number_of_line_6_exemptions = 1 # defaults to 1 since self checkbox is always true
-        if calculate_line_6_spouse_checkbox
-          number_of_line_6_exemptions += 1
-        end
-        # DF data does not contain Domestic Partner information, so we are not supporting the Domestic Partner exemption
+        self_exemption = 1
+        number_of_line_6_exemptions = self_exemption + number_of_true_checkboxes([line_6_spouse_checkbox])
         number_of_line_6_exemptions * 1_000
       end
 
-      def calculate_line_7_self_checkbox
-        over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
-        return true if @intake.primary_birth_date <= Date.new(over_65_birth_year, 12, 31)
-        false
+      def line_7_self_checkbox
+        is_over_65(@intake.primary_birth_date)
       end
 
-      def calculate_line_7_spouse_checkbox
-        over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
+      def line_7_spouse_checkbox
         return false unless @intake.spouse_birth_date.present?
-        return true if @intake.spouse_birth_date <= Date.new(over_65_birth_year, 12, 31)
-        false
+        is_over_65(@intake.spouse_birth_date)
       end
 
       def calculate_line_7
-        number_of_line_7_exemptions = 0
-        if calculate_line_7_self_checkbox
-          number_of_line_7_exemptions += 1
-        end
-        if calculate_line_7_spouse_checkbox
-          number_of_line_7_exemptions += 1
-        end
+        number_of_line_7_exemptions = number_of_true_checkboxes([line_7_self_checkbox, line_7_spouse_checkbox])
         number_of_line_7_exemptions * 1_000
       end
 
@@ -64,6 +51,17 @@ module Efile
       def analytics_attrs
         {
         }
+      end
+
+      private
+
+      def is_over_65(birth_date)
+        over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
+        birth_date <= Date.new(over_65_birth_year, 12, 31)
+      end
+
+      def number_of_true_checkboxes(checkbox_array_for_line)
+        checkbox_array_for_line.sum { |a| a == true ? 1 : 0 }
       end
     end
   end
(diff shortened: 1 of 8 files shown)


app/lib/efile/nj/nj1040.rb
```diff
@@ -3,10 +3,60 @@ module Nj
...
+      end
+
+      def calculate_line_7
+        number_of_line_7_exemptions = 0
+        if calculate_line_7_self_checkbox
+          number_of_line_7_exemptions += 1
+        end
+        if calculate_line_7_spouse_checkbox
+          number_of_line_7_exemptions += 1
+        end
```
aloverso, line 52: [sand] This if/else structure seems pretty wordy, especially if we ever have more than two options. We're also using the same structure in `calculate_line_6`. Could we consider extracting to a private method that counts the trues and sums the total?

If you pass a list of true/false values (the checkboxes) you can for example do something like 
```
checkboxes_for_line.sum{ |a| a==true ? 1 : 0 }
```

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -54,7 +54,19 @@ def hash_for_pdf
...
+
+        # line 6 exemptions
+        "Check Box39": @xml_document.at("Exemptions SpouseCuRegular")&.text == "X" ? "Yes" : "Off",
+        "Check Box40": "Off",
+        "Domestic": get_line_6_exemption_count,
+        "x  1000": get_line_6_exemption_count * 1000,
+
+        # line 7 exemptions
+        "Check Box41": @xml_document.at("Exemptions YouOver65")&.text == "X" ? "Yes" : "Off",
+        "Check Box42": @xml_document.at("Exemptions SpouseCuPartner65OrOver")&.text == "X" ? "Yes" : "Off",
```
aloverso, line 67: [pebble] Didn't realize the PDF value for a checkbox was "Yes" or "Off" hahaha. Can we extract this into a private utility function for getting the PDF checkbox value so that we're not repeating this ternary in many places?

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -54,7 +54,19 @@ def hash_for_pdf
...
         "CountyMunicipality Code See Table page 50": @xml_document.at("ReturnHeaderState Filer USAddress CityNm")&.text,  # city / town text field
         "State": @xml_document.at("ReturnHeaderState Filer USAddress StateAbbreviationCd")&.text,
-        "ZIP Code": @xml_document.at("ReturnHeaderState Filer USAddress ZIPCd")&.text
+        "ZIP Code": @xml_document.at("ReturnHeaderState Filer USAddress ZIPCd")&.text,
+
+        # line 6 exemptions
+        "Check Box39": @xml_document.at("Exemptions SpouseCuRegular")&.text == "X" ? "Yes" : "Off",
+        "Check Box40": "Off",
+        "Domestic": get_line_6_exemption_count,
+        "x  1000": get_line_6_exemption_count * 1000,
```
aloverso, line 63: [pebble/dust] For this and the `x 1000_2` below, can we get this value from the XML instead of redoing the calculation here?

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  context 'when filing status is single' do
+    it "sets line 6 to 1000" do
+      instance.calculate
+      expect(instance.lines[:NJ1040_LINE_6].value).to eq(1000)
+    end
+
+    context 'when filer is older than 65' do
+      before do
+        intake.primary_birth_date = Date.new(1900, 1, 1)
```
aloverso, line 20: [pebble/dust] what are the tradeoffs of changing the value on an existing intake as opposed to re-creating an intake with the desired field in the setup?

My gut says that I would always rather recreate using the factory because otherwise you open yourself up to issues regarding state leaking between tests and test order mattering to whether it passes/fails that can get really hard to debug and spot.

Was there a reason for this approach? Otherwise I'd prefer to recreate the intake rather than modify it

spec/factories/state_file_nj_intakes.rb
```diff
@@ -110,5 +111,21 @@
...
     end
+
+    trait :married do
+      filing_status { "married_filing_jointly" }
+      spouse_birth_date { Date.new(1990, 1, 1) }
+      spouse_ssn { "123456789" }
+    end
+
+    trait :primary_over_65 do
+      primary_birth_date { Date.new(1900, 1, 1) }
```
aloverso, line 122: [pebble] Here and in the other tests, I would recommend using a test year that is on the edge of 65: 1959. Otherwise you're not testing the "over 65" logic that you wrote and the tests would still pass even if we accidentally flubbed our logic in the code and made it so over 55 still gets checked or something.

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      it 'checks the self 65+ checkbox and sets line 7 to 1000' do
+        expect(instance.lines[:NJ1040_LINE_7_SELF].value).to eq(true)
+        expect(instance.lines[:NJ1040_LINE_7_SPOUSE].value).to eq(false)
+        expect(instance.lines[:NJ1040_LINE_7].value).to eq(1000)
+      end
+    end
+
+    context 'when filer is younger than 65' do
+      before do
+        intake.primary_birth_date = Date.new(2000, 1, 1)
```
aloverso, line 32: [pebble] same comment as above that 1960 should be the year under test for younger than 65.

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      end
+      it 'does not check the self 65+ checkbox and sets line 7 to 0' do
+        expect(instance.lines[:NJ1040_LINE_7_SELF].value).to eq(false)
+        expect(instance.lines[:NJ1040_LINE_7_SPOUSE].value).to eq(false)
+        expect(instance.lines[:NJ1040_LINE_7].value).to eq(0)
+      end
+    end
+  end
+
+  context 'when filing status is married filing jointly' do
```
aloverso, line 43: this set of tests from here below is really nice & clean and well done

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -92,6 +92,71 @@
...
     end
 
+    describe "exemptions" do
+      context "single filer" do
+        let(:submission) {
+          create :efile_submission, tax_return: nil, data_source: create(
+            :state_file_nj_intake,
+            primary_first_name: "Bert",
+            primary_last_name: "Muppet",
+            primary_middle_initial: "S",
```
aloverso, line 102: [pebble] these name lines don't seem relevant to what's being tested, can they be deleted? Here and in the tests below as well.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -92,6 +92,71 @@
...
+          expect(pdf_fields["x  1000"]).to eq "1000"
+        end
+
+        it "fills pdf with Line 7 fields" do
+          expect(pdf_fields["Check Box41"]).to eq "Off"
+          expect(pdf_fields["Check Box42"]).to eq "Off"
+          expect(pdf_fields["x  1000_2"]).to eq "0"
+          expect(pdf_fields["undefined_9"]).to eq "0"
+        end
+        context "primary over 65" do
```
aloverso, line 117: [nitpick] I prefer a line of whitespace in between context blocks / describe blocks / it blocks

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -84,7 +78,6 @@
         it "populates line 7 XML fields" do
           expect(xml.at("Exemptions YouOver65").text).to eq("")
           expect(xml.at("Exemptions SpouseCuPartner65OrOver").text).to eq("")
-          expect(xml.at("Exemptions Line7").text).to eq("0")
```
aloverso: Why are all the calculated value tests being removed from this file?

app/lib/pdf_filler/nj1040_pdf.rb
jachan: This value is not stored in the XML, only the calculator. Guidance from CfA was to recalculate this value here since they don't want us pulling PDF values from the calculator.

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
jachan: Line7 and Line6 are not fields supported in the XML Schema, only the checkboxes are.

app/lib/efile/nj/nj1040.rb
```diff
@@ -3,10 +3,60 @@ module Nj
...
       def calculate
+        set_line(:NJ1040_LINE_6_SPOUSE, :calculate_line_6_spouse_checkbox)
+        set_line(:NJ1040_LINE_6, :calculate_line_6)
+        set_line(:NJ1040_LINE_7_SELF, :calculate_line_7_self_checkbox)
+        set_line(:NJ1040_LINE_7_SPOUSE, :calculate_line_7_spouse_checkbox)
+        set_line(:NJ1040_LINE_7, :calculate_line_7)
         @lines.transform_values(&:value)
       end
 
+      def calculate_line_6_spouse_checkbox
```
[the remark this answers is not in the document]
jachan: I'm in favor of a naming convention that clarifies true/false. I was using the `checkbox` suffix for that. I see your perspective on `calculate` being misleading, happy to simplify to `line_6_spouse_checkbox`

app/lib/pdf_filler/nj1040_pdf.rb
jachan: Yeah "Yes" and "Off" is not my favorite either. Refactoring...

spec/lib/efile/nj/nj1040_spec.rb
jachan: Original concern was that if we chose 1957 for younger than 65, next year that would be OLDER than 65 and the test would break. I can fix this by making it dynamically generate the birthdate.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
jachan: ran rubocop, thanks for reminder

spec/lib/efile/nj/nj1040_spec.rb
aloverso: Hm. I'm not sure. It feels weird to just recreate the logic of our code in our tests (`current_year - 65` or whatever) but I also see your point about it breaking next year. 

I think I prefer the dynamic to the set-it-to-2000 approach, but I wonder if there's any other option. Hm.

spec/lib/efile/nj/nj1040_spec.rb
jachan: No reason! Didn't realize it was possible to define `primary_birth_date` in the `create(:state_file_nj_intake)` call. Thanks for demonstrating the pattern in your PR and advocating for it here.

spec/factories/state_file_nj_intakes.rb
```diff
@@ -105,5 +111,21 @@
...
+    end
+
+    trait :married_spouse_over_65 do
```
mmazanec22, line 125: [dust] should this be `married_filing_jointly` or `mfj` over 65 instead so it's clearer, since there's also a married filing separately possibility?

mmazanec22 (approved):

spec/factories/state_file_nj_intakes.rb
jachan: Great callout, thanks!

jachan:
Rebased on top of PDF and County Code commits on main

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -17,5 +17,77 @@
...
+
+      context "when filer is over 65" do
+        let(:intake) { create(:state_file_nj_intake, :primary_over_65) }
+        it "populates line 7 XML fields" do
+          expect(xml.at("Exemptions YouOver65").text).to eq("X")
+          expect(xml.at("Exemptions SpouseCuPartner65OrOver")).to eq(nil)
+        end
+      end
+      context "when filer is younger than 65" do
+        let(:intake) { create(:state_file_nj_intake) }
```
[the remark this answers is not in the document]
jachan: Thanks - was a holdover from when I was creating a trait for `:under_65`. Fixed!

PR #4753, opened 2024-09-10
Reduce refresh window when waiting to load IRS data

mpidcock:
## Link to pivotal/JIRA issue
- No story
## Is PM acceptance required?
## What was done?
- I reduced the refresh on the page with the loading spinner from 20 seconds to 5
- Settled on 5 because it is lower than what we had, but not so low that it would result in spamming the refresh
- See the previous PR here for the previous change
   - https://github.com/codeforamerica/vita-min/pull/3896
   - We are seeing the second scenario happen a majority of the time, resulting in the 20 second wait to refresh happening regularly
## How to test?
- Go though a return, set a timer before the data loading screen, and see that it only takes 5 seconds instead of the previous 20
This is the spinner in question:
image


mrotondo (approved):
5️⃣

PR #4754, opened 2024-09-10
update webpack-dev-server and shakapacker to allow transitive dependency on path-to-regexp 0.1.10, resolving dependabot high severity alert

mrotondo:


package.json
```diff
@@ -23,8 +23,9 @@
     "jquery-mask-plugin": "^1.14.16",
     "jquery-ui": "^1.13.2",
     "mini-css-extract-plugin": "^2.7.6",
+    "prop-types": "^15.7.2",
     "regenerator-runtime": "^0.13.7",
-    "shakapacker": "7.1.0",
+    "shakapacker": "7.2.3",
```
mpidcock, line 28: any reason to not have "^7.2.3"?

mpidcock (approved):
Fine how it is, but if we can avoid locking to a specific version that would be nice

package.json
mrotondo: The version derived from the yarnfile/package.json and the gemfile need to match, and I don't trust the process of their respective ~/^ operators to keep the two versions in sync. I don't know why the version was previously hardcoded, but I assume it was a similar reason.

mrotondo:
Note to self: don't merge this, it seems very broken on heroku. Issues with flipper_features not being a table in the DB, and with css/js loading?

PR #4757, opened 2024-09-10
Fyst 184 1099int subtraction

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-184
## What was done?
- The story specifies that the DF field should be a fake one for now because we don't have the real one (hence all the "fake" comments)
- Note: all of the AZ persona PDFs are changed because I had to regenerate them. Otherwise the persona_spec fails because the field `"28"` in the PDF is now `""` instead of `nil`
## How to test?
- I have 4 specs:
  - a feature spec for the federal info page
  - a unit spec for the az140 pdf filler
  - a unit spec for the az xml builder
  - a unit spec for the df data getter/setter even though it is currently "fake"
- A note for acceptance testers: Mike and I took a look at the output PDFs for the AZ personas and they look the same even though apparently there are some differences between the files. The tests also say that all the fields are filled in correctly but it would be good for someone to double check those (johnny, leslie, martha, rory) because we're expecting them to be accurate.


app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
```diff
@@ -107,6 +107,7 @@ def documents_wrapper
               end
               xml.AzAdjSubtotal calculated_fields.fetch(:AZ140_LINE_19)
               xml.Subtractions do
+                xml.IntUSObligations @submission.data_source.direct_file_data.interest_reported_amount if @submission.data_source.direct_file_data.interest_reported_amount > 0
```
mpidcock, line 110: There's a helper method called `add_non_zero_value` that could be used here

mpidcock (approved):
seems fine since we don't have any info about how/if we will get 1099-int data

PR #4758, opened 2024-09-11
Fix error on state file FAQ page by adding filing years to NJ

jenny-heath:
## Link to Sentry issue
- https://codeforamerica.sentry.io/issues/5827535300/?alert_rule_id=1046555&alert_type=issue&notification_uuid=a00e7a97-90ae-469c-9b42-8a110921010c&project=1880327&referrer=slack
## What was done?
- like the title says
- also:
  - refactored state info service and specs so that they test all the states
  - refactored the schema file loader because it doesn't seem like we need an attribute on the info service just for that but maybe i'm missing something?
## How to test?
- one reservation about the info service tests: it doesn't feel like we *need* to require all the methods be defined but it also felt like overkill to separate them into required and not required. open to suggestions!
- Before
image

- After
image

### DIFF
--- a/app/services/state_file/state_information_service.rb
+++ b/app/services/state_file/state_information_service.rb
@@ -47,10 +47,6 @@ module StateFile
       def state_intake_class_names
         @_state_intake_class_names ||= state_intake_classes.map(&:to_s).freeze
       end
-
-      def state_schema_file_names
-        @_state_schema_file_names ||= STATES_INFO.map { |_, attrs| attrs[:schema_file_name] }.freeze
-      end
     end
 
     private
@@ -76,7 +72,6 @@ module StateFile
         vita_link: "https://airtable.com/appnKuyQXMMCPSvVw/pag0hcyC6juDxamHo/form",
         voucher_form_name: "Form AZ-140V",
         voucher_path: "/pdfs/AZ-140V.pdf",
-        optional: false
       },
       nc: {
         intake_class: StateFileNcIntake,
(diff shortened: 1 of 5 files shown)


mrotondo (approved):

app/services/state_file/state_information_service.rb
```diff
@@ -101,23 +103,24 @@ def state_code_to_name_map
...
+                              "Processing Center<br/>" \
+                              "Trenton, NJ".html_safe,
+        pay_taxes_link: "https://www.nj.gov/treasury/taxation/payments-notices.shtml",
+        survey_link: "",
+        tax_payment_info_url: "https://www.nj.gov/treasury/taxation/payments-notices.shtml",
+        tax_payment_url: "https://www.nj.gov/treasury/taxation/payments-notices.shtml",
+        tax_refund_url: "https://www.tax.ny.gov/pit/file/refund.htm",
+        vita_link: "",
+        voucher_form_name: "NJ Voucher Form",
+        voucher_path: "/pdfs/it201v_1223.pdf", # TODO
```
jachan, line 116: [DUST] Is it possible to leave voucher_path commented? Or at least blank if not? Concerned about pointing to a nonexistent path.

jachan (commented):
Had a question about the NJ Voucher Path - otherwise looks good! Thanks for taking the initiative to help fix the FAQ page!

app/services/state_file/state_information_service.rb
aloverso: Or we could set it to an empty string?

aloverso (approved):

app/services/state_file/state_information_service.rb
jenny-heath: done!

PR #4759, opened 2024-09-11
Remove python dependency and update node version

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-95
## Is PM acceptance required?
## What was done?
- Removed python and python-opencv dependencies and the blur detection code & rake task that used them
- Removed long-unused pipx dependency
- Updated dev & prod node versions to 20 (from 18 and 16 respectively), since 20 is the currently active LTS version of node (see chart on https://nodejs.org/en/about/previous-releases)
- Did not restore jdk dependency to brew file - that can get resolved in another story
## How to test?
- Should just be a matter of making sure the site still works with some manual testing, as well as observing that all the automated tests pass.


jenny-heath (approved):
this approval means nothing looks wrong to me but given the infrequency of these kinds of tasks i'm not 100% confident i would know if something were missing from this (hopefully that makes sense)

PR #4760, opened 2024-09-11
update webpack-cli dependency to fix dev server crash, and update shakapacker options to avoid dev server warning

mrotondo:


jnf (approved):
lgtm

PR #4762, opened 2024-09-11
Fyst 328 Remove W2s from NC xml

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-328 kind of
## What was done?
- Realized during acceptance testing that W2s are actually not in the documents list in the NC schema, removing them for that reason
- Had to fix a failing spec: state_return_spec checks for the presence of W2s so I excluded NC from the list. Would appreciate any input on how that affects the readability of the test
## How to test?
- Acceptance test: fill out NC intake, see that xml contains no W2s

### DIFF
--- a/spec/lib/submission_builder/state_return_spec.rb
+++ b/spec/lib/submission_builder/state_return_spec.rb
@@ -1,8 +1,8 @@
 require 'rails_helper'
 
 describe SubmissionBuilder::StateReturn do
-  # NC does not include W2s in the state return xml
-  StateFile::StateInformationService.active_state_codes.excluding("nc").each do |state_code|
+  states_requiring_w2s = StateFile::StateInformationService.active_state_codes.excluding("nc")
+  states_requiring_w2s.each do |state_code|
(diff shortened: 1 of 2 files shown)


spec/lib/submission_builder/state_return_spec.rb
```diff
@@ -1,8 +1,9 @@
 require 'rails_helper'
 
 describe SubmissionBuilder::StateReturn do
-  StateFile::StateInformationService.active_state_codes.each do |state_code|
-    describe '.build', required_schema: state_code do
+  # NC does not include W2s in the state return xml
+  StateFile::StateInformationService.active_state_codes.excluding("nc").each do |state_code|
```
mrotondo, line 5: No change necessary, totally subjective - you could avoid the comment and maybe improve readability by saying `states_requiring_w2s = StateFile::StateInformationService.active_state_codes.excluding("nc")` and then iterating over that list

mrotondo (approved):

PR #4763, opened 2024-09-12
Skip nj xml validation

mrotondo:
## Link to pivotal/JIRA issue
- None
## Is PM acceptance required?
## What was done?
- Viewing XML on any servers without the NJ schemas has been crashing when trying to load the schemas for our validation. This change skips validation for all NJ return XMLs, and should be reverted when the NJ schemas are added to the application.


jnf (approved):
go forth and merge

PR #4764, opened 2024-09-12
Nj 32 easy fields

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/32

## What was done?
- Add tests for filing status XML
- Fix spouse xml generation and test
- Add filing status to PDF
- Light refactor to move NJ 1040 specific xml tests into documents folder

## How to test?
Walk through the tax filing process with different directfile data. The filing status on the PDF should change and the XML should have spouse and filing status filled out correctly.

filing_status


app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -20,45 +20,46 @@ def hash_for_pdf
...
-        "Your Social Security Number required": taxpayer_ssn[2],
-        "Text3": taxpayer_ssn[3],
-        "Text4": taxpayer_ssn[4],
-        "Text5": taxpayer_ssn[5],
-        "Text6": taxpayer_ssn[6],
-        "Text7": taxpayer_ssn[7],
-        "Text8": taxpayer_ssn[8],
+        undefined: taxpayer_ssn[0],
+        undefined_2: taxpayer_ssn[1],
+        'Your Social Security Number required': taxpayer_ssn[2],
```
aloverso, line 31: [dust] I thought CFA said they prefer double-quoted strings

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -197,5 +208,53 @@
         expect(pdf_fields["ZIP Code"]).to eq "08037"
       end
     end
+
+    describe "filing status" do
+      context "single" do
+        before do
+          submission.data_source.direct_file_data.filing_status = 1
+        end
+        it "checks the appropriate box" do
```
aloverso, line 217: [dust/pebble] cam we rename these tests to include the expectation? I like being able to read test names to know what behavior to expect, so when it just says "appropriate box" it means you have to actually read the expect line of the test itself to understand which box is appropriate. Could update to something like "checks the Choice1 box" just to be clearer and more readable

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -197,5 +208,53 @@
...
+          expect(pdf_fields["Group1"]).to eq "Choice4"
+        end
+      end
+
+      context "qualifying widow" do
+        before do
+          submission.data_source.direct_file_data.filing_status = 5
+          submission.data_source.direct_file_data.spouse_date_of_death = "2023-09-30"
```
aloverso, line 252: [pebble] Why this date as a magic string? Could we add a variable name for this to be more self-documenting? For example:
```
date_within_the_prior_year = "2023-09-30"
submission.data_source.direct_file_data.spouse_date_of_death = date_within_the_prior_year
```

aloverso, line 252: [pebble] Why this date as a magic string? Could we add a variable name for this to be more self-documenting? For example:
```
date_within_the_prior_year = "2023-09-30"
submission.data_source.direct_file_data.spouse_date_of_death = date_within_the_prior_year
```

spec/lib/pdf_filler/nj1040_pdf_spec.rb
aloverso, line 217: [dust/pebble] cam we rename these tests to include the expectation? I like being able to read test names to know what behavior to expect, so when it just says "appropriate box" it means you have to actually read the expect line of the test itself to understand which box is appropriate. Could update to something like "checks the Choice1 box" just to be clearer and more readable

app/lib/pdf_filler/nj1040_pdf.rb
aloverso, line 31: [dust] I thought CFA said they prefer double-quoted strings

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+describe SubmissionBuilder::Ty2024::States::Nj::Documents::Nj1040, required_schema: "nj" do
+  describe ".document" do
+    let(:intake) { create(:state_file_nj_intake, filing_status: "single", municipality_code: "0101") }
+    let(:submission) { create(:efile_submission, data_source: intake) }
+    let(:build_response) { described_class.build(submission, validate: false) }
+    let(:xml) { Nokogiri::XML::Document.parse(build_response.document.to_xml) }
+
+    context "single filer" do
+      it "correctly fills answers" do
+        expect(xml.document.at('FilingStatus Single').text).to eq "X"
```
aloverso, line 12: [pebble] 
- similar feedback to before, can the test name be more descriptive?
- for this and the tests below, can we exhaustively test that the other filing statuses are NOT "X"? These tests would all still pass if we changed the code to just set an X on every filing status.

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        expect(xml.document.at('FilingStatus MarriedCuPartFilingSeparate SpouseName MiddleInitial').text).to eq(intake.spouse.middle_initial)
+        expect(xml.document.at('FilingStatus MarriedCuPartFilingSeparate SpouseName LastName').text).to eq(intake.spouse.last_name)
+      end
+    end
+
+    context "qualifying widow/er filers" do
+      let(:intake) { create(:state_file_nj_intake, filing_status: "qualifying_widow") }
+      
+      before do
+        intake.direct_file_data.spouse_date_of_death = "2024-07-30"
```
aloverso, line 39: [pebble] same feedback about explaining date of death in a variable name

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    context "single filer" do
+      it "correctly fills answers" do
+        expect(xml.document.at('FilingStatus Single').text).to eq "X"
+      end
+    end
+
+    context "married filing jointly filers" do
+      let(:intake) { create(:state_file_nj_intake, filing_status: "married_filing_jointly") }
+      
+      it "correctly fills spouse-specific answers" do
```
aloverso, line 19: [pebble] copy paste error in test name

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    context "qualifying widow/er filers" do
+      let(:intake) { create(:state_file_nj_intake, filing_status: "qualifying_widow") }
+      
+      before do
+        intake.direct_file_data.spouse_date_of_death = "2024-07-30"
+      end
+      
+      it "correctly fills qualifying-widow-specific answers" do
+        expect(xml.document.at('FilingStatus QualWidOrWider QualWidOrWiderSurvCuPartner').text).to eq "X"
+        expect(xml.document.at('FilingStatus QualWidOrWider LastYear').text).to eq "X"
```
aloverso, line 44: [pebble] 
- Can we test exhaustively that "TwoYearPrior" is not checked?
- Can we also include a test for when the date is 2023, that TwoYearPrior is checked and LastYear is not
- Can we have a test to handle what happens if the date of death is greater than 2 years prior?

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -26,21 +30,22 @@ def document
...
                     when :qualifying_widow
                       yod = Date.parse(@submission.data_source.direct_file_data.spouse_date_of_death)&.strftime("%Y")
                       xml.QualWidOrWider do
                         xml.QualWidOrWiderSurvCuPartner 'X'
-                        if yod == Time.now.year - 1
+                        case yod
+                        when '2024'
                           xml.LastYear 'X'
-                        elsif yod == Time.now.year - 2
+                        when '2023'
```
aloverso, line 48: [dust] should we be worried about hardcoding these years from a future-proofing perspective?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -197,5 +208,53 @@
...
+      context "head of household" do
+        before do
+          submission.data_source.direct_file_data.filing_status = 4
+        end
+        it "checks the appropriate box" do
+          expect(pdf_fields["Group1"]).to eq "Choice4"
+        end
+      end
+
+      context "qualifying widow" do
```
aloverso, line 249: [dust/pebble] I don't see year of spouse death in the PDF part of this code, but I see it in the XML. Should that be in scope for this story if it's in the XML scope?

app/lib/pdf_filler/nj1040_pdf.rb
mmazanec22: Huh, this was a rubocop autofix thing, and when I tried to change it to double quotes I got `Style/QuotedSymbols: Prefer single-quoted symbols when you don't need string interpolation or special symbols.`

I wonder if the default rubocop config is not aligned with CfA preferred style and there's no override for this one in the local config?

spec/factories/state_file_nj_intakes.rb
```diff
@@ -103,6 +103,26 @@
       intake.raw_direct_file_data = intake.direct_file_data.to_s
     end
 
+
+    trait :mfs_complete_spouse do
```
jachan, line 107: [Pebble] introduced something similar in my commit - can we merge the two traits?

https://github.com/codeforamerica/vita-min/blob/main/spec/factories/state_file_nj_intakes.rb#L115

spec/lib/pdf_filler/nj1040_pdf_spec.rb
jachan: Also curious if it's worth dynamically generating this date using

`MultiTenantService.new(:statefile).current_tax_year - 1`

so it doesn't break next year.

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
jachan, line 1: [Pebble] We should figure out where we want the tests to live: this file or `documents/nj1040_spec.rb`. If we want to stay consistent with the rest of the states, we should make our asserts in `nj_return_xml_spec.rb` rather than in `documents/nj1040_spec.rb`. It'll be testing the parent XML that uses the NJ1040 rather than the child XML generator, but I think that's a good thing as it ensures nothing is overwritten before we submit the return.

mmazanec22: Looks like CfA is using the documents folder for document specific tests for NC now. I was thinking that a test file structure that parallels our application file structure will make it easier to find what we're looking for. Should we maybe add a test in nj_return_xml_spec that ensures it does not overwrite document info?

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -111,5 +112,17 @@ def format_no_last_name(first_name, middle_initial, suffix)
...
+    FILING_STATUS_OPTIONS = {
+      "Single" => "Choice1",
+      "MarriedCuPartFilingJoint" => "Choice2",
+      "MarriedCuPartFilingSeparate" => "Choice3",
+      "HeadOfHousehold" => "Choice4",
+      "QualWidOrWider" => "Choice5"
+    }
+
+    def filing_status
+      FILING_STATUS_OPTIONS[@xml_document.at('FilingStatus')&.children&.first&.name]
```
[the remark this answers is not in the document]
mmazanec22: Getting the name of the first child element of filing status, since a child element is created for the X or other info

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
jachan: I'm open to that! When you do the rebase onto main, we could move tests into the `documents/nj1040_spec`?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
mmazanec22: I started to test the greater than 2 years prior case, but realized that begs the question of what we should do with filing status in that case. Since the fed requirements for spouse death within last two years are the same, and we should match fed filing status in call cases, is this still a case worth testing for? If so, should we add eligibility offboarding after the data import for this and other potential data quality issues?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mmazanec22: Wow, yes, ty for catching!

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
mmazanec22: Marking resolved per team sync discussion

spec/fixtures/state_file/fed_return_xmls/2023/nj/head_of_household.xml
jenny-heath, line 1: fyi i'm not reviewing the fixtures but let me know if you think i should!

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -253,5 +295,77 @@
...
+            date_two_years_prior = "#{MultiTenantService.new(:statefile).current_tax_year - 1}-09-30"
+            submission.data_source.direct_file_data.spouse_date_of_death = date_two_years_prior
+          end
+  
+          it "checks the Choice5 box" do
+            expect(pdf_fields["Group1"]).to eq "Choice5"
+          end
+  
+          it "checks the two years prior spouse date of death" do
+            expect(pdf_fields["Group1qualwi5ab"]).to eq "0"
```
jenny-heath, line 365: [DUST] this says "checks the 2 years prior" box but is really just checking that the one year box is _not_ checked, so i'd update the description accordingly

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+        end
+      end
+
+      context "when filer is under 65 and spouse is over 65" do
+        let(:intake) { create(:state_file_nj_intake, :mfj_spouse_over_65) }
+        it "populates line 7 XML fields" do
+          expect(xml.at("Exemptions YouOver65")).to eq(nil)
+          expect(xml.at("Exemptions SpouseCuPartner65OrOver").text).to eq("X")
+        end
+      end
```
jenny-heath, line 93: [DUST] imo it might be overkill to test all 4 of these scenarios? i think it just needs one test each for the checked and unchecked case for each box. that said, i'm never really too fussed about more tests

spec/lib/pdf_filler/nj1040_pdf_spec.rb
jenny-heath: ohhh i've just read `nj1040_pdf.rb#spouse_death_year` and i understand now. nvm!

jenny-heath (approved):
looks good!

spec/fixtures/state_file/fed_return_xmls/2023/nj/head_of_household.xml
mmazanec22: Do you have any advice on generating these and validating them? I'm manually copying/pasting/editing for now and it feels like a matter of time before we run into issues with invalid directfile example data

PR #4765, opened 2024-09-13
Fyst 387 Add phone number to NC return header

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-387
## What was done?
- Adds the phone number to the return header; pulled from DF per the story
- This is the first state-divergent thing in the return header so would appreciate any input on the approach there
## How to test?
- Added a phone_number writer to df data and the ability to edit from the fed info page so acceptance testers can see whatever phone number they add show up in the xml
- unit specs for df data and return header
- feature spec for editing the df xml
image


app/models/direct_file_data.rb
```diff
@@ -764,6 +769,7 @@ def attributes
       mailing_apartment
       mailing_zip
       cell_phone_number
+      phone_number
```
jenny-heath, line 772: having to add this really got me in terms of adding the ability to edit via the fed info page. could we maybe generate this from `SELECTORS` instead?

mpidcock: that's a great idea

app/lib/submission_builder/return_header.rb
```diff
@@ -26,6 +26,9 @@ def document
             end
             xml.TaxpayerSSN @submission.data_source.primary.ssn if @submission.data_source.primary.ssn.present?
             xml.DateOfBirth date_type(@submission.data_source.primary.birth_date) if @submission.data_source.primary.birth_date.present?
+            if @submission.data_source.state_code == "nc" && @submission.data_source.direct_file_data.phone_number.present?
```
mpidcock, line 29: I'd really rather not do this. preferably we have the same return header for all states, but if not, I like your idea that we split out the states that are different to implement their own version

mpidcock (approved):

jenny-heath:
@mpidcock ready for re-review! i checked with tiffany and it should be okay to include the phone number in all the headers. also, since it's now present everywhere i had to regenerate the persona output pdf's

mpidcock (approved):

mpidcock (approved):

PR #4767, opened 2024-09-13
NJ 42 - add dependents name, birth year, SSN to XML and PDF

aloverso:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/42

## What was done?
- [Code] Add zeus_two_deps so that we can test multiple without going through 11 dependents
- Include `BirthYear` in XML output for each dependent
- Fill in first 4 dependents with name, SSN, and birth year in PDF

## How to test?
- Use `zeus_one_dep` to test output with 1 dependent
- Use `zeus_two_deps` to test output with 2 dependents
- Use `zeus_many_deps` to test output with 11 dependents (validates all 11 show on XML, only first 4 show on PDF) 
  
image

### DIFF
--- a/app/lib/pdf_filler/nj1040_pdf.rb
+++ b/app/lib/pdf_filler/nj1040_pdf.rb
@@ -59,88 +59,20 @@ module PdfFiller
       }
 
       dependents = get_dependents
-      if dependents[0]
-        dep0 = dependents[0]
-        name = format_name(dep0[:first_name], dep0[:last_name], dep0[:middle_initial], dep0[:suffix])
-        answers.merge!({
-          "Last Name First Name Middle Initial 1": name,
-          "undefined_18": dep0[:ssn][0],
-          "undefined_19": dep0[:ssn][1],
-          "undefined_20": dep0[:ssn][2],
-          "Text54": dep0[:ssn][3],
-          "Text55": dep0[:ssn][4],
-          "Text56": dep0[:ssn][5],
-          "Text57": dep0[:ssn][6],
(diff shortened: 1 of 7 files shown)


app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -57,6 +57,92 @@ def hash_for_pdf
         "undefined_9": get_line_7_exemption_count,
         "x  1000_2": get_line_7_exemption_count * 1000,
       }
+
+      dependents = get_dependents
```
mmazanec22, line 61: [dust] rubocop is underlining some of the funcs in this file with "Naming/AccessorMethodName: Do not prefix reader method names with `get_`." IMO this is fine, but we could decide to change it for consistency, or turn that rubocop rule off.

mmazanec22 (approved):

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -57,6 +57,92 @@ def hash_for_pdf
...
+      if dependents[0]
```
mrotondo, line 62: What are your thoughts on packaging this up into a function with a map from {dependent indices -> [various pdf field indices]}? I don't think a change is necessary, as the code here feels much more explicit & readable, and maybe the duplication is fine given that this code shouldn't change very much.

mrotondo (approved):
Looks good, just one question about approach in the PDF filler on duplication vs a less-explicit shared method

mrotondo (approved):

PR #4768, opened 2024-09-14
remove usage of removed optional flag on efile schemas

mrotondo:


app/lib/schema_file_loader.rb
```diff
@@ -36,14 +36,14 @@ def download_and_unzip_schemas_from_s3(dest_dir)
...
-      get_missing_downloads(dest_dir).each do |(download_path, _, optional)|
+      get_missing_downloads(dest_dir).each do |(download_path, _)|
         s3_client.get_object(
           response_target: download_path,
           bucket: BUCKET,
           key: File.basename(download_path),
         )
       rescue Aws::S3::Errors::NoSuchKey => e
-        raise e unless optional
+        raise e
```
powersurge360, line 46: Can delete all three lines, I think.

powersurge360 (approved):
Suggested a minor change that I think should be done but it's small enough I don't necessarily need to come back for another review, I think.

app/lib/schema_file_loader.rb
mrotondo: oh right, thanks 🤦

PR #4770, opened 2024-09-16
FYST-162 Custom money field & validation for up to 2 decimal places

arinchoi03:
## Link to pivotal/JIRA issue
[FYST-162](https://codeforamerica.atlassian.net/browse/FYST-162)
## What was done?

**`vita_min_money_field`**
- We added a form builder element for money inputs: with prefix `$` in front & a `0.00` placeholder
- Replace any cfa_input_field within `vita-min/app/views/state_file/questions` that takes in a money amount with this new `vita_min_money_field` 
- Alternatives considered: considered building from scratch, decided to conform to honeycrisp as much as possible in case we lose any built-in options considered by honeycrisp not knowing how we're currently using cfa_input_field across all kinds of money input scenarios. Another option could be to use javascript? but felt unnecessarily messy since honeycrisp already has an example of a prefix. 

**validations**
- "validating" the correct number of decimals in the controller instead of in the model since setting the attribute already conforms any value to the correct number of decimal digits
-  Alternatives considered:
1. adding a Form & pre-emptively catching the value before it gets changed? not sure if this would help since the assignment itself/not save (which leads to validation) is what’s causing the change in value
2. wait until we have a front-end implementation in place to not even allow clients to enter this number format, although from previous convo I believe that may take a long time to align on

## How to test?
  - looking at the changes on localhost/deployed env and enter whole numbers/2 decimal numbers/1 decimal number/no decimals
  - unit test on AzPublicSchoolContributionsController
  - added specs that were missing for Az322Contribution
- Risk Assessment
  - Changes were made in SCSS, tried to minimize impact but other inputs that are _not_ money fields should look same as before
  - Please test also the validation on contribution date (all three should be filled; otherwise warning)
- Before
Screenshot 2024-09-18 at 3 47 25 PM
- After
Screenshot 2024-09-18 at 3 46 08 PM

[FYST-162]: https://codeforamerica.atlassian.net/browse/FYST-162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/vita_min_form_builder.rb
```diff
@@ -278,6 +278,40 @@ def cfa_file_field(method, label_text, help_text: nil, options: {}, classes: [],
...
+      )
+
+    html_output = <<~HTML
+      <div class="form-group#{error_state(object, method)} money-input-form-group">
+      #{label_and_field_html}
+        #{errors_for(object, method)}
+      </div>
+    HTML
+    html_output.html_safe
+  end
```
arinchoi03, line 313: we copied the code for `cfa_input_field` from honeycrisp, but took out some of the stuff that we weren't needing. 

We did add the `placeholder` piece & the `money-input-group`, `money-input`, `money-input-form-group` classes to better target a wrapper.

spec/models/az322_contribution_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        az.assign_attributes(amount: 100.113)
+        az.valid?
+        expect(az.errors[:amount]).not_to be_empty
+      end
+    end
+
+    describe "#date_of_contribution_is_valid_date" do
+      it 'should validate presence' do
+        # TODO
```
arinchoi03, line 120: @tahsinaislam I wasn't sure how to write this test, you probably have more context on what that validation was expected to do

app/models/az322_contribution.rb
```diff
@@ -42,10 +42,6 @@ def set_date_of_contribution
     end
   end
 
-  def parse_date_params(year, month, day)
-    Date.new(year.to_i, month.to_i, day.to_i) rescue nil
-  end
-
```
arinchoi03, line 48: removed since putting in a `nil` year led to this returning a valid date. 

```
irb(main):002> Date.new(nil.to_i, 6, 20)
=> Sun, 20 Jun 0000
```

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -65,6 +67,11 @@ def destroy
 
       private
 
+      def has_valid_cents?
+        amount = az322_contribution_params[:amount]
+        amount.to_s.match?(/^(\d+)?\.?\d{0,2}$/)
```
arinchoi03, line 72: played around with this https://regex101.com/r/fUi2Wb/1 to figure out regex for what we want to consider "valid"
Screenshot 2024-09-20 at 2 12 41 PM

spec/models/az322_contribution_spec.rb
```diff
@@ -99,6 +99,30 @@
...
+
+        az.amount = '.12'
+        az.valid?
+        expect(az.errors[:amount]).to be_empty
+      end
+
+      it "should clean up format that leads to validation errors" do
+        az.amount = '1.'
+        az.valid?
+        expect(az.errors[:amount]).to be_empty
```
jenny-heath, line 124: should it also test what it changes the value to? (in addition to just not being invalid)

app/models/az322_contribution.rb
jenny-heath, line 1: does this also work for something like `az211_contribution.update(amount: "10.")`? looks like all the specs use "="

arinchoi03: ah...yes, I was just following the previous pattern of assigning by `=` but this also works for `update`

spec/models/az322_contribution_spec.rb
arinchoi03: done ❇️  https://github.com/codeforamerica/vita-min/pull/4770/commits/bd9ec35e1eadda8e63f5712d492bc5370e3a06a3

arinchoi03:
@jenny-heath @powersurge360 @mpidcock we discussed this via tuple and decided we can remove the validations on the money field, rely on the db rounding to hundredth place, and showing the rounded value in the review screen 😌

arinchoi03:
TODO: implement this for new [app/views/state_file/questions/nc_subtractions/edit.html.erb](https://github.com/codeforamerica/vita-min/pull/4781/files/682b1b541aa242e9ced4b5f01be038b361712726#diff-13976a88eb71be12d7f11cccc1557389ac8161a41c00993e6eee515faa07bbae)

arinchoi03:
Currently **blocked** until all the money columns are converted to decimal type

PR #4772, opened 2024-09-17
NJ 47 - add w2 wages sum to PDF and XML

aloverso:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/47

- Yes - don't merge until issue is accepted!

## What was done?
- Sum `StateWagesAmt` from all W2s from federal DF, and round to nearest whole number
- Output sum in NJ XML field `WagesSalariesTips`
- Output sum in Line 15 of PDF
- [code] update zeus_two_w2s and zeus_many_w2s to use more interesting wage numbers with decimals

## How to test?
- Use `zeus_minimal` for a user with no DF W2s to confirm XML are not filled in when no W2
- Use `zeus_two_w2s` to test with wages: `12345.67` + `50000`
- Use `zeus_many_w2s` to test with wages: `50000.33` + `50000.33` + `50000.33` + `50000.33`
  
image
image

### DIFF
--- a/app/lib/efile/nj/nj1040.rb
+++ b/app/lib/efile/nj/nj1040.rb
@@ -13,6 +13,7 @@ module Efile
         set_line(:NJ1040_LINE_7_SELF, :line_7_self_checkbox)
         set_line(:NJ1040_LINE_7_SPOUSE, :line_7_spouse_checkbox)
         set_line(:NJ1040_LINE_7, :calculate_line_7)
+        set_line(:NJ1040_LINE_15, :calculate_line_15)
         @lines.transform_values(&:value)
       end
 
@@ -54,6 +55,19 @@ module Efile
       end
 
       private
+
+      def calculate_line_15
+        if @direct_file_data.w2s.empty?
+          return -1
+        end
+
+        sum = 0
+        @direct_file_data.w2s.each do |w2|
+          state_wage = w2.node.at("W2StateLocalTaxGrp StateWagesAmt").text.to_f
+          sum += state_wage
+        end
+        sum.round
+      end
 
       def is_over_65(birth_date)
         over_65_birth_year = MultiTenantService.new(:statefile).current_tax_year - 65
--- a/app/models/state_file_nj_intake.rb
+++ b/app/models/state_file_nj_intake.rb
@@ -29,7 +29,6 @@
 #  last_sign_in_ip                   :inet
 #  locale                            :string           default("en")
 #  locked_at                         :datetime
-#  love_taxes                        :string
 #  message_tracker                   :jsonb
 #  municipality_code                 :string
 #  municipality_name                 :string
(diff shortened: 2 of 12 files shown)


app/models/state_file_nj_intake.rb
```diff
@@ -29,6 +29,7 @@
 #  last_sign_in_ip                   :inet
 #  locale                            :string           default("en")
 #  locked_at                         :datetime
+#  love_taxes                        :string
```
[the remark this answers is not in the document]
aloverso: 🤦

app/lib/efile/nj/nj1040.rb
```diff
@@ -55,6 +56,19 @@ def analytics_attrs
...
       private
 
+      def calculate_line_15
+        if @direct_file_data.w2s.empty?
+          return -1
+        end
+
+        sum = 0
+        @direct_file_data.w2s.each do |w2|
+          state_wage = w2.node.at("W2StateLocalTaxGrp StateWagesAmt").text.to_f
```
jenny-heath, line 66: [PEBBLE] you can use the accessor on `DfW2` here https://github.com/codeforamerica/vita-min/blob/main/app/models/df_w2_accessor.rb#L18

so it would look like `state_wage = w2.StateWagesAmt.to_f`

aloverso: Replied in slack, agreed we're not changing for now

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -402,5 +402,79 @@
         end
       end
     end
+
+    describe "line 15 wages" do
```
jachan, line 406: [SAND] Is it reasonable to try a test case where we have an overflow? e.g. W2 wages sum to 10000000000.00? I understand that someone earning that much would probably be screened out of our tool but I'm curious how our application would handle a case like that.

jachan (approved):
Looks good! Left a comment about an edge case (very big W2 sum).

embarnard:
This code all looked good to me, excellent work! Obviously there was a test that was failed when merged to main but doubtful we would have caught that since it didn't show in circle. 

No biggie but for future practice, I think its important to have an explicit acceptance from our side (meaning you get an actual approved status from github) before merging the PR even if it seems that someone else might have looked at the PR or left a comment. Sometimes we might start looking at a PR but get pulled away to other things but will usually try to get back to it by EOD. @aloverso @jachan

spec/lib/pdf_filler/nj1040_pdf_spec.rb
aloverso: Nope, no need! You can see a conversation about this question in the story, and also more context [in this slack thread](https://njcio.slack.com/archives/C07GYAQ4E1E/p1726515777378459)

PR #4774, opened 2024-09-17
Fyst 316/add nc county page

powersurge360:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-316
## What was done?
- Adds a county page for NC intakes
## How to test?
- Unit tests included

image
image
image


app/models/concerns/nc_residence_county_concern.rb
```diff
@@ -0,0 +1,53 @@
+# Handles the bits of the North Carolina state file intake that deals with the
+# county.
+module NcResidenceCountyConcern
+  extend ActiveSupport::Concern
+
+  COUNTIES = {
+    # This first one is unofficial and for our uses
+    "unfilled" => "Unfilled",
```
powersurge360, line 8: I chose this specifically because it is not a number. I didn't want to have an invalid `000` suddenly become valid and this makes that chance more remote.

app/models/concerns/nc_residence_county_concern.rb
```diff
@@ -0,0 +1,53 @@
+# Handles the bits of the North Carolina state file intake that deals with the
+# county.
+module NcResidenceCountyConcern
```
powersurge360, line 3: A little pre-emptive but I expect that our NCIntake is going to become huge over time so I went ahead and broke this bit out. I can move it into the intake instead if it is felt strongly.

app/controllers/state_file/questions/nc_county_controller.rb
```diff
@@ -0,0 +1,11 @@
+module StateFile
+  class Questions::NcCountyController < Questions::QuestionsController
+    include ReturnToReviewConcern
+    before_action :set_ivars
+
+    def set_ivars
```
embarnard, line 6: @powersurge360 What is ivars?

powersurge360: Rubyist [shorthand for instance variables](https://docs.ruby-lang.org/en/master/contributing/glossary_md.html#:~:text=Instance%20Variable.%20Refers%20to%20a%20Ruby%20instance%20variable%20like%20%E2%80%98@foo%60).

embarnard (approved):
This is amazing! Great work, I left some nitpicky comments but definitely not blockers 🎉

app/forms/state_file/nc_county_form.rb
```diff
@@ -0,0 +1,15 @@
...
+    set_attributes_for :intake,
+      :residence_county
+
+    validates :residence_county,
+      presence: true,
+      inclusion: { in: StateFileNcIntake::COUNTIES.keys }
+
+    def save
+      @intake.assign_attributes(attributes_for(:intake))
+      @intake.save
```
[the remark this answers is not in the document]
powersurge360: Might as well while I'm in there!

app/forms/state_file/nc_county_form.rb
```diff
@@ -0,0 +1,14 @@
+module StateFile
+  class NcCountyForm < QuestionsForm
+    set_attributes_for :intake,
+      :residence_county
+
+    validates :residence_county,
+      presence: true,
+      inclusion: { in: StateFileNcIntake::COUNTIES.keys }
```
arinchoi03, line 8: this is also a question that I generally have about validations -- if we have validations like this on the `Form` and the value eventually gets saved on the given `intake` record in the DB, there is no validation on the record (i.e. you're not adding the same validation on the `residence_county` on the `intake` model), is this our accepted pattern and what would be the reason for not adding that on the intake model?

powersurge360: I very much agree. I think with the validators being spread around it's pretty wonky. When I initially wrote this I actually included this validation in the concern and was going to keep the form 'dumb' to leverage the machinery in place to hook into the form objects. This is the first page I've implemented using form objects so I am unwilling to mount a strong case _right now_ but after I have a couple under my belt I would like to suggest something new.

I bet we could write an easy form adapter and then pass the validations in concerns to keep them nice and sliced up/segmented and minimize rewrite as well as reduce risks of the database getting into an inconsistent state. But again, I need more exposure before I can competently argue in any direction.

arinchoi03: ooh yeah great idea, ty for confirming I'm not alone here :)

PR #4780, opened 2024-09-18
Fix failing state_return_spec: use Nokogiri to pull expected value instead of DfW2 class method

jenny-heath:
## Link to pivotal/JIRA issue
N/A
## What was done?
- A [PR from NJ](https://github.com/codeforamerica/vita-min/pull/4772) uncovered an issue with this spec: the methods defined by DfXmlAccessor convert nodes that end with _amt to integers but when we construct W2s we just copy over the whole chunk of xml from df without converting the values. The spec compares the constructed xml values with the values pulled using the accessor class, so one is rounded and the other is not. #4772 uncovered it by adding a fixture that included a decimal value as one of the amounts.
- This PR pulls the expected value from the DF fixture without rounding it
- There is a bigger conversation to be had I think about the fact that all of our df xml accessor methods call `.to_i` but this step is just to fix main

### DIFF
--- a/spec/lib/submission_builder/state_return_spec.rb
+++ b/spec/lib/submission_builder/state_return_spec.rb
@@ -51,12 +51,12 @@ describe SubmissionBuilder::StateReturn do
 
               # w2 at index 0 remains the same
               w2_from_xml = xml.css('IRSW2')[0]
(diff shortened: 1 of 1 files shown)


embarnard (approved):
Nice catch!

PR #4782, opened 2024-09-19
Nj 49 line 40b

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/49

## What was done?
Add a question for homeowner or renter status. Add it to XML and PDF.

## How to test?
Go through the filing process, selecting any sample directfile data. You should see a question after county and municipality that asks whether you were a renter or a homeowner in the last tax year. The choice you select should show up in the XML and PDF submissions when you are done.

image
image
image
image


config/locales/en.yml
```diff
@@ -2517,6 +2517,13 @@ en:
         edit:
           county: County
           title: Select the county where you lived on December 31, %{filing_year}
+      nj_household_rent_own:
+        edit:
+          helper_description: If you own a mobile home that is located in a mobile home park, you are considered a tenant.
```
aloverso, line 2522: [dust] Why a helper description instead of in the tenant option like in the designs?

aloverso: Actually, in looking at the designs, seems like most of the content doesn't follow the Figma?
- title: You might be eligible for the New Jersey Property Tax credit or deduction!
- question label: In 2024, which were you?
- Option text is slightly different

app/models/state_file_nj_intake.rb
```diff
@@ -84,6 +85,8 @@ class StateFileNjIntake < StateFileBaseIntake
 
   encrypts :account_number, :routing_number, :raw_direct_file_data, :raw_direct_file_intake_data
 
+  enum household_rent_own: { unfilled: 0, rent: 1, own: 2, neither: 3 }, _prefix: :household_rent_own
```
aloverso, line 88: [pebble[ I believe "both" is also supposed to be an option according to the designs? We should at least capture that here in the model (and it looks like an option on the XML as well?)

config/locales/es.yml
```diff
@@ -2490,6 +2490,13 @@ es:
         edit:
           county: Condado
           title: Selecciona el condado donde viviste el 31 de diciembre de %{filing_year}
+      nj_household_rent_own:
+        edit:
+          helper_description: Si usted es dueño de una casa móvil que está ubicada en un parque de casas móviles, se le considera un inquilino.
+          homeowner: Propietario de vivienda en %{filing_year}
+          neither: Tampoco
+          tenant: Inquilino en %{filing_year}
+          title: "¿Cuál te describe mejor?"
```
aloverso, line 2499: [dust] double checking this was added as a line item to our spanish review ticket

spec/controllers/state_file/questions/nj_household_rent_own_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:intake) { create :state_file_nj_intake }
+  before do
+    sign_in intake
+  end
+
+  describe "#edit" do
+    render_views
+    it 'succeeds' do
+      get :edit
+      expect(response).to be_successful
```
aloverso, line 13: [pebble] is this test sufficient? other controller tests (looking at county and municipality) are also testing that the update function is in fact updating the correct intake params

spec/forms/state_file/nj_household_rent_own_form_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  let(:intake) { create :state_file_nj_intake }
+
+  describe "validations" do
+    let(:form) { described_class.new(intake, invalid_params) }
+
+    context "invalid params" do
+      context "all fields are required" do
+        let(:invalid_params) do
+          {
+            :household_rent_own => nil,
```
aloverso, line 13: [dust] this feels like a dumb question but isn't the value 0 when unfilled, not nil?

aloverso: same silly question below, when is the value `"rent"` vs when is it `1`?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -229,5 +229,43 @@
...
+      context "when taxpayer is a renter" do
+        let(:intake) { create(:state_file_nj_intake, :df_data_minimal, household_rent_own: 'rent',) }
+
+        it "adds a checked tenant element to property tax deduct or credit" do
+          expect(xml.at("PropertyTaxDeductOrCredit Tenant").text).to eq("X")
+        end
+
+        it "does not add a checked homeowner element to property tax deduct or credit" do
+          expect(xml.at("PropertyTaxDeductOrCredit Homeowner")).to eq(nil)
+        end
```
aloverso, line 243: [dust] if you want, it might be easier / more readable to combine these into a single it statement with 2 expect statements (here and below)

aloverso (approved):
Left a few pebbles around the text content and the "both" option - assuming those are addressed, looks good to go!

jachan (approved):
Hey Melanie, great work here! Left a question about the data modeling on neither vs unfilled.

config/locales/en.yml
mmazanec22: I followed the ticket acceptance criteria on this one. The design is still WIP, so there's a follow-up ticket coming soon to a GitHub project board near you

app/models/state_file_nj_intake.rb
mmazanec22: Re: both: same response as the other design question - those are WIP so I implemented the case described by the ticket acceptance criteria and followed the version of the design that matched those.

Re: neither vs unfilled: all versions of the design seemed to have this option, so I assumed it meant that we didn't want to check anything by default, and did want the user to definitively answer the question.

tahsinaislam (approved):

PR #4787, opened 2024-09-20
Nj 39 blind disabled exemption

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/39

## Is PM acceptance required? 
- Yes

## What was done?
- Updated DirectFileData class to allow checking for Primary and Spouse blind individually
- Updated calculator to store total exemption for primary/spouse
- Updated XML submission filler to call DirectFileData class to determine whether checkboxes should be filled
- Updated PDF to check XML for checkboxes and calculate total exemption

## How to test?
  - Unit tests
  - Manual Testing using the `blind_primary` and `blind_primary_and_spouse` personas

https://github.com/user-attachments/assets/7489e2ed-2208-433c-9b4a-0dc7b7dc7177


spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -21,33 +21,48 @@ def over_65_birth_year
...
-      let(:intake) do
-        create(:state_file_nj_intake,
-               primary_birth_date: Date.new(over_65_birth_year, 1, 1))
-      end
-
-      it 'checks the self 65+ checkbox and sets line 7 to 1000' do
-        expect(instance.lines[:NJ1040_LINE_7_SELF].value).to eq(true)
-        expect(instance.lines[:NJ1040_LINE_7_SPOUSE].value).to eq(false)
-        expect(instance.lines[:NJ1040_LINE_7].value).to eq(1000)
+    context "when filer is not blind" do
```
mmazanec22, line 24: [dust] This file might be more readable if we grouped related tests. Maybe with a parent context `blindness exemption` or something like that

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -144,6 +144,13 @@
           expect(pdf_fields["undefined_9"]).to eq "0"
         end
 
+        it "fills pdf with Line 8 fields" do
```
mmazanec22, line 147: [dust] Maybe we should specify the scenario when we expect these fields to be present but not checked

app/models/direct_file_data.rb
```diff
@@ -502,14 +504,33 @@ def fed_w2_state=(value)
...
+  end
+
+  def spouse_blind
+    create_or_destroy_df_xml_node(__method__, true, 'VirtualCurAcquiredDurTYInd')
+    write_df_xml_value(__method__, "X")
+  end
+
+  def is_primary_blind?
+    return false unless parsed_xml.at('PrimaryBlindInd')
+    true
```
aloverso, line 519: [dust/pebble] why this style over just `return parsed_xml.at('PrimaryBlindInd')`. I find the "false unless" logic pretty head-bending to read if we can avoid it

aloverso, line 519: [dust] can you explain this section? It feels like we have two functions here `primary_blind` and `is_primary_blind` based on two different pieces of data in the XML? Can you help me understand?

spec/lib/efile/nj/nj1040_spec.rb
aloverso: agree with this, and same for the older/younger than 65 ones as well. would be nice to group in a a `describe`. IMO describe is for groups of related tests and context is for specific situations, so "blindness exemption" should be a describe

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -21,33 +21,48 @@ def over_65_birth_year
...
+      it "sets line 8 deductions to 0" do
+        expect(instance.lines[:NJ1040_LINE_8].value).to eq(0)
       end
     end
 
-    context 'when filer is younger than 65' do
+    context "when filer is blind" do
```
aloverso, line 30: [pebble] this is the same exact test as "when filer is blind and spouse is not blind" I believe? That one is more specific wording-wise so I think this one should be removed

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -180,6 +202,30 @@
           expect(pdf_fields["x  1000_2"]).to eq "0"
           expect(pdf_fields["undefined_9"]).to eq "0"
         end
+
+        it "fills pdf with Line 8 fields" do
+          expect(pdf_fields["Check Box43"]).to eq "Off"
+          expect(pdf_fields["Check Box44"]).to eq "Off"
+          expect(pdf_fields["x  1000_3"]).to eq "0"
+          expect(pdf_fields["undefined_10"]).to eq "0"
+        end
```
aloverso, line 211: [pebble] I was about to make the comment that this is the same test as the one on line 147. Then I looked into why they both existed, and I see that they exist indepedently under single filer vs married filing jointly. I'm a little confused about this breakdown of test organization. I feel like I'd much prefer to see tested logic grouped by feature (by line) because that's how I expect to read the tests when I'm looking for documentation on a specific behavior. Having to jump up and down the test file for blindness or other features and understand that it's in a filing status context seems confusing to me. Could we refactor this test file to remove "single filer" and "married filing jointly" as top-level contexts altogether and rather, test for those states _within_ specific lines as they are applicable? The fact that this test is duplicated in both places to me shows that these filing statuses are not always applicable to the outcome (like in this case) so I would rather organize by the topic that is affecting the test outcome (the line number and the specific exemption).

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -32,6 +32,19 @@
         expect(xml.at("Exemptions DomesticPartnerRegular")).to eq(nil)
       end
 
+      it "populates line 8 XML fields" do
+        expect(xml.at("Exemptions YouBlindOrDisabled")).to eq(nil)
+        expect(xml.at("Exemptions SpouseCuPartnerBlindOrDisabled")).to eq(nil)
```
aloverso, line 37: [pebble] this sounds like a nitpick but I do think test clarity is important so pebble it is (sorry). I think "populates fields" doesn't match with the behavior of expecting these fields to be nil and not exist, can we update test name to match behavior?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -32,6 +32,19 @@
...
+      end
+
+      context "when filer is blind" do
+        let(:intake) { create(:state_file_nj_intake, :primary_blind) }
+        it "populates line 8 XML fields" do
```
aloverso, line 42: [pebble] for this and other tests in this file, can we update the test name to reflect the behavior? Since they all have the same exact test name, it doesn't give me any clues to what behavior to expect from reading it. Something more specific to each situation like "it populates line 8 XML for self" or similar might make sense

app/models/direct_file_data.rb
jachan: Yeah, thanks for the callout there, will fix. Had more complex logic originally that I thought warranted the guard statement but it's confusing now.

jachan: `primary_blind` is used as a setter to modify the XML. We only use it in `factories` to set the trait to true. The name is `primary_blind` to match the convention of the other setters in this class.

`is_primary_blind?` is the method actually used to determine the blindness of the filer.

spec/lib/efile/nj/nj1040_spec.rb
jachan: Got it. Refactoring tests based on feedback above, will change.

app/models/direct_file_data.rb
aloverso: @jachan why do they set different values? As in, if `is_primary_blind` is reading the value of `'PrimaryBlindInd'` in the XML, I would expect that this is the value to set if we want to indicate they are blind. Why does the trait setter use `'VirtualCurAcquiredDurTYInd'` - where does that come from?

jachan: The `'VirtualCurAcquiredDurTYInd'` arg in the `create_or_destroy_df_xml_node` method is the `after` arg. See https://github.com/codeforamerica/vita-min/blob/f3c0c353190a556f35c76c3ed7592a91a2b4af79/app/helpers/df_xml_crud_methods.rb#L6. 

The `after` arg specifies what the preceding node should be. In this case, we want to insert the `primary_blind` node (with node name fetched from the `selectors` dict in line 69 of this class) after the `VirtualCurAcquiredDurTYInd` node.

aloverso: @jachan ooh thanks for explaining. Huh! That is really unintuitive. Not your fault of course. I assume we can trust that the order of nodes doesn't change here?

jachan: Yeah! It is weird. Asked CfA about how they construct the XML samples they use, and they said that their PMs get it from the IRS DF folks, since there isn't a public schema for the DF data. Existing DF samples with `<PrimaryBlindInd>` always have this field after `<VirtualCurAcquiredDurTYInd>` so I put it there.

I don't think the order matters necessarily, just the hierarchy of where this node is placed. And it's important to make sure that the sibling node `<VirtualCurAcquiredDurTYInd>` always exists in the base XML document. @tahsinaislam can you speak to how we can guarantee that this insert doesn't ever fail because `<VirtualCurAcquiredDurTYInd>` doesn't exist?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
jachan: @aloverso does this logic apply to the XML tests as well? I see that they're organized by trait rather than by feature right now, I can refactor those tests to be feature-based.

tahsinaislam (approved):

app/models/direct_file_data.rb
jachan: Yep, the `PrimaryBlindInd` element in the XML only shows up if the applicant is blind. Otherwise it's missing and `write_df_xml_value` fails (I believe). A similar pattern with the `create_or_destroy_df_xml_node` method is used to set spouse information, which might also be missing (if the filer is single, for example). 

https://github.com/codeforamerica/vita-min/blob/main/app/models/direct_file_data.rb#L120

Where we differ from the existing pattern is that this value is boolean. In `mailing_street`, for example, we can just write the value, if `primary_blind` should be set, then we'll need to set it to `X` to indicate that this value is true. There's no real point in accepting a "value" here.:

```
  def mailing_street=(value)
  end
```

That's my current understanding of how this works based on my conversation with Tahsina a month or so ago. Happy to talk more about this and refactor if needed!

PR #4789, opened 2024-09-20
NJ 55 - Add taxpayer SSN and name to header

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/55

## What was done?
- Update PDF filler to take name and SSN and put in the header

## How to test?
  - Unit tests
  - Testing locally through the statefile app using the `married_filing_jointly` persona

- Before
image

- After
image
image
image
image

### DIFF
--- a/app/lib/pdf_filler/nj1040_pdf.rb
+++ b/app/lib/pdf_filler/nj1040_pdf.rb
@@ -20,6 +20,14 @@ module PdfFiller
       mfj_spouse_ssn = get_mfj_spouse_ssn
       mfs_spouse_ssn = get_mfs_spouse_ssn
       answers = {
+        # header
+        'Your Social Security Number': taxpayer_ssn.to_s,
+        'Your Social Security Number_2': taxpayer_ssn.to_s,
+        'Your Social Security Number_3': taxpayer_ssn.to_s,
+        'Names as shown on Form NJ1040': get_name,
+        'Names as shown on Form NJ1040_2': get_name,
+        'Names as shown on Form NJ1040_3': get_name,
+
         # county code
         CM4: county_code[1],
         CM3: county_code[2],
--- a/spec/lib/pdf_filler/nj1040_pdf_spec.rb
+++ b/spec/lib/pdf_filler/nj1040_pdf_spec.rb
@@ -38,6 +38,12 @@ RSpec.describe PdfFiller::Nj1040Pdf do
         )
       }
 
+      it 'sets taxpayer SSN fields on header' do
+        expect(pdf_fields["Your Social Security Number"]).to eq "123456789"
+        expect(pdf_fields["Your Social Security Number_2"]).to eq "123456789"
+        expect(pdf_fields["Your Social Security Number_3"]).to eq "123456789"
+      end
+
       it 'sets taxpayer SSN fields' do
         expect(pdf_fields["undefined"]).to eq "1"
         expect(pdf_fields["undefined_2"]).to eq "2"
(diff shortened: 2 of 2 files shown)


spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -273,8 +308,12 @@
               spouse_ssn: "123456789"
             )
           }
+
           it 'fills pdf with LastName FirstName & LastName FirstName' do
             expect(pdf_fields[name_field]).to eq "Lively Blake E & Reynolds Ryan"
+            expect(pdf_fields["Names as shown on Form NJ1040"]).to eq "Lively Blake E & Reynolds Ryan"
+            expect(pdf_fields["Names as shown on Form NJ1040_2"]).to eq "Lively Blake E & Reynolds Ryan"
+            expect(pdf_fields["Names as shown on Form NJ1040_3"]).to eq "Lively Blake E & Reynolds Ryan"
```
aloverso, line 316: [dust] In this test and the previous, if you wanted slightly less repetition and prevent copy-paste errors, you could add a local variable like `expected_name = ` instead of copying the expected name in every one.

aloverso (approved):
Nice, I really like how you added the name tests in with the existing ones instead of making new header tests. Makes it really clean

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -20,6 +20,14 @@ def hash_for_pdf
       mfj_spouse_ssn = get_mfj_spouse_ssn
       mfs_spouse_ssn = get_mfs_spouse_ssn
       answers = {
+        # header
+        'Your Social Security Number': taxpayer_ssn.to_s,
```
embarnard, line 24: is this header getting reflected in the XML as well? @jachan

jachan: Nope! Just in the PDF, it reuses existing XML data and puts it in the header of the PDF on pages 2-4.

jachan: Seeing that the Primary Name/SSN appears to be stored as part of the return header: https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/return_header.rb#L20. Testing shows that the XML doc generated by the application stores the name under:

```
<ReturnHeaderState>
  <Filer>
    <Primary>
      <TaxpayerName>
        <First Name>
        <Middle Initial>
        <Last Name>
        <Name Suffix>
```

To combine these fields for the PDF, we use the standard NJ format, which differs based on information provided: https://github.com/codeforamerica/vita-min/blob/main/spec/lib/pdf_filler/nj1040_pdf_spec.rb#L186

embarnard: @jachan Yes nice! Since its part of the XML in the return header XML I think it makes more sense to grab the fields from there like so `@xml_document.at('Primary TaxpayerSSN')&.text` or `@xml_document.at('Secondary TaxpayerSSN')&.text` (see `app/lib/pdf_filler/az140_pdf.rb`)

embarnard (approved):
just that one change, good work! 🎉

app/lib/pdf_filler/nj1040_pdf.rb
jachan: Yep! That's already getting done, `taxpayer_ssn = get_taxpayer_ssn` which sources from the XML itself https://github.com/codeforamerica/vita-min/blob/3f5ff0e4eab5c7ebd81a87deaf1bba9ea14d665a/app/lib/pdf_filler/nj1040_pdf.rb#L146

PR #4790, opened 2024-09-20
Nj 50 property taxes line 40a

aloverso:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/50

## What was done?
- Add controllers for two new pages, control visibility based on `household_rent_own` question. Add form, validations, and content for questions.
- Edit `household_rent_own` question and form to reset property_tax_paid and rent_paid on change, and edit review page version so it will allow for changing tax paid on change to tenancy status from review
- Calculate line 40a based on 18% of rule, and based on whether filing status is MFS
- Insert into PDF and XML

## How to test?
- Use `minimal` for testing non-MFS
- Use `married filing separately` for testing MFS
- Select "homeowner" to enter property taxes
- Select "renter" to enter rent paid

tenant page: image
---
homeowner page: image
---
review page: 
image
---
XML (18% of rent): 
image
---
PDF (18% of rent): 
image


spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -229,5 +229,55 @@
...
+
+        it "adds a checked tenant element to property tax deduct or credit" do
+          expect(xml.at("PropertyTaxDeductOrCredit Tenant").text).to eq("X")
+        end
+
+        it "does not add a checked homeowner element to property tax deduct or credit" do
+          expect(xml.at("PropertyTaxDeductOrCredit Homeowner")).to eq(nil)
+        end
+
+        it 'inserts property tax rent calculation on line 40a' do
```
jachan, line 245: [DUST] Should we explain how we get to 9778 here like we do in the calculator? Seems confusing to have a different value here other than `rent_paid`.

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -229,5 +229,55 @@
         end
       end
     end
+
+    describe "property tax - lines 40a and 40b" do
+      context "when taxpayer is a renter" do
```
jachan, line 234: [DUST] More of a testing best practices question - noticed that we're only testing MFS logic in the calculator, and nowhere else. To me this makes sense as the XML submission builder treats the values the same once they're saved to `calculated_fields`.

Is the takeaway here to not functionally test behavior in all scenarios for Calculator, XML, and PDF, but only in the classes that have relevant logic? (e.g. PDF filler and XML need to check a different box for homeowner vs renter, but not for MFS, it's just the calculated value that changes)

jachan (approved):
Looks good! Minor comments, mostly questions for my own understanding.

spec/controllers/state_file/questions/nj_household_rent_own_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe "#edit" do
+    render_views
+    it 'succeeds' do
+      get :edit
+      expect(response).to be_successful
+    end
+  end
+
+  def get_next_controller(intake)
```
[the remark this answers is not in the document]
aloverso: Good callout! Was doing a refactoring but ultimately just changed approach altogether, this is a remnant. Will remove

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
aloverso: Hm. I wound up not explaining it here because the process that calculates that one isn't under test here, the way it is in the calculator. I used the same thinking in the XML builder test. Neither the XML nor PDF controls the conversion of one number to another, so I don't really want the tests to explain that conversion because it might imply that this is the class you'd want to look at to change that logic.

I used the word "calculation" in the test name to indicate that the value is derived from the calculator. Any ideas on how to make that clearer?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
aloverso: When I'm doing TDD, it allows me to write exactly the right number of tests for the behavior under test. 

Under the TDD paradigm, there's no reason for me to write a test in the PDF or XML for the MFS scenario because it would not drive out in the PDF or XML classes themselves. I can't write a failing test to describe that. 

So yes, my roundabout answer is - there's no need to test unrelated functionality if I have a unit test of the XML or PDF behavior. TDD is nice because I don't have to actually _make that decision_ - I just follow the principle of writing a failing test first and if I can't write a failing test, then you could argue that behavior should not be under test here.

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
jachan: Got it. After our conversation about TDD below, I think I agree. "Calculation" is probably sufficient here - alternative would be something like "inserts property tax rent amount from calculator on line 40a", but I don't think it's necessary.

spec/controllers/state_file/questions/nj_household_rent_own_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+require "rails_helper"
+
+RSpec.describe StateFile::Questions::NjHouseholdRentOwnController do
+  let(:intake) { create :state_file_nj_intake }
+  before do
+    sign_in intake
```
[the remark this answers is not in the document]
aloverso: I think so... I was actually borrowing from other controller test code... maybe a CFA eng can answer this for sure?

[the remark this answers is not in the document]
tahsinaislam: I dont think sign_in intake instantiates the controller but you will need it though to simulate a user being signed in because some of your tests might redirect to the login page if the test does not find an authenticated user.

tahsinaislam (approved):

PR #4791, opened 2024-09-23
Fyst 462 implement form d 400 ctc lines for both pdf and xml pt 2

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-462
## What was done?
- Adds the rest of the lines from the story to the calculator, xml, and pdf for form d-400
- If you want to review by commit, I basically kept everything for each line in one commit
## How to test?
- Calculator tests actual different tax cases
- Xml tests for presence of node, mocking calc method
- Pdf tests for filled in field


spec/lib/efile/nc/d400_calculator_spec.rb
```diff
@@ -87,6 +87,43 @@
...
+  describe "Line 15: North Carolina Income Tax" do
+    context "value (line 14 (which = line 12b) * 0.045 rounded to the nearest dollar) is positive" do
+      it "returns the value rounded up" do
+        allow(instance).to receive(:calculate_line_12b).and_return 100
+
+        instance.calculate
+        expect(instance.lines[:NCD400_LINE_15].value).to eq 5
+      end
+
+      it "returns the value rounded down" do
```
embarnard, line 109: Im confused by the context in which it should be rounded down vs up, should these be in separate context blocks?

spec/lib/submission_builder/ty2024/states/nc/documents/d400_spec.rb
```diff
@@ -33,6 +37,10 @@
         expect(xml.document.at('FAGIPlusAdditions').text).to eq "10000"
         expect(xml.document.at('NCStandardDeduction').text).to eq standard_deduction.to_s
         expect(xml.document.at('NCAGIAddition').text).to eq nc_agi_addition.to_s
+        expect(xml.document.at('NCAGISubtraction').text).to eq nc_agi_subtraction.to_s
+        expect(xml.document.at('NCTaxableInc').text).to eq nc_agi_subtraction.to_s # TODO: validate assumption that we only have full year residents at this stage
```
embarnard, line 41: is this TODO for now or later?

embarnard (commented):
Nice work! just a few questions before approval

spec/lib/efile/nc/d400_calculator_spec.rb
jenny-heath: oh sorry this description is probably confusing. it just means that it's always rounded (so not floored or rounded up or anything). it might be overkill to include both examples (so in the first case the number is 4.5 which rounds to 5 and in the other case it's 2.25 which rounds to 2), my thought was just that i could have accidentally implemented it as `.to_i` instead of `.round` and a test that only has the rounding-down case would not have caught that?

spec/lib/submission_builder/ty2024/states/nc/documents/d400_spec.rb
jenny-heath: oops it's for now

jenny-heath: ok i checked and it is true. comment removed!

PR #4792, opened 2024-09-23
Move all logic to model instead of controller

arinchoi03:
Refactor to move all validations/clean up related to `amount`

Things I tried before arriving at this solution:
1. trying to add validation error in the writer `def amount=(value)`. This seems to get overwritten when validators run
2. using `validates_format_of`: the value has already been converted to a decimal(12, 2) value

### DIFF
--- a/app/controllers/state_file/questions/az_public_school_contributions_controller.rb
+++ b/app/controllers/state_file/questions/az_public_school_contributions_controller.rb
@@ -26,11 +26,10 @@ module StateFile
           return redirect_to next_path
         end
 
-        if @az322_contribution.valid? && has_valid_money_format?
+        if @az322_contribution.valid?
           @az322_contribution.save
           redirect_to action: :index, return_to_review: params[:return_to_review]
         else
-          @az322_contribution.errors.add(:amount, 'must be a valid dollar amount') unless has_valid_money_format?
           render :new
         end
       end
@@ -48,11 +47,10 @@ module StateFile
           return redirect_to action: :index, return_to_review: params[:return_to_review]
         end
 
-        if @az322_contribution.valid? && has_valid_money_format?
+        if @az322_contribution.valid?
           @az322_contribution.save
           redirect_to action: :index, return_to_review: params[:return_to_review]
         else
-          @az322_contribution.errors.add(:amount, 'must be a valid dollar amount') unless has_valid_money_format?
           render :edit
         end
       end
--- a/app/controllers/state_file/questions/az_public_school_contributions_controller.rb
+++ b/app/controllers/state_file/questions/az_public_school_contributions_controller.rb
@@ -65,11 +65,6 @@ module StateFile
 
       private
 
-      def has_valid_money_format?
-        amount = az322_contribution_params[:amount]
-        amount.to_s.match?(/^(\d+)?\.?\d{0,2}$/)
-      end
-
       def az322_contribution_params
         params.require(:az322_contribution).permit(
           :made_contribution,
--- a/app/models/az322_contribution.rb
+++ b/app/models/az322_contribution.rb
@@ -39,10 +39,7 @@ class Az322Contribution < ApplicationRecord
 
   # Custom validation for handling values before they are coerced into decimal(12, 2) type
   def amount=(value)
-    if value&.to_s&.match?(/^(\d+)?\.$/) # Remove trailing decimal. Ex: '10.'
-      value = value.chop
-    end
-    write_attribute :amount, value
+    write_attribute :amount, value&.to_s&.chomp('.') # Remove trailing decimal. Ex: '10.'
   end
 
   private
--- a/app/models/az322_contribution.rb
+++ b/app/models/az322_contribution.rb
@@ -29,6 +29,7 @@ class Az322Contribution < ApplicationRecord
   validates :ctds_code, presence: true, format: { with: /\A\d{9}\z/, message: -> (_object, _data) { I18n.t("validators.ctds_code") }}, if: -> { made_contribution == "yes" }
   validates :district_name, presence: true, if: -> { made_contribution == "yes" }
   validates :amount, presence: true, numericality: { greater_than: 0 }, if: -> { made_contribution == "yes" }
+  validate :amount_format
 
   validates :date_of_contribution,
             inclusion: {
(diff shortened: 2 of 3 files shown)


app/models/az322_contribution.rb
```diff
@@ -45,4 +46,13 @@ def amount=(value)
   end
 
   private
+
+  VALID_MONEY_REGEXP = /\A(\d+)?\.?\d{0,2}\z/.freeze
+  def amount_format
+    return true if amount.blank?
+
+    unless VALID_MONEY_REGEXP.match?(read_attribute_before_type_cast(:amount).to_s)
```
arinchoi03, line 54: `read_attribute_before_type_cast` because the amount gets coerced into decimal type & `to_s` because the regex matching can't happen on valid floats.

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -67,11 +65,6 @@ def destroy
 
       private
 
-      def has_valid_money_format?
-        amount = az322_contribution_params[:amount]
-        amount.to_s.match?(/^(\d+)?\.?\d{0,2}$/)
-      end
-
```
arinchoi03, line 74: there are some tests on the controller related to this validation which I left

powersurge360 (commented):
Why not rely on the regular typecasting of the value to determine validity instead of writing a regex?

I think if you have `validates :amount, presence: true` it would catch everything and cast correctly, excluding values with trailing decimal points (which would become `nil`).

arinchoi03:
@powersurge360 we want to show an error to the user if they enter more than 2 decimals, so the existing validation & type casting would just swallow that amount & return the typecast amount.

app/models/az322_contribution.rb
```diff
@@ -29,6 +29,7 @@ class Az322Contribution < ApplicationRecord
   validates :ctds_code, presence: true, format: { with: /\A\d{9}\z/, message: -> (_object, _data) { I18n.t("validators.ctds_code") }}, if: -> { made_contribution == "yes" }
   validates :district_name, presence: true, if: -> { made_contribution == "yes" }
   validates :amount, presence: true, numericality: { greater_than: 0 }, if: -> { made_contribution == "yes" }
+  validate :amount_format
```
powersurge360, line 32: Changing these two lines to:

```ruby
  validates :amount, presence: {message: 'must be a valid dollar amount'}, numericality: { greater_than: 0 }
  # validate :amount_format
```
Makes most tests pass. There are a couple of failures.

[One test fails](https://github.com/codeforamerica/vita-min/blob/be299371d99c8d4d258d3198b312150da3e629c6/spec/models/az322_contribution_spec.rb#L89) because the error message has changed (and is now "must be a valid dollar amount"), which is fine, imo.

[Another test fails](https://github.com/codeforamerica/vita-min/blob/be299371d99c8d4d258d3198b312150da3e629c6/spec/models/az322_contribution_spec.rb#L127-L131) because it rounds to the nearest hundreth, which I think is probably also fine.

[Notably, the test with the trailing decimal](https://github.com/codeforamerica/vita-min/blob/be299371d99c8d4d258d3198b312150da3e629c6/spec/models/az322_contribution_spec.rb#L121-L125) does _not_ fail, instead becoming value `1.0` instead.

jenny-heath: i'm not quite understanding your suggestion, @powersurge360? are you saying that the `VALID_MONEY_REGEXP` does the same thing as `numericality: { greater_than: 0 }`? even if it does i think it still should incorporate `read_attribute_before_type_cast` which i can't figure out how to do without adding a validator method...

jenny-heath (commented):
i think i like it but i'm also not understanding all of the comments

app/models/az322_contribution.rb
powersurge360: I am mistaken then! I did try `'100.'.match(<regex here>)` and it matched that as well. I think because the final bit after decimal and the decimal bit are not grouped together.

app/models/az322_contribution.rb
powersurge360: Yea, wanting to truncate vs allowing it to round in the way the `BigDecimal` does naturally is for sure the trickiest bit. The typecasting goes a long way towards making sure we're receiving valid numbers but if we're deviating things get wonky. Hrrrrrrm.

app/models/az322_contribution.rb
arinchoi03: oh, here i was trying to make sure `100.` is considered a "valid" money format (b/c that's what Product/Design wanted to consider valid) -- so the matching should be ok?

powersurge360: Oh I gotcha. Sorry getting turned around a little on requirements!

app/models/az322_contribution.rb
arinchoi03: >[Notably, the test with the trailing decimal](https://github.com/codeforamerica/vita-min/blob/be299371d99c8d4d258d3198b312150da3e629c6/spec/models/az322_contribution_spec.rb#L121-L125) does not fail, instead becoming value 1.0 instead.

this is b/c we take away the trailing decimal here:
>  def amount=(value)
    if value&.to_s&.match?(/^(\d+)?\.$/) # Remove trailing decimal. Ex: '10.'
      value = value.chop
    end
    write_attribute :amount, value
  end

outside of the validation. hope that clears things up!

arinchoi03:
@jenny-heath @powersurge360 
I think prod/design was pretty clear that they want _some_ sort of user error message shown when they enter more than 2 decimals, and this is the only way I can think of doing it without doing some javascript/front-end validation which is still under discussion.

I think we can:
1. (likely) Push back on the `'10.'` being a valid number input, if we think overriding the setter is janky
2. (unlikely) Push back on "no more than 2 decimal places" validation, ask for prioritization of front-end validations
3. do both validations, which means we apply this to all the money fields we have.

Would like to present this to the team after I merge this implementation to the PR this one is branched off from. Lmk if you disagree.

app/models/az322_contribution.rb
powersurge360: Speaking of, I see that's not part of this but you might consider dropping the regex and just doing `value = value.chomp('.')`. That will remove the trailing decimal if and only if it exists, otherwise returning the string unmodified. Or maybe as with side effects, if you prefer.

`value.chomp!('.')`

arinchoi03: ooh yes I had initially used this before but ran into issues but looks like I was just using this completely wrong

powersurge360: It modifies the value in place. What it returns is just whether it changed and what it changed to, the actual value changes in place without needing to reassign the variable.

PR #4796, opened 2024-09-25
Fyst 582 add Idaho to codebase

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-582
## What was done?
- Idaho was born
- You can review each step of the process (more or less) in each commit.
## How to test?
- I added a feature spec and updated any failing specs to pass
- Before:

- After:
image
image
image


db/migrate/20240924213820_create_state_file_id_intakes.rb
```diff
@@ -0,0 +1,64 @@
...
+      t.integer :withdraw_amount
+      t.date :date_electronic_withdrawal
+
+      # federal fields
+      t.text :raw_direct_file_data
+      t.datetime :df_data_imported_at
+      t.datetime :df_data_import_failed_at
+      t.string :federal_submission_id
+      t.string :federal_return_status
+      t.string :hashed_ssn # also add an index (algorithm: :concurrently)
```
arinchoi03, line 64: is this a to-do for later?

spec/lib/submission_builder/ty2024/states/id/documents/id40_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+describe SubmissionBuilder::Ty2024::States::Id::Documents::Id40, required_schema: "id" do
+  describe ".document" do
+    let(:intake) { create(:state_file_id_intake, filing_status: "single") }
+    let(:submission) { create(:efile_submission, data_source: intake) }
+    let(:build_response) { described_class.build(submission, validate: false) }
```
arinchoi03, line 7: b/c `validate` is false, wouldn't the response always be a valid xml? Is this intentional b/c the fields haven't been filled out yet?

jenny-heath: true! it's mostly empty so at this point will never be valid. actually i just copied the md equivalent for this and didn't realize that the test itself expects it to be "valid" haha—will modify it to make a bit more sense

tahsinaislam (approved):

PR #4799, opened 2024-09-26
Enable InsightAppSec security scans

ddefigueiredo:
*Please forgive my lack of process here. This is my first pull request for this repo, so I do not know what I am doing.*
Please advise what I need to do. This is a minor change to allow us to scan the app with a security scanner we want to evaluate. 

## Link to pivotal/JIRA issue
- I do not have a JIRA issue for this change. Please let me know if I need to create one.
## Is PM acceptance required? 
- I only added a `meta` tag
## What was done?
- Added a small `meta` entry to enable the vendor to validate that we own the app. The logic is not environment agnostic, but it does not seem we are adopting that practice. Do let me know otherwise.
## How to test?
- This is such a small change I did not write a test for it and just tested it locally. Please let me know if I should do so.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -35,7 +35,7 @@
     <meta property="twitter:url" content="<%= canonical_url %>">
 
     <% # Site verification for InsightAppSec DAST vulnerability scanner %>
-    <% if not Rails.env.production? %>
+    <% unless Rails.env.production? %>
       <meta name="insight-app-sec-validation" content="a08ae307-c25f-4fc4-9dff-0c4ae0c7dff4">
     <% end %>
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -34,6 +34,11 @@
     <meta property="twitter:card" content="summary_large_image">
     <meta property="twitter:url" content="<%= canonical_url %>">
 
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -34,6 +34,11 @@
     <meta property="twitter:card" content="summary_large_image">
     <meta property="twitter:url" content="<%= canonical_url %>">
 
+    <% # Site verification for InsightAppSec DAST vulnerability scanner %>
+    <% if not Rails.env.production? %>
```
jenny-heath, line 38:
```suggestion
    <% unless Rails.env.production? %>
```
so we don't actually follow any [ruby style guides](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#if-vs-unless) _but_ we also don't have any instances of `if not` and tend to use `unless` around the codebase

jenny-heath:
> The logic is not environment agnostic, but it does not seem we are adopting that practice. Do let me know otherwise.

i'm not sure what this means. could you elaborate? it looks like you want to have the meta tag only in non-prod environments (so, development, demo, test, staging, heroku), is that correct?

ddefigueiredo:
That's right, I just want to add this `meta` tag to non-prod environments. That's all I need.

jenny-heath (approved):

PR #4800, opened 2024-09-26
Remove tax_return_year from md, nc, nj intake tables (should be delegated to direct_file_data)

jenny-heath:
## Link to pivotal/JIRA issue
- N/A was gonna make a story but this literally took 5 minutes 
## What was done?
- i had included `tax_return_year` on my initial migration for "Warshington" state in that spike PR because i was moving fast and didn't realize A) that value is delegated to `direct_file_data` and B) that my PR would be used as a template in the future
## How to test?
- i elected not to do a strong migration here because none of this stuff is in prod and, because the delegation to direct file data happens on the base class, any references to the column should be covered.


tahsinaislam (approved):

jachan (approved):
Hi! Looks good!

[DUST] General question separate from this PR but tied to `tax_return_year`.

We're using return_year [here](https://github.com/codeforamerica/vita-min/blob/main/app/lib/efile/nj/nj1040.rb#L99) to determine what year the over 65 exemption should be.

It looks like `tax_return_year` is defined in `intake`, `direct_file_data`, and the `MultiTenantService`. If we remove this from `intake`, should we also remove from `MultiTenantService` or modify that to source from `direct_file_data` rather than the Rails config file that it currently reads from?

jenny-heath:
@jachan you know, that's an interesting question. i guess the only time i would expect the two to differ is for last year's intakes, but i'm not sure we're even going to keep those in the database. and for any current intakes i would assume that the year coming from DF is _always_ the current tax year. let me confer with the team and get back to you!

PR #4803, opened 2024-09-26
nj 35 - added lines 13 and 30 logic

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/35#issue-2503540877

## What was done?
- Sum lines 6-8 and store value in `calculated_fields`, XML fields for 13/30, and the PDF

## How to test?
- Manual Testing in Heroku, selected a variety of filing situations from Minimal to Blind Primary and Spouse, validated that PDF and XML were populated with the right values.
- 

image
image

image
image

### DIFF
--- a/app/lib/efile/nj/nj1040.rb
+++ b/app/lib/efile/nj/nj1040.rb
@@ -79,7 +79,7 @@ module Efile
       end
 
       def calculate_line_13
-        calculate_line_6 + calculate_line_7 + calculate_line_8
+        line_or_zero(:NJ1040_LINE_6) +  line_or_zero(:NJ1040_LINE_7) +  line_or_zero(:NJ1040_LINE_8) 
       end
 
       def calculate_line_15
--- a/app/lib/efile/nj/nj1040.rb
+++ b/app/lib/efile/nj/nj1040.rb
@@ -16,11 +16,22 @@ module Efile
         set_line(:NJ1040_LINE_7_SPOUSE, :line_7_spouse_checkbox)
         set_line(:NJ1040_LINE_7, :calculate_line_7)
         set_line(:NJ1040_LINE_8, :calculate_line_8)
+        set_line(:NJ1040_LINE_13, :calculate_line_13)
         set_line(:NJ1040_LINE_15, :calculate_line_15)
         set_line(:NJ1040_LINE_40A, :calculate_line_40a)
         @lines.transform_values(&:value)
       end
 
+      def refund_or_owed_amount
+        0
+      end
+
+      def analytics_attrs
+        {
+        }
+      end
+
+      private
       def line_6_spouse_checkbox
         @intake.filing_status_mfj?
       end
@@ -51,7 +62,7 @@ module Efile
                                                                  @direct_file_data.is_spouse_blind?])
         number_of_line_8_exemptions * 1_000
       end
-      
+
       def calculate_line_40a
         is_mfs = @intake.filing_status == :married_filing_separately
 
@@ -67,20 +78,9 @@ module Efile
         is_mfs ? (property_tax_paid / 2.0).round : property_tax_paid.round
       end
 
-      def total_exemption_amount
-        0
+      def calculate_line_13
+        calculate_line_6 + calculate_line_7 + calculate_line_8
       end
-
-      def refund_or_owed_amount
-        0
-      end
-
-      def analytics_attrs
-        {
-        }
-      end
-
-      private
 
       def calculate_line_15
         if @direct_file_data.w2s.empty?
--- a/spec/lib/efile/nj/nj1040_spec.rb
+++ b/spec/lib/efile/nj/nj1040_spec.rb
@@ -105,15 +105,15 @@ describe Efile::Nj::Nj1040 do
   end
 
   describe 'line 13 - total exemptions' do
-    let(:intake) { create(:state_file_nj_intake) }
-    it 'sets line 13 to the sum of lines 6-12' do
-      self_exemption = 1000
+    let(:intake) { create(:state_file_nj_intake, :primary_over_65, :primary_blind) }
+    it 'sets line 13 to the sum of lines 6-8' do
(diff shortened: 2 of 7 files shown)


spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -104,6 +104,19 @@ def over_65_birth_year
     end
   end
 
+  describe 'line 13 - total exemptions' do
+    let(:intake) { create(:state_file_nj_intake) }
+    it 'sets line 13 to the sum of lines 6-12' do
```
aloverso, line 109: [pebble] might suggest that it should say the sum of lines 6-8, given that's what has been coded. Other lines are not being included in the sum nor tested yet

aloverso, line 109: [pebble] might suggest that it should say the sum of lines 6-8, given that's what has been coded. Other lines are not being included in the sum nor tested yet

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -104,6 +104,19 @@ def over_65_birth_year
...
+      self_exemption = 1000
+      expect(instance.lines[:NJ1040_LINE_6].value).to eq(self_exemption)
+      not_over_65 = 0
+      expect(instance.lines[:NJ1040_LINE_7].value).to eq(not_over_65)
+      not_blind = 0
+      expect(instance.lines[:NJ1040_LINE_8].value).to eq(not_blind)
+      expect(instance.lines[:NJ1040_LINE_13].value).to eq(self_exemption + not_over_65 + not_blind)
```
aloverso, line 116: [pebble] Could we adjust the input for this test to have values in all three fields that get summed? Right now, since two of them are zero, I could go and change the calculate_line_13 method to just sum the value of line 6 alone, and the test would still pass.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -676,6 +676,39 @@
       end
     end
 
+    describe "line 13/30 total exemptions" do
+      let(:submission) {
+        create :efile_submission, tax_return: nil, data_source: create(
+          :state_file_nj_intake
+        )
+      }
+      it "totals line 6-12 and writes it to line 13" do
```
aloverso, line 685: [pebbles] same comment about "lines 6-12"

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -267,6 +267,18 @@
       end
     end
 
+    describe "total exemption - lines 13 and 30" do
+      let(:intake) { create(:state_file_nj_intake) }
+      it "totals lines 6-12 and stores the result" do
```
aloverso, line 272: [pebble] 
- same comment about "lines 6-12"
- I might suggest that the test name should indicate in some way that the result gets stored in two places in the XML, like "and stores the result in both TotalExemptionAmountA and TotalExemptionAmountB"

spec/lib/efile/nj/nj1040_spec.rb
aloverso: Specifically I would recommend it just for this test - no need to do this change for the XML and PDF tests since they are not actually testing the summing behavior.

aloverso (commented):
Looks good! Just a couple small notes on test naming and test values.

spec/lib/efile/nj/nj1040_spec.rb
jachan: That makes sense, good catch.

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -104,6 +104,19 @@ def over_65_birth_year
     end
   end
 
+  describe 'line 13 - total exemptions' do
+    let(:intake) { create(:state_file_nj_intake, :primary_over_65, :primary_blind) }
+    it 'sets line 13 to the sum of lines 6-8' do
+      self_exemption = 1_000
```
mmazanec22, line 110: TIL you can put underscores in numbers in Ruby!

mmazanec22 (approved):
Looks great!

app/lib/efile/nj/nj1040.rb
```diff
@@ -67,21 +78,10 @@ def calculate_line_40a
         is_mfs ? (property_tax_paid / 2.0).round : property_tax_paid.round
       end
 
-      def total_exemption_amount
-        0
+      def calculate_line_13
+        calculate_line_6 + calculate_line_7 + calculate_line_8
```
embarnard, line 82: [pebble] since some of the calculations rely on each other and need to be done in a procedural fashion we follow the practice of only calling the calculate methods when we are setting the lines in the calculate method at the top and then calling the set line everywhere else so that it would look like this: 
```suggestion
        line_or_zero(:NJ1040_LINE_6) +  line_or_zero(:NJ1040_LINE_7) +  line_or_zero(:NJ1040_LINE_8) 
```

embarnard (changes requested):
Looks great! just that one requested change

app/lib/efile/nj/nj1040.rb
jachan: Thanks for letting me know about this! Can you elaborate on the danger of leaving this as a method call? Happy to change, just trying to understand this hazard better.

embarnard (approved):

app/lib/efile/nj/nj1040.rb
embarnard: Yeah I wish I could find the example of where we ran into this problem years ago (maybe I'll look for it if it doesn't chew up too much time and send it over) but basically we've run into problems where calling a calculation multiple times would change its value which is why we only call it once and run through the lines in order. I think there could also potentially be infinite loop/recursion issues hypothetically since lines depend on each other.

PR #4804, opened 2024-09-27
Nj 48 property tax eligibility

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/48

## What was done?
- Add ineligible/unsupported pages
- Add "both" option to rent/own page that navigates to Unsupported
- "neither" option now routes to Ineligible
- Add new homeowner eligibility checkboxes screen that routes to either Property Tax, Ineligible, or Unsupported depending on logic

## How to test?
use `minimal`

Note: this recording shows the incorrect logic for the combo of multi-unit yes and multi-unit-max-four yes - it should give a soft no, not an advance. That's been fixed in the PR but I don't want to re-record the whole thing

https://github.com/user-attachments/assets/1e17409d-1258-4cb4-a6bd-8495097b5ecd


app/views/state_file/questions/nj_homeowner_eligibility/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <p><%= t(".label") %></p>
+
+    <div class="tight-checkboxes spacing-above-0 question-with-follow-up">
+      <%= f.cfa_checkbox(:homeowner_home_subject_to_property_taxes, t(".homeowner_home_subject_to_property_taxes"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:homeowner_more_than_one_main_home_in_nj, t(".homeowner_more_than_one_main_home_in_nj"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:homeowner_shared_ownership_not_spouse, t(".homeowner_shared_ownership_not_spouse"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <div class="question-with-follow-up__question">
+        <%= f.cfa_checkbox(:homeowner_main_home_multi_unit, t(".homeowner_main_home_multi_unit"), options: { checked_value: "yes", unchecked_value: "no", "data-follow-up": "#homeowner_multi_unit_followup"}) %>
+      </div>
+      <div class="question-with-follow-up__follow-up" id="homeowner_multi_unit_followup">
```
mmazanec22, line 18: [pebble] This will need some ARIA markup to meet [WCAG 4.1.2, which mandates that users be notified of changes](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html).

[The UK digital services team has put a lot of care and research into this topic.](https://accessibility.blog.gov.uk/2021/09/21/an-update-on-the-accessibility-of-conditionally-revealed-questions/) The [TLDR in their component lib seems to be](https://design-system.service.gov.uk/components/radios/) that associating one input is fine (though I'm a little skeptical of our choice to not visually differentiate it) and we should use `aria-controls`.

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      def self.show?(intake)
+        intake.household_rent_own == 'own'
+      end
+
+      def next_path
+        options = {}
+        options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
+
+        is_hard_no = (current_intake.homeowner_home_subject_to_property_taxes == 'no' or (current_intake.homeowner_main_home_multi_unit == 'yes' and current_intake.homeowner_main_home_multi_unit_max_four_one_commercial == 'no'))
+        is_soft_no = (current_intake.homeowner_more_than_one_main_home_in_nj == 'yes' or current_intake.homeowner_shared_ownership_not_spouse == 'yes' or (current_intake.homeowner_main_home_multi_unit == 'yes' and current_intake.homeowner_main_home_multi_unit_max_four_one_commercial == 'yes'))
```
aloverso, line 17: This logic gets repeated in 3 different components. I want to externalize it so that it's shared, but I'm not sure _where_ the best place for that to live is. I think a question for CFA engineers

app/views/state_file/questions/nj_homeowner_eligibility/edit.html.erb
mmazanec22: [dust/pebble] It looks like gov UK is also using aria-expanded on their example, and voiceover announced whether or not something was expanded when I tried it on their example. So it might be worth looking into adding that too, or figuring out whether we need that

spec/controllers/state_file/questions/nj_homeowner_eligibility_controller_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      let(:intake) { create :state_file_nj_intake, household_rent_own: "both" }
+      it "does not show" do
+        expect(described_class.show?(intake)).to eq false
+      end
+    end
+  end
+
+  describe "#next_path" do
+    describe "hard no ineligible states" do
+      context "when NO to homeowner_home_subject_to_property_taxes" do
```
mmazanec22, line 70: [dust] Maybe instead of using the field name, it makes sense to make the test context human readable for additional context, since the field name is in the test anyway

mmazanec22: Good point. I think I'm just finding it hard to remember what some fields mean without referring to en.yml-- not for any fault in naming, but bc this is such dense subject matter! 😅 If you wanted to change it, maybe something like "when a taxpayer paid neither rent nor property taxes"?

app/controllers/state_file/questions/nj_household_rent_own_controller.rb
```diff
@@ -2,10 +2,8 @@ module StateFile
   module Questions
     class NjHouseholdRentOwnController < QuestionsController
       include ReturnToReviewConcern
-      def edit
-        @filing_year = Rails.configuration.statefile_current_tax_year
-        super
-      end
+
+      before_action -> { @filing_year = Rails.configuration.statefile_current_tax_year }
```
jachan, line 6: [DUST] Flagging that we have an open discussion on how to source `@filing_year` in this PR: https://github.com/codeforamerica/vita-min/pull/4800#pullrequestreview-2332008557

app/models/state_file_nj_intake.rb
```diff
@@ -86,12 +91,18 @@ class StateFileNjIntake < StateFileBaseIntake
...
+  enum homeowner_main_home_multi_unit: { unfilled: 0, yes: 1, no: 2}, _prefix: :homeowner_main_home_multi_unit
+  enum homeowner_main_home_multi_unit_max_four_one_commercial: { unfilled: 0, yes: 1, no: 2}, _prefix: :homeowner_main_home_multi_unit_max_four_one_commercial
+  enum homeowner_more_than_one_main_home_in_nj: { unfilled: 0, yes: 1, no: 2}, _prefix: :homeowner_more_than_one_main_home_in_nj
+  enum homeowner_shared_ownership_not_spouse: { unfilled: 0, yes: 1, no: 2}, _prefix: :homeowner_shared_ownership_not_spouse
 
   def disqualifying_df_data_reason
     w2_states = direct_file_data.parsed_xml.css('W2StateLocalTaxGrp W2StateTaxGrp StateAbbreviationCd')
     return :has_out_of_state_w2 if w2_states.any? do |state|
-      (state.text || '').upcase != state_code.upcase
+      !(state.text || '').casecmp(state_code).zero?
```
jachan, line 105: [NIT] What's the purpose of changing this? I find the old one more readable.

jachan (changes requested):
Seems good on the whole! Minor comments. Would want to see that repeated `is_soft_no` and `is_hard_no` logic abstracted out into it's own class and tests before merging.

app/controllers/state_file/questions/nj_household_rent_own_controller.rb
aloverso: @jachan it seems like that's talking about a separate value? This value is coming from the Rails config. Is there something I'm missing?

app/models/state_file_nj_intake.rb
aloverso: I'm guessing this was a rubocop autofix.

aloverso: So problem being, if I change it back, it'll just get changed again in the future

app/controllers/state_file/questions/nj_household_rent_own_controller.rb
jachan: I guess it's `return_year` vs `filing_year` which are separate, got the two mixed up! We can disregard.

app/models/state_file_nj_intake.rb
jachan: Yep, if it's a rubocop fix, we can leave it.

app/views/state_file/questions/nj_homeowner_eligibility/edit.html.erb
aloverso: There exists `cfa_checkbox_set` but it's not being used in the codebase, and doesn't support HTML options (including aria-expanded but also just the ones we need to get the checkbox values and the expand/collapse function working). I think we should give this feedback to CFA but it's not part of this codebase, it's a library being pulled in.

mmazanec22: Maybe we should use checkbox set for now (assuming it outputs the correct html) and ask them to support aria-expanded? And if that's not happening, perhaps consider a design change?

aloverso: @mmazanec22 Sorry I didn't explain correctly - that was my plan, but the checkbox set doesn't appear to support the expand/collapse function. Like, I can't give a particular checkbox a particular class/data tag which is what allows one checkbox to trigger the other one appearing. So at least with the current way the set works, I don't think we can use it in this page because it can't support the appear/disappear functionality

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
mrotondo: I would move it into a method (that takes an intake) in a new module in app/helpers (possibly in a new state_file subdirectory and named something like `nj_homeowner_helper.rb`), and include that module in each class where you currently have this logic.

app/forms/state_file/nj_homeowner_eligibility_form.rb
```diff
@@ -0,0 +1,21 @@
...
+                       :homeowner_main_home_multi_unit_max_four_one_commercial,
+                       :homeowner_more_than_one_main_home_in_nj,
+                       :homeowner_shared_ownership_not_spouse
+
+    def save
+      is_hard_no = (self.homeowner_home_subject_to_property_taxes == 'no' or (self.homeowner_main_home_multi_unit == 'yes' and self.homeowner_main_home_multi_unit_max_four_one_commercial == 'no'))
+      is_soft_no = (self.homeowner_more_than_one_main_home_in_nj == 'yes' or self.homeowner_shared_ownership_not_spouse == 'yes' or (self.homeowner_main_home_multi_unit == 'yes' and self.homeowner_main_home_multi_unit_max_four_one_commercial == 'yes'))
+
+      if is_soft_no || is_hard_no
+        @intake.property_tax_paid = nil
```
mrotondo, line 15: should this be `@intake.update(property_tax_paid: nil)`? it looks like this will currently be saved by the `@intake.update` call at the end of this method, but i wouldn't want to rely on the way the code is currently ordered.

app/forms/state_file/nj_ineligible_property_tax_form.rb
```diff
@@ -0,0 +1,5 @@
+module StateFile
+  class NjIneligiblePropertyTaxForm < QuestionsForm
+    def save; end
```
mrotondo, line 3: are these empty save methods necessary (here and in nj_unsupported_property_tax_form)?

app/forms/state_file/nj_homeowner_eligibility_form.rb
aloverso: @mrotondo This is a useful callout, is your suggestion the preferred approach? I was borrowing from the approach in other conditional screens where changing one answer would reset the following dependent screen - specifically was seeing it in `NyCountyForm`. Happy to do either approach, what is the preference here?

mrotondo: I think I may have miscommunicated - I think this method is well structured & doing all the right things, I was just noting that it combines assignment (`@intake.property_tax_paid = nil`, which doesn't result in anything being saved to the database) with `update` (`@intake.update(attributes_for(:intake))`, which calls `save` behind the scenes, and so ensures that the previous assignment will get persisted).

It's a minor stylistic concern, and one you should feel free to ignore, as the code will currently do exactly what you meant for it to do - should have marked it [DUST], I think is the right word?

I'd prefer either
1. multiple assigns followed by an explicit `save` call (as `NyCountyForm` does, which ensures that the whole method's changes go in a single database transaction), or
2. no assigns, but multiple `updates` and no `save` call, which doesn't rely on the order of lines staying fixed.

jachan (approved):
Looks good, the homeowner eligibility helper is very nice.

app/helpers/state_file/nj_homeowner_eligibility_helper.rb
```diff
@@ -0,0 +1,19 @@
+module StateFile
+  module NjHomeownerEligibilityHelper
+    class << self
+
+      def determine_eligibility(intake)
+        is_ineligible = intake.homeowner_home_subject_to_property_taxes == 'no' || (intake.homeowner_main_home_multi_unit == 'yes' && intake.homeowner_main_home_multi_unit_max_four_one_commercial == 'no')
+        is_unsupported = intake.homeowner_more_than_one_main_home_in_nj == 'yes' || intake.homeowner_shared_ownership_not_spouse == 'yes' || (intake.homeowner_main_home_multi_unit == 'yes' && intake.homeowner_main_home_multi_unit_max_four_one_commercial == 'yes')
+
+        if is_ineligible
```
mrotondo, line 9: [PEBBLE] I think these return values would be good candidates to turn into constants, to reduce the risk that someone mis-types a string when checking the output of this function. i.e. at the top of this class you could add something like
```
Ineligible = :ineligible
Unsupported = :unsupported
Advance = :advance
```
and elsewhere, when checking output, you could do things like `if ...determine_eligibility(intake) == NjHomeownerEligibilityHelper::Unsupported`. a bit verbose, but less error-prone

mrotondo (approved):

PR #4805, opened 2024-09-27
Fyst 394 do not allow greater than 10 contributions on public school contributions page

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-394
## Is PM acceptance required?
## What was done?
- Added logic to the add button that disables it if there are already 10 contributions, and also shows a warning below it
## How to test?
- Add 10 public school contributions & verify that the button is disabled and the warning is shown
- There are also automated tests verifying this
image

### DIFF
--- a/app/models/state_file_az_intake.rb
+++ b/app/models/state_file_az_intake.rb
@@ -119,6 +119,9 @@ class StateFileAzIntake < StateFileBaseIntake
 
   validates :made_az321_contributions, inclusion: { in: ["yes", "no"]}, on: :az321_form_create
   validates :az321_contributions, length: { maximum: 10 }
+
+  # validates :made_az322_contributions, inclusion: { in: ["yes", "no"]}, on: :az322_form_create
+  validates :az322_contributions, length: { maximum: 10 }
   def federal_dependent_count_under_17
     self.dependents.select{ |dependent| dependent.age < 17 }.length
   end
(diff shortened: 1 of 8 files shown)


app/models/state_file_az_intake.rb
```diff
@@ -119,6 +119,8 @@ class StateFileAzIntake < StateFileBaseIntake
 
   validates :made_az321_contributions, inclusion: { in: ["yes", "no"]}, on: :az321_form_create
   validates :az321_contributions, length: { maximum: 10 }
+
+  validates :az322_contributions, length: { maximum: 10 }, on: :az322
```
arinchoi03, line 123: why is the implementation for this different from the one above `:az321_contributions`?

mrotondo: The az322 (and originally, az321) contributions both were implemented with a boolean value on the contributions model which should have been on the intake ("made_contribution"), largely because of misleading mocks which referenced another similar page without noting the important difference. The az_321 contributions model was then refactored to move that boolean up to the intake, but (I learned while doing this story), that refactor never happened for the az322 contributions model. I'll go more into this at computer gardening next week, but let me know if that's not clear.

arinchoi03 (approved):
I think I'm still wrapping my brain around the diff between 321 & 322 but I think I can get behind this PR without the full explanation.

PR #4810, opened 2024-09-30
TBE-114 Stop submitting mixpanel events in dev

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-114
## What was done?
- Currently, all events fired from dev, demo, staging, heroku environments will all be received at the GetYourRefund.dev mixpanel. ~~I am disabling the `run` command which appears to actually send the tracking events to mixpanel.~~ Trying to mock out a `@tracker` by using a Struct per Jey's suggestion
- Alternatives considered: 
   - I thought about just removing the config variables `mixpanel_token` from the rails credentials files and storing that in LastPass. This would minimize any code changes but would be less transparent about why/how mixpanel is being disabled for the non-prod environments.
   - I also tried to see if there was a way to conveniently disable the Dev environment on mixpanel's end, but alas I didn't find anything :( 
- Risks:
   - It'll be harder to do acceptance testing for mixpanel events on staging/demo/heroku environments once we turn these off. I think the best course of option would be to locally manipulate the code in MixpanelService & pair with a prod person to show what the events will look like in mixpanel. We can also turn on the mixpanel events via a commit (adding certain environments to the `Rails.env.production?` block as needed on a temporary basis).
## How to test?
  - Unit tests
  - Try navigating to a page on a demo page & see that reflected on the Dev Mixpanel & switch over to this branch, go to the same page & see that the event does not come through.

### DIFF
--- a/app/services/mixpanel_service.rb
+++ b/app/services/mixpanel_service.rb
@@ -39,7 +39,7 @@ class MixpanelService
   #
   def run(distinct_id:, event_name:, data: {})
     unless Rails.env.production?
-      Rails.logger.error "You are currently on non-prod environment. if You want to track mixpanel events in non-prod environments, please uncomment this check in Mixpanel Service"
+      Rails.logger.error "You are currently on non-prod environment. If you want to track mixpanel events in non-prod environments, please uncomment this check in Mixpanel Service"
       return
     end
 
--- a/app/services/mixpanel_service.rb
+++ b/app/services/mixpanel_service.rb
@@ -18,23 +18,25 @@ class MixpanelService
 
     @buffer = []
     @mutex = Mutex.new
+
     @tracker =
       if Rails.env.production?
         Mixpanel::Tracker.new(mixpanel_key) do |type, message|
           buffer_event_for_send(type, message)
         end
-      elsif Rails.env.development? # for debugging purposes
-        Struct.new(:track) do
+      elsif Rails.env.development?
+        Struct.new("DummyTracker") do
           def track(distinct_id, event_name, data)
+            # for debugging purposes
             Rails.logger.info("Sending Mixpanel event: id #{distinct_id}, event_name #{event_name}, data #{data}")
           end
-        end
+        end.new
       else # demo, staging, heroku
-        Struct.new(:track) do
+        Struct.new("DummyTracker") do
           def track(_distinct_id, _event_name, _data)
             # do nothing
           end
-        end
+        end.new
       end
 
     # silence local SSL errors
--- a/app/services/mixpanel_service.rb
+++ b/app/services/mixpanel_service.rb
@@ -18,9 +18,24 @@ class MixpanelService
 
     @buffer = []
     @mutex = Mutex.new
-    @tracker = Mixpanel::Tracker.new(mixpanel_key) do |type, message|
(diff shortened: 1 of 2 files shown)


app/services/mixpanel_service.rb
```diff
@@ -38,6 +38,11 @@ def initialize
   # @param [Hash] data: (optional, defaults to {}) data to be sent to mixpanel
   #
   def run(distinct_id:, event_name:, data: {})
+    unless Rails.env.production?
+      Rails.logger.error "You are currently on non-prod environment. If you want to track mixpanel events in non-prod environments, please uncomment this check in Mixpanel Service"
```
jnf, line 42: is this going to clutter our sentry/datadog logs for non-prod environments? I don't know if we need to log an error for every event we don't track.

spec/services/mixpanel_service_spec.rb
```diff
@@ -65,6 +68,33 @@
         procs.pop.call
       end
     end
+
+    context "in non-prod environment" do
```
jnf, line 72: love these tests!

jnf (commented):
just an idea -- what do you think about maybe changing `@tracker` in non-prod environments. It could be a no-op in remote non-prod (like a Struct that responds to `#track` but does nothing), and a basic logger in dev (like a Struct that responds to `#track` but just funnels the event to Rails.logger.info). 

Maybe overkill? I'm curious if there's value in not treating the tracker as an error in non-prod environments.

arinchoi03:
@jnf lmk if I used Struct correctly, I have actually never used them before 😅

app/services/mixpanel_service.rb
```diff
@@ -18,9 +18,26 @@ def initialize
...
-      buffer_event_for_send(type, message)
-    end
+
+    @tracker =
+      if Rails.env.production?
+        Mixpanel::Tracker.new(mixpanel_key) do |type, message|
+          buffer_event_for_send(type, message)
+        end
+      elsif Rails.env.development?
+        Struct.new("DummyTracker") do
```
arinchoi03, line 28: if i don't pass it any options it was yelling at me, so I had to pass in a string here

jnf:
@arinchoi03 this looks pretty great! thank you for giving Struct a try. anything else there would be polish/preference (like i really hate if/else blocks but that's literally just my personality flaw).

jnf (approved):
love it

arinchoi03:
we discussed async using hashes

```
TRACKERS = Hash.new(
  Struct.new("DummyTracker") do 
    def track(_distinct_id, _event_name, _data); end
  end.new
)

TRACKERS["development"] = Struct.new("DevTracker") do
  def track(distinct_id, event_name, data)
    Rails.logger.info("Sending Mixpanel event: id #{distinct_id}, event_name #{event_name}, data #{data}")
  end
end.new

TRACKERS["production"] = Mixpanel::Tracker.new(mixpanel_key) do |type, message|
  buffer_event_for_send(type, message)
end

@tracker = TRACKERS[Rails.env]
```
but since this ran into an error `dynamic constant assignment` I opted to keep the if/else syntax. Open to refactoring down the line, if desired

PR #4811, opened 2024-09-30
FYST 558 - XML Editor bug for W2s

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-558

## Is PM acceptance required?

## What was done?
The form was accidentally erasing important tax nodes, like EmployeeSSN, whenever someone opened the editor to look at the data. This happened even if they didn't change anything.
This was fixed by editing the w2 attributes method to only update a field if it has changed and the new value is present. This prevents overwriting existing data with blank or unchanged values.

## How to test?
Use the heroku link to do the following: 

1- Transfer fake DF data from [Nick](https://github.com/codeforamerica/vita-min/blob/main/spec/fixtures/state_file/fed_return_xmls/2023/nc/nick.xml)

2- Click on the “I’m on a demo environment, let me edit the response XML” and press “Continue” (without editing anything)

3 - Continue through the flow without changing anything - submit & download the pdf

4 - In the NC XML  <IncTaxWith> should be equal to “15”

5 - In the PDF Line 20a (Primary NC Income tax withheld) should also be qual to “15”

### DIFF
--- a/app/forms/state_file/federal_info_form.rb
+++ b/app/forms/state_file/federal_info_form.rb
@@ -114,19 +114,19 @@ module StateFile
     end
 
     def w2s_attributes=(attributes)
+      index = 0
       attributes.each do |_form_number, w2_attributes|
-        w2 = w2s.find { |w| w.id == w2_attributes['id'] }
-        if w2
-          w2_attributes.except('id', '_destroy').each do |field, value|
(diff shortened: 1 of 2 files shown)


mrotondo (approved):
nice fix! 🙌

PR #4815, opened 2024-09-30
fix error messages on sales use tax form

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-318
## Is PM acceptance required?
## What was done?
- Updated the error messages for the manually calculated sales & use tax form field to reflect design's specifications in conversation on the JIRA story after the original story was accepted & the PR was merged.
- Added `allow_blank: true` to the numericality validations so that they don't show up when the field is left blank (this will still get caught by the `presence: true` validation on the field), resulting in us only ever showing one error message at a time
## How to test?
1. Go through the NC flow til you get to the sale & use tax page.
1. Select "Yes" and "I kept a complete record..."
1. Submit the form without entering anything in the "Enter calculated use tax" field. _Verify that only "Can't be blank." is shown_
1. Submit the form with a non-number in that field. _Verify that "Please enter numbers only." is shown_
1. Submit the form with a negative number in that field. _Verify that "Please enter numbers only." is shown_
1. Change /en to /es in the url, and do the last two steps again, _verifying that we show "Por favor ingresa únicamente números." in those cases
**Before**
image
image
image

**After**
image
image
image

### DIFF
--- a/app/forms/state_file/sales_use_tax_form.rb
+++ b/app/forms/state_file/sales_use_tax_form.rb
@@ -10,6 +10,7 @@ module StateFile
     validates :sales_use_tax,
       presence: true,
       numericality: {
+        allow_blank: true,
         greater_than_or_equal_to: 0
       },
       if: -> { sales_use_tax_calculation_method == "manual" }
--- a/app/forms/state_file/sales_use_tax_form.rb
+++ b/app/forms/state_file/sales_use_tax_form.rb
@@ -11,7 +11,8 @@ module StateFile
       presence: true,
       numericality: {
         allow_blank: true,
-        greater_than_or_equal_to: 0
+        greater_than_or_equal_to: 0,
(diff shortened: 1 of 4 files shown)


app/forms/state_file/sales_use_tax_form.rb
```diff
@@ -10,6 +10,7 @@ class SalesUseTaxForm < QuestionsForm
     validates :sales_use_tax,
       presence: true,
       numericality: {
+        allow_blank: true,
         greater_than_or_equal_to: 0
```
tahsinaislam, line 14: question: why wouldnt you just add one message in the numericality hash do you don't have to have two strings in the yaml? is it to make it more clear?
````
numericality: {
        allow_blank: true,
        greater_than_or_equal_to: 0,
        message:  .not_a_number }
```

tahsinaislam (approved):

PR #4816, opened 2024-10-01
Fyst 451/add state id nc

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-451

## What was done?
- Added in NC primary & spouse state ID pages. XML seems to be covered already and no action was necessary to implement that.

## How to test?
- Unit tests provided
- Stepping through the NC flow and entering valid data in the NC state pages
  - Spiderman persona xml seems to work for married filing jointly
  - Tucker single persona seems to work for single filer

image
image
image


app/forms/state_file/nc_state_id_concern.rb
```diff
@@ -0,0 +1,36 @@
+module StateFile
+  module NcStateIdConcern
```
powersurge360, line 2: I'm aware that there's a `StateIdForm`. I chose to do things this way for a few reasons. `StateIdForm` seems to have a lot of crossed logic from multiple states and I wasn't sure if I wanted or needed to contribute to that. I also didn't want to create a form which _could_ be used to map to a model but shouldn't (a concern is unusable by itself) and lastly I needed to break out into my own code to take advantage of `DateAccessor` for simpler validation of dates.

app/forms/state_file/nc_state_id_concern.rb
```diff
@@ -0,0 +1,36 @@
...
+      date_accessor :expiration_date, :issue_date
+
+      validates :issue_date, presence: true, unless: -> { id_type == "no_id" }
+      validates :expiration_date, presence: true, unless: -> { id_type == "no_id" || non_expiring == "1" }
+      validates :state, presence: true, inclusion: { in: States.keys }, unless: -> { id_type == "no_id" }
+      validates :id_type, presence: true
+      validates :id_number,
+                presence: true,
+                # Seems to just be for New York
+                # alphanumeric: true, length: {is: 9},
```
powersurge360, line 32: It seems weird that this is only for New York, but that's what `StateIdForm` seems to indicate.

app/views/state_file/questions/ny_primary_state_id/_state_id.html.erb
```diff
@@ -56,6 +56,7 @@
       </div>
     </div>
   </div>
+  <%= content_for :info_box %>
```
powersurge360, line 59: Not ideal but a concession I made to not interfere with the reusable templates more than I have to.

tahsinaislam (commented):
I think you still need to implement the XML part of this. Let me know if you want to pair on this

tahsinaislam (approved):

PR #4818, opened 2024-10-01
Fyst 524 add MD, ID, and NJ FAQ

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-524
## What was done?
- in production, modified visible FAQ states to include *only* those that have been open before AND will be open in the coming season (the upshot is that in production only AZ should be visible, not NY because it is not reopening)
- in non-prod envs, visible FAQ states include all that will be open in the coming season, regardless of whether they were open before (so this will now include MD, ID, NJ, NC in addition to AZ)
## How to test?
- tried to test all scenarios in a way that is not too completely unreadable and confusing. "all scenarios" = 
  - before, during, and after filing season in a given year
  - prod and non-prod
  - states that are open 2 years in a row, closing, opening for the first time
- acceptance testing will be weird because A) behavior is specific to environment and B) session toggles can't change the year apparently
- Before
  image
- After
  image

### DIFF
--- a/app/controllers/state_file/faq_controller.rb
+++ b/app/controllers/state_file/faq_controller.rb
@@ -2,32 +2,8 @@ class StateFile::FaqController < ApplicationController
   layout "state_file"
 
   def index
-    intake_start = Rails.configuration.state_file_start_of_open_intake
-    intake_end = Rails.configuration.state_file_end_of_in_progress_intakes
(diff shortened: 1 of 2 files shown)


embarnard (approved):
Looks good!

PR #4823, opened 2024-10-02
Fyst-866 1099G: use xml builder for more states

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-866
## What was done?
- Reused the AZ 1099G xml builder for other states (ID, MD, NC):
  - moved the code into a generic not-state-namespaced file
  - set that builder as the default for `form1099g_builder` (which is fine i think because the way it actually gets used for a state is `form1099gs` is added to `supported_documents` on each subclass; also it can be overwritten)
  - added 1099gs to supported docs for ID, MD, NC
  - added a spec for the new builder because those lines weren't tested
  - in `state_return_spec.rb`, added test that 1099g's are present for all states that require them
## How to test?
- see description of automated tests under "what was done"
- acceptance testing should be straightforward: just go through each state's flow and add 1099g's, see that they show up in the xml


tahsinaislam (approved):
ohhh soo cleann ✨

mrotondo (approved):

PR #4824, opened 2024-10-03
Nj 53 child dependent

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/53

## What was done?
Use directfile data to fill line 64 in calculator, XML, PDF

## How to test?
Use zeus one dep, navigate through flow, check PDF and XML

Screenshot 2024-10-03 at 12 55 00 PM
Screenshot 2024-10-03 at 12 55 16 PM

### DIFF
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
@@ -131,8 +131,8 @@ module SubmissionBuilder
                     xml.NewJerseyTaxableIncome calculated_fields.fetch(:NJ1040_LINE_42)
                   end
                   
-                  xml.ChildDependentCareCredit calculated_fields.fetch(:NJ1040_LINE_64)
-                  
+                  xml.ChildDependentCareCredit calculated_fields.fetch(:NJ1040_LINE_64) if calculated_fields.fetch(:NJ1040_LINE_64)
+
                   line_65 = calculated_fields.fetch(:NJ1040_LINE_65)
                   xml.NJChildTCNumOfDep calculated_fields.fetch(:NJ1040_LINE_65_DEPENDENTS) if line_65
                   xml.NJChildTaxCredit line_65 if line_65
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
@@ -131,9 +131,12 @@ module SubmissionBuilder
                     xml.NewJerseyTaxableIncome calculated_fields.fetch(:NJ1040_LINE_42)
                   end
                   
+                  xml.ChildDependentCareCredit calculated_fields.fetch(:NJ1040_LINE_64)
+                  
                   line_65 = calculated_fields.fetch(:NJ1040_LINE_65)
                   xml.NJChildTCNumOfDep calculated_fields.fetch(:NJ1040_LINE_65_DEPENDENTS) if line_65
                   xml.NJChildTaxCredit line_65 if line_65
+
                 end
               end
             end
--- a/spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
+++ b/spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
@@ -147,6 +147,9 @@
       <StudentLoanInterestDedAmt>1300</StudentLoanInterestDedAmt>
       <TotalAdjustmentsAmt>1500</TotalAdjustmentsAmt>
     </IRS1040Schedule1>
+    <IRS1040Schedule3>
+      <CreditForChildAndDepdCareAmt>1000</CreditForChildAndDepdCareAmt>
+    </IRS1040Schedule3>
     <IRS1040Schedule8812 documentName="IRS1040Schedule8812" documentId="IRS8812001">
       <AdjustedGrossIncomeAmt>56300</AdjustedGrossIncomeAmt>
       <ModifiedAGIAmt>56300</ModifiedAGIAmt>
--- a/spec/lib/efile/nj/nj1040_calculator_spec.rb
+++ b/spec/lib/efile/nj/nj1040_calculator_spec.rb
@@ -297,34 +297,44 @@ describe Efile::Nj::Nj1040Calculator do
     end
 
     context 'with an income of 150k or less' do
-      it 'returns 10% of federal credit for an income of 150k or less' do
-        allow(instance).to receive(:calculate_line_42).and_return 150_000
+      before do
+        allow(instance).to receive(:calculate_line_42).and_return nj_taxable_income
         instance.calculate
-        expect(instance.lines[:NJ1040_LINE_64].value).to eq(100)
       end
 
-      it 'returns 20% of federal credit for an income of 120k or less' do
-        allow(instance).to receive(:calculate_line_42).and_return 120_000
-        instance.calculate
-        expect(instance.lines[:NJ1040_LINE_64].value).to eq(200)
+      context "with an income of 150k or less" do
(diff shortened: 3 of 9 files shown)


spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
```diff
@@ -147,6 +147,9 @@
       <StudentLoanInterestDedAmt>1300</StudentLoanInterestDedAmt>
       <TotalAdjustmentsAmt>1500</TotalAdjustmentsAmt>
     </IRS1040Schedule1>
+    <IRS1040Schedule3>
```
jachan, line 150: [DUST] How was this XML sourced? Did we copy it from another state's XML example?

spec/lib/efile/nj/nj1040_spec.rb
```diff
@@ -265,4 +265,53 @@ def over_65_birth_year
       expect(instance.lines[:NJ1040_LINE_42].value).to eq(instance.lines[:NJ1040_LINE_39].value)
     end
   end
+
+  describe 'line 64 - child and dependent care credit' do
```
jachan, line 269: [DUST] These are great tests, thank you!

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -130,6 +130,8 @@ def document
                   if calculated_fields.fetch(:NJ1040_LINE_42) > 0
                     xml.NewJerseyTaxableIncome calculated_fields.fetch(:NJ1040_LINE_42)
                   end
+
+                  xml.ChildDependentCareCredit calculated_fields.fetch(:NJ1040_LINE_64)
```
jachan, line 134: [DUST] In the case where the `federal_child_and_dependent_care_credit` = 0, this will write 0 into the XML. Should we have an if statement that only writes it if it's greater than 0?

jachan (approved):
Left a comment on the ticket for Tristan about the PDF checkbox on Line 64. Some questions here about how we want to handle when the case where `federal_child_and_dependent_care_credit` is 0. If we add some logic to handle the less than 0 case, we should probably add a test as well!

spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
mmazanec22: I inferred it from the directfile import. Feels sketchy but I was not sure how else to accomplish this

jachan: Gotcha, seems fine.

aloverso: Did you look in some of the other DF state imports?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -1057,6 +1057,33 @@
...
+          'Text194',
+          'Text195',
+          'Text196',
+        ]
+        digit_fields.each do |field_name, _i|
+          digits_in_pdf << pdf_fields[field_name]
+          digits_in_pdf << "." if field_name == digit_fields[-3]
+        end
+        tax_credit = digits_in_pdf.to_f
+        expect(tax_credit).to eq 400
```
aloverso, line 1084: [boulder] I am not a big fan of this testing approach. Because the amount of logic that is written here, it feels like the test itself needs a test to confirm that it properly constructs a number out of independent digit strings in the PDF. Philosophically, I would strongly prefer to test the data as it appears in the PDF, rather than constructing a new third-party data type from the object under test, and testing that data type. I think it gives a lot more confidence if we just test the output of the object under test.

You could pull in the refactor work I already started on the `nj-73-refactor-insert-digits-pdf` branch here, to simplify these expect statements, if that still feels important. 

However, I've been thinking about it more, and I'm not entirely sure what's to be gained by the refactor at all - either the one in the code here OR the one on my branch there. It's not as if it's fewer lines of code or something. What are we really gaining in test readability? And is it worth it for what we're losing in test confidence and directness?

Marking this as a boulder because I just want to talk about it as a team as we pick a path forward, and that's kind of what I mean by re-review

aloverso (commented):
Left a comment about refactoring, would love to have a conversation there. I wrote it up as a separate ticket [here](https://github.com/newjersey/affordability-pm/issues/73) so that we could keep the context of refactoring separate from the context of working on a particular story. So it would be fine to merge this in without the test refactor and we could talk about the refactor separately - it doesn't have to be a blocker on this story.

spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
mmazanec22: Yes, this was not present in any others

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -279,7 +279,56 @@ def over_65_birth_year
...
+    end
+
+    context 'with an income of 150k or less' do
+      it 'returns 10% of federal credit for an income of 150k or less' do
+        allow(instance).to receive(:calculate_line_42).and_return 150_000
+        instance.calculate
+        expect(instance.lines[:NJ1040_LINE_64].value).to eq(100)
+      end
+
+      it 'returns 20% of federal credit for an income of 120k or less' do
```
embarnard, line 306: [pebble] I think these would be more readable if we lost the "context 'with an income of 150k or less'" context and instead put each of these in their own context like 
```suggestion
context "with an income of 120k or less"
   let(:nj_taxable_income){ 120_000 }
      it 'returns 20% of federal credit' do
        expect(instance.lines[:NJ1040_LINE_64].value).to eq(200)
     end
end
```
with the following at the top of the describe block:
```
before do
        allow(instance).to receive(:calculate_line_42).and_return nj_taxable_income
        instance.calculate
end
```

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -391,6 +391,13 @@
       end
     end
     
+    describe "child and dependent care credit - line 64" do
+      let(:intake) { create(:state_file_nj_intake, :df_data_one_dep, :fed_credit_for_child_and_dependent_care) }
+      it "adds 40% of federal credit for an income of 60k or less" do
+        expect(xml.at("ChildDependentCareCredit").text).to eq(400.0.to_s)
```
embarnard, line 397: [dust]
```suggestion
        expect(xml.at("ChildDependentCareCredit").text).to eq('400')
```

embarnard (approved):
Looks really good! Approved but left some comments 🎉

spec/fixtures/state_file/fed_return_xmls/2023/nj/zeus_one_dep.xml
mmazanec22: This field was not in another DF import. I manually added it here because we needed it to exist to test this PR, and it could be inferred from the df import code.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -279,7 +279,56 @@ def over_65_birth_year
...
+
+  describe 'line 64 - child and dependent care credit' do
+    let(:intake) { 
+      create(
+      :state_file_nj_intake,
+      :df_data_one_dep,
+      :fed_credit_for_child_and_dependent_care)
+    }
+    
+    context 'with an income of over 150k' do
```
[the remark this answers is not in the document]
mmazanec22: That's line 65. Line 64 deals with a wider range of incomes

PR #4828, opened 2024-10-07
NJ 68 - tenant property tax eligibility screen

aloverso:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/68

## What was done?
- Add checkbox screen for tenant property tax eligibility
- Add tenant eligibility logic

## How to test?
- Use any persona for testing non-MFS
- Use MFS for testing "spouse in same home" question

https://github.com/user-attachments/assets/2b082e0c-dda5-401b-82d0-27c27506f859


app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
```diff
@@ -0,0 +1,44 @@
+<%
+    title = t(".title", filing_year: @filing_year)
+    content_for :page_title, title
+%>
+
+<% content_for :card do %>
+  <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+    <h1 class="h2"><%= title %></h1>
+    <p><%= t(".label") %></p>
```
mmazanec22, line 9: [pebble] I know we still need to go back and fix the individual checkboxes being in fieldsets thing with CfA, and I  added that to the sundry a11y issues ticket. In the mean time we can get closer to level A WCAG compliance by programmatically associating the question and the options here with a fieldset and legend rather than a div and a p tag.

app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
```diff
@@ -0,0 +1,44 @@
...
+      <%= f.cfa_checkbox(:tenant_more_than_one_main_home_in_nj, t(".tenant_more_than_one_main_home_in_nj"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <%= f.cfa_checkbox(:tenant_shared_rent_not_spouse, t(".tenant_shared_rent_not_spouse"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+
+      <% if current_intake.filing_status == :married_filing_separately %>
+        <%= f.cfa_checkbox(:tenant_same_home_spouse, t(".tenant_same_home_spouse"), options: { checked_value: "yes", unchecked_value: "no" }) %>
+      <% end %>
+
+    </div>
+
+    <div class="reveal">
```
mmazanec22, line 31: [pebble] We should use the native html summary/details tags for these

app/forms/state_file/nj_tenant_eligibility_form.rb
```diff
@@ -0,0 +1,21 @@
...
+                       :tenant_more_than_one_main_home_in_nj,
+                       :tenant_shared_rent_not_spouse,
+                       :tenant_same_home_spouse
+
+    def save
+      @intake.update(attributes_for(:intake))
+
+      if StateFile::NjTenantEligibilityHelper.determine_eligibility(@intake) != StateFile::NjTenantEligibilityHelper::ADVANCE
+        @intake.update({
+                         rent_paid: nil
```
mrotondo, line 16: [dust] funky whitespace, could be one line

app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
mrotondo: Would the honeycrisp accordion be appropriate here & have the right accessibility? https://honeycrisp.herokuapp.com/cfa/styleguide#molecules-accordion

mrotondo (approved):

app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
mmazanec22: That example does seem to have the right markup!

aloverso: @mmazanec22 I've been looking into this and I don't think it makes sense to fix, at least not in this ticket:
- The "reveal" class is what's styled in `state-file.scss` to have the "blue-box" styling, accordion does not have the application styling. I could edit the scss but I'd want to discuss this with the larger CFA team to ensure that's how the team wants accordions to be styled across the app
- I tried just updating this html to use summary/details html markup, but it seems like that interferes with the honeycrisp javascript and it no longer opens/closes when clicked.
- The "reveal" class is being used in this same way extensively across the application and across states. I'd rather have an accessibility ticket (either for our team or the CFA team) to fix the issue broadly in all occurrences, and I'm not sure it should block merging of this story. 

What do you think?

mmazanec22: Noting here what we just discussed that CfA has a reveal component in honeycrisp with aria-expanded, so I assume the honeycrisp js properly changes that value and we just need to add the attr. We can also add the [optional aria-controls](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) (like in the accordion example in honeycrisp)

aloverso: Update - will bring up with CFA and fix in future ticket

PR #4829, opened 2024-10-07
nj 72 - add automated test for XML schema validation

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/72

## What was done?
- Add automated test for asserting no NJ XML errors
- Fix existing NJ XML errors by introducing `NJW2` document and mapping to replace `IRSW2` and update specs to match
- Fix existing NJ fake IRS DF data that does not conform to schema (w2 wages must be integers) and associated tests

## How to test?
n/a

n/a

### DIFF
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
@@ -3,28 +3,12 @@ module SubmissionBuilder
     module States
       module Nj
         module Documents
-          class NjW2 < SubmissionBuilder::Document
+          class NjW2 < SubmissionBuilder::ReturnW2
             def document
-              w2 = @kwargs[:w2]
-              intake_w2 = @kwargs[:intake_w2]
-              xml_node = Nokogiri::XML(w2.node.to_xml)
+              xml_node = super
+
               xml_node.at("IRSW2")["documentName"] = "NJW2"
               xml_node.at("IRSW2").name = "NJW2"
-
-              if intake_w2.present?
-                state_local_tax_grp_node = xml_node.at(:W2StateLocalTaxGrp)
-                state_tax_group_xml = intake_w2.state_tax_group_xml_node
-                if state_tax_group_xml.present?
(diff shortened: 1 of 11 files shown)


spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -8,6 +8,50 @@
     let!(:submission_efile_device_info) { create :state_file_efile_device_info, :submission, :filled, intake: intake }
     let(:xml) { Nokogiri::XML::Document.parse(described_class.build(submission).document.to_xml) }
 
+    describe "XML schema" do
+
+      context "with one dep" do
+        let(:intake) { create(:state_file_nj_intake, municipality_code: "0101", raw_direct_file_data: StateFile::XmlReturnSampleService.new.read('nj_zeus_one_dep')) }
```
mmazanec22, line 14: [dust] Why specify municipality code here?

mmazanec22 (approved):

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
aloverso: Municipality is a required field in the XML, and by default it is nil (until the user enters it in the app) so to simulate that and get the test to pass, it has to have a value in the test data input.

app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
```diff
@@ -0,0 +1,46 @@
+module SubmissionBuilder
+  module Ty2024
+    module States
+      module Nj
+        module Documents
+          class NjW2 < SubmissionBuilder::Document
```
mrotondo, line 6: [pebble - this looks good, and no change is necessary] It looks like this class largely duplicates the functionality of ReturnW2 and then renames and removes a bunch of nodes. What do you think about having it subclass ReturnW2 and have a `document` method that looks like
```
def document
  super
  rename and remove nodes
end
```
?

mrotondo (approved):

PR #4832, opened 2024-10-07
Update setup things

jenny-heath:
## Link to pivotal/JIRA issue
- N/A
## What was done?
- add postgresql@14 to brewfile because [postgis just changed](https://github.com/Homebrew/homebrew-core/commit/430c4e9ea3449e40d0dc8016d32a3ffe19a31005#diff-a436d87c42acd56e9af1b5d6ea108d20b1516e66bee10bb08ae2dda34f37b8dcL39) to no longer include postgresql as a dependency
  - fun fact: we used to have postgres in the brewfile but [removed it in 2021](https://github.com/codeforamerica/vita-min/commit/2f5d3e0f) because people kept running into issues where the versions of postgres and postgis would get out of sync. sadly i guess we can't rely on postgis to set up that dependency for us, but according to the compatibility matrix we should be good for a few years.
- re-order some steps in readme because bin/setup will fail without credential files and schema files
- add the cloning step to the readme because it's not in there
  - note: i specified cloning with SSH because it will set you up with the right kind of remote for pushing to the repo
## How to test?
- no functional changes except the dependency. i guess it will be tested live when the contractors set up their machines?


jnf (approved):
looks great, thank you!

mrotondo (approved):

PR #4838, opened 2024-10-08
FYST-785 Implement ID Form40 for Household Info

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-785

## Is PM acceptance required? 

## What was done?
- ID40 Calculator was updated to include exemption lines 
- New PDF fillers add for for ID40 and ID39R forms
- Added the following to ID40xml: filing status and exemption details, and dependent information 
- Additional ID39R pdf filler was added to handle additional dependents past 7 dependents because the first 39R will be used for other info later
- The DirectFileData model now checks if a spouse is claimed as a dependent for exemption calculations

## How to test?
- Tests were added for ID XML Return class, ID40 xml and pdf, ID39R pdfs for additional dependents
- XML and json blob was added for john_mfs_8_deps to test handling of additional dependents 

Screenshot 2024-10-10 at 12 26 14 PM
Screenshot 2024-10-10 at 12 26 00 PM


app/lib/pdf_filler/id39r_additional_dependents_pdf.rb
```diff
@@ -0,0 +1,32 @@
+module PdfFiller
+  class Id39rAdditionalDependentsPdf
```
tahsinaislam, line 2: this pdf filler was added in additional to the first ID39R pdf filler because additional ID39R are only used to handle additional dependents while the first ID39R filled out will have other info

app/lib/efile/line_data.yml
```diff
@@ -166,6 +166,14 @@ NYC_EIC_RATE_WK_LINE_5:
...
 NYC_EIC_RATE_WK_LINE_6:
   label: 'Subtract line 4 from line 5 or enter the decimal from the NYC EIC rate table. Also enter on Worksheet C, line 2.'
+ID40_LINE_6A:
+  label: '6a Yourself'
+ID40_LINE_6B:
+  label: '6b Spouse'
+ID40_LINE_6C:
+  label: '6c Dependents'
+ID40_LINE_6D:
+  label: '6d Total household'
```
embarnard, line 176: could these be a little more descriptive, something like" "primary claimed as dependent", "spouse claimed as dependents", "dependents count" and "total dependent count"

app/lib/submission_builder/ty2024/states/id/id_return_xml.rb
```diff
@@ -32,11 +32,25 @@ def supported_documents
...
                 include: true
               },
+              {
+                xml: nil, # TODO 39RForm xml to be added in future story
+                pdf:  PdfFiller::Id39rPdf,
+                include: @submission.data_source.dependents.count > 4
+              },
             ]
-            
+            # If there are more than 7 dependents total (the State XML field can handle a max of 20), another copy of the ID Form 39R PDF is used to handle these cases
```
embarnard, line 44: Sorry I'm confused about this so just going to make sure I understand correctly. The first 4 are on the Id40pdf, the next three are on the id39r and then an additional 13 can be added in set of 3 using the Id39r-additional pdf right?

embarnard (approved):
Very nice work! 🎉

app/lib/submission_builder/ty2024/states/id/id_return_xml.rb
tahsinaislam: yes!

app/lib/efile/line_data.yml
tahsinaislam: ohhh thats a good call i'll fix this!

PR #4840, opened 2024-10-08
Update DOB handling between different guidelines

embarnard:
## Link to pivotal/JIRA issue
- e.g. [JIRA link](https://codeforamerica.atlassian.net/browse/FYST-828?atlOrigin=eyJpIjoiMGZlZDBjNzJkYTFjNDFiZjg0M2ZmMmUxNGI0MTkzOGIiLCJwIjoiaiJ9)
## What was done?
- there are different age calculations depending on the state or what benefits you're calculating the age for
  - most states follow federal guidelines which consider individuals to attain their age the day before their birthday for benefits that need you to be a certain age (for example, senior benefits) but **not** for benefits you age out (for example some dependent benefits)
  - Maryland is an outlier and always considers an individual to attain their age on the day of their birthday
  - this means that when calculating folks' age at the end of the tax year sometimes we have to include people with Jan 1st birthdays and sometimes we don't
  - I added a `calculate_age(inclusive_of_jan_1: true)` method to make it clear that this is a calculation that changes based on whether or not you want to include Jan 1st birthdays in the previous year
  - In the story description it said there were some options to use DF indicators instead of calculating it but I decided against this after talking to Courtney who said we can't reliably count on these always being provided in the DF xml and since Maryland calculates things differently we would have to calculate it anyways (except for <17 benefits)
## How to test?
- unit tests for state-file-dependent, state-file-az-intake and az-140 calculator specs


spec/lib/efile/az/az301_calculator_spec.rb
```diff
@@ -35,10 +35,10 @@
     end
 
     it "calculates AZ301 part 2 values" do
-      expect(instance.lines[:AZ301_LINE_27].value).to eq(2114) # Line 46 from AZ140
-      expect(instance.lines[:AZ301_LINE_32].value).to eq(2114)
+      expect(instance.lines[:AZ301_LINE_27].value).to eq(2167) # Line 46 from AZ140
```
embarnard, line 38: these numbers changed because they are impacted by the number of people (out of primary and spouse) that are 65+. The factory instance is using `df_return_sample.xml` which indicates that the primary is 65+ (`Primary65OrOlderInd`) but factory instance of `state_file_az_intake` doesn't have a primary_birth_date

spec/factories/state_file_az_intakes.rb
```diff
@@ -102,6 +102,7 @@
     raw_direct_file_data { File.read(Rails.root.join('app', 'controllers', 'state_file', 'questions', 'df_return_sample.xml')) }
     primary_first_name { "Ariz" }
     primary_last_name { "Onian" }
+    primary_birth_date { Date.new((MultiTenantService.statefile.current_tax_year - 65), 12, 1) }
```
embarnard, line 105: there were a few calculations that were changing across the codebase because they are impacted by the number of people (out of primary and spouse) that are 65+. Previously we were grabbing this number from the df data and now we are calculating it based on the birthdate we get since there are different rules for different states. 

The factory instance is using df_return_sample.xml which indicates that the primary is 65+ (Primary65OrOlderInd) but doesn't have a primary_birth_date. I added a 65+ birthdate here so that it would be consistent with the DF data used in this instance and to resolve the failing tests.

tahsinaislam (approved):
wooo

app/models/state_file_base_intake.rb
```diff
@@ -304,4 +304,16 @@ def sanitize_bank_details
       self.date_electronic_withdrawal = nil
     end
   end
+
+  def calculate_age(inclusive_of_jan_1: true, dob: nil)
+    # federal guidelines: you qualify for age related benefits the day before your birthday
+    # that means for a given tax year those born on Jan 1st the following tax-year will be included
+    # this does not apply for benefits you age out of or any age calculations for Maryland
+    birth_year = dob.year
+    if inclusive_of_jan_1 && state_code != 'md'
```
mpidcock, line 313: is there a way to avoid having the 'md' state_code here? should it be enough to use `inclusive_of_jan_1`?

embarnard: Why would we want to avoid having md state code in here? But yes we could have Maryland overwrite the function in their own state intake class @mpidcock

mpidcock: Not using individual state codes in conditional logic was a principle we decided on after last tax season and the main reason we introduced the StateInformationService. It makes maintenance harder if State-specific rules are hidden in shared components making them harder to find, for example if we want to remove MD as a supported State. Overriding in the MD intake class would make the logic visible and isolated to the only State that is concerned with the exception.

app/lib/efile/az/az140_calculator.rb
```diff
@@ -102,6 +102,13 @@ def calculate_line_2c
         @intake.charitable_noncash_amount&.round
       end
 
+      def calculate_line_8
+        # Age 65 or over (you and/or spouse) count
```
embarnard, line 106: note: check line 8

PR #4841, opened 2024-10-09
md502 xml

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-653
## Is PM acceptance required?
## What was done?
- Added income fields to the MD 502 XML & PDF 
- Attach a blank MD 502R to the XML & PDF if there is income from taxable pensions, IRAs or annuities
## How to test?
- added automated tests that should cover everything
- submit with federal returns with various values in the fields described in the story, and download the output XML & PDFs to make sure the correct values are present


app/lib/efile/line_data.yml
```diff
@@ -490,6 +490,16 @@ IT214_LINE_33:
   label: '33 Enter the amount from line 32 or 31, whichever is less. This is the credit for your household.'
 IT227_PART_2_LINE_1:
   label: ''
+MD502_LINE_1:
+  label: "Adjusted gross income from your federal return"
```
embarnard, line 494: Could you add the line number before the line description like the other labels, for example:
```suggestion
  label: "1 Adjusted gross income from your federal return"
```

app/lib/pdf_filler/md502_r_pdf.rb
```diff
@@ -0,0 +1,24 @@
...
+      "md502R-TY2023"
+    end
+
+    def initialize(submission)
+      @submission = submission
+
+      # Most PDF fields are grabbed right off the XML
+      builder = StateFile::StateInformationService.submission_builder_class(:md)
+      @xml_document = builder.new(submission).document
+      @calculator = submission.data_source.tax_calculator
```
embarnard, line 15: we shouldn't call the calculator here, the only place the calculator should be called is in the submission builder. Instead we should grab pdf elements from the `@xml_document` directly or intake/df_data when its not present in the XML.

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
```diff
@@ -29,12 +29,20 @@ def schema_file
...
             supported_docs = [
               {
                 xml: SubmissionBuilder::Ty2024::States::Md::Documents::Md502,
                 pdf: PdfFiller::Md502Pdf,
                 include: true
               },
+              {
+                xml: SubmissionBuilder::Ty2024::States::Md::Documents::Md502R,
+                pdf: PdfFiller::Md502RPdf,
+                include: has_income_from_taxable_pensions_iras_annuities
```
embarnard, line 44: can we remove lines 32 and 33 and instead just do
```suggestion
                include:  @submission.data_source.direct_file_data.fed_taxable_pensions&.to_i.positive?
```

db/schema.rb
```diff
@@ -9,6 +9,7 @@
 # migrations use external dependencies or application code.
 #
 # It's strongly recommended that you check this file into your version control system.
+
```
embarnard, line 12: remove?

embarnard (approved):
Looks good! Just a few requested changes, small things need no re-review

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
mrotondo: I would prefer to access the calculator, since it lets us avoid a very small amount of code duplication. I was following the example that Jenny shared with me of [`ny_return_xml.rb`](https://github.com/codeforamerica/vita-min/blob/main/app/lib/submission_builder/ty2022/states/ny/ny_return_xml.rb#L135) where the calculator is used to determine conditions for including various documents

PR #4843, opened 2024-10-09
Synchronize dependents on tin and ssn

arinchoi03:
## Link to pivotal/JIRA issue
- N/A; [See convo here](https://cfastaff.slack.com/archives/C0544ERAFQV/p1728495837684339?thread_ts=1727999227.602799&cid=C0544ERAFQV)
## What was done?
- When we synchronize dependents, use `tin` on json data of a dependent to match the xml dependent.
- this is based off of MD502 household info branch bc I added a bunch of fixtures that needed tin added. I think that PR will be mergeable soon
- Throw an error when xml/json people do not match -- we will remove this exception throwing eventually, but this was added during testing season to see if this is an issue we need to be alert about.
## How to test?
  - Unit tests


app/models/direct_file_json_data.rb
```diff
@@ -68,8 +68,13 @@ def find_matching_json_dependent(dependent)
...
     dependents.find do |json_dependent|
-      # TODO: find match based on SSN
-      json_dependent["firstName"] == dependent.first_name
+      next unless json_dependent.present?
+
+      json_tin = json_dependent["tin"]&.tr("-", "")
+      xml_ssn = dependent.ssn
+
+      next unless json_tin && xml_ssn
+      json_tin == xml_ssn
```
arinchoi03, line 77: _just in case_ either json blob or xml does not have a `tin` value... I don't want to say it's a match for instance if both `json_tin` and `xml_ssn` are `nil`

app/models/state_file_base_intake.rb
```diff
@@ -83,12 +83,19 @@ def synchronize_filers_to_database
...
   def synchronize_df_dependents_to_database
     direct_file_data.dependents.each do |direct_file_dependent|
       dependent = dependents.find { |d| d.ssn == direct_file_dependent.ssn } || dependents.build
       dependent.assign_attributes(direct_file_dependent.attributes)
 
       dependent_json = direct_file_json_data.find_matching_json_dependent(dependent)
+
+      if direct_file_json_data.data.present? && dependent_json.nil?
+        raise SynchronizeError, "Could not find matching dependent #{dependent.id} with #{state_name} intake id: #{id}"
+      end
```
arinchoi03, line 97: should we also thrown an error if number of dependents on xml / json doesn't match? It's possible that xml has dependents that match some of the json dependents, but some that end up not having a json counterpart. Is that overkill?

app/models/direct_file_json_data.rb
```diff
@@ -1,15 +1,19 @@
 class DirectFileJsonData
 
   def initialize(json)
-    @json = JSON.parse(json || "{}")
+    @data = JSON.parse(json || "{}")
+  end
+
+  def data
+    @data
   end
```
mpidcock, line 9: Consider replacing with `attr_reader :data`, but no strong feelings

spec/fixtures/state_file/fed_return_jsons/2023/md/riley_claimedasdep.json
```diff
@@ -6,7 +6,6 @@
       "middleInitial": null,
       "lastName": "Gilmore",
       "dateOfBirth": "1980-01-01",
-      "isPrimaryFiler": true
```
mpidcock, line 9: Did you confirm this is accurate? I would expect this filer to still have `"isPrimaryFiler": true` on this return even if they are claimed as a dependent on another return, but I can accept I might be wrong if you've already asked about this somewhere else.

app/models/state_file_base_intake.rb
mpidcock: I think that's overkill, if we identify data integrity issues like that with IRS data during testing we can submit issues with them

mpidcock (approved):
Approving, but there are some minor comments to look at. Thanks for doing this 🙏

PR #4853, opened 2024-10-12
fixes verification code flakes

mrotondo:
The tests use a regex to pull the verification code out of the emails that was occasionally picking up a string of 6 numbers in the URL for the logo attachment, which is a random string of hex characters. I updated the regex in the test helpers to look more specifically for 6 digits surrounded by `<strong></strong>` html tags. 

(There were a few places where this was already handled by looking for a whitespace character before the verification code - I don't have strong feelings about which way is better, I just wanted to make them all uniform. The right thing to do would probably be to create a helper so this code doesn't get repeated so much, but I wanted to make a narrow change for now.)


mrotondo:
image

embarnard (approved):

PR #4864, opened 2024-10-16
add a general structure for wrapping up DF JSON, refactor dependents and filers to use it, and use it to wrap 1099 INT data

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-950
## Is PM acceptance required?
## What was done?
- I made a DfJsonAccessor class that mirrors the structure we use to wrap the DF XML, and created subclasses of it to access filers, dependents and 1099-INTs that we extract from the DF JSON
## How to test?
- Automated tests should validate that the 1099-INT data is being captured correctly, and that the refactor of the filer & dependent accessors is not changing any functionality


app/models/df_json_accessor.rb
```diff
@@ -0,0 +1,40 @@
+class DfJsonAccessor
+
+  def initialize(json)
+    @json = json
+  end
+
+  def self.define_json_accessors
```
powersurge360, line 7: I don't 100% like the use of the word `accessor` because to me it implies an api parity with `attr_accessor` somewhere. It is internally consistent with our xml accessor though so I'll not block on it but I did want to say that explicitly.

app/models/df_json_accessor.rb
```diff
@@ -0,0 +1,40 @@
...
+      else
+        define_method(selector_name) do
+          df_json_value(__method__)
+        end
+      end
+    end
+  end
+
+  def df_json_value(selector_name)
+    @json[selectors[selector_name][:key]]
```
powersurge360, line 32: You may consider using 

```ruby
keys = selectors[selector_name][:key].split(' ')
@json.dig(*keys)
```

Which will enable deep traversal of the JSON structure without an API change.

You may or may not need to splat. You'll have to play with it to see. [Documentation linked for convenience](https://apidock.com/ruby/Hash/dig).

app/models/state_file_base_intake.rb
```diff
@@ -64,19 +64,19 @@ def synchronize_filers_to_database
...
     end
 
     if filing_status_mfj? && direct_file_json_data.spouse_filer.present?
       attributes_to_update.merge!(
-        spouse_first_name: direct_file_json_data.spouse_first_name,
-        spouse_middle_initial: direct_file_json_data.spouse_middle_initial,
-        spouse_last_name: direct_file_json_data.spouse_last_name,
-        spouse_birth_date: direct_file_json_data.spouse_dob
+        spouse_first_name: direct_file_json_data.spouse_filer.first_name,
+        spouse_middle_initial: direct_file_json_data.spouse_filer.middle_initial,
```
powersurge360, line 77: I'm surprised to see this notation. I expected it to be a hash. How do we get to object? Am I overlooking something?

powersurge360 (commented):
Have a question and once answered I'll approve.

app/models/df_json_accessor.rb
mrotondo: awesome - that's exactly what i wanted and just got lazy and didn't look up how to do it/think of using dig :D

app/models/state_file_base_intake.rb
mrotondo: If you're talking about the `spouse_filer.first_name` notation - that's a side effect of using the DfJsonAccessor (soon to be renamed) wrapper, as the `define_method` calls create a method "API" to the underlying data

app/models/df_json_accessor.rb
powersurge360: Oh, I mean in the sense that since:

```ruby
attr_accessor :foo, :bar, :baz
attr_reader :foo, :bar, :baz
attr_writer :foo, :bar, :baz
```

I'd expect to be able to

```ruby
json_accessor foo: { type: :string, key: 'some path' }
```

And the equivalent read/write APIs. But it's consistent with how we're doing xml stuff and I don't want to _block_ on it, necessarily.

app/models/state_file_base_intake.rb
powersurge360: Ah, I see. I was thinking `direct_file_json_data` was the entry point to the json but it's actually an object with many entry points. Cool cool cool.

spec/models/df_json_wrapper_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+end
+
+describe DfJsonWrapper do
+  let(:json) { JSON.parse(json_string) }
+  let(:instance) { DfJsonTest.new(json) }
+
+  describe "#define_json_readers" do
+    context "when the JSON has the correct structure" do
+      let(:json_string) {
+        <<~JSON_STRING
```
powersurge360, line 19: NIT: In some editors, if you use the name of the format for the heredoc (eg, `JSON` here), it will engage the LSP and give goodies like syntax highlighting and even linting/completion

image

powersurge360 (approved):
Seems good. Minor note about heredoc that may not affect anyone but me anyways lol

spec/models/df_json_wrapper_spec.rb
mrotondo: oooh, it works in rubymine too! thanks, that's awesome.

app/models/df_json_accessor.rb
mrotondo: ah, gotcha! yes, that's _much_ cleaner & i've switched over to that syntax. thanks again for the extremely thoughtful review, and for getting into detail about your suggestions (loved the screenshot explaining the heredoc change - i've never used the <<-/<<~ syntax before so it helped me understand what you meant)!

mrotondo: and of course, lmk if i was still not quite understanding what you were thinking

mrotondo: in particular, i've seen you put multiple attribute names on a attr_reader/writer/accessor line before, but i wonder how you feel about that style now that each `json_reader` specification is a whole line unto itself. i switched from having multiple `json_reader`s to just one, but could switch back if you think it'd be clearer

app/models/state_file_base_intake.rb
```diff
@@ -98,9 +98,9 @@ def synchronize_df_dependents_to_database
 
       if dependent_json.present?
         json_attributes = {
-          middle_initial: dependent_json["middleInitial"],
-          relationship: dependent_json["relationship"]&.humanize,
-          dob: dependent_json["dateOfBirth"]
+          middle_initial: dependent_json.middle_initial,
+          relationship: dependent_json.relationship&.humanize,
```
mpidcock, line 102: @mrotondo noting here that there's another PR about to be merged that is removing this `humanize` and to look out for the conflict when merging. CC @jenny-heath

mpidcock: This could be avoided if you also remove it in this PR

app/models/df_json_accessor.rb
powersurge360: I may be inclined to break the hashes up over multiple lines but having them all be one call seems alright to me. Yea, this is exactly the sort of thing I was thinking about and looks good to me. Ultimately, either way you're having to set up a hash of keys -> configuration so may as well make declaring it and registering it a single step.

powersurge360 (approved):
Like the new accessor formatting, looks good.

PR #4856, opened 2024-10-15
Nj 67 disabled exemption

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/67#issue-2553710897

> Acceptance criteria

>GIVEN I have gotten past the screener and the directfile import AND I have not claimed the blind exemption on my federal return
WHEN I am answering NJ specific questions
THEN I see radio buttons that ask me if I was disabled on the last day of the tax year

>GIVEN I have gotten past the screener and the directfile import and I am filing jointly with my spouse/CU partner AND my spouse has not claimed the blind exemption on our federal return
WHEN I am answering NJ specific questions
THEN I see radio buttons that ask me if my spouse was disabled on the last day of the tax year

>GIVEN I have gotten past the screen and the directfile import
WHEN I am answering NJ specific questions
THEN I see helper text explaining the definition of disabled

>GIVEN I am about to submit my state file data
WHEN I see the page that allows me to review my answers
THEN I can go back and edit my disability status

>GIVEN I am about to submit my state file data and I am filing jointly
WHEN I see the page that allows me to review my answers
THEN I can go back and edit my partner's disability status

>GIVEN I have submitted my state return
WHEN I download the PDF
THEN it shows my and my partner's disability status represented by Line 8, "Blind/Disabled" being checked or unchecked for myself and my partner

>GIVEN I have submitted my state return
WHEN I view the XML
THEN it shows my and my partner's disability status

>GIVEN I have claimed the disability exemption
WHEN I view Line 13 and 30 (Total Exemption Amount) in the PDF and XML
THEN my disability exemption amount is reflected in this total sum

## Is PM acceptance required?
- Yes - don't merge until JIRA issue is accepted!

## What was done?
- Added intake question to check if primary or spouse are disabled
- Updated logic in calculator to update calculated exemption when primary and/or spouse are disabled
- Updated logic in XML submission builder to check Line 8 checkboxes if primary and/or spouse are disabled

## How to test?
  - Unit/integration/local testing
  - Created minimal import, validated that disabled question appears, it is required, it shows up on the review screen, and that it populates the XML and the PDF correctly when yes/no is selected
  - Created married filing jointly import, validated that disabled questions appears for primary and spouse, both fields are required, it shows up on the review screen, and that it populates the XML and the PDF correctly when yes/no is selected
  - Created blind_primary import, validated that disabled question does not appear and that the XML and PDF are populated correctly
[...]

PR #4869, opened 2024-10-17
Nj 75 medical

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/75

## What was done?
- Add medical expenses column
- Add calculated medical expenses deduction
- Include medical expenses deduction in line 38

There are a few issues that need to be resolved in this PR or a subsequent one:
- ~The help text and label overlapping (we could do a quick fix with css obviously, but I would rather figure out why this is happening and fix that)~ Saw that this is also corrected in honeycrisp compact, so I corrected it for the whole application
- ~The placeholder text returning to 0.00 even though I gave an empty placeholder to the currency field~ should be fixed
- We need to update honeycrisp so we can use the accessible reveal component here and elsewhere

## How to test?
Select any directfile import and proceed to the medical expenses screen. Enter a value that is greater than 2% of gross income. When you get to the review screen, you should see medical expenses and be able to edit them.

A note for screen reader testing: one change I made here is on the review screen, where I included sr-only text to distinguish the medical expenses edit link from the others. We can use this pattern going forward and change it when we do the sundry a11y bugs ticket. The review screen should also use headers instead of styled text to indicate sections.

https://github.com/user-attachments/assets/51d9a3d5-b81d-43c4-99e9-543ff8b06566
Screenshot 2024-10-17 at 11 35 56 AM

### DIFF
--- a/app/controllers/state_file/questions/nj_medical_expenses_controller.rb
+++ b/app/controllers/state_file/questions/nj_medical_expenses_controller.rb
@@ -1,0 +1,9 @@
+module StateFile
+  module Questions
+    class NjMedicalExpensesController < QuestionsController
+      include ReturnToReviewConcern
+
+      before_action -> { @filing_year = Rails.configuration.statefile_current_tax_year }
+    end
+  end
+end
--- a/app/forms/state_file/nj_medical_expenses_form.rb
+++ b/app/forms/state_file/nj_medical_expenses_form.rb
@@ -1,0 +1,15 @@
+module StateFile
+  class NjMedicalExpensesForm < QuestionsForm
+    set_attributes_for :intake,
+                       :medical_expenses
+
+    validates_numericality_of :medical_expenses, only_integer: true, message: :round_to_whole_number, if: -> { medical_expenses.present? }
+    validates :medical_expenses, presence: true, allow_blank: false, numericality: { greater_than_or_equal_to: 0 }, if: -> { medical_expenses.present? }
+                   
+    def save
+      unless medical_expenses.nil?
+        @intake.update(attributes_for(:intake))
+      end
+    end
+  end
+end
--- a/app/forms/state_file/nj_medical_expenses_form.rb
+++ b/app/forms/state_file/nj_medical_expenses_form.rb
@@ -3,8 +3,8 @@ module StateFile
     set_attributes_for :intake,
                        :medical_expenses
 
-    validates_numericality_of :medical_expenses, only_integer: true, message: :round_to_whole_number, if: -> { medical_expenses.present? }
-    validates :medical_expenses, presence: true, allow_blank: false, numericality: { greater_than_or_equal_to: 0 }, if: -> { medical_expenses.present? }
+    validates_numericality_of :medical_expenses, only_integer: true, message: :round_to_whole_number
+    validates :medical_expenses, presence: true, allow_blank: false, numericality: { greater_than_or_equal_to: 0 }
                    
     def save
       unless medical_expenses.nil?
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -20,6 +20,7 @@ module Efile
         set_line(:NJ1040_LINE_15, :calculate_line_15)
         set_line(:NJ1040_LINE_27, :calculate_line_27)
         set_line(:NJ1040_LINE_29, :calculate_line_29)
+        set_line(:NJ1040_LINE_31, :calculate_line_31)
         set_line(:NJ1040_LINE_38, :calculate_line_38)
         set_line(:NJ1040_LINE_39, :calculate_line_39)
         set_line(:NJ1040_LINE_40A, :calculate_line_40a)
@@ -186,8 +187,16 @@ module Efile
         calculate_line_27
       end
 
+      def calculate_line_31
+        two_percent_gross = calculate_line_29 * 0.02
+        difference_with_med_expenses = @intake.medical_expenses - two_percent_gross
+        rounded_difference = difference_with_med_expenses.round
+        return rounded_difference if rounded_difference.positive?
+        nil
+      end
+
       def calculate_line_38
-        calculate_line_13
+        calculate_line_13 + line_or_zero(:NJ1040_LINE_31)
       end
 
       def calculate_line_39
--- a/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
+++ b/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
@@ -6,7 +6,7 @@
 <% content_for :card do %>
   <h1 class="h2"><%= t(".title") %></h1>
   <ul class="list--bulleted">
-    <%= t(".description") %>
+    <%= t(".description_html") %>
   </ul>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
--- a/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
+++ b/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
@@ -1,0 +1,41 @@
+<%
+  title = t(".title", filing_year: @filing_year)
+  content_for :page_title, title
+%>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= t(".title") %></h1>
+  <ul class="list--bulleted">
+    <%= t(".description") %>
+  </ul>
+
+  <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
+    <div class="white-group">
+      <%= f.vita_min_money_field(
+        :medical_expenses,
+        t('.label', filing_year: @filing_year),
+        help_text: t('.help_text'),
+        classes: ["form-width--long"],
+        options: { placeholder: "" }
+    ) %>
+    </div>
+
+    <div class="reveal">
+        <h2 style="margin-bottom: 0">
+            <a class="reveal__link" aria-controls="nj-medical-details" aria-expanded="false">
+                <%= t('.learn_more_title') %>
+            </a>
+        </h2>
+        <div class="reveal__content" id="nj-medical-details">
+            <p><%= t('.learn_more_meaning') %>:</p>
+            <ul class="list--bulleted"><%= t('.learn_more_content_html') %></ul>
+            <p><%= t('.learn_more_note_html') %></p>
+        </div>
+    </div>
+
+    <% if params[:return_to_review].present? %>
+      <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
+    <% end %>
+    <%= f.continue %>
+  <% end %>
+<% end %>
--- a/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
+++ b/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
@@ -5,9 +5,7 @@
 
 <% content_for :card do %>
   <h1 class="h2"><%= t(".title") %></h1>
-  <ul class="list--bulleted">
-    <%= t(".description_html") %>
-  </ul>
+  <%= t(".description_html") %>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
     <div class="white-group">
--- a/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
+++ b/app/views/state_file/questions/nj_medical_expenses/edit.html.erb
@@ -18,6 +18,8 @@
     ) %>
     </div>
 
+    <p><%= t(".do_not_claim") %></p>
+
     <div class="reveal">
         <h2 style="margin-bottom: 0">
             <a class="reveal__link" aria-controls="nj-medical-details" aria-expanded="false">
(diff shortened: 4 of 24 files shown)


app/views/state_file/questions/nj_medical_expenses/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+  <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
+    <div class="white-group">
+      <%= f.vita_min_money_field(
+        :medical_expenses,
+        t('.label', filing_year: @filing_year),
+        help_text: t('.help_text'),
+        classes: ["form-width--long"],
+        options: { placeholder: "" }
+        # TODO: WHY DOES HELP TEXT OVERLAP LABEL?
```
aloverso, line 22: [pebble] If we're okay with merging and fixing these todos in the future (which I'm assuming from the fact that this is in review?) I'd like for each of these TODO comments to be tracked in backlog tickets and not in comments in the code. It's unlikely we're going to be coming back to this file and seeing these in the future

app/views/state_file/questions/nj_medical_expenses/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  title = t(".title", filing_year: @filing_year)
+  content_for :page_title, title
+%>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= t(".title") %></h1>
+  <ul class="list--bulleted">
+    <li><%= t(".includes_whom") %></li>
+    <li><%= t(".percent_income") %></li>
+    <li><%= t(".no_reimbursed") %></li>
```
aloverso, line 11: [pebble] You can put this bullet list in a single content item, which would probably be easier to read and edit than tracking them as independent content items. See `terms_and_conditions.edit.ny_text_html` as an example

app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -46,5 +46,16 @@
     </div>
   <% end %>
 
+  <div id="rent_paid" class="white-group">
```
aloverso, line 49: [pebble] copy paste error "rent paid"

config/locales/en.yml
```diff
@@ -2586,6 +2586,20 @@ en:
         edit:
           helper_description: Please continue to see what else we can guide you to qualify for.
           title: Looks like you’re not eligible for this credit.
+      nj_medical_expenses:
+        edit:
+          help_text: If you do not want to claim this deduction, click “Continue”.
+          includes_whom: This includes expenses for yourself, your spouse, and your dependents.
+          label: Non-reimbursed medical expenses paid in %{filing_year}
+          learn_more_content_html: "<li>Physicians, dental, and other medical fees</li><li>Prescription eyeglasses and contact lenses</li><li>Hospital care</li><li>Nursing care</li><li>Medicines and drugs</li><li>Prosthetic devices</li><li>X‑rays and other diagnostic services conducted by or directed by a physician or dentist</li><li>Amounts paid for transportation primarily for and essential to medical care</li><li>Insurance (including amounts paid as premiums under Part B of Title XVIII of the Social Security Act, relating to supplementary medical insurance for the aged) covering medical care</li>"
```
aloverso, line 2594: [dust/pebble] Would be nice to format this using the `|` so that each `li` can be its only line in the yaml

db/migrate/20241007210254_add_medical_expenses_to_nj_intake.rb
```diff
@@ -0,0 +1,5 @@
+class AddMedicalExpensesToNjIntake < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_nj_intakes, :medical_expenses, :integer, default: 0, null: false
```
aloverso, line 3: [dust] I see we're doing integer here for this money field, which is fine given our discussion yesterday but will be another field to track for updating if we switch to accepting non-rounded numbers?

spec/forms/state_file/nj_medical_expenses_form_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  describe ".save" do
+    let(:intake) {
+      create :state_file_nj_intake, medical_expenses: 0
+    }
+    let(:form) { described_class.new(intake, valid_params) }
+
+    context "when saving medical expenses" do
+      let(:valid_params) do
+        { medical_expenses: 12345 }
```
aloverso, line 53: [dust] Not sure if it matters but just noting two different representations of params in the same file. `{ medical_expenses: 12345 }` here and using symbol/rocket `{ :medical_expenses => "-123" }` up above. Might be nice to be consistent. I have a slight preference for the syntax here if it's all the same anyways.

app/forms/state_file/nj_medical_expenses_form.rb
```diff
@@ -0,0 +1,15 @@
+module StateFile
+  class NjMedicalExpensesForm < QuestionsForm
+    set_attributes_for :intake,
+                       :medical_expenses
+
+    validates_numericality_of :medical_expenses, only_integer: true, message: :round_to_whole_number, if: -> { medical_expenses.present? }
+    validates :medical_expenses, presence: true, allow_blank: false, numericality: { greater_than_or_equal_to: 0 }, if: -> { medical_expenses.present? }
```
aloverso, line 7: [dust] I'm a little confused by the validating presence-true if present. It seems like nil is a valid value here, ie, they could move forward with an empty field. Is that correct?

If so, might be nice to have an explicit test to document that nil is a valid value

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -163,6 +163,49 @@ def over_65_birth_year
...
+      it 'sets line 31 to the entire cost of medical expenses' do
+        expect(instance.lines[:NJ1040_LINE_31].value).to eq(123_456)
+      end
+
+      it 'includes medical expenses in line 38' do
+        expect(instance.lines[:NJ1040_LINE_38].value).to eq(124_690)
+      end
+    end
+
+    context 'when gross income is 10_000' do
```
aloverso, line 186: [pebble] can we reframe these test context names to describe the case it represents rather than the specific value? From reading this, I'm not sure why 10,000 and 100,000,000 have this behavior. Looking at the ticket, I assume it's when income * .02 is less than the medical expenses?

Can we:
a) rephrase the context strings to explain this, ie "when medical expenses is greater than 2% of income" or something like that
b) test the edge cases, like when medical expenses is exactly the same as 2% income, one dollar above, and one dollar below

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -899,6 +899,54 @@
...
+    describe "line 31 medical expenses" do
+      context "with a gross income of 200k" do
+        let(:submission) {
+          create :efile_submission, tax_return: nil, data_source: create(
+            :state_file_nj_intake,
+            :df_data_many_w2s,
+            medical_expenses: 567_890
+          )
+        }
+        it "writes sum $563,890.00 to fill boxes on line 31" do
```
aloverso, line 911: [dust] not to be a pedant but technically this is not a sum, it's a subtraction?

aloverso (commented):
Looks good! Left a couple comments, and looks like a couple failing tests and `line_data` needs to be updated.

app/forms/state_file/nj_medical_expenses_form.rb
jachan: The way I understand it - `if medical_expenses.present?` THEN `presence: true, allow_blank: false, greater_than_or_equal_to: 0`. Melanie can you confirm?

I tested something similar here: https://github.com/codeforamerica/vita-min/pull/4856/files#diff-62e5f3eca04b36aa7e3dd487d951844767312fbae9e56015100adc09a80f5a57, and Kevin from CFA recommended [shoulda-matchers](https://github.com/thoughtbot/shoulda-matchers) as a good way to test these.

app/views/state_file/questions/nj_medical_expenses/edit.html.erb
jachan: +1, one thing I've done in the past is having the backlog ticket in the comment itself

`# TODO github.com/issue/whatever`

config/locales/es.yml
```diff
@@ -2560,6 +2560,20 @@ es:
         edit:
           helper_description: Continúe viendo para qué más podemos ayudarlo a calificar.
           title: Parece que no eres elegible para este crédito.
+      nj_medical_expenses:
+        edit:
+          help_text: If you do not want to claim this deduction, click “Continue”.
```
jachan, line 2565: [SAND] Is it possible to stick all this in Google Translate? I've been doing that for all my placeholder Spanish text. Does it matter, since we're replacing it in the future anyway?

db/migrate/20241007210254_add_medical_expenses_to_nj_intake.rb
jachan: @aloverso @mmazanec22 do we have a ticket/notepad tracking where we're using non-rounded numbers somewhere?

spec/controllers/state_file/questions/nj_medical_expenses.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  describe "#edit" do
+    render_views
+    it 'succeeds' do
+      get :edit
+      expect(response).to be_successful
+    end
+
+    context "when a user has medical expenses" do
```
jachan, line 16: [SAND] This should be in a `describe "#update"` block, right?

app/controllers/state_file/questions/nj_medical_expenses_controller.rb
```diff
@@ -0,0 +1,9 @@
+module StateFile
+  module Questions
+    class NjMedicalExpensesController < QuestionsController
+      include ReturnToReviewConcern
+
+      before_action -> { @filing_year = Rails.configuration.statefile_current_tax_year }
```
jachan, line 6: [DUST] Is @filing_year being used at all in the page?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -97,8 +98,15 @@ def calculate_line_29
...
+      def calculate_line_31
+        two_percent_gross = calculate_line_29 * 0.02
+        difference_with_med_expenses = @intake.medical_expenses - two_percent_gross
+        return difference_with_med_expenses.round if difference_with_med_expenses.positive?
+        nil
+      end
+
       def calculate_line_38
-        calculate_line_13
+        calculate_line_13 + line_or_zero(:NJ1040_LINE_31)
```
jachan, line 109: [DUST] Thanks for updating this!

app/controllers/state_file/questions/nj_medical_expenses_controller.rb
jachan: Nvm I see it is used on line 18 of [app/views/state_file/questions/nj_medical_expenses/edit.html.erb](https://github.com/codeforamerica/vita-min/pull/4869/files#diff-260d83dc4b3add3a84c8feaab5e058bc20cb28d5f653465bb41406f32e1195fe)

jachan (approved):
A few code style things that should be fixed before merging.

app/views/state_file/questions/nj_review/edit.html.erb
mmazanec22: Thanks for the catch. Any idea why we have ids on these anyway?

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mmazanec22: Subtraction is just a sum with a negative number 😛

db/migrate/20241007210254_add_medical_expenses_to_nj_intake.rb
mmazanec22: Made a ticket https://github.com/newjersey/affordability-pm/issues/98

app/forms/state_file/nj_medical_expenses_form.rb
mmazanec22: Currently this defaults to 0. This could be a good idea because of the requirement to show medical expenses on the review screen regardless of value and let the user go back and edit.

However, we can change it to allow blank, default to nil, and just show a 0 on the review screen if there is no value.

Maybe we should make some blanket decisions about defaults vs nil and add them to our internal eng practices doc?

config/locales/es.yml
mmazanec22: Per our slack conversation, leaving this for now. Added this to the translation ticket in the backlog

app/views/state_file/questions/nj_medical_expenses/edit.html.erb
aloverso: I would very much prefer to not have the comment exist at all. Not sure having the comment reference a GH issue will improve the usefulness here

app/forms/state_file/nj_medical_expenses_form.rb
mpidcock: [dust] I am also confused why you have `validates :medical_expenses, presence: true ... if: -> { medical_expenses.present? }` - is this just "validate true if true"?

I think we have also been wondering about when to allow blank vs requiring a 0 and it's more challenging than I expected to make a blanket decision so we often handle it on a case by case basis :/

mpidcock (approved):
Just the tiniest comment, otherwise looks great!

PR #4870, opened 2024-10-17
[FYST-523] Add md screener questions

squanto:
## Link to pivotal/JIRA issue
-  https://codeforamerica.atlassian.net/browse/FYST-523

## What was done?
- Add Maryland screener questions
## How to test?
  - Added integration specs for screener questions

Screenshot 2024-10-16 at 3 14 09 PM


squanto:
note - restarted pr from this (seemingly) unrelated error https://github.com/codeforamerica/vita-min/actions/runs/11372349014/job/31641989119

app/models/state_file_md_intake.rb
```diff
@@ -75,6 +79,14 @@
 class StateFileMdIntake < StateFileBaseIntake
   encrypts :account_number, :routing_number, :raw_direct_file_data, :raw_direct_file_intake_data
 
+  enum eligibility_lived_in_state: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_lived_in_state
+  enum eligibility_out_of_state_income: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_out_of_state_income
```
mrotondo, line 83: we should use a consistent type for the `_prefix` parameter to `enum`. i know i suggested passing `true` because it has the same effect as passing the name of the enum & is less repetitive/DRYer, but it's also inconsistent with the rest of the codebase. for that reason, i think it's probably worth changing lines 84-87 to pass the name of the enum as the `_prefix` argument, as lines 82-83 do

spec/factories/state_file_md_intakes.rb
```diff
@@ -83,6 +87,8 @@
     primary_middle_initial { "A" }
     primary_last_name { "Lando" }
 
+    eligibility_filing_status_mfj { "unfilled" }
```
mrotondo, line 90: is this necessary, given that "unfilled" is the default?

mrotondo (approved):
Looks good! Made a few minor comments but it can go in after you take a look at those.

spec/factories/state_file_md_intakes.rb
squanto: that's true, it's not strictly necessary, updated

app/models/state_file_md_intake.rb
squanto: Good catch, updated to pass the enum name to prefix

squanto:
Updated the eligibility checks for Q2 and Q3, with the following behavior:

[Watch the video](https://github.com/user-attachments/assets/954db27b-f5ee-4fce-be83-0ea3f720015a)

mpidcock (commented):
Hi @squanto please take a look at the test failures before merging. Ideally these should be passing before code review & acceptance testing, but in this case it does look like it's a small translations related fix.

image

PR #4886, opened 2024-10-18
FYST-907 W-2 refactor: creating W-2s on DF import and add to income review page

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-907

## What was done?
- add method to copy all W2s from df xml into db on import
- W2 controller 
  - remove from navigation flows
  - remove index action from
  - remove concept of "invalid" W2s
  - just keep edit and update
- remove W2 section from final review page
- add W2s to new income review page
  - summary of each includes employee and employer name so added new columns to state w2
  - each summary links to the edit page for that w2 (all will have records in db already from import)
- in tax calculator, instead of combining state file w2s from the db with w2 xml from df, reference db only because we now copy them all over on import
## How to test?
- Controllers are tested (income review looks for list in view, w2 edit and update are tested)
- Import copying is tested (method is tested in model spec, method is mocked in job spec)
- Some calculator (or downstream of calculator) tests had to be changed to accommodate new db-only w2 references

- Before
- After
Screenshot 2024-10-17 at 3 15 27 PM


db/migrate/20241017215227_add_columns_to_state_file_w2.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumnsToStateFileW2 < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_w2s, :employer_name, :string
+    add_column :state_file_w2s, :employee_name, :string
```
arinchoi03, line 4: as discussed, these should never be `null` -- `EmployeeNm` and `EmployerName` seem to be required fields on the federal IRS schema

app/models/state_file_base_intake.rb
```diff
@@ -118,6 +119,25 @@ def synchronize_df_1099_rs_to_database
     end
   end
 
+  def synchronize_df_w2s_to_database
+    direct_file_data.w2s.each_with_index do |direct_file_w2, i|
+      state_file_w2 = state_file_w2s[i] || state_file_w2s.build
```
arinchoi03, line 124: are we certain that `state_file_w2s` will be indexed the same way as the order of w2s in xml? Do we need to do any kind of "matching" like we do on dependents? Maybe on employerStateIdNum and EmployeeNm?

app/controllers/state_file/questions/w2_controller.rb
```diff
@@ -1,113 +1,34 @@
...
     class W2Controller < QuestionsController
-      include ReturnToReviewConcern
-      before_action :load_w2s
-      before_action :load_w2, only: [:edit, :update]
+      before_action :load_w2
 
-      def self.show?(intake)
-        invalid_df_w2s(intake).any?
+      def self.show?(_)
+        false
```
arinchoi03, line 7: can we just remove W2Controller from the `_navigation.rb` files which would let us not have to define this `show` method?

app/models/state_file_base_intake.rb
jenny-heath: oh good catch! we actually have a column on state_file_w2 called w2_index that we can use here that corresponds to the index of a w2 in the direct file xml

app/controllers/state_file/questions/w2_controller.rb
jenny-heath: i forgot to remove this!

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -31,6 +31,7 @@ def perform(authorization_code:, intake:)
 
         intake.synchronize_df_dependents_to_database
         intake.synchronize_df_1099_rs_to_database
+        intake.synchronize_df_w2s_to_database
```
mpidcock, line 34: This should also be called on `federal_info_form #save`, along with those other synchronize methods. Mike had pointed out that it's probably time to make a `synchronize_df_data_to_database` method that calls all of these since they are called in multiple places.

app/models/df_w2_accessor.rb
```diff
@@ -1,6 +1,8 @@
 class DfW2Accessor < DfXmlAccessor
   SELECTORS = {
     EmployeeSSN: 'EmployeeSSN',
+    EmployeeNm: 'EmployeeNm',
+    EmployerNm: 'EmployerName BusinessNameLine1Txt',
```
mpidcock, line 5: [DUST] I just saw that `EmployerNm` is duplicate with `EmployerName` that was already added on line 7. Ideally should keep the one that's there and consider changing `EmployeeNm` to `EmployeeName`.

spec/models/efile_error_spec.rb
```diff
@@ -29,8 +29,8 @@
   end
 
   it 'converts paths to controllers' do
-    controller = EfileError.path_to_controller("w2")
-    expect(controller).to eq StateFile::Questions::W2Controller
+    controller = EfileError.path_to_controller("name-dob")
+    expect(controller).to eq StateFile::Questions::NameDobController
```
mpidcock, line 33: the name-dob page's days are limited - could you choose a different controller? "income-review" should be reliable

mpidcock: oh, I didn't notice it was used in other tests too. nevermind, that can all be fixed when the page is removed

spec/factories/state_file_w2s.rb
mpidcock, line 1: [Sand] I'm noting here that we discussed adding `synchronize_df_w2s_to_database` to the `state_file_xx_intakes` FactoryBots, which might be nice in particular for NJ, but it's acceptable to me to address that in a follow-up.

app/models/state_file_w2.rb
```diff
@@ -3,6 +3,9 @@
 # Table name: state_file_w2s
 #
 #  id                          :bigint           not null, primary key
+#  employee_name               :string           not null
+#  employee_ssn                :string
+#  employer_name               :string           not null
```
mpidcock, line 8: [Sand] Noting that we discussed adding `WagesAmt` and `WithholdingAmt` here as well, but fine to leave for now and follow up later.

mpidcock (approved):
I left some non-blocking comments and did not thoroughly review the NJ tests. Really excited to see this happening.

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -265,19 +265,19 @@
     end
 
     describe "wages" do
-      context "when no w2 wages (line 15 is -1)" do
-        let(:intake) { create(:state_file_nj_intake, :df_data_minimal) }
+      let(:intake) { create(:state_file_nj_intake) }
 
+      context "when no w2 wages (line 15 is -1)" do
         it "does not include WagesSalariesTips item" do
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_15).and_return -1
```
jachan, line 272: [DUST] Seeing a change in testing strategy here. Previously we would create the intake and then let the calculator run before checking the XML, now we're mocking out the calculator values and validating that the XML is generated properly. Personally, I'm fine with this, since we test the calculator separately in its own spec, we don't need to have "end-to-end" testing here.

@mmazanec22 @aloverso worth looking at!

jachan (commented):
Hi! Thanks for helping with this refactor, it should make it easier to reference W2 data in the future. I just looked at files beginning with `nj`, left a few comments on testing strategy.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -307,48 +307,51 @@ def over_65_birth_year
...
 
   describe 'line 27 - total income' do
-    let(:intake) { create(:state_file_nj_intake, :df_data_2_w2s) }
+    let(:intake) { create(:state_file_nj_intake) }
+
     it 'sets line 27 to the sum of all state wage amounts' do
-      line_15_w2_wages = 12345 + 50000
-      expect(instance.lines[:NJ1040_LINE_15].value).to eq(line_15_w2_wages)
-      expect(instance.lines[:NJ1040_LINE_27].value).to eq(line_15_w2_wages)
+      allow(instance).to receive(:calculate_line_15).and_return 50000
```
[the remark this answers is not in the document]
jenny-heath: yup, the other way to set up the test is to create `state_file_w2`s. i generally prefer to mock calculator methods in tests when they're not the thing that's being explicitly tested, though, in part because it's just easier to understand the test setup. since the calculation for line 15 is tested, i think it would be redundant to recreate the db records just to check that the output for calculate_line_15 gets copied over to line 27, but maybe i'm missing something! happy to tuple about it or just refactor the test if that's the strategy y'all prefer

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -641,6 +641,8 @@
         }
 
         it "includes the sum 200,000 split into each box on line 15" do
+          submission.data_source.synchronize_df_w2s_to_database
```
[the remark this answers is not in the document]
jenny-heath: i think so, but open to other approaches? i landed on this because it was easier to call the synchronize method than to manually create the w2 records based on the fixture contents. i think it would also be acceptable to create state file w2s from the factory

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
jachan: Got it! I agree that mocking out the calculator is the right way to go for the XML builder spec. This is something that we've discussed internally as a team, but never took the time to refactor previous tests.

I have no objections here, just flagging for the team to look over.

spec/lib/efile/nj/nj1040_calculator_spec.rb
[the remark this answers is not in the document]
jachan: Ah gotcha. I reviewed the business logic for line 27, which appears to be the total income field, so as we add additional income sources, mocking each of them out and validating that `calculate_line_27` sums them all seems appropriate here.

And it looks like we follow a similar approach elsewhere in the spec for mocking out other lines in the calculator. I'm fine with the existing approach!

https://github.com/codeforamerica/vita-min/blob/main/spec/lib/efile/nj/nj1040_calculator_spec.rb#L510

jachan (approved):
Thanks for talking through the testing strategy with me!

spec/lib/pdf_filler/nj1040_pdf_spec.rb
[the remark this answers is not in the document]
aloverso: Could this be in a before each for every test in this file?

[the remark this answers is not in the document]
jenny-heath: @aloverso i was also thinking that i didn't love calling this method in a bunch of places so impulsively did an experiment at the end of yesterday: https://github.com/codeforamerica/vita-min/pull/4886/commits/a504b987e3f7351a777607c0e6a4db1d7291390b

let me know what you think!

spec/fixtures/state_file/persona_approved_outputs/2023/az/johnny_return_xmls/xml/submission.xml
```diff
@@ -263,18 +263,17 @@
...
+  <StateAbbreviationCd>AZ</StateAbbreviationCd>
+  <EmployerStateIdNum>12345</EmployerStateIdNum>
+  <StateWagesAmt>45000</StateWagesAmt>
+  <StateIncomeTaxAmt>400</StateIncomeTaxAmt>
+  <W2LocalTaxGrp>
+    <LocalWagesAndTipsAmt>0</LocalWagesAndTipsAmt>
+    <LocalIncomeTaxAmt>0</LocalIncomeTaxAmt>
+  </W2LocalTaxGrp>
+</W2StateTaxGrp>
+</W2StateLocalTaxGrp>
```
jenny-heath, line 276: i think i tried a little too hard and wasted a little too much time trying to make this output the w2s with the right formatting because the contents haven't changed. potentially not worth it? i'm wondering:
A) have we just never built the persona outputs with combined w2s (did they all come from the df xml? i think maybe they did) which is why this hasn't come up
B) should i manually reformat it or is that dumb because the next person who regenerates these will have to do it too

PR #4889, opened 2024-10-18
Nj 90 external links

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/90

## What was done?
Copy the [USWDS external link indicator](https://designsystem.digital.gov/components/link/) pattern to visually show which links go to other sites and add sr-only text that indicates external links and links that open in new tabs, as recommended by USWDS, [W3C](https://www.w3.org/TR/WCAG20-TECHS/G201.html), and [Deque](https://dequeuniversity.com/checklists/web/links).

This PR marks any link that starts with http as external, excluding any link containing fileyourstatetaxes.org. We could add more urls to the exclusion list, or we could make this a class and manually apply it to every external link. The latter approach seems like it might be tedious and prone to being forgotten.

This PR also corrects the z index of the skip nav link to make it visible on top of the header.

## How to test?
Check out the footer links

Screenshot 2024-10-18 at 5 26 10 PM

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -70,6 +70,7 @@
 @import "components/info-box";
 @import "components/form-cards";
 @import "components/offboarding";
+@import "components/external-link";
 @import "pages/form_13614c";
 @import "pages/provider-list";
 @import "pages/provider-details";
@@ -243,6 +244,10 @@ main {
   @include sr-only;
 }
 
+a[href^='http']:not([href*="fileyourstatetaxes.org" i]) {
+  @include external-link;
+}
+
 ul.with-bullets {
   margin-left: 20px;
   list-style-type: disc;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -243,6 +244,10 @@ main {
   @include sr-only;
 }
 
+a[href^='http']:not([href*="fileyourstatetaxes.org" i]) {
```
jachan, line 247: [DUST] CFA, is this a sufficient list of external links?

aloverso (approved):
lgtm

jachan (approved):
Cool to change this in styling so it's taken care of for all future links! Left a comment for CFA to address.

app/assets/stylesheets/application.scss
mpidcock: In order to support GYR as well I'd like to parameterize this based on the request domain. I tested some steps to follow:
1. add a definition to `application_helper.rb`
```rb
  def request_domain
    request.domain
  end
```
2. update the logic: `a[href^='http']:not([href*=request_domain i])`

I think that should be sufficient to catch external links.

mpidcock (approved):

PR #4892, opened 2024-10-18
Fyst 548/MD State ID Screen and XML

afiakyalo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-548

## What was done?
- Added MD State ID screen for primary and spouse
## How to test?
  - Added unit tests
  - Manually test with single file and mfj personas
Screenshot 2024-10-18 at 6 17 08 PM
Screenshot 2024-10-18 at 6 16 57 PM


app/controllers/state_file/questions/md_primary_state_id_controller.rb
```diff
@@ -0,0 +1,8 @@
+module StateFile
+	module Questions
+		class MdPrimaryStateIdController < QuestionsController
+			include ReturnToReviewConcern
+		end
+	end
+end
+
```
anisharamnani, line 8: 🤔 this is the _tiniest_ of things, but i think most of the codebase is 2 spaces instead of 4.

anisharamnani:
I had the tiniest suggestion & don't feel qualified to approve, but this looks great! I learned a lot reviewing this PR.

app/controllers/state_file/questions/md_primary_state_id_controller.rb
afiakyalo: VScode was messing me up.

app/views/state_file/questions/md_primary_state_id/edit.html.erb
```diff
@@ -0,0 +1,24 @@
+<% title = t("state_file.questions.primary_state_id.edit.title") %>
+<% content_for :page_title, title %>
+
+<%# Must be above :card content_for %>
```
arinchoi03, line 4: hmm not sure why this comment is needed? what is the reason this needs to be above?
`content_for :card`

spec/controllers/state_file/questions/md_spouse_state_id_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    context "with an intake that does not reports married filed jointly" do
+      let!(:intake) { create :state_file_nc_intake, filing_status: :single}
+
+      it "returns false" do
+        expect(described_class.show?(intake)).to eq false
+      end
+    end
+  end
```
arinchoi03, line 20: should this also have a similar `#edit` action test as well (similar to what is in `MdPrimaryStateIdController`)?

app/views/state_file/questions/md_spouse_state_id/edit.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <%= t('.encourage_sharing') %>
+        </p>
+      </div>
+    </div>
+  </div>
+<% end %>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= title %></h1>
+  <%= render 'state_file/questions/ny_primary_state_id/state_id', options: {} %>
```
arinchoi03, line 23: since this form is used for all states, can we think about moving this into a more general state-specific folder?

arinchoi03 (commented):
some thoughts! great work ✨

app/views/state_file/questions/md_primary_state_id/edit.html.erb
afiakyalo: For some reason it gets hidden if it's not above. The comment helps to provide that context.

app/views/state_file/questions/primary_state_id/edit.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if current_state_code == "az" %>
+  <% render 'state_file/questions/az_primary_state_id/az_primary' %>
+<% elsif current_state_code == "md" %>
+  <% render 'state_file/questions/md_primary_state_id/md_primary' %>
+<% elsif current_state_code == "nc" %>
+  <% render 'state_file/questions/nc_primary_state_id/nc_primary' %>
+<% end %>
```
jnf, line 7: [pebble -- avoiding state-code dependent conditionals] could we deprecate this file by assigning the partial/view to render in the controller action?

app/views/state_file/questions/spouse_state_id/edit.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if current_state_code == "az" %>
+  <% render 'state_file/questions/az_spouse_state_id/az_spouse' %>
+<% elsif current_state_code == "md" %>
+  <% render 'state_file/questions/md_spouse_state_id/md_spouse' %>
+<% elsif current_state_code == "nc" %>
+  <% render 'state_file/questions/nc_spouse_state_id/nc_spouse' %>
+<% end %>
```
jnf, line 7: ditto above comment about controller action.

app/controllers/state_file/questions/primary_state_id_controller.rb
```diff
@@ -1,6 +1,6 @@
 module StateFile
   module Questions
-    class AzPrimaryStateIdController < QuestionsController
+    class PrimaryStateIdController < QuestionsController
```
jnf, line 3: [pebble-ish, i guess?] given this is a shared place now, is it better to have differentiating logic in the controller instead of the view? For example, we could do

```ruby
class PrimaryStateIdController < QuestionsController
  include ReturnToReviewConcern

  def edit
    super
    render "state_file/questions/#{current_state_code}_primary_state_id/_#{current_state_code}_primary" 
  end
end
```

i think that would remove the need for the views that just have if/elsif blocks, which i much prefer (both fewer view files and fewer conditionals). But I am not sure if it adds complication in the controller/form flow. I don't _think_ it does, but I'm not the expert on the controller magic.

jnf (changes requested):
requesting info and maybe a small cleanup

app/controllers/state_file/questions/primary_state_id_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        if @form.valid?
+          @form.save
+          after_update_success
+          track_question_answer
+          redirect_to(next_path)
+        else
+          after_update_failure
+          track_validation_error
+          render "state_file/questions/#{current_state_code}_primary_state_id/_#{current_state_code}_primary"
+        end
```
jnf, line 22: [dust] oof, is thia side effect of pulling the render into the controller? sorry about that. :/

tahsinaislam: i dont hate it ?

mrotondo: we could extract lines 8 & 21 into a a method named something like `state_specific_view`

mrotondo (commented):
hey @tahsinaislam i'm hopping into pairing so i only had time to leave one comment

jnf (approved):
this looks good to me.

app/views/state_file/questions/md_primary_state_id/edit.html.erb
powersurge360: It's because the file executes in render order. If the content for this bit isn't provided by the time the partial is rendered, it will simply render without it.

PR #4894, opened 2024-10-21
Remove unemployment from NJ navigation

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/99

## What was done?
Remove unemployment from NJ navigation controller

## How to test?
Go through the flow, observe the lack of unemployment question

### DIFF
--- a/app/views/state_file/landing_page/edit.html.erb
+++ b/app/views/state_file/landing_page/edit.html.erb
@@ -65,7 +65,7 @@
               <% end %>
             <% else %>
               <div class="white-group spacing-below-60">
-                <%= t(".help_text_html") %>
+                <%= t(".#{@state_code}.help_text_html", default: t(".help_text_html")) %>
               </div>
               <%= form_with model: @form, url: { action: :update }, local: true, method: :put, builder: VitaMinFormBuilder do |f| %>
                 <%= f.submit t("general.get_started"), class: "button button--primary button--wide", id: "firstCta" %>
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -2076,6 +2076,15 @@ es:
           closed_html: |
             <strong>FileYourStateTaxes</strong> se creó en asociación con el New Jersey Office of Innovation para integrarse con la herramienta Direct File del IRS.<br /><br />
             Estamos cerrados por la temporada de impuestos. Desafortunadamente no puede presentar su declaración estatal con nosotros este año.
+          help_text_html: |
+            <strong>Lo que necesitarás para comenzar</strong>
+            <br />
+            <ul class="list--bulleted">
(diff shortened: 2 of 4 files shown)


config/locales/es.yml
```diff
@@ -2047,7 +2047,7 @@ es:
             <li>Una declaración federal de 2023 <strong>aceptada</strong> mediante <a href="https://directfile.irs.gov/">IRS Direct File</a>. (No necesitas una copia de esta declaración; transferirás tu declaración aquí).</li>
             <li>Dirección de correo electrónico o número de teléfono.</li>
             <li>Licencia de conducir o identificación emitida por el estado.</li>
-            <li>Formulario 1099-G (si recibiste beneficios de desempleo en 2023).</li>
+            <li class="hide--nj">Formulario 1099-G (si recibiste beneficios de desempleo en 2023).</li>
```
embarnard, line 2050: I had some trouble formulating my thoughts arounds this. On one hand it does seem to solve the issue concisely and work well which I like but on the other hand its a departure from what we usually do in putting this kind of logic in the view and might obscure it a bit or get lost in transifex. [boulder] I think we definitely should add some sort of test or explore doing this in some other way like putting it into the view. What do you think?

embarnard (changes requested):
I'm not sure what the best approach on this is, happy to chat about it more but if you want to go this approach I think we should have a test

config/locales/es.yml
mmazanec22: Totally agree. I took this approach because I have been trying to use only the lightest touch possible in any files that aren't NJ specific. Is it possible instead to have default text and override if there's state specific text?

embarnard: Yes! Do mean adding a new jersey specific `help_text_html` key in the yml and then saving the original `help_text_html` as the default and displaying them in some sort of:
```
<% if current_state_code == 'nj' %>
<%= t("nj.help_text_html") %>
<% else %>
<%= t(".help_text_html") %>
<% end %>
```
I think that would work! Sucks to have to add the whole block of text again but does help with keeping the logic in one place

app/views/state_file/landing_page/edit.html.erb
```diff
@@ -65,7 +65,7 @@
               <% end %>
             <% else %>
               <div class="white-group spacing-below-60">
-                <%= t(".help_text_html") %>
+                <%= t(".#{@state_code}.help_text_html", default: ".help_text_html") %>
```
embarnard, line 68: [boulder]We still need to include the default text since this will now leave the other landing pages in a blank place for this section
Screenshot 2024-10-28 at 12 47 13 PM

embarnard (changes requested):
Add default text options for help text

app/views/state_file/landing_page/edit.html.erb
mmazanec22: @embarnard fixed the error, thanks for catching! Didn't realize default expects the actual text string

embarnard (approved):
Looks good! Thanks for making those changes 🙌🏼

PR #4897, opened 2024-10-22
FYST-792: ID Grocery Credit

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-792
## What was done?
- Added the Idaho grocery credit - two new pages, calculator code, and XML + PDF output
- We used the form & controller code from `NameDob` and `AzSeniorDependents` as examples
## How to test?
- We both tested manually and added automated unit tests for most scenarios
- Risk Assessment
- We added parameters to our "question-with-follow-up" javascript, so please also test some other places in the app where subsequent questions are revealed based on answers to make sure we didn't break them
image
image
image


arinchoi03 (commented):
Had a few questions! My head's spinning a bit tbh, might need to pair review tomorrow

app/views/state_file/questions/id_grocery_credit/edit.html.erb
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+      .grocery-credit-checkbox .form-group {
+          margin-bottom: 1.5rem;
+      }
+
+      ul.list--bulleted {
+          padding-left: 3rem !important;
+      }
+  </style>
```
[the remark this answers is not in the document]
mrotondo: I was just being a little lazy :) I moved them into our stylesheet, let me know if you think the names and locations are reasonable & safe.

app/views/state_file/questions/id_grocery_credit/edit.html.erb
```diff
@@ -0,0 +1,118 @@
...
+
+  <h1 class="h2"><%= t(".title_html") %></h1>
+
+  <% household_size = 1 %>
+  <% household_size += 1 if current_intake.filing_status_mfj? %>
+  <% household_size += current_intake.dependents.count %>
```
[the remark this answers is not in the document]
mrotondo: good point. turns out we already had the hard part there with `#filer_count`, so i added a `#household_count` method following on from that and used it here.

db/migrate/20241025002538_add_donate_grocery_credit_to_id_intake.rb
```diff
@@ -0,0 +1,5 @@
+class AddDonateGroceryCreditToIdIntake < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_id_intakes, :donate_grocery_credit, :integer, default: 0, null: false
+  end
+end
```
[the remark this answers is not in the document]
mrotondo: there's definitely not a reason to keep them, but i'm not going to merge them - i know it's a little sloppy, but we're already planning to compress our migrations into a schema snapshot so there will be no benefit to manually merging these ones first

app/lib/efile/id/id40_calculator.rb
```diff
@@ -49,6 +57,43 @@ def calculate_line_29
...
+        spouse_credit = spouse_eligible_months * spouse_credit_per_month
+
+        dependents_credit = @intake.dependents.sum do |dependent|
+          dependent_eligible_months = 12
+          if dependent.id_has_grocery_credit_ineligible_months_yes?
+            dependent_eligible_months -= dependent.id_months_ineligible_for_grocery_credit
+          end
+          dependent_credit_per_month = 10
+          dependent_eligible_months * dependent_credit_per_month
+        end
```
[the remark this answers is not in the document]
mrotondo: Agreed - I cleaned it up a bit, let me know what you think. More importantly, this comment helped me catch _two_ bugs with how the calculator was implemented, which were being masked by a too-fancy test setup, so I fixed those and simplified the tests.

app/lib/submission_builder/ty2024/states/id/documents/id40.rb
```diff
@@ -25,13 +25,17 @@ def document
                   xml.DependentGrid do
                     xml.DependentFirstName sanitize_for_xml(dependent.first_name, 20)
                     xml.DependentLastName sanitize_for_xml(dependent.last_name, 20)
-                    unless dependent.ssn.nil?
+                    unless dependent.ssn.nil?  # this will produce invalid xml if it's missing
```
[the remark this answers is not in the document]
mrotondo: This was a note to myself, I've deleted it. I was just noting that if we ever do hit this branch and omit the `DependentSSN` in the output XML, we'll get a validation error - but we shouldn't hit this branch since there should always be an ssn (TIN really) on the dependent

app/forms/state_file/id_grocery_credit_form.rb
```diff
@@ -0,0 +1,47 @@
...
+                greater_than_or_equal_to: 0,
+                message: :blank
+              },
+              if: -> { primary_has_grocery_credit_ineligible_months == "yes" }
+    validates :spouse_months_ineligible_for_grocery_credit,
+              numericality: {
+                greater_than_or_equal_to: 0,
+                message: :blank
+              },
+              if: -> { spouse_has_grocery_credit_ineligible_months == "yes" }
```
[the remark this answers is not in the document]
mrotondo: good call - i just hadn't looked to see what the norm was. i've added a form test now

app/views/hub/clients/_client_header.html.erb
```diff
@@ -138,3 +138,4 @@
       }
   </script>
 <% end %>
+
```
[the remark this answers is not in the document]
mrotondo: accidental change, reverted

app/javascript/lib/honeycrisp.js
```diff
@@ -103,16 +103,26 @@ var followUpQuestion = (function() {
                     }
                 });
 
+                let reset_other_follow_ups = $(this).attr('data-reset-other-follow-ups') !== "false";
```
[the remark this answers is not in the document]
mrotondo: let's chat about this in the morning. my intention was that these attributes should be optional, and that the default should be the previous behavior, so that reset_other_follow_ups should be true unless the `data-reset-other-follow-ups` attribute is both present and exactly the value "false", but i'm not certain that i got it right

app/assets/stylesheets/_state-file.scss
```diff
@@ -298,6 +314,11 @@
   .bulleted ul {
     padding-left: 3rem;
     list-style: disc;
+
+    li {
+      list-style: disc;
+      margin-top: 2rem;
+    }
```
arinchoi03, line 321: i thought the `li` items inherit from the `ul`, do we need to set the `list-style` again?

Also, is `.bulleted` a class that was not used for any elements previously but was defined? can we use `.list--bulleted` instead which seems more widely used?

spec/forms/state_file/id_grocery_credit_form_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+
+RSpec.describe StateFile::IdGroceryCreditForm do
+  describe "#valid?" do
+    context "without answering the top-level question" do
+      let!(:intake) { create :state_file_id_intake, :single_filer_with_json }
+      let(:invalid_params) do
+        {
+          household_has_grocery_credit_ineligible_months: "",
+          primary_has_grocery_credit_ineligible_months: "",
+          primary_months_ineligible_for_grocery_credit: "",
```
arinchoi03, line 11: [dust] do we need to include `primary_has_grocery_credit_ineligible_months` and `primary_months_ineligible_for_grocery_credit` in this test?

spec/forms/state_file/id_grocery_credit_form_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+              id_months_ineligible_for_grocery_credit: ""
+            },
+            '4': {
+              id: second_dependent.id,
+              id_months_ineligible_for_grocery_credit: ""
+            },
+            '5': {
+              id: third_dependent.id,
+              id_months_ineligible_for_grocery_credit: ""
+            }
```
arinchoi03, line 92: ooh super fascinating to see this is how the params come through for the dependents

spec/forms/state_file/id_grocery_credit_form_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+          }
+        }
+      end
+
+      it "is invalid and has errors for dependent months ineligible" do
+        form = described_class.new(intake, invalid_params)
+        expect(form).not_to be_valid
+        expect(form.dependents.first.errors).to include :id_months_ineligible_for_grocery_credit
```
arinchoi03, line 100: can we also add an expectation that the `form.dependents.second.errors` is empty as well as `form.dependents.third.errors` is empty?

spec/lib/efile/id/id40_calculator_spec.rb
```diff
@@ -58,6 +57,7 @@
 
   describe "Line 6d: Total Exemptions" do
     it "sums lines 6a, 6b, and 6c" do
+      allow(instance).to receive(:line_or_zero).and_return(nil)
```
arinchoi03, line 60: the purpose of this?

arinchoi03 (commented):
will chat tomorrow but looked over the changes

app/javascript/lib/honeycrisp.js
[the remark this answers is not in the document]
arinchoi03: oh I think i get it now! So most other use cases of followup questions never even have `data-reset-other-follow-ups` or `data-reset-hide-other-follow-ups` attributes, so they would always evaluate to true for `reset_other_follow_ups` & `hide_other_follow_ups` which retains previous behavior

For this particular feature, you wanted to _not_ hide/reset the followup questions on the page on a click of a question with a followup b/c you have multiple on a page that should remain open/shown and not reset. 

Took me a while to understand it but seems to make sense to me now!

app/lib/efile/id/id40_calculator.rb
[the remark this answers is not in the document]
arinchoi03: looks great to me -- glad that caught bugs, sorry i didn't catch it during review 😓

db/migrate/20241025002538_add_donate_grocery_credit_to_id_intake.rb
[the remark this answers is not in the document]
arinchoi03: i just had a hard time rolling back the migrations when i was switching branches, but just reset the db 🤷 not sure if having a single migration would've helped, but fine with leaving if we're going to compress the migrations soon.

app/lib/efile/id/id40_calculator.rb
```diff
@@ -13,13 +13,21 @@ def calculate
...
+        set_line(:ID40_LINE_43_DONATE, :calculate_line_43_donate)
+        set_line(:ID40_LINE_43, :calculate_line_43)
         @lines.transform_values(&:value)
       end
 
       def refund_or_owed_amount
         0
       end
 
+      def grocery_credit_amount
```
anisharamnani, line 27: is this just to make the method not private? what is the reason for the additional method?

anisharamnani: if so, i would say, wouldn't it be better to have the method be public instead so there is less obfuscation of logic?

app/assets/stylesheets/_state-file.scss
mrotondo: You're right - we didn't need to duplicate the list-style. Also, we did some investigating and learning and found that we could simplify things somewhat, using a combination of existing styles with one override.

spec/forms/state_file/id_grocery_credit_form_spec.rb
mrotondo: Technically no - the test still passes without them - but they'll always be part of the page and so I'd like to leave them here to be more reflective of real minimal params for this form.

spec/forms/state_file/id_grocery_credit_form_spec.rb
mrotondo: Sure! Done.

spec/lib/efile/id/id40_calculator_spec.rb
mrotondo: This is necessary since we now call `line_or_zero` in the calculator with params other than the ones specified here, and if you don't provide a default you get an error saying that one is required:
```
RSpec::Mocks::MockExpectationError: #<Efile::Id::Id40Calculator ...> received :line_or_zero with unexpected arguments
  expected: (:ID40_LINE_6C)
  got: (:ID40_LINE_43_WORKSHEET)
...
Please stub a default value first if message might be received with other args as well. 
```

If you know of another way to deal with this, let me know! But afaik, once you mock any return values for a method, no other calls to that method will ever get to the actual code and also must return a mocked value.

mrotondo: Nevermind, found a way to have it call the original code by default!

app/lib/efile/id/id40_calculator.rb
mrotondo: (discussed on tuple - this is so that we can show the amount in the title of the review page, and we don't allow calling the individual `calculate_x` methods because they frequently rely on the values of lines calculated above them)

changed this to assume that calculate has already been called, avoiding the possible bug described above

app/forms/state_file/id_grocery_credit_form.rb
```diff
@@ -0,0 +1,86 @@
...
+      # set household months to nil/0 if their "has months" answer is no
+      if base_attrs[:primary_has_grocery_credit_ineligible_months] == 'no'
+        base_attrs[:primary_months_ineligible_for_grocery_credit] = ''
+      end
+      if base_attrs[:spouse_has_grocery_credit_ineligible_months] == 'no'
+        base_attrs[:spouse_months_ineligible_for_grocery_credit] = ''
+      end
+
+      updated_dependent_data = updated_dependent_data.to_h do |k, v|
+        credit_count_key = v.key?(:id_months_ineligible_for_grocery_credit)
```
arinchoi03, line 67: can we call this `has_credit_count_key` to be more clear

app/forms/state_file/id_grocery_credit_form.rb
```diff
@@ -0,0 +1,86 @@
...
+        has_matching_no = updated_dependent_data.any? do |_, v2|
+          v[:id] == v2[:id] && v2[:id_has_grocery_credit_ineligible_months] == 'no'
+        end
```
arinchoi03, line 70: why is this going through the `updated_dependents_data` again to find the matching `v`? Isn't this the same as:

```
has_matching_no = v[:id_has_grocery_credit_ineligible_months] == 'no'
```
can we simplify this whole section to
```
      updated_dependent_data = updated_dependent_data.to_h do |k, v|
         if v.key?(:id_months_ineligible_for_grocery_credit) && v[:id_has_grocery_credit_ineligible_months] == 'no'
            [k, v.merge(id_months_ineligible_for_grocery_credit: '')]
         else
            [k, v]

      end
```

app/models/df_json_wrapper.rb
```diff
@@ -4,32 +4,62 @@ def initialize(json)
...
+  def self.json_accessor(mappings)
+    json_reader(mappings)
+    json_writer(mappings)
+  end
+
   def self.json_reader(mappings)
     mappings.each do |selector_name, selector_info|
-      selector_info => {type:, key:}
+      selector_info => { type:, key: }
+      key_path = key.split
```
arinchoi03, line 15: can you clarify why we need this `key.split`? which keys do we currently have that will have multiple words separated by a whitespace? is this a pre-emptive implementation (i see the tests)

app/models/df_json_wrapper.rb
```diff
@@ -4,32 +4,62 @@ def initialize(json)
...
   end
 
-  def df_json_value(key)
-    key_path = key.split
+  def self.json_writer(mappings)
+    mappings.each do |selector_name, selector_info|
+      selector_info => { type:, key: }
+      key_path = key.split
+      case type
+      when :date
```
arinchoi03, line 42: should we also handle setting a boolean value ("true") or a `money_amount` value?

app/views/state_file/questions/id_grocery_credit_review/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      %>
+
+      <div class="notice spacing-below-0">
+        <p><%= t('.donate_warning_p1') %>
+        <p><%= t('.donate_warning_p2') %>
+      </div>
+    </div>
+
+    <div class="reveal">
+      <p><a href="#" class="reveal__link"><%= t('.donate_impact_heading') %></a></p>
```
arinchoi03, line 28: change to `<button class="reveal__button">` after merging in main

app/views/state_file/questions/id_grocery_credit/edit.html.erb
```diff
@@ -0,0 +1,97 @@
...
+                <%= ff.cfa_select(:id_months_ineligible_for_grocery_credit, t(".total_number_months"), (1..12).to_a, include_blank: t('general.select_prompt')) %>
+              </div>
+            </div>
+          <% end %>
+        </div>
+      </div>
+    </div>
+
+    <div class="reveal">
+      <p><a href="#" class="reveal__link"><%= t('.why_are_you_asking_heading') %></a></p>
```
arinchoi03, line 79: change to `<button class="reveal__button">` after merging main

app/models/direct_file_json_data.rb
```diff
@@ -36,6 +36,10 @@ def initialize(json)
     @data = JSON.parse(json || "{}")
   end
 
+  def to_s
+    @data.to_s
+  end
```
arinchoi03, line 41: why was this added?

arinchoi03 (commented):
just a few more q's to clarify my understanding...this story got complex 😵

anisharamnani:
since @mrotondo is out today & we spent time working on this, i am going to give it a go addressing merge conflicts & handle some of the questions. 🤞

mrotondo:
Thanks for hopping in @anisharamnani! Just popping my head up to answer a few questions.

app/forms/state_file/id_grocery_credit_form.rb
```diff
@@ -0,0 +1,86 @@
...
+    def save
+      @intake.update!(attributes_to_save)
+    end
+
+    def attributes_to_save
+      base_attrs = attributes_for(:intake)
+      updated_dependent_data = dependents_attributes
+
+      # set household member "has months" answers to no if household "has months" answer is no
+      if base_attrs[:household_has_grocery_credit_ineligible_months] == 'no'
```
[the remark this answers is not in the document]
mrotondo: Enums let us track yes/no/unfilled instead of just true/false, which I think is primarily useful when generating output for optional pages, though we don't check the unfilled value of this answer anywhere IIRC. Someone who's been on the team longer than me can give a better answer here :) I don't feel very strongly about it either way, mostly just following convention.

app/forms/state_file/id_grocery_credit_form.rb
mrotondo: It's because the `dependent_params` hash has separate entries for the `id_has_grocery_credit_ineligible_months` and `id_months_ineligible_for_grocery_credit` params. I believe that this is because they're in separate `fields_for :dependents`, which generate their own param "sub-hashes", but which can be corresponded by the database IDs of the dependents (as we do here). Looking in the same place for both keys won't work unless we do a prior step of merging the params together so that each dependent has just one entry. I felt like the complexity was about the same either way, so Hugo and I went with this approach.

arinchoi03: Oh yeah! Forgot about that piece 🤪

app/models/df_json_wrapper.rb
mrotondo: yep, I was just trying to have some kind of feature parity with the XML version of this (in `DfXmlAccessor.define_xml_readers`/`DfXmlCrudMethods#df_xml_value`) where spaces can be used in a key and Nokogiri will look it up as a css path in the nested XML structure. I know we're not getting nested objects in the DF JSON currently, but I don't think this is too much speculative complexity to bear, personally.

app/models/df_json_wrapper.rb
mrotondo: I believe but am not sure that dates are the only value whose string representation doesn't just already look the way we want, so they were the only one that needed extra formatting. But I was pretty deep into this story by the time I did this and was just trying to finish up, so maybe the others do need to be there! We don't use this for anything but strings & dates yet, in any case.

app/models/direct_file_json_data.rb
mrotondo: Good catch. I originally added it because I incorrectly used it on `state_file_id_intakes.rb:122` instead of `to_json`, but now that I look more closely, I introduced a bug when I switched over to calling `to_json`, as it's now returning a json string of the entire `DirectFileJsonData` object (i.e. it has `{data: {...}}`) instead of just its stored json object. So this needs to be changed to a `to_json` that exposes `@data`, e.g.

```
def to_json
  @data.to_json
end
```

or
```
delegate :to_json, to: :data
```

anisharamnani:
hello @mrotondo! this is just a note for when you return, @mpidcock and i were able to make a merge in main and resolve the merge conflicts. 🎉 

we've found that this issue you mentioned [in slack](https://cfastaff.slack.com/archives/C0544ERAFQV/p1730394953591739) is still occurring where if you select yes, then select no, the calculations still appear on the following page. 

something may have changed with the merge conflicts to create this error? i am not sure, but we felt like we got as far as we could go with it.

app/forms/state_file/id_grocery_credit_form.rb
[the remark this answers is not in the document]
arinchoi03: @DrewProebstel for more context 

also @jenny-heath @embarnard @mpidcock @tahsinaislam if there was more reasoning for this choice (unfilled is similar to `nil`, no?)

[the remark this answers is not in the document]
mrotondo: FWIW I think that the decision to move away from using "unfilled" rather than nil as a default value in the app is well outside the scope of this PR, so discussion on that should probably happen elsewhere for posterity & visibility

app/models/direct_file_json_data.rb
arinchoi03: @mrotondo does this still need to be addressed?

app/forms/state_file/id_grocery_credit_form.rb
arinchoi03: @mrotondo bump

app/views/state_file/questions/id_grocery_credit_review/edit.html.erb
arinchoi03: @mrotondo bump

app/forms/state_file/id_grocery_credit_form.rb
mrotondo: done

app/views/state_file/questions/id_grocery_credit_review/edit.html.erb
mrotondo: done

app/views/state_file/questions/id_grocery_credit/edit.html.erb
mrotondo: done

app/models/direct_file_json_data.rb
mrotondo: done & added a test to verify

app/forms/state_file/id_grocery_credit_form.rb
```diff
@@ -0,0 +1,99 @@
+module StateFile
+  class IdGroceryCreditForm < QuestionsForm
+    attr_accessor :dependents_attributes
+    delegate :dependents, to: :intake
```
anisharamnani, line 4: @aloverso look here at lines 3 and 4!

app/views/state_file/questions/id_grocery_credit/edit.html.erb
```diff
@@ -0,0 +1,97 @@
...
+        <div class="question-with-follow-up">
+          <div class="question-with-follow-up__question">
+            <div class="white-group">
+              <p class="spacing-below-15"><strong><%= t(".select_household_members") %></strong></p>
+              <div class="tight-checkboxes spacing-above-0 spacing-below-0">
+                <%= f.cfa_checkbox(:primary_has_grocery_credit_ineligible_months, current_intake.primary.full_name, options: { checked_value: "yes", unchecked_value: "no", "data-follow-up": "#primary_months_ineligible_for_grocery_credit" }) %>
+                <% if current_intake.filing_status_mfj? %>
+                  <%= f.cfa_checkbox(:spouse_has_grocery_credit_ineligible_months, current_intake.spouse.full_name, options: { checked_value: "yes", unchecked_value: "no", "data-follow-up": "#spouse_months_ineligible_for_grocery_credit" }) %>
+                <% end %>
+                <%= f.fields_for :dependents do |ff| %>
```
anisharamnani, line 41: @aloverso please look at this line as well!

app/controllers/state_file/questions/id_grocery_credit_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      private
+
+      def form_params
+        params
+          .require(:state_file_id_grocery_credit_form)
+          .permit(
+            form_class.attribute_names +
+              [{ dependents_attributes: [:id, :id_has_grocery_credit_ineligible_months, :id_months_ineligible_for_grocery_credit] }])
```
anisharamnani, line 15: & here as well @aloverso

PR #4899, opened 2024-10-22
NJ 76 - sales and use tax question & calculation

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/76

## What was done?
- Add sales use tax question to NJ flow
- Calculate use tax automatically based on income when applicable

## How to test?
- `zeus_many_w2s` for $200,000 income => $170 use tax when calculated automatically
- `zeus_two_w2s` for $100,000 income => $106 use tax when calculated automatically
  
image
image
image
image
image
image


mluedke2:
@aloverso this is no longer draft, right?

app/forms/state_file/nj_sales_use_tax_form.rb
```diff
@@ -0,0 +1,37 @@
...
+      super(intake, params)
+    end
+
+    def save
+      attributes = attributes_for(:intake)
+
+      @intake.update!(attributes)
+    end
+  end
+end
```
mluedke2, line 37: sand - no line at end of file

aloverso:
@mluedke2 I don't know if this is necessary, but I've been using draft status until we send it to CFA for review

app/lib/efile/line_data.yml
```diff
@@ -590,6 +590,8 @@ NJ1040_LINE_42:
   label: '42 New Jersey Taxable Income (Subtract line 41 from line 39)'
 NJ1040_LINE_43:
   label: '43 Tax on amount on line 42 (Tax Table page 54)'
+NJ1040_LINE_51:
+  label: '51 Use Tax Due on Internet, Mail-Order, or Other Out-of-State Purchases (See instructions) If no Use Tax, enter 0.00 .'
```
mluedke2, line 594: dust - extra space in `0.00 .'`  , and it looks like most of these do not end in a `.` so maybe remove that to be consistent? I'm not totally sure what these are used for so not sure how much it matters

spec/controllers/state_file/questions/nj_sales_use_tax_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        {
+          state_file_nj_sales_use_tax_form: {
+            untaxed_out_of_state_purchases: "yes",
+            sales_use_tax_calculation_method: "automated"
+          }
+        }
+      end
+    end
+  end
+end
```
mluedke2, line 29: pebble/sand newline end-of-file

app/views/state_file/questions/nj_sales_use_tax/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    <div class="question-with-follow-up spacing-below-25">
+      <div class="question-with-follow-up__question">
+        <div class="white-group">
+          <%=
+            f.cfa_radio_set(
+              :untaxed_out_of_state_purchases,
+              label_text: t('.use_tax_radio_label'),
+              collection: [
+                { value: :yes, label: t("general.affirmative"), input_html: { "data-follow-up": "#sut-method-field" } },
+                { value: :no, label: t("general.negative") },
```
jachan, line 17: [DUST] How does the `sut-method-field` get shown or hidden? I see a way to hide the `#sut-field` onclick
```
input_html: { "onclick": "document.querySelector('#sut-field').style.display = 'none'"} },
```

but not for this first question. How is that visibility controlled?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -109,6 +110,29 @@ def calculate_tax_liability_without_deduction
...
+          84
+        when 75_000..100_000
+          106
+        when 100_000..150_000
+          134
+        when 150_000..200_000
+          170
+        when 200_000..Float::INFINITY
+          [0.000852 * nj_gross_income, 494].min.round
+        else
```
jachan, line 131: [DUST] Do we ever expect this to be 0? Looking at the tests, it seems like this should always have a positive value if we're autocalculating for them. Is there an alternative way to capture this error state?

jachan (commented):
Hi! Great work here. A few questions and minor comments.

app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -62,5 +62,17 @@
...
+      <% if current_intake.untaxed_out_of_state_purchases_yes? %>
+        <p class="text--bold spacing-below-5"><%=t(".amount") %></p>
+        <p><%=number_to_currency(current_intake.sales_use_tax, precision: 0)%></p>
+      <% end %>
+      <%= link_to t("general.edit"), StateFile::Questions::NjSalesUseTaxController.to_path_helper(return_to_review: "y"), class: "button--small" %>
+    </div>
+  </div>
+
   <%= render "state_file/questions/shared/review_footer" %>
 <% end %>
```
mluedke2, line 78: dust - outside the scope of this diff but noticing newline end-of-file here

app/views/state_file/questions/nj_sales_use_tax/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+      </div>
+    </div>
+
+    <% if params[:return_to_review].present? %>
+      <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
+    <% end %>
+    <%= f.continue %>
+  <% end %>
+<% end %>
```
mluedke2, line 51: pebble/sand - newline end-of-file

config/locales/en.yml
```diff
@@ -2647,13 +2647,31 @@ en:
...
+          followup_radio_label: Select one of the following
+          manual_use_tax_label: Enter calculated use tax. (Round to the nearest whole number)
+          subtitle_html: |
+            This includes:
+             <ul class="list--bulleted">
+               <li>Items purchased or services performed out-of-state, where there was no sales tax or the sales tax rate was below the NJ rate of 6.625%.</li>
+               <li>Online purchases where sales or use tax was not applied.</li>
+               <li>Purchases made by you or your spouse.</li>
+               <li><a href="https://nj.gov/treasury/taxation/su_10.shtml" target="_blank" rel="noopener noreferrer">Learn more</a>.
+             </ul>
```
mluedke2, line 2670: dust - is this the practice used throughout this repo, to include HTML in the localization strings directly? the alternative would be to interpolate just the text into HTML defined in `app/views/state_file/questions/nj_sales_use_tax/edit.html.erb` and then it would be the same HTML for `en`, `es`, etc. Probably not a problem in one case, but if it's the general practice it does sort of create a need to eventually have more E2E tests across all languages

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: maybe the first case could be `when -Float::INFINITY..15_000` instead of `when 0..15_000`

and then the `when 200_000..Float::INFINITY` could become the `else` (with a comment of `# above 200_000`) and then the current `else` case could just be removed?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -109,6 +110,29 @@ def calculate_tax_liability_without_deduction
...
+        when 50_000..75_000
+          84
+        when 75_000..100_000
+          106
+        when 100_000..150_000
+          134
+        when 150_000..200_000
+          170
+        when 200_000..Float::INFINITY
+          [0.000852 * nj_gross_income, 494].min.round
```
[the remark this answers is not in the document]
aloverso: I hear you. I'm leaning keeping it as-is. I don't think the variable name adds a lot. This file is full of magic numbers, that's kind of the nature of it. And the test does validate the magic number by explicitly writing the expectation of 0.0852%

spec/controllers/state_file/questions/nj_sales_use_tax_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  describe "#update" do
+    it_behaves_like :return_to_review_concern do
+      let(:form_params) do
+        {
+          state_file_nj_sales_use_tax_form: {
+            untaxed_out_of_state_purchases: "yes",
+            sales_use_tax_calculation_method: "automated"
+          }
+        }
+      end
```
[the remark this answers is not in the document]
aloverso: We cannot. The conditional logic is not the same as an if/else in the ruby code. It's controlled by the `question-with-follow-up` classes from Honeycrisp. Because honeycrisp is a JS library, it does not get executed during a rspec unit test. So the content in question will always be on the page.

Discussed this in slack with CFA a little while ago ([link](https://njcio.slack.com/archives/C07GVSC8X3P/p1727295165359129)) and we agreed that in order to test these kinds of conditional follow-ups, we need to use the end-to-end feature tests. The plan is for us to add NJ end-to-end tests once we have some of the specific test personas that the PMs are developing over the next couple weeks.

spec/forms/state_file/nj_sales_use_tax_form_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      context "with a non numeric sales-use-tax" do
+        let(:params) do
+          {
+            untaxed_out_of_state_purchases: "yes",
+            sales_use_tax_calculation_method: "manual",
+            sales_use_tax: "NaN",
+          }
+        end
+
+        it "is valid" do
```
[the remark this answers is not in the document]
aloverso: What's the typo?

app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -62,5 +62,17 @@
     </div>
   <% end %>
 
+  <div id="use-tax" class="white-group">
+    <div class="spacing-below-5">
+      <p class="text--bold spacing-below-5"><%=t(".use_tax_applied") %></p>
+      <p><%=current_intake.untaxed_out_of_state_purchases_yes? ? t("general.affirmative") : t("general.negative") %></p>
```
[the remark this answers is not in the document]
aloverso: Stole this from existing CFA code!

app/views/state_file/questions/nj_sales_use_tax/edit.html.erb
aloverso: It's controlled by the honeycrisp classes and attributes. `question-with-follow-up` and the `"data-follow-up": "#sut-method-field"` on the initial Yes radio button

aloverso: If you want to see how it works you can check out the `followUpQuestion` function in the `honeycrisp.js` file

app/lib/efile/nj/nj1040_calculator.rb
aloverso: The only case we're solving for here is if income is negative, right? Since that should never happen, I'd rather have the explicit `else` to just set this to 0. Even though it should never happen. But seems better than capturing it in the first case and assigning $14 of use tax

config/locales/en.yml
aloverso: What do you mean? Like, having three separate strings for each list item and doing the `<li>` and `<ul>` in the erb file?

Funny because just a few days ago I gave the feedback to @mmazanec22 to consolidate li strings into an HTML like this https://github.com/codeforamerica/vita-min/pull/4869#discussion_r1805051551

This pattern (with the html) is what I've seen in the code elsewhere. It seems easier to manage than trying to keep several items straight. But we could talk about this and refactor across our codebase if there's some reason we shouldn't use this pattern?

mluedke2: Yep my earlier comment is suggesting separate strings for each item so that there isn't HTML in the translation files. The reasons not to use this pattern would be to DRY the HTML, so refactors to it would only happen in one place. I think if it's the general practice we'd need to increase E2E tests across all languages since the code would be duplicated. Like for when we support 7 languages we'd have to make sure the HTML did not break for any of them. Thoughts?

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: throwing an error in the "should never happen" case instead of returning `0` then?

spec/forms/state_file/nj_sales_use_tax_form_spec.rb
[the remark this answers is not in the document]
jachan: This test case should be `it "is not valid"` right?

config/locales/en.yml
aloverso: Let's leave as-is for now and discuss as a team during our thursday sync. Depending on the outcome we can do a follow-up refactor

app/lib/efile/nj/nj1040_calculator.rb
aloverso: @mluedke2 idk... it just feels weird to throw an error in the app when I don't know enough about how it would get handled for the user experience. Something just feels weird about throwing an error. Maybe it doesn't matter since it should never happen, but it just makes me worried I'm missing something and then a user will end up with an error.

What do you think is problematic with the return 0 case?

mluedke2: i don't think it's blocking for this PR, but i think it's better to fail loudly and early-- if in the weird case we *are* missing something, i'd much rather give an error that will hopefully trigger our monitoring etc than quietly give the user a wrong tax return. the problem with returning `0` for a negative value is that it turns an invalid input into a valid output that can look normal and cause problems downstream.

app/helpers/vita_min_form_builder.rb
```diff
@@ -286,7 +286,7 @@ def vita_min_money_field(
     )
     text_field_options = standard_options.merge(
       class: (classes + ["text-input money-input"]).join(" "),
-      ).merge(options).merge(error_attributes(method: method)).merge(placeholder: '0.00')
+      ).merge(error_attributes(method: method)).merge(placeholder: '0.00').merge(options)
```
embarnard, line 289: What happens when with the change of order?

aloverso: Should have called this out! This allows me to override the placeholder text, which you can see in app/views/state_file/questions/nj_sales_use_tax/edit.html.erb:38

Because we're only allowing whole rounded numbers, and throwing an error if someone inputs a decimal (this is an ongoing discussion from a UX perspective, might change, but is our pattern for now) - it seemed important to change the placeholder text here so that the user isn't seeing `0.00` on a field that won't accept a decimal number.

config/locales/en.yml
embarnard: Yeah this is interesting. We do have this pattern in our codebase. I think I worry most about it being hard for the translators in Transifex to make changes to the text when its embedded in HTML but since historically it hasn't been an issue with them I would say its acceptable although I'm not sure preferred.

embarnard (approved):
Great work! No notes for changes ✅

app/lib/efile/nj/nj1040_calculator.rb
aloverso: @embarnard That's interesting! In what circumstances could that happen?

mluedke2: also interested how that might happen -- regardless, that leads me to believe that `when -Float::INFINITY..15_000` might be the route to go after all (and then we could remove the `else` since our cases would cover all possibilities)

config/locales/en.yml
mluedke2: at thursday sync we came for now to a pattern of putting HTML involving `class` or headers to be in the `.erb` for easier validation, but links and simple elements were acceptable in `.yml`

mluedke2 (dismissed):
Only two things that seem unresolved at the moment-- rest looks great to me!

* resolving the negative gross income input to `calculate_use_tax()` and what is invalid/valid for input and outputs there
* class or header html in .erb, only simple elements in .yml

app/lib/efile/nj/nj1040_calculator.rb
embarnard: Yes I think you're right @mluedke2, for some reason I thought if AGI was negative it should return 0 which the else case catches but found this comment from @mrotondo which leads be to believe it should instead use the `when -Float::INFINITY..15_000` case:
```
heads up/reminder for anyone writing calculator code, especially anyone creating one of our many tables that take AGI and match it against a bunch of ranges: AGI can be negative, so anytime we have a chart where the lowest band starts at 0, we need to either use [0, agi].max or replace the 0's in the range with -Float::INFINITY (the latter is what we've tended to do in the codebase)
```
I'm realizing I need to change this for NC sales use tax calculation as well

jachan: Brought this up today with Taxation, their guidance is to return 14$ for negative AGI.

app/forms/state_file/nj_sales_use_tax_form.rb
mluedke2: for my own understanding -- do we have a repo policy to go either way on newlines? github always flags but if we don't care i can ignore them

mluedke2 (approved):
awesome and thanks for using this one to help show me around the codebase!

PR #4902, opened 2024-10-23
Nj 100 101 json

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/100
https://github.com/newjersey/affordability-pm/issues/101

## What was done?
Build on the work CfA did to use directfile JSON, add fake JSON data, add tests to make sure primary and spouse name and dob are filled.

## How to test?
Go through flow, select married filing jointly


app/lib/navigation/state_file_nj_question_navigation.rb
```diff
@@ -28,7 +28,6 @@ class StateFileNjQuestionNavigation < Navigation::StateFileBaseQuestionNavigatio
                                           # Federal info does not show to users
                                           Navigation::NavigationStep.new(StateFile::Questions::FederalInfoController),
                                           Navigation::NavigationStep.new(StateFile::Questions::DataTransferOffboardingController, false),
-                                          Navigation::NavigationStep.new(StateFile::Questions::NameDobController),
```
mluedke2, line 31: 🎉

aloverso (approved):
Looks good! cool how it Just Works™️

spec/factories/state_file_nj_intakes.rb
```diff
@@ -138,16 +145,35 @@
...
+      raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_married_filing_jointly') }
+    end
+
+    trait :df_data_mfs do
+      filing_status { "married_filing_separately" }
+      raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_married_filing_separately') }
+      raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_married_filing_jointly') }
+    end
+
     trait :married_filing_jointly do
```
jachan, line 180: [DUST] Should we remove this trait? What happens if someone tries to use this trait now? Is it still a valid test case without the DF data?

spec/fixtures/state_file/fed_return_jsons/2023/nj/blind_primary.json
```diff
@@ -0,0 +1,12 @@
+{
```
jachan, line 1: [DUST] Doesn't have to be necessarily part of this PR, but I'm curious how we plan on generating these json files moving forward. Is part of the PR strategy to add to these test JSON files every time we introduce a new feature that requires reading a value from JSON? Is there a way to get these JSON files directly from DF instead of needing to construct them ourselves? As is, it feels like we're making these up based on what we think is necessary for our logic, which seems backwards.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -326,7 +326,8 @@
             create :efile_submission, tax_return: nil, data_source: create(
               :state_file_nj_intake,
               primary_first_name: "Grace",
-              primary_last_name: "Hopper"
+              primary_last_name: "Hopper",
+              primary_middle_initial: ""
```
jachan, line 330: [DUST] Why do we need to add `primary_middle_initial` here if it's an optional field?

app/assets/stylesheets/_state-file.scss
```diff
@@ -761,6 +761,10 @@ $state-colors: (
     .slab--footer {
       background-color: $color-nj-blue-darker;
     }
+
+    .review-header #household-info div:last-of-type {
```
jachan, line 765: [DUST] Can you elaborate on what this hides and why this is necessary?

jachan (commented):
Overall looks good! Left some questions for my own understanding.

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mmazanec22: To overwrite the base intake, which has a value for middle initial

app/assets/stylesheets/_state-file.scss
mmazanec22: This hides the edit button for name and dob, per the most recent comments on the ticket

spec/fixtures/state_file/fed_return_jsons/2023/nj/blind_primary.json
mmazanec22: Agree that manual construction is annoying. These represent data we are getting from df, but that just isn't in the federal return in a way we can predictably access or parse

spec/factories/state_file_nj_intakes.rb
mmazanec22: Maybe! I didn't want to rewrite literally every test so I left this for now. It is still valid for some cases, where we don't need the full df import with all the values aligned

spec/controllers/state_file/questions/nj_tenant_eligibility_controller_spec.rb
```diff
@@ -35,7 +35,7 @@
     end
 
     context "when a user is MFS" do
-      let(:intake) { create :state_file_nj_intake, :married_filing_separately }
+      let(:intake) { create :state_file_nj_intake, :df_data_mfs }
```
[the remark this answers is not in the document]
mmazanec22: The tests failed for some cases without the full df import. I'm a little fuzzy on why, but it's possible we can simplify the intake factory going forward

spec/fixtures/state_file/fed_return_jsons/2023/nj/blind_primary.json
jachan: Question for CFA: How do you deal with this? Do y'all just manually construct each time too?

jachan (approved):

spec/fixtures/state_file/fed_return_jsons/2023/nj/blind_primary.json
mmazanec22: It looks like they've got a similar setup for Maryland but I'd be curious to hear too!

app/assets/stylesheets/_state-file.scss
mpidcock: [dust] In the future, could you reach out when you are changing a shared component could you flag it for our attention? This change is acceptable since we will also be removing this button, but in the future I'd like to make sure we're on the same page.

In addition, I would prefer that the button be removed from the template entirely rather than being hidden so that it doesn't collect dust as a hidden element.

spec/controllers/state_file/questions/nj_tenant_eligibility_controller_spec.rb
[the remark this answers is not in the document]
mpidcock: [sand] Just to clarify, by "full DF import" do you mean df json data that matches the df xml data and is fully fleshed out to match the test case? The factory makes a default version of both json and xml, but that does mean it doesn't always match if MFS/MFJ is changed in the test setup. Some tests are ok with that, but some aren't if they need to use more spouse information. 

I think in the future we should aim for one baseline test trait for each filing status to avoid mismatched data between data sources.

spec/factories/state_file_nj_intakes.rb
mpidcock: See above, we believe this should be removed in follow-up work.

spec/fixtures/state_file/fed_return_jsons/2023/nj/blind_primary.json
mpidcock: Yes, right now we manually create the json test data with values that match the xml data when applicable. The expectation is that early next month we will have real test data once we are able to generate it from DF submissions.

mpidcock (approved):
Left some minor comments, looks good to me!

app/assets/stylesheets/_state-file.scss
mmazanec22: This only hides the button for nj

spec/controllers/state_file/questions/nj_tenant_eligibility_controller_spec.rb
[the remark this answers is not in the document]
mmazanec22: Made a ticket to dedupe traits

PR #4912, opened 2024-10-25
Nj 77 review a11y

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/77

## What was done?
Edits the review screen to mitigate accessibility issues:

- [WCAG 1.3.1 info and relationships](https://www.w3.org/TR/WCAG22/#info-and-relationships) - there were bold p tags used to convey structure to sighted users only, so these have been updated to headers. This also helps meet [2.4.10](https://www.w3.org/TR/WCAG22/#section-headings)
- [WCAG 2.4.4](https://www.w3.org/TR/WCAG22/#link-purpose-in-context) and [2.4.9](https://www.w3.org/WAI/WCAG22/quickref/#link-purpose-link-only) link purpose - there were edit links that all had the same text, but went to different places. Adding sr-only text makes it easier for someone to tab through the page or navigate using the rotor

This PR also replaces the deprecated axe gem with the current ones and adds an axe test just for the content in the main tag on the NJ review page. There were a few other a11y issues that were flagged by axe that should be fixed in follow up tickets.

Note: as mentioned in Slack, this modifies a shared file to use headers instead of p tags. To make sure the experience is consistent for all AT users, CfA could edit other review screen files. Alternately, NJ could alter the review screen to not use the shared header.

## How to test?
Navigate to the review screen. Turn on voiceover and navigate through the page ([VO cheat sheet](https://dequeuniversity.com/assets/pdf/screenreaders/voiceover-macos-guide.pdf)).
  
There should be no visual changes


spec/features/state_file/complete_intake_spec.rb
```diff
@@ -123,7 +125,7 @@
       click_on I18n.t("general.continue")
 
       expect(page).to have_text "Good news, you're getting a New York state tax refund of $1468. How would you like to receive your refund?"
-      expect(page).to_not have_text "Your responses are saved. If you need a break, you can come back and log in to your account at fileyourstatetaxes.org."
+      expect(page).not_to have_text "Your responses are saved. If you need a break, you can come back and log in to your account at fileyourstatetaxes.org."
```
mmazanec22, line 128: Linter autofix changes

config/locales/en.yml
```diff
@@ -2938,10 +2938,10 @@ en:
           nth_dependent_name: Your %{ordinal} dependent's name
           spouse_name: Your spouse's name
           taxes_owed_html: |
-            <p>
-              <strong>Your refund or taxes owed</strong> <br/>
-              $0
-            </p>
+            <h2 class="text--body text--bold spacing-below-5">
```
aloverso, line 2941: [pebble] per our discussion, we should move this h2 tag to the HTML and not the YML, right?

jachan: +1 since this line has class styling in the tag

jachan (approved):
Looks great. Glad we're making this change. Requesting that we address Anne's comment before merging.

PR #4914, opened 2024-10-25
FYST-788 Add interest income from obligations of the us

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-788

## Is PM acceptance required?

## What was done?
- Added ID 39R calculator 
- Added in ID 39R XML 
- added in field for Section B Line 3 for ID39R pdf
- added persona for idaho with 1099 INT

## How to test?
- units tests were added for calculator, xml and pdf
- PM can you use persona Tim 1099 INT to test 

Screenshot 2024-10-25 at 5 10 25 PM
Screenshot 2024-10-25 at 5 10 12 PM


app/lib/efile/id/id39_r_calculator.rb
```diff
@@ -0,0 +1,30 @@
...
+        @lines.transform_values(&:value)
+      end
+
+
+      private
+
+      def calculate_sec_b_line_3
+        sum = 0
+        @direct_file_json_data.interest_reports.each do |interest_report|
+          sum += interest_report.interest_on_government_bonds
```
arinchoi03, line 24: note to self: this will be `0` if the value was `nil` in the JSON, due to the DfJsonWrapper reader method, so we don't have to protect against a potential nil value 🙌

spec/fixtures/state_file/fed_return_jsons/2023/id/tim_1099_int.json
```diff
@@ -0,0 +1,26 @@
+{
+  "familyAndHousehold": [],
+  "filers": [
+    {
+      "firstName": "Tim",
+      "middleInitial": "T",
+      "lastName": "Interest",
+      "dateOfBirth": "1980-01-01",
+      "isPrimaryFiler": true
```
arinchoi03, line 9: can the filer also have a `tin` field as well since that's something we should always expect (hopefully) from IRS?

arinchoi03 (approved):
just one comment on the fixture but otherwise looks good!

spec/fixtures/state_file/fed_return_jsons/2023/id/tim_1099_int.json
tahsinaislam: thanks i'll add in the tin values!

PR #4924, opened 2024-10-29
Nj 104 taxable interest income

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/104

## What was done?
Add line 16a to PDF and XML

## How to test?
Use the Zeus two dependents profile, see 16a filled in pdf and xml

### DIFF
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -19,6 +19,7 @@ module Efile
         set_line(:NJ1040_LINE_8, :calculate_line_8)
         set_line(:NJ1040_LINE_13, :calculate_line_13)
         set_line(:NJ1040_LINE_15, :calculate_line_15)
+        set_line(:NJ1040_LINE_16A, :calculate_line_16a)
         set_line(:NJ1040_LINE_27, :calculate_line_27)
         set_line(:NJ1040_LINE_29, :calculate_line_29)
         set_line(:NJ1040_LINE_31, :calculate_line_31)
@@ -46,7 +47,6 @@ module Efile
       end
 
       def get_tax_rate_and_subtraction_amount(income)
-
         if @intake.filing_status_mfs? || @intake.filing_status_single?
           case income
           when 1..20_000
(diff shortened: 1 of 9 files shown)


app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -234,6 +223,25 @@ def is_ineligible_or_unsupported_for_property_tax
         StateFile::NjHomeownerEligibilityHelper.determine_eligibility(@intake) != StateFile::NjHomeownerEligibilityHelper::ADVANCE
       end
 
+      def calculate_line_40a
```
mmazanec22, line 226: Moved to be in sequential order by line number, did not change as part of this PR

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -234,6 +223,25 @@ def is_ineligible_or_unsupported_for_property_tax
...
+        when "rent"
+          if @intake.rent_paid.nil?
+            return nil
+          end
+          property_tax_paid = @intake.rent_paid * RENT_CONVERSION
+        else
+          return nil
+        end
+
+        is_mfs_same_home ? (property_tax_paid / 2.0).round : property_tax_paid.round
```
powersurge360, line 242: You may consider `2.to_d` since this is money. I'm not sure that it would matter with a single calculation that's being rounded anyways, but converting to decimal is more correct because the decimal type uses (to my understanding) two integers to represent the decimal in a fraction notation until the number is ultimately intended to be used when it finally does the math that converts back to a float for display.

In this specific case it probably doesn't matter but it might be good to begin using decimals by default with taxes?

powersurge360 (commented):
Looks good to me. Left a comment wrt to decimal v. floats. I don't think it actually matters in this specific case but I'd like an explicit decision to either punt on it or to make the change and then I'll approve.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -391,6 +391,29 @@ def over_65_birth_year
...
+    context 'with interest reports, but no interest on government bonds' do
+      let(:intake) { create(:state_file_nj_intake, :df_data_one_dep) }
+      it 'does not set line 16a' do
+        expect(instance.lines[:NJ1040_LINE_16A].value).to eq(nil)
+      end
+    end 
+
+    context 'with interest on government bonds' do
+      let(:intake) { create(:state_file_nj_intake, :df_data_two_deps) }
+      it 'sets line 16a to 300 (fed taxable income minus sum of bond interest)' do
```
aloverso, line 411: [dust/pebble] could we add numbers to these calculation descriptions? Helps me track down where they're coming from in the test data. 

ie, `"sets line 16a to $300 (fed taxable income $XXX minus sum of bond interest $XXX)"`

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -689,6 +689,48 @@
...
+          ["112",
+           "111",
+           "110",
+           "109",
+           "108",
+           "Text107",
+           "undefined_41",
+           "undefined_40",
+           "undefined_39",
+           "undefined_43"].each { |pdf_field| expect(pdf_fields[pdf_field]).to eq "" }
```
aloverso, line 709: Love this

aloverso (approved):
Looks good!

app/lib/efile/nj/nj1040_calculator.rb
mmazanec22: This was not part of this PR, I just moved the method to put them in sequential order

powersurge360 (approved):

PR #4925, opened 2024-10-29
NJ 140 - fix XML errors and add schema validation to all XML tests

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/140

## What was done?
- Fix error introduced by W2 changes (need for `to_i`)
- Add tests that would have caught this error to `nj_return_xml_spec`, automatically sync W2s in NJ intake
- Add tests that will catch future issues in XML including client questions by adding XML validation to all nj1040 specs

## How to test?
n/a

Before:
```
113:0: ERROR: Element '{http://www.irs.gov/efile}WagesSalariesTips': '50000.0' is not a valid value of the atomic type '{http://www.irs.gov/efile}NJAmountNNType'.
115:0: ERROR: Element '{http://www.irs.gov/efile}TotalIncome': '50000.0' is not a valid value of the atomic type '{http://www.irs.gov/efile}NJAmountNNType'.
116:0: ERROR: Element '{http://www.irs.gov/efile}GrossIncome': '50000.0' is not a valid value of the atomic type '{http://www.irs.gov/efile}NJAmountNNType'.
120:0: ERROR: Element '{http://www.irs.gov/efile}TaxableIncome': '45000.0' is not a valid value of the atomic type '{http://www.irs.gov/efile}NJAmountNNType'.
124:0: ERROR: Element '{http://www.irs.gov/efile}NewJerseyTaxableIncome': '45000.0' is not a valid value of the atomic type '{http://www.irs.gov/efile}NJAmountNNType'.
```

### DIFF
--- a/spec/factories/state_file_nj_intakes.rb
+++ b/spec/factories/state_file_nj_intakes.rb
@@ -110,6 +110,7 @@ FactoryBot.define do
     raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_zeus_one_dep') }
     
     after(:build) do |intake, evaluator|
+      intake.municipality_code = "0101"
       numeric_status = {
         single: 1,
         married_filing_jointly: 2,
@@ -133,14 +134,11 @@ FactoryBot.define do
       end
       intake.update(overrides)
 
+      intake.synchronize_df_w2s_to_database
       intake.synchronize_df_dependents_to_database
       intake.dependents.each_with_index do |dependent, i|
         dependent.update(dob: i.years.ago)
       end
-    end
-
-    trait :with_w2s_synced do
-      after(:create, &:synchronize_df_w2s_to_database)
     end
 
     trait :df_data_2_w2s do
--- a/spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
+++ b/spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
@@ -2,14 +2,22 @@ require 'rails_helper'
 
 describe SubmissionBuilder::Ty2024::States::Nj::Documents::Nj1040, required_schema: "nj" do
   describe ".document" do
-    let(:intake) { create(:state_file_nj_intake, filing_status: "single", municipality_code: "0101") }
+    let(:intake) { create(:state_file_nj_intake, filing_status: "single") }
     let(:submission) { create(:efile_submission, data_source: intake) }
-    let(:build_response) { described_class.build(submission, validate: false) }
+    let(:build_response) { described_class.build(submission, validate: true) }
     let(:xml) { Nokogiri::XML::Document.parse(build_response.document.to_xml) }
 
-    it "includes municipality code with a prepending 0" do
-      xml = described_class.build(submission).document
-      expect(xml.document.at("CountyCode").to_s).to include("00101")
+    after(:each) do
+      expect(build_response.errors).not_to be_present
+    end
+
+    context "with municipality code" do
+      let(:intake) { create(:state_file_nj_intake, municipality_code: "0304") }
+
+      it "includes municipality code with a prepending 0" do
+        xml = described_class.build(submission).document
+        expect(xml.document.at("CountyCode").to_s).to include("00304")
+      end
     end
 
     context "when filer has no spouse" do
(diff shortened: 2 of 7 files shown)


spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -2,14 +2,22 @@
...
     let(:submission) { create(:efile_submission, data_source: intake) }
-    let(:build_response) { described_class.build(submission, validate: false) }
+    let(:build_response) { described_class.build(submission, validate: true) }
     let(:xml) { Nokogiri::XML::Document.parse(build_response.document.to_xml) }
 
-    it "includes municipality code with a prepending 0" do
-      xml = described_class.build(submission).document
-      expect(xml.document.at("CountyCode").to_s).to include("00101")
+    after(:each) do
+      expect(build_response.errors).not_to be_present
```
mmazanec22, line 11: Nice, clean and simple!

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -61,22 +65,28 @@
...
           builder_response = described_class.build(submission)
           expect(builder_response.errors).not_to be_present
+          expect(builder_response.document.at("WagesSalariesTips").text).not_to eq(nil)
+          expect(builder_response.document.at("NewJerseyTaxableIncome").text).not_to eq(nil)
         end
       end
 
       context "with many w2s" do
-        let(:intake) { create(:state_file_nj_intake, municipality_code: "0101", raw_direct_file_data: StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_zeus_many_w2s')) }
+        let(:intake) { create(:state_file_nj_intake, raw_direct_file_data: StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_zeus_many_w2s')) }
```
mmazanec22, line 74: [pebble] Can we remove the raw_direct_file_data here and just use :df_data_many_w2s?

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -61,22 +65,28 @@
...
           builder_response = described_class.build(submission)
           expect(builder_response.errors).not_to be_present
+          expect(builder_response.document.at("WagesSalariesTips").text).not_to eq(nil)
+          expect(builder_response.document.at("NewJerseyTaxableIncome").text).not_to eq(nil)
         end
       end
 
       context "with two w2s" do
-        let(:intake) { create(:state_file_nj_intake, municipality_code: "0101", raw_direct_file_data: StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_zeus_two_w2s')) }
+        let(:intake) { create(:state_file_nj_intake, raw_direct_file_data: StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_zeus_two_w2s')) }
```
mmazanec22, line 84: Same question here

mmazanec22 (approved):

spec/factories/state_file_nj_intakes.rb
```diff
@@ -133,16 +134,13 @@
       end
       intake.update(overrides)
 
+      intake.synchronize_df_w2s_to_database
```
aloverso, line 137: This line seems to be the one causing issues with the PDF spec and the state_return spec. I will probably need some help from CFA debugging why that is

jachan (approved):
Looks great. Appreciate you adding muni code by default, definitely simplifies future testing.

spec/factories/state_file_nj_intakes.rb
aloverso: @jachan this comment addresses your question about circleci failures

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -26,10 +26,12 @@
       end
 
       context "with one dep" do
-        let(:intake) { create(:state_file_nj_intake, :df_data_one_dep, municipality_code: "0101") }
+        let(:intake) { create(:state_file_nj_intake, :df_data_one_dep) }
         it "does not error" do
           builder_response = described_class.build(submission)
           expect(builder_response.errors).not_to be_present
+          expect(builder_response.document.at("WagesSalariesTips").text).not_to eq(nil)
+          expect(builder_response.document.at("NewJerseyTaxableIncome").text).not_to eq(nil)
```
jenny-heath, line 34: question: the description of this says "does not error" but i can't tell why it cares about these two fields specifically (and just that they are not nil?)

aloverso: Ah, good catch. We can probably delete this. We wanted some way to validate that it was building "as expected" - ie, that if there were W2 wages in the original DF XML, there should be output elements in the state return that reflect those W2 wages. Another test to ensure that they were actually getting synced and used.

I think we're now testing this sufficiently in other places though

jenny-heath:
@aloverso looks like the issue was that one of the W2s in the arizona fixture for `:df_data_2_w2s` lacked a `LocalityNm` (and actually just doesn't have contents in `W2LocalTaxGrp`). according to our validations on the model it's valid to be missing those things, so i just added in an if statement. i'd like to flesh out the test a little more but i think what i'll do as a follow-up to this is check with our product people to get a more comprehensive idea of what we care about re: building the w2s, now that we have them all in the db. then can refactor it out into a separate spec for `return_w2.rb` (which i've been meaning to do anyway).

PR #4926, opened 2024-10-29
Update AZ140 test coverage

jenny-heath:
i noticed while working on a different thing that our test coverage for this file could be better so i used the small remainder of my friday to add to this file

the commits do make sense individually so you can review them that way if it would help to see where things were added vs simply re-ordered


mpidcock (approved):
Thanks for doing this! Much better coverage and really clear setup.

PR #4941, opened 2024-11-01
Nj 97 income eligibility

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/97

## Is PM acceptance required?

## What was done?
- I moved logic that was going to be reused (wage calculation, age check, property tax eligibility) into `lib/efile/`

- I edited both `next_path()` and `show?()` for several of the property tax-related screens to skip questions not needed due to taxpayer not being eligible for what they were asking. (I learned that `next_path()` sort of overrides `show?()`, and had to use both to properly build the flow)

- I considered adding a line to the calculator (and a bunch of related tests) to make super explicit the requirement `40a, and 41 should be blank`. Those lines are blank if `property_tax_paid` or `rent_paid` are `nil`, which is always true when the screens this PR skips are skipped. So in the end it felt over-constrained and would not happen in practice so I removed the extra code. You can see what I removed in [this commit](https://github.com/codeforamerica/vita-min/commit/192b455313b4da2a69a37289aba44aed7c306c1a) if interested.

## How to test?
- The current flow stays the same for everyone with income above the minimum, so that can be still experienced with `Zeus many w2s` (video below)
- The `Minimal` transfer has an income below the minimum, so using that you can see that the user is still asked about being homeowner/tenant but then is moved along (video below)
- The `Blind primary` transfer has income below minimum but qualifies for credit-only, so you can see user is asked about homeowner/tenant and the housing details eligibility, but skips the screen asking for property tax/rent paid (video below)
- The new `Married filing jointly 15k wages` transfer has an income that is below the $20k minimum for MFJ
- The new `Married filing separately 15k wages` transfer has an income that is above the $10k minimum for MFS

### Zeus many w2s

https://github.com/user-attachments/assets/bda9ab82-ccfe-43bc-80c5-b1143b387f70

---
### Minimal

https://github.com/user-attachments/assets/266c1a76-0060-43e4-923c-3254f29562d3

---
### Blind primary

https://github.com/user-attachments/assets/ab4a0f14-8c47-479e-807c-17eac2b60c5e

---
### MFJ $15k Wages (below $20k minimum)

https://github.com/user-attachments/assets/1291d313-bd8b-4557-8c64-e205027b1de1

---
### MFS $15k Wages (above $10k minimum)

https://github.com/user-attachments/assets/14a1275b-9d65-41e7-87a9-44f03d96c4f0


app/lib/efile/nj/nj_property_tax_eligibility.rb
```diff
@@ -0,0 +1,38 @@
...
+        def determine_eligibility(intake)
+          state_wages = Efile::Nj::NjStateWages.calculate_state_wages(intake)
+
+          wage_minimum = if intake.filing_status == :married_filing_separately || intake.filing_status == :single
+            10_000
+          else
+            20_000
+          end
+
+          return NOT_INELIGIBLE if state_wages > wage_minimum
```
mmazanec22, line 18: [dust] The double negative is throwing me. I wonder if an alternative like "possibly eligible," or "eligible for credit and deduction" or something else more specific might make this easier to reason through

app/lib/efile/nj/nj_property_tax_eligibility.rb
```diff
@@ -0,0 +1,38 @@
...
+                            Efile::Nj::NjSenior.is_over_65(intake.primary_birth_date)
+
+          spouse_meets_exception = intake.filing_status == :married_filing_jointly &&
+                              (
+                                intake.direct_file_data.is_spouse_blind? ||
+                                intake.spouse_disabled_yes? ||
+                                Efile::Nj::NjSenior.is_over_65(intake.spouse_birth_date)
+                              )
+
+          return INELIGIBLE_FOR_DEDUCTION if meets_exception || spouse_meets_exception
```
mmazanec22, line 31: [pebble] Same comment about clarity of language - maybe "eligible for credit only"? To clarify the distinction between this and just plain "ineligible"?

app/lib/efile/nj/nj_state_wages.rb
```diff
@@ -0,0 +1,15 @@
+module Efile
+  module Nj
+    class NjStateWages
+      NO_WAGES = -1
+
+      def self.calculate_state_wages(intake)
```
mmazanec22, line 6: [dust] Is this not covered in the calculator class already?

spec/lib/efile/nj/nj_property_tax_eligibility_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      { filing_status: :single, income: 9_999, expected: :ineligible },
+      { filing_status: :married_filing_jointly, income: 19_999, expected: :ineligible },
+      { filing_status: :married_filing_separately, income: 9_999, expected: :ineligible },
+      { filing_status: :head_of_household, income: 19_999, expected: :ineligible },
+      { filing_status: :qualifying_widow, income: 19_999, expected: :ineligible },
+    ].each do |test_case|
+      context "when filing #{test_case}" do
+        let(:intake) do
+          traits = []
+          traits << test_case[:filing_status] if test_case[:filing_status] != :single
```
mmazanec22, line 37: [pebble] Making the shape of the objects above something like { traits: { traitx: value }, expected: blah } would allow you to ditch lines 36-44

app/lib/efile/nj/nj_senior.rb
```diff
@@ -0,0 +1,11 @@
+module Efile
+  module Nj
+    class NjSenior
+      def self.is_over_65(birth_date)
```
mmazanec22, line 4: [pebble] The base intake model already has methods `primary_senior?` and `spouse_senior?`

app/lib/efile/nj/nj_state_wages.rb
mluedke2: i moved the wages logic (and the age check logic) out of the calculator since it now needs to be used in two separate places-- let me know if that choice sounds ok

app/lib/efile/nj/nj_senior.rb
mluedke2: i moved the age check logic (and wages logic) out of the calculator since it now needs to be used in two separate places-- let me know if that choice sounds ok

app/lib/efile/nj/nj_property_tax_eligibility.rb
mluedke2: yeah i considered "possibly eligible" -- since actual eligibility depends on additional answers to questions we can't just simply say "eligible" for anything yet. agree that double-negative is not great. i think i'll change to "possibly eligible"

app/lib/efile/nj/nj_property_tax_eligibility.rb
mluedke2: yeah makes sense. i'll change this one to "possibly eligible for credit"

spec/lib/efile/nj/nj_property_tax_eligibility_spec.rb
mluedke2: good idea-- will do that

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -6,20 +6,25 @@ class NjHomeownerEligibilityController < QuestionsController
       before_action -> { @filing_year = Rails.configuration.statefile_current_tax_year }
 
       def self.show?(intake)
-        intake.household_rent_own_own?
+        intake.household_rent_own_own? &&
+        Efile::Nj::NjPropertyTaxEligibility.determine_eligibility(intake) != Efile::Nj::NjPropertyTaxEligibility::INELIGIBLE
       end
 
       def next_path
```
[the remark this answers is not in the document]
mluedke2: ok will add little comments on the `show?`s

app/lib/efile/nj/nj_senior.rb
mluedke2: oh i think i may have misread your comment -- i simply was just taking our existing logic that explicitly compares birth year... so moving to a `primary_senior?` would be a refactor to our existing code. i'm ok with that but just noting that it would effect beyond this PR, so I'd be relying on existing tests to make sure that refactor is ok. I'll try it though, since it would be nice to just not have this util at all...

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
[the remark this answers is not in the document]
mmazanec22: I think next_path was the one that was bending my mind a little more. But also maybe the variable renaming will help and make this unnecessary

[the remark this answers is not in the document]
jachan: [DUST ]+1 the show makes sense to me as-is

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -6,20 +6,25 @@ class NjHomeownerEligibilityController < QuestionsController
...
+        
         case StateFile::NjHomeownerEligibilityHelper.determine_eligibility(current_intake)
         when StateFile::NjHomeownerEligibilityHelper::INELIGIBLE
           NjIneligiblePropertyTaxController.to_path_helper(options)
         when StateFile::NjHomeownerEligibilityHelper::UNSUPPORTED
           NjUnsupportedPropertyTaxController.to_path_helper(options)
         else
-          NjHomeownerPropertyTaxController.to_path_helper(options)
+          if Efile::Nj::NjPropertyTaxEligibility.determine_eligibility(current_intake) == Efile::Nj::NjPropertyTaxEligibility::INELIGIBLE_FOR_DEDUCTION
+            super
```
jachan, line 24: [DUST] If adding a comment, this is the confusing part. When the applicant is INELIGIBLE_FOR_DEDUCTION, it calls next_path()? If adding comment, perhaps clarify how this is meant to skip certain eligibility questions.

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -147,6 +147,14 @@ def calculate_line_6
         number_of_line_6_exemptions * 1_000
       end
 
+      def line_7_self_checkbox
+        Efile::Nj::NjSenior.is_over_65(@intake.primary_birth_date)
```
jachan, line 151: [DUST] Cool to see this logic broken out!

app/lib/efile/nj/nj_property_tax_eligibility.rb
```diff
@@ -0,0 +1,38 @@
...
+    module NjPropertyTaxEligibility
+      INELIGIBLE = :ineligible
+      INELIGIBLE_FOR_DEDUCTION = :ineligible_for_deduction
+      NOT_INELIGIBLE = :not_ineligible
+
+      class << self
+        def determine_eligibility(intake)
+          state_wages = Efile::Nj::NjStateWages.calculate_state_wages(intake)
+
+          wage_minimum = if intake.filing_status == :married_filing_separately || intake.filing_status == :single
```
jachan, line 12: [SAND] Prefer `intake.filing_status_mfs?` and `filing_status_single?` helper methods rather than checking directly.

app/lib/efile/nj/nj_property_tax_eligibility.rb
jachan: [DUST] Agree with Melanie here, possibly eligible feels better to me.

app/lib/efile/nj/nj_state_wages.rb
jachan: Moving shared logic out into its own class seems fine to me to avoid rewriting logic in two different places.

spec/controllers/state_file/questions/nj_tenant_eligibility_controller_spec.rb
```diff
@@ -81,29 +81,125 @@
...
+        let(:intake) {
+          create(
+            :state_file_nj_intake,
+            :df_data_minimal,
+            :primary_disabled,
+            household_rent_own: "rent",
+            tenant_home_subject_to_property_taxes: "no"
+          )
+        }
+        it "next path is ineligible page" do
```
jachan, line 136: [DUST] Great tests for next path, these are very legible.

jachan (approved):
Agree with Melanie's comments on naming to avoid NOT_INELIGIBLE, and then have some minor comments on code style. Great work with your first (I think?) PR!

app/lib/efile/nj/nj_property_tax_eligibility.rb
mluedke2: good catch - done

app/lib/efile/nj/nj_senior.rb
mluedke2: great catch - using those methods does indeed still pass tests, so i will refactor the original in the calculator to just use that method too

spec/lib/efile/nj/nj_property_tax_eligibility_spec.rb
mluedke2: was able to do something along those lines to reduce test lines. thanks for the suggestion!

app/lib/efile/nj/nj_property_tax_eligibility.rb
mluedke2: changed to 

```
INELIGIBLE = :ineligible
POSSIBLY_ELIGIBLE_FOR_CREDIT = :possibly_eligible_for_credit
POSSIBLY_ELIGIBLE_FOR_DEDUCTION_OR_CREDIT = :possibly_eligible_for_deduction_or_credit
```

app/lib/efile/nj/nj_property_tax_eligibility.rb
mluedke2: changed to

```
INELIGIBLE = :ineligible
POSSIBLY_ELIGIBLE_FOR_CREDIT = :possibly_eligible_for_credit
POSSIBLY_ELIGIBLE_FOR_DEDUCTION_OR_CREDIT = :possibly_eligible_for_deduction_or_credit
```

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
mluedke2: This PR add three uses of `super` within `next_path`, so I just added comments to those three lines to make more clear what the behavior is

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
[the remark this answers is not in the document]
mluedke2: added three comments to the uses of `super` within `next_path` to make behaviors there more clear

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: ended up using `@intake.primary_senior?` and `@intake.spouse_senior?` after all

app/lib/efile/nj/nj_state_wages.rb
mluedke2: I was able to get rid of the age check util altogether but kept this wages one

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -6,20 +6,25 @@ class NjHomeownerEligibilityController < QuestionsController
...
-
+        
         case StateFile::NjHomeownerEligibilityHelper.determine_eligibility(current_intake)
         when StateFile::NjHomeownerEligibilityHelper::INELIGIBLE
           NjIneligiblePropertyTaxController.to_path_helper(options)
         when StateFile::NjHomeownerEligibilityHelper::UNSUPPORTED
           NjUnsupportedPropertyTaxController.to_path_helper(options)
         else
-          NjHomeownerPropertyTaxController.to_path_helper(options)
+          if Efile::Nj::NjPropertyTaxEligibility.determine_eligibility(current_intake) == Efile::Nj::NjPropertyTaxEligibility::POSSIBLY_ELIGIBLE_FOR_CREDIT
```
arinchoi03, line 23: would it be cleaner/easier to read if there a way to just call `current_intake.possibly_eligible_for_credit?` instead of doing comparisons like this everywhere? i.e. define these methods on NJ intake?

arinchoi03 (approved):
[Suggestion] I'm a bit newer to the team so I'm still figuring this out but wondering if all the specific housing-related controllers need to be in the `StateFileNjQuestionNavigation`? This may help with all the logic in the `show` method -- if that controller is not in the navigation, only time `show?` will be called if you explicitly navigate to that controller.

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -6,20 +6,25 @@ class NjHomeownerEligibilityController < QuestionsController
...
-        intake.household_rent_own_own?
+        intake.household_rent_own_own? &&
+        Efile::Nj::NjPropertyTaxEligibility.determine_eligibility(intake) != Efile::Nj::NjPropertyTaxEligibility::INELIGIBLE
       end
 
       def next_path
         options = {}
         options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
-
+
```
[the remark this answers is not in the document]
mluedke2: removed

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
mluedke2: Ooh thanks for this suggestion. I don't feel too strongly about having the logic live in the util vs intake model, especially since this is my first PR so there may be a pattern I'm not aware of, but I lean util to keep it separate from everything else.

That being said I'll make the functions like `.possibly_eligible_for_credit?(intake)` in the util, so it can be

```
Efile::Nj::NjPropertyTaxEligibility.possibly_eligible_for_deduction_or_credit?(intake)
```

instead of

```
Efile::Nj::NjPropertyTaxEligibility.determine_eligibility(intake) == Efile::Nj::NjPropertyTaxEligibility::POSSIBLY_ELIGIBLE_FOR_DEDUCTION_OR_CREDIT
```

mluedke2:
@arinchoi03 that is helpful to know that the questions don't all need to be in the Navigation-- our process thus far has been to add all of them to it. I think since this logic is not too tedious I'll keep it, but I'll remember that for next time

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
arinchoi03: sounds fine to use the util

mluedke2: added these in [the latest commit](https://github.com/codeforamerica/vita-min/pull/4941/commits/65719ab69a0cfe2c459421dab6ad90f2dd9f96e3)

arinchoi03: Thanks for humoring me!

PR #4944, opened 2024-11-01
[FYST-1089] refactor twilio service

jnf:
## Link to pivotal/JIRA issue
[FYST-1089](https://codeforamerica.atlassian.net/browse/FYST-1089)

### Updates credentials file structure
The structure below would appear in each environment's credentials file. ~This structure reflects a simplified Twilio organization that isn't reflective of the current setup;~ restructuring the Twilio services and setup is a blocking dependency of this refactor (completed 11 Nov 24).

```
twilio:
  voice_phone_number: <i think there's just the one>
  gyr:
    account_sid: <value>
    auth_token: <value>
    messaging_service_id: <value>
  statefile:
    account_sid: <value>
    auth_token: <value>
    messaging_service_id: <value>
  ctc:
    account_sid: <value>
    auth_token: <value>
    messaging_service_id: <value>
```

### Updates `TwilioService` API
Enables vita-min to interact with many Twilio messaging services. The responsibility of determining which is positioned in the MultiTenantService. Now, to interact with Twilio using their REST api, you'll need to create an instance of `TwilioService` to represent a specific vita-min tenant, i.e., `TwilioService.new(:statefile)`. The returned instance knows how to do all the things the prior singleton version did, just scoped to the appropriate messaging service (one per tenant).

### Updates all the callsites and their specs
The bulk of the PR is updating all the places we interact with Twilio to use this updated interface. A bunch of related specs needed tweaking too.

## How to test?
Assuming all specs pass, a full test involves getting this into an environment with a valid Twilio setup (staging, demo, or production). Once there, the regular Twilio interactions (auth codes, document transfer, chats) should work as they always have. _New_ Twilio functionality (like FYST support) will come soon™️, in a separate PR/testing effort. 

[FYST-1089]: https://codeforamerica.atlassian.net/browse/FYST-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/twilio_service.rb
```diff
@@ -12,104 +13,101 @@ class TwilioService
     delivered
     undelivered
     failed
-  ].freeze
+  ).unshift(nil) # why do we need nil in this list, and why must it be first?
```
jenny-heath, line 16: [DUST] i assume you're planning to remove this comment before merging? would prefer not to leave stuff like this around the codebase

spec/services/twilio_service_spec.rb
```diff
@@ -62,18 +63,19 @@
...
 
       before do
-        allow(TwilioService).to receive(:fetch_attachment).with(image_url).and_return({
+        @twilio_service = TwilioService.new(:gyr)
+        allow(@twilio_service).to receive(:fetch_attachment).with(image_url).and_return({
                                                                                         filename: "a_real_image.jpg",
                                                                                         body: "image body"
                                                                                       })
-        allow(TwilioService).to receive(:fetch_attachment).with(pdf_url).and_return({
+        allow(@twilio_service).to receive(:fetch_attachment).with(pdf_url).and_return({
```
jenny-heath, line 95: i'm unfamiliar with using instance variables in rspec specs—typically we'd put something like this in a `let` block (not to say it's better, just that it's the practice). happy to discuss!

app/models/signup.rb
```diff
@@ -34,7 +34,7 @@ def self.send_message(message_name, batch_size=nil, after: nil)
       end
 
       if signup.phone_number.present?
-        TwilioService.send_text_message(to: signup.phone_number, body: message.sms_body)
+        TwilioService.new(:ctc).send_text_message(to: signup.phone_number, body: message.sms_body)
```
jenny-heath, line 37: this is actually not just a CTC thing! see: https://www.getyourrefund.org/en/sign-up

app/jobs/send_outgoing_text_message_job.rb
```diff
@@ -4,7 +4,7 @@ class SendOutgoingTextMessageJob < ApplicationJob
   def perform(outgoing_text_message_id)
     outgoing_text_message = OutgoingTextMessage.find(outgoing_text_message_id)
     begin
-      message = TwilioService.send_text_message(
+      message = TwilioService.new.send_text_message( # TODO: set explicit service scope; is this job for :ctc or :gyr?
```
jenny-heath, line 7: [DUST] again, would like to remove comments before merging

jenny-heath:
just a general question about this PR: what was the thinking behind refactoring this from a singleton into a class that gets instantiated with a service type (as opposed to, say, passing the service type into `send_text_message`)? not opposed to this approach, just wondering about the thinking behind it.

related to that: do we want to consolidate the twilio subaccounts into just the main account? i feel like we talked about this on friday but wasn't sure how serious we are about it. this approach accounts for having different account info (account sid + auth token) per service type but if we just have the one maybe that would affect the implementation

anyway, assuming we will discuss verbally just wanted to write down my thoughts before i forget them

app/services/twilio_service.rb
jnf: yes! for sure. i literally don't know the answer but would like to. 😅

app/models/signup.rb
jnf: that makes me moderately unhappy. but it's kinda sorta ok as right now ctc and gyr seem to be sharing the same twilio setup.

jnf:
I prefer an instantiated class for a couple of reasons:
- send_text_message isn't the only method on TwilioService that needs info about which tenant service it's working with (client_metadata does too).
- limits what caller's need to know about the internals. i like the general pattern of saying "i need to talk to twilio as fyst" and the TwilioService managing what that means behind the scenes (which is why i also replaced direct calls to Twilio::REST::Client with an instance of TwilioService

re: consolidating accounts. Yes, that's a very real thing and this PR assumes that work is _done_. I think we'll end up with 2 accounts (down from 5)--one main account that'll house the FYST-related services/assets, and one subaccount that'll do the same for GYR (with CTC as a technically also there deprecated sidecar). Each account will have many messaging services--one per supported environment--with each messaging service having a specific sender pool and environment configuration.

Hope this all makes sense and isn't too disrupting. Lemme know if you wanna talk through it more; happy to get on a call about any/all of it!

app/jobs/send_outgoing_text_message_job.rb
jnf: absolutely. do you know if/how this job is aware of it's tenant service/environment?

app/services/twilio_service.rb
```diff
@@ -12,104 +13,101 @@ class TwilioService
...
-    end
+    outgoing_text_message&.update(status_key => "twilio_error")
 
-    def client
-      @@_client ||= Twilio::REST::Client.new(
-        EnvironmentCredentials.dig(:twilio, :account_sid),
-        EnvironmentCredentials.dig(:twilio, :auth_token)
-      )
+    unless e.code == 21211 # Invalid 'To' Phone Number https://www.twilio.com/docs/api/errors/21211
+      raise # should we include the original exception here (e)??
```
jenny-heath, line 48: fix as a follow-up task & remove comment

spec/controllers/twilio_webhooks_controller_spec.rb
```diff
@@ -1,6 +1,12 @@
 require "rails_helper"
 
 RSpec.describe TwilioWebhooksController do
+  let (:twilio_service) { instance_double(TwilioService) }
```
jenny-heath, line 4: delete space after let

spec/jobs/get_phone_metadata_job_spec.rb
```diff
@@ -37,14 +37,14 @@
...
-          expect(TwilioService).not_to have_received(:get_metadata)
+          expect(twilio_service).not_to have_received(:get_metadata)
         end
       end
     end
 
     context "when Twilio raises an exception" do
       before do
-        allow(Twilio::REST::Client).to receive(:new).and_raise(Twilio::REST::RestError.new(400, OpenStruct.new(body: {}, status_code: 21211)))
+        allow(twilio_service).to receive(:send_text_message).and_raise(Twilio::REST::RestError.new(400, OpenStruct.new(body: {}, status_code: 21211)))
```
jenny-heath, line 47: should probably mock `:get_metadata` instead

jenny-heath (approved):

PR #4950, opened 2024-11-06
FYST-1027 ID Form39R xml and pdf additions/subtractions

arinchoi03:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1027

## What was done?
- Implemented `TotalAdditions` and `TotalSubtractions` to ID39R xml/pdf
- filled out Additions (always 0 b/c OOS), Subtractions lines 7, 8f (placeholder 0), 24 (sum of 3, 6, 7, 8f, and 18)

## How to test?
  - Unit tests
  - Test data used: tim_1099_int
- Specify any relevant testing environments used: development/heroku
- Risk Assessment
  - ~~I did refactor the ChildCareCreditAmt a tiny bit, make sure the XML should still fill out as before when it should~~
 
Screenshot 2024-11-06 at 3 12 59 PM
Screenshot 2024-11-06 at 3 13 15 PM
Screenshot 2024-11-06 at 3 13 19 PM


app/lib/pdf_filler/id39r_pdf.rb
```diff
@@ -16,9 +16,13 @@ def initialize(submission)
 
     def hash_for_pdf
       answers = {
-        "BL6" => @xml_document.at('Form39R ChildCareCreditAmt')&.text,
```
arinchoi03, line 19: i removed the "Form39R" to be more consistent here

mpidcock: [dust] I think "Form39R" was a nice guard to have in place in case there's another `ChildCareCreditAmt` in the xml. Fortunately, the federal amount is called `FedChildCareCreditAmt`, so there's no chance of collision in this case, but it would have been reasonable to leave it in.

mpidcock (approved):
Looks great!

app/lib/pdf_filler/id39r_pdf.rb
arinchoi03: Hmm would it be good practice to always prepend the selector with the form name?

arinchoi03: cc @powersurge360 since I asked why you did this, maybe this will be just good practice

mpidcock: It's a good practice, but without a known naming collision I don't think changing it is needed.

arinchoi03: 👍 just changed it since it was easy enough to do & I want people to follow this precedence just in case

PR #4953, opened 2024-11-07
Send auto response on join keyword

squanto:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1006

## What was done?
- Adds a condition in `IncomingTextMessageService` so that incoming messages that are just the text "join" or "start" receive an automated reply
- The ticket mentions just sending the message, with follow up for actual "join" logic to follow up
## How to test?
- Unsure, but if there are numbers registered to the same credentials that our heroku pr apps use, we can test by texting one of their registered numbers the keyword "JOIN" or "START" and verify the response.

### DIFF
--- a/app/models/automated_message/join_response.rb
+++ b/app/models/automated_message/join_response.rb
@@ -1,0 +1,11 @@
+module AutomatedMessage
+  class JoinResponse < AutomatedMessage
+    def self.name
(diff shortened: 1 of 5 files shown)


squanto:
Closing in favor of messaging via the Twilio console

PR #4957, opened 2024-11-08
Fix signature date tests

tahsinaislam:

## What was done?
Refactored some tests for signature dates on pdfs so that it will pass no matter when it is run. Previously tests could not be run after midnight UTC because the signature date would be set on the date of the timezone. Now all signature dates set to specific time instead of time.now.

## How to test?
- Updated unit tests


mrotondo (approved):
This fixes the issue where certain tests would fail when run after midnight in the state's time zone, but before midnight in the test runner's time zone.

PR #4958, opened 2024-11-08
NJ sales use tax - fix bug preventing field input

aloverso:
## Link to pivotal/JIRA issue
?

## What was done?
- Fix field so that you can type in it

## How to test?
- Go to sales use tax screen, select yes/manual, type in field

Before:

https://github.com/user-attachments/assets/231aedb4-5282-4255-ab42-3e794677acc4

### DIFF
--- a/app/views/state_file/questions/nj_sales_use_tax/edit.html.erb
+++ b/app/views/state_file/questions/nj_sales_use_tax/edit.html.erb
@@ -25,8 +25,8 @@
       </div>
       <div class="question-with-follow-up__follow-up" id="sut-method-field">
         <div class="question-with-follow-up">
-          <div class="question-with-follow-up__question">
-            <div class="white-group">
+          <div class="white-group">
+            <div class="question-with-follow-up__question">
               <%=
                 f.cfa_radio_set(
                   :sales_use_tax_calculation_method,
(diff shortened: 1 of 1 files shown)


aloverso:
I'm not actually sure why this works (or why it broke in the first place). It was just a hunch that the div arrangement seemed wrong. Hunch fixed the issue so here we are.

mluedke2:
I notice in the `After` video that it's possible to put three numbers after the decimal -- seems like other fields allow that too so it's out of scope for this PR, but is there a ticket someplace for that?

aloverso:
@mluedke2 Hm, **should** there a limit on decimal places? We round it anyways so do we actually need to limit to two places?

tristanvanech:
@aloverso the one thing I noticed playing around with decimals is if you put $10.499 it rounds to $11 which doesn't seem right. Seems like it'd be a typo and user meant to put $10.49, which would round down. This will probably never happen though lol

mluedke2 (approved):

mluedke2:
@aloverso it would be for user convenience and to limit entry errors in money fields. We can use the paper tax forms as inspiration too-- there aren't more than two boxes after decimals in those. I would kind of flip the question around and expect there to be a specific reason to allow three or more decimals (I'm no accountant so maybe there is, but I'd be surprised).

arinchoi03:
@aloverso i'm guessing this might have to do with the honeycrisp update since there were javascript updates related to the follow up questions component -- may have to do some digging

PR #4964, opened 2024-11-12
Updates to AZ-140 XML & PDF

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-259

## What was done?
This PR makes the following changes: 
* adds the line 28 to the calculation for line 35 in az-301
* populates line 51 in az-140 from line 62 in az-140
* adds testing for line 52 in az-140 in the calculator 
* refactor of lines 43, 45, 46, 49, 50, 51 in az140 pdf

## How to test?
Test that data shows up correctly for lines 28 and 35 in az-140.


app/lib/efile/az/az140_calculator.rb
```diff
@@ -141,9 +141,9 @@ def calculate_line_32
 
       def calculate_line_35
         # Subtotal after additions and subtractions
-        subtractions = line_or_zero(:AZ140_LINE_29A) + line_or_zero(:AZ140_LINE_29B)
-        (30..32).each do |line_num|
-          subtractions += line_or_zero("AZ140_LINE_#{line_num}")
+        subtraction_lines = ["28", "29A", "29B"] + (30..32).to_a
```
anisharamnani, line 144: i don't know if this is more misleading than the previous implementation.

arinchoi03: this is nice imo! 30..32 are subtraction lines too so this makes sense.

app/lib/pdf_filler/az140_pdf.rb
```diff
@@ -106,6 +106,7 @@ def hash_for_pdf
         "48" => @xml_document.at('Form140/DeductionAmt/SubTotal')&.text,
         "49" => @xml_document.at('DepTaxCredit')&.text,
         "50" => @xml_document.at('FamilyIncomeTaxCredit')&.text,
+        "51" => @xml_document.at('DeductionAmt CreditsFromAZ301')&.text,
```
arinchoi03, line 109: [DUST] Noticing here that there are variety of ways we're specifying the node that is nested within another one (see `Form140/DeductionAmt/SubTotal`) -- the directly sibling selectors (ex: `FamilyIncomeTaxCredit`, `DepTaxCredit`) are all not specified with `DeductionAmt` -- what do you think about adding the specificity to the other selectors, if we're going to do this for `CreditsFromAZ301`?

arinchoi03 (approved):
just a comment, i don't think it matters too much but wondering what you think -- otherwise looks good! 

note for future: discussed async that personas were updated due to the failing tests that updated `nil` values to `'0'`

app/lib/pdf_filler/az140_pdf.rb
anisharamnani: ah, i don't mind adding the specificity to the other selectors. i prefer the specificity! 

tbh, i also did not realize that the specificity was not needed.

arinchoi03 (approved):

app/lib/pdf_filler/az140_pdf.rb
anisharamnani: just an update here! i refactored the subtraction nodes into their own section as can be seen on line 121 of this file. the only one not included is for "48." my theory is that Subtotal is used throughout the XML, and so there is a need to be specific by calling `Form140 DeductionAmt SubTotal`.

anisharamnani: also slashes seem to add more specificity (even opposed to spaces). i do not know why at this time.

PR #4970, opened 2024-11-13
Fyst 971 pull household information directly from df json

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-971
- https://codeforamerica.atlassian.net/browse/FYST-989

## What was done?
- Removed the NameDob controller/form/view
- Pull in the "months in home" value from the DF JSON into `StateFileDependent.months_in_home`, and handle the fact that we will only get values of 6 and up (or null) in that JSON field
- Changed all our uses of `StateFileDependent.relationship` to use the new non-gendered relationship names that we pull in from the DF JSON
- Removed "edit" link from household info on the review page
- Changed the default EfileError controller to DataReview, which is the page immediately after DF import. I don't love this, but I don't know of a better option. I've asked PM to weigh in on the story
- Updated many many tests to account for the above changes (different relationship names, use DF JSONs instead of specifying name/dob/months in home in test config, don't expect to see the NameDob page in feature tests, etc)
 
## How to test?
- Ensure that the NameDob page doesn't show up in any state flows
- Ensure that (when you use a persona that has a JSON set up) all filer & dependent info from that JSON is used correctly in our output XML & PDF. You can see which persona have JSONs by looking inhttps://github.com/codeforamerica/vita-min/tree/FYST-971-pull-household-information-directly-from-df-json/spec/fixtures/state_file/fed_return_jsons/2023


mpidcock:
I'm just passing through, but pls fill out the PR template and address failing tests 🙏

anisharamnani (approved):
🎉 i don't have any comments, things look good & i appreciate that the tests have been updated to reflect the way of handling the data. thanks @mrotondo :)

app/forms/state_file/name_dob_form.rb
```diff
@@ -1,127 +0,0 @@
...
-    validate :spouse_birth_date_is_valid_date, if: -> { @intake.ask_spouse_dob? }
-    validates :primary_first_name, format: { with: /\A[a-zA-Z]{1}([A-Za-z\-\s']{0,15})\z/.freeze, message: ->(_object, _data) { I18n.t('errors.attributes.first_name.invalid_format') }}
-    validates :primary_middle_initial, length: { maximum: 1 }, format: { with: /\A[A-Za-z]\z/.freeze, allow_blank: true }
-    validates :primary_last_name, format: { with: /\A[a-zA-Z]{1}([A-Za-z\-\s']{0,137})\z/.freeze, message: ->(_object, _data) { I18n.t('errors.attributes.last_name.invalid_format') }}
-    validates :spouse_first_name, format: { with: /\A[a-zA-Z]{1}([A-Za-z\-\s']{0,15})\z/.freeze, message: ->(_object, _data) { I18n.t('errors.attributes.first_name.invalid_format') }}, if: -> { @intake.ask_spouse_name? }
-    validates :spouse_middle_initial,  length: { maximum: 1 }, format: { with: /\A[A-Za-z]\z/.freeze, allow_blank: true }, if: -> { @intake.ask_spouse_name? }
-    validates :spouse_last_name, format: { with: /\A[a-zA-Z]{1}([A-Za-z\-\s']{0,137})\z/.freeze, message: ->(_object, _data) { I18n.t('errors.attributes.last_name.invalid_format') }}, if: -> { @intake.ask_spouse_name? }
```
arinchoi03, line 38: bunch of validations that were previously being enforced is no longer active -- this is fine b/c we're just trusting DF json data right?

app/models/state_file_dependent.rb
```diff
@@ -32,20 +32,33 @@
...
+    "grandParent" => "Grandparent",
+    "otherAncestorOfParent" => "Grandparent",
+    "stepParent" => "Parent",
+    "parentInLaw" => "Parent",
+    "noneOfTheAbove" => "Other",
+    "siblingOfParent" => "?",
+    "otherDescendantOfHalfSibling" => "?",
+    "otherDescendantOfStepSibling" => "?",
+    "fosterParent" => "?",
+    "siblingsSpouse" => "?",
```
arinchoi03, line 61: how is the `?` value used?

spec/factories/state_file_ny_intakes.rb
```diff
@@ -116,6 +116,7 @@
 
     raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.old_xml_sample }
     raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.old_json_sample }
+    df_data_import_succeeded_at { DateTime.now }
```
arinchoi03, line 119: oh maybe i should've added this when i added the column, it makes sense -- except in situations we may want an unimported intake for testing? but then I guess we manually set the `df_data_import_succeeded_at` to `nil`

app/models/direct_file_json_data.rb
```diff
@@ -17,6 +18,29 @@ class DfJsonDependent < DfJsonPerson
...
+
+    WORDS_TO_NUMBERS = {
+      "six" => 6,
+      "seven" => 7,
+      "eight" => 8,
+      "nine" => 9,
+      "ten" => 10,
+      "eleven" => 11,
+      "twelve" => 12
+    }
```
arinchoi03, line 30: since not having 1-5 is a bit mystifying here, could we add a comment to clarify why those don't exist here?

arinchoi03 (approved):
real quick review as i'm working off of spotty internet but i promised some comments so here are some! nothing blocking i think

spec/lib/efile/az/az140_calculator_spec.rb
jenny-heath, line 1: i like these changes but i also can't tell how they're related to this PR

spec/fixtures/state_file/fed_return_jsons/2023/ny/zeus_8_deps.json
jenny-heath, line 1: man it is such a bummer that we're having to do things to maintain NY code

app/models/state_file_dependent.rb
```diff
@@ -78,17 +90,25 @@ def full_name
...
+  end
+
+  def months_in_home_for_xml
+    months_in_home || 5
+  end
+
   def ask_senior_questions?
     return false if dob.nil?
-    senior? && months_in_home == 12 && ['PARENT', 'GRANDPARENT'].include?(relationship)
+    senior? && months_in_home == 12 && ['parent', 'grandParent', 'otherAncestorOfParent'].include?(relationship)
```
arinchoi03, line 103: do `stepParent`, `siblingOfParent`, `parentInLaw`, `fosterParent` not count for these scenarios?

app/models/state_file_dependent.rb
```diff
@@ -78,17 +90,25 @@ def full_name
...
   end
 
   def is_qualifying_parent_or_grandparent?
     ask_senior_questions? && needed_assistance_yes?
   end
 
   def is_hoh_qualifying_person?
-    relationship == 'PARENT' || (relationship != 'NONE' && (months_in_home || 0) >= 6)
+    relationship == 'parent' || (relationship != 'noneOfTheAbove' && (months_in_home || 0) >= 6)
```
arinchoi03, line 111: does `stepParent` or `parentInLaw` not count here?

app/forms/state_file/name_dob_form.rb
mrotondo: IMO yes, we shouldn't throw validation errors on information we don't collect from the user but get directly from DF. if there are incompatibilities between this information and our output schemas, those should be addressed (as best as possible) with output filtering/truncation, which I believe is already in place

app/models/state_file_dependent.rb
mrotondo: Good catch - I meant to replace these with something and didn't get back to that before asking for review. I'll map these to what I think are sensible values and ask for PM input on the story.

spec/factories/state_file_ny_intakes.rb
mrotondo: Agreed, the default should be success and we can override with nil on this and the raw_*_data columns when testing failure

spec/lib/efile/az/az140_calculator_spec.rb
mrotondo: Because we got rid of `NameDobForm`, I changed the validation in `StateFileDependent` from
```
validates_presence_of :first_name, :last_name, :dob, on: :dob_form
```
to
```
validates_presence_of :first_name, :last_name, :dob
```
which meant that all the dependents we were creating without names were no longer valid and weren't getting saved to the database. Some tests still passed (if they didn't query for dependents but just looked at the variables created in the test), but enough broke that I decided to make a whole-codebase change to use the factory, which provides a default name (and we should be using factories for test data anyway)

app/models/state_file_dependent.rb
mrotondo: Dunno - forwarding to PM/program

PR #4974, opened 2024-11-13
Fyst 1088 move financial transaction to shared xml other cleanup

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1088

## What was done?
- Moved the financial transaction out into a shared folder and then modified XML to include it.
- Also includes a fix for a broken tax refund form
- Stubs for refunds have been changed to a positive value to allow for it to show up

## How to test?
- For each of ID, NC, and MD step through the process off creating a tax return
- At the end, fill out the routing number and account number and mark direct deposit. Note that these cannot match
- After clicking through, click on the xml and near the bottom it should show the `FinancialTransaction` element.

image


app/lib/efile/md/md502_calculator.rb
```diff
@@ -79,7 +79,7 @@ def calculate
       end
 
       def refund_or_owed_amount
-        0
+        50
```
powersurge360, line 82: These are stubs anyways so I updated them to be positive so that the financial transaction element will show up.

anisharamnani: [dust] is there any way to note that these are stubbed possibly with a comment? or, do you feel like that's not completely necessary?

anisharamnani (approved):
I had the tiniest of comments, but this looks good, thanks @powersurge360 !

app/lib/efile/md/md502_calculator.rb
powersurge360: These are pretty big functionality methods so I'm probably going to leave it alone for now. I think I'll be touching these again on my next ticket anyways.

PR #4976, opened 2024-11-13
Handle DirectFile Data Import failures while syncing

arinchoi03:
[Computer Garden Seed](https://www.notion.so/cfa/Should-ImportFromDirectFileJob-synchronize_-_to_database-failures-be-more-obvious-for-non-prod-12f373fd79b2809fad32e845a58c9d92?pvs=4)
[Slack convo](https://cfastaff.slack.com/archives/C0544ERAFQV/p1731530208892769)
Add df_data_import_succeeeded to replace failed_at since if the sync fails the failed_at field does not update
With this PR, if an intake has ever succeeded with data transfer, it will always remain successful (allow user to proceed with rest of the intake flow)

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1171

## What was done?
- When the one of the `synchronize_df_*` methods fails due to one of the validation checks failing on an intake's association, the `df_data_import_failed_at` can never be properly set. Therefore, we are pivoting to use the `df_data_import_succeeded_at` flag instead, which _will_ set properly after all the syncs run through _successfully_ (if anything fails during sync, it'll raise an error but not attempt to save the intake/associated records)
- I added a fixture to demonstrate the failure (adding unaccepted/invalid characters on a w2 `miranda_1099r_with_df_w2_error`) -- is it too much? I had a hard time trying to write the spec that would catch the exact thing i was seeing
## How to test?
  - Unit/manual tests
  - Test data used:
    - `miranda_1099r_with_df_w2_error`: it allows you to proceed with incomplete data on demo, it should show the direct file data transfer failed page on this heroku environment.

- Risk Assessment
  - There was existing logic to ensure that if the transfer was successful, setting the `df_data_import_failed_at` to `nil` (if it previously failed). I don't know if there's a case where we would clear the `df_data_import_succeeded_at` to `nil` on an unsuccessful transfer after a successful one? Is this even possible?? need more thoughts on this. 
  - **other recommendations**: Mike mentioned that if we can transfer the same intake multiple times, we should create `df_import_attempts` for an intake, each of which would have multiple `df_data_import_errors`, instead of the current set-up where a single intake has multiple `df_data_imports_errors`.


arinchoi03:
Still need to add `df_data_import_failed_at` to `ignored_columns` if we decide this is the path forward

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -34,11 +34,9 @@ def perform(authorization_code:, intake:)
         intake.synchronize_df_w2s_to_database
         intake.synchronize_filers_to_database
 
-        # Clear this timestamp if it failed before but succeeded this time
-        intake.update(df_data_import_failed_at: nil)
```
arinchoi03, line 38: how/why would this happen?

db/migrate/20241113224054_add_df_data_import_succeeded_at_to_intakes.rb
```diff
@@ -0,0 +1,10 @@
+class AddDfDataImportSucceededAtToIntakes < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_nc_intakes, :df_data_import_succeeded_at, :datetime, null: true
+    add_column :state_file_az_intakes, :df_data_import_succeeded_at, :datetime, null: true
+    add_column :state_file_md_intakes, :df_data_import_succeeded_at, :datetime, null: true
+    add_column :state_file_id_intakes, :df_data_import_succeeded_at, :datetime, null: true
+    add_column :state_file_ny_intakes, :df_data_import_succeeded_at, :datetime, null: true
```
arinchoi03, line 7: adding to `ny` for completeness but not sure if we need this

app/jobs/state_file/import_from_direct_file_job.rb
arinchoi03: found the implementation/comment here https://github.com/codeforamerica/vita-min/pull/4295

I'm unfamiliar with how we "create a return" in the DF system, come into FYST and when the intake gets created during that flow -- if they're coming from DF -> FYST, do they always hit the landing page (thus invalidating a previous intake & creating a current intake)

```
# app/controllers/state_file/landing_page_controller.rb
    def update
      sign_out current_intake if current_intake.present?
      intake = StateInformationService.intake_class(params[:us_state]).new(
        visitor_id: cookies.encrypted[:visitor_id],
        source: session[:source],
        referrer: session[:referrer]
      )
      intake.save
      intake.create_state_file_analytics!
      sign_in intake
```

This logging out of current session/intake seem like a new behavior _added_ after this comment ([here's the PR](https://github.com/codeforamerica/vita-min/pull/4655)), so maybe that concern is no longer present.

app/models/state_file_id_intake.rb
```diff
@@ -76,6 +77,8 @@
 #  index_state_file_id_intakes_on_spouse_state_id_id   (spouse_state_id_id)
 #
 class StateFileIdIntake < StateFileBaseIntake
+  self.ignored_columns = [:df_data_import_failed]
```
embarnard, line 80: Should we just add this line to the `StateFileBaseIntake` instead if its ignored on all the states intakes?

arinchoi03: oh great suggestion, will do!

arinchoi03:
uncommented the acceptance specs and xml is failing to build b/c of below errors:

```
+<html><head><meta name="color-scheme" content="light dark"></head><body><pre style="word-wrap: break-word; white-space: pre-wrap;">199:0: ERROR: Element '{http://www.irs.gov/efile}QualCharityContrDate': [facet 'minInclusive'] The value '2023-06-21' is less than the minimum value allowed ('2024-01-01').
+214:0: ERROR: Element '{http://www.irs.gov/efile}SchoolContrDate': [facet 'minInclusive'] The value '2023-06-21' is less than the minimum value allowed ('2024-01-01').
```

this has to do with the `config.statefile_current_tax_year = 2023` -- may need help fixing this. Can I update the current tax year?

arinchoi03:
accidentally added commits to the wrong PR & reverted them 😕 hopefully this doesn't cause issues down the line

embarnard (approved):

app/jobs/state_file/import_from_direct_file_job.rb
mpidcock: belatedly capturing what we discussed here:

* The error we saw before was that DF would have a brief outage, resulting in a timestamp for `df_data_import_failed_at`
* This error was ephemeral, and after simply hitting back and trying again, the df data would appear on the intake
* However, the error timestamp was never cleared, so the user would still see the failed import page, despite it suceeeding
* The only way to proceed was tho make a new intake that hadn't failed, then the user could continue with a successful import

Arin and I agreed that this new implementation to store `succeeded_at` will not have the same issue, since once the data has been successfully transferred it will not be deleted, so there is no need to update the timestamp once it has been set.

PR #4977, opened 2024-11-14
Fix id xml order

mrotondo:
This addresses an XML schema validation error generated by elements which were added out-of-order due to me not checking the schema during a manual merge. Because this would have been caught by a test that checked XML schema validity, I added those tests for all states where our XML is currently schema-valid (all except MD, which I left commented out)


tahsinaislam (approved):
lgtm!

PR #4981, opened 2024-11-14
Fix a crash in DirectFileJsonData#initialize that was breaking DF import

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1169

## What was done?
- In `DirectFileJsonData#initialize`, we were calling `JSON.parse` on the argument, despite only ever using this class to to dig into a value stored in `StateFileXxIntake::raw_direct_file_intake_data`, which is a `jsonb` column. This was causing a crash when `JSON.parse` was called with a `Hash` argument. I removed the `JSON.parse` and fixed our tests by adding a JSON.parse to `DirectFileApiResponseSampleService#read_json`, which should have been there this whole time.
## How to test?
- To test that this fix is correct, you should go on demo (once this is merged to main) & attempt to import the return described in the ticket


mrotondo:
> Looks good and thank you for updating the factories to have `raw_direct_file_intake_data`! My only question is if we should mention this change to NJ for any reason, but that doesn't block this.

@mpidcock NJ is (as usual :]) way ahead of us and already had jsons on all their intake factories & traits

PR #4988, opened 2024-11-15
Correct blinking disabled for nested question-with-follow-up

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1075

## What was done?
- This changes the nested follow up normalization code to only work on direct descendants. This stops the issue of nested question with follow up boxes from being picked up
- 🚨 This feature is not unit tested. It may be that there are places where this will break existing functionality. I went through and checked as much as I could think of but it's something to be vigilant for.

## How to test?
- Navigate through NC flow
- Transfer any persona
- Select `yes` on `nc-sales-use-tax` page
- Select second option `I kept a complete
- Click money field input box

### DIFF
--- a/app/javascript/lib/honeycrisp.js
+++ b/app/javascript/lib/honeycrisp.js
@@ -121,7 +121,7 @@ var followUpQuestion = (function () {
             });
         },
         update: function ($container) {
-            $container.find('> .question-with-follow-up__follow-up input, .question-with-follow-up__follow-up select').attr('disabled', true);
+            $container.find('> .question-with-follow-up__follow-up input, > .question-with-follow-up__follow-up select').attr('disabled', true);
             $container.find('> .question-with-follow-up__follow-up').hide();
 
             $container.find('.question-with-follow-up__question input').each(function (index, input) {
--- a/app/javascript/lib/honeycrisp.js
+++ b/app/javascript/lib/honeycrisp.js
@@ -121,8 +121,8 @@ var followUpQuestion = (function () {
             });
         },
         update: function ($container) {
-            $container.find('.question-with-follow-up__follow-up input, .question-with-follow-up__follow-up select').attr('disabled', true);
-            $container.find('.question-with-follow-up__follow-up').hide();
+            $container.find('> .question-with-follow-up__follow-up input, .question-with-follow-up__follow-up select').attr('disabled', true);
+            $container.find('> .question-with-follow-up__follow-up').hide();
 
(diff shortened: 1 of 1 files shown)


app/javascript/lib/honeycrisp.js
```diff
@@ -121,8 +121,8 @@ var followUpQuestion = (function () {
             });
         },
         update: function ($container) {
-            $container.find('.question-with-follow-up__follow-up input, .question-with-follow-up__follow-up select').attr('disabled', true);
-            $container.find('.question-with-follow-up__follow-up').hide();
+            $container.find('> .question-with-follow-up__follow-up input, .question-with-follow-up__follow-up select').attr('disabled', true);
+            $container.find('> .question-with-follow-up__follow-up').hide();
```
arinchoi03, line 125: just a note that this _used_ to exist in the previous followup implementation but it was removed in this PR https://github.com/codeforamerica/honeycrisp-gem/pull/312/files we might not have any instances of followup questions after none of the above checkboxes

powersurge360: I'm not sure I see the direct descendent selector in that diff?

arinchoi03: Oops wrong one -- I'll post the right one after lunch

powersurge360:
Mike noted that even though the UX is correct, the form fields in nested attributes may still be active. Users will have to go out of their way to get to that funky functionality but it is possible. A full refactor of the JS for follow up questions (and perhaps even some changes into the structure of the HTML) is likely to solve all sharp edges.

app/javascript/lib/honeycrisp.js
arinchoi03: https://github.com/codeforamerica/vita-min/commit/4f6da694d8e943320ac1536dbd6c19830933a499

I was mistaken- it seemed like we solved for this case a long time ago and it was a tax-specific code addition to honeycrisp.js that was never implemented in the gem, so when I copied it over we lost the existing fix

arinchoi03:
@powersurge360 can we document how we might get into that funky state & create a ticket/issue?

arinchoi03 (approved):

powersurge360:
> @powersurge360 can we document how we might get into that funky state & create a ticket/issue?
> 
> cc @aloverso on the input blinking issue

https://codeforamerica.atlassian.net/browse/FYST-1211

PR #4991, opened 2024-11-15
Refactor refund_or_owed_amount to be defined on parent class and use constants on child classes

jenny-heath:
This is some preliminary work for FYST-799. It doesn't contain any functional changes so I'd like to merge it in.

## What was done?
- The `refund_or_owed_amount` method, as far as I can tell, should be the same everywhere. It is used to determine whether clients see the refund or owed page in the flow, and to display an amount on the review page. It should subtract the owed amount from the refund amount, one of which should always be 0 because you can't owe and get a refund at the same time, so that you get a positive amount if refund and negative if owed.
- Open to other approaches to this, e.g. not using constants. I figure this is pretty flexible because child classes can just overwrite the method if it works differently for a state.
## How to test?
- I added tests for all states even the ones that are stubbed for now. Hopefully they can be a guide so that when this calc is added for states that don't have it yet, people remember to actually test it. I think it makes sense to have one test per state just to check that the correct lines are defined for refund/owed.

### DIFF
--- a/app/lib/efile/az/az140_calculator.rb
+++ b/app/lib/efile/az/az140_calculator.rb
@@ -2,6 +2,8 @@ module Efile
   module Az
     class Az140Calculator < ::Efile::TaxCalculator
       attr_reader :lines
+      REFUND_LINE_NUM = :AZ140_LINE_79
+      OWED_LINE_NUM = :AZ140_LINE_80
 
       def initialize(year:, intake:, include_source: false)
         super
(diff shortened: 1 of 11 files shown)


embarnard (approved):
looks good!

powersurge360 (approved):
Looks good! You may consider some porcelain to make the API a little nicer, but I'm not going to get in the way on it.

```ruby
class TaxCalculator
  delegate :refund_line, :owe_line, to: :class

  class << self
    attr_accessor :refund_line, :owe_line

    def set_refund_lines(refund: 0, owe: 0)
      self.refund_line = refund
      self.owe_line = owe
    end
  end
end

class IdCalculator < TaxCalculator
  set_refund_lines refund: :foo, owe: :bar
end

class NcCalculator < TaxCalculator
  set_refund_lines refund: :foo
end
```

PR #5002, opened 2024-11-19
Fyst 1077 remove DF API server public certs, as they are not needed

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1077
## Is PM acceptance required?

## What was done?
- The original story said to add the 2025 API server, intermediate and root CA certificates to our http requests, but it turns out that we don't seem to actually need to those attached to our requests - when I pushed the code to staging with neither the old nor the new certificate chain, it still worked: When using `ImportFromDirectFileJob` or `IrsApiService#import_federal_data` from the rails console, I was able to make API requests and import DF data.
- I also added some additional error-handling code around in `IrsApiService#import_federal_data`, as the `JSON.parse` could fail if response.body was nil


squanto (approved):
lgtm

PR #5006, opened 2024-11-19
add income review card to final review screen

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1084

## What was done?
- Added a new card to the review header, which will show up for all states.
 - Note that this is not correct for NY, since the income review page was never added to the NY Question Navigation (flow), but I'm not addressing that since we aren't planning to ship with NY
## How to test?
- Go through the full flow for each state, and verify that the income details review card shows up, has an edit button that leads back to the income review page, and that clicking continue from that page leads back to the final review page
image

### DIFF
--- a/app/controllers/state_file/questions/income_review_controller.rb
+++ b/app/controllers/state_file/questions/income_review_controller.rb
@@ -1,6 +1,7 @@
 module StateFile
   module Questions
     class IncomeReviewController < QuestionsController
+      include ReturnToReviewConcern
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/state_file/questions/income_review_controller.rb
```diff
@@ -1,6 +1,7 @@
 module StateFile
   module Questions
     class IncomeReviewController < QuestionsController
+      include ReturnToReviewConcern
```
arinchoi03, line 4: looks good to me -- can we add the spec to test `it_behaves_like :return_to_review_concern`?

arinchoi03 (approved):
one suggestion but code looks good

PR #5009, opened 2024-11-19
Implement ID Form 40 XML & PDF - Payments, Tax Due or Refund, Direct Deposit

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-1068?atlOrigin=eyJpIjoiNWRmNDllMTNjZDg2NGNkY2E5NzU0NTg5M2M1ZmM5M2IiLCJwIjoiaiJ9
## What was done?
-  implement the Payments & Other Credits, Tax Due or Refund, Direct Deposit sections of ID Form 40 along with the related ID state XML fields.
## How to test?
- unit tests, can test in heroku using intakes where line 42 is greater than line 50 and where line 50 is greater than line 42

### DIFF
--- a/app/lib/efile/id/id40_calculator.rb
+++ b/app/lib/efile/id/id40_calculator.rb
@@ -36,6 +36,11 @@ module Efile
         set_line(:ID40_LINE_43_DONATE, :calculate_line_43_donate)
         set_line(:ID40_LINE_43, :calculate_line_43)
         set_line(:ID40_LINE_46, :calculate_line_46)
+        set_line(:ID40_LINE_50, :calculate_line_50)
+        set_line(:ID40_LINE_51, :calculate_line_51)
+        set_line(:ID40_LINE_54, :calculate_line_54)
+        set_line(:ID40_LINE_55, :calculate_line_55)
+        set_line(:ID40_LINE_56, :calculate_line_56)
         @id39r.calculate
         @lines.transform_values(&:value)
       end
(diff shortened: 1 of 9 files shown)


app/lib/efile/id/id40_calculator.rb
```diff
@@ -207,8 +210,29 @@ def calculate_line_46
...
-      def calculate_line_54; end
-      def calculate_line_56; end
+      def calculate_line_50
+        line_or_zero(:ID40_LINE_43) + line_or_zero(:ID40_LINE_46)
+      end
+
+      def calculate_line_51
+        if line_or_zero(:ID40_LINE_42) > line_or_zero(:ID40_LINE_50)
+          line_or_zero(:ID40_LINE_42) - line_or_zero(:ID40_LINE_50)
+        end
```
arinchoi03, line 220: should the `else` be specified here as `0` (in the case line 50 > line 40 value) or is that not a significant difference at the end of the day since `line_or_zero` is being called on the `:ID40_LINE_51` value?

app/lib/efile/id/id40_calculator.rb
```diff
@@ -207,8 +210,29 @@ def calculate_line_46
...
+      end
+
+      def calculate_line_54
+        line_or_zero(:ID40_LINE_51)
+      end
+
+      def calculate_line_55
+        if line_or_zero(:ID40_LINE_42) < line_or_zero(:ID40_LINE_50)
+          line_or_zero(:ID40_LINE_50) - line_or_zero(:ID40_LINE_42)
+        end
```
arinchoi03, line 230: same as above

arinchoi03 (commented):
i see we handle cases of 
`line_or_zero(:ID40_LINE_42) < line_or_zero(:ID40_LINE_50)`
and 
`line_or_zero(:ID40_LINE_42) > line_or_zero(:ID40_LINE_50)` 
but what should happen if they equal each other? Are both values 0?

app/lib/submission_builder/document.rb
```diff
@@ -99,6 +99,11 @@ def add_non_zero_value(xml, elem_name, line)
       end
     end
 
+    def add_element_if_present(xml, tag, line_id)
+      value = calculated_fields.fetch(line_id)
+      xml.send(tag, value) if value.present?
+    end
```
arinchoi03, line 105: if `value` is not present here, it won't `send` to the xml which means that `TaxDue` node will _not_ exist? (because it hasn't been "added"? Or does that node already exist but the value is not being filled? If latter, i'd rename the method to something like `fill_element_if_present`

arinchoi03: i see this was a method that already existed -- i'd still love to know if I understand this correctly though!

embarnard: @arinchoi03 Yes to the first. If the value doesn't exist (is nil), then it won't create the element and add it to the XML

app/lib/efile/id/id40_calculator.rb
embarnard: @arinchoi03 these elements would not appear if they are equal values which is the intention. I'll double check with Margaret but I don't think we want to display a refund or payment if they don't have one (is zero). And since the refund_or_owed_amount method in the base calculator class also uses `line_or_zero` I don't think we'll run into issues there either. Sorry did that cover your concerns or did I miss something?

PR #5011, opened 2024-11-20
NJ 51 - dependents attending college (line 12)

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/51

## What was done?
- Add client question that shows all dependents under 22 and asks user to select a series of checkboxes used to determine if they qualify for the dependent attending college exemption
- include count and total exemption amount in PDF
- include count in XML
- update line 13 calculation to include line 12 in its sum

## How to test?
- `zeus_two_deps` for 2 deps, one of whom is over 22 and will not show on question page
- `zeus_many_deps` for 11 deps, one of whom is over 22 and will not show on question page


spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -317,202 +452,143 @@
...
           let(:submission) {
             create :efile_submission, tax_return: nil, data_source: create(
-              :state_file_nj_intake,
-              primary_first_name: "Bert",
-              primary_last_name: "Muppet",
-              primary_middle_initial: "S",
-              spouse_first_name: "Ernie",
-              spouse_last_name: "Muppet",
-              spouse_ssn: "123456789"
+              :state_file_nj_intake, :two_dependents_in_college
```
aloverso, line 523: I rearranged a lot of the tests in here because they were out of order with the PDF. No existing tests have been modified though, and this line 12 test is the only actual change that needs review. Everything else is just rearranging describe blocks

app/forms/state_file/nj_college_dependents_exemption_form.rb
```diff
@@ -0,0 +1,17 @@
+module StateFile
+  class NjCollegeDependentsExemptionForm < QuestionsForm
+    attr_accessor :dependents_attributes
+    delegate :dependents, to: :intake
+
+    set_attributes_for(:intake)
```
powersurge360, line 6: [sand] Without a second argument, does this do anything? I think it will just create a `nil` entry in the scoped attributes hash.

app/forms/state_file/nj_college_dependents_exemption_form.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def save
+      @intake.update!(attributes_to_save)
+    end
+
+    def attributes_to_save
+      base_attrs = attributes_for(:intake)
+      base_attrs.merge({ dependents_attributes: dependents_attributes.to_h })
```
powersurge360, line 14: [sand] Does `base_attrs` contain anything? I think it might be an empty hash. If so, you may be able to just return `{dependent_attributes: dependent_attributes.to_h}`

app/controllers/state_file/questions/nj_college_dependents_exemption_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      def self.show?(intake)
+        intake.dependents.count(&:under_22?).positive?
+      end
+
+      def form_params
+        params
+          .require(:state_file_nj_college_dependents_exemption_form)
+          .permit(
+            form_class.attribute_names +
+              [{ dependents_attributes: [:id, :nj_dependent_attends_accredited_program, :nj_dependent_enrolled_full_time, :nj_dependent_five_months_in_college, :nj_filer_pays_tuition_for_dependent] }])
```
powersurge360, line 15: [dust] P long line. Can we break this up a little to make it easier to scan? Not a major thing though.

powersurge360 (commented):
This comment is basically approved but with a gut check. I think there's a couple of things being used that aren't by the code. I tried it locally on the branch with those changes made and the tests included still seem to pass.

powersurge360 (approved):
Looks good!

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -117,6 +117,17 @@ def hash_for_pdf
                                                  ]))
       end
 
+      # line 12
+      if @xml_document.at("Exemptions DependAttendCollege")
+        count = @xml_document.at("Exemptions DependAttendCollege").text.to_i
+        digits = count.digits
+        answers[:undefined_14] = digits[0]
+        answers[:'x  1000_4'] = count * 1_000
```
mluedke2, line 125: i ran into a similar decision [on my upcoming PR](https://github.com/codeforamerica/vita-min/pull/5014/files#diff-2062992eaee6859ddbb1cc1b5096eeae8efbc7e474aa5ee72ef286e35618a239R129) (i haven't yet merged this work into mine to resolve conflicts), where I couldn't get a value for the PDF from XML because it wasn't in the XML. I ended up getting it from the calculator-- otherwise there are two sources of truth for the calculation

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -117,6 +117,17 @@ def hash_for_pdf
...
+        if digits.length.positive?
+          answers[:Text49] = digits[1]
+        end
```
mluedke2, line 128: [dust|sand] also on my PR I have [this change](https://github.com/codeforamerica/vita-min/pull/5014/files#diff-2062992eaee6859ddbb1cc1b5096eeae8efbc7e474aa5ee72ef286e35618a239R575-R586) to `insert_digits_into_fields` so that it works for non-decimals too. if you want to do it this way and i'll just remember to refactor in my PR that is one way we could do this?

app/models/state_file_dependent.rb
```diff
@@ -55,6 +59,10 @@ class StateFileDependent < ApplicationRecord
   enum eic_disability: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eic_disability
   enum eic_student: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eic_student
   enum id_has_grocery_credit_ineligible_months: { unfilled: 0, yes: 1, no: 2 }, _prefix: :id_has_grocery_credit_ineligible_months
+  enum nj_dependent_attends_accredited_program: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nj_dependent_attends_accredited_program
+  enum nj_dependent_enrolled_full_time: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nj_dependent_enrolled_full_time
+  enum nj_dependent_five_months_in_college: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nj_dependent_five_months_in_college
+  enum nj_filer_pays_tuition_for_dependent: { unfilled: 0, yes: 1, no: 2 }, _prefix: :nj_filer_pays_tuition_for_dependent
```
mluedke2, line 65: [dust] i think since these ultimately end up decided by checkboxes that only have the checked/unchecked state as yes/no, something feels odd to me about using the same enums as we do for radio buttons where the user can say yes, or no, or leave unfilled.

But since the answer implicitly becomes `no` if a user goes through the form without clicking the box, I guess it's ultimately a UX question -- if we're ok not being able to have the same validations as we do on an unfilled radio button, then i guess having three states for this is fine after all. But just wanted to mention because it's the first time I'm encountering this in our UX.

spec/models/state_file/state_file_dependent_spec.rb
```diff
@@ -206,6 +206,53 @@
...
+          intake: intake
+        }
+        let(:intake) { create :state_file_nj_intake }
+
+        it "returns #{test_case[:expected]}" do
+          expect(dependent.nj_qualifies_for_college_exemption?).to eq(test_case[:expected])
+        end
+      end
+    end
+  end
```
mluedke2, line 254: 👍 👍

mluedke2 (dismissed):
looks good to me! only change i would ask for before approving would be resolving the duplicated logic in the pdf filler, but feel free to dismiss this request once that's resolved

app/models/state_file_dependent.rb
mmazanec22: Agree that checkboxes only have filled and unfilled states and we might want to figure out if there's a better enum

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -174,14 +179,18 @@ def calculate_line_9
         number_of_line_9_exemptions * 6_000
       end
 
+      def calculate_line_12
```
mmazanec22, line 182: [dust] Is line 12 left unset above on purpose?

app/lib/pdf_filler/nj1040_pdf.rb
mmazanec22: I like the idea of getting this value from the calculator

app/views/state_file/questions/nj_college_dependents_exemption/edit.html.erb
```diff
@@ -0,0 +1,55 @@
...
+      <button class="reveal__button"><%= t('.tuition_books_helper_heading') %></button>
+      <div class="reveal__content">
+        <p><%= t('.tuition_books_helper_description_html') %></p>
+      </div>
+    </div>
+
+    <%= f.fields_for :dependents do |ff| %>
+      <% dependent = ff.object %>
+      <%  if dependent.under_22? %>
+        <div class="white-group">
```
mmazanec22, line 36: [pebble] we're visually grouping each dependent's fields together for sighted users, and should use a fieldset and legend to group them for assistive tech users

app/views/state_file/questions/nj_college_dependents_exemption/edit.html.erb
```diff
@@ -0,0 +1,55 @@
+<%
+    title = t(".title")
+    content_for :page_title, title
+%>
+
+<% content_for :card do %>
+  <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
+    <h1 class="h2"><%= title %></h1>
+    <p><%= t(".label_1") %></p>
```
mmazanec22, line 9: [dust] Calling this "label" is a little confusing to read. I would expect a label to be something that is used in a label element. Maybe subheader? Description?

app/lib/pdf_filler/nj1040_pdf.rb
aloverso: Let's talk about this one in our eng sync!

app/forms/state_file/nj_college_dependents_exemption_form.rb
aloverso: Good, catch, you're right. Will remove.

app/forms/state_file/nj_college_dependents_exemption_form.rb
aloverso: You're right, this can be removed. Will fix

app/models/state_file_dependent.rb
aloverso: This is a good point, can we discuss what it *should* be?

Also, I'm thinking I might like to resolve this in a follow-up refactor because we should also change the other checkboxes we have that following the same pattern, which I believe would be the `homeowner_*` and `tenant_*` ones

app/lib/pdf_filler/nj1040_pdf.rb
aloverso: Oh neat! I don't have strong feelings but I think using your modification is slightly nicer. How should it work logistically? Are you suggesting I keep it this way and you refactor this code with your new pattern after I merge? Or something else?

mluedke2: either way is fine with me, but since my PR already has the new function introduced, maybe let's just keep that and i'll do the refactor in that PR after this is merged? it will just be replacing the lines here with `digits` with a call to `insert_digits_into_fields`

spec/factories/state_file_nj_intakes.rb
```diff
@@ -315,6 +315,40 @@
...
+            dob: 1.years.ago,
+            nj_dependent_attends_accredited_program: "yes",
+            nj_dependent_enrolled_full_time: "yes",
+            nj_dependent_five_months_in_college: "yes",
+            nj_filer_pays_tuition_for_dependent: "yes"
+          )
+        end
+      end
+    end
+
```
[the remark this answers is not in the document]
mluedke2: [pebble] actually, while merging this into [my work on qualifying children](https://github.com/codeforamerica/vita-min/pull/5014), the fact that these traits involve getting a specific XML+JSON overwrites other needs of setting up a mock... is there a way to do this without importing a specific XML+JSON? 

if you do the below (replacing `2` with `11` for the second trait), then i think that would cycle through that many dependents (if they already exist)

```
intake.dependents.take(2).each_with_index do |dependent, i|
  dependent.update(
    dob: 1.year.ago,
    nj_dependent_attends_accredited_program: "yes",
    nj_dependent_enrolled_full_time: "yes",
    nj_dependent_five_months_in_college: "yes",
    nj_filer_pays_tuition_for_dependent: "yes"
  )
end
```

app/lib/pdf_filler/nj1040_pdf.rb
aloverso: Convo result: keep in this PR for now, Anne or Matt will do a fast follow PR to refactor all instances of this in the PDF to pull from calc instead

spec/factories/state_file_nj_intakes.rb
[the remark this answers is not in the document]
mluedke2: recap of discussion is that it works for me to merge this as-is, since hopefully it just runs into a conflict only once. in my PR i'll resolve the conflict by having two separate tests

app/models/state_file_dependent.rb
mluedke2: recap of discussion is that:
* we'll verify that submitting the form with an unchecked box indeed turns this value to `no`
* keep the word `unfilled` to match other values... but to avoid confusion with an unfilled checkbox we can add comments and group enums for checkboxes vs radio buttons separately in the code

spec/factories/state_file_nj_intakes.rb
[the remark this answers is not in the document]
mluedke2: [pebble] @aloverso  actually still one small edit can still be made -- the `.each_with_index` can be refactored to just `.each`

app/views/state_file/questions/nj_college_dependents_exemption/edit.html.erb
aloverso: From slack: we will add a `fieldset` and replace the `h2` with a `legend` element instead

app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -167,5 +167,18 @@
     </div>
   </div>
 
+  <div id="use-tax" class="white-group">
+    <div class="spacing-below-5">
+      <h2 class="text--body text--bold spacing-below-5">COLLEGE DEPENDNET</h2>
```
mluedke2, line 172: [pebble] typo `DEPENDENT`

aloverso: oops. more importantly this was just test code that i accidentally pushed!!!

mluedke2 (approved):

PR #5014, opened 2024-11-20
Nj 116 qualifying child

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/116

## What was done?
- Added calculator functions to get the two categories of dependents, and added to line 13 and later relevant lines
- The new calculator functions will consider `qualifyingChild` missing from the JSON to be falsey and thus in the "other" category. (In other words, `qualifyingChild` must be both present and `true` to be a qualifying child, and any others are "other")
- No new question or review item-- all happens behind the scenes for the user
- The XML for `NumOfQualiDependChild` and `NumOfOtherDepend` already existed but was meant to be later removed. [This was discussed on the ticket](https://github.com/newjersey/affordability-pm/issues/116#issuecomment-2483766125) and so I removed that for this new logic.
- I had to change the `dependents` function in `direct_file_json_data` from `private` to public in order to be able to call it in the calculator
- Since the exemption amounts are not in the XML (just the number of dependents are), I broke our rule to populate the PDF purely from the XML and get it from the calculator-- also meaning i needed to add `calculated_fields` to the pdf filler
- Edited existing JSONs to get certain results:
  - Zeus One Dep, which is the default, is unchanged and thus has 1 "other" dependent
  - Minimal is unchanged has 0 dependents
  - Zeus Two Deps now has 1 qualifying child and 1 "other" dependent
  - Zeus Many Deps now has 10 qualifying child and 1 "other" dependent
 
## How to test?
- Try with Minimal, Zeus One Dep, Zeus Two Deps, and Zeus Many Deps to see that lines 10, 11, and 13 are correct.

### Minimal

https://github.com/user-attachments/assets/2a351968-9f10-463a-89ef-a35c65caeab0

---

### Zeus One Dep

https://github.com/user-attachments/assets/5008629a-4fad-4f04-a82f-c2dbd6c41728

---

### Zeus Two Deps

https://github.com/user-attachments/assets/954b3d67-7835-4439-8dc5-26946e8544d7

---

### Zeus Many Deps

https://github.com/user-attachments/assets/5d3f59d3-b98e-4ab2-9e4b-3be2016cbcc8


app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -104,7 +104,46 @@ def hash_for_pdf
...
+        count = @xml_document.at("Exemptions DependAttendCollege").text.to_i
+        answers.merge!(
+          insert_digits_into_fields(
+            count,
+            [ "undefined_14", "Text49" ],
+            as_decimal: false
+          )
+        )
+        answers[:'x  1000_4'] = count * 1_000
+      end
```
mluedke2, line 144: i moved this from below in the file to get the lines a little bit more in order (they aren't throughout the file but every little bit helps). And I also `insert_digits_into_fields` now supports `as_decimal: false` so i refactored this as discussed on [the PR for line 12](https://github.com/codeforamerica/vita-min/pull/5011#discussion_r1852421036)

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -516,6 +516,79 @@
...
+        end
+
+        context "1 qualified dependent child" do
+          let(:submission) {
+            create :efile_submission, tax_return: nil, data_source: create(
+              :state_file_nj_intake,
+              :df_data_two_deps,
+            )
+          }
+          it "fills in properly" do
```
mmazanec22, line 540: [dust] this test name could be more descriptive. Maybe "fills in number and exemption value" or something

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -668,5 +701,8 @@ def get_line_65_nj_ctc
       @xml_document.at("Body NJChildTaxCredit")&.text.to_i
     end
 
+    def calculated_fields
+      @calculated_fields ||= @submission.data_source.tax_calculator.calculate
+    end
```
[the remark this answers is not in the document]
mmazanec22: [dust] I wonder if we should somehow indicate in the code that we should not use these when avoidable? Maybe a comment on line 705, or naming this something like `calculated_fields_not_in_xml`?

mmazanec22 (approved):
So much test refactoring! Looks great. Thanks for setting the calculator -> PDF pattern and calling that out

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -1390,7 +1463,7 @@
       )
       }
 
-      it "writes rounded tax amount $7,615.10 based on income $200,000 with 2,000 exemptions 15,000 property tax deduction and 0.0637 tax rate minus 4,042.50 subtraction" do
+      it "writes rounded tax amount $7,519.00 based on income $200,000 with 3,500 exemptions 15,000 property tax deduction and 0.0637 tax rate minus 4,042.50 subtraction" do
```
aloverso, line 1466: [dust] this will cause a merge conflict / test failure with my new Lines 45,49,50,54 commit based on whoever merges first, because that also uses this number. That's fine, just a flag for whoever merges second

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mluedke2: good idea, i'll update

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mluedke2: sorry! just merged.

this does bring up a good point though. i had to redo these numbers which was not all that tedious, just bringing out a manual calculator for a couple minutes to make sure the new test output is correct... but it feels maybe dangerously error prone? like if i had just accepted what the test was outputting and updated the expectation, that would defeat the purpose... i don't have any ideas on how to make this easier though

PR #5015, opened 2024-11-21
FYST-1073 update supporting forms conditionality

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1073

## What was done?
- Added some conditions for attaching MD forms to the return (see story for details)
- Took the opportunity to refactor MD502CR into its own calculator (because I thought I'd be adding a helper method to that calculator—I'm not anymore)
## How to test?
- Tests for 502CR calculator are just copied from where they were before (502 calculator). I added newlines but nothing else.
- `md_return_xml_spec.rb` has gained some tests:
  - tests for when to attach 502CR and 502SU
  - test for new method that finds non-zero lines in a given sub-form using just a line prefix matcher (since we've been very consistent with naming)


app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
```diff
@@ -93,6 +94,13 @@ def financial_transaction
               kwargs: { refund_amount: calculator.refund_or_owed_amount }
             ).document.at("*")
           end
+
+          def form_has_non_zero_amounts(form_prefix, calculated_fields)
+            lines = calculated_fields.keys.select { |line_name| line_name.match?(/^"#{form_prefix}"/) }
+            lines.any? do |line_num|
+              calculated_fields.fetch(line_num) != 0
```
arinchoi03: if the `calculated_fields.fetch(line_num)` returns `nil`, this will also evaluate to true, right? Is that a situation that we don't expect?

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -85,7 +87,7 @@
           it "attaches a 502R" do
             expect(xml.at("Form502R")).to be_present
             expect(instance.pdf_documents).to be_any { |included_document|
-              included_document.pdf == PdfFiller::Md502RPdf
+              included_document.pdf = PdfFiller::Md502RPdf
```
arinchoi03, line 125: looking at the rest of the file, looks like maybe just a typo!
-- 
why is the equal sign removed here? was this just a typo? I feel like the `==` comparison is necessary to satisfy `be_any` matcher, but I can't find good docs on this. I only found:
https://gist.github.com/mcmire/5860047

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
 
       context "502CR" do
-        it "attaches a 502CR" do
-          expect(xml.at("Form502CR")).to be_present
-          expect(instance.pdf_documents).to be_any { |included_documents|
-            included_documents.pdf = PdfFiller::Md502CrPdf
-          }
+        context "L24" do
+          before do
+            allow(instance).to receive(:form_has_non_zero_amounts).with("MD502CR_", anything).and_return false
```
arinchoi03, line 147: i've never seen `anything` as an argument to pass in, that's interesting!

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+
+          context "Form 502 L24 has an amount" do
+            before do
+              allow_any_instance_of(Efile::Md::Md502Calculator).to receive(:calculate_line_24).and_return 50
+            end
+
+            it "attaches a 502CR" do
+              expect(xml.at("Form502CR")).to be_present
+              expect(instance.pdf_documents).to be_any { |included_documents|
+                included_documents.pdf = PdfFiller::Md502CrPdf
```
arinchoi03, line 158: use `==` here

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+
+          context "502CR has non-zero amounts" do
+            before do
+              allow(instance).to receive(:form_has_non_zero_amounts).with("MD502CR_", anything).and_return true
+            end
+
+            it "attaches a 502CR" do
+              expect(xml.at("Form502CR")).to be_present
+              expect(instance.pdf_documents).to be_any { |included_documents|
+                included_documents.pdf = PdfFiller::Md502CrPdf
```
arinchoi03, line 190: same

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+
+          context "Form 502 L13 has an amount" do
+            before do
+              allow_any_instance_of(Efile::Md::Md502Calculator).to receive(:calculate_line_13).and_return 50
+            end
+
+            it "attaches a 502SU" do
+              expect(xml.at("Form502SU")).to be_present
+              expect(instance.pdf_documents).to be_any { |included_documents|
+                included_documents.pdf = PdfFiller::Md502SuPdf
```
arinchoi03, line 224: same

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+              }
+            end
+          end
+
+          context "Form 502 L13 does not have an amount" do
+            before do
+              allow_any_instance_of(Efile::Md::Md502Calculator).to receive(:calculate_line_13).and_return 0
+            end
+
+            it "attaches a 502SU" do
```
arinchoi03, line 234: does not attach

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+              }
+            end
+          end
+
+          context "Form 502 L24 does not have an amount" do
+            before do
+              allow_any_instance_of(Efile::Md::Md502Calculator).to receive(:calculate_line_24).and_return 0
+            end
+
+            it "attaches a 502CR" do
```
arinchoi03, line 168: does not attach

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
```diff
@@ -105,11 +142,134 @@
...
+
+          context "502SU has non-zero amounts" do
+            before do
+              allow(instance).to receive(:form_has_non_zero_amounts).with("MD502_SU_", anything).and_return true
+            end
+
+            it "attaches a 502SU" do
+              expect(xml.at("Form502SU")).to be_present
+              expect(instance.pdf_documents).to be_any { |included_documents|
+                included_documents.pdf = PdfFiller::Md502SuPdf
```
arinchoi03, line 256: use `==`

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
```diff
@@ -41,7 +47,8 @@ def form1099g_builder
           def supported_documents
             calculated_fields = @submission.data_source.tax_calculator.calculate
             has_income_from_taxable_pensions_iras_annuities = calculated_fields.fetch(:MD502_LINE_1D)&.to_i.positive?
-            has_md_su_subtractions = calculated_fields.fetch(:MD502_SU_LINE_1).positive?
+            has_md_su_subtractions = calculated_fields.fetch(:MD502_LINE_13).positive? || form_has_non_zero_amounts("MD502_SU_", calculated_fields)
```
arinchoi03, line 50: can `:MD502_LINE_13` even be positive if `form_has_non_zero_amounts("MD502_SU_", calculated_fields)` is true? Just wondering if only the one check is necessary. Same for below.

arinchoi03 (commented):
A few questions!

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
jenny-heath: oh that 's a really great point. should i:
`calculated_fields.fetch(line_num).present? && calculated_fields.fetch(line_num) != 0`
or
`calculated_fields.fetch(line_num).to_i != 0`
i'm leaning towards the first option just in case there's a value like 0.05 or something stupid

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
jenny-heath: yeah i asked erica and i guess the condition for both 502SU and 502CR is that either:
1) the given line from the 502 has a value
OR
2) the attached form itself has any values
i'm not sure what the odds are that one would be true and the other false, but i did clarify that we want to check both

app/lib/submission_builder/ty2024/states/md/md_return_xml.rb
arinchoi03: first option sounds good due to the case you mentioned 👍

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
jenny-heath: tbh i was just trying to follow the existing pattern which seemed to be `=` for `.to be_any` and `==` for `.not_to be_any`.

i have no idea how `be_any` works though and as pointed out in that gist, it doesn't really grammatically make sense. what i've figured out from trial and error is that both `==` and `=` work for the positive case (`.to be_any`) and only `==` works for the negative case (`.not_to be_any`). so we could either leave it as is or change everything to `==`

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
jenny-heath: i discovered it as part of this!

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
jenny-heath: dang too much copy-paste

spec/lib/submission_builder/ty2024/states/md/md_return_xml_spec.rb
jenny-heath: actually please ignore that i think i maybe understand now why `=` was passing for the positive case? because setting `.pdf` returns true?

anyway changing them all to `==`

app/lib/efile/md/md502_calculator.rb
```diff
@@ -16,6 +16,12 @@ def initialize(year:, intake:, include_source: false)
           lines: @lines,
           intake: @intake
         )
+
+        @md502cr = Efile::Md::Md502crCalculator.new(
+          value_access_tracker: @value_access_tracker,
+          lines: @lines,
+          intake: @intake
+        )
       end
```
tahsinaislam, line 25: sorry was lurking because i was trying to do the same thing on my PR but i think this needs to be Efile::Md::Md502CrCalculator.new

jenny-heath: ah no i named the file `md502cr_calculator` which means the cr is treated like part of the same word as Md502. we kind of had two different conventions going on already (`md502b`, `md502_su`) so i just picked one arbitrarily

PR #5029, opened 2024-11-25
Move initial efile device collection to income review page

embarnard:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1193?atlOrigin=eyJpIjoiODYzMzAzN2ViODEzNDNiNWE3YWM4ZGMzYTBkMTM0YjIiLCJwIjoiaiJ9

## What was done?
- We now redirect away from the data review page, moved efile device id collection to the income review page
- Initially we tried changing the hidden button on the waiting to load page to submit the device id but this proved risky so opted for taking another approach
- The data review page still exists but only for non-prod environments since we needed a page that would hold that edit DF XML button in testing environments 
## How to test?
- Test that AZ, ID, MD, NC and NJ all skip the data review page and that we still have have these fields filled out in the xml auth header:
```
xml.TransmissionDetail do
          xml.InitialCreation do
            xml.IPAddress do
              xml.IPv4AddressTxt 
              xml.IPv6AddressTxt 
            end
            xml.IPTs 
            xml.DeviceId 
            xml.DeviceTypeCd 
          end
end
```


config/locales/en.yml
```diff
@@ -2354,12 +2354,6 @@ en:
           title: Next, create your account with a quick code
       data_review:
         edit:
-          basic_household_info_title: Basic household information
-          complete_description: Complete
-          filing_status_title: Filing status
-          income_and_deductions_title: Income and deductions
-          partially_complete_description: Partially complete
-          state_tax_withheld_title: "%{name} state tax withheld"
           title: Your federal tax return is now transferred. We completed some sections to save you time.
```
anisharamnani, line 2357: [dust] 🤔 since this page is now only going to be visible to those on non-prod environments, it might be helpful to see what in the label. this description is more filer saving.

spec/features/state_file/login_spec.rb
```diff
@@ -52,8 +52,6 @@
     expect(page).to have_text "Code verified! Authentication needed to continue."
     fill_in "Enter your Social Security number or ITIN. For example, 123-45-6789.", with: ssn
     click_on "Continue"
-
-    expect(page).to have_text "Your federal tax return is now transferred."
   end
```
anisharamnani, line 55: [pebble] removing these expectations leaves the test in a weird state. the last line in each of these tests is clicking on "Continue." I think it would be preferable to have an expectation. 

Is there an expectation that can be added here after clicking on continue? Possibly testing that the SSN is being added to the DB?

anisharamnani (approved):
i've got two comments. one pebble & dust. otherwise this looks good! thanks @embarnard!

PR #5030, opened 2024-11-25
Update sinatra, sass-rails and rack

tahsinaislam:
## Is PM acceptance required? 

## What was done?
- To update sinatra to the latest version to address the[ security vulnerability ](https://github.com/codeforamerica/vita-min/security/dependabot/144), we needed to update rack. For the rack update we also needed a sass-rails update to be compatible. 

## How to test?
- 
- Before
- After


app/assets/config/manifest.js
```diff
@@ -1,3 +1,2 @@
 //= link_tree ../images
-//= link_directory ../javascripts .js
-//= link_directory ../stylesheets .css
+//= link application.css
```
arinchoi03, line 2: https://github.com/rails/sprockets/blob/main/UPGRADING.md#manifestjs

Looks like from this documentation, the `javascript` part _is_ ok to remove.

PR #5034, opened 2024-11-26
Preliminary work for FYST-1221: remove old workarounds and update tests

jenny-heath:
Ticket: https://codeforamerica.atlassian.net/browse/FYST-1221

Functional changes:
- removes workarounds we put in place temporarily last year (the AZ one was for one specific client, the redirect from efile thing was because DF had temporarily had this as the redirect URL)

Non-functional changes: 
- the spec reflects a little more of what's in the controller, I hope


squanto (approved):
lgtm 👍

PR #5037, opened 2024-11-26
Adds persona import/export as a CLI

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1216

## What was done?
- This adds both a `personas:export` and a `personas:import` command. No tests included, due to being timeboxed, but if it looks good I'd like to polish it.

## How to test?
The two commands can be run locally to test. Just find an intake with a json blob and xml blob that you want to export as a persona. For example, for a `StateFileIdIntake` you can export either the xml or the json by passing either the `x` or `j` flag as appropriate.

`bin/thor personas:export --xml --state id 30` or `bin/thor personas:export -xs id 30` for a little more brevity.

This will output to stdout which can conveniently be imported with `bin/thor personas:import -s id <some_persona_slug>`. Note that the file extension is not given, it will detect if it is a json blob or an xml blob automatically by checking if it begins with an angle bracket.

The idea is that you can do something like:

`aptible ssh --app vita-app-demo bin/thor personas:export -xs id 30 | bin/thor personas:import -s id penelope_pipsqueak` to get the xml added.

and

`aptible ssh --app vita-app-demo bin/thor personas:export -js id 30 | bin/thor personas:import -s id penelope_pipsqueak`  to get the json persona added.

Note the only difference between the commands is whether an xml or json flag is passed. I was racking my brain trying to think of a way to get both at once but I think it being split between local and remote unfortunately necessitates two runs.

- Before
- After


powersurge360:
Actually, I think what I'd like to do is hand write a `bin/import-demo-persona` command that will take in a state id and intake id and run both of these commands but it will need to be a fast follow on this so I can test it as it will require these scripts to exist on demo.

bin/bundle
```diff
@@ -1,3 +1,114 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
-load Gem.bin_path('bundler', 'bundle')
+# frozen_string_literal: true
```
powersurge360, line 2: I needed to regenerate bundler's binstubs to create binstubs for thor. Not sure why.

powersurge360:
Not strictly part of this PR but for the fast-follow I'm thinking something like this in `scripts/import-demo.rb` which can be called like `scripts/import-demo.rb persona nc 30 penelope_pipsqueak` which will remotely fire the export and pipe it into a local invocation which will put the files in the right place.

```ruby
#!/usr/bin/env ruby

require_relative "../config/environment"

# Really, it's import from aptible demo environment, but abbreviated to make it less
# annoying to type
class ImportDemo < Thor
  desc 'persona', 'Imports a persona from aptible. Uses thor tasks under the hood as defined in the lib/tasks/personas.rb file'
  def persona(state, id, slug)
    say "Importing XML persona", :green
    `aptible ssh --app vita-min-demo bin/thor personas:export -xs #{state} #{id} | bin/thor personas import -s #{state} #{slug}`

    say "Importing JSON persona", :green
    `aptible ssh --app vita-min-demo bin/thor personas:export -js #{state} #{id} | bin/thor personas import -s #{state} #{slug}`
  end
end

ImportDemo.start
```

mpidcock (approved):

PR #5023, opened 2024-11-22
NJ 118 - Lines 45, 49, 50, 54: Calculating Total Tax Due

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/118

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

## What was done?
- Basic arithmetic calculations for lines 45, 49, 50, 54
- Add to XML and PDF

## How to test?
`zeus_many_w2s` will get you some numbers in here. Especially if you add sales/use tax and property tax credit

## Screenshots (for visual changes)
image

image

PR #5040, opened 2024-11-26
NJ 171 - refactor calculator to use lines instead of calculate functions

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/171

Context here: https://github.com/codeforamerica/vita-min/pull/5023

## What was done?
Update nj calculator to follow pattern to reference lines instead of calculate functions

### DIFF
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -116,7 +116,7 @@ module Efile
 
       def calculate_property_tax_deduction
         limit = is_mfs_same_home ? 7_500 : 15_000
-        if calculate_line_40a.nil?
+        if @lines[:NJ1040_LINE_40A]&.value.nil?
           return nil
         end
 


spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -904,6 +912,7 @@ def over_65_birth_year
     }
     it 'returns line 39 times tax rate' do
       allow(instance).to receive(:calculate_line_39).and_return 36_000
+      instance.calculate
```
mluedke2, line 915: [dust] wondering how these were passing without the `instance.calculate`?

mluedke2 (approved):

spec/lib/efile/nj/nj1040_calculator_spec.rb
aloverso: I think that because they were calling the calculate functions, the mock was able to take effect there, but now it needs the calculate method so that if it's referencing that line, that line value is actually up to date to reflect the mock

mluedke2: ah ok makes sense

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -120,18 +120,18 @@ def calculate_property_tax_deduction
           return nil
```
embarnard, line 120: I guess using line_or_zero wouldn't work for line 119 but could use something like `@lines[:NJ1040_LINE_40A]&.value`

PR #4951, opened 2024-11-07
nj 107 - question for estimated tax payments line 57 and shared money test

aloverso:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/109
- https://github.com/newjersey/affordability-pm/issues/147

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

## What was done?
- Question for estimated tax payments, include in review and PDF and XML
- Shared money test refactor for all money fields

## How to test?
- use `minimal` DF
- test both including a value, and that you can move past the screen without providing a value

## Screenshots (for visual changes)
image
image
image
image

PR #5045, opened 2024-11-27
Nj 122 driver license

mluedke2:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/122

## What was done?
- I reuse the existing `state_id` feature used by other states as much as possible, which is why this PR doesn't contain any code for XML input or building the form
- The only overwrites were for the new names for driver license/state ID that our designs introduced

## How to test?
- Test with a single persona and an MFJ persona
- You should always be asked the primary Driver License question, and spouse too if MFJ
- XML and PDF filled with info if provided

https://github.com/user-attachments/assets/b911e437-86f2-465a-b674-e06acfb407f0


spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -1967,5 +1967,73 @@
...
+        it "fills in the ID number" do
+          expect(pdf_fields["Drivers License Number Voluntary Instructions page 44"]).to eq "1"
+          expect(pdf_fields["Text246"]).to eq "2"
+          expect(pdf_fields["Text247"]).to eq "3"
+          expect(pdf_fields["Text248"]).to eq "4"
+          expect(pdf_fields["Text249"]).to eq "5"
+          expect(pdf_fields["Text250"]).to eq "6"
+          expect(pdf_fields["Text251"]).to eq "7"
+          expect(pdf_fields["Text252"]).to eq "8"
+          expect(pdf_fields["Text253"]).to eq "9"
```
aloverso, line 2005: [dust] where are these numbers coming from? I don't see them in the setup but maybe it's the default for those factories?

aloverso (approved):
Looks good! Cool how much of this is already built for us

spec/lib/pdf_filler/nj1040_pdf_spec.rb
mluedke2: yep exactly it is the default for a mock `state_id`

https://github.com/codeforamerica/vita-min/blob/0e91274961bd506da2922e5f5e9368804a61f081/spec/factories/state_ids.rb#L19

app/views/state_file/questions/nj_spouse_state_id/_nj_spouse.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <%= t('.not_required') %>
+        </p>
+      </div>
+    </div>
+  </div>
+<% end %>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= t(".title") %></h1>
+  <p> <%= t(".subtitle") %></p>
```
embarnard, line 19: [dust] extra space?

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -572,6 +572,31 @@ def hash_for_pdf
                                                    "undefined_162",
                                                  ]))
       end
+
```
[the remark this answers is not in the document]
mluedke2: correct -- there's only one PDF spot for an ID number, so we are moving forward with only putting the primary's ID in there, if present. not sure if you have visibility to our discussion on the ticket below but we are double-checking with NJ Taxation so if they do eventually come back saying the spouse can go in there we will need to update.

image


https://github.com/newjersey/affordability-pm/issues/122#issuecomment-2494258914

config/locales/es.yml
```diff
@@ -2959,6 +2959,15 @@ es:
           helper_heading: El lugar donde vivo no está en la lista
           municipality: Municipio
           title: Seleccione el municipio donde residía el 31 de diciembre de %{filing_year}
+      nj_primary_state_id:
```
embarnard, line 2962: [dust] What are New Jersey's practices for translation files? We usually will run our English copy through google translate to get Spanish copy as a stand-in before transfix translators update it (which will include NJ strings) but that might not happen for awhile so good to have a stand-in. I see that some NJ copy has been translated and some has not. Is this an agreed upon practice?

embarnard (approved):
@anisharamnani and I paired reviewed this. Looks good! Just left a few question/comments

config/locales/es.yml
mluedke2: good question - it's our NJ team's practice to just drop in English for now and know that it will be translated later. i asked the same question [here on an earlier PR](https://github.com/codeforamerica/vita-min/pull/4951#discussion_r1833034706)

PR #5049, opened 2024-12-02
NJ 119 - total income tax withheld line 55

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/119

## What was done?
- Add line 55 summing box 17 across w2s

## How to test?
- `zeus 2 w2s` or `zeus many w2s` 

image

image

### DIFF
--- a/spec/lib/pdf_filler/nj1040_pdf_spec.rb
+++ b/spec/lib/pdf_filler/nj1040_pdf_spec.rb
@@ -1615,20 +1615,22 @@ RSpec.describe PdfFiller::Nj1040Pdf do
 
     describe "line 55 - total income tax withheld" do
       context 'when TaxWithheld has a value' do
-        let(:intake) { create(:state_file_nj_intake, :df_data_many_w2s) }
+        before do
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_55).and_return 12_345_678
+        end
 
         it "fills line 55 with sum of income tax withheld" do
           # millions
-          expect(pdf_fields["undefined_1471qerw"]).to eq ""
-          expect(pdf_fields["undefined_114"]).to eq ""
+          expect(pdf_fields["undefined_1471qerw"]).to eq "1"
+          expect(pdf_fields["undefined_114"]).to eq "2"
           # thousands
-          expect(pdf_fields["undefined_143"]).to eq ""
-          expect(pdf_fields["undefined_144"]).to eq ""
-          expect(pdf_fields["undefined_145"]).to eq "2"
+          expect(pdf_fields["undefined_143"]).to eq "3"
+          expect(pdf_fields["undefined_144"]).to eq "4"
+          expect(pdf_fields["undefined_145"]).to eq "5"
           # hundreds
-          expect(pdf_fields["Text153"]).to eq "0"
-          expect(pdf_fields["Text154"]).to eq "0"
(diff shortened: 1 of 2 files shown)


aloverso:
@mluedke2 the `undefined_114` PDF duplicated field name issue is included in here - it will be resolved when we update to the 2024 PDF anyways, so no worries there.

mluedke2 (approved):
all clean to me! as you noted, we'll have to revisit to make sure the non-unique field works on the new pdf

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -1613,6 +1613,50 @@
       end
     end
 
+    describe "line 55 - total income tax withheld" do
+      context 'when TaxWithheld has a value' do
+        let(:intake) { create(:state_file_nj_intake, :df_data_many_w2s) }
```
jenny-heath, line 1618: [pebble] sorry to keep harping on this but i do think it's really useful to mock the calculator method. that way if the fixture changes it doesn't break a bunch of specs beyond just the calculator spec

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -759,6 +759,23 @@
       end
     end
 
+    describe "total income tax withheld - line 55" do
+      context 'when has w2s' do
+        let(:intake) { create(:state_file_nj_intake, :df_data_many_w2s)}
+        it 'sets TaxWithheld to sum of state_income_tax_amount' do
```
jenny-heath, line 765: [pebble] same comment as on pdf spec

jenny-heath (approved):
looks good aside from the one comment about test setup!

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -358,6 +359,12 @@ def calculate_line_54
         [sum, 0].max
       end
 
+      def calculate_line_55
+        return nil if @intake.state_file_w2s.empty?
+
+        @intake.state_file_w2s.inject(0) { |sum, w2| sum + w2.state_income_tax_amount }.round
```
noah-marcus, line 365: [Pebble]: This might be a tax feature I am not aware of but is using `.round` here appropriate? That seems to remove any decimal value. Marking this a pebble because it could have a pretty large effect if not intentional

[Dust]: Also, curious what the `inject(0)` is doing here and if there is a more explicit way to pull out the values needed from the `state_file_w2s` object. (I don't know ruby super well so I could also just be reading this wrong 🙏 )

noah-marcus (commented):
I am missing some context overall, specifically on the `nj1040_pdf.rb` code. While I don't fully understand the __why__ things are happening, it all looks good to me functionally 🙂

Left a comment around the use of rounding. Based off the tests, this seems intentional but wanted to call it out just in case!

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: good callout on the `round`.. we decided as a team to always round since that is what the dept. of taxation ultimately does anyway, so we round and then on the review page inform the user that we rounded.

and `inject()` is the same as `reduce()` in ruby, so if you're used to reduce then all is well. (i also had to look up since i'm used to the `reduce()` syntax)

PR #5067, opened 2024-12-03
use the 'keep it open forever' configs in dev like we do in other non-prod environments

jnf:
## Link to pivotal/JIRA issue
n/a

## What was done?
This adds the overrides to the dev environment so, by default, FYST and GYR are open for business. We've already done this in other non-prod environments, but hadn't brought it to dev. _until now_

## How to test?
Pull the branch, fire up local dev and rejoice that you don't have to use the session toggles to be able to start an intake.


powersurge360 (approved):
Looks good to me!

PR #5082, opened 2024-12-03
Fix 1099-INT XML output crash

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1281

## What was done?
- Make our 1099-int xml output not crash when Payer TIN is missing. 
- Update ID 1099 int intake factory to use new (real) DF return, which does not include Payer TIN
- Fix tests that break due to new factory
## How to test?
- Go through the app with a DF return that includes real, current 1099-INT in the JSON (such as Idaho > Estrada) and confirm that you don't get a crash when viewing our output XML

### DIFF
--- a/app/lib/submission_builder/state1099_int.rb
+++ b/app/lib/submission_builder/state1099_int.rb
@@ -13,7 +13,7 @@ module SubmissionBuilder
             xml.BusinessNameLine1Txt sanitize_for_xml(form1099int.payer.tr('-', ' '), 75)
           end
         end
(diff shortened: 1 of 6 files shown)


arinchoi03 (approved):
confirmed `minOccurs=0` for this field on `StateIndividual/IRSForms/State1099Int.xsd`

mrotondo:
> confirmed `minOccurs=0` for this field on `StateIndividual/IRSForms/State1099Int.xsd`

Thanks for double checking @arinchoi03 !

PR #5083, opened 2024-12-03
FYST-1270 Tax period not included in return header for NC

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1270

## Is PM acceptance required?

## What was done?
- Added a method on intakes on whether return header tax period begin and end date is to be shown
- Fixed d400 pdf so that it is not looking for tax period in the return head but rather from the intake

## How to test?
- Unit test was edited and a new one was added for the return header spec 

- Before
Screenshot 2024-12-03 at 3 55 47 PM

- After
Screenshot 2024-12-03 at 3 56 25 PM


arinchoi03 (approved):
this looks good to me but I wish there was more in the story about why NC doesn't want these dates in their xml?

PR #5085, opened 2024-12-04
Prevent 500 errors on unemployment submit without values

powersurge360:
## Link to pivotal/JIRA issue
- No jira

## What was done?
`params#require` causes failures when a form is submitted with no values sent. Because radio values when unchecked send no values, when a user clicks nothing on the unemployment page and hits submit, the page 500s. This adds a hidden input that will always be sent and removes the 500 issue.

## How to test?
- Navigate through any flow that includes `questions/unemployment`. Click 'Continue' without clicking a radio. See no 500 error.

- Before
image

- After
image


arinchoi03 (commented):
is this preferable to adding some kind of validation?

powersurge360:
It fails at a level before validation can happen. The `require(:symbol_for_model)` breaks because there is nothing in the parameters at all.

arinchoi03 (approved):
Discussed in person that this is due to the `require` that is looking at empty params, ty for explaining @powersurge360 -- this wouldn't happen if we used `permit` instead, but implementing that would be involved & require a bigger refactor. Maybe we could put a TODO here to revisit this workaround?

PR #5090, opened 2024-12-04
Nj 120 refund/balance due amounts

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/120

## Is PM acceptance required?

## What was done?
- I stubbed out several of the lines that get added into Lines 66 and 78, instead of leaving them out.
- I added `set_refund_owed_lines` in the calculator, which then populates the screens that tells the taxpayer about their refund/owed.
- Because the refund screen then asks about direct deposit/check, we now need default values in the intake mock for payment (I took the default values used multiple places [elsewhere](https://github.com/codeforamerica/vita-min/blob/70b369e398c846562eb7c95b373dee4af8d16a89/spec/lib/submission_builder/ty2024/states/id/id_return_xml_spec.rb#L5) in the repo)
- The one place I just left a comment is for calculating Line 80, where once we implement the inputs to Line 78 we will have to find some way to keep that value from going negative.
 
## How to test?
- Minimal profile, with $1000 already paid should mean they get a refund and see the first set of screenshots below
- Zeus many w2s should mean they owe a balance and see the second set of screenshots below

### Minimal with $1000 already paid

Refund screen:
image

---

XML
image

---

PDF
image
image

### Zeus Many W2s

Balance Owed Screen:

image

---

XML
image

---

PDF
image
image


aloverso (approved):
Looks great, no issues

PR #5101, opened 2024-12-04
Upcase suffixes for XML to pass schema validation

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1316

## What was done?
- suffixes are pulled solely from json data; although this information comes from DF and we expect it to be in capitalized format, there is no guarantee of what case it'll be in so it'll be safer to be more defensive here
- I asked in jira comments whether we want to preserve the data and Tiffany said it doesn't matter in DB -- so instead of mutating data from DirectFile during synchronization, I'm going to just upcase when we need to (at the XML output)
- This impacts all states, not just NC
- I also saw we fill in the `NameSuffix` field for `account_holder_suffix` & `joint_account_holder_suffix` information (see `app/lib/submission_builder/ty2024/states/md/documents/md502.rb`), but these values are filled in via intake flow (our screens) and it's a dropdown field that will not allow a lower case suffix. Should I also upcase these although there's likely no way these will be lowercase? My opinion is no -- although I can be persuaded to just preserve some uniformity in the suffix handling at xml level.
- Updated NJ1040 b/c the suffixes from json get used there too. Should ping NJ folk to make sure they can see it.

## How to test?
  - Unit tests

No bundle failure observed with `nc_zeus_mfj_deps` fixture we were seeing before.
Screenshot 2024-12-05 at 12 29 11 AM


spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -61,15 +61,20 @@
...
         let(:spouse_middle_initial) { nil }
         let(:spouse_last_name) { nil }
+        let(:spouse_suffix) { nil }
+
         before do
           intake.direct_file_data.primary_ssn = primary_ssn
+          intake.primary_suffix = primary_suffix
           intake.direct_file_data.spouse_ssn = spouse_ssn
           intake.direct_file_data.phone_number = "5551231234"
+          intake.spouse_suffix = spouse_suffix
```
arinchoi03, line 77: `direct_file_data` does not contain `suffix` information

spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -61,15 +61,20 @@
         let(:primary_first_name) { "Prim" }
         let(:primary_middle_initial) { "W" }
         let(:primary_last_name) { "Filerton" }
+        let(:primary_suffix) { "jr" }
```
anisharamnani, line 64: after line 88, does it make to add this expectation?
```
expect(doc.at('Filer Primary TaxpayerName NameSuffix').content).to eq("JR")
```
i see the test data is set up but there's no expectation.

anisharamnani: if i'm missing something please let me know, don't mean to be nit picky. i just didn't see the expectation.

spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -106,13 +111,15 @@
...
                 intake.direct_file_json_data.primary_filer.last_name = nil
+                intake.primary_suffix = primary_suffix
               end
 
               if intake.direct_file_json_data&.spouse_filer.present?
                 intake.direct_file_json_data.spouse_filer.dob = nil
                 intake.direct_file_json_data.spouse_filer.first_name = nil
                 intake.direct_file_json_data.spouse_filer.middle_initial = nil
                 intake.direct_file_json_data.spouse_filer.first_name = nil
+                intake.spouse_suffix = nil
```
anisharamnani, line 122: & same here does this expectation need to be added, since the test data is being set up: 
```
expect(doc.at('Filer Secondary TaxpayerName LastName')).not_to be_present
```

anisharamnani (approved):
i have some fairly minor comments, but overall this looks good! thanks @arinchoi03!

spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -153,6 +161,21 @@
...
+          let(:spouse_birth_date) { 42.years.ago }
+          let(:spouse_ssn) { "200000030" }
+          let(:spouse_first_name) { "Sec" }
+          let(:spouse_middle_initial) { "Z" }
+          let(:spouse_last_name) { "Filerton" }
+          let(:spouse_suffix) { "sr" }
+
+          it "should upcase suffixes" do
+            expect(doc.at("Filer Primary TaxpayerName NameSuffix").text).to eq("JR")
+            expect(doc.at("Filer Secondary TaxpayerName NameSuffix").text).to eq("SR")
```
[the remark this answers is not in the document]
arinchoi03: i left this there b/c I think this is a weird edge-case (we hopefully shouldn't receive lower-cased suffixes from DF...hopefully?) and i dont want to make that the "norm" in our tests

PR #5118, opened 2024-12-05
FYST-1365 Fix md address

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1365

## Is PM acceptance required? 

## What was done?
- Add in methods similar to ny ones to cut out the apartment from the df mailing address if the address is longer than 35 characters 

## How to test?
- Added unit test for MD502 xml spec

- Before
Screenshot 2024-12-05 at 12 14 51 PM

- After
Screenshot 2024-12-05 at 12 30 52 PM

### DIFF
--- a/app/lib/submission_builder/document.rb
+++ b/app/lib/submission_builder/document.rb
@@ -114,14 +114,14 @@ module SubmissionBuilder
       end
     end
 
-    def process_address(xml)
+    def extract_apartment_from_mailing_street(xml)
       return unless @submission.data_source.direct_file_data.mailing_street.present?
 
       mailing_street = sanitize_for_xml(@submission.data_source.direct_file_data.mailing_street)
       key_position = mailing_street.index(ADDRESS_ABBREV_REGEX)
 
       if key_position
-        truncated_mailing_street = mailing_street[0, key_position].rstrip
+        truncated_mailing_street = mailing_street[0...key_position].rstrip
         excess_characters = mailing_street[key_position..].lstrip
 
         xml.AddressLine1Txt sanitize_for_xml(truncated_mailing_street, 35)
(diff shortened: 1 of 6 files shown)


squanto (approved):

app/lib/submission_builder/document.rb
```diff
@@ -114,6 +114,34 @@ def email_from_intake_or_df
...
+        else
+          xml.AddressLine2Txt excess_characters
+        end
+      else
+        xml.AddressLine1Txt sanitize_for_xml(mailing_street, 35)
+        xml.AddressLine2Txt sanitize_for_xml(@submission.data_source.direct_file_data.mailing_apartment, 35) if @submission.data_source.direct_file_data.mailing_apartment.present?
+      end
+    end
+
     def process_mailing_street(xml)
```
mrotondo, line 146: a comment here noting that these are NY-only would help clarify the difference from the new code above

app/lib/submission_builder/document.rb
```diff
@@ -114,6 +114,34 @@ def email_from_intake_or_df
...
     end
 
+    def process_md_mailing_street(xml)
+      return unless @submission.data_source.direct_file_data.mailing_street.present?
+
+      mailing_street = sanitize_for_xml(@submission.data_source.direct_file_data.mailing_street)
+      key_position = mailing_street.index(ADDRESS_ABBREV_REGEX)
+
+      if key_position
+        truncated_mailing_street = mailing_street[0, key_position].rstrip
```
mrotondo, line 124: [dust] minor style point, but for consistency with the below line (`[key_position..]`), can you use `[0...key_position]` here to denote the range between 0 inclusive and key_position non-inclusive? ([Range literal docs](https://ruby-doc.org/3.3.6/syntax/literals_rdoc.html#label-Range+Literals) because I always forget which one is which)

mrotondo (approved):
couple requested changes, but looks good once those are made!

PR #5126, opened 2024-12-05
remove i18n string that is basically a dupe

jenny-heath:
now that we're using transifex we should avoid having people put in extra labor, for example having to translate 2 strings that are the same but for one link


app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
```diff
@@ -1,7 +1,7 @@
-<% @main_heading = t('.main_heading', filing_year: current_tax_year) %>
+<% @main_heading_no_html = strip_tags(t('.main_heading_html')) %>
 <% show_yes_no = @contribution_count == 0  %>
 
-<% content_for :page_title, @main_heading %>
+<% content_for :page_title, @main_heading_no_html %>
```
powersurge360, line 4: How does this look in the browser tab?

squanto (approved):
🧼

app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
jenny-heath: like what does it look like without the html or with it?

PR #5133, opened 2024-12-06
allow back link to to previous navigation actions in the current controller

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1296

## What was done?
- I added a branch to `StateFile::QuestionsController.prev_path` so that in case the current controller has multiple `navigation_actions` (think 1099-Gs, AZ public school contributions and AZ qualifying organization contributions) our app's back link will go to the first action in that list when clicked, instead of the previous behavior of jumping to the previous controller.
## How to test?
- Go through every possible flow for the unemployment, public school and qualifying org contribution pages and use the back link from every one of them to verify that it does a sensible thing.
- The only behavior with this change that I believe is slightly worse than the previous behavior is when you go to an empty index page (e.g. unemployment), submit without selecting a radio button and are shown a validation error, and then try to use the back link - it now shows you the same page again but without the validation error, because you're technically on the update action after submitting the blank radio buttons and the back link is bringing you back to the index action.


app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -56,8 +57,30 @@ def prev_step
...
+
+        # look up action corresponding to the submitted form if we hit a validation error on submission and are re-rendering new/edit
+        replacements = { update: :edit, create: :new }
+        current_action = replacements[current_action] if replacements.key? current_action
+
+        action_index = self.class.navigation_actions.index(current_action)
+
+        case action_index
+        when 0 then nil
+        when 1.. then self.class.navigation_actions[action_index - 1]
```
arinchoi03, line 73: I think I'm ok with this as-is and as this is built upon current patterns of how we use `navigation_actions` in our app --

but just thoughts here...
wondering if this makes sense for all sorts of `navigation_actions` array, for example, if I made a `navigation_actions` that looks like `[:edit, :delete, :index]`, this flow wouldn't really work, right? 
Is more of what we're looking for that if we have more than one `navigation_actions` defined, we want to return to the `:index` action as a `prev_action`?

```
when 1.. then :index
```
or
```
when 1.. then self.class.navigation_actions[0]
```
with the understanding that if we're defining multiple actions on a controller and an index is specified, the first should always be an index?

spec/controllers/state_file/questions/questions_controller_spec.rb
```diff
@@ -1,5 +1,105 @@
...
+end
+
+class QuestionFourController < StateFile::Questions::QuestionsController; end
 
+class QuestionNavigation < Navigation::StateFileBaseQuestionNavigation
+  include ControllerNavigation
+  FLOW = [QuestionOneController, QuestionTwoController, QuestionThreeController, QuestionFourController]
 end
+
+RSpec.describe StateFile::Questions::QuestionsController do
```
arinchoi03, line 20: i really appreciate this testing set-up!

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
```diff
@@ -213,6 +213,9 @@
       get :index
       expect(response.body).not_to include(I18n.t('state_file.questions.az_qualifying_organization_contributions.index.maximum_records'))
 
+      # remove the built-but-invalid az321_contribution relation created when #index renders the :new template
+      subject.current_intake.reload
```
arinchoi03, line 217: is this not a problem in the real implementation? does the `current_intake` reload on the create action in the controller?

arinchoi03 (approved):
more questions than real suggestions to change anything!

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -56,8 +57,30 @@ def prev_step
...
+        else self.class.navigation_actions[0]
```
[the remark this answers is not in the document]
mrotondo: yes, this handles situations where `current_action` is not in the `navigation_actions` list (or in the `replacements` list above). for example, if you hit `edit` on a public school contribution, the `edit` action isn't in the `navigation_actions` list but you should still go back to `index` (the first item in the list) before going back to the previous controller

app/controllers/state_file/questions/questions_controller.rb
mrotondo: Good call - you're right, this code assumes both that:
- the `navigation_actions` will be a list that should be visited in order, and
- the first item in the `navigation_actions` should be some kind of "default" page that is visited before every other action in the list, and should be returned to if we can't figure out where else to send you.

That second assumption is already in the code in `StateFile::Questions::QuestionsController#path_for_step` where we say `options = { action: step.navigation_actions.first }`. 

However, the first assumption is new and really has no basis - it just happens to work right now because all the places where we've overridden `navigation_actions`, we've only included 2 actions. I'll change this code to jump back to the first action in the list if the current action is anything other than the first action, and to the previous controller if the current_action is the first action in the list. Does that sound good to you?

spec/controllers/state_file/questions/az_qualifying_organization_contributions_controller_spec.rb
mrotondo: I spent a while figuring out what was going on here :(

As I understand it, this problem will never pop up in real usage. The issue is that the `:index`/`:new` actions create blank contributions for the form to use when rendering. However, because we call multiple actions within a single test, the controller (and its references) are never garbage collected, and so when `create` is called, the blank (and therefore invalid) contribution from `index` is still present in the `current_intake`'s `az321_contributions` relation. This prevents the subsequent contributions from being saved, because the intake is invalid by virtue of the blank unsaved contribution in its relation. In production, new instances of the controller would be created for each request so this could not occur. The `reload` simulates a new controller instance being created for the subsequent requests, as would happen in a real deployment.

I've confirmed this understanding by calling `get :new` before making the create calls in the test code without the other changes in this PR and reproducing the failure I saw when I added the blank contribution creation to `index`.

arinchoi03 (approved):
🚀 yay!

PR #5139, opened 2024-12-06
[FYST-1392] Update state information service

squanto:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1392

## What was done?
- Update `StateInformationService` values for 2024


app/services/state_file/state_information_service.rb
```diff
@@ -83,49 +83,55 @@ def state_intake_class_names
         w2_include_local_income_boxes: false
       },
       id: {
+        mail_voucher_address: "Idaho State Tax Commission<br/>" \
```
embarnard, line 86: I see that is getting called on the return status page, we never show the addresses in Spanish right? I don't think we do but might want to check in with product on that if they haven't already said something

embarnard (approved):
Looks good! Just had a question but if it seems good to you feel free to proceed

app/lib/pdfs/ID-VP.pdf
mpidcock, line 1: Since this pdf is used for a public link, we need to store it under `public/pdfs`

app/services/state_file/state_information_service.rb
```diff
@@ -83,49 +83,55 @@ def state_intake_class_names
...
         tax_payment_info_url: "https://www.marylandtaxes.gov/individual/individual-payments.php",
         tax_payment_url: "Marylandtaxes.gov",
-        tax_refund_url: "",
+        tax_refund_url: "https://interactive.marylandtaxes.gov/INDIV/refundstatus/home.aspx",
         department_of_taxation: "Comptroller of Maryland",
         timezone: 'America/New_York',
         vita_link: "",
-        voucher_form_name: "",
+        voucher_form_name: "Form PV",
         voucher_path: "",
```
mpidcock, line 135: We'll also need to add the MD voucher pdf and set the path here

app/services/state_file/state_information_service.rb
```diff
@@ -144,11 +150,11 @@ def state_intake_class_names
...
         submission_builder_class: SubmissionBuilder::Ty2024::States::Nc::NcReturnXml,
-        survey_link: "",
-        tax_payment_info_url: "",
-        tax_payment_url: "",
+        survey_link: "https://codeforamerica.co1.qualtrics.com/jfe/form/SV_1MM0vBfZ5N2OMLA",
+        tax_payment_info_url: "https://www.ncdor.gov/file-pay/pay-individual-income-tax",
+        tax_payment_url: "https://www.ncdor.gov/file-pay/pay-individual-income-tax",
         tax_refund_url: "https://eservices.dor.nc.gov/wheresmyrefund/SelectionServlet",
-        department_of_taxation: "",
+        department_of_taxation: "https://www.ncdor.gov/",
```
mpidcock, line 157: This was a mistake in the jira ticket - the value should be the name of the Department (North Carolina Department of Revenue)

mpidcock (changes requested):
Looks like there were some details missing from the jira ticket so some minor fixes need to be made

app/services/state_file/state_information_service.rb
mpidcock: We can follow up on this later

mpidcock (approved):

PR #5140, opened 2024-12-06
Updates send message for email support

powersurge360:
## Link to pivotal/JIRA issue
- N/A
## Is PM acceptance required? (delete one)
- No - merge after code review approval

**Reminder**: merge main into this branch and get green tests before merging to main
## What was done?
- I bounced off of the [ticket](https://codeforamerica.atlassian.net/browse/FYST-1345) for adding a new message but I did, as part of that, get emails working with `scripts/send_message.rb`. Figured I'd extract that to get merged separately.

## How to test?
- The harness has instructions built into it.

## Screenshots (for visual changes)
- Before
- After

PR #5146, opened 2024-12-06
Fyst 1345/correct terminal sms and email

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1345

## What was done?
- Adds a new terminal rejected status message
- Adds a link between rejected and cancelled for rejected submissions

## How to test?
- Unit tests provided
- See ticket for recreation

- Before
- After


app/jobs/state_file/after_transition_tasks_for_rejected_return_job.rb
```diff
@@ -6,7 +6,11 @@ def perform(submission, transition)
       Efile::SubmissionErrorParser.persist_errors(transition)
 
       if transition.efile_errors.any?
-        submission.transition_to!(:cancelled) if transition.efile_errors.any?(&:auto_cancel)
+        if transition.efile_errors.any?(&:auto_cancel)
+          submission.transition_to!(:notified_of_rejection)
+          submission.transition_to!(:cancelled)
+        end
```
jenny-heath, line 12: remind me whether you ran this by Tiffany? i think it makes sense to me but the notified of rejection state is still slightly confusing to me so we should make sure this is sanctioned by product

embarnard (approved):
I'm a little rusty on efile state machine things but I think this makes sense to me!

config/initializers/letter_opener.rb
jenny-heath, line 1: what's dis

embarnard: I think thats the thing that opens messages in a file when it gets sent locally

spec/services/state_file/after_transition_messaging_service_spec.rb
```diff
@@ -101,6 +101,19 @@
     end
   end
 
+  describe "#send_efile_submission_terminal_rejected_message" do
+    let(:message) { StateFile::AutomatedMessage::TerminalRejected }
+
+    it "sends the accepted refund" do
+      expect do
+        messaging_service.send_efile_submission_terminal_rejected_message
+      end.to change(StateFileNotificationEmail, :count).by(1)
```
jenny-heath, line 110: should we have a `StateFileNotificationEmail` equivalent for texts?

jenny-heath: (i think the answer is probably yes)

config/initializers/letter_opener.rb
powersurge360: Should be removed though. Originally this had some stuff with letter opener in it but when I parked it due to it seemly being blocked, I ported those changes [to another branch](https://github.com/codeforamerica/vita-min/pull/5140). I will remove, thanks for noticing.

spec/services/state_file/after_transition_messaging_service_spec.rb
powersurge360: If you mean for the test case, sure I can add. If you mean the `AutomatedMessage` hierarchy may need to be refactored I'm not so sure. I'd need to think about it. I'll make an adjustment based on the former assumption though and let me know if you want to talk about moving around automated messages.

app/jobs/state_file/after_transition_tasks_for_rejected_return_job.rb
powersurge360: Tiffany approved of this! Just demo'd on tuple.

spec/factories/state_file_az_intakes.rb
```diff
@@ -102,6 +102,7 @@
     primary_last_name { "Onian" }
     primary_birth_date { Date.new((MultiTenantService.statefile.current_tax_year - 65), 12, 1) }
     state_file_analytics { StateFileAnalytics.create }
+    # phone_number: "+15551115511" # number taken from specs for Phony
```
powersurge360, line 105: I had to dig through `Phony` specs to find a valid number for testing since `555-555-5555` was rejected. I put it where people might expect to find it but I left it in a comment in case tests were depending on phone numbers being nil by default.

spec/services/state_file/after_transition_messaging_service_spec.rb
jenny-heath: copying over my slack message because i realized it's good to have the record in github:

"so sorry my comment about this wasn't clearer but what i meant about a StateFileNotificationEmail equivalent was: the record we keep for an outgoing email in FYST is StateFileNotificationEmail and since we didn't send texts last year we didn't make one for texts. but i think it would make sense to make a minimal model called StateFileNotificationTextMessage (based on the email one) and create a new one when we send text notifications. and if that feels like it's adding too much scope for this story i think a follow-up would be fine"

PR #5148, opened 2024-12-07
add buttons for transitioning to rejected with auto wait and auto cancel errors

jenny-heath:
i found that while working on the return status content updates that it was annoying to try to get my local submissions into the right state so i added these buttons but maybe we don't like them, you tell me

image

### DIFF
--- a/app/controllers/hub/state_file/efile_submissions_controller.rb
+++ b/app/controllers/hub/state_file/efile_submissions_controller.rb
@@ -81,7 +81,13 @@ module Hub
         authorize! :update, @efile_submission
         metadata = { initiated_by_id: current_user.id }
         if to_state == "rejected"
-          metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}").last.code
+          if params[:auto_cancel]
+            metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}", auto_cancel: true).last.code
+          elsif params[:auto_wait]
+            metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}", auto_wait: true).last.code
+          else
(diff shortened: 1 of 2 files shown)


app/controllers/hub/state_file/efile_submissions_controller.rb
```diff
@@ -81,7 +81,13 @@ def transition_to
...
         metadata = { initiated_by_id: current_user.id }
         if to_state == "rejected"
-          metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}").last.code
+          if params[:auto_cancel]
+            metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}", auto_cancel: true).last.code
+          elsif params[:auto_wait]
+            metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}", auto_wait: true).last.code
+          else
+            metadata[:error_code] = EfileError.where(service_type: "state_file_#{@efile_submission.data_source.state_code}", auto_cancel: false, auto_wait: false).last.code
+          end
```
arinchoi03, line 90: I don't really understand `auto_cancel` and `auto_wait`, but I assume these are different from the "regular rejection" which means these don't cancel and these don't wait -- they just _sit_ in rejected state/limbo and need manual handling?

arinchoi03 (approved):
the changes look larger b/c of html elements moving about but I'm assuming the only relevant changes are the two new "auto_" buttons? If so, looks great 💯

app/controllers/hub/state_file/efile_submissions_controller.rb
jenny-heath: yeah basically `auto_cancel` = we know we cannot help the client and they need to file another way and `auto_wait` = the client needs to fix something so we can give them the go-ahead to do that. the goal is for the majority of our errors to go into one of these categories because everything else needs to be investigated by our team

PR #5150, opened 2024-12-09
NJ 66 - add PDF for additional dependents

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/66

## What was done?
- Add nj_additional_dependents PDF filler
- Include additional dependents in PDF if greater than 4

## How to test?
- Use `zeus_many_deps` to test more than 4 dependents
- Use any other persona to test fewer than 4

image


DrewProebstel (approved):
Looks great!

embarnard (approved):

app/lib/pdfs/nj_additional_dependents.pdf
jachan, line 1: [PEBBLE] When tabbing through the PDF fields using Adobe Acrobat, the ordering after the first row of the Dependents Table is column by column, not row by row, as one would expect.

i.e. Name_Row1 -> SSN_Row1 -> BirthYear_Row1 -> HealthInsurance_Row1 -> HealthInsurance2...HealthInsuranceX -> Name2....NameX -> SSN2...SSNX.

This suggests something wrong with the underlying structure of the PDF, which could be an accessibility issue: https://www.w3.org/TR/WCAG20-TECHS/PDF3.html.

jachan, line 1: [DUST] Great job on giving sane names to these fields though, a breath of fresh air after some of the NJ1040 field names. Only question here is why not 0-index the rows?

app/lib/pdf_filler/nj_additional_dependents_pdf.rb
```diff
@@ -0,0 +1,57 @@
...
+    def initialize(submission)
+      @submission = submission
+
+      builder = StateFile::StateInformationService.submission_builder_class(:nj)
+      @xml_document = builder.new(submission).document
+    end
+
+    def hash_for_pdf
+      answers = {
+        "Names(s) as shown on NJ 1040": get_name(@xml_document),
```
jachan, line 19: [DUST] Great use of nj_pdf_helper - glad you put that helper function there.

jachan (commented):
Great code style as always, I flagged something about the structure of the nj_additional_dependents PDF that might make it hard to navigate with a keyboard/screenreader and linked a resource that might help.

app/lib/pdf_filler/nj_additional_dependents_pdf.rb
```diff
@@ -0,0 +1,57 @@
...
+        }
+      end
+    end
+
+    def get_taxpayer_ssn
+      @xml_document.at("ReturnHeaderState Filer Primary TaxpayerSSN")&.text
+    end
+
+
+    def fill_dependents
```
jachan, line 47: [DUST] Validating that "No Health Insurance" is meant to always be unchecked on this addendum?

jachan: If so, can we add that assert to `spec/lib/pdf_filler/nj_additional_dependents_pdf_spec.rb`?

app/lib/pdfs/nj_additional_dependents.pdf
aloverso: I was thinking that field names are meant to be human-readable, not engineer-readable!

jachan: Point taken, that makes sense!

aloverso: Ugh, great callout, I know this happened because of the order in which I added the fillable fields to the PDF to make the manual process faster. Will have to just do a some manual work to fix this, but appreciate the callout.

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -164,5 +164,49 @@
...
       end
     end
+
+    describe "additional dependents PDF" do
+      context "when there are more than 4 dependents" do
+        let(:intake) { create(:state_file_nj_intake, :df_data_minimal) }
+        let(:nj_return) { described_class.new(submission) }
+
+        before do
+          create :state_file_dependent, intake: intake
```
mmazanec22, line 174: [dust] This could be simplified to `5.times { create :state_file_dependent, intake: intake } ` here and below. Having the actual number specified might improve readability in this case

mmazanec22 (approved):

app/lib/pdfs/nj_additional_dependents.pdf
jachan, line 1: [PEBBLE] Still one lingering issue with a single field (sorry). 

SSN_Row13 is at the end of everything instead of in-between Name_Row13 and BirthYear_Row13.

So right now the order is Name_Row16, SSN_Row16, BirthYear_Row16, and then SSN_Row13 before looping back to the beginning.

jachan (approved):
PDF looks great!

PR #5157, opened 2024-12-09
FYST-1340 state specific copy on post submit screens

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1340

## What was done?
- A bunch of content updates to the return status page
## How to test?
- I started to write some controller specs to check that the right content shows up but it was starting to feel little time consuming to cover all cases that way and I'm not sure what our philosophy on testing for content is? Do we believe that tests should check all the copy or is it enough to have acceptance testers just look at it?
there's a little variation between states but not enough to get screenshots of all of them

pending:
Screenshot 2024-12-06 at 11 35 27 AM

accepted (owe)
Screenshot 2024-12-03 at 2 24 07 PM

accepted (refund)
Screenshot 2024-12-03 at 11 31 28 AM

rejected
Screenshot 2024-12-06 at 11 40 22 AM


app/lib/default_error_messages.rb
```diff
@@ -50,6 +50,14 @@ def self.generate!(service_type:)
...
         service_type: service_type_name
       )
+      EfileError.find_or_create_by!(
+        code: "AN UNFIXABLE ERROR",
+        message: "Something about your return disqualifies you from filing",
+        source: :intake,
+        expose: true,
+        auto_cancel: true,
+        service_type: service_type_name
+      )
```
jenny-heath, line 60: meant for this to be in my other PR actually but i'm pretty sure these errors are used just in the seeder and i wanted access to an auto_cancel error locally without having to manually create one every time

app/views/state_file/questions/return_status/_pending.html.erb
```diff
@@ -9,15 +9,31 @@
...
+  <% elsif current_intake.email_notification_opt_in_yes? %>
+    <%= t('state_file.questions.submission_confirmation.edit.email_update') %>
+  <% elsif current_intake.sms_notification_opt_in_yes? %>
+    <%= t('state_file.questions.submission_confirmation.edit.text_update') %>
+  <% end %>
+</p>
+
+<p>
+  <%= t('.check_spam') %>
+</p>
```
arinchoi03, line 27: does this `.check_spam` message make sense if they haven't opted in to any communications?

app/lib/default_error_messages.rb
arinchoi03: [q] I don't fully understand how you used this in this PR so would love to chat about it 🙏

arinchoi03 (approved):
In regards to the testing piece of it, is there a way to force the submission into certain submission states & check for snippets of text in `complete_intake_spec`? 
If it becomes too extensive, maybe just picking one state and testing at least the shared messaging across all states for different return statuses? I guess not testing the content at all makes me pause in case for some reason any of those return screens break/dont load we won't know until alerted by the user.

There's "pending" state specs in `spec/controllers/state_file/questions/return_status_controller_spec.rb` and maybe we can just add to them too.

app/views/state_file/questions/return_status/_pending.html.erb
arinchoi03: confirmed with Anu/Madison that this should only show if they signed up with email
https://codeforamerica.atlassian.net/browse/FYST-1340?focusedCommentId=13932

app/lib/default_error_messages.rb
arinchoi03: discussed in person:
> Arin: how do you link this error to your local return?
> Jenny: oh so that button i added for "rejected autocancel" will find the first auto-cancel error so it end up being this one because it's the only one i have

arinchoi03 (approved):
looks good to me! phew

PR #5158, opened 2024-12-10
don't output parts B, M or AA of MD 502CR in case deduction method is N

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1386

## What was done?
- Changed the MD502CR XML generator to only output parts B, M and AA if the "deduction method" we calculated as part of the MD502 was "S" and not "N". Added tests to validate this change to the XML. Since the PDF gets all its values from the XML, the same change should take effect there.
## How to test?
- Go through the MD flow such that your deduction method is calculated as "N". Verify that parts B, M and AA of the 502CR are missing from both the XML & PDF. Do the same thing with an "S" deduction method and verify that they are present.


anisharamnani (approved):
thanks @mrotondo, no notes!

PR #5162, opened 2024-12-10
NJ 195 - fix bundle fail error by validate w2 field present

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/195

## What was done?
When creating NJ W2, check field present before changing field name

## How to test?
Use `IRS_test_1` persona which matches the failure test case


mrotondo (approved):

app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
```diff
@@ -12,10 +12,10 @@ def document
 
               xml_node.at("EmployerNameControlTxt").name = "EmployerNameControl"
               xml_node.at("ControlNum").name = "ControlNumber" if xml_node.at("ControlNum").present?
-              xml_node.at("EmployeeNm").name = "EmployeeName"
+              xml_node.at("EmployeeNm").name = "EmployeeName" if xml_node.at("EmployeeNm").present?
               xml_node.at("PriorUSERRAContributionYr").name = "PriorYearUserraContribution" if xml_node.at("PriorUSERRAContributionYr").present?
               xml_node.at("OtherDeductionsBenefitsGrp").name = "OtherDeductsBenefits" if xml_node.at("OtherDeductionsBenefitsGrp").present?
-              xml_node.at("EmployerStateIdNum").name = "EmployersStateIdNumber"
+              xml_node.at("EmployerStateIdNum").name = "EmployersStateIdNumber" if xml_node.at("EmployerStateIdNum").present?
```
mluedke2, line 18: [dust] this makes me wonder -- are there any XML attributes that we actually do want to fail on (in a caught way of course, not just 500) because it would mean something is wrong in the input?

mluedke2 (approved):

app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
mluedke2: that makes sense to me 👍

PR #5168, opened 2024-12-10
FYST-1393 Change notifications preference to show provided email

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1393

## Is PM acceptance required? 

## What was done?
- Added in a condition to show the email set previously 
- Instead of showing a field to change the email address for notification, there is now statement showing the email that was previously inputted in the code verification section

## How to test?
- This was tests in the flow locally 

- Before
Screenshot 2024-12-10 at 3 24 55 PM

- After
Screenshot 2024-12-10 at 3 23 45 PM


app/forms/state_file/notification_preferences_form.rb
```diff
@@ -4,16 +4,23 @@ class NotificationPreferencesForm < QuestionsForm
...
-    validates :email_address, presence: true, if: -> { attributes_for(:intake)[:email_notification_opt_in] == "yes" }
+    validates :email_address, 'valid_email_2/email': true, if: -> { should_validate_email? }
+    validates :email_address, presence: true, if: -> { should_validate_email? }
     validate :at_least_one_selected
 
     def save
-      @intake.update(attributes_for(:intake))
+      email_attributes = should_validate_email? ? {} : {email_address: @intake.email_address}
+
+      @intake.update(attributes_for(:intake).merge(email_attributes))
```
arinchoi03, line 14: [q] do we need to update the intake with the existing `email_address` if it's already on the intake?

app/forms/state_file/notification_preferences_form.rb
```diff
@@ -4,16 +4,23 @@ class NotificationPreferencesForm < QuestionsForm
...
     end
 
     private
 
+    def should_validate_email?
+      attributes_for(:intake)[:email_notification_opt_in] == "yes" &&
```
arinchoi03, line 20: [q] how come we can't do `email_notification_opt_in == "yes"` here?

spec/forms/state_file/notification_preferences_form_spec.rb
```diff
@@ -60,63 +60,92 @@
...
+          it "is invalid" do
+            form.valid?
+            expect(form).not_to be_valid
+          end
+
+          it "adds an error to the email_address attribute" do
+            form.valid?
+            expect(form.errors[:email_address]).to be_present
+          end
+        end
```
arinchoi03, line 101: [dust] can we fold the "is invalid" and "adds an error to the email_address attribute" into one `it` block? I do see it as a pattern here but wondering if there's a good reason why these need to be separate...open to either, interested to hear what your thoughts are here

spec/forms/state_file/notification_preferences_form_spec.rb
```diff
@@ -60,63 +60,92 @@
...
+
+          it "is invalid" do
+            form.valid?
+            expect(form).not_to be_valid
+          end
+
+          it "adds a presence error to the email_address attribute" do
+            form.valid?
+            expect(form.errors[:email_address]).to include "Can't be blank."
+          end
```
arinchoi03, line 120: can we also just fold this into one it block?

spec/forms/state_file/notification_preferences_form_spec.rb
```diff
@@ -60,63 +60,92 @@
...
-      context "when email address is missing" do
-        let(:invalid_params) do
+      context "with existing email address" do
+        let!(:intake) { create :state_file_az_intake, email_address: "existing@example.com" }
+
+        let(:params) do
           {
             email_notification_opt_in: "yes",
-            email_address: ""
+            email_address: "" # Empty email address should be valid when there's an existing one
```
arinchoi03, line 130: [pebble] is there also a spec where the params look like:

```
{email_notification_opt_in: "yes"}
```
b/c I'm assuming this is how the params look when user has an email already & clicks opt-in. could be wrong tho

arinchoi03 (commented):
few comments!

config/locales/es.yml
```diff
@@ -3140,6 +3140,7 @@ es:
           message_title: Detalles de la mensajería
           opt_in_label: "¡Seleccione la(s) opción(es) que funcionen mejor para usted!"
           phone_number: Tu número de teléfono
+          provided_email: Le enviaremos actualizaciones sobre su declaración de impuestos a %{email_address}.
```
arinchoi03, line 3143: should flag this translation needs to be vetted & added in transifex

app/forms/state_file/notification_preferences_form.rb
tahsinaislam: yeah it updates to nil otherwise

spec/forms/state_file/notification_preferences_form_spec.rb
tahsinaislam: the user can click this regardless of whether the user had an email already

app/forms/state_file/notification_preferences_form.rb
arinchoi03: oh right, i guess the email param still comes through even for the statement-only version

arinchoi03 (approved):
nice! 👏

PR #5176, opened 2024-12-11
NJ 194 - solve IRS import 500 error by allow NJ w2 state amount to exceed federal

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/194

## What was done?
- Add testing persona to match IRS failure case - when the state wage amount on the w2 (box 16) exceeds the federal wage amount (box 1)
image

- update logic for state_file_w2s to allow for state wage to exceed federal only when state code is NJ

per Gabriel, this is NJ special logic:
> he reminded me that NJ taxable income can indeed be higher than federal taxable income on a W2. This isn't the case in the other states CFA supports, because they are all calculating based off the federal adjusted gross income (AGI), but NJ does not.

## How to test?
- use `irs_test_w2_state_wage_exceeds_federal` and validate that DF import is successful and you can move into intake.

n/a

### DIFF
--- a/app/models/state_file_base_intake.rb
+++ b/app/models/state_file_base_intake.rb
@@ -147,7 +147,6 @@ class StateFileBaseIntake < ApplicationRecord
         local_income_tax_amount: direct_file_w2.LocalIncomeTaxAmt,
         local_wages_and_tips_amount: direct_file_w2.LocalWagesAndTipsAmt,
         locality_nm: direct_file_w2.LocalityNm,
-        state_code: direct_file_w2.StateAbbreviationCd,
         state_income_tax_amount: direct_file_w2.StateIncomeTaxAmt,
         state_wages_amount: direct_file_w2.StateWagesAmt,
         state_file_intake: self,
--- a/spec/models/state_file_base_intake_spec.rb
+++ b/spec/models/state_file_base_intake_spec.rb
@@ -104,7 +104,6 @@ describe StateFileBaseIntake do
       expect(w2.state_wages_amount).to eq 35000
       expect(w2.wages).to eq 35000
       expect(w2.employer_ein).to eq "234567891"
-      expect(w2.state_code).to eq "AZ"
     end
 
     it "reads in w2s and removes dash/hyphen from employer_state_id_num" do


app/models/state_file_w2.rb
```diff
@@ -91,7 +92,7 @@ def validate_tax_amts
         errors.add(:local_income_tax_amount, I18n.t("state_file.questions.w2.edit.wages_amt_error", wages_amount: w2.WagesAmt))
         errors.add(:state_income_tax_amount, I18n.t("state_file.questions.w2.edit.wages_amt_error", wages_amount: w2.WagesAmt))
       end
-      if state_wages_amount.present? && state_wages_amount > w2.WagesAmt
+      if state_wages_amount.present? && state_wages_amount > w2.WagesAmt && state_code != "NJ"
```
mluedke2, line 95: [dust] is it worth making an enum or list out of state codes to not have to do string matching? I'm assuming they are all the official state abbreviations as used by USPS, which would make this pretty universal and low-risk... but my sense is it would be slightly more rubyish to do `state_code.is_nj?` etc?

mluedke2 (approved):
looks good to me! does this mean it's not only the first NJ-specific but the first rule specific to any state when it comes to the XML??

app/models/state_file_w2.rb
```diff
@@ -15,6 +15,7 @@
 #  local_income_tax_amount     :decimal(12, 2)
 #  local_wages_and_tips_amount :decimal(12, 2)
 #  locality_nm                 :string
+#  state_code                  :string
```
mmazanec22, line 18: [pebble] Why add state code instead of getting it from the intake? And maybe adding a convenience method on this model?

app/models/state_file_w2.rb
mmazanec22: Agree that this seems like too high an effort/impact ratio

mluedke2: works for me 👍

app/models/state_file_w2.rb
mluedke2: could a single intake conceivably have several w2s that are each from different states?

aloverso: It felt weird to me if the validation is on a W2 level, then we should be considering the state code as listed on the W2, not the intake it is attached to. I can confirm with product though

aloverso: Confirmed with Amani that this is a W2-specific validation, not based on the taxpayer's state (ie, intake). Moot point for this filing year but for the logical approach, I think it is correct to keep as is.

tahsinaislam: The W2's StateAbbreviationCd refers to the employer's address. If this change should apply to all W2s for any NJ filer, than we should be determining whether to use it based on the intake's state code, not the W2's.

tahsinaislam: If this is just a NJ filer specific rule and not dependent on NJ employers, we can use the method def validate_state_specific_w2_requirements to make this a state specific rule instead of adding the state code to the model.

tahsinaislam (commented):
I think we can refactor this using the validate_state_specific_w2_requirements method. but would like to discuss!

app/models/state_file_nj_intake.rb
```diff
@@ -194,5 +194,7 @@ def disqualifying_eligibility_rules
       health_insurance_eligibility: "ineligible"
     }
   end
+
+  def validate_state_specific_w2_requirements(w2); end
```
mmazanec22, line 198: [dust] A comment calling out that we're intentionally overriding the base class might be helpful, but then again maybe that should be obvious

mmazanec22 (approved):
This is a great refactor!

app/models/state_file_nj_intake.rb
aloverso: I was hoping that would be clear from the test for this - ie, that test will fail without this line, but could go either way here

tahsinaislam (approved):

PR #5191, opened 2024-12-12
sms terms only show to intakes with phone numbers

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1436

## What was done?
- made sms terms page only show to intakes with phone numbers
## How to test?
- start an intake and verify that you are not shown sms page

### DIFF
--- a/app/controllers/state_file/questions/sms_terms_controller.rb
+++ b/app/controllers/state_file/questions/sms_terms_controller.rb
@@ -1,6 +1,9 @@
 module StateFile
   module Questions
     class SmsTermsController < QuestionsController
+      def self.show?(intake)
+        intake.phone_number?
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/state_file/questions/sms_terms_controller.rb
```diff
@@ -1,6 +1,9 @@
 module StateFile
   module Questions
     class SmsTermsController < QuestionsController
+      def self.show?(intake)
+        intake.phone_number?
```
squanto, line 5: I think you'll also want to update how the terms and condition controller shows if this one is skipped

DrewProebstel: don't we always want Terms and Conditions to show?

PR #5202, opened 2024-12-13
Validate XML for 1099Gs and add tests for MD 1099G

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1306

## What was done?
* checked validations for 1099G (MD & all other states) 
* truncated name string to 35 characters as per the validation

### DIFF
--- a/app/lib/submission_builder/state1099_g.rb
+++ b/app/lib/submission_builder/state1099_g.rb
@@ -25,7 +25,7 @@ module SubmissionBuilder
                       form1099g.intake.spouse
                     end
         xml.RecipientSSN recipient.ssn
-        xml.RecipientName sanitize_for_xml(recipient.full_name)
+        xml.RecipientName sanitize_for_xml(recipient.full_name, 35)
         xml.RecipientUSAddress do
           xml.AddressLine1Txt sanitize_for_xml(form1099g.recipient_address_line1, 35)
           xml.AddressLine2Txt sanitize_for_xml(form1099g.recipient_address_line2, 35) if form1099g.recipient_address_line2.present?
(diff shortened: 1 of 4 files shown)


app/lib/submission_builder/state1099_g.rb
```diff
@@ -25,21 +25,23 @@ def document
...
-        xml.UnemploymentCompensation form1099g.unemployment_compensation_amount&.round
-        xml.FederalTaxWithheld form1099g.federal_income_tax_withheld_amount&.round
-        xml.State1099GStateLocalTaxGrp do
-          xml.StateTaxWithheldAmt form1099g.state_income_tax_withheld_amount&.round
-          xml.StateAbbreviationCd state_abbreviation
-          if form1099g.state_identification_number && form1099g.state_identification_number != ''
-            xml.PayerStateIdNumber form1099g.state_identification_number
+        xml.UnemploymentCompensation form1099g.unemployment_compensation_amount&.round if form1099g.unemployment_compensation_amount.present?
+        xml.FederalTaxWithheld form1099g.federal_income_tax_withheld_amount&.round if form1099g.federal_income_tax_withheld_amount.present?
+        if form1099g.state_income_tax_withheld_amount.present?
```
arinchoi03, line 38: so it looks like `StateAbbreviationCd` is the only required node within the `State1099GStateLocalTaxGrp`, so maybe we don't need to check if the `state_income_tax_withheld_amount` is present here?

app/lib/submission_builder/state1099_g.rb
```diff
@@ -25,21 +25,23 @@ def document
...
+          xml.State1099GStateLocalTaxGrp do
+            xml.StateTaxWithheldAmt form1099g.state_income_tax_withheld_amount&.round
+            xml.StateAbbreviationCd state_abbreviation
+            if form1099g.state_identification_number && form1099g.state_identification_number != ''
```
arinchoi03, line 42: can we change this to `form1099g.state_identification_number.present?`

arinchoi03 (approved):
few questions but non-blocking

app/lib/submission_builder/state1099_g.rb
arinchoi03: I tried this locally -- 
we cannot leave state_income_tax_withheld_amount field blank (we get a validation error)
however, we _can_ enter `0` -- not sure if this is intentional or if we should add a validation error to not allow it? (Q for product)

PR #5205, opened 2024-12-13
Nj 204 year

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/204

## What was done?
- Require the pin as part of the signature. There is a separate ticket to do this after confirming with taxation how it should function, but we wanted to see if adding the signature and date would get NJ returns accepted.
- Remove some redundant tests in nj return xml spec. Some of the tests that still exist look silly now, but this is an improvement because we're only checking for build errors once

## How to test?
Submit one of the valid returns (Jennifer Brown etc)

I did not make any visual changes, but just added db fields and copied the func that enables pin from the MD intake

### DIFF
--- a/spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
+++ b/spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
@@ -9,6 +9,13 @@ describe SubmissionBuilder::Ty2024::States::Nj::NjReturnXml, required_schema: "n
     let(:build_response) { described_class.build(submission, validate: true) }
     let(:xml) { Nokogiri::XML::Document.parse(build_response.document.to_xml) }
 
+    before do
+      intake.primary_signature_pin = '12345'
+      if intake.filing_status_mfj?
+        intake.spouse_signature_pin = '12345'
+      end
+    end
+
     after do
       expect(build_response.errors).not_to be_present
     end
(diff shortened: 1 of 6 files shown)


embarnard (approved):
✅

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -33,10 +33,8 @@
       context "with one dep" do
         let(:intake) { create(:state_file_nj_intake, :df_data_one_dep) }
         it "does not error" do
-          builder_response = described_class.build(submission)
-          expect(builder_response.errors).not_to be_present
```
jachan, line 37: [DUST] You mention that we're removing these tests because they're redundant, where else are we validating that the XML is built without errors?

jachan (approved):
Looks good! Curious how we'll use the PIN in the future. Left a non-blocking question about the redundant testing just so I can learn a bit more about the codebase.

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
mmazanec22: We're validating that on line 12 with the `after do`

mluedke2: [dust] i know this is after the fact and not important but if we revisit this for any reason it would probably be worth rewording `it "does not error"` more to what is in the expectations... `it "has not-nil salary and income"`, maybe?

mluedke2:
👍  but i have a question. The ticket mentions `You need to remove the xmlns="http://www.irs.gov/efile"`... was that incorrect?

PR #5208, opened 2024-12-13
Gyr1 608/remove adopted and buyer

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-608

## What was done?
- Removes the controllers from GYR flow and sets the `::show?` method to be false on both controllers to be double sure they won't pop up.

## How to test?
- Navigate through the flow, being careful to click both that you have owned a home and that you do have dependents.
- After the dependent care question, you should not see the adopted child question
- After the mortgage interest question, you should not see the homeowner buyer credit question

- Before
- After

### DIFF
--- a/app/controllers/questions/adopted_child_controller.rb
+++ b/app/controllers/questions/adopted_child_controller.rb
@@ -2,9 +2,7 @@ module Questions
   class AdoptedChildController < QuestionsController
     include AuthenticatedClientConcern
 
-    def self.show?(intake)
-      intake.had_dependents_yes?
-    end
+    def self.show?(intake) = false
 
     layout "yes_no_question"
   end
--- a/app/controllers/questions/homebuyer_credit_controller.rb
+++ b/app/controllers/questions/homebuyer_credit_controller.rb
@@ -4,9 +4,7 @@ module Questions
 
     layout "yes_no_question"
 
-    def self.show?(intake)
-      intake.ever_owned_home_yes?
-    end
+    def self.show?(intake) = false
 
     private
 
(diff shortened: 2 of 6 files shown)


app/controllers/questions/adopted_child_controller.rb
```diff
@@ -2,10 +2,8 @@ module Questions
   class AdoptedChildController < QuestionsController
     include AuthenticatedClientConcern
 
-    def self.show?(intake)
-      intake.had_dependents_yes?
-    end
+    def self.show?(intake) = false
```
anisharamnani, line 5: these two controllers seem to be intentionally not removed, but rather changing show to false. what is the logic behind that? why not delete them?

powersurge360: I thought about this yesterday and I think I'm leaning towards just disabling because I am reluctant to remove too much on my first ticket on unfamiliar code that isn't well socialized among the team. If you feel _really_ strongly that I should fully remove controllers, remove tests, remove forms, and remove views I might be persuaded, but I currently am preferring a light touch.

spompea-cfa: Good question here. I normally love-love-love tossing out code, but the light touch might be better here, for a couple reasions: this feels like an unusual situation: a bunch of last-minute changes that could, conceivably, be reversed again in next year's form -- or, heaven forbid, even reversed in some sort of 11th-hour update in the next few weeks (which is unlikely but not impossible). and also, keeping the per-ticket surface area of change small will help us meet the tight deadline we have. that said, i am open to discussing further 😄

spompea-cfa (approved):
lgtm

anisharamnani (approved):
also putting a mark of approval! sorry for the delayed review

app/controllers/questions/adopted_child_controller.rb
anisharamnani: thanks to you both for explaining!

PR #5216, opened 2024-12-16
Fyst 1457 update behavior of software id field in xml

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1457

## Is PM acceptance required? 

## What was done?
- A seperate software ID was added to each credential file for MD because it requires a u nique software id only in use for Maryland. A default Software ID is used for all other states

## How to test?
- updated return header test


spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -22,7 +22,8 @@
           intake.direct_file_data.mailing_zip = mailing_zip
           intake.direct_file_data.tax_return_year = tax_return_year
           allow(EnvironmentCredentials).to receive(:irs).with(:efin).and_return efin
-          allow(EnvironmentCredentials).to receive(:irs).with(:sin).and_return sin
+          software_id = StateFile::StateInformationService.software_id_key(state_code).to_sym
+          allow(EnvironmentCredentials).to receive(:irs).with(software_id).and_return sin
```
arinchoi03, line 26: can we check in the test that the `SoftwareID` gets filled in with the correct value here?

arinchoi03 (approved):
one suggestion for test

PR #5219, opened 2024-12-16
Fyst 1305 w2 validate against schema

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1305
- No - merge after code review approval; I think this is sufficient but happy to have acceptance test if we think we need it

## What was done?
- Minor clean up
- Removed upcasing of locality name, as that should already be handled in the `before_validation` part of locality_name column update. I think this is fine, we always sync w2 data when we receive it from DF & if the user updates it, this before validation action will run.
- Discussed with Jenny that the W2 xml filling (submission builder) looks so different from other forms b/c it was built quickly in 2023 to accommodate tight deadlines for FYST. Ideally we'd at least update the submission builder to be more like the other builders.
## How to test?
  - all tests should still pass
  - If federal XML contains `LocalityNm` that is not all capitalized, it should be at the state XML output for `LocalityNm`


mrotondo (approved):
looks good - i agree that the model should get its `locality_nm` upcased in the before_validation hook. the only issue i could imagine is if the `locality_nm` is overwritten in a test by assigning the attribute without actually saving the object to the database, but that's not a case worth designing the code around imo.

PR #5222, opened 2024-12-17
Add all box 14 nodes to w2

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/209

## What was done?
- Add all edited box 14 nodes for each state to state file w2 xml
- Update rubocop

## How to test?
- go through any persona with W2s (ex. Zeus Many W2s)
- edit to add Ui and FLI holdings

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,7 +106,7 @@ group :development, :test do
   gem 'rails-controller-testing'
   gem 'pry-byebug'
   gem 'guard-rspec', require: false
-  gem 'rubocop', '~> 1.47.0', require: false
+  gem 'rubocop', '~> 1.53.0', require: false
   gem 'rubocop-performance', '~> 1.16.0', require: false
   gem 'rubocop-rspec', '~> 2.18.0', require: false
   gem 'i18n-tasks', require: false
--- a/app/lib/submission_builder/return_w2.rb
+++ b/app/lib/submission_builder/return_w2.rb
@@ -14,10 +14,7 @@ module SubmissionBuilder
           state_local_tax_grp_node.remove
         end
 
-        node_after_box_14_codes = state_local_tax_grp_node
-        unless state_local_tax_grp_node.present?
-          node_after_box_14_codes = @xml_node.at(:StandardOrNonStandardCd)
-        end
+        node_after_box_14_codes = state_local_tax_grp_node || @xml_node.at(:StandardOrNonStandardCd)
         state_code = @intake_w2.state_file_intake.state_code
         box14_codes = StateFile::StateInformationService.w2_supported_box14_codes(state_code)
         box14_codes.each do |code|
--- a/app/lib/submission_builder/return_w2.rb
+++ b/app/lib/submission_builder/return_w2.rb
@@ -35,17 +35,17 @@ module SubmissionBuilder
       if existing_xml_node
         existing_xml_node.at('Amt').content = @intake_w2[field_name].round.to_s
       else
-        this_code_node = Nokogiri::XML::Node.new('OtherDeductionsBenefitsGrp', @xml_node)
+        new_xml_node = Nokogiri::XML::Node.new('OtherDeductionsBenefitsGrp', @xml_node)
 
         desc_node = Nokogiri::XML::Node.new('Desc', @xml_node)
         desc_node.content = field_as_desc
         amt_node = Nokogiri::XML::Node.new('Amt', @xml_node)
         amt_node.content = @intake_w2[field_name].round.to_s
 
-        this_code_node.add_child(desc_node)
-        this_code_node.add_child(amt_node)
+        new_xml_node.add_child(desc_node)
+        new_xml_node.add_child(amt_node)
 
-        node_after_box_14_codes.add_previous_sibling(this_code_node)
+        node_after_box_14_codes.add_previous_sibling(new_xml_node)
       end
     end
   end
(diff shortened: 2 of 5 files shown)


app/lib/submission_builder/ty2024/states/nj/documents/nj_w2.rb
```diff
@@ -4,17 +4,26 @@ module States
...
         module Documents
           class NjW2 < SubmissionBuilder::ReturnW2
+            def schema_file
+              SchemaFileLoader.load_file("us_states", "unpacked", "NJIndividual2024V0.1", "NJCommon", "NJW2.xsd")
+            end
+
             def document
               xml_node = super
-
+
```
aloverso, line 13: [pebble] remove whitespace here

spec/lib/submission_builder/ty2024/states/nj/documents/nj_w2_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        box14_ui_hc_wd: 99,
+        box14_fli: 100, 
+        employer_ein: '123456789', 
+        wages: '999'
+        )
+      }
+
+      it "adds box 14 nodes" do
+        box_14_nodes = xml.css("OtherDeductsBenefits")
+        expect(box_14_nodes.length).to eq(3)
```
aloverso, line 27: [dust] could we expect the correct values too?

app/lib/submission_builder/return_w2.rb
```diff
@@ -2,47 +2,58 @@ module SubmissionBuilder
...
-            if state_local_tax_grp_node.present?
-              state_local_tax_grp_node.add_previous_sibling(stpickup_node)
-            else
-              xml_node.at(:StandardOrNonStandardCd).add_previous_sibling(stpickup_node)
-            end
-          end
+        node_after_box_14_codes = state_local_tax_grp_node
+        unless state_local_tax_grp_node.present?
+          node_after_box_14_codes = @xml_node.at(:StandardOrNonStandardCd)
+        end
```
mluedke2, line 20: [dust] - could the above four lines be condensed like this?

`node_after_box_14_codes = state_local_tax_grp_node || @xml_node.at(:StandardOrNonStandardCd)`

app/lib/submission_builder/return_w2.rb
```diff
@@ -2,47 +2,58 @@ module SubmissionBuilder
...
+          desc_node.content = field_as_desc
+          amt_node = Nokogiri::XML::Node.new('Amt', @xml_node)
+          amt_node.content = @intake_w2[field_name].round.to_s
+
+          this_code_node.add_child(desc_node)
+          this_code_node.add_child(amt_node)
+
+          node_after_box_14_codes.add_previous_sibling(this_code_node)
+        end
+      end
```
mluedke2, line 56: [not sure how to categorize this] - I'm surprised/wondering why there isn't a test file in `spec/lib/submission_builder/return_w2`, that could have test expectations to confirm the new code here is working correctly. Is that something that should be in the scope of this PR (which would make this a boulder) or am I missing a reason why that's not needed (which would make this dust)?

Gemfile
```diff
@@ -106,7 +106,7 @@ group :development, :test do
   gem 'rails-controller-testing'
   gem 'pry-byebug'
   gem 'guard-rspec', require: false
-  gem 'rubocop', '~> 1.47.0', require: false
+  gem 'rubocop', '~> 1.53.0', require: false
```
jenny-heath, line 109: i'm not sure what the implications of this are, if any! since we don't enforce use of rubocop i assume it's okay though

mmazanec22: Just fixing some local rubocop struggles for meeeee

spec/lib/submission_builder/ty2024/states/nj/documents/nj_w2_spec.rb
jenny-heath: ^ second that

app/lib/submission_builder/return_w2.rb
```diff
@@ -2,42 +2,54 @@ module SubmissionBuilder
...
+      field_name = "box14_#{code.downcase}".to_sym
+      field_as_desc = code.tr '_', '/' 
+
+      return if !@intake_w2[field_name].present? || !@intake_w2[field_name].positive?
+
+      existing_xml_node = @xml_node.at_xpath("//OtherDeductionsBenefitsGrp[Desc='#{field_as_desc}']")
+      if existing_xml_node
+        existing_xml_node.at('Amt').content = @intake_w2[field_name].round.to_s
+      else
+        this_code_node = Nokogiri::XML::Node.new('OtherDeductionsBenefitsGrp', @xml_node)
```
jenny-heath, line 41: [dust] could we call this `new_xml_node` to be consistent with `existing_xml_node`?

app/lib/submission_builder/return_w2.rb
jenny-heath: i think this is actually tested in `state_return_spec.rb` (it should probably be moved into a file called `return_w2_spec.rb` but that hasn't happened yet). i _think_ that should make sure this is still working correctly...

jenny-heath (approved):
i think the one thing this is missing is a test that checks the specific nj box 14 nodes (similar to "Maryland W2 Box 14 STPICKUP handling" in `state_return_spec`) but i guess that's kind of up to you all?

PR #5178, opened 2024-12-11
Nj 191 property tax worksheet

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/191
 
## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

**Reminder**: merge main into this branch and get green tests before merging to main
## What was done?
- Replaced concept of "soft no - unsupported" with a screen that prompts the user to fill out a worksheet and tell us the value

## How to test?
- Use Zeus 1 dependent, navigate to the New Jersey Property Tax credit, select Homeowner
- Check the "My main home was subject to property taxes" and the "Did you share ownership of your main home with someone who is not your spouse" checkboxes.
- Verify that you are taken to the Property Tax Worksheet page, rather than the unsupported page
- Enter 10000 for property tax paid
- Validate that XML includes
```
<PropertyTaxDeductOrCredit>
    <TotalPropertyTaxPaid>10000</TotalPropertyTaxPaid>
    <Homeowner>X</Homeowner>
    <PropertyTaxDeduction>10000</PropertyTaxDeduction>
</PropertyTaxDeductOrCredit>
```
- Validate that PDF has 10000 for line 41, property tax deduction and line 40a.
image

Repeat for each of the different ACs, including renters.

## Screenshots (for visual changes)
- Before - Unsupported
Pasted Graphic

- After - Worksheet page
image

PR #5224, opened 2024-12-17
Nj 192 property tax both

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/192

## What was done?
- Support "both" flow - when both homeowner and tenant, show homeowner flow and then tenant flow
- Handle if ineligible homeowner but eligible tenant (when low income exemption, this means show tenant eligibility)
- Handle if eligible homeowner but ineligible tenant (when low income exemption, this means do not show tenant eligibility)
- NJ1040 calculations (needs PM review)
- "back" button in property tax flow always goes back to household rent/own page
- review screen combination of items
- ineligible page tracks rent/own param, and calculates reason for ineligiblity
- refactor: remove all "show" methods from property tax flow and calculate next controller dynamically always based on current flow status
- add e2e tests for property tax flow route variations

## How to test?
thoroughly 

Too many possible flows to show all of them, but here is the most basic "both" flow.
The end property tax deduction is $13,600 because 10,000 + (0.18 * 20,000)

https://github.com/user-attachments/assets/c46c4aed-fc00-4419-9138-e0780c20b98d


spec/features/state_file/complete_intake_spec.rb
```diff
@@ -738,5 +768,303 @@
...
+      property = I18n.t("state_file.questions.nj_ineligible_property_tax.edit.on_rental")
+      reason = I18n.t("state_file.questions.nj_ineligible_property_tax.edit.reason_property_taxes")
+      expect(page).to have_text strip_html_tags(I18n.t("state_file.questions.nj_ineligible_property_tax.edit.title_html", property: property))
+      expect(page).to have_text strip_html_tags(I18n.t("state_file.questions.nj_ineligible_property_tax.edit.reason_html", reason: reason))
+      click_on I18n.t("general.continue")
+
+      expect(page).to have_text I18n.t("state_file.questions.nj_estimated_tax_payments.edit.title", filing_year: filing_year)
+    end
+
+    it "handles property tax eligible tenant flow when low income but meets exemption", required_schema: "nj" do
```
jachan, line 832: [DUST] Can we use `context` blocks to group the income exception test cases?

app/helpers/state_file/nj_property_tax_flow_helper.rb
```diff
@@ -0,0 +1,13 @@
+module StateFile
+  module NjPropertyTaxFlowHelper
```
jachan, line 2: [DUST] `property_tax_flow_helper` feels generic, prefer `nj_property_tax_off_ramp` or `nj_property_tax_final_page`

jachan (commented):
Capturing comments from our sync PR review

spec/features/state_file/complete_intake_spec.rb
jachan, line 1: [DUST] These are very cool tests!

jachan (approved):
Minor non-blocking comment about a potentially unused logic path. Great work on this PR, it was super complex but your approach with TDD and deterministic paths through the Property Tax controller made it much easier to understand.

app/controllers/state_file/questions/nj_household_rent_own_controller.rb
```diff
@@ -8,20 +8,18 @@ def next_path
...
-        when 'own'
+        when 'own', 'both'
           NjHomeownerEligibilityController.to_path_helper(options)
-        when 'both'
-          NjIneligiblePropertyTaxController.to_path_helper(options)
         when 'neither'
           NjIneligiblePropertyTaxController.to_path_helper(options)
         else
-          super
+          StateFile::NjPropertyTaxFlowHelper.next_controller(options)
```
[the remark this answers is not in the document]
aloverso: No, it is not possible to ever hit this case, but removing the else block causes a linting error. 
image

app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
```diff
@@ -3,8 +3,66 @@ module Questions
...
+        end
+
+        if params[:on_home_or_rental] == "rental"
+          if current_intake.tenant_home_subject_to_property_taxes_no?
+            return "property_taxes"
+          elsif current_intake.tenant_building_multi_unit_yes? && current_intake.tenant_access_kitchen_bath_no?
+            return "multi_unit_conditions"
+          end
+        end
+      end
```
mluedke2, line 29: [sand] - (mentioned on live review) since `determine_reason()` is dependent on non-view logic, should probably live in `NjPropertyTaxEligibility` so there is one source of truth

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -738,5 +768,303 @@
...
+
+      # tenant eligibility page
+      expect(page).to have_text I18n.t("state_file.questions.nj_tenant_eligibility.edit.title", filing_year: filing_year)
+      check I18n.t('state_file.questions.nj_tenant_eligibility.edit.tenant_home_subject_to_property_taxes')
+      click_on I18n.t("general.continue")
+
+      # skips rent paid page
+
+      expect(page).to have_text I18n.t("state_file.questions.nj_estimated_tax_payments.edit.title", filing_year: filing_year)
+    end
```
mluedke2, line 1068: [boulder] - sand if engineer-only, but if requesting PM review of this too, we should probably DRY these now to be more readable. One way to do it would be anywhere there is currently a comment in this diff could be a helper function instead

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -325,11 +325,38 @@ def calculate_line_39
...
+        when "own"
+          StateFile::NjHomeownerEligibilityHelper.determine_eligibility(@intake) == StateFile::NjHomeownerEligibilityHelper::INELIGIBLE
+        when "rent"
+          StateFile::NjTenantEligibilityHelper.determine_eligibility(@intake) == StateFile::NjTenantEligibilityHelper::INELIGIBLE
+        when "both"
+          StateFile::NjTenantEligibilityHelper.determine_eligibility(@intake) == StateFile::NjTenantEligibilityHelper::INELIGIBLE &&
+            StateFile::NjHomeownerEligibilityHelper.determine_eligibility(@intake) == StateFile::NjHomeownerEligibilityHelper::INELIGIBLE
+        else
+          nil
+        end
```
mluedke2, line 340: [boulder/sand] - discussed this with @jachan  on this PR https://github.com/codeforamerica/vita-min/pull/5178#discussion_r1884267898

since we have three utils for this: `NjHomeownerEligibilityHelper`, `NjTenantEligibilityHelper`, and `NjPropertyTaxEligibility`... we should consolidate the logic at some point and it being a sand on multiple PRs makes it harder for the next person, so sooner would be less debt.

We could start by moving `is_ineligible_or_unsupported_for_property_tax_credit()` out of the calculator, and then expanding the result types from `INELIGIBLE` to have more context.

app/lib/navigation/state_file_nj_question_navigation.rb
```diff
@@ -41,14 +41,14 @@ class StateFileNjQuestionNavigation < Navigation::StateFileBaseQuestionNavigatio
...
-                                          Navigation::NavigationStep.new(StateFile::Questions::NjHomeownerPropertyTaxWorksheetController),
                                           Navigation::NavigationStep.new(StateFile::Questions::NjHomeownerPropertyTaxController),
-                                          Navigation::NavigationStep.new(StateFile::Questions::NjTenantPropertyTaxWorksheetController),
+                                          Navigation::NavigationStep.new(StateFile::Questions::NjHomeownerPropertyTaxWorksheetController),
+                                          Navigation::NavigationStep.new(StateFile::Questions::NjTenantEligibilityController),
                                           Navigation::NavigationStep.new(StateFile::Questions::NjTenantRentPaidController),
-                                          Navigation::NavigationStep.new(StateFile::Questions::NjEitcQualifyingChildController),
+                                          Navigation::NavigationStep.new(StateFile::Questions::NjTenantPropertyTaxWorksheetController),
+                                          Navigation::NavigationStep.new(StateFile::Questions::NjIneligiblePropertyTaxController),
```
mluedke2, line 49: [sand] - we could test the flow after removing these screens from the navigation since they only get navigated to "manually" using `next_step()` anyway. That may also free us from the need to tie the offramp from the Property Tax flow to the question that happens to come next.

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -325,11 +325,38 @@ def calculate_line_39
...
       end
 
       def calculate_line_40a
+        if @intake.household_rent_own_both?
+          return nil unless @intake.rent_paid&.positive? || @intake.property_tax_paid&.positive?
+          property_tax_paid = @intake.property_tax_paid&.positive? ? @intake.property_tax_paid : 0
+          rent_paid = @intake.rent_paid&.positive? ? @intake.rent_paid * RENT_CONVERSION : 0
```
mluedke2, line 347: [dust] - two above lines could be

```
property_tax_paid = [@intake.property_tax_paid, 0].max
rent_paid = [@intake.rent_paid * RENT_CONVERSION, 0].max
```

jnf (approved):
sorry it took so long, but this looks great.

app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
aloverso: Captured in follow-up ticket https://github.com/newjersey/affordability-pm/issues/227

app/lib/efile/nj/nj1040_calculator.rb
aloverso: Captured in https://github.com/newjersey/affordability-pm/issues/227

app/lib/navigation/state_file_nj_question_navigation.rb
aloverso: captured in https://github.com/newjersey/affordability-pm/issues/227

app/lib/efile/nj/nj1040_calculator.rb
aloverso: Hm, this doesn't actually work because of the case where `@intake.property_tax_paid` or `@intake.rent_paid` could be nil. Either the integer comparison fails or the `*` operation fails on nil. It feels like I'd still need a ternary to test for `present` in this case, so I might as well just do the ternary as is? Any ideas on how to use the max syntax if possibly nil?

mluedke2: i missed that one could be nil while the other is positive! you'd have to add `.to_i` like this then (`nil.to_i == 0`) 

```
property_tax_paid = [@intake.property_tax_paid.to_i, 0].max
rent_paid = [@intake.rent_paid.to_i * RENT_CONVERSION, 0].max
```

aloverso: Thanks so much! Updated it to use `to_f` so that our rounding logic still works, but this is great, thanks

mluedke2 (approved):
looks good! i really like the e2e refactor and thanks for capturing other refactor suggestions in the follow-up ticket

PR #5226, opened 2024-12-17
FYST-1322 Validate MD 502CR against schema & MD validations sheet

arinchoi03:
- [MD Validations sheet
](https://docs.google.com/spreadsheets/d/1kqVWAw2HoqjBTzBLgDAAjpZMjzZhDTuG8hyjNfYRdcs/edit?gid=1894126376#gid=1894126376)

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1322

## What was done?
- Quite a lot of changes to the calculations, please see the comments I left in the PR to point to the validation rules coming from the MD validation sheet (linked at the top)
- Not adding elements to the xml if they are blank; take special attention to the Decimal field. This line _could_ be 0.
## How to test?
  - Unit tests


spec/lib/efile/md/md502cr_calculator_spec.rb
```diff
@@ -330,39 +412,44 @@
       main_calculator.calculate
     end
 
-    context "when single filer" do
-      let(:filing_status) { "single" }
+    context "when filing_status is not mfj" do
+      %w[single married_filing_separately head_of_household qualifying_widow dependent].each do |filing_status|
```
arinchoi03, line 416: according to prod/program, "dependent" filing status should never be in a situation where they're calculating this value. However, just to be thorough, I'm going to include it to follow the rule

app/lib/submission_builder/document.rb
```diff
@@ -101,6 +101,13 @@ def add_non_zero_value(xml, elem_name, line)
       end
     end
 
+    def add_positive_decimal_value(xml, elem_name, line)
+      value = calculated_fields.fetch(line)
+      if value.present? && value.positive?
+        xml.send(elem_name, value)
+      end
+    end
```
arinchoi03, line 109: i was trying to do this in the submission builder, i guess it's only used it one place so happy to put it back there but maybe it'll be useful for decimal values?

app/lib/efile/md/md502cr_calculator.rb
```diff
@@ -132,7 +136,8 @@ def calculate_md502_cr_part_b_line_3
       end
 
       def calculate_md502_cr_part_b_line_4
-        (line_or_zero(:MD502CR_PART_B_LINE_2) * @lines[:MD502CR_PART_B_LINE_3]&.value).round(0)
+        calculated_value = (line_or_zero(:MD502CR_PART_B_LINE_2) * @lines[:MD502CR_PART_B_LINE_3]&.value).round(0)
+        [calculated_value, 1_344].min
```
arinchoi03, line 140: see [jira comment](https://codeforamerica.atlassian.net/browse/FYST-1322?focusedCommentId=14163)

>Part B Line 4 Amount of Credit for Child and Dependent Care expenses - Multiply Line 2 by decimal amount on Line 3 not to exceed $1,344.

app/lib/efile/md/md502cr_calculator.rb
```diff
@@ -42,13 +42,17 @@ def calculate
...
             credit = 1750
           elsif @intake.primary_senior? ^ @intake.spouse_senior?
             credit = 1000
           end
-        elsif (filing_status_single? || filing_status_mfs?) && agi <= 100_000
+        elsif (filing_status_qw? || filing_status_hoh?) && agi <= 150_000
+          if @intake.primary_senior?
+            credit = 1750
+          end
+        elsif (filing_status_single? || filing_status_mfs? || filing_status_dependent?) && agi <= 100_000
```
arinchoi03, line 55: from validation doc:
> Part M, Line 1 Senior Tax Credit - If filing Married Filing Jointly, and federal adjusted gross income does not exceed $150,000, amount cannot exceed $1,750 when both individuals are at least 65 or $1,000 if only one of the individuals is at least 65.
> Part M, Line 1 Senior Tax Credit - Amount cannot exceed $1,000 if individual is at least 65, filing Single, Married Filing Separately, or Dependent, and federal adjusted gross income does not exceed $100,000.
> Part M, Line 1 Senior Tax Credit - If filing as Head of Household or Qualifying Widow(er) who is at least 65 and federal adjusted gross income does not exceed $150,000, amount cannot exceed $1,750.

app/lib/efile/md/md502cr_calculator.rb
```diff
@@ -148,13 +153,9 @@ def calculate_part_aa_line_14
...
       def calculate_part_cc_line_7
-        qualified_for_credit = if filing_status_single? || filing_status_hoh? || filing_status_qw?
-                                 @direct_file_data.fed_agi <= 59_401
-                               elsif filing_status_mfj?
-                                 @direct_file_data.fed_agi <= 89_101
-                               end
+        qualifying_fed_agi_limit = filing_status_mfj? ? 89_100 : 59_400
 
-        return unless qualified_for_credit
+        return unless @direct_file_data.fed_agi <= qualifying_fed_agi_limit
```
arinchoi03, line 158: > Part CC Line 7 Amount of Refundable Credit for Child and Dependent Care expenses - to claim this credit, taxpayers FAGI cannot exceed $89,100 for filing status Married Filing Joint and cannot exceed $59,400 for all other filing statuses.

app/lib/submission_builder/document.rb
jenny-heath: i have a question/suggestion that's hopefully not too out of scope: is `add_non_zero_value` really just excluding 0 and intending to fill in negative numbers? if so i think it would make sense to:
A) remove the `to_i` in `add_non_zero_value`, as i don't believe it's serving any purpose
B) rename this to just `add_positive_value`, since it doesn't seem specific to decimals

jenny-heath (approved):
i think this looks good! it is all very confusing though 😵‍💫

app/lib/submission_builder/document.rb
arinchoi03: hmm i had to rethink this, I don't think we need a separate method for a positive value, I was just tryign to avoid the `to_i` conversion in `add_non_zero_value` method which is why I created that extra method.
what I actually need is for the `add_non_zero` method to accommodate float values that may come through. I think still want to convert the `value` to float since technically anyone can pass in a string as a value (or a non-numeric value).

arinchoi03: after brief slack convo with Jenny, here's my decision:
although we could probably use `add_non_zero_float_value` for `value` that is supposed to be an integer type, I separated it out since it's scary to change `add_non_zero` method which is used in so many places, right before launch. Maybe we can revisit though!

PR #5231, opened 2024-12-17
White space bug: auto strip address attributes on 1099G

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-541

## What was done?
- This was reported as a bug where valid zip codes were being rejected by the validations but actually it was just an issue of leading whitespace
- I found this `auto_strip_attributes` thing which seems to do the trick. The unemployment controller doesn't have a corresponding form but this seems appropriate to put at the model level anyway
## How to test?
- I added a model spec
## Screenshots
before
image
after (had to trigger a separate validation bug so you can see it got fixed)
image


arinchoi03 (approved):
✨  amazing

PR #5235, opened 2024-12-18
Fyst 1429 update md two income subtraction to use new df json format for 1099 gs

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1429

## What was done?
- Added a new JSON accessor for 1099G data based on the federal returns we have imported from DF
   - The previous implementation was based on a draft of the potential fields, this is _the real deal_
- Reduced unemployment income by the amount the filer paid back
- Replaced uses of the old field with the new one (only MD Two Income Subtraction is affected)
## How to test?
- Existing unit test updated
- The persona "Ariel mfj 1099 g" has updated json data
N/A


spec/lib/efile/md/two_income_subtraction_worksheet_spec.rb
```diff
@@ -94,8 +120,11 @@
         intake.direct_file_json_data.interest_reports[1].recipient_tin = spouse_ssn
         intake.direct_file_json_data.interest_reports[0].amount_1099 = "1.00"
         intake.direct_file_json_data.interest_reports[1].amount_no_1099 = "2.00"
-        intake.direct_file_json_data.primary_filer&.form_1099_gs_total = "100.00"
-        intake.direct_file_json_data.spouse_filer&.form_1099_gs_total = "200.00"
+        intake.raw_direct_file_intake_data["form1099Gs"] = [{}, {}]
```
anisharamnani, line 123: what do all four types of income refer to?

app/lib/efile/md/two_income_subtraction_worksheet.rb
```diff
@@ -45,7 +45,9 @@ def calculate_fed_income(primary_or_spouse)
           .select { |form1099r| form1099r.recipient_ssn == filer_ssn }
           .sum { |form1099r| form1099r.taxable_amount&.round }
 
-        unemployment_income = find_filer_json_for(primary_or_spouse)&.form_1099_gs_total&.round || 0
+        unemployment_income = @direct_file_json_data.form_1099gs
```
anisharamnani, line 48: i think this can be tackled in another story or even just discussed, but it may make sense to create methods for these calculations on the @direct_file_json_data, so we can test the calculations. i know they might not be used elsewhere, but it could be nice to have testable chunks.

i _think_ we discussed this last time, and decided it was OK here. this might just be because i'm looking this over another time.

spec/lib/efile/md/two_income_subtraction_worksheet_spec.rb
anisharamnani: i ask because if unemployment is included, do we want to add data to the 1099G array? the comment says that we want to include the least amount of data to make this pass.

anisharamnani: so with that in mind, i may just be misunderstanding the test purpose.

app/lib/efile/md/two_income_subtraction_worksheet.rb
mpidcock: I agree with this instinct, and maybe a future refactor is still a good idea, but I'd like to know for sure how it might get re-used in the future before preemptively moving methods out.

spec/lib/efile/md/two_income_subtraction_worksheet_spec.rb
mpidcock: I updated the comments, hopefully it's more clear. The "minimum" refers to only populating fields used by the two income subtraction, not all the fields we expect to be present in the json. 

Also to clarify, calling `[{}, {}]` is to initialize an array json objects, and then the subsequent lines populate values. it's not the most graceful, and I'd prefer to make factory methods, but it keeps the scope low for setting up the test.

anisharamnani: ah, i missed the subsequent line! & that's on me, thank you for explaining.

PR #5237, opened 2024-12-18
Nj 204 2 xml validity

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/209

## What was done?
According to NJ taxation:

> All your opening tags have the xmlns = portion which should not be there. You need to remove the xmlns="http://www.irs.gov/efile" from those fields in order for your returns to reach the correct parser on our end. This is causing the current error you are receiving. I am anticipating other rejects once your returns get to the proper parser but we will work through them once we get through this issue.

> Under ReturnState you can have the xmlns entries but any tag under that cannot have it as our system can’t decipher the correct tag when it’s there.

However, we can't validate the xml without the default namespace. So we are removing the default namespace in demo and prod for NJ only to see whether that generates valid submissions.

## How to test?
After merging this, submit a return to NJ on the demo site.

No visual changes

### DIFF
--- a/app/lib/submission_builder/document.rb
+++ b/app/lib/submission_builder/document.rb
@@ -56,9 +56,9 @@ module SubmissionBuilder
 
     def build_xml_doc(tag_name, **root_node_attributes)
       default_attributes = { 'xmlns:efile' => 'http://www.irs.gov/efile' }
-      return_state_attributes = { 'xmlns' => 'http://www.irs.gov/efile' }
+      return_state_attributes = get_return_state_attributes(tag_name)
       xml_builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
-        default_attributes.merge!(return_state_attributes) if merge_state_attrs_for_tag?(tag_name)
+        default_attributes.merge!(return_state_attributes) if return_state_attributes
         xml.send(tag_name, default_attributes.merge(root_node_attributes)) do |contents_builder|
           yield contents_builder if block_given?
         end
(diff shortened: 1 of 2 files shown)


app/lib/submission_builder/document.rb
```diff
@@ -56,28 +56,28 @@ def self.build(*args)
...
-        end
+    def get_return_state_attributes(tag_name)
+      return_state_attributes = { 'xmlns' => 'http://www.irs.gov/efile' }
+      intake_class = self.submission.data_source_type
+      if intake_class == 'StateFileNyIntake' && !%w[ReturnState efile:ReturnState StateSubmissionManifest].include?(tag_name)
+        # This method is requirement from NY to remove the namespace attribute from
+        # any tag unrelated to the root XML tag. StateSubmissionManifest seems to
+        # need it also in order for the XML to be valid.
+        return false
+      elsif intake_class == 'StateFileNjIntake' && (Rails.env.production? || Rails.env.demo?)
```
mluedke2, line 77: [dust] do we want to remove the `&& (Rails.env.production? || Rails.env.demo?)`, either now or eventually?

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -8,11 +8,18 @@
...
+    
     after do
       expect(build_response.errors).not_to be_present
     end
 
+    it "does not include the xmlns attribute on any element in the demo env" do
+      allow(Rails.env).to receive(:demo?).and_return true
+      nodes_with_namespace = xml.document.xpath('//*[namespace-uri()="http://www.irs.gov/efile"]')
+      expect(nodes_with_namespace.count).to eq(0)
+      # Do not check for build errors because schema cannot be validated against without namesapce
```
mluedke2, line 20: [dust] From my reading, `expect(build_response.errors).not_to be_present` would run after this test, so this comment would not be adhered to... am I missing something?

mmazanec22: Good point, moved!

aloverso (approved):

app/lib/submission_builder/document.rb
mmazanec22: We need to leave the default xml namespace on local and test so we can validate the xml

PR #5242, opened 2024-12-18
FYST-1471-md-senior-tax-credit-incorrect-detailed-return-info-for-non-tax-clients

DrewProebstel:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1471

## What was done?
- md502_cr_part_m_line_1 is now 0 when deduction method is nonstandard
## How to test?
- use spongebob on heroku
- Risk Assessment
  - I repeated the `deduction_method_is_standard?` method from the MD502 calculator in the MD502CR calculator. I didn't think it need a concern but it isn't DRY


spec/lib/efile/md/md502cr_calculator_spec.rb
```diff
@@ -35,6 +35,17 @@
               expect(instance.lines[:MD502CR_PART_M_LINE_1].value).to eq(1_750)
               expect(instance.lines[:MD502CR_PART_AA_LINE_13].value).to eq(1_750)
             end
+
+            context "deduction method is nonstandard" do
+              before do
```
anisharamnani, line 40: i'm curious why the test above doesn't fail since you don't explicitly stub out `deduction_method_is_standard` to equal true.

arinchoi03: I'm guessing the fixture that is used to calculate (`calculate_deduction_method`) whether it's standard/nonstandard returns `S`. Maybe we should be explicit and stub it out in the block above to be super clear?

arinchoi03 (approved):
🚀 just one suggestion from me & Anisha but looks great

PR #5251, opened 2024-12-19
FYST-541 update MD502 to 2024

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-541

## What was done?
- Replaced 2023 PDF with 2024 and updated the names of the fields in our PDF filler
- Had to do some editing of the PDF itself in acrobat because some checkboxes were text fields, some fields had the same name (not allowed), some fields were named "undefined" etc. Tbh we did leave some of the stupid field names alone because we didn't want to edit the whole dang PDF
## How to test?
- Pdf filler test should at least catch cases where a text value is wrong and check that we used existing field names
- The one thing it doesn't catch is if you try to fill in an invalid checkbox value i.e. a lot of them used to be "Yes" but in the new pdf they are "On" and we had to update the test to check that it's "On", it did not error when using "Yes" (even though that would not check the box)


anisharamnani (approved):
thanks @jenny-heath for taking this on <3

PR #5266, opened 2024-12-20
NJ 229 - Add static paid preparer fields to return header

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/229

## What was done?
- NJ Taxation is reporting MEF errors with all submissions since `PaidPreparerInformationGrp` is not present in the `StateReturnHeader` of XML submissions
- This change adds a static `PaidPreparerInformationGrp` block that corresponds to self prepared taxes.

## How to test?
- Unit Tests
- Run through `devito_single` persona on localhost/heroku, validate that XML contains `PaidPreparerInformationGrp`

- After
image

### DIFF
--- a/app/lib/submission_builder/return_header.rb
+++ b/app/lib/submission_builder/return_header.rb
@@ -12,9 +12,11 @@ module SubmissionBuilder
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
-        xml.PaidPreparerInformationGrp do
-          xml.PTIN "P99999999"
-          xml.PreparerPersonNm "Self Prepared"
+        if @submission.data_source.state_code.upcase == "NJ"
+          xml.PaidPreparerInformationGrp do 
+            xml.PTIN "P99999999"
+            xml.PreparerPersonNm "Self Prepared"
+          end
         end
         xml.DisasterReliefTxt @intake.disaster_relief_county if @intake.respond_to?(:disaster_relief_county)
         xml.OriginatorGrp do
--- a/app/lib/submission_builder/return_header.rb
+++ b/app/lib/submission_builder/return_header.rb
@@ -12,6 +12,10 @@ module SubmissionBuilder
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
+        xml.PaidPreparerInformationGrp do
+          xml.PTIN "P99999999"
+          xml.PreparerPersonNm "Self Prepared"
+        end
         xml.DisasterReliefTxt @intake.disaster_relief_county if @intake.respond_to?(:disaster_relief_county)
         xml.OriginatorGrp do
           xml.EFIN EnvironmentCredentials.irs(:efin)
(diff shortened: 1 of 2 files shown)


app/lib/submission_builder/return_header.rb
```diff
@@ -12,6 +12,10 @@ def document
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
+        xml.PaidPreparerInformationGrp do
```
jachan, line 15: Lmk if we should gate this behind a `if @submission.data_source.state_code_nj?` or similar

app/lib/submission_builder/return_header.rb
```diff
@@ -33,7 +37,7 @@ def document
             xml.DateSigned date_type_for_timezone(@submission.data_source.primary_esigned_at)&.strftime("%F") if @submission.data_source.ask_for_signature_pin?
             xml.USPhone @submission.data_source.direct_file_data.phone_number if @submission.data_source.direct_file_data.phone_number.present?
           end
-          if @submission.data_source&.spouse.ssn.present? && @submission.data_source&.spouse.first_name.present? && !@intake.filing_status_mfs?
+          if @submission.data_source&.spouse&.ssn.present? && @submission.data_source&.spouse&.first_name.present? && !@intake.filing_status_mfs?
```
jachan, line 40: Rubocop autoformat

app/lib/submission_builder/return_header.rb
amanifarooque: it seems safer to gate it, if there's a precedent/model for doing so -- how difficult would that be?

jachan: Just added conditional logic! Will pull in a reviewer from CfA to see if this approach is okay.

spec/lib/submission_builder/ty2022/states/return_header_spec.rb
```diff
@@ -42,6 +42,25 @@
         end
       end
 
+      context "paid preparer information group" do
+        if state_code == "nj"
```
mmazanec22, line 46: [pebble] Could use the existing schema required mechanism instead

mmazanec22 (approved):
Looks good

app/lib/submission_builder/return_header.rb
mpidcock: We strongly prefer not to use state codes in conditional branches if it can be avoided. Another different approach could be to:
* define methods for `ptin` and `preparer_person_name` in `nj_return_xml.rb` that return these hardcoded values
* define a default implementation in `document.rb` that returns nil
* check for the existence of the fields before adding them

If you need to rush this for testing though, this can be a follow up ticket

app/lib/submission_builder/return_header.rb
```diff
@@ -12,6 +12,12 @@ def document
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
+        if @submission.data_source.state_code.upcase == "NJ"
```
jnf, line 15: [sand] if we don't have a better way of disambiguating state-specific blocks, we should make one.

jnf (approved):
i wanna chat about how to reliably, consistently manage state-specific concerns in these shared builders, but that's a conversation larger than this pr.

app/lib/submission_builder/return_header.rb
jachan: Appreciate the guidance on an alternate approach. Agree that that'd be less brittle than relying on `state_code`. After talking with our PM, we'd like to merge this for vendor testing next week.

We'll refactor with this ticket, prioritized for next week: https://github.com/newjersey/affordability-pm/issues/230#issue-2753443632

spec/lib/submission_builder/ty2022/states/return_header_spec.rb
jachan: I'm so sorry that I missed this comment! Added it in the follow up ticket that I will take a look at next week.

https://github.com/newjersey/affordability-pm/issues/230

PR #5268, opened 2024-12-20
Nj 146 detail review

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/146

## What was done?
Add calculated fields to review screen

## How to test?
Pick a persona and walk through to the review screen, toggle the reveal

image

### DIFF
--- a/app/controllers/state_file/questions/nj_review_controller.rb
+++ b/app/controllers/state_file/questions/nj_review_controller.rb
@@ -12,7 +12,7 @@ module StateFile
             { text_key: '27_total_income', value: line_or_zero(:NJ1040_LINE_27) },
             # { text_key: '28c_retirement_excluded_from_taxation', value: line_or_zero(:NJ1040_LINE_28C) },
             { text_key: '29_nj_gross_income', value: line_or_zero(:NJ1040_LINE_29) },
-            # { text_key: '30_exemptions', value: line_or_zero(:NJ1040_LINE_30) },
+            { text_key: '30_exemptions', value: line_or_zero(:NJ1040_LINE_13) }, # same value for 13 and 30
             { text_key: '31_medical', value: line_or_zero(:NJ1040_LINE_31) },
             { text_key: '41_property', value: line_or_zero(:NJ1040_LINE_41) },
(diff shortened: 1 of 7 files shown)


spec/features/state_file/complete_intake_spec.rb
```diff
@@ -732,7 +732,7 @@
 
       groups = page.all(:css, '.white-group').count
       h2s = page.all(:css, 'h2').count
-      expect(groups).to eq(h2s)
+      expect(groups).to eq(h2s - 1) # account for the h2 reveal button
```
jachan, line 735: [DUST] is there a test we can add here to validate that the detailed review section contains all the different fields we want it to?

app/controllers/state_file/questions/nj_review_controller.rb
```diff
@@ -1,6 +1,44 @@
 module StateFile
   module Questions
     class NjReviewController < BaseReviewController
+      before_action :get_detailed_return_info
+
+      def get_detailed_return_info
```
jachan, line 6: [DUST] Is there a way to pull the XML values rather than the calculated values? This is something that CfA made a point of during our pairing sessions. Since XML is what we actually submit, we'd ideally use those rather than calculator values.

jachan (approved):
Looks good. Two non-blocking comments to consider.

app/controllers/state_file/questions/nj_review_controller.rb
mrotondo: Not sure who you paired with John, and I may not be aware of a planned refactor, but the other "detail" sections for (at least some of) the other states pull from the calculator as well.

mrotondo: Also, I appreciate this refactor from explicitly referencing each value in the view.

mrotondo (approved):

app/controllers/state_file/questions/nj_review_controller.rb
jachan: The stuff I paired on was specifically the PDF generation, so good to know that the review screen pulls from the calculator!

PR #5274, opened 2024-12-23
Nj 210 qss

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/210

## What was done?
Copy NC QSS question; replace QSS intake testing profile with realistic Sinatra profile

## How to test?
Use Sinatra QSS profile

image


jnf (approved):
nice!

spec/fixtures/state_file/fed_return_xmls/nj/qualified_widow.xml
jachan, line 1: [DUST] Great to remove these files after adding personas in!

spec/controllers/state_file/questions/nj_year_of_death_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        expect(described_class).not_to be_show(intake)
+      end
+    end
+  end
+
+  describe "#update" do
+    let(:form_params) do
+      {
+        state_file_nj_year_of_death_form: {
+          spouse_death_year: 2023
```
mluedke2, line 34: [dust] will eventually have to update this test when 2023 is no longer a valid choice... but will be quick and easy and in two years from now

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -211,14 +211,13 @@
     end
 
     context "qualifying widow/er filers" do
-      let(:intake) { create(:state_file_nj_intake, filing_status: "qualifying_widow") }
+      let(:intake) { create(:state_file_nj_intake, :df_data_qss) }
```
mluedke2, line 214: [dust] - is this change needed? importing a whole profile is more test setup than just setting the `filing_status` so i'd think we'd want the less-setup route... unless I'm missing something?

jachan (approved):
Looks good!

app/controllers/state_file/questions/nj_year_of_death_controller.rb
```diff
@@ -0,0 +1,10 @@
+module StateFile
+  class Questions::NjYearOfDeathController < Questions::QuestionsController
+    include ReturnToReviewConcern
+
+    def self.show?(intake)
+      intake.filing_status_qw?
+    end
+
+  end
+end
```
mluedke2, line 10: [dust] - this is only 10 lines, but the tests are 87 lines, and it's almost identical to the NC version... is there a way we can reuse the code for our purposes to be more DRY? It's not so much code that I'd consider it to be a boulder, but thinking it'll likely be easier to maintain (and could always break back out if functionality diverges in the future). Thoughts?

mmazanec22: I generally agree that it feels silly to have these two identical files, but the two things holding me back from DRYing it out in this case are 1) I generally don't think a refactor is worth the effort until the third repetition, especially for something with low complexity like this 2) this PR is necessary for accurate returns and it'll be harder to get PRs through for the next week

PR #5275, opened 2024-12-23
Authentication header xsd validations

embarnard:
## Link to pivotal/JIRA issue
- e.g. [link](https://codeforamerica.atlassian.net/browse/FYST-1304?atlOrigin=eyJpIjoiY2IxMWY1NzAyZmRjNGFkNmI5NzhhNzlhZGNlYTVhZDYiLCJwIjoiaiJ9)

## What was done?
- Added validation to id number on state id form to check for strings between 1-40 characters
- check for device id to match pattern, if not we send the default
- Also found a bug where it wouldn't load the follow-up if you selected "Driver license" and reloaded the page or had a validation issue upon submit. You had to re-select the radio button for it to show. check that when you select the "driver's license" option it will still load the followup with validation errors on submit or page reload
## How to test?
- Check if you can submit a id number more than 40 characters
- select driver license option with validation issues and continue to next page, this change is working if it reloads with the follow up reveals
Added validation to id field
Screenshot 2024-12-23 at 12 39 55 PM

This is how I reviewed the XML field validations for auth header:
Screenshot 2024-12-23 at 7 12 07 PM


app/javascript/lib/honeycrisp.js
```diff
@@ -103,13 +103,17 @@ var followUpQuestion = (function () {
 
                 // set initial state of follow-ups based on the page
                 $(this).find('input').each(function (index, input) {
-                    if ($(this).attr('data-follow-up') != null) {
-                        if ($(this).is(':checked')) {
-                            $($(this).attr('data-follow-up')).find('input, select').attr('disabled', false);
-                            $($(this).attr('data-follow-up')).show();
+                    var $input = $(this);
```
embarnard, line 106: The question-follow-up pattern only assumes that there is one option with the show follow-up option. 
So if two or more radio buttons have data-follow-up="#id-details-fields", the final iteration might be one that is not checked—so it will hide #id-details-fields. This change makes it so that we will load the follow-up if any of the options are checked, not just the last one.

app/lib/submission_builder/authentication_header.rb
```diff
@@ -42,7 +42,7 @@ def document
               xml.IPv6AddressTxt device_info&.ip_address if device_info&.ip_address&.ipv6?
             end
             xml.IPTs datetime_type(device_info&.updated_at)
-            xml.DeviceId device_info&.device_id || 'AB' * 20
+            xml.DeviceId /^[A-F0-9]{40}$/.match?(device_info&.device_id) ? device_info&.device_id : 'AB' * 20
```
anisharamnani, line 45: can we write a test for this?

app/forms/state_file/state_id_concern.rb
```diff
@@ -28,8 +28,7 @@ module StateIdConcern
       validates :id_type, presence: true
       validates :id_number,
                 presence: true,
-                # Seems to just be for New York
-                # alphanumeric: true, length: {is: 9},
```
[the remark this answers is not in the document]
embarnard: @mpidcock not sure why it was commented out but technically the xml field according to the XSD validation definition accepts characters that are not just alphanumeric. I can add it in if you are sure all states ID number strings are just alphanumeric?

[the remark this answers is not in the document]
mpidcock: Thanks for checking! I guess we should leave it out, I'm just surprised.

app/javascript/lib/honeycrisp.js
anisharamnani: @mrotondo and i paired on reviewing this. 

@embarnard, oof good catch on this bug! this is quite hairy code, so thank you for updating it. 

do you mind adding a comment at line 115 just to explain that two questions can refer to the same followup? since this is so tricky, a comment would help. 

for future notice, @mrotondo and i spoke about moving this logic to the update command, but that doesn't need to be included in this PR.

anisharamnani (approved):
i made two comments--1 to add a test & the other to add a comment to honeycrisp.js. neither of them are blocking. thank you @embarnard!

app/javascript/lib/honeycrisp.js
embarnard: definitely! Thank you for reviewing it so thoroughly

app/lib/submission_builder/authentication_header.rb
embarnard: @anisharamnani done! thanks for adding that, it actually helped me catch that I hadn't added it to the initial device info block so glad you pointed that out

PR #5278, opened 2024-12-23
Remove unnecessary attempt to set attrs that wasn't doing anything

arinchoi03:
## Link to pivotal/JIRA issue
- N/A offshoot from conversation while pairing with @jachan https://github.com/codeforamerica/vita-min/pull/5280

## What was done?
- Removed some set up code that wasn't doing anything that was highlighted by John
- https://github.com/codeforamerica/vita-min/pull/5264#discussion_r1896003586
> re: discussion on transient

> https://stackoverflow.com/a/41626919
> I think we're using transient wrong here
>
> transient attributes allow you to pass in data that isn’t an attribute on the model.

See more https://thoughtbot.github.io/factory_bot/ref/transient.html

All these traits (except for filing_status which makes sense here) should not be in this transient block.

## How to test?
  - all tests are still passing
- Risk Assessment
  - I don't think so since that was just test set up


jachan (approved):
Thank you for follow up on this!

PR #5280, opened 2024-12-23
nj 203 - added signature date to pdf

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/203#issue-2738881181

## What was done?
- Paired Arin from CfA, thanks @arinchoi03 for the help!
- NJ does not plan to use PIN for FY2024, but wants Signature Date to be filled in XML and PDF
- Make `primary_signature_pin` and `spouse_signature_pin` ignored so the columns can be dropped in the future

## How to test?
- Validated using `zeus_one_dep` (married) and `devito_single` (single), clicking through the flow, and checking the following:
- Signature PIN is no longer required
- XML contains Esign Date on under `ReturnHeaderState > Filer > Primary > Date Signed`
- PDF contains Signature Dates on Page 4.

- Single Filer
image
image

- Married Filer
image
image

### DIFF
--- a/app/models/state_file_nj_intake.rb
+++ b/app/models/state_file_nj_intake.rb
@@ -60,7 +60,6 @@
 #  primary_last_name                                      :string
 #  primary_middle_initial                                 :string
 #  primary_signature                                      :string
-#  primary_signature_pin                                  :text
 #  primary_ssn                                            :string
 #  primary_suffix                                         :string
 #  primary_veteran                                        :integer          default("unfilled"), not null
@@ -84,7 +83,6 @@
 #  spouse_first_name                                      :string
 #  spouse_last_name                                       :string
 #  spouse_middle_initial                                  :string
-#  spouse_signature_pin                                   :text
 #  spouse_ssn                                             :string
 #  spouse_suffix                                          :string
 #  spouse_veteran                                         :integer          default("unfilled"), not null
@@ -113,6 +111,7 @@
 #  index_state_file_nj_intakes_on_spouse_state_id_id   (spouse_state_id_id)
 #
 class StateFileNjIntake < StateFileBaseIntake
+  self.ignored_columns += ["primary_signature_pin", "spouse_signature_pin"]
   encrypts :account_number, :routing_number, :raw_direct_file_data, :raw_direct_file_intake_data
 
   enum household_rent_own: { unfilled: 0, rent: 1, own: 2, neither: 3, both: 4 }, _prefix: :household_rent_own
@@ -202,7 +201,8 @@ class StateFileNjIntake < StateFileBaseIntake
   def validate_state_specific_w2_requirements(w2); end
 
   def ask_for_signature_pin?
-    true
+    false
   end
+
 end
 
--- a/spec/factories/state_file_md_intakes.rb
+++ b/spec/factories/state_file_md_intakes.rb
@@ -112,7 +112,7 @@ FactoryBot.define do
     end
 
     factory :state_file_md_refund_intake do
-      after(:build) do |intake, evaluator|
+      after(:build) do |intake|
         intake.direct_file_data.fed_agi = 10000
         intake.raw_direct_file_data = intake.direct_file_data.to_s
         intake.payment_or_deposit_type = "direct_deposit"
@@ -130,8 +130,6 @@ FactoryBot.define do
     primary_middle_initial { "A" }
     primary_last_name { "Lando" }
     primary_birth_date { Date.new(1950, 01, 01) } # matches the bday in md_minimal.json
-    primary_signature_pin { '12345' }
-    primary_esigned_at { DateTime.now }
     subdivision_code { "0111" }
     political_subdivision { "Mt Savage" }
     confirmed_permanent_address { "yes" }
@@ -193,8 +191,6 @@ FactoryBot.define do
       spouse_middle_initial { "B" }
       spouse_last_name { "Lando" }
       spouse_birth_date { MultiTenantService.statefile.end_of_current_tax_year - 40 }
-      spouse_signature_pin { '54321' }
-      spouse_esigned_at { DateTime.now }
     end
 
     trait :with_senior_spouse do
--- a/spec/lib/pdf_filler/nc_d400_pdf_spec.rb
+++ b/spec/lib/pdf_filler/nc_d400_pdf_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe PdfFiller::NcD400Pdf do
 
     context "pulling fields from xml" do
       let(:tomorrow_midnight) { DateTime.tomorrow.beginning_of_day }
-      let(:mailing_apartment) { 'Apt 2B'}
+      let(:mailing_apartment) { 'Apt 2'}
       let(:intake) {
         create(:state_file_nc_intake,
                filing_status: "single",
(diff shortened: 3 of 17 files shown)


app/models/state_file_nj_intake.rb
```diff
@@ -202,7 +201,8 @@ def disqualifying_eligibility_rules
   def validate_state_specific_w2_requirements(w2); end
 
   def ask_for_signature_pin?
-    true
+    false
```
arinchoi03, line 204: [dust] this is false on the base intake, so you shouldn't need to define it

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
```diff
@@ -43,36 +48,36 @@
...
           expect(pdf_fields['y_d400wf_fstat1']).to eq 'Yes'
           expect(pdf_fields['y_d400wf_fstat2']).to eq 'Off'
           expect(pdf_fields['y_d400wf_fstat3']).to eq 'Off'
           expect(pdf_fields['y_d400wf_fstat4']).to eq 'Off'
           expect(pdf_fields['y_d400wf_fstat5']).to eq 'Off'
 
-          expect(pdf_fields['y_d400wf_li6_good']).to eq '9000'
-          expect(pdf_fields['y_d400wf_li8_good']).to eq '9000'
+          expect(pdf_fields['y_d400wf_li6_good']).to eq '76902'
+          expect(pdf_fields['y_d400wf_li8_good']).to eq '76902'
```
arinchoi03, line 65: notes for self: both of these fields equal federal agi
`      <AdjustedGrossIncomeAmt>76902</AdjustedGrossIncomeAmt>`

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
```diff
@@ -43,36 +48,36 @@
...
           expect(pdf_fields['y_d400wf_li11_page1_good']).to eq '12750'
 
           expect(pdf_fields['y_d400wf_lname2_PG2']).to eq 'Carolinian'
-          expect(pdf_fields['y_d400wf_li20a_pg2_good']).to eq '15'
-          expect(pdf_fields['y_d400wf_li23_pg2_good']).to eq '15'
-          expect(pdf_fields['y_d400wf_li25_pg2_good']).to eq '15'
-          expect(pdf_fields['y_d400wf_dayphone']).to eq '9845559876'
+          expect(pdf_fields['y_d400wf_li20a_pg2_good']).to eq '4394'
+          expect(pdf_fields['y_d400wf_li23_pg2_good']).to eq '4394'
+          expect(pdf_fields['y_d400wf_li25_pg2_good']).to eq '4394'
```
arinchoi03, line 72: notes for self: line 20a= sum of all state income tax amount, which this fixture only has one of `<StateIncomeTaxAmt>4394</StateIncomeTaxAmt>`
line 23 =line_or_zero(:NCD400_LINE_20A) + line_or_zero(:NCD400_LINE_20B)
line 25 = same as line 23 (line 24 OOS)

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
```diff
@@ -43,36 +48,36 @@
...
           expect(pdf_fields['y_d400wf_Consumer Use Tax']).to eq 'Yes'
           expect(pdf_fields['y_d400wf_li18_pg2_good']).to eq '0'
-          expect(pdf_fields['y_d400wf_li19_pg2_good']).to eq '0'
+          expect(pdf_fields['y_d400wf_li19_pg2_good']).to eq '2887'
           expect(pdf_fields['y_d400wf_li26a_pg2_good']).to eq ''
           expect(pdf_fields['y_d400wf_li27_pg2_good']).to eq '0'
-          expect(pdf_fields['y_d400wf_li28_pg2_good']).to eq '15'
-          expect(pdf_fields['y_d400wf_li34_pg2_good']).to eq '15'
+          expect(pdf_fields['y_d400wf_li28_pg2_good']).to eq '1507'
+          expect(pdf_fields['y_d400wf_li34_pg2_good']).to eq '1507'
```
arinchoi03, line 80: notes for self: line 25 - line 19 ✅

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
```diff
@@ -133,7 +139,7 @@
           expect(pdf_fields['y_d400wf_fstat4']).to eq 'Off'
           expect(pdf_fields['y_d400wf_fstat5']).to eq 'Off'
 
-          expect(pdf_fields['y_d400wf_li20b_pg2_good']).to eq '15'
+          expect(pdf_fields['y_d400wf_li20b_pg2_good']).to eq '4394'
```
arinchoi03, line 142: notes for self: the w2 was modified to match spouse ssn in the `before` block

arinchoi03 (approved):
looks good to me! I'm cc'ing @mpidcock in case she wants to take a look, although we didn't really end up implementing any big cross-states changes

spec/factories/state_file_md_intakes.rb
```diff
@@ -130,8 +130,6 @@
     primary_middle_initial { "A" }
     primary_last_name { "Lando" }
     primary_birth_date { Date.new(1950, 01, 01) } # matches the bday in md_minimal.json
-    primary_signature_pin { '12345' }
-    primary_esigned_at { DateTime.now }
```
arinchoi03, line 134: moved setting of the signature-related attributes in the actual specs that need them, instead of treating them as a base-line state. We'd ideally would like to have intakes that start with no signatures, since for the most of our in-app flow the app is not signed

app/models/state_file_nj_intake.rb
jachan: Yep, we just thought it'd be clearer to define it again explicitly here.

jachan:
Perfect, I'll give Martha the rest of the day in case she wants to take a look, and merge tomorrow morning if there are no blocking comments.

arinchoi03:
also just noting this PR originated from https://github.com/codeforamerica/vita-min/pull/5264 for historical link

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
```diff
@@ -29,6 +30,10 @@
                primary_esigned_at: tomorrow_midnight)
       }
 
+      before do
+        intake.direct_file_data.mailing_apartment = mailing_apartment
+      end
```
mpidcock, line 35: [dust] I won't block on this, but why was this added? is it related to changing the state_file_nc_intake factory? it makes it seem like there is something special related to the mailing_apartment in this test, which there shouldn't be. 

also, the test shows `expect(pdf_fields['y_d400wf_apartment number']).to eq 'Apt 2'` - why not 'Apt 2B'?

mpidcock (approved):
I appreciate the thoughtful update to the esigned fields in the tests! Just a bit confused about what's going on with the mailing_apartment

spec/lib/pdf_filler/nc_d400_pdf_spec.rb
arinchoi03: hmm yeah i can go either way here -- I guess I was worried I was losing the apartment number part of the test as I changed the fixture used in this story. `nc_nick` had a apartment address but `nc_daffy_single` doesn't. Maybe I should break this out into its own context if I want to make sure to test it?

arinchoi03: @mpidcock i think you uncovered a bug here, I wasn't thinking when I made these tests but looks like NCD400 Apartment number has a limit of 5 characters, so if we try to set it to "Apt 2B" it only is able to set it to "Apt 2". Not sure how important it is but since the PDF looks like it has "Apartment Number" over the input field, wondering if we should try to remove any `COMMON_ADDRESS_ABBREV`?

arinchoi03: For this test I'll update the test to fit the box, so it doesn't lead to confusion

PR #5281, opened 2024-12-23
Add financial transaction to NJ return xml

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/232

## What was done?
Add the financial transaction element to NJ return xml

## How to test?
Generate any return, and add acct and routing number in the final screens

### DIFF
--- a/app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
@@ -26,6 +26,12 @@ module SubmissionBuilder
             "NJIndividual2024V0.1"
           end
 
+          def build_state_specific_tags(document)
+            if !@submission.data_source.routing_number.nil? && !@submission.data_source.account_number.nil?
+              document.at("ReturnState").add_child(financial_transaction)
+            end
+          end
+
           def documents_wrapper
             nil
           end
@@ -102,7 +108,20 @@ module SubmissionBuilder
           end
 
           def calculator 
-            @submission.data_source.tax_calculator
+            calculator = @submission.data_source.tax_calculator
+            calculator.calculate
+            calculator
+          end
+
+          def financial_transaction
+            calculator = @submission.data_source.tax_calculator
+            calculator.calculate
+
+            FinancialTransaction.build(
+              @submission,
+              validate: false,
+              kwargs: { refund_amount: calculator.refund_or_owed_amount }
+            ).document.at("*")
           end
         end
       end
--- a/app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
@@ -59,6 +59,7 @@ module SubmissionBuilder
           def nj_2450s
             docs = []
 
+            # Must call calculator directly because these aren't lines on the NJ 1040
             if calculator.line_59_primary&.positive? || calculator.line_61_primary&.positive?
               docs << {
                 xml: SubmissionBuilder::Ty2024::States::Nj::Documents::Nj2450,
@@ -83,7 +84,7 @@ module SubmissionBuilder
           def schedule_nj_hcc
             docs = []
 
-            if calculator.line_53c_checkbox
+            if calculated_fields.fetch(:NJ1040_LINE_53C_CHECKBOX)
               docs << {
                 xml: SubmissionBuilder::Ty2024::States::Nj::Documents::ScheduleNjHcc,
                 pdf: PdfFiller::ScheduleNjHccPdf,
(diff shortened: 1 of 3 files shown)


app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
```diff
@@ -102,7 +108,20 @@ def additional_dependents
...
           def calculator 
-            @submission.data_source.tax_calculator
+            calculator = @submission.data_source.tax_calculator
+            calculator.calculate
+            calculator
+          end
+
+          def financial_transaction
+            calculator = @submission.data_source.tax_calculator
+            calculator.calculate
```
mpidcock, line 118: [pebble] I know this was copied from elsewhere, but rather than call `.calculate` here, it would be better to follow the pattern from AZ and use [calculated_fields](https://github.com/codeforamerica/vita-min/blob/e9eca3906046dd89e1d65e7585daa2e99081d9ca/app/lib/submission_builder/ty2022/states/az/az_return_xml.rb#L223-L225) to reduce the calls to `.calculate`. 

This should be updated in ID, MD, and NC - we could update NJ at the same time we do the others if you'd rather merge and move on.

spec/factories/state_file_nj_intakes.rb
```diff
@@ -268,6 +268,17 @@
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_zeus_box_14') }
     end
 
+    factory :state_file_nj_refund_intake do
```
mpidcock, line 271: [dust] perhaps rename this factory - it's not a refund intake specifically, and the test that uses it configures the `refund_or_owed_amount` in the test.

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -24,6 +24,16 @@
         expect(xml.document.at('ReturnDataState FormNJ1040 Header')).to be_an_instance_of Nokogiri::XML::Element
       end
 
+      context "when there is a refund with banking info" do
+        let(:intake) { create(:state_file_nj_refund_intake) }
+  
+        it "generates FinancialTransaction xml without RefundAmt" do
```
mpidcock, line 30: [dust] what does "without RefundAmt" refer to here?

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -24,6 +24,16 @@
         expect(xml.document.at('ReturnDataState FormNJ1040 Header')).to be_an_instance_of Nokogiri::XML::Element
       end
 
+      context "when there is a refund with banking info" do
```
mpidcock, line 27: [dust] consider testing the owed case as well

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
```diff
@@ -24,6 +24,16 @@
...
+        let(:intake) { create(:state_file_nj_refund_intake) }
+  
+        it "generates FinancialTransaction xml without RefundAmt" do
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:refund_or_owed_amount).and_return 500
+          xml = Nokogiri::XML::Document.parse(described_class.build(submission).document.to_xml)
+          expect(xml.at("FinancialTransaction")).to be_present
```
jachan, line 33: [Pebble] Can we also assert on that the children of this element exist and are correct?

RefundDirectDeposit exists
RoutingTransitNumber matches the factory value
BankAccountNumber matches the factory value
Amount matches the mocked value on line 31
Savings is checked based on factory
NotIATTransaction is set correctly (not sure how that logic works)

app/lib/submission_builder/ty2024/states/nj/nj_return_xml.rb
```diff
@@ -26,6 +26,12 @@ def state_schema_version
             "NJIndividual2024V0.1"
           end
 
+          def build_state_specific_tags(document)
```
jachan, line 29: [dust] great pattern introduced here, might be reusable for https://github.com/newjersey/affordability-pm/issues/230

jachan (commented):
Great work here - one requested addition to the tests

spec/lib/submission_builder/ty2024/states/nj/nj_return_xml_spec.rb
mmazanec22: That's covered in financial transaction spec

jachan: Makes sense. Feel free to disregard and resolve then.

mluedke2 (commented):
👍  learning from this - thanks for figuring this out!

PR #5284, opened 2024-12-24
Update MD502B pdf fieldnames - clashing with other PDF field name in bundle

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1498
## Is PM acceptance required? (delete one)
- No - merge after code review approval

**Reminder**: merge main into this branch and get green tests before merging to main
## What was done?
- Noticed [test failing on `main`](https://app.circleci.com/jobs/github/codeforamerica/vita-min/76642)
- The "1" field kept getting set to `"0"` which was perplexing as `hash_for_pdf` method in the `Md502bPdf` had a value of `"2"` when I put a byebug in.
- Noticed that https://github.com/codeforamerica/vita-min/pull/5251 was merged in which _also_ has a fieldname of `"1"` (name clash) which made it hard for ` let(:pdf_fields) { filled_in_values(submission.generate_filing_pdf.path) }` to return the correct value we were looking for, since it looks at the entire PDF, not just MD502
- Updated the PDF to have more specific names for the fields
## How to test?
  - See the MD502B tests pass
## Screenshots (for visual changes)
Still filling the form correctly
Screenshot 2024-12-23 at 8 49 58 PM

PR #5286, opened 2024-12-24
Fix mdEL101 spec failure in the evening again

arinchoi03:
[It's...back again:
](https://app.circleci.com/pipelines/github/codeforamerica/vita-min/27794/workflows/248ada5c-21b2-4bf6-a5e2-4fded9b403e5/jobs/76659/tests
)
[See that it wasn't an issue (this PR didn't have anything to do with MD EL101) but it starts being an issue ~9PM PT
](https://app.circleci.com/pipelines/github/codeforamerica/vita-min?branch=FYST-1469-update-misc-nc-copy-for-translations)
Taking a page out of NJ playbook to set any timestamp we manually set to a noon at UTC timestamp so that it will always be converted to the same day. [See q & a here](https://github.com/codeforamerica/vita-min/pull/5264#discussion_r1895942471)

## Link to pivotal/JIRA issue
- N/A

## What was done?
- Setting the signed_at timestamp to noon time will ensure it will always return the same Date because the UTC/other timezone difference will be within the boundaries of changing days (~7 hours difference between UTC / Arizona time zone)
## How to test?
- Run tests at all hours & make sure it does not fail again
- Risk Assessment
  - Not any really -- just trying to get tests to consistently pass regardless of when its run


arinchoi03:
note the MDEL101 spec passes but Md502b pdf spec fails (addressed [in another PR](https://github.com/codeforamerica/vita-min/pull/5284) 😓 ) 
https://app.circleci.com/pipelines/github/codeforamerica/vita-min/27796/workflows/f2c80c3a-5ab0-490a-a778-900b2c09a864/jobs/76667/tests

spec/lib/pdf_filler/md_el101_pdf_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:intake) { create(:state_file_md_intake) }
   let(:submission) { create :efile_submission, tax_return: nil, data_source: intake }
   let(:pdf) { described_class.new(submission) }
-  let(:expected_signature_date) { DateTime.now.in_time_zone(StateFile::StateInformationService.timezone('md')).strftime("%F") }
+  let(:expected_signature_date) { DateTime.new(2024, 12, 19, 12).strftime("%F") }
```
tahsinaislam, line 10: i feel like we still need to use in_time_zone(StateFile::StateInformationService.timezone('md') on the esigned at because its stored in UTC time and then we sign in MD timezone. id40 pdf has a good example i think

arinchoi03: hmm ok, i can add the timezone. will probably also add to the two line changes below?

arinchoi03: this approach is just a little different b/c it sets UTC 12/19/2024 12PM as the time, which should translate to "12-19-2024" which ever timezone we're in. I think this approach as-is still makes sense, but happy to chat about it more. timezone stuff is so confusing to me

tahsinaislam: ohhh noon time does make sense, but would it be better to do something like let(:today_noon) { Time.current.noon } instead of hardcoding the date

arinchoi03:
will close in favor of mike's PR coming soon

spec/lib/pdf_filler/md_el101_pdf_spec.rb
arinchoi03: @tahsinaislam just flagging here for you this test failed b/c i had set the signed_at timestamp to 1.hour.ago so there was always a time span where tests would break 😅 anyways see the most recent PR for the way we want to test timestamps going forward!

PR #5288, opened 2024-12-24
Nj 125 income editing

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/125

## What was done?
- Adapted to the new XML schema for `UI/WF/SWF` to `UIWFSWF` and `UIHCWD`
- An overwrite method is now in the W2 model to capture intended behavior that combines `UIHCWD` into `UIWFSWF` if `UIWFSWF` is empty... and downstream calculators and XML for NJ1040 and NJ2450 also needed adapting for this.
- The W2 model also has validation checks for the limits for those specific types of income. The validation has to be conditional to allow for importing any info, but then turned on by the Income Review and W2 Editing controllers to not allow users to proceed without fixing.
- I decided to reuse the existing income and W2 editing controllers and forms, and put in checks for the existence of info about a given state not taxing a type of income.
- There weren't E2E tests yet specifically for just income review so I added NJ-specific tests there for this ticket's AC
## How to test?
- Use Lucky to see W2 Box 14 editing, and O'Neal Walker for seeing messaging for other income forms.

image

image

after trying to save values above the limits
image

image

image


config/locales/en.yml
```diff
@@ -3754,14 +3755,14 @@ en:
           verification_code_label: Enter the 6-digit code
       w2:
         edit:
-          box14_explanation:
-            md_html: For tax year %{year}, we will only process the <strong>STPICKUP</strong> code, which represents any contributions you made to the state retirement system.
+          box14_explanation_md_html: For tax year %{year}, we will only process the <strong>STPICKUP</strong> code, which represents any contributions you made to the state retirement system.
```
mmazanec22, line 3758: Typo toward the end of the text - "those don''t need to be reported"

mluedke2: It looks wrong, but the double apostrophe `''` is to escape the `'` since the whole string is wrapped in single-quotes `'`. This was actually done automatically by the i18n-normalize script, and you can see in one of the screenshots that it renders as a single apostrophe.

app/views/state_file/questions/income_review/edit.html.erb
```diff
@@ -81,10 +85,15 @@
...
+    <div class="white-group" id="formssa1099s">
       <p class="text--bold spacing-below-0"><%= t(".ssa_title") %></p>
       <p class="text--grey-bold spacing-above-15 spacing-below-0">
-        <%= t(".no_info_needed") %>
+        <%= case current_state_code
+        when "nj"
+          t(".nj_not_taxed")
+        else
+          t(".no_info_needed")
+        end %>
```
arinchoi03, line 96: [dust] I think we generally have been trying to steer away from checking `current_state_code` to enter in state-specific content. Ways we've been trying to avoid it:
1. adding flags on the `StateFileBaseIntake` and `StateFile[state]Intake`  
2. creating partials
3. creating state-specific translations

For this above scenario, is it possible to do something like below? (from `app/views/state_file/questions/return_status/_accepted.html.erb`:

```
      <% # i18n-tasks-use t("state_file.questions.return_status.accepted.check_payable_html") # hint for the i18n linter that, yes, we are using this key (sometimes) %>
      <%= t(".#{current_state_code}.check_payable_html", default: :'.check_payable_html', payable_to: @department_of_taxation, filing_year: current_tax_year, form_number: @tax_form_number) %>
```

we have a `default` (`.no_info_needed`) here and state-specific copy (`.nj_not_taxed` -> `.nj.no_info_needed`)?

app/views/state_file/questions/w2/edit.html.erb
```diff
@@ -46,13 +49,17 @@
       <% end %>
     </div>
 
-    <% if t(".box14_explanation.#{@state_code}_html", default: nil) %>
+    <% explanation_key = ".box14_explanation_#{@state_code}_html" %>
+    <% button_text = @state_code == "nj" ? t(".what_is_box14_nj") : t(".what_is_box14") %>
```
arinchoi03, line 53: only two states support box 14, is there a way to make this dynamic?
`    <% button_text =  t(".what_is_box14_#{state_code}") %>`

arinchoi03 (commented):
Some questions/suggestions

app/services/state_file/state_information_service.rb
```diff
@@ -202,7 +202,7 @@ def state_intake_class_names
         vita_link_es: "",
         voucher_form_name: "NJ Voucher Form",
         voucher_path: "",
-        w2_supported_box14_codes: ["UI_WF_SWF", "UI_HC_WD", "FLI"],
+        w2_supported_box14_codes: ["UI_WF_SWF", "FLI"],
```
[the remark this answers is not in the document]
mluedke2: oh good catch, will do

app/views/state_file/questions/income_review/edit.html.erb
mluedke2: thanks -- that is a great tip about the linter hint; i definitely wrestled a bit with that on this. i'll update to use your suggestion to not check state code there...

app/views/state_file/questions/w2/edit.html.erb
mluedke2: sounds good, will do!

spec/features/state_file/income_review_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      # UIWFSWF 350
+      # UIHCWD 450
+      # FLI 550
+
+      expect(page).to have_field('state_file_w2_box14_ui_wf_swf', with: '350.0')
+      expect(page).to have_field('state_file_w2_box14_fli', with: '550.0')
+      expect(page).to have_field('state_file_w2_employer_state_id_num', with: '12345')
+      expect(page).to have_field('state_file_w2_state_wages_amount', with: '12345.0')
+      expect(page).to have_field('state_file_w2_state_income_tax_amount', with: '500.0')
+      expect(page).to have_button("What should I put in Box 14?")
```
jachan, line 57: [PEBBLE] Original tests in `review_page_spec` test that clicking continue takes you back to the review screen. Since we're removing NJ from those tests, mind putting that check here as well?

jachan (commented):
Great work! Had some small questions/comments!

spec/controllers/state_file/questions/w2_controller_spec.rb
```diff
@@ -198,4 +198,32 @@
...
+    context "when box14_ui_wf_swf is nil and box14_ui_hc_wd is not" do
+      let(:intake) { create :state_file_nj_intake }
+      let!(:state_file_w2) do
+        create :state_file_w2, state_file_intake: intake, w2_index: 1,
+               box14_ui_wf_swf: nil, box14_ui_hc_wd: 340
+      end
+  
+      it "sets box14_ui_wf_swf to box14_ui_hc_wd" do
+        get :edit, params: { id: state_file_w2.id }
+        expect(assigns(:w2).box14_ui_wf_swf).to eq 340
```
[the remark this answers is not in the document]
mluedke2: that's a good point -- the value of `box14_ui_hc_wd` doesn't end up mattering because it is just there as a backup in case `box14_ui_wf_swf` is empty, then isn't used anymore. but i think I still want an expectation of it's value for the same reason as you. so i'll consider setting it to `nil` after it's used and then put that in the expectation...

app/views/state_file/questions/income_review/edit.html.erb
```diff
@@ -27,7 +27,11 @@
   <% if current_intake.direct_file_data.fed_unemployment > 0 %>
     <div class="white-group unemployment" id="form1099gs">
       <p class="text--bold spacing-below-0"><%= t(".unemployment_title") %></p>
-      <% if current_intake.state_file1099_gs.count > 0 %>
+      <% if current_state_code == "nj" %>
```
mmazanec22, line 30: [dust] This is another place where you might refactor to remove the state code check

app/views/state_file/questions/w2/edit.html.erb
```diff
@@ -9,6 +9,9 @@
       <p class="spacing-below-25"><%= @w2.employer_name %></p>
 
       <% if @box14_codes.present? %>
+        <% if @state_code == "nj" %>
```
mmazanec22, line 12: [dust] State code check

config/locales/en.yml
```diff
@@ -3758,14 +3759,14 @@ en:
...
...d FLI.<br/><br/><strong>If you do not see these, please leave these fields blank.</strong><br/><br/><strong>Definitions:</strong><br/><br/><strong>UI/WF/SWF</strong> (may also appear as <strong>"UI/HC/WD"</strong> on your W2) represents:<ul><li><strong>UI</strong> refers to unemployment insurance contributions</li><li><strong>WF</strong> and <strong>WD</strong> refer to  workforce development partnership fund contributions</li><li><strong>SWF</strong> refers to supplemental workforce fund contributions</li><li><strong>HC</strong> refers to healthcare contributions</li><li><strong>WD</strong> refers to workforce development contributions</li></ul><strong>FLI</strong>: Represents contributions withheld for Family Leave Insurance. Family Leave Insurance (FLI) provides paid leave benefits to bond with a new child, care for a loved one, or cope with domestic/sexual violence. If you have a number for this on your paycheck, it means you have paid into this program and may be eligible to receive cash benefits.<br/><br/>We use the amounts to determine if you  might be eligible for a credit.<br/><br/>There may be other information in box 14 - those don''t need to be reported with this tool.'
```
mmazanec22, line 3763: [dust] Will taxpayers see UI/WF/SWF on their returns or UIWFSWF? Might want to update copy to reflect the most likely case

mmazanec22 (approved):

spec/controllers/state_file/questions/w2_controller_spec.rb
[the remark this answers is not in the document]
mluedke2: That's a good point-- `box14_ui_hc_wd` doesn't matter from here on out so I didn't bother adding an expectation for it. I will be making an update to the w2 model that will help make it clear that we aren't using that value...

spec/features/state_file/income_review_spec.rb
mluedke2: true - will make sure that everything that was tested before is still covered...

mluedke2:
Thanks all for the comments (and the discussion at our check-in). I'll be adding a couple error messages and making a change to the w2 model to make sure Line 59 doesn't get broken (I suspect it is actually broken right now and just wasn't caught by tests).

Since that will likely be a pretty big set of changes I'll re-tag people when ready for review again. But for now will go back to Draft while I'm working...

config/locales/en.yml
mluedke2: [Amani confirmed](https://github.com/newjersey/affordability-pm/issues/125#issuecomment-2565789608) that the slashes are there visually. The version without slashes seems to just be for XML

app/views/state_file/questions/w2/edit.html.erb
mluedke2: thanks -- fixed

app/views/state_file/questions/income_review/edit.html.erb
mluedke2: thanks! fixed

spec/controllers/state_file/questions/w2_controller_spec.rb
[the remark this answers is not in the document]
mluedke2: to resolve this, i introduced a function called `get_box14_ui_overwrite()` in `state_file_w2` so this logic is not in this controller anymore

app/views/state_file/questions/w2/edit.html.erb
mluedke2: made this dynamic (added `_md` to the existing string)

app/views/state_file/questions/income_review/edit.html.erb
mluedke2: used the linter hint and default setup to avoid checking for "nj" specifically. thanks!

app/services/state_file/state_information_service.rb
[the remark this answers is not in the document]
mluedke2: I actually decided to end up keeping this, because it all maps back to the DB. It could be renamed, but would require a db migration to rename the field there. (And there were about 70 uses of this throughout the app so I didn't want this diff to grow). And it is only important to not have underscores in the XML alone. If we still want to do that db field name change and related rename I'm not against it but a follow-up would be better in my opinion.

spec/features/state_file/income_review_spec.rb
mluedke2: i did this by putting the NJ option back into that existing test, but then in order to make it work I made part of it conditional

mluedke2:
@jachan @mmazanec22 this is ready for review again!

mluedke2:
[boulder] - i need to change how i am validating the limit on W2 Box 14 because as a validation on the model I am blocking even the import of invalid values, which causes a 500 on the Lucky profile since it has a too-high value... Will likely move the validation to just be on the w2 editing form step

mluedke2:
[boulder] - there's also a boulder on the XML export which is still exporting both UI values for box14 instead of consolidating them.

mluedke2:
the two boulders have been addressed!

anisharamnani (approved):
looks good, also tested MD box 14 to see if anything changed - tahsina

app/controllers/state_file/questions/w2_controller.rb
```diff
@@ -9,6 +9,7 @@ def edit
 
       def update
         @w2.assign_attributes(form_params)
+        @w2.check_box14_limits = true
```
mluedke2, line 12: note from review: can we turn on this flag in one place when the import is done?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -251,14 +251,12 @@ def line_53c_checkbox
...
-        get_personal_excess(@intake.primary.ssn, :box14_ui_wf_swf, EXCESS_UI_WF_SWF_MAX) +
-          get_personal_excess(@intake.primary.ssn, :box14_ui_hc_wd, EXCESS_UI_WF_SWF_MAX)
+        get_personal_excess(@intake.primary.ssn, :box14_ui_wf_swf, EXCESS_UI_WF_SWF_MAX)
       end
 
       def line_59_spouse
         if @intake.filing_status_mfj?
-          get_personal_excess(@intake.spouse.ssn, :box14_ui_wf_swf, EXCESS_UI_WF_SWF_MAX) +
-            get_personal_excess(@intake.spouse.ssn, :box14_ui_hc_wd, EXCESS_UI_WF_SWF_MAX)
+          get_personal_excess(@intake.spouse.ssn, :box14_ui_wf_swf, EXCESS_UI_WF_SWF_MAX)
```
mluedke2, line 259: review comment: instead of passing excess type, could just pass the value itself

app/models/state_file_w2.rb
```diff
@@ -113,4 +115,22 @@ def state_tax_group_xml_node
...
+    validate_limit(:box14_ui_wf_swf, 179.78)
+    validate_limit(:box14_fli, 145.26)
+  end
+
+  def validate_limit(field, limit)
+    value = send(field)
+    if value.present? && value > limit
+      errors.add(field, I18n.t("validators.dollar_limit", limit: '%.2f' % limit))
+    end
+  end
```
mluedke2, line 135: review comment: check if it would be cleaner to put this in the NJ-specific W2

app/views/state_file/questions/w2/edit.html.erb
```diff
@@ -9,14 +9,24 @@
...
         <fieldset>
           <legend class="spacing-below-5"><%= t(".box14_html") %></legend>
           <% @box14_codes.each do |code| %>
             <div class="form-question spacing-below-25">
               <% field_name = "box14_#{code.downcase}" %>
+              <% value = if code == "UIWFSWF"
+                @w2.get_box14_ui_overwrite
+              else
+                @w2.send(field_name)
+              end %>
```
mluedke2, line 25: review comment: could a method on the W2 model prevent us from having to do all these if-else to get these values?

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: @mmazanec22 actually, i looked at this again and because `get_excess_value()` is called within two different iterators (ie needs to look at several w2s), it actually makes that function not so easy to get rid of...

app/models/state_file_w2.rb
mluedke2: @mmazanec22 it looks like we don't actually have an NJ-specific W2 model yet. we could add one, but i think i will make a follow-up ticket for this

app/controllers/state_file/questions/w2_controller.rb
mluedke2: @mmazanec22 I did manage to try out an `after_create` callback on the W2 model to toggle `check_box14_limits` just after creation, which does allow import... but it's a bit of a challenge to get it to *stay* set, I think because the controllers re-fetch them from the database when the forms load. So I think I'd either need to

1- find a way to pass the w2 model instances from one place to another instead of loading from DB
2- add a db column for this and set it
3- keep how i have it, setting in each controller each time

I think the second option is good but is not so much better that I want to add it to this PR.

app/views/state_file/questions/w2/edit.html.erb
mluedke2: @mmazanec22 I want to keep this one open too, but since it's only used in a few spots (and in different ways) I think I want to leave as a follow-up

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: @mmazanec22  aha! I tried again and was able to make it work by figuring out lambda syntax better in this commit https://github.com/codeforamerica/vita-min/pull/5288/commits/f33f5e5abdfeae9f8365406426c9b7937551edf1

PR #5292, opened 2024-12-24
remove 1 hour gap that was breaking us in MD, and simplify our expectations about signature dates on pdfs everywhere

mrotondo:

## What was done?
- Removed all adjustments to signature dates in PDF tests
  - "1.hour.ago" in an MD test
  - "tomorrow.midnight" in a bunch of other tests
- This simplifies the test logic and avoids time-dependent failures
## How to test?
- We should make sure that all tests pass regardless of when they are run
- This is hard to verify on CircleCi, so we set the system time locally to different times near midnight and made sure the tests passed with e.g. `sudo date 1224200124 && bundle exec rspec spec/lib/pdf_filler/nc_d400_pdf_spec.rb`, where the command line argument to `date` is Month-Day-Hour-Minute-Year

### DIFF
--- a/spec/lib/pdf_filler/id40_pdf_spec.rb
+++ b/spec/lib/pdf_filler/id40_pdf_spec.rb
@@ -3,12 +3,13 @@ require 'rails_helper'
 RSpec.describe PdfFiller::Id40Pdf do
   include PdfSpecHelper
 
-  let(:tomorrow_midnight) { DateTime.tomorrow.beginning_of_day }
+  let(:signature_date) { DateTime.now }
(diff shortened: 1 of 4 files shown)


arinchoi03:
@jachan this is how we want to test signatures going forward, up to you if you want to refactor tests to be more exact instead of relying on staying within bounds of the time difference amount

jachan:
Appreciate the heads up, Arin and Mike! I like that we're converting into a state's local time zone when creating `expected_signature_date`. I am curious why we're using `DateTime.now()` instead of having a static DateTime that we can assert against, though. Doesn't that make the tests more likely to be flaky?

Sorry for all the TimeZone headaches....

PR #5297, opened 2024-12-27
NJ 230 - refactor return_header to remove state_code

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/230

## What was done?
- Refactor based on Martha's comment on PR: https://github.com/codeforamerica/vita-min/pull/5266#discussion_r1894355150
- Previously, we relied on an `if state_code == 'nj'` statement to conditionally place PaidPreparerFields. This refactor makes the logic consistent across all states - if `ptin` and `preparer_person_name` are overwritten in the state's submission builder, it places the relevant fields in the Header. 

## How to test?
- Run unit tests
- Ran through FYST with `jones_mfj` persona, validated that NJ XML had PaidPreparerField block

### DIFF
--- a/app/lib/submission_builder/return_header.rb
+++ b/app/lib/submission_builder/return_header.rb
@@ -16,7 +16,7 @@ module SubmissionBuilder
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
-        if !state_submission_builder.ptin.nil? && !state_submission_builder.preparer_person_name.nil?
+        if state_submission_builder.ptin.present? && state_submission_builder.preparer_person_name.present?
           xml.PaidPreparerInformationGrp do 
             xml.PTIN state_submission_builder.ptin
             xml.PreparerPersonNm state_submission_builder.preparer_person_name
--- a/app/lib/submission_builder/return_header.rb
+++ b/app/lib/submission_builder/return_header.rb
@@ -3,6 +3,10 @@ module SubmissionBuilder
     include SubmissionBuilder::FormattingMethods
     include SubmissionBuilder::BusinessLogicMethods
 
+    def state_submission_builder
+      StateFile::StateInformationService.submission_builder_class(@submission.data_source.state_code)
+    end
+
     def document
       build_xml_doc("ReturnHeaderState") do |xml|
         xml.Jurisdiction "#{@submission.data_source.state_code.upcase}ST"
@@ -12,10 +16,10 @@ module SubmissionBuilder
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
-        if @submission.data_source.state_code.upcase == "NJ"
+        if !state_submission_builder.ptin.nil? && !state_submission_builder.preparer_person_name.nil?
           xml.PaidPreparerInformationGrp do 
-            xml.PTIN "P99999999"
-            xml.PreparerPersonNm "Self Prepared"
+            xml.PTIN state_submission_builder.ptin
+            xml.PreparerPersonNm state_submission_builder.preparer_person_name
           end
         end
         xml.DisasterReliefTxt @intake.disaster_relief_county if @intake.respond_to?(:disaster_relief_county)
--- a/app/lib/submission_builder/state_return.rb
+++ b/app/lib/submission_builder/state_return.rb
@@ -1,18 +1,20 @@
 module SubmissionBuilder
   class StateReturn < SubmissionBuilder::Document
     def document
-      @document = state_schema_version.present? ?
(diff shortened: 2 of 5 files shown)


app/lib/submission_builder/state_return.rb
jachan, line 1: @mmazanec22 Per discussion on https://github.com/codeforamerica/vita-min/pull/5266#discussion_r1894348513. I decided to leave the tests as is. If I added a NJ specific test using the `required_schema: nj` syntax, I had no easy way to assert that all non-NJ states should have an empty block. I felt that the current approach felt most extensible:

```
if state_code == "nj"
# assert behavior for PaidPreparerInformation states
else
# assert behavior for other states
end
```

I also considered making this test conditional off whether the `state_submission_builder` had `ptin` and `preparer_person_name` defined, similar to `return_header`. I decided not to do this, as it felt less error-prone to explicitly define expectations for each state.

mmazanec22: You could iterate through the other required schemas in an approach similar to [the one in `state_return_spec`](https://github.com/codeforamerica/vita-min/blob/7d8ebc8a5ca943b4489cf3cadeae6926c2138951/spec/lib/submission_builder/state_return_spec.rb#L17)

app/lib/submission_builder/state_return.rb
```diff
@@ -1,18 +1,20 @@
 module SubmissionBuilder
   class StateReturn < SubmissionBuilder::Document
     def document
-      @document = state_schema_version.present? ?
-                    build_xml_doc(build_xml_doc_tag, stateSchemaVersion: state_schema_version) :
+      @document = if state_schema_version.present?
```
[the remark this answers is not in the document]
jachan: Yeah, can manually disable the linter and fix if folks prefer the old way.

app/lib/submission_builder/return_header.rb
```diff
@@ -12,10 +16,10 @@ def document
           xml.TaxPeriodEndDt date_type(Date.new(@submission.data_source.tax_return_year, 12, 31))
         end
         xml.TaxYr @submission.data_source.tax_return_year
-        if @submission.data_source.state_code.upcase == "NJ"
+        if !state_submission_builder.ptin.nil? && !state_submission_builder.preparer_person_name.nil?
```
mluedke2, line 19: [dust] - would this work the same and avoid a double-negative?

`if state_submission_builder.ptin.present? && state_submission_builder.preparer_person_name.present?`

app/lib/submission_builder/state_return.rb
mluedke2: i do think that's a `pebble` from me, to do that refactor on the tests too

jachan: Yep, working on that now! It'd be generally useful for any `none_of_the_above` checkboxes.

mluedke2 (approved):
i like this to remove checks to state codes! once done in the test too, i approve 👍

app/lib/submission_builder/return_header.rb
```diff
@@ -3,6 +3,10 @@ class ReturnHeader < SubmissionBuilder::Document
     include SubmissionBuilder::FormattingMethods
     include SubmissionBuilder::BusinessLogicMethods
 
+    def state_submission_builder
```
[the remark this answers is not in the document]
jachan: I think the original plan was to try and remove all `state_code` references entirely, including referencing the `StateInformationService`. However, that can't be done within the existing structure, since this `ReturnHeader` is shared between all states. It has no knowledge of what state it's representing without using state_code. The solution I chose uses the existing state_submission_builder attribute in the `StateInformationService` to figure out what state it's representing.

The other ideas you suggest both seem like viable alternatives!
1) Add to state intake, and then pull from intake here, rather than referencing the `StateInformationService`.

I prefer to avoid using intake, since conceptually, this doesn't match with how we utilize `intake`. `intake` is used for user input, while this is something static that appears on every NJ return, regardless of what the user selects.

2) Adding a separate NJ Return Header

This was the approach I originally had in mind. I'm not quite sure what it would take to implement though. I believe it would involve modifying the `NjReturnXml` child class to override the `return_header` method in the parent `StateReturn` class. We'd also need to create a new `NjReturnHeader` class that is essentially identical to `return_header.rb`, but adds the PaidPreparerFields. This approach feels most correct to me, but I recognize it would involve a lot of duplicated XML between the `ReturnHeader` and the `NjReturnHeader` classes.

Happy to discuss further with CfA about which option feels right here.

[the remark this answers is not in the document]
mpidcock: @mmazanec22 I think you make a good point that this is a new pattern, and methods on the default and NJ intakes would be a better way to stay consistent but @jachanalso makes a good point that intakes are primarily for user input. Also, part of the refactor to implement `StateInformationService` was specifically to extract the static data configured in the intake classes to somewhere centralized. 

I think I was reluctant to add more fields to the `StateInformationService`, but after further consideration I think that is the correct home for state-specific static data such as this. Does that sound alright to you?

PR #5301, opened 2024-12-30
NJ 44 - update NJ voucher info

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/44

## What was done?
- Update NJ voucher and payment info

## How to test?
n/a

n/a


mluedke2 (approved):
i'm curious how much just works automatically or not by including the PDF and the filename in `voucher_path`?

jachan (approved):

aloverso:
@mluedke2 Not entirely sure how to test it (I think Amani is doing that) but all info I have points to just using the path for the voucher. It's the same thing other states are doing, and the ticket included this comment

image

app/lib/pdfs/nj1040v-TY2024.pdf
mpidcock, line 1: [pebble] Because voucher pdfs are served to the user without being filled, it needs to be moved to `/public/pdfs/`. Apologies that this is confusing and not documented!

mpidcock (approved):

PR #5305, opened 2024-12-31
first stab at a data backfill for StateFileArchivedIntake

jnf:
## Link to pivotal/JIRA issue
[FYST-1504](https://codeforamerica.atlassian.net/browse/FYST-1504)

## What was done?
- Adds a rake task (and corresponding service object) to identify and create records for intakes/submissions from TY23 that are 'archiveable.'
- Defining 'archiveable' as an intake + _accepted_ efile submission for TY23.
  - Not concerned w/ efile submissions in any other state, as, for the purposes of the [larger project](https://codeforamerica.atlassian.net/browse/FYST-1466), we're creating this archived records for the explicit purpose of providing PDFs to clients who successfully submitted a return.
- The rake task is batched at 10. That's super safe mode and we can tune it if we want. I chose such a low number as 'archiving' requires attaching a PDF blob to the archive record. It's still unclear to me whether this _duplicates_ the actual file in S3 or just adds an additional reference. Need to figure out which it is and ensure that the later task of removing old intakes from the production table doesn't accidentally break this association.

## How to test?
- The service object is testable in a rails console. Creating an instance for a state exposes the core methods of `#query_archiveable`, `#current_batch`, and `#archive_batch`.
- The rake task can be run locally, but you'll need some TY23 returns with an 'accepted' efile submission. jey has a database dump you can use with NY and AZ returns in this state.

[FYST-1504]: https://codeforamerica.atlassian.net/browse/FYST-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,74 @@
+module StateFile
```
anisharamnani, line 1: at a high level, i think it would be helpful to write tests for this file.

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,74 @@
...
+        archive = StateFileArchivedIntake.new(record.without('source_intake_id'))
+        archive.submission_pdf.attach(data_source.find(record['source_intake_id']).submission_pdf.blob)
+        archive.save!
+        archived_ids << record['source_intake_id']
+      rescue StandardError => e
+        Rails.logger.warn("Caught exception #{e} for record #{record}. Continuing with batch")
+        next
+      end
+      Rails.logger.info("Archived #{archived_ids.count} #{data_source} intakes: [#{archived_ids.join(', ')}]")
+      @current_batch = nil # reset the batch
```
anisharamnani, line 37: why is the `@current_batch` set to nil for this method?

anisharamnani: ah, i think it is to exit the `while` loop in the rake task.

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      Rails.logger.info("Found #{current_batch.count} #{data_source} intakes to archive.")
+    end
+
+    def archive_batch
+      archived_ids = []
+      current_batch.each do |record|
+        archive = StateFileArchivedIntake.new(record.without('source_intake_id'))
+        archive.submission_pdf.attach(data_source.find(record['source_intake_id']).submission_pdf.blob)
+        archive.save!
+        archived_ids << record['source_intake_id']
```
anisharamnani, line 31: 🤔 i am assuming that we don't want to archive and delete in the same rake task.

maybe it would be helpful to have a method that after the `archive_batch` is run, we verify that the archivable intakes now have an archivable record. 

i don't if that makes sense or is easy.

anisharamnani:
also, the annotate error should be resolved if you merge in `FYST-1466-prior-year-access-mvp` 🎉

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      # two subqueries in this :/ there's probably a better way
+      # i could eliminate the 2nd subquery if records in state_file_archived_intakes contained a reference to
+      # their intake source data. however that reference would eventually/quickly become orphaned once the
+      # source table is cleaned out in preparation for the next tax year
+      <<~SQL
+        SELECT
+          #{tax_year} AS tax_year, '#{state_code}' AS state_code,
+          email_address, hashed_ssn, id AS source_intake_id
+        FROM
+          state_file_az_intakes
```
anisharamnani, line 50: 🤔 just realized this query is just for AZ. since this is a WIP PR, i'm sure you were gonna change this, but just pointing out for other reviewers.

app/services/state_file/ty23_archiver_service.rb
anisharamnani: also how were you thinking of saving the intake IDs?

app/services/state_file/ty23_archiver_service.rb
jnf: yeah, fixed in the next commit.

app/services/state_file/ty23_archiver_service.rb
jnf: not deleting in this task. collecting the ids for logging purposes only. we'll do the actual table truncation in a different effort.

app/services/state_file/ty23_archiver_service.rb
jnf: setting it to `nil` to represent that the batch has been processed; those intakes are no longer eligible for archiving because they've been archived :)

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,92 @@
...
+      @data_source = INTAKE_MAP[state_code.to_sym]
+      @tax_year = 2023
+      @cutoff = cutoff
+      @current_batch = nil
+      raise ArgumentError, "#{state_code} isn't an archiveable state. Expected one of #{INTAKE_MAP.keys.join(', ')}" unless data_source
+    end
+
+    def find_archiveables
+      query_result = ActiveRecord::Base.connection.exec_query(query_archiveable_intakes)
+      @current_batch = query_result.as_json
```
mpidcock, line 22: The main reason I changed this to json was to make it easier to mock in the tests, but can revert if there's an objection

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+      Rails.logger.info("Archived #{archived_ids.count} #{data_source.name.pluralize}: [#{archived_ids.join(', ')}]")
+      @current_batch = nil # reset the batch
+      archived_ids
+    end
+
+    def query_archiveable_intakes
+      <<~SQL
+        SELECT
+          id AS intake_id, hashed_ssn
```
mpidcock, line 66: hashed_ssn can also be extracted from the intake, but returning it here makes it possible to cross-reference with existing archived intakes for sanity checks

app/services/state_file/ty23_archiver_service.rb
jnf: i chose `exec_query` so the results would be a hash, which makes for easy assignment in the batch processing. it's also why i'm SELECTing tax_year and state_code is the archiveable query. if you're gonna make this json, should probably change the query too.

app/services/state_file/ty23_archiver_service.rb
```diff
@@ -1,49 +1,69 @@
...
-        intake = data_source.find(record['source_intake_id'])
-        archive = StateFileArchivedIntake.new(record.without('source_intake_id'))
-        # TODO: pull mailing address destails off the intake; populate relevant fields on the archived intake record
-        archive.submission_pdf.attach(intake.blob)
-        archive.save!
-        archived_ids << record['source_intake_id']
+      @current_batch&.each do |batch_data|
+        archived_intake = StateFileArchivedIntake.new(hashed_ssn: batch_data['hashed_ssn'])
+        archived_intake.state_code = @state_code
+        archived_intake.tax_year = @tax_year
```
jnf, line 31: prefer you use the accessors for reads; generally good practice to use defined getters/setters over accessing properties directly.

app/services/state_file/ty23_archiver_service.rb
jnf: ah, i see you already updated the query. i am curious how `as_json` makes things easier, but, again, for a one off task i don't think it much matters.

mpidcock: i ran the debugger to see what @current_batch was set to so I could mock it in the test for `#archive_batch`. The object type is `ActiveRecord::Result`, which I found challenging to mock compared to json.
image

mrotondo (approved):

PR #5312, opened 2025-01-04
Nj 224 check claimed as dependent for lines 10 and 11

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/224

## Is PM acceptance required?

## What was done?
- Added requirement that dependent must be explicitly claimed as dependent in JSON to count for lines 10 or 11
- The rest is just adjusting test cases. The XML and PDF tests for this were coupled with the calculator so I instead set them to have a mocked calculator.

## How to test?
- Latifah is the only profile that will show new changes: will have 0 for lines 10 or 11 (before would have had 1 on line 10)
- Others remain the same:
  -  Minimal will have 0 for lines 10 or 11
  - Sinatra will have 1 on line 10
  - Zeus many deps will have 10 on line 10 and 1 on line 11

## Screenshots

Latifah:

image


app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -224,7 +224,7 @@ def calculate_line_9
 
       def calculate_line_10_count
         @intake.direct_file_json_data.dependents.count do |dependent|
-          dependent.qualifying_child
+          dependent.qualifying_child && dependent.is_claimed_dependent
```
mpidcock, line 227: [dust] I'm ok with how this is, but so you know, the `qualifying_child` column in `StateFileDependent` is populated from `direct_file_json_data` during import, and all of the dependents in `intake.dependents` have already been filtered to be only claimed dependents.

It would be a better practice to avoid extracting data from the json when it is available in the database, especially, for example, if you needed to access other dependent fields from the database that aren't in the json.

mpidcock (approved):

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -553,40 +553,26 @@
 
       describe "Line 10 exemptions" do
         context "0 qualified dependent children" do
-          let(:submission) {
-            create :efile_submission, tax_return: nil, data_source: create(
-              :state_file_nj_intake
-            )
-          }
           it "does not fill in" do
+            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_10_count).and_return 0
```
mmazanec22, line 557: This is an improvement!

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -383,25 +375,10 @@
     end
 
     describe "total exemption - lines 13 and 30" do
-      let(:intake) { create(:state_file_nj_intake, :primary_over_65, :primary_blind, :primary_veteran, :two_dependents_in_college) }
-      it "totals lines 6-12 and stores the result in both TotalExemptionAmountA and TotalExemptionAmountB" do
-        line_6_single_filer = 1_000
```
mmazanec22, line 388: Makes sense to remove this, as it is covered in the calculator tests

mmazanec22 (approved):

app/lib/efile/nj/nj1040_calculator.rb
mluedke2: @mpidcock I didn't know that! When I search the codebase for either `is_claimed_dependent` or `isClaimedDependent`, I didn't come across any references that pointed me towards logic like that? Can you help me find it?

But regardless, I tried simplifying the logic as you said and sure enough the tests still pass. So I'll push a change to

```
def calculate_line_10_count
  @intake.dependents.count do |dependent|
    dependent.qualifying_child
  end
end

...
def calculate_line_11_count
  @intake.dependents.count do |dependent|
    !dependent.qualifying_child
  end
end
```

mpidcock: @mluedke2 Well, we don't actually have a method to filter claimed dependents, because in the list of dependents in the XML is specifically a list of claimed dependents. It was new this year to have json data for other members of the household, but since our database gets instantiated with the xml data and augmented by the json data, the non-dependent household members are simply never added to the list of dependents. I hope that helps and let me know if I can clarify anything else!

mluedke2: Oh ok that makes sense. Thanks!

PR #5318, opened 2025-01-06
update twilio auth token to newly rotated primary token, in hopes that it fixes webhook request validation

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1533

## What was done?
- Some time in mid-December, we started failing to validate Twilio webhook requests. The only input to the validation method that we pass from our code is the Twilio auth token. This auth token was rotated in this same time period and the auth token we were using became the "secondary" auth token - the credentials were never updated to use the new primary auth token. 
- We believe that while the secondary auth token seems to work for sending messages, its use may be the problem with validating the signature of the webhook requests - maybe they are signed with the same private key that generated the primary auth token, and therefore the signature can't be validated with the secondary auth token?
## How to test?
- We will monitor [datadog](https://app.datadoghq.com/logs?query=outgoing_text_messages&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice%2C%40status&context_event=AZOOKdY1AAD0BxTMHBeJlAAe&event=AwAAAZQ9kYrA2i-lwAAAABhBWlE5a2RiLUFBQVZQc1VrSjJKYkVnQWcAAAAkMDE5NDNkOTItZDJlMC00NzQwLTk3NzQtMjE4ZTRhNTVmMGZjAAAH1g&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=time%2Cdesc&viz=&from_ts=1736198856045&to_ts=1736199756045&live=true) to see if the 403s are resolved after this PR is merged to main and deployed to demo by running our repro case (sending a text from demo & watching the logs for the webhook requests).


jnf (approved):
woof. thanks for catching this. even if this hypothesis is refuted it's good to close the loop on the credential updates.

PR #5200, opened 2024-12-13
Add line 16a to line 27

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/202

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

## What was done?
Add line 16a to line 27 and alter the tests for line 27 and line 29 (which is just the value from line 27)

## How to test?
See steps in ticket

PR #5319, opened 2025-01-07
NJ 239 State Wages Bug

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/239

## What was done?
- a `-1` indicator was being used to indicate when there were no state wages, and this had been fine because it was the only value considered for XML, which would skip it since it was not positive.
- but then in https://github.com/codeforamerica/vita-min/pull/5200 , the `-1` value itself was then added to Line 16a
- this work just sets the sum to be `0` instead of `-1` 
- When no W2s or when sum is 0: XML element is not present and Line 15 is left blank.

## How to test?
- Jon bon Jovi persona, will see `TotalIncome` and `GrossIncome` of `10200` as expected, not `10199`
### Before

image

```
<Body>
<TaxableInterestIncome>10200</TaxableInterestIncome>
<TotalIncome>10199</TotalIncome>
<GrossIncome>10199</GrossIncome>
<TotalExemptionAmountB>4000</TotalExemptionAmountB>
<TotalExemptDeductions>4000</TotalExemptDeductions>
<TaxableIncome>6199</TaxableIncome>
```
### After

image

image

image


mmazanec22 (approved):

jenny-heath (approved):
makes sense!

aloverso (approved):
Looks good! My only question is to validate the correctness of the approach - ie, is there ever a time where the W2 wages could actually sum to 0 somehow (and thus the zero should be included on the XML)?

mluedke2:
@aloverso good question -- in [the ticket discussion](https://github.com/newjersey/affordability-pm/issues/239#issuecomment-2578038434) I clarified with Amani and she approved always having it blank for either reason of it being 0

PR #5321, opened 2025-01-07
FYST-1450 Add an sms equivalent to StateFileNotificationEmail objects

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1450

## What was done?
- Made a new model so we can have records of outgoing notification texts for state file
- Tried to replicate the same patterns we use for email which includes putting the actual sending of the text in a job and triggering the job when the record is persisted
- Doesn't include leveraging the twilio webhooks to update the message sending status; can add as a follow-up?
## How to test?
- Added some coverage for emails (new spec file for delayed job, model spec for #deliver)
- Coverage for texts:
  - delayed job spec mocks twilio methods (tried briefly to use MockTwilio but couldn't figure out how to mock response with sid)
  - model spec checks after_create and deliver
  - messaging service now looks at record count instead of using MockTwilio
- Risk Assessment
  - the only real risk i guess is that texts stop working so acceptance testing should include trying to send texts
- N/A

### DIFF
--- a/app/jobs/state_file/send_notification_text_message_job.rb
+++ b/app/jobs/state_file/send_notification_text_message_job.rb
@@ -1,0 +1,22 @@
+module StateFile
+  class SendNotificationTextMessageJob < ApplicationJob
+    def perform(state_file_notification_text_message_id)
+      state_file_notification_text_message = StateFileNotificationTextMessage.find(state_file_notification_text_message_id)
+      twilio = TwilioService.new(:statefile)
+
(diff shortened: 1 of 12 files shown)


squanto (approved):
lgtm 👍

jnf (approved):
appreciate all the tests. agree that leveraging webhooks to update status can be a followup effort.

PR #5324, opened 2025-01-08
Nj 78 2 checkbox

mmazanec22:
## Link to pivotal/JIRA issue
Addresses some of the fieldset issues identified here https://github.com/newjersey/affordability-pm/issues/78
And explained in depth here:
- https://github.com/newjersey/accessibility-pm/issues/28
- https://github.com/newjersey/accessibility-pm/issues/30
- https://github.com/newjersey/accessibility-pm/issues/31

## What was done?
Add a fieldset component that allows us to include a follow up question. Replace all of our weirdly nested fieldsets with this one. Abstract some fieldset checkbox logic to controllers where it's more complex.

Also deal with an existing title html issue that was making the title appear strangely at the top of the page (seemingly on local only??).

Note: this does apply the tight checkbox styling to all checkboxes using this component. We could change that but I thought the more uniform spacing might be an improvement, esp since our implementation was not a pixel perfect match of the designs to begin with.

## How to test?
For homeowner and tenant eligibility, select a married filing separately profile and make sure the conditional checkboxes still show up.

For college dependents and dependent health insurance, select the "many deps claimed as dep" and make sure those checkboxes still work.

- Before
image

- After
image

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -247,6 +247,14 @@
       }
     }
 
+    .tight-ish-checkboxes {
+      @extend .tight-checkboxes;
+
+      .checkbox {
+        margin: 1rem 0
+      }
+    }
+
     .radio-button {
       border: none;
       background-color: transparent;
--- a/app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
+++ b/app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
@@ -22,12 +22,12 @@ module StateFile
           {
             method: :homeowner_main_home_multi_unit,
             label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_main_home_multi_unit"),
-            has_follow_up_id: 'homeowner_multi_unit_followup'
+            opens_follow_up_with_id: 'homeowner_multi_unit_followup'
           },
           {
             method: :homeowner_main_home_multi_unit_max_four_one_commercial,
             label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_main_home_multi_unit_max_four_one_commercial"),
-            is_follow_up_id: 'homeowner_multi_unit_followup'
+            follow_up_id: 'homeowner_multi_unit_followup'
           },
         ]
         if current_intake.filing_status_mfs?
--- a/app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
+++ b/app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
@@ -3,6 +3,45 @@ module StateFile
     class NjHomeownerEligibilityController < QuestionsController
       include ReturnToReviewConcern
 
+      before_action -> { set_checkboxes }
+
+      def set_checkboxes
+        @checkbox_collection = [
+          {
+            method: :homeowner_home_subject_to_property_taxes,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_home_subject_to_property_taxes")
+          },
+          { 
+            method: :homeowner_more_than_one_main_home_in_nj,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_more_than_one_main_home_in_nj"),
+          },
+          { 
+            method: :homeowner_shared_ownership_not_spouse,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_shared_ownership_not_spouse"),
+          },
+          {
+            method: :homeowner_main_home_multi_unit,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_main_home_multi_unit"),
+            has_follow_up_id: 'homeowner_multi_unit_followup'
+          },
+          {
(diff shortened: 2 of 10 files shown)


app/helpers/vita_min_form_builder.rb
```diff
@@ -401,6 +401,70 @@ def vita_min_date_text_fields(method, label_text, help_text: nil, options: {}, o
...
+        <label class="#{joined_classes}">
+        #{check_box(*checkbox_args)} <span class="text--normal">#{item[:label]}</span>
+        </label>
+      </div>
+    HTML
+  end
+
+  # Coped from Honeycrisp form builder v1 cfa_checkbox_set, modified to allow checkboxes to have follow up
+  def vita_min_checkbox_set(
+    method,
```
mmazanec22, line 434: I copied this from honeycrisp and tried not to change the method signature to avoid making this difficult/annoying to use. One question I have: how can we use the method here to do error handling when the checkboxes themselves are associated with db fields, not the element overall?

app/assets/stylesheets/_state-file.scss
```diff
@@ -247,6 +247,14 @@
       }
     }
 
+    .tight-ish-checkboxes {
```
mmazanec22, line 250: This change was made to make the checkboxes taller to meet the minimum click target size. CfA might want to consider changing the tight-checkboxes class

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
```diff
@@ -3,6 +3,45 @@ module Questions
...
+          },
+          {
+            method: :homeowner_main_home_multi_unit,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_main_home_multi_unit"),
+            has_follow_up_id: 'homeowner_multi_unit_followup'
+          },
+          {
+            method: :homeowner_main_home_multi_unit_max_four_one_commercial,
+            label: I18n.t("state_file.questions.nj_homeowner_eligibility.edit.homeowner_main_home_multi_unit_max_four_one_commercial"),
+            is_follow_up_id: 'homeowner_multi_unit_followup'
```
mluedke2, line 30: [dust] - i was initially confused by the difference between `has_follow_up_id` and `is_follow_up_id`... to me, when something *has* an ID of `12345` that also means it *is* the thing with that ID.

Thoughts about wording like this?

`has_follow_up_with_id`
`follow_up_id`

config/locales/en.yml
```diff
@@ -3143,6 +3143,7 @@ en:
           continue: Click "continue" if all these people had health insurance.
           coverage_heading: What is health insurance with minimum essential coverage?
           label: 'Check all the people that did NOT have health insurance:'
+          title: Please tell us which dependents were missing health insurance (with minimum essential health coverage) in 2024.
```
mluedke2, line 3146: [dust] do you mind including the before/after screenshots for the use of this? Without the HTML link this text reads awkwardly to me.

mluedke2 (commented):
no big concerns from me. i left a couple comments -- I will check back again later on the comments and after the tests pass

app/controllers/state_file/questions/nj_homeowner_eligibility_controller.rb
mmazanec22: Yeah that seems better! Thanks for the suggestions!

config/locales/en.yml
mmazanec22: This just changes the title of the page so that there's no html gore in the browser tab

mluedke2: makes sense, thanks!

app/helpers/vita_min_form_builder.rb
```diff
@@ -401,6 +401,70 @@ def vita_min_date_text_fields(method, label_text, help_text: nil, options: {}, o
...
+      container_class = "question-with-follow-up__follow-up"
+      container_id = item[:follow_up_id]
+    end
+  
+    joined_classes = ((item[:classes] || []) + ["checkbox"]).join(" ")
+    checkbox_args = [item[:method], item_options]
+    checkbox_args += ["yes", "no"] if enum
+
+    <<~HTML.html_safe
+      <div id="#{container_id}" class="#{container_class}">
```
powersurge360, line 424: For these variables, do you think you can try, eg, `#{h container_id}`? It's not something we're currently doing but I am concerned that whitelisting with `html_safe` and then interpolating variables is going to bite us at some point. If it works with that we may consider doing that going forward. `#h` is an alias for `raw` and is basically the way you un-`html_safe`.

app/helpers/vita_min_form_builder.rb
powersurge360: These form builders are getting kind of complex. I'm not sure how you'd do it either. May require stepping back and creating a new control that _does_ deviate.

powersurge360 (approved):
I'm having a little trouble reasoning about this but I think it looks ok. I have a suggestion about escaping arguments that I've been thinking about that although I mention here, isn't unique to here.

app/helpers/vita_min_form_builder.rb
mmazanec22: Following this suggestion threw a method undefined error. I made a follow up ticket to evaluate the whitelisting/interpolating issue

PR #5326, opened 2025-01-08
Fyst 1517 add ssn itin validation step to prior year access flow

anisharamnani:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1517

## What was done?
- Added ssn page verification
- Added validations to check for existing archived intakes with matching ssn and email address
- Added in the conditions for allowing access to the ssn page which includes checking in the session that the code was verified in the previous screen and checking that the request has not already been locked out

## How to test?
- Add units tests for the new form and controller
- Updated verification code controller spec for adding condition to check if code was verified before accessing the identification number page
- Added to the feature spec 
- After
Screenshot 2025-01-17 at 2 31 01 PM


app/forms/state_file/completed_return_email_address_controller.rb
```diff
@@ -0,0 +1,5 @@
+module StateFile
+  class CompletedReturnEmailAddressForm < QuestionsForm
+
+  end
```
anisharamnani, line 4: we have to delete this too.

app/lib/navigation/completed_tax_return_navigation.rb
```diff
@@ -0,0 +1,12 @@
+module Navigation
+  class CompletedTaxReturnNavigation < Navigation::StateFileBaseQuestionNavigation
```
anisharamnani, line 2: we also must delete this.

app/views/state_file/questions/validate_identification_number/edit.html.erb
```diff
@@ -0,0 +1,2 @@
+<% content_for :page_title, t(".title") do %>
```
anisharamnani, line 1: i think we delete this tooo.

db/data/20241226211010_add_address_to_completed_2023_tax_return.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
anisharamnani, line 1: bye bye bye

app/controllers/state_file/archived_intakes/identification_number_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        if @form.valid?
+          create_state_file_access_log("correct_ssn_challenge")
+          current_request.reset_failed_attempts!
+          session[:ssn_verified] = true
+          redirect_to root_path
+          # need to change to address controller
+        else
+          create_state_file_access_log("incorrect_ssn_challenge")
+          current_request.increment_failed_attempts
```
mrotondo, line 23: It looks like we're re-using the same mechanism for marking "failed attempts" here as on the verification code page. If the user mis-enters a verification code and then gets it right on their second try, will that count against their total allowed failed attempts? In other words, should we reset failed attempts to zero after a user succeeds verification code authorization?

spec/controllers/state_file/archived_intake/identification_number_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        expect(assigns(:form)).to be_valid
+
+        access_log = StateFileArchivedIntakeAccessLog.last
+        expect(access_log.state_file_archived_intake_request).to eq(intake_request)
+        expect(access_log.event_type).to eq("correct_ssn_challenge")
+        expect(session[:ssn_verified]).to eq(true)
+        expect(intake_request.reload.failed_attempts).to eq(0)
+
+        expect(response).to redirect_to(root_path)
+        # need to change to address path
```
mrotondo, line 66: Can you replace these comments with a TODO that reference the story ID where the work to change the path will be done?

mrotondo (commented):
One hunch of a potential bug in the interaction between this story and the previous one that I'd like you to check out, and one request for changes to comments, but looks good other than that! LMK when you've gotten a chance to check out the potential interaction with failed verification code attempts & I'll approve.

app/controllers/state_file/archived_intakes/identification_number_controller.rb
tahsinaislam: oh I thought we did this in both controllers when the form is valid by line 17:           current_request.reset_failed_attempts! but i could be wrong would be down to discuss on tuesday!

mrotondo: You're totally right, I missed that! Thanks for pointing it out - looks good!

mrotondo (approved):

PR #5332, opened 2025-01-09
FYST-431 AZ Final Review navigation

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1229

## Is PM acceptance required? 

## What was done?
- When navigating from the review page to add/edit for the public school and qualifying contributions flows, there was inconsistent behavior with returning to the review page when trying to add another contribution for each flow. It worked when editing existing contributions or adding the first contribution, but it failed to return to the review page when adding additional contributions with existing records.
- For the add contribution button, the params for return for review was added to the actual form submission data that gets sent when the button is clicked. We currently only passed it into the URL parameter but we needed both URL parameters and form parameters to ensure the return_to_review value wasn't lost during the form submission adding a new contribution

## How to test?
- This was tested by going through the flow locally for the AZ flow

### DIFF
--- a/app/views/state_file/questions/az_public_school_contributions/index.html.erb
+++ b/app/views/state_file/questions/az_public_school_contributions/index.html.erb
@@ -37,7 +37,7 @@
   <% end %>
 
   <%= button_to(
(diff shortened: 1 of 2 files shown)


mrotondo (approved):
🎸

PR #5348, opened 2025-01-10
NJ 244 - Fix property tax credit income threshold bug

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/244

## What was done?
- Currently, the income threshold for Property Tax is based on the sum of all W2s. This is incorrect, and the threshold should be based on the NJ Gross Income (line 29). See [original ticket.](https://github.com/newjersey/affordability-pm/issues/97)

## How to test?
- Use `Streep single inv limit`
- Get to property tax screen (see below)
- Select either homeowner or renter
- Check the first box to be eligible for the credit and click continue
- Verify that you see the Rent Paid or Property Tax paid screen

- Before (skips to the property tax offramp)

https://github.com/user-attachments/assets/2a8184c8-c346-43f1-a709-60edbbf52110

- After (asks user how much property tax was paid to your township)

https://github.com/user-attachments/assets/5f96d960-5c37-4aa8-ac5d-f130f68b1c46


spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -758,7 +758,7 @@
         context "when qualifies for income exemption" do
           it "does not check 53c Schedule NJ-HCC checkbox and leaves 53a, 53b, and 53c amount blank" do
             single_income_threshold = 10_000
-            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_54).and_return single_income_threshold
+            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_29).and_return single_income_threshold
```
mluedke2, line 761: [dust] can you help me understand why this change was needed? seems very separate from Line 15/29

mluedke2: upgrading to [pebble/sand] - Out of interest I tried running this file of tests the way had it been before, with line 54, and it still passes that way... as well as with lines 760 and 761 removed entirely!

So this test doesn't seem to be testing what we think it's testing, which still might be a separate bug from this ticket but we may want to resolve while we are looking at it now, but at least make a follow-up bug ticket.

My guess is we need to add a `context "when does not qualify for income exemption" do` next to this test, making sure the income is mocked as being high.

app/lib/efile/nj/nj_property_tax_eligibility.rb
```diff
@@ -19,26 +19,26 @@ def possibly_eligible_for_deduction_or_credit?(intake)
         end
 
         def determine_eligibility(intake)
-          state_wages = Efile::Nj::NjStateWages.calculate_state_wages(intake)
```
mluedke2, line 22: [sand] - The reason that `NjStateWages.calculate_state_wages()` is in its own util is because it was called both here and in the calculator, and so was easier to break out for mocking purposes. But if we don't call it here anymore and don't need it for this util's tests, then `NjStateWages.calculate_state_wages()` would only be called in one place in one file... and given that it's only two lines of code anyway, it's now more boilerplate than it's worth.

So I'd say that either in this ticket or a follow-up, we should simplify and get rid of the `NjStateWages` util.

mluedke2 (approved):
i think this bug arose because for a while Line 15 and Line 29 were the same. Line 16a was added in but this was still pointing at the same source. glad it was caught and easy to fix!

there are two follow up tickets i think we should make from my comments, and I think one of the changed lines doesn't need to be changed in this and should be in the follow-up. but otherwise all looks good to me

PR #5372, opened 2025-01-15
NJ 223 - add ineligible screen for low income property tax (no exemption)

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/223

## What was done?
- Reroute household_rent_own to ineligible page when low-income and does not meet exemption
- Handle new low-income reasons

## How to test?
- Use `minimal` for single low-income (select "no" for disabled)
- Use `mfj 15k wages` for MFJ low-income (select "no" for disabled)

image

### DIFF
--- a/app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
+++ b/app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
@@ -7,6 +7,14 @@ module StateFile
       helper_method :on_home_or_rental
 
       def determine_reason
+        if Efile::Nj::NjPropertyTaxEligibility.ineligible?(current_intake)
+          if current_intake.filing_status_single? || current_intake.filing_status_mfs?
+            return "income_single_mfs"
+          else
+            return "income_mfj_qss_hoh"
+          end
+        end
+
         if current_intake.household_rent_own_neither?
           return "neither"
         end
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -3262,6 +3262,8 @@ es:
           reason_multi_unit_conditions: this main home in a multi-unit building did not meet the requirements to claim this credit or deduction.
           reason_neither: you must own rent or own a main home in %{filing_year} to receive this credit or deduction.
           reason_property_taxes: did not pay property taxes (directly, or through rent) on this main home.
+          reason_income_single_mfs: you did not meet the income filing threshold of $10,000.
+          reason_income_mfj_qss_hoh: you did not meet the income filing threshold of $20,000.
           title_html: It looks like you're not eligible for the Property Tax Deduction or Credit <strong>%{property}</strong>
       nj_medical_expenses:
         edit:
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -3259,11 +3259,11 @@ es:
           on_home: on the home you lived in and owned.
           on_rental: on the home you lived in and rented.
           reason_html: You are not eligible for this because <b>%{reason}</b>
+          reason_income_mfj_qss_hoh: you did not meet the income filing threshold of $20,000.
+          reason_income_single_mfs: you did not meet the income filing threshold of $10,000.
           reason_multi_unit_conditions: this main home in a multi-unit building did not meet the requirements to claim this credit or deduction.
           reason_neither: you must own rent or own a main home in %{filing_year} to receive this credit or deduction.
           reason_property_taxes: did not pay property taxes (directly, or through rent) on this main home.
-          reason_income_single_mfs: you did not meet the income filing threshold of $10,000.
-          reason_income_mfj_qss_hoh: you did not meet the income filing threshold of $20,000.
           title_html: It looks like you're not eligible for the Property Tax Deduction or Credit <strong>%{property}</strong>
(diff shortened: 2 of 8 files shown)


config/locales/es.yml
```diff
@@ -3264,6 +3264,8 @@ es:
           reason_multi_unit_conditions: this main home in a multi-unit building did not meet the requirements to claim this credit or deduction.
           reason_neither: you must own rent or own a main home in %{filing_year} to receive this credit or deduction.
           reason_property_taxes: did not pay property taxes (directly, or through rent) on this main home.
+          reason_income_single_mfs: you did not meet the income filing threshold of $10,000.
+          reason_income_mfj_qss_hoh: you did not meet the income filing threshold of $20,000.
```
aloverso, line 3268: I went back and forth on whether to make these two separate content lines, or store it as one like like `"you did not meet the income filing threshold of %{income_threshold}"`

Ultimately I went with this because in localizations, we might need to change how the income numbers are represented (ie, using `20,000` vs `20.000`) so it seemed like the income strings would have to be stored as content keys anyways. So it came down to the complexity of having *three* content strings (`reason_income`, `income_10k`, and `income_20k`) vs the two you see here, and then having do a bit MORE logic every time we use them (ie, we'd have to store the decision-logic of single=10k, MFJ=20k anywhere we reference these strings, which would include the e2e tests). It seemed simpler for the code to just reference the filing status and allow the specific numbers to live in the content, making it easier for them to change in the future.

Curious what folks think/prefer on this rationale.

mmazanec22: This seems fine. It does make me wonder how we can remind ourselves when we update content to consider what related content needs to be changed too

mluedke2: I do think think you can do this with just one content string

`reason_income: you did not meet the income filing threshold of %{threshold}.`

and then we could use [number_to_currency](https://api.rubyonrails.org/v8.0.1/classes/ActiveSupport/NumberHelper.html#method-i-number_to_currency) to feed in the value.

That way also the threshold constants used in the eligibility logic could be one source of truth, exported and referenced here, and the UI strings would adapt.

This is done in a few other places in the app ([like this](https://github.com/codeforamerica/vita-min/blob/4296257d7f0feadb6c3fc2722a31b23ba1a52b92/app/views/ctc/questions/income/edit.html.erb#L15) for instance) so we'd be following a set pattern.

app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
```diff
@@ -7,6 +7,14 @@ class NjIneligiblePropertyTaxController < QuestionsController
...
 
       def determine_reason
+        if Efile::Nj::NjPropertyTaxEligibility.ineligible?(current_intake)
+          if current_intake.filing_status_single? || current_intake.filing_status_mfs?
+            return "income_single_mfs"
+          else
+            return "income_mfj_qss_hoh"
+          end
+        end
+
```
mluedke2, line 17: [dust] - this seems like the cleanest way to do this for this PR, but I do wonder long-term if `determine_reason` would be better living in the `NjPropertyTaxEligibility` util... though that refactor would be good to do alongside combining with the `NjHomeownerEligibilityHelper` and `NjTenantEligibilityHelper` into one central brain for property tax eligibility. but again, probably for another day.

spec/features/state_file/nj/complete_intake_spec.rb
```diff
@@ -460,5 +460,29 @@ def advance_to_property_tax_page(skip_health_insurance_eligibility: false)
         expect_estimated_tax_page
       end
     end
+
+    context "when low income and does NOT meet exception" do
```
mmazanec22, line 464: [dust] This test is a little vague

mmazanec22 (approved):

mrotondo (approved):
Looks good! I agree with Matt that an eventual refactoring of the ineligible logic into the single utility class would be good, but that's non-blocking.

config/locales/es.yml
aloverso: @mluedke2 Started going down the path of this refactor, but unfortunately I don't think it will work because `number_to_currency` is a ruby ActionView helper which means it is only available in views (and this would be trying to use it inside the controller).

You could argue that this is indicative of a larger code smell - ie, the way that the ineligible_controller is resolving translation strings should live in the view instead per a more Ruby-esque approach - but that seems like a larger refactor.

Okay with keeping it as-is for now?

mluedke2: yeah i agree. does feel like a slight code smell as you said but since it would require deeper refactor it sounds like a less important follow-up ticket. thanks for trying it out!

app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
mluedke2: linking to the ticket which captures this work for cross-reference

https://github.com/newjersey/affordability-pm/issues/227

config/locales/es.yml
mluedke2: I've added this to the ticket of property tax refactors:

https://github.com/newjersey/affordability-pm/issues/227

PR #5374, opened 2025-01-15
FYST-1607 Add require_verification: false to the AfterTransitionMessagingService notifications that don't have it

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1607
- Yes

## What was done?
- Realized while testing the `StateFile::AutomatedMessage::StillProcessing` message that in cases where a client opts into both contact methods for notifications, they only receive messages with the contact info they did the verification step with. This is inconsistent with our other after transition messages that do not require contact methods be verified to send the message.
## How to test?
- Updated `after_transition_messaging_service_spec.rb` to not have verified contact info
  - the most "realistic" setup here would be to have one of them verified—open to changing that
- Updated `after_transition_messaging_service_spec.rb` to look for text message records

### DIFF
--- a/spec/services/state_file/after_transition_messaging_service_spec.rb
+++ b/spec/services/state_file/after_transition_messaging_service_spec.rb
@@ -1,16 +1,15 @@
 require "rails_helper"
 
 describe StateFile::AfterTransitionMessagingService do
-  include MockTwilio
   include ActiveJob::TestHelper
 
   let(:intake) do
     create :state_file_az_intake,
            primary_first_name: "Mona",
            email_address: "mona@example.com",
-           email_address_verified_at: 1.minute.ago,
+           email_address_verified_at: nil, # none of these notifications require verification
            phone_number: "+15551115511",
-           phone_number_verified_at: 1.minute.ago,
+           phone_number_verified_at: nil, # ^ so this tests that require_verification is passed in as false
            message_tracker: {}
   end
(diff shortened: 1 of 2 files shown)


spec/services/state_file/after_transition_messaging_service_spec.rb
```diff
@@ -115,20 +118,11 @@
...
-
-      expect(efile_submission.message_tracker).to include "messages.state_file.terminal_rejected"
-      expect(StateFile::MessagingService).to have_received(:new).with(intake: intake, submission: efile_submission, message: message, body_args: body_args)
-    end
-
-    it "sends the terminal rejected refund sms" do
-      expect do
-        messaging_service.send_efile_submission_terminal_rejected_message
-        perform_enqueued_jobs
-      end.to change(FakeTwilioClient.messages, :count).by(1)
```
jenny-heath, line 131: since the actual twilio-message-sending is tested for the `StateFileNotificationTextMessage#after_create_commit` hook (and the job that it kicks off), it seems unnecessary to use `MockTwilio` here and should be sufficient to check that the `StateFileNotificationTextMessage` was created

arinchoi03 (approved):
i think having the realistic test set up would be great, but don't want to block this PR, maybe have a small follow up chore if possible :) ty!

PR #5379, opened 2025-01-16
NJ 160 - Add underpayment notice

jachan:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/160

## What was done?
- Forked the TaxesOwedController and TaxesOwedForm, creating NjTaxesOwedController/Form.
- Added logic to NjTaxesOwed page to display a message when taxpayer has underpaid by more than 400$

## How to test?
- Use Jones MFJ persona
- Seelct Other on homeowner/renter screen
- Taxes owed should be 677

To test behavior at $400 in taxes owed
- Do above
- Add 277 in estimated tax payments page

Testing behavior for less than $400 in taxes owed
- Do above
- Add 300 in estimated tax payments

- Before

image

- After

image

image


app/controllers/state_file/questions/nj_taxes_owed_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        current_intake.calculated_refund_or_owed_amount.abs
+      end
+      helper_method :taxes_owed
+
+      private
+
+      def card_postscript; end
+    end
+  end
+end
```
mluedke2, line 19: [potential boulder] - If I'm reading it right, I think most of the diff of this PR can be avoided by reusing the existing `TaxesOwedController`, `TaxesOwedForm`, and the `edit.html.erb` inside `/taxes_owed/`. We could add a bit of logic to `TaxesOwedController` to look for a constant (maybe in `StateInformationService`?) to denote if/when to show the instructions about Taxation reaching out, and then that bit could conditionally display on the existing screen. That way we don't need to maintain the duplicated code going forward.

Thoughts?

(Didn't really matter on which line I put this comment, just wanted to start a thread as opposed to ping-ponging comments.)

amanifarooque: unsolicited PM perspective here -- I would really advise _against_ duplicating global pages wherever we can, especially for a screen that's complicated like Taxes Owed. There could be global changes to this component to handle tax payments after the filing season is ended, for example, and it could be difficult to remember to keep these two in-sync.

I'll keep track of whichever route we go as a note for future tickets, may be worth pairing with CFA for a sustainable solution here. LMK how I can help!

mluedke2: I think knowing that there's definite things that will be added I think that upgrades my comment to a true boulder as opposed to an potential boulder

jachan: The existing page has conditional logic based on `current_state_code` for Maryland: 

https://github.com/codeforamerica/vita-min/blob/main/app/views/state_file/questions/taxes_owed/edit.html.erb#L6

It seemed like the codebase would get messy if I added additional `state_code` checks for NJ, so I followed the NC example of splitting this out into a separate class. However, I recognize your point (Amani and Matt) that this makes it difficult to maintain in the future, as any change in form logic needs to be made in multiple places. Here are some alternatives that I wanted to run by CFA (cc: @mpidcock)

1 - Preferred: Use state specific translation tags, similar to https://github.com/codeforamerica/vita-min/blob/main/app/views/state_file/questions/w2/edit.html.erb#L12

2 - Use the state information service to determine whether or not to show the additional text, similar to https://github.com/codeforamerica/vita-min/blob/c52c32ca86b5130b84795c64c3a6fba390a799d2/app/lib/submission_builder/return_header.rb#L19

3 - Follow the MD example:

```
<% if taxes_owed >= 400 && current_state_code == "NJ"%>
    <p><%= t('.underpayment_notice_html') %></p>
<% end %>
```

What is CFA's preferred way to do this?

mpidcock: @jachan Thanks for the thoughtful outline of options here! I agree that 1 would be _ideal_, however, seeing as there are already numerous examples along the lines of 3, I would be okay with that option. (And hopefully in the future we can revisit a bulk clean up, [similar to what this comment alludes to](https://github.com/codeforamerica/vita-min/blob/40c3456e703a74cdfbeab5a302c52f9dc0791394/app/views/state_file/questions/esign_declaration/edit.html.erb#L6).)

I took a peek through the code base for residual `state_code ==` checks, and the majority seem to be copy/template related. When we starting pursuing refactor projects, this is on my radar as something to iterate on 🤞

mpidcock (approved):
Wait, you already did the ideal implementation! Looks great!

app/views/state_file/questions/taxes_owed/edit.html.erb
```diff
@@ -7,6 +7,18 @@
     <p><%= t('.subtitle_html') %></p>
   <% end %>
 
+  <% underpayment_notice_key = "state_file.questions.taxes_owed.edit.underpayment_notice_#{current_state_code}_html" %>
```
jachan, line 10: [i18n context](https://www.i18next.com/translation-function/context) doesn't seem like the right fit here, since we don't have a default to use. It would need to be something like 

```
i18next.t('state_file.questions.taxes_owed.edit.underpayment_notice_html', { context: current_state_code} });
```

But then we'd need to define `state_file.questions.taxes_owed.edit.underpayment_notice_html` to provide a default of empty string in case the key doesn't exist. Seems more trouble than it's worth.

mluedke2: I guess it's a matter of whether it helps us to have this already set up when a second state needs this or not; but I don't know if another state will actually need this so this approach makes sense to me 👍

app/lib/navigation/state_file_nj_question_navigation.rb
```diff
@@ -63,7 +63,7 @@ def self.show_progress?(controller_class)
       Navigation::NavigationSection.new("state_file.navigation.nj.section_5", [
                                           Navigation::NavigationStep.new(StateFile::Questions::PrimaryStateIdController), # Footer
                                           Navigation::NavigationStep.new(StateFile::Questions::SpouseStateIdController), # Footer
-                                          Navigation::NavigationStep.new(StateFile::Questions::NjReviewController), # review should come before taxes owed / refund due screens
+                                          Navigation::NavigationStep.new(StateFile::Questions::NjReviewController), # REVIEW: should come before taxes owed / refund due screens
```
mluedke2, line 66: [pebble] why this change? was this a linter thing? I think the comment was more clear the way it was

mluedke2 (approved):
thanks so much for addressing the earlier concern! all good to me!

app/lib/navigation/state_file_nj_question_navigation.rb
jachan: I removed it, but the linter is running on auto-save so it added it back in. Thanks for catching it. Let me see how I can fix it...

PR #5384, opened 2025-01-17
GYR1-632: Add written langauge question to the interview scheduling page and map client communication and donation preferences to the 2024 13614-C form

embarnard:
## Link to pivotal/JIRA issue
- e.g. [link](https://codeforamerica.atlassian.net/browse/GYR1-632?atlOrigin=eyJpIjoiZGU4YTUxMDgxOGNmNDA4OGIxYmUwNWE4NTI5ZDk5NDkiLCJwIjoiaiJ9)

## What was done?
- Added written language question to the interview scheduling flow and mapped it to the 13614c form
## How to test?
- Select language on the interview scheduling page and make sure it is on the "Would you like to receive written communications from the IRS in a language other than English" line of the form
- can also test that the value looks correct on hub form if its set in the flow and that changing to spanish won't effect how its entered onto the form
- Before
- (no written language question)
Screenshot 2025-01-17 at 6 24 42 PM

- After
**`/interview-scheduling` page**
Screenshot 2025-01-17 at 6 23 13 PM
**drop down options:**
Screenshot 2025-01-17 at 6 23 17 PM
**mapping to 13614c form:**
Screenshot 2025-01-17 at 6 26 06 PM


app/forms/interview_scheduling_form.rb
```diff
@@ -9,8 +12,13 @@ def language_options
     I18n.backend.translations[I18n.locale][:general][:language_options].invert
   end
 
+  def written_language_options
+    I18n.backend.translations[I18n.locale][:general][:written_language_options].invert
```
embarnard, line 16: I'm assuming we are saving these languages by the their shortened code names like we do for the`preferred_interview_language` field, but for some of the test cases I see the whole string used like "Greek". Also took the liberty of searching the code names myself but not sure if there is a more official place to grab them

embarnard: Hmm might run into some issues because the hub form for the 13614c saves the language as a whole string, should we try to change the hub form or make the intake flow align with the hub form b/c of time crunch?

embarnard: i went down the path of saving it as a string but hated how clunky everything was so went back to the original way and just converted the codes to string in the pdf and hub form which seemed a lot cleaner

this was the other way:
```
class InterviewSchedulingForm < QuestionsForm
  set_attributes_for :intake,
                     :interview_timing_preference,
                     :preferred_interview_language,
                     :preferred_written_language

  def save
    preferred_written_language_not_english = [I18n.t("general.written_language_options.en")].include?(preferred_written_language) ? 'no' : 'yes' # will this work for spanish too?
    @intake.update(receive_written_communication: preferred_written_language_not_english)
    @intake.update(attributes_for(:intake))
  end

  def language_options
    I18n.backend.translations[I18n.locale][:general][:language_options].invert
  end

  def written_language_options
    # we are saving these as the full language string instead of the acronym like we do for `preferred_interview_language`
    # b/c the `preferred_written_language` field is an open field in the hub's 13614c form
    I18n.backend.translations[I18n.locale][:general][:written_language_options].values
  end

  def self.existing_attributes(intake)
    intake.preferred_interview_language ||= I18n.locale
    intake.preferred_written_language = existing_written_language_preference(intake)
    HashWithIndifferentAccess.new(intake.attributes)
  end

  private

  def self.existing_written_language_preference(intake)
    return intake.preferred_written_language if (I18n.t("general.written_language_options").values).include?(intake.preferred_written_language)

    if intake.written_language_preference_english?
      I18n.t("general.language_options.en")
    else
      # ensures if the user switches languages it will select the correct value
      opposite_locale = I18n.locale == :en ? :es : :en
      key_match = I18n.t("general.written_language_options", locale: opposite_locale).key(intake.preferred_written_language)
      if key_match.present?
        I18n.t("general.written_language_options.#{key_match}")
      else
        I18n.t("general.written_language_options.#{I18n.locale}", default: "general.written_language_options.en")
      end
    end
  end
end

```

spompea-cfa (approved):
approved. one small comment above

app/forms/hub/update_13614c_form_page3.rb
```diff
@@ -50,5 +50,10 @@ def save
       @client.intake.update(attributes_for(:intake))
       @client.touch(:last_13614c_update_at)
     end
+
+    def self.existing_attributes(intake)
+      intake.preferred_written_language = intake.preferred_written_language_string
+      super
+    end
   end
```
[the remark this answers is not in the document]
embarnard: @spompea-cfa ooo good catch! thanks for pointing that out

PR #5385, opened 2025-01-17
NJ 259 - remove custom CSS for household details review

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/259

## What was done?
From [this PR](https://github.com/codeforamerica/vita-min/pull/4902/files#diff-296123afaf4374655b7c97f7c22f4966c01cb8320dc72f62e047ea9f8f013aee), it seems like were hiding the last element when it was the "Edit" button. Now that the edit button has been removed globally from the component, we should remove our CSS fix so that it doesn't hide the last element in the household details.

## How to test?
Single - confirm section is not blank
MFJ - confirm spouse is present

- Before
image

- After
image

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -864,10 +864,6 @@ $state-colors: (
       background-color: $color-nj-blue-darker;
     }
 
-    .review-header #household-info div:last-of-type {
-      display: none;
-    }
-
     .nj-review-outer .reveal hr {
       border:0;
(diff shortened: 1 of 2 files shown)


mluedke2 (approved):
that's a wild side effect!

a [dust] wondering if we should add an expectation to our E2E test that looks for the "Your name" string on the review page, that would have caught this bug?

arinchoi03 (approved):

PR #5390, opened 2025-01-17
FYST-1644 Make sure request record is persisted before navigating to verification page

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1644
- [Sentry Error](https://codeforamerica.sentry.io/issues/6221991756/?alert_rule_id=1046574&alert_type=issue&notification_uuid=0a7fce48-0d4b-4b96-8ad1-4f6c7bde7aa1&project=1880321)

## What was done?
- We tried to recreate the bug locally by putting a `sleep(5)` on a `before_create` hook for StateFileArchivedIntakeRequest and was able to see the same error. Our suspicion is that the db creation happens in the background and is not completed by the time we arrive at the VerificationCodeController `edit` action.
- On demo, when we refreshed the page (after seeing the error), we saw the page load successfully (few seconds later). This further supports our theory.
- We are mostly certain that saving the record to a variable & referencing the record forces us to wait for the record to be fully created. 
## How to test?
- We weren't sure how to recreate a slow database operation in a test; if anyone has ideas let us know!


anisharamnani:
🤔 this is a really interesting problem & clever solution! thanks @jenny-heath & @arinchoi03 for finding a solution. 

i agree--i can't think of way to test this, but i would say adding a comment here would be helpful. i think this mistake would be easy to revert, since it's not clear that there is additional complexity when it comes to setting a variable.

PR #5395, opened 2025-01-20
Az 8879 validations

embarnard:
## Link to pivotal/JIRA issue

## What was done?
- validates that AZAdjGrossIncome and BalanceOfTaxDue are not longer than 15 digits
- validates that AzIncTaxWithheld, RefundAmt and AmtOwed are not longer than 15 digits and is a positive 
## How to test?
- use numbers larger than 15 digits for these values or negative for applicable fields


app/lib/submission_builder/document.rb
```diff
@@ -106,6 +106,22 @@ def add_non_zero_float_value(xml, elem_name, line)
...
+
+    def add_non_negative_int_value(xml, elem_name, line)
+      # validates USAmountNNType, 15 digit limit
+      value = calculated_fields.fetch(line).to_i
+      if value.present? && value.to_s.length <= 15 && value >= 0
+        xml.send(elem_name, value)
+      end
+    end
+
     def add_element_if_present(xml, tag, line_id)
```
tahsinaislam, line 125: I was trying to do something similar with md but Tiffany and Erica mentioned they would want to pass value that are negative and over the digit limit to investigate those failures. Maybe worth checking in with them?

embarnard: Ooo okay yes thanks for bringing this up, I will check in with them

embarnard:
Closing this PR since we don't want to add validations for numbers longer than 15 digits or negative numbers because we would want to investigate those cases

PR #5401, opened 2025-01-21
Fyst [1519&1518] add mailing address validation step to flow

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1519
- https://codeforamerica.atlassian.net/browse/FYST-1518

## What was done?
- adds mailing address validation step to get your pdf flow
## How to test?
- in heroku make sure the `get_your_pdf` flipper tag is enabled and that a statefile archived intake from the seeder file is used the go though the get your pdf flow til this page is hit.
Screenshot 2025-01-23 at 12 45 26 PM


app/controllers/state_file/archived_intakes/mailing_address_validation_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+        @form = MailingAddressValidationForm.new(addresses: @addresses, current_address: current_archived_intake.full_address)
+      end
+
+      def update
+        @form = MailingAddressValidationForm.new(mailing_address_validation_form_params, addresses: @addresses, current_address: current_archived_intake.full_address)
+        @addresses = address_challenge_set
+
+        if @form.valid?
+          create_state_file_access_log("correct_mailing_address")
+          # this should take us to the download
```
anisharamnani, line 21: I like the pattern that we did in our ticket (which was @mrotondo's suggestion) of adding the ticket that will fix this.

app/controllers/state_file/archived_intakes/mailing_address_validation_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+          redirect_to root_path
```
anisharamnani, line 22: I like the pattern that we did in our ticket (which was @mrotondo's suggestion) of adding the ticket that will fix this.

app/controllers/state_file/archived_intakes/mailing_address_validation_controller.rb
tahsinaislam, line 21: do we want to add something to the session saying this was successful for the right people to access the download page?

spec/features/state_file/prior_year_access_spec.rb
```diff
@@ -30,9 +30,9 @@
       fill_in I18n.t("state_file.archived_intakes.identification_number.edit.ssn_label"), with: "123456789"
       click_on I18n.t("general.continue")
 
-      # TODO: https://codeforamerica.atlassian.net/browse/FYST-1518
-      # need to change to address path
-      expect(current_path).to eq(root_path)
+      expect(current_path).to eq(state_file_archived_intakes_edit_mailing_address_validation_path)
+
+
```
tahsinaislam, line 35: should you add more to this like selecting the right address and pressing continue?

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,80 @@
...
   end
 
   def increment_failed_attempts
     super
-    if attempts_exceeded? && !access_locked?
-      lock_access!
+    lock_access! if attempts_exceeded? && !access_locked?
+  end
+
+  def fake_addresses
```
anisharamnani, line 38: why did you choose to add the logic for fetching random addresses on the request class itself? as opposed to creating a separate class/service for this?

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,80 @@
...
+    [fake_address_1, fake_address_2]
+  end
+
+  private
+
+  def populate_fake_addresses
+    self.fake_address_1, self.fake_address_2 = fetch_random_addresses
```
anisharamnani, line 45: also sorry if this is obvious, and i missed it, how do you determine if the address if NY or AZ?

app/controllers/state_file/archived_intakes/mailing_address_validation_controller.rb
anisharamnani: do you want to add session data for the mailing challenge to be completed?

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      it "renders the edit template with a new MailingAddressValidationForm" do
+        get :edit
+
+        expect(assigns(:form)).to be_a(StateFile::ArchivedIntakes::MailingAddressValidationForm)
+        expect(response).to render_template(:edit)
+      end
+    end
+
+    it "redirects to the lockout path when the request is locked" do
```
anisharamnani, line 45: what's the difference between this and the test on line 25?

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        post :update, params: {
+          state_file_archived_intakes_mailing_address_validation_form: { selected_address: intake.full_address, addresses: controller.address_challenge_set}
+        }
+        expect(assigns(:form)).to be_valid
+
+        access_log = StateFileArchivedIntakeAccessLog.last
+        expect(access_log.state_file_archived_intake_request).to eq(current_request)
+        expect(access_log.event_type).to eq("correct_mailing_address")
+
+        #this will be the download page
```
anisharamnani, line 83: adding a TODO with the ticket here as well!

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      get :edit
+
+      expect(response).to redirect_to(root_path)
+      expect(StateFileArchivedIntakeAccessLog.last.event_type).to eq("unauthorized_mailing_attempt")
+    end
+  end
+
+  describe "PATCH #update" do
+    context "with a valid chosen address" do
+      it "creates an access log and redirects to the root path" do
```
anisharamnani, line 73: 🤔 i may be missing something here, but how come the session data doesn't need to be set for this test to pass?

spec/controllers/state_file/archived_intake/archived_intake_controller_spec.rb
```diff
@@ -3,7 +3,8 @@
 describe StateFile::ArchivedIntakes::ArchivedIntakeController, type: :controller do
   let(:ip_address) { '192.168.0.1' }
   let(:email_address) { 'test@example.com' }
-  let!(:request_instance) { create :state_file_archived_intake_request, ip_address: ip_address, email_address: email_address }
+  let(:intake) {create :state_file_archived_intake}
```
anisharamnani, line 6: [nit] can we call this an `archived_intake` since we use `intake` to refer to a completely different db model.

anisharamnani:
👋  thanks for doing this work! sorting out how to make the fake addresses seems to be quite complicated! 

i've left a bunch of comments. let me know if its easier to pair and walk through them.

app/models/state_file_archived_intake_request.rb
DrewProebstel: because this is only used here, If it was used somewhere else  I would have made a service.

app/models/state_file_archived_intake_request.rb
DrewProebstel: On line 66. There is a NY and an AZ csv.

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
DrewProebstel: it is set on lines 15 and 16.

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
DrewProebstel: nothing oops.

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,80 @@
...
+    addresses = CSV.read(file_path, headers: false).flatten
+    addresses.sample(2)
+  end
+
+  def determine_csv_file_path
+    if Rails.env.development? || Rails.env.test?
+      Rails.root.join('app', 'lib', 'challenge_addresses', 'test_addresses.csv')
+    else
+      file_path = Rails.root.join('tmp', 'challenge_addresses.csv')
+      download_file_from_s3(file_path) unless File.exist?(file_path)
```
mpidcock, line 59: this should be called from `fetch_random_addresses`, not from `determine_csv_file_path`. currently it looks like the csv is not getting captured and used from s3

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,80 @@
...
+      file_path
+    end
+  end
+
+  def download_file_from_s3(file_path)
+    bucket = select_bucket
+    file_key = "challenge_addresses/#{state_file_archived_intake&.mailing_state&.downcase}_addresses.csv"
```
mpidcock, line 66: I think it would be cleaner to set this in `fetch_random_addresses` and pass it into the method

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,80 @@
...
+
+  private
+
+  def populate_fake_addresses
+    self.fake_address_1, self.fake_address_2 = fetch_random_addresses
+  end
+
+  def fetch_random_addresses
+    file_path = determine_csv_file_path
+    addresses = CSV.read(file_path, headers: false).flatten
```
mpidcock, line 50: you'll need to either read from a local file path OR download from s3

mpidcock (commented):
Let me know if you want to talk through any of my comments

app/models/state_file_archived_intake_request.rb
```diff
@@ -5,6 +5,8 @@
 #  id                            :bigint           not null, primary key
 #  email_address                 :string
 #  failed_attempts               :integer          default(0), not null
+#  fake_address_1                :string
```
anisharamnani, line 8: can you also explain the logic of storing the fake addresses on the record itself as opposed to sampling two random ones from the CSV?

app/models/state_file_archived_intake_request.rb
```diff
@@ -15,19 +17,82 @@
...
+        file_path = File.join(Rails.root, "tmp", File.basename(file_key))
+
+        download_file_from_s3(bucket, file_key, file_path) unless File.exist?(file_path)
+      end
+      addresses = CSV.read(file_path, headers: false).flatten
+      addresses.sample(2)
+    end
+  end
+
+  def download_file_from_s3(bucket, file_key, file_path)
```
anisharamnani, line 66: this is the [line](https://github.com/codeforamerica/vita-min/pull/5401/files#r1927755452) you're referring to? 

i once again must be misunderstanding how do you choose NY or AZ

app/controllers/state_file/archived_intakes/archived_intake_controller.rb
```diff
@@ -6,18 +6,32 @@ def current_request
...
+      end
+
       def create_state_file_access_log(event_type)
         StateFileArchivedIntakeAccessLog.create!(
           event_type: event_type,
           state_file_archived_intake_request: current_request
         )
       end
 
+      def address_challenge_set
```
anisharamnani, line 20: why is the address_challenge_set defined here? are we trying to hide the method? 

🤔 if that's the case, since it's not shared across controllers, how do you feel about putting it at the model level?

app/models/state_file_archived_intake_request.rb
mpidcock: the idea is that we don't want people to be able to refresh or start a new session and see different fake addresses but the same real address, but that reasoning is not obvious, so I agree that it should be capture somewhere, perhaps a comment on the `populate_fake_addresses` method

app/models/state_file_archived_intake_request.rb
mpidcock: it's now line 55: `"#{state_file_archived_intake.mailing_state.downcase}_addresses.csv"`

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        access_log = StateFileArchivedIntakeAccessLog.last
+        expect(access_log.state_file_archived_intake_request).to eq(current_request)
+        expect(access_log.event_type).to eq("incorrect_mailing_address")
+        expect(current_request.access_locked?).to eq(true)
+        expect(response).to redirect_to(state_file_archived_intakes_verification_error_path)
+      end
+    end
+
+    context "without a chosen address" do
```
anisharamnani, line 97: can you explain the case when this would happen where a post to update would occur & we wouldn't want to lock them out?

mpidcock (approved):

app/controllers/state_file/archived_intakes/mailing_address_validation_controller.rb
anisharamnani: i have a similar comment in my comments, so re-opening tahsina's here.

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
anisharamnani: handles the case where the user is hits submit without selecting an addresss

spec/controllers/state_file/archived_intake/mailing_address_validation_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    context "with a valid chosen address" do
+      it "creates an access log and redirects to the root path" do
+        post :update, params: {
+          state_file_archived_intakes_mailing_address_validation_form: { selected_address: intake.full_address, addresses: controller.address_challenge_set}
+        }
+        expect(assigns(:form)).to be_valid
+
+        access_log = StateFileArchivedIntakeAccessLog.last
+        expect(access_log.state_file_archived_intake_request).to eq(current_request)
+        expect(access_log.event_type).to eq("correct_mailing_address")
```
anisharamnani, line 74: can you add a test verifying that the session data has been updated?

app/models/state_file_archived_intake_request.rb
anisharamnani: i don't think this needs to be handled in this PR, but...even if it's not used elsewhere, i think a more object oriented approach would be to have this in it's own class. we could also then test some of meatier logic in these methods individually. 

...something to keep in mind for the future!

anisharamnani (approved):
thanks @DrewProebstel! this is a monster of a PR, and therefore a difficult challenge. thank you for handling this!

PR #5403, opened 2025-01-22
FYST-1650 Allow non-AdminRole accounts to be logged into using @codeforamerica oauth

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1650

#### Context
- the hub role `StateFileNjStaff` is the only role with access to view NJ submissions
- the standard `AdminRole` given to CfA employees only has access to non-NJ submissions
- In order for some select CfA staff to have access to NJ's hub view, we have provisioned some email addresses intended only for having the NJ role and accessing the hub (e.g. `*-njhub@codeforamerica.org`)
- Users with @codeforamerica.org emails were unable to log in as anything other than an Admin

#### Changes
- When logging in with oauth, look for a user to login to with **any** role type, but **prioritize** "AdminRole" if by some wild circumstance there are multiple
- Preserved existing logic as best I could

## How to test?
- locally I created nj user with nj email and logged in with it
   - this test is challenging to repeat in heroku because it requires the url to be a registered, authorized uri with our Google oauth account
      - update: I was able to add this specific heroku instance to enable oauth testing
   - it can be tested in demo by anyone with a `-njhub` email
- also still could login with normal email to admin account
- Risk: I think there is a very mild risk that this alters the login functionality to be slightly more permissive
   - ihowever, it is allowing accounts with cfa (& gyr) emails to be non-admin accounts, which is more restrictive access, so not a security concern
#### digging into data
I wanted to know if there are any active users that could be affected by this change, other than the known NJ users, in either prod or demo.

There are no users with duplicate emails in either prod or demo 👍 
```
User.all.group_by(&:email).select { |email, matching_users| matching_users.count > 1 }
=> {}
```

I also wanted to know if there are other existing non-Admin accounts with CFA or GYR (oauth required) domains that are in the same situation where they most likely cannot log in
```
User.where("email LIKE ?", "%@codeforamerica.org").where(suspended_at: nil).each { |user| puts user.email + ", " + user.role_name if user.role_name != "Admin" }
User.where("email LIKE ?", "%@getyourrefund.org").where(suspended_at: nil).each { |user| puts user.email + ", " + user.role_name if user.role_name != "Admin" }
```

- I discovered no prod accounts, but ~20 CFA accounts that were not admins
- a majority of them are using "+" to modify their email, from before oauth was used, and can't be logged into anymore anyway
- there are a few client success users in demo that may be affected, so I will reach out to them separately about logging in to the hub (Andrew, Elena, and Alyssa)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -322,10 +322,10 @@ class User < ApplicationRecord
     return nil unless google_login_domain?(email) && google_login_domain?(auth_hash.extra.id_info["hd"])
 
     matching_users = User.where(email: email, external_provider: [nil, oauth2_provider_name], external_uid: [nil, auth_hash['uid']], suspended_at: nil)
+    # NOTE: Duplicate emails should never happen because there is a `validates_uniqueness_of` email constraint
+    #   However, I wanted to preserve the previous implementation as closely as possible, which means prioritizing
+    #   returning a user account with AdminRole, if one exists
     user = matching_users.where(role_type: "AdminRole").first || matching_users.first
-    if matching_users.count > 1
-      Rails.logger.warn("Found multiple hub accounts for #{email}, logging into #{user.role_type} account")
-    end
     user.update!(external_provider: oauth2_provider_name, external_uid: auth_hash['uid']) if user.present? && user.external_uid.nil?
     user
   end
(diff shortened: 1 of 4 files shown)


jnf (approved):
let's go!

PR #5404, opened 2025-01-22
FYST-1603 updates to NY intake flow

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1603

## What was done?
- Remove get started button on landing page for NY and display some copy about not supporting it anymore
- Redirect from NY intake pages to landing page
  - note about this: it's already the case that pages in the intake flow require an intake in the session, so the only real purpose of redirecting is for the edge case where a client logs in using an existing NY intake and tries to manually visit the pages. this seems fairly unlikely but also it's fine to implement it temporarily until we delete more NY code?
  - update to this note: i only just realized that we won't have any ny intakes to log in with because we deleted them :( so maybe it was all for naught
  - another update: we're now just skipping the tests for ny controllers
## How to test?
- Added a feature spec
- Before
image

- After
image
image

### DIFF
--- a/app/controllers/concerns/state_file/state_file_intake_concern.rb
+++ b/app/controllers/concerns/state_file/state_file_intake_concern.rb
@@ -38,8 +38,7 @@ module StateFile
     end
 
     def redirect_deprecated_state
-      # this just handles controllers with "ny" in the name; other controllers are shared and should redirect to login
-      if params[:controller].include?("ny")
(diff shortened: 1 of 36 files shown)


anisharamnani (approved):
jenny and i paired on reviewing this. thanks jenny!

PR #5405, opened 2025-01-22
add AWS IP ranges file to config, and load it into trusted proxies during initialization

mrotondo:
## JIRA Story
https://codeforamerica.atlassian.net/browse/FYST-1644
- Yes, but I'm not sure this will work on heroku

## What was done?
- It turns out we have (at least as far as I can tell, based on IPs associated with state file intakes on demo and prod) never written down correct client IPs - rather we've been recording the IPs of AWS loadbalancers and other proxies. This is because ActiveDispatch has a mechanism to ignore "trusted proxies" by filtering out their IPs (see logic in `ActionDispatch::RemoteIp`) from the list of IPs in the `X_FORWARDED_FOR` header, but we weren't providing the list of AWS IPs, so their proxies weren't being filtered out. This PR adds those IPs (downloaded [straight from AWS](https://docs.aws.amazon.com/vpc/latest/userguide/aws-ip-ranges.html#aws-ip-download) to that filter, and we've tested it on staging to confirm that `request.remote_ip` is now correctly set to the actual client IP.
## How to test?
- This fixes the intermittent failure on staging where an error page would pop up after entering an email address into the archived intakes login flow. This was happening because the "client ID" was changing with each request, due to us not filtering out AWS proxies and interpreting a load balancer IP as the client, causing the database query looking for an Archived Intake Request with the current IP to fail. This change is currently pushed to staging, so you can go verify that you're able to get through the Archived Intake email page 100% of the time there to verify that this change does what it's supposed to.
- Before
```
[vita-min-staging-web]: request.remote_ip: 18.68.29.117
[vita-min-staging-web]:   HTTP_X_FORWARDED_FOR: 157.131.197.166, 18.68.29.117, 10.210.0.208
```
- After
```
[vita-min-staging-web]: request.remote_ip: 157.131.197.166
[vita-min-staging-web]:   HTTP_X_FORWARDED_FOR: 157.131.197.166, 18.68.29.117, 10.210.0.208
```


anisharamnani (approved):
this is _very_ interesting! thanks @mrotondo (& i believe @arinchoi03) for find a solution to this.

arinchoi03 (approved):

PR #5411, opened 2025-01-22
NJ 240 - calculate 2% of medical expenses for content

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/240

## What was done?
Add calculation for 2% of income and display on medical expenses page

## How to test?
Use any persona with income and visit medical expenses page

Screenshot 2025-01-22 at 15 06 14

### DIFF
--- a/app/models/state_file_nj_intake.rb
+++ b/app/models/state_file_nj_intake.rb
@@ -203,5 +203,10 @@ class StateFileNjIntake < StateFileBaseIntake
     false
   end
 
+  def medical_expenses_threshold
+    nj_gross_income = calculator.lines[:NJ1040_LINE_29].value
+    (nj_gross_income * 0.02).floor
+  end
+
 end
 
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -3258,7 +3258,7 @@ es:
           description_html: |-
             <p>If you had at least %{two_percent_of_income} in non-reimbursed medical expenses, you are eligible for a deduction. Enter the full amount of your non-reimbursed expenses here. We’ll deduct the right amount for your situation.</p>
             <p>If your medical expenses were less than %{two_percent_of_income}, you are not eligible for the deduction. Click "Continue."</p>
-            
+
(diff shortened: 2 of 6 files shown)


mmazanec22 (approved):
Looks good! And seems like a nice clarification

config/locales/es.yml
```diff
@@ -3256,6 +3256,9 @@ es:
       nj_medical_expenses:
         edit:
           description_html: |-
+            <p>If you had at least %{medical_expenses_threshold} in non-reimbursed medical expenses, you are eligible for a deduction. Enter the full amount of your non-reimbursed expenses here. We’ll deduct the right amount for your situation.</p>
+            <p>If your medical expenses were less than %{medical_expenses_threshold}, you are not eligible for the deduction. Click "Continue."</p>
+
```
aloverso, line 3261: What is the process for Spanish content updates, knowing that will be on the CFA Transifex side?

app/models/state_file_nj_intake.rb
```diff
@@ -203,5 +203,10 @@ def ask_for_signature_pin?
     false
   end
 
+  def medical_expenses_threshold
+    nj_gross_income = calculator.lines[:NJ1040_LINE_29].value
```
mpidcock, line 207: [dust] since this is used in multiple places you could consider making a calculator method for "gross income" - just as a readable/reusable thing

config/locales/es.yml
mpidcock: I'm checking in on this, for now english is ok

mpidcock (approved):

mluedke2:
FYI, it sounds like the issue flagged above was caused by a user doing something interesting with their browser tabs -- probably navigating as far as the medical expenses page, then in a different tab signing out/starting over, and then refreshing or submitting medical expenses on the first tab. They were able to recover and have a successful submission after this error.

So it seems like a generic issue to address with app state and not something specific to NJ Medical Expenses...

PR #5428, opened 2025-01-23
Nj 217 income warnings

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/217

## What was done?
- I decided to edit the format of the Box 14 data in `StateInformationService` so that the types might also have a limit attached, so there is one source of truth
- Then `StateFileW2` has a class method called `find_limit` that gets that limit, and other places in code reference `find_limit`
## How to test?
- With the Blackburn persona, two W2s will display with warnings. This is because these two AC apply there:

```
GIVEN the primary taxpayer on the return has 2 or more W2s
WHEN the UI/WF/SWF and UI/HC/WD field are blank on a given W2
THEN we should show the "double check" error on that W2 on the income form overview page

GIVEN the primary taxpayer on the return has 2 or more W2s
WHEN the FLI field is blank on a given W2
THEN we should show the "double check" error on that W2 on the income form overview page
```

Blackburn:
image


config/locales/en.yml
```diff
@@ -2924,6 +2924,7 @@ en:
           title: Here are the income forms we transferred from your federal tax return.
           unemployment_title: Unemployment benefits (1099-G)
           w2s_title: Jobs (W-2)
+          warning: "⚠️ We need to double-check some information"
```
aloverso, line 2927: [dust] are there any potential issues with using an emoji here instead of an svg or something?

spec/controllers/state_file/questions/income_review_controller_spec.rb
```diff
@@ -100,6 +104,120 @@
...
       end
     end
+
+    context "when no W2 box 14 warnings" do
+      let(:intake) { create(:state_file_nj_intake) }
+      
+      shared_examples "does not display W2 warnings" do
+        it "does not display W2 warnings" do
+          get :edit, params: params
+          expect(response.body).not_to have_text "We need to double-check some information"
```
aloverso, line 114: [pebble] I think the recommended pattern would be to reference the `I18n.t` rather than the string directly

spec/controllers/state_file/questions/income_review_controller_spec.rb
```diff
@@ -100,6 +104,120 @@
...
+
+    context "when W2 box 14 warnings are present" do
+      let(:intake) { create(:state_file_nj_intake) }
+
+      shared_examples "displays at least one W2 warning" do
+        it "displays at least one W2 warning" do
+          get :edit, params: params
+          expect(response.body).to have_text "We need to double-check some information"
+        end
+      end
```
aloverso, line 172: [dust/pebble] should we test for the correct number of W2 warnings? "at least one" feels imprecise.

I think you can do this by passing a `count` to the matcher

aloverso (approved):
Nice work! Well done navigating a pretty complex space. Just had a couple small comments

tahsinaislam (approved):
lgtm!

spec/controllers/state_file/questions/income_review_controller_spec.rb
mluedke2: ooh i like that. I edited it to expect exactly one warning. thanks!

config/locales/en.yml
mluedke2: I don't think it's an issue for this project, but it's a little bit up to the browser how exactly to render the emoji as opposed to an image we know will always look exactly the same. And for accessibility we don't want to rely on it, but we are ok since the rest of our text gives the actual information. So using this should be all good.

PR #5431, opened 2025-01-24
GYR1-616 Add Hub additional notes p5 + pertinent mappings

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-616

## Is PM acceptance required?

## What was done?

- Added an editable 14-C page 5 into the Hub, which has just the "Additional 
  Notes/Comments" field. This field should "start out" with whatever comments 
  the client entered into the FinalInfoController screen back during the intake 
  flow.
- The mapping was also added so that the comments/notes get properly written to 
  the PDF.
- For this PR, mappings were also added to write out dependent records onto the 
  5th page when there are 4 or more dependents. 
  - This includes the new dependent IPPIN field as well.

## Technical Notes

- `additional_notes_comments` has been created, which is a `text` column, to 
  replace the legacy `final_info` `string` column. I think `text` is the 
  preferable data type here.
- It looks like `additional_info` wasn't really being used, so it's no longer
   included in the string that gets written out to the PDF.

## How to test?

- Ensure that the text entered into the FinalInfoController shows up when 
  navigating to page 5 in the Hub
- Ensure that if changes/additions are made to the text in the Hub, it will all 
  show up properly on the PDF.
- Enter in at least 5 dependents and ensure that the 4th and 5th records show up 
  properly on page 5 of the PDF. 
  - Switch up the values and spot-check that no wires were crossed in terms of 
    the order of how values were written out on the PDF.
- Ensure it all works when there's *only* a note but only 3 or fewer dependent 
  records.

## Example

image

### DIFF
--- a/app/views/hub/clients/edit_13614c_form_page5.html.erb
+++ b/app/views/hub/clients/edit_13614c_form_page5.html.erb
@@ -1,0 +1,31 @@
+<% @title = 'Additional Notes/Comments' %>
+<% content_for :page_title, @title %>
+<% content_for :card do %>
+  <div class="form_13614c slab slab--not-padded">
+    <h1><%= @title %></h1>
+    <p>Last client 13614-C
+      update: <%= @client.last_13614c_update_at&.in_time_zone("America/Los_Angeles")&.strftime("%b %-d %l:%M %p") %></p>
+
+    <%= form_with model: @form,
+                  url: edit_13614c_form_page5_hub_client_path,
+                  method: :put, local: true, builder: VitaMinFormBuilder,
+                  html: { class: 'form-card form-question-13614-page5' } do |f| %>
+      <div>
+        <hr style="margin-top: 0;"/>
+        <%= f.cfa_textarea :additional_notes_comments, '', options: {rows: '2', columns: 100} %>
+      </div>
+
+      <div style="display: flex; justify-content: space-between;">
+        <div>
+          <%= f.submit t("general.save"), class: "button button--cta"%>
+          <%= f.submit t("general.save_and_exit"), class: "button button--cta"%>
+          <%= link_to t("general.cancel"), cancel_13614c_hub_client_path, class: "button button--danger",
+                      data: { confirm: t("general.confirm_exit_without_saving") } %>
+        </div>
+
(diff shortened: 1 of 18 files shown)


app/views/hub/clients/edit_13614c_form_page5.html.erb
```diff
@@ -0,0 +1,31 @@
+<% @title = 'Additional Notes/Comments' %>
+<% content_for :page_title, @title %>
+<% content_for :card do %>
+  <div class="form_13614c slab slab--not-padded">
+    <h1><%= @title %></h1>
+    <p>Last client 13614-C
+      update: <%= @client.last_13614c_update_at&.in_time_zone("America/Los_Angeles")&.strftime("%b %-d %l:%M %p") %></p>
```
embarnard, line 7: Where is this timezone coming from?

spec/features/hub/edit_13614c_spec.rb
```diff
@@ -652,5 +652,29 @@
...
+        click_on 'Edit 13614-C'
+      end
+
+      within '.form_13614c-page-links', match: :first do
+        click_on '5'
+      end
+      header = 'Additional Notes/Comments'
+      expect(page).to have_text header
+
+      note = 'It was very late and everyone had left the café except an old man who sat in the shadow the leaves of the tree made against the electric light.'
```
embarnard, line 668: 🥹

embarnard (approved):
Looks good!

app/views/hub/clients/edit_13614c_form_page5.html.erb
spompea-cfa: that's a good question. it's just what was in the preexisting ERBs in the editable 14-C flow. now that you mention it, i do wonder if it being hardcoded like that is the best idea. but afaik, that's the way it was for ty2023. could be worth following up on though.

spompea-cfa:
@embarnard also, i meant to have the write-up ready before putting into review. i've added that now.

spec/features/hub/edit_13614c_spec.rb
spompea-cfa: agreed! great story.

PR #5432, opened 2025-01-24
Fyst 1581 md add 1099 r subtraction UI

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1581

## What was done?
- Add the MD 1099R page, and a concern that can be reused which enables the repeated question logic
## How to test?
- I don't think there's an MD persona with multiple 1099-Rs right now - we should create one with new/realistic/expected DF output for 1099Rs. But once we have that, you should go through the MD flow with that persona (and with the `show_retirement_ui` flipper flag turned on) and verify that the pages show up, and can be edited and that those answers are persisted (they show up if you go back through the flow.
  - Pay special attention to the behavior when using the continue and back buttons to go back and forth through the 1099Rs.
image


tahsinaislam (approved):
I think you just have on test to add the controller to? But lgtm!

app/controllers/state_file/questions/md_retirement_income_subtraction_controller.rb
jenny-heath, line 1: i wonder if all this could go in a parent controller called like `RetirementIncomeSubtractionController` and that way each state-specific child controller could just define something like this:
```
def followup_class
  StateFileMd1099RFollowup
end
```
and then line 38 could be `@state_file_1099r.update(state_specific_followup: followup_class.create)`
and then rest of these methods, since they are generic 1099r stuff, could be shared across states doing retirement income subtraction pages

spec/controllers/state_file/questions/md_retirement_income_subtraction_controller_spec.rb
jenny-heath, line 1: sorry to keep requesting more work but these tests would be so great as shared examples that we could use for the various states doing the exact same navigation pattern

app/controllers/state_file/questions/md_retirement_income_subtraction_controller.rb
mrotondo: Good idea, done!

spec/controllers/state_file/questions/md_retirement_income_subtraction_controller_spec.rb
mrotondo: Also good idea, also done!

app/controllers/state_file/questions/md_retirement_income_subtraction_controller.rb
```diff
@@ -0,0 +1,9 @@
+module StateFile
+  module Questions
+    class MdRetirementIncomeSubtractionController < RetirementIncomeSubtractionController
+
+      def followup_class = StateFileMd1099RFollowup
```
jenny-heath, line 5: wow this...works?

spec/forms/state_file/md_retirement_income_subtraction_form_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    context "validations" do
+      context "with invalid amounts" do
+        let(:params) { {} }
+        it "returns false and adds an error to the form" do
+          expect(form.valid?).to eq false
+          expect(form.errors).to include(:income_source, :service_type)
+        end
+      end
+
+      context "with valid amounts" do
```
jenny-heath, line 19: nit: i assume this description is copied because "military" is not an amount

app/controllers/state_file/questions/md_retirement_income_subtraction_controller.rb
mrotondo: Learned it from Kevin ⛳

jenny-heath (approved):
this looks great! i realized i would have made one other request if i'd thought of it: because the return to review stuff is so complicated, we have a feature spec `spec/features/state_file/review_page_spec.rb` that kind of tests the whole flow through review->income review->income form etc. and i think it would make sense to add 1099r to that spec. happy for it to be a follow-up though!

PR #5433, opened 2025-01-24
NJ 267 - route ineligible both owner/renter out of flow after ineligible page

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/267

## What was done?
- when user is ineligible due to low income, they should proceed to use tax page after ineligible page regardless of whether they are a "both"

## How to test?
- use bon jovi health dependents persona (NOT disabled)
- go through to property tax credit screens
- select you were "both" a homeowner and renter

### DIFF
--- a/app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
+++ b/app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
@@ -60,6 +60,10 @@ module StateFile
         options = {}
         options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
 
+        if Efile::Nj::NjPropertyTaxEligibility.ineligible?(current_intake)
+          return StateFile::NjPropertyTaxFlowOffRamp.next_controller(options)
+        end
+
         if current_intake.household_rent_own_both? && params[:on_home_or_rental] != "rental"
           NjTenantEligibilityController.to_path_helper(options)
         else
--- a/spec/controllers/state_file/questions/nj_ineligible_property_tax_controller_spec.rb
+++ b/spec/controllers/state_file/questions/nj_ineligible_property_tax_controller_spec.rb
@@ -133,6 +133,15 @@ RSpec.describe StateFile::Questions::NjIneligiblePropertyTaxController do
       end
     end
 
+    context "when indicated both rent and own and ineligible" do
+      let(:intake) { create :state_file_nj_intake, household_rent_own: "both" }
+      it "next path is next_controller for property tax flow" do
+        allow(Efile::Nj::NjPropertyTaxEligibility).to receive(:ineligible?).and_return true
+        get :edit, params: {on_home_or_rental: "home"}
+        expect(subject.next_path).to eq(StateFile::NjPropertyTaxFlowOffRamp.next_controller({}))
+      end
+    end
+
     context "when not both rent and own" do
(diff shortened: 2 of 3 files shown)


app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
```diff
@@ -60,6 +60,10 @@ def next_path
         options = {}
         options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
 
+        if Efile::Nj::NjPropertyTaxEligibility.ineligible?(current_intake)
+          return StateFile::NjPropertyTaxFlowOffRamp.next_controller(options)
+        end
```
aloverso, line 65: I thought that adding this early return was more readable than modifying the `if` below to include a `&& !Efile::Nj::NjPropertyTaxEligibility.ineligible?(current_intake)` - both for length and for the cognitive load of parsing "not ineligible". 

But I acknowledge that makes it a little duplicative. Let me know what you think

embarnard: I'm always a fan of readability!

embarnard: Nice controller names! Hadn't seen `NjPropertyTaxFlowOffRamp` yet

embarnard (approved):
Looks good!

app/controllers/state_file/questions/nj_ineligible_property_tax_controller.rb
jachan: I like this early return - do not find it duplicative!

spec/controllers/state_file/questions/nj_ineligible_property_tax_controller_spec.rb
```diff
@@ -133,6 +133,15 @@
       end
     end
 
+    context "when indicated both rent and own and ineligible" do
+      let(:intake) { create :state_file_nj_intake, household_rent_own: "both" }
+      it "next path is next_controller for property tax flow" do
```
jachan, line 138: [Pebble] Should this be something like `it sends the taxpayer to the OffRamp` instead?

jachan (approved):
Left a pebble about a test's name, otherwise approved!

spec/controllers/state_file/questions/nj_ineligible_property_tax_controller_spec.rb
aloverso: Hm. I'm following the patterns in other tests here (but open to renaming all of them if we want). But I actually think that OffRamp is a little too specific here and I'd rather name it describing the thing it is: "the next controller for the flow". Whereas if we ever refactored the name of the OffRamp to something else, these tests would get out-of-date and confusing. You alright with that?

jachan: I'd prefer to describe the behavior for the taxpayer, so anyone reading the test can understand what expected behavior is. In this case - we are offboarding the taxpayer when they are ineligible. Is it possible to convey that in the `it` clause? This is a small thing, happy to go with whatever you think is best, but wanted to share my preference for more descriptive documentation here rather than just "we go to the next step".

mluedke2 (commented):
👍

PR #5442, opened 2025-01-24
get all box 14 fields within fieldset

mluedke2:
## Link to pivotal/JIRA issue

followup to https://github.com/newjersey/affordability-pm/issues/217 for https://github.com/newjersey/accessibility-pm/issues/27

## What was done?
- the `<fieldset>` was within the iterator creating fields instead of outside of it
- since the other recommendations in the accessibility ticket are design changes or shared components I think we should keep that ticket open but address separately

## How to test?
- Open up W2 editing, and while everything looks and behaves the same you can see the difference with Inspector
- Before

image

- After
image

### DIFF
--- a/spec/features/state_file/income_review_spec.rb
+++ b/spec/features/state_file/income_review_spec.rb
@@ -1,4 +1,6 @@
 require "rails_helper"
+require 'axe-capybara'
+require 'axe-rspec'
 
 RSpec.feature "Income Review", active_job: true do
   include StateFileIntakeHelper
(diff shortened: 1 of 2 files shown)


spec/features/state_file/income_review_spec.rb
```diff
@@ -54,6 +56,7 @@ def advance_to_income_edit
     it "displays w2 info on edit screen" do
       advance_to_income_edit
 
+      expect(page).to be_axe_clean
```
mluedke2, line 59: this passes even without the above change, but I wanted to add anyway while looking at this.

PR #5443, opened 2025-01-24
Do not attempt to fill out account holder names if taxes are owed (ac…

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1695

## What was done?
- For the 502xml, `NameOnBankAccount` is only filled out if: `if @intake.payment_or_deposit_type.to_sym == :direct_deposit && @intake.refund_or_owe_taxes_type == :refund`
- this logic was not duplicated for the filling out of the PDF:
   -   See code:
   ```
    def full_names_of_bank_account_holders
         intake = @submission.data_source
         return nil unless intake.payment_or_deposit_type.to_sym == :direct_deposit
    ```

## How to test?
  - added a test to catch this bug
  - manually loaded a `owed` MD intake and loaded up the pdf 
  - Use Nellie, choose direct_deposit at the `taxes-owed` controller, and hit download state return
  - confirmed in demo this was an error; confirmed on heroku that this no longer occurs
  
Screenshot 2025-01-24 at 12 40 43 PM

Screenshot 2025-01-24 at 12 40 38 PM

### DIFF
--- a/app/lib/pdf_filler/md502_pdf.rb
+++ b/app/lib/pdf_filler/md502_pdf.rb
@@ -178,7 +178,7 @@ module PdfFiller
 
     def full_names_of_bank_account_holders
       intake = @submission.data_source
-      return nil unless intake.payment_or_deposit_type.to_sym == :direct_deposit
+      return nil unless intake.payment_or_deposit_type.to_sym == :direct_deposit && intake.refund_or_owe_taxes_type == :refund
 
       if intake.has_joint_account_holder_yes?
         account_holder_full_name + " and " + account_holder_full_name(for_joint: true)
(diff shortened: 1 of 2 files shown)


app/lib/pdf_filler/md502_pdf.rb
```diff
@@ -191,6 +191,8 @@ def account_holder_full_name(for_joint: false)
       attributes = %w[FirstName MiddleInitial LastName NameSuffix]
       account_holder_xmls = @xml_document.css('Form502 NameOnBankAccount')
       account_holder_xml = for_joint ? account_holder_xmls[1] : account_holder_xmls[0]
+      return if account_holder_xml.nil?
```
arinchoi03, line 194: guarding in case the `NameOnBankAccount` is for some reason `nil` when we actually do expect it so it doesn't crash

mrotondo (approved):
Nice catch!

PR #5455, opened 2025-01-27
Truncate employer name after 35 characters

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/275

## What was done?
Truncate the employer name to avoid bundle failures

## How to test?
- Use the Zeus two w2s profile
- Edit the UI box 14 data to be under the threshold but add up to more than max (try 90 in both boxes)

### DIFF
--- a/app/helpers/state_file/nj_2450_helper.rb
+++ b/app/helpers/state_file/nj_2450_helper.rb
@@ -9,5 +9,22 @@ module StateFile
       ssn = primary_or_spouse == :primary ? intake.primary.ssn : intake.spouse.ssn
       intake.state_file_w2s.all&.select { |w2| w2.employee_ssn == ssn }
     end
+
+    def get_employer_name(w2, truncate: false)
+      truncate ? w2.employer_name[0...35] : w2.employer_name
+    end
+
+    def get_wages(w2)
+      w2.wages&.round
+    end
+
+    def get_column_a(w2)
+      column_a = w2.box14_ui_wf_swf&.positive? ? w2.box14_ui_wf_swf : w2.box14_ui_hc_wd
+      column_a&.round || 0
+    end
+
+    def get_column_c(w2)
+      w2.box14_fli&.round || 0
+    end
   end
 end
--- a/app/helpers/state_file/nj_2450_helper.rb
+++ b/app/helpers/state_file/nj_2450_helper.rb
@@ -11,7 +11,7 @@ module StateFile
     end
 
     def get_employer_name(w2, truncate: false)
-      truncate ? w2.employer_name[0...35] : w2.employer_name
+      truncate ? sanitize_for_xml(w2.employer_name, 35) : w2.employer_name
     end
 
     def get_wages(w2)
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
@@ -21,13 +21,13 @@ module SubmissionBuilder
                   xml.Body do
                     column_a = w2.box14_ui_wf_swf&.positive? ? w2.box14_ui_wf_swf : w2.box14_ui_hc_wd
 
-                    xml.EmployerName w2.employer_name[0...35]
+                    xml.EmployerName get_employer_name(w2, truncate: true)
                     xml.FedEmployerId w2.employer_ein
-                    xml.Wages w2.wages&.round
+                    xml.Wages get_wages(w2)
                     xml.Deductions do
-                      xml.ColumnA column_a&.round || 0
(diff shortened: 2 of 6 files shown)


app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
```diff
@@ -21,7 +21,7 @@ def document
                   xml.Body do
                     column_a = w2.box14_ui_wf_swf&.positive? ? w2.box14_ui_wf_swf : w2.box14_ui_hc_wd
 
-                    xml.EmployerName w2.employer_name
+                    xml.EmployerName w2.employer_name[0...35]
```
mpidcock, line 24: I recommend using `sanitize_for_xml(w2.employer_name, 35)` because that method will
* ensure no excess spaces (only single spaces allowed in most (all?) xml strings)
* trim a trailing space if the truncation lands after a space (trailing whitespace can also cause a validation error)

amanifarooque:
@mmazanec22 seeing XML errors now generating for the w2 > businessline1 field

the IRS _is_ doing validations on employer name for w2s, that limit is 70 characters, so we should never receive an employer name longer than that limit

```
182:0: ERROR: Element '{http://www.irs.gov/efile}BusinessNameLine1Txt': [facet 'pattern'] The value 'Wharton State Forest, the best place to mountain bike in New Jersey' is not accepted by the pattern '(([A-Za-z0-9#\-\(\)]|&|') ?)*([A-Za-z0-9#\-\(\)]|&|')'.
```

app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
mmazanec22: Fixed, thank you!

mpidcock (approved):

app/helpers/state_file/nj_2450_helper.rb
```diff
@@ -9,5 +9,22 @@ def get_persons_w2s(intake, primary_or_spouse)
...
+    def get_employer_name(w2, truncate: false)
+      truncate ? sanitize_for_xml(w2.employer_name, 35) : w2.employer_name
+    end
+
+    def get_wages(w2)
+      w2.wages&.round
+    end
+
+    def get_column_a(w2)
+      column_a = w2.box14_ui_wf_swf&.positive? ? w2.box14_ui_wf_swf : w2.box14_ui_hc_wd
```
mluedke2, line 22: [dust] works either way, but this can be `column_a = w2.get_box14_ui_overwrite` since that is used elsewhere and is 

```
def get_box14_ui_overwrite
  box14_ui_wf_swf || box14_ui_hc_wd
end
```

app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
```diff
@@ -21,13 +21,13 @@ def document
                   xml.Body do
                     column_a = w2.box14_ui_wf_swf&.positive? ? w2.box14_ui_wf_swf : w2.box14_ui_hc_wd
```
mluedke2, line 22: [pebble] i think this would want to use `get_column_a` to follow the same pattern as the others, right? (and I am on a semi-quest to DRY box 14 as much as possible)

spec/lib/submission_builder/ty2024/states/nj/documents/nj2450_spec.rb
```diff
@@ -24,7 +25,10 @@
         body_elements.each do |body|
           w2 = [w2_1, w2_2].find { |test_w2| test_w2.employer_ein == body.at('FedEmployerId').text}
 
-          expect(body.at('EmployerName').text).to eq(w2.employer_name)
+          expect(body.at('EmployerName').text).to eq(w2.employer_name[0...35])
+          if w2.employer_name == long_employer_name
+            expect(body.at('EmployerName').text.length).to eq(35)
```
mluedke2, line 30: [pebble] putting expectations inside an `if` feels off to me, since there's an outside chance the test is only passing because the expectation isn't being tested. I think the expectation on line 28 is probably enough on its own, but if you feel there needs to be more then you could test explicitly the two expected names outside of the `.each`?

mluedke2 (approved):
two pebbles and a dust but otherwise looks great to me!

mmazanec22:
@mluedke2 I'm going to address your comments in a fast follow since this is an urgent PR to fix bundle failures

app/lib/submission_builder/ty2024/states/nj/documents/nj2450.rb
mmazanec22: Unused variable, whoops!

PR #5456, opened 2025-01-27
remove state file w2 validations

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1710

## What was done?
- We were failing to fully import DF data for people who had W2 information that failed our ActiveRecord validations. Because we don't currently have a way for them to fix their W2s (and even once we do, we'll want those validations to apply to that UI, not at import), this PR removes the validations from the StateFileW2 model. They should be reintroduced in the future on a W2 form, and ideally only for fields that we allow users to edit.
## How to test?
- Go through the DF import flow and use the XML editor to generate invalid W2 XML. For example, make state wages greater than box 1 (was causing validation error in states other than NJ) or make Locality Nm blank.


mpidcock (approved):
LGTM

PR #5457, opened 2025-01-27
NJ 271 - fix when review screen displays health insurance section

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/271

## What was done?
- Update NJ Review screen to have the same logic as the controller, only displays if there are dependents

https://github.com/codeforamerica/vita-min/blob/da604920eee7df593281810b9b36bcb3be9a9c7f/app/controllers/state_file/questions/nj_dependents_health_insurance_controller.rb#L7

## How to test?
- Use Springsteen MFJ, navigate to review screen, ensure that "Dependents DO NOT have Health Insurance" section does not appear

- Before
image

- After
image

### DIFF
--- a/app/views/state_file/questions/nj_review/edit.html.erb
+++ b/app/views/state_file/questions/nj_review/edit.html.erb
@@ -23,7 +23,7 @@
     </section>
   <% end %>
 
-  <% if current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no? %>
+  <% if current_intake.dependents.any? && (current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no?) %>
     <section id="missing-health-insurance" class="white-group">
       <div class="spacing-below-5">
         <h2 class="text--body text--bold spacing-below-5"><%=t(".missing_health_insurance") %></h2>
--- a/app/views/state_file/questions/nj_review/edit.html.erb
+++ b/app/views/state_file/questions/nj_review/edit.html.erb
@@ -23,7 +23,7 @@
     </section>
   <% end %>
 
-  <% if current_intake.dependents.any? && (current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no?) %>
+  <% if current_intake.dependents.any? && current_intake.has_health_insurance_requirement_exception? %>
     <section id="missing-health-insurance" class="white-group">
       <div class="spacing-below-5">
         <h2 class="text--body text--bold spacing-below-5"><%=t(".missing_health_insurance") %></h2>
--- a/spec/controllers/state_file/questions/nj_dependents_health_insurance_controller_spec.rb
+++ b/spec/controllers/state_file/questions/nj_dependents_health_insurance_controller_spec.rb
@@ -29,14 +29,6 @@ RSpec.describe StateFile::Questions::NjDependentsHealthInsuranceController do
           expect(described_class.show?(intake)).to eq true
         end
       end
-     
-      context "and did not have a health insurance requirement exception, but all members did not have health insurance" do  
-        it "shows" do
-          allow_any_instance_of(StateFileNjIntake).to receive(:has_health_insurance_requirement_exception?).and_return(false)
-          allow_any_instance_of(StateFileNjIntake).to receive(:eligibility_all_members_health_insurance_no?).and_return(true)
-          expect(described_class.show?(intake)).to eq true
-        end
-      end
     end
   end
 
--- a/spec/controllers/state_file/questions/nj_review_controller_spec.rb
+++ b/spec/controllers/state_file/questions/nj_review_controller_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe StateFile::Questions::NjReviewController do
+  let(:intake) { create :state_file_nj_intake, :df_data_minimal }
+  before do
+    sign_in intake
+  end
+
+  describe "#edit" do
+    render_views
+    it 'succeeds' do
+      get :edit
+      expect(response).to be_successful
+    end
+
+    context 'when no dependents and gross income at threshold' do
+      it 'does not show the dependents without health insurance block' do
+        allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_29).and_return 20_000
+        get :edit       
+        expect(response.body).not_to have_text "Dependents who DO NOT have health insurance"
+      end
+    end
+  end
+end
--- a/spec/controllers/state_file/questions/nj_review_controller_spec.rb
+++ b/spec/controllers/state_file/questions/nj_review_controller_spec.rb
@@ -1,23 +1,38 @@
 require "rails_helper"
 
 RSpec.describe StateFile::Questions::NjReviewController do
-  let(:intake) { create :state_file_nj_intake, :df_data_minimal }
   before do
     sign_in intake
   end
 
   describe "#edit" do
     render_views
-    it 'succeeds' do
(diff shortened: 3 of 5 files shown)


app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -23,7 +23,7 @@
     </section>
   <% end %>
 
-  <% if current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no? %>
+  <% if current_intake.dependents.any? && (current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no?) %>
```
tristanvanech, line 26: is the second conditional necessary? If they answered no to the health insurance eligibility question, they either will be offboarded (and so will never get to the review screen) or they meet an exception

spec/controllers/state_file/questions/nj_review_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  describe "#edit" do
+    render_views
+    it 'succeeds' do
+      get :edit
+      expect(response).to be_successful
+    end
+
+    context 'when no dependents and gross income at threshold' do
+      it 'does not show the dependents without health insurance block' do
+        allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_29).and_return 20_000
```
tristanvanech, line 18: I'm not sure if this is a test or app functionality but just highlighting that the income threshold is just one of the exceptions to health insurance eligibility (the other being claimed as dependents on another return) and it's 20k for some filing statuses, 10k for others

mluedke2: That is a good point - instead of getting into this or that reason for eligibility, which is all tested elsewhere, the mock here can have something like:

```
allow_any_instance_of(StateFileNjIntake).to receive(:has_health_insurance_requirement_exception?).and_return(true)
allow_any_instance_of(StateFileNjIntake).to receive(:eligibility_all_members_health_insurance_no?).and_return(true)
```

spec/controllers/state_file/questions/nj_review_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+        get :edit       
+        expect(response.body).not_to have_text "Dependents who DO NOT have health insurance"
```
mluedke2, line 20: [pebble] - I think if we have this test, we should pair it with a test case for the opposite when we *do* expect the text

spec/controllers/state_file/questions/nj_review_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe StateFile::Questions::NjReviewController do
+  let(:intake) { create :state_file_nj_intake, :df_data_minimal }
```
mluedke2, line 4: [dust] - should we use one of the more realistic profiles here? and then use mocks below as needed?

app/views/state_file/questions/nj_review/edit.html.erb
mluedke2: That's a great catch -- it was in the previous code so not introduced here, but I suspect that if we took it out then tests would probably still pass and would be cleaner.

spec/controllers/state_file/questions/nj_review_controller_spec.rb
jachan: Yeah, it's a good point - happy to make a new import for springsteen that represents no dependents.

app/views/state_file/questions/nj_review/edit.html.erb
jachan: After some digging, I think you're right! This second condition is redundant because anyone at this point would have been screened out if they answered yes and didn't have an exception. Truth table below to validate that the logic behaves identically:

`current_intake.has_health_insurance_requirement_exception? || current_intake.eligibility_all_members_health_insurance_no?`
Exception True, Insurance_no? True = conditional True
Exception False, Insurance_no? True - screened out
Exception True, Insurance_no? False = conditional True
Exception False, Insurance_no? False = conditional False

`current_intake.has_health_insurance_requirement_exception?`
Exception True, Insurance_no? True = conditional True
Exception False, Insurance_no? True - screened out
Exception True, Insurance_no? False = conditional True
Exception False, Insurance_no? False = conditional False

spec/controllers/state_file/questions/nj_dependents_health_insurance_controller_spec.rb
```diff
@@ -29,14 +29,6 @@
           expect(described_class.show?(intake)).to eq true
         end
       end
-     
-      context "and did not have a health insurance requirement exception, but all members did not have health insurance" do
```
jachan, line 33: Test case would never happen as someone with no exceptions and who answered no to the eligibility question would be screened out.

https://github.com/codeforamerica/vita-min/blob/d02c9011da076940673f58964643fa32753e3582/app/models/state_file_nj_intake.rb#L188

spec/factories/state_file_nj_intakes.rb
```diff
@@ -178,6 +178,13 @@
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_minimal') }
     end
 
+    trait :df_data_no_deps do
+      raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_springsteen_mfj') }
+      raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_springsteen_mfj') }
+      primary_birth_date { Date.new(1990, 1, 1) }
```
jachan, line 184: Have to add birth dates or EITC check fails when trying to check age of filer/spouse.

spec/factories/state_file_nj_intakes.rb
```diff
@@ -181,8 +173,6 @@
     trait :df_data_no_deps do
       raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_springsteen_mfj') }
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_springsteen_mfj') }
-      primary_birth_date { Date.new(1990, 1, 1) }
-      spouse_birth_date { Date.new(1990, 1, 1) }
```
jachan: Spouse birth date was not set because filing status was incorrectly set to `single`.

spec/factories/state_file_nj_intakes.rb
```diff
@@ -128,14 +128,6 @@
...
     after(:build) do |intake, evaluator|
       intake.municipality_code = "0101"
-      numeric_status = {
-        single: 1,
-        married_filing_jointly: 2,
-        married_filing_separately: 3,
-        head_of_household: 4,
-        qualifying_widow: 5,
-        }[evaluator.filing_status.to_sym] || evaluator.filing_status
-      intake.direct_file_data.filing_status = numeric_status
```
jachan, line 138: This code would always override filing status to be `single`

jnf (approved):
lgtm

spec/factories/state_file_nj_intakes.rb
jachan: Moving this issue to a separate ticket: https://github.com/newjersey/affordability-pm/issues/279. Going to revert the logic for now, and use a persona with `filing_status` single (nj_lucky_single) and no dependents for the test.

mluedke2: thanks for making that follow-up ticket!

mluedke2 (approved):
looks good!

PR #5464, opened 2025-01-28
make one more local tax validation only apply in states where we show local tax boxes for w2s

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1710

## What was done?
- I missed one W2 validation of local tax info that should only be applied in states where we show local tax boxes for w2s, and therefore where users can actually address the error (only MD for now)
## How to test?
- Go through the flow for AZ (or any state other than MD), choose a persona with a w2, edit the XML and set LocalityNm to 0, then verify that you can proceed through the Income Review page

### DIFF
--- a/spec/models/state_file_w2_spec.rb
+++ b/spec/models/state_file_w2_spec.rb
@@ -71,10 +71,10 @@ describe StateFileW2 do
     end
 
     context "states where we don't show local boxes" do
-      let(:intake) { create :state_file_nc_intake }
+      let(:intake) { create :state_file_az_intake }
 
       it "doesn't validate local tax fields" do
-        w2.locality_nm = "asdf"
+        w2.locality_nm = "0"
         w2.local_wages_and_tips_amount = -1
         w2.local_income_tax_amount = -1
         expect(w2).to be_valid(:state_file_edit)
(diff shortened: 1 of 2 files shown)


spec/models/state_file_w2_spec.rb
```diff
@@ -71,10 +71,10 @@
     end
 
     context "states where we don't show local boxes" do
-      let(:intake) { create :state_file_nc_intake }
+      let(:intake) { create :state_file_az_intake }
 
       it "doesn't validate local tax fields" do
-        w2.locality_nm = "asdf"
+        w2.locality_nm = "0"
```
arinchoi03, line 77: why did this change from nc to az? both of them do not have local boxes shown, right?

spec/models/state_file_w2_spec.rb
```diff
@@ -71,10 +71,10 @@
     end
 
     context "states where we don't show local boxes" do
-      let(:intake) { create :state_file_nc_intake }
+      let(:intake) { create :state_file_az_intake }
```
arinchoi03, line 74: why did this change from nc to az? both of them do not have local boxes shown, right?

mrotondo: Yes, this is a no-op change and I just wanted to double-check that there wasn't something weird about AZ, since that was the first acceptance failure I heard about on the previous PR.

PR #5473, opened 2025-01-29
[FYST-1727] special program az

jnf:
## Link to pivotal/JIRA issue
[FYST-1727](https://codeforamerica.atlassian.net/browse/FYST-1727)

## What was done?
- Adds a new element to the return header in AZ xml submissions.
- `SpecialProgram` is an optional node that has remain unused; AZ folks asked if we could start including it to help them in their record keeping/stat building.
## How to test?
- Make an AZ return; it should have `<SpecialProgram>Direct File</SpecialProgram>` right after `<ReturnType>` in the `<ReturnHeaderState>` block.
- No other state return should include `<SpecialProgram>`

[FYST-1727]: https://codeforamerica.atlassian.net/browse/FYST-1727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
```diff
@@ -185,6 +187,10 @@ def schema_file
             SchemaFileLoader.load_file("us_states", "unpacked", "AZIndividual2024v2.0", "AZIndividual", "IndividualReturnAZ140.xsd")
           end
 
+          def special_program
+            "<SpecialProgram>Direct File</SpecialProgram>"
```
jnf, line 191: i dunno if there's a fancy way to do this, but this works.

mpidcock: I think it would be something kinda ridiculous like 
```
Nokogiri::XML::Builder.new(encoding: 'UTF-8').send("SpecialProgram", "Direct File")
```

what you have is probably more clear

mpidcock (approved):

PR #5474, opened 2025-01-29
Nj a11y 19 helper text

mmazanec22:
## Link to pivotal/JIRA issue
Project issue: https://github.com/newjersey/affordability-pm/issues/246
A11y issue: https://github.com/newjersey/accessibility-pm/issues/19

## What was done?
- Update honeycrisp
- Update components affected by the honeycrisp update
- Update help text and errors for non-honeycrisp components: money field and checkbox set, the latter of which is only used by NJ
- Update help text color
- Move box 14 help text to associate it with the input - **I did this for all states, so it does affect Maryland STPICKUP**

## How to test?
- This change can be observed on the contact info pages and in the box 14 editing section, as well as on the dependents attending college page

Before
image
image
After
image
image

This is what the vita min select looks like with help text. We are not using this component with help text anywhere in the application.
image

### DIFF
--- a/app/helpers/vita_min_form_builder.rb
+++ b/app/helpers/vita_min_form_builder.rb
@@ -291,20 +291,11 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
       help_text: nil
     )
 
-    describedby_ids = []
-    
-    if help_text
-      help_id = help_text_id(method)
-      describedby_ids << help_id
-    end
-    
-    if object.errors[method].any?
-      describedby_ids << error_label(method)
-    end
+    describedby = get_describedby(method, help_text: help_text)
 
     text_field_options = standard_options.merge(error_attributes(method: method)).merge(
       class: (classes + ["text-input money-input"]).join(" "),
-      'aria-describedby': describedby_ids.join(' ')
+      'aria-describedby': describedby
     ).merge(placeholder: '0.00').merge(options)
     
     text_field_options[:id] ||= sanitized_id(method)
@@ -470,17 +461,12 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
 
     fieldset_classes = ["input-group", "form-group#{error_state(object, method)}"]
 
-    describedby_ids = []
+    describedby = get_describedby(method, help_text: help_text)
+
     if help_text.present?
       help_id = help_text_id(method)
       help_html = help_text_html(help_text, help_id)
-      describedby_ids << help_id
     end
-
-    if object.errors[method].any?
-      describedby_ids << error_label(method)
-    end
-    describedby = describedby_ids.join(' ')
 
     <<~HTML.html_safe
       <fieldset class="#{fieldset_classes.join(' ')}" aria-describedby="#{describedby}">
@@ -517,6 +503,21 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
       style: "display:none",
       'data-permitted': permitted_values.to_json
     )
+  end
+
+  def get_describedby(method, help_text: nil)
+    describedby_ids = []
+
+    if help_text
+      help_id = help_text_id(method)
+      describedby_ids << help_id
+    end
+
+    if object.errors[method].any?
+      describedby_ids << error_label(method)
+    end
+
+    describedby_ids.join(' ')
   end
 
   ##### Methods that copy from or override honeycrisp
--- a/app/helpers/vita_min_form_builder.rb
+++ b/app/helpers/vita_min_form_builder.rb
@@ -472,6 +472,7 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
     HTML
   end
 
+  alias v1_cfa_input_field cfa_input_field
   def cfa_input_field(
     method,
     label_text,
@@ -479,24 +480,41 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
     help_text: nil,
     options: {},
     autofocus: nil,
-
     classes: [],
-    prefix: nil,
-    postfix: nil,
     optional: false,
-    notice: nil
+    prefix: nil
   )
+    if prefix
+      return v1_cfa_input_field(
+        method,
+        label_text,
+        type: type,
+        help_text: help_text,
+        options: options,
+        autofocus: autofocus,
+        classes: classes,
+        optional: optional,
+        prefix: prefix,
+        postfix: nil, # not used in this application
+        notice: nil, # not used in this application
+      )
+    end
+
     text_field_options = options.merge(
       type: type,
       autofocus: autofocus,
-      # required: options.required || !optional  todo fix
+      required: options[:required]
     )
+
+    if !options[:required] && optional
+      text_field_options = text_field_options.merge(required: false)
+    end
 
     @v2_builder.cfa_text_field(
       method,
       label_text,
       help_text: help_text,
-      wrapper_options: {},
+      wrapper_options: { class: classes.join(' ') },
       label_options: {},
       **text_field_options
     )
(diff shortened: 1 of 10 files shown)


spec/support/helpers/state_file_intake_helper.rb
```diff
@@ -68,18 +68,28 @@ def step_through_eligibility_screener(us_state:)
     end
   end
 
-  def step_through_initial_authentication(contact_preference: :text_message)
+  def expect_programmatically_associated_help_text
```
jachan, line 71: This is very cool!

app/helpers/vita_min_form_builder.rb
```diff
@@ -485,4 +517,120 @@ def warning_for_select(element_id, permitted_values, msg)
...
+      label_text,
+      help_text: help_text,
+      wrapper_options: { class: classes.join(' ') },
+      label_options: {},
+      **text_field_options
+    )
+  end
+
+  # Added help text to label and field method instead to remove it from label
+  def label_contents(label_text, help_text, optional: false, include_help_text: true)
```
mmazanec22, line 572: I had to include help text by default to account for a select dropdown in `app/views/hub/clients/edit_take_action.html.erb`. NJ is not using help text with either the `cfa_select` or `vita_min_state_file_select` components currently, so I did not alter these components. However, these components also need the same label/error/help text changes.

app/helpers/vita_min_form_builder.rb
```diff
@@ -285,13 +291,27 @@ def vita_min_money_field(
       classes: [],
       help_text: nil
     )
-    text_field_options = standard_options.merge(
-      class: (classes + ["text-input money-input"]).join(" "),
-      ).merge(error_attributes(method: method)).merge(placeholder: '0.00').merge(options)
 
+    describedby_ids = []
```
jachan, line 295: [DUST] Is there a way to pull out this logic into a shared function that can be shared between `vita_min_money_field` and `vita_min_checkbox_set`?

app/helpers/vita_min_form_builder.rb
```diff
@@ -485,4 +517,120 @@ def warning_for_select(element_id, permitted_values, msg)
       'data-permitted': permitted_values.to_json
     )
   end
+
+  # Methods that override honeycriscp v1
+
+  alias v1_cfa_input_field cfa_input_field
+  def cfa_input_field(
```
jachan, line 524: [Pebble] This override is meant to convert v1 honeycrisp to v2 honeycrisp, since V1 has this accessibility issue while V2 fixes it.

Question for CFA - how much effort is it to upgrade from honeycrisp_v1 to honeycrisp_v2? Is there a reason to stay with v1 right now? That seems like a cleaner solution if possible.

jachan (commented):
Without this change, users with screen readers are not able to know which field an error message or help text corresponds to. This is a WCAG Level A violation, which means the site is unusable for screen reader users.

Honeycrisp_v2 resolves the violation, but right now we're still using honeycrisp_v1. The changes in this PR provide a crosswalk that allows the use honeycrisp_v2 components without a full repo-wide upgrade. If there are concerns from CFA about this "crosswalk" approach (which I understand might feel hacky), I think the alternative would be prioritizing the repo-wide upgrade.

app/helpers/vita_min_form_builder.rb
```diff
@@ -485,4 +517,120 @@ def warning_for_select(element_id, permitted_values, msg)
...
+    if help_text && include_help_text
+      label_text << <<~HTML
+        <p class="text--help">#{help_text}</p>
+      HTML
+    end
+
+    label_text.html_safe
+  end
+
+  def label_and_field(
```
mmazanec22, line 586: Overrides this honeycrisp method: https://github.com/codeforamerica/honeycrisp-gem/blob/8f20565dd9354362491000ede2bf7c242d7d39a4/app/helpers/cfa/styleguide/cfa_form_builder.rb#L455

mmazanec22:
A little elaboration on how users are affected by the current state of form fields, and why I made the changes I did, since not all the recommendations in the ticket are covered in the description.

- Change: Error text programmatically associated with inputs and fieldsets
- Why: From Deque: "People who are blind cannot use the visual layout of a form to determine which form element contains an error. Because of this, error messages must specifically identify the field other by containing the name of the form element, or by being programmatically associated with the form element. When error messages and form elements are textually or programmatically associated, a screen reader user can be certain to which form element an error message applies."
  - More about the WCAG standard: [Information, structure, and relationships ](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html)conveyed through presentation can be programmatically determined or are available in text.
  - If all of our error messages used the names of the fields, then we could skip this change. But since they don’t, programmatic association seemed like a lighter lift than changing all the “can’t be blank” messages to specify the field


- Change: P tags no longer used in labels
- Why: [Phrasing content elements are the only valid child elements of labels.](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) Divs and paragraphs are flow content. We are not sure if this affects all assistive tech users the same way, but our accessibility consultant has indicated that it is an issue for the Voiceover screen reader.
  - Without this fix, some screen reader users currently cannot perceive any form labels in the entire application, making the whole application unusable
  - Even if this does not affect all screen readers, fixing this is consistent with [the WCAG requirement that applications be “robust,”](https://www.w3.org/TR/WCAG22/#robust) or compatible with many types of assistive technology.


- Change: Helper text moved outside of labels (and programmatically associated with fields)
- Why: This change makes it so the accessible name of the form control matches what users are seeing. Including the help text, but [distinguishing it by color](https://www.w3.org/TR/WCAG22/#sensory-characteristics) when it’s not actually separate, is potentially confusing for people using voice inputs.
  - Including help text in the label might also be less usable for people using screen readers, who will have to listen to additional text– especially when the help text is long.
  - Including help text in the label means that we can’t include links in help text


Additional recommendations
- Apply these changes to select elements (I didn’t because we’re not using help text with select elements)
- Apply aria-invalid to fields with errors (https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21)

arinchoi03:
this is not viewable to NJ folk but for cfa team reference, here is when GCF implemented accessible error to inputs 
https://github.com/codeforamerica/gcf-backend/pull/4486/files

app/helpers/vita_min_form_builder.rb
```diff
@@ -23,9 +24,10 @@ def vita_min_field_in_label(
 
     formatted_label = label(
         method,
-        label_contents(label_text, help_text, optional: optional) + field_html,
+        label_contents(label_text, has_help_text: help_text, optional: optional) + field_html,
```
jenny-heath, line 27: is there a reason we don't do `!!` here like in `vita_min_state_file_select` below?

app/helpers/vita_min_form_builder.rb
```diff
@@ -63,16 +65,19 @@ def vita_min_state_file_select(
...
         label_text,
-        options[:help_text],
+        has_help_text: !!options[:help_text],
         optional: options[:optional],
         ),
       class: label_class,
       )
     html_options_with_errors = html_options.merge(error_attributes(method: method))
 
+    help_html = help_text_html(options[:help_text], :method)
```
jenny-heath, line 75: shouldn't we be using the method variable?
```suggestion
    help_html = help_text_html(options[:help_text], method)
```

app/helpers/vita_min_form_builder.rb
```diff
@@ -63,16 +65,19 @@ def vita_min_state_file_select(
...
+
     html_output = <<~HTML
       <div class="form-group#{error_state(object, method)}">
         #{formatted_label}
+        #{help_html}
```
jenny-heath, line 80: can we have a screenshot in the PR description of what the select looks like with help text?

mmazanec22: Sure, though FWIW it looks like this select + help text is not in use anywhere in the app

jenny-heath: ohh didn't realize we don't have one. good to know!

PR #5475, opened 2025-01-29
FYST-1721 AZ include qual child dependent only if filer is hoh

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1721

## What was done?
- Added a condition to make sure filer is hoh before adding qualifying dependent info

## How to test?
- Updated units test

### DIFF
--- a/app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
+++ b/app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
@@ -44,7 +44,7 @@ module SubmissionBuilder
             xml_doc = build_xml_doc("Form140") do |xml|
               xml.LNPriorYrs sanitize_for_xml(@submission.data_source.prior_last_names)
               xml.FilingStatus filing_status
-              if hoh_qualifying_person.present?
+              if hoh_qualifying_person.present? && @submission.data_source.filing_status_hoh?
                 xml.QualChildDependentName do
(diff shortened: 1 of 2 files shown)


app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
```diff
@@ -44,7 +44,7 @@ def documents_wrapper
             xml_doc = build_xml_doc("Form140") do |xml|
               xml.LNPriorYrs sanitize_for_xml(@submission.data_source.prior_last_names)
               xml.FilingStatus filing_status
-              if hoh_qualifying_person.present?
+              if hoh_qualifying_person.present? && @submission.data_source.filing_status_hoh?
```
arinchoi03, line 47: can `hoh_qualifying_person` exist if filing status is _not_ hoh?
i.e. can we fold in this check into the `hoh_qualifying_person` logic?

arinchoi03 (commented):
one question/suggestion

app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
tahsinaislam: yeah we had a hoh qualifiying dependent on a single filer

arinchoi03 (approved):
ty for clarifying! I should've read the story a little more closely

PR #5476, opened 2025-01-29
FYST-1728 Fix MD client address updates

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1728

## What was done?
- Make sure MD is hardcoded when the client changes their address in the permanent address controller. This is because in the form MD is hardcoded because permanent address must be within MD 

## How to test?
- Updated permanent address controller because there shouldn't be a case where a CO/out of state address is shown in the xml because in the flow it stops you to change the address to a MD one 

- Before
- After

### DIFF
--- a/app/lib/submission_builder/ty2024/states/md/documents/md502.rb
+++ b/app/lib/submission_builder/ty2024/states/md/documents/md502.rb
@@ -54,7 +54,7 @@ class SubmissionBuilder::Ty2024::States::Md::Documents::Md502 < SubmissionBuilde
           xml.AddressLine1Txt sanitize_for_xml(@intake.permanent_street, 30)
           xml.AddressLine2Txt sanitize_for_xml(@intake.permanent_apartment, 30) if @intake.permanent_apartment.present?
           xml.CityNm sanitize_for_xml(@intake.permanent_city, 20)
-          xml.StateAbbreviationCd @intake.direct_file_data.mailing_state.upcase
+          xml.StateAbbreviationCd "MD"
           xml.ZIPCd sanitize_zipcode(@intake.permanent_zip)
         end
       end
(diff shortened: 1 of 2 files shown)


arinchoi03 (commented):
> (Remember: this change should not impact mailing address from DF, which can be out-of state)

Found this in the note, seeing the tests be deleted, it looks like this change will make DF mailing address that is out of state hard-coded to "MD"

tahsinaislam:
> > (Remember: this change should not impact mailing address from DF, which can be out-of state)
> 
> Found this in the note, seeing the tests be deleted, it looks like this change will make DF mailing address that is out of state hard-coded to "MD". What I _think_ should happen is that if _not_ `:permanent_address_outside_md`, we could hardcode? Happy to tuple!

Mailing address is that in the return header. We are hardcoding the Maryland permanent address in md502 which must always be in MD. This is the address we correct in the flow if it is out of state and the state is hardcoded to MD in the flow.

app/lib/submission_builder/ty2024/states/md/documents/md502.rb
```diff
@@ -54,7 +54,7 @@ def document
           xml.AddressLine1Txt sanitize_for_xml(@intake.permanent_street, 30)
           xml.AddressLine2Txt sanitize_for_xml(@intake.permanent_apartment, 30) if @intake.permanent_apartment.present?
           xml.CityNm sanitize_for_xml(@intake.permanent_city, 20)
-          xml.StateAbbreviationCd @intake.direct_file_data.mailing_state.upcase
+          xml.StateAbbreviationCd "MD"
```
arinchoi03, line 57: confirmed that although the DF _can_ send us the out of state `mailing_state` -- if the user confirms `:permanent_address_outside_md` it will lead the user to an ineligible screen. Which means that by the time we fill out the MD502, the user should've provided an address within MD (state hard coded on permanent address controller).

arinchoi03 (approved):
🙏

PR #5483, opened 2025-01-30
Fyst 1720 md disqualify dependent taxpayers from claiming poverty level credit

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1720

## What was done?
- The MD 502 calculator is supposed to make certain decisions based on whether the filer will use the filing status "dependent" on their Maryland return. However, this filing status only exists for Maryland, and will never be present on the federal return - we decide whether to use it in MD based on whether the filer's direct file data indicates that they are claimed as a dependent.
- This means that the previous implementation of `Md502Calculator#filing_status_dependent?` would never return true for a real user, since it checked the intake's filing status (which is their federal filing status) for a value it could never have.
- This PR changes that method to instead check the `direct_file_data.claimed_as_dependent?` method, which is the same way we decide to set their MD filing status to "dependent".
- Several tests started failing when we made this change, since they were incorrectly setting the federal filing status of test data to 6 (dependent) in their setup methods, which it can never be. These methods (MD502 line 22 and line 42) both relate to EIC and now both return if the filer is claimed as a dependent, which from talking with Courtney I believe is correct, but we should double check.
- I think that we need a subsequent change that will uniformize how we use filing status in the calculators. It sounds like we calculate a state filing status that could differ from the federal filing status in at least one other state. IMO we should have a `federal_filing_status` and `state_filing_status` method on intakes, the latter of which in MD would return `dependent` if the filer is claimed as a dependent (as we do in Md502.rb on lines 63-74), and switch our calculators over to use `#state_filing_status` everywhere (or almost everywhere, if there are calculations that explicitly use federal filing status)
  - Notably, there are still a number of tests for Md502Calculator that pass `:dependent` as a filing status to the StateFileMdIntake constructor, which I believe will never happen in real usage since that is not a valid federal filing status.
## How to test?
- We need to be careful with this one, since we failed to catch this bug in acceptance previously.
- We should reproduce the error on demo with a persona that is claimed as a dependent but has a valid federal filing status and is otherwise eligible for poverty level credit (and EIC, since i suspect we had the same kind of bug with that credit)
- We should then use the same persona and do the same checks on this PR's heroku instance and verify that we are not able to claim those credits

### DIFF
--- a/spec/lib/efile/md/md502_calculator_spec.rb
+++ b/spec/lib/efile/md/md502_calculator_spec.rb
@@ -940,8 +940,10 @@ describe Efile::Md::Md502Calculator do
       )
     }
     let(:federal_eic) { 1001 }
+    let(:claimed_as_dependent) { false }
 
     before do
+      allow_any_instance_of(DirectFileData).to receive(:claimed_as_dependent?).and_return claimed_as_dependent
       intake.direct_file_data.fed_eic = federal_eic
       instance.calculate
     end
@@ -970,9 +972,8 @@ describe Efile::Md::Md502Calculator do
 
     context "when filing as a dependent and no qualifying children" do
       let(:df_xml_key) { "md_zeus_two_w2s" }
-      before do
-        intake.direct_file_data.primary_claim_as_dependent = "X"
-      end
+      let(:claimed_as_dependent) { true }
+
       it 'EIC is nil' do
         expect(instance.lines[:MD502_LINE_22].value).to eq nil
       end
@@ -1023,20 +1024,21 @@ describe Efile::Md::Md502Calculator do
   end
 
   describe "#calculate_line_23" do
+    let (:deduction_method_standard) { true}
+    let (:claimed_as_dependent) { false }
+
     before do
       intake.direct_file_data.fed_wages_salaries_tips = line_1b
       allow_any_instance_of(described_class).to receive(:calculate_line_7).and_return(line_7)
-      allow_any_instance_of(described_class).to receive(:deduction_method_is_standard?).and_return(true)
+      allow_any_instance_of(described_class).to receive(:deduction_method_is_standard?).and_return(deduction_method_standard)
+      allow_any_instance_of(DirectFileData).to receive(:claimed_as_dependent?).and_return claimed_as_dependent
       instance.calculate
     end
 
     context "deduction method is nonstandard" do
-      let(:line_1b) { 20_000 }
-      let(:line_7) { 15_000 }
-
-      before do
-        allow_any_instance_of(described_class).to receive(:deduction_method_is_standard?).and_return(false)
-      end
+      let(:line_1b) { 5_000 }
+      let(:line_7) { 10_000 }
+      let(:deduction_method_standard) { false }
 
       it "returns 0" do
         expect(instance.lines[:MD502_LINE_23].value).to eq 0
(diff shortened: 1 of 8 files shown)


spec/lib/efile/md/md502_calculator_spec.rb
```diff
@@ -1018,30 +1024,31 @@
...
     end
 
     context "deduction method is nonstandard" do
-      let(:line_1b) { 20_000 }
-      let(:line_7) { 15_000 }
-
-      before do
-        allow_any_instance_of(described_class).to receive(:deduction_method_is_standard?).and_return(false)
-      end
+      let(:line_1b) { 5_000 }
```
mrotondo, line 1039: These needed to change because the previous values made the intake ineligible for PLC - the test still passed even when I changed the filing status to not-dependent

spec/lib/efile/md/md502_calculator_spec.rb
```diff
@@ -1018,30 +1024,31 @@
...
       it "returns 0" do
         expect(instance.lines[:MD502_LINE_23].value).to eq 0
       end
     end
 
     context "when filing as dependent" do
-      let(:filing_status) { "dependent" }
-      let(:line_1b) { 20_000 }
-      let(:line_7) { 15_000 }
+      let(:line_1b) { 5_000 }
```
mrotondo, line 1049: Same as above

arinchoi03:
> Notably, there are still a number of tests for Md502Calculator that pass :dependent as a filing status to the StateFileMdIntake constructor, which I believe will never happen in real usage since that is not a valid federal filing status.

should we fix this up as a fast follow up to this story?

spec/lib/efile/md/md502_calculator_spec.rb
```diff
@@ -1018,30 +1024,31 @@
   end
 
   describe "#calculate_line_23" do
+    let (:deduction_method_standard) { true}
```
arinchoi03, line 1027: space after `true`

spec/lib/efile/md/md502cr_calculator_spec.rb
```diff
@@ -227,7 +227,17 @@
...
     %w[single married_filing_separately dependent].each do |filing_status|
       context "when filing status is #{filing_status}" do
-        let(:filing_status) { filing_status }
+        if filing_status == "dependent"
+          let(:filing_status) { "single" }
+          let(:claimed_as_dependent) { true}
+        else
+          let(:filing_status) { filing_status }
+          let(:claimed_as_dependent) { false }
+        end
```
arinchoi03, line 236: i'm guessing we'll do something similar for the MD502 calculator tests when we have time to fix them up?

arinchoi03 (approved):

mrotondo:
> should we fix this up as a fast follow up to this story? Are we not testing these scenarios properly currently because `dependent` filing_status doesn't really mean anything -- for instance, we're never entering the `if` statement here in the test when we determine deduction method?:

Yes, the tests should be fixed but IMO that should go hand in hand with changing how we check filing status (e.g. the federal_filing_status vs state_filing_status suggestion in the pr description). And that work should be very hand in hand with program & PM to make sure it is acceptance-tested really precisely.

And as for the if block you pasted - we _will_ go into that one because that (`direct_file_data.claimed_as_dependent?`) is actually the correct check (rather than intake.filing_status == :dependent), but I'd rather that the method do something like `intake.state_filing_status == :dependent` so that we're not duplicating the logic of checking the direct file data field all over the place.

spec/lib/efile/md/md502cr_calculator_spec.rb
mrotondo: Yes, that's my thinking

PR #5484, opened 2025-01-30
Do not include spouse ssn if there is not one

mmazanec22:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/280

- Yes - don't merge until JIRA issue is accepted! - after CfA has handled the NRA code in the header

## What was done?
Do not try to include mfs spouse ssn if there isn't one

## How to test?
Use the married filing separately 15k wages profile

### DIFF
--- a/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
+++ b/app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
@@ -28,7 +28,7 @@ module SubmissionBuilder
                     case status
                     when :married_filing_separately
                       xml.MarriedCuPartFilingSeparate do
-                        xml.SpouseSSN intake.spouse.ssn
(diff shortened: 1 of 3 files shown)


aloverso (approved):

mluedke2 (approved):

amanifarooque:
@mmazanec22 I noticed we didn't make any changes to the json data file -- this can be cleanup later, but in this case we would not get Spouse SSN in either the XML or JSON.
https://github.com/codeforamerica/vita-min/blob/main/spec/fixtures/state_file/fed_return_jsons/nj/married_filing_separately_15k_wages.json

It doesn't seem like we're reading the spouse SSN from the JSON to populate anything anyway, but just flagging in case it's used.

arinchoi03 (approved):

PR #5488, opened 2025-01-30
FYST-1395 Validate year for state id page

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1395

## What was done?
- Right now the state id page does not validate year for issue dat and expiration date. To use the date accessor class for validations a range needs to be used for year. 
- The validations for the stat id concern were updated to include a range for the year and have a proper error message saying "Please enter a valid date"

## How to test?
- Added a test for state id classes from primary id and spouse id

Before
Screenshot 2025-01-30 at 2 31 35 PM
After
Screenshot 2025-01-30 at 2 29 33 PM

### DIFF
--- a/app/forms/state_file/state_id_concern.rb
+++ b/app/forms/state_file/state_id_concern.rb
@@ -24,12 +24,12 @@ module StateFile
 
       validates :issue_date,
                 inclusion: {
-                  in: Date.new(2000)..DateTime.now, message: I18n.t('errors.attributes.birth_date.blank')
+                  in: Date.new(2000)..DateTime.now, message: ->(_object, _data) {I18n.t('errors.attributes.birth_date.blank')}
                 },
                 presence: true, unless: -> { id_type == "no_id" }
       validates :expiration_date,
                 inclusion: {
-                  in: Date.new(Time.now.year - 3)..Date.new(Time.now.year + 50), message: I18n.t('errors.attributes.birth_date.blank')
+                  in: Date.new(Time.now.year - 3)..Date.new(Time.now.year + 50), message: ->(_object, _data) { I18n.t('errors.attributes.birth_date.blank') }
                 },
                 presence: true, unless: -> { id_type == "no_id" || non_expiring == "1" }
       validates :state, presence: true, inclusion: { in: States.keys }, unless: -> { id_type == "no_id" }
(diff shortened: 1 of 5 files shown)


app/forms/state_file/state_id_concern.rb
```diff
@@ -22,8 +22,16 @@ module StateIdConcern
...
-      validates :issue_date, presence: true, unless: -> { id_type == "no_id" }
-      validates :expiration_date, presence: true, unless: -> { id_type == "no_id" || non_expiring == "1" }
+      validates :issue_date,
+                inclusion: {
+                  in: Date.new(2000)..DateTime.now, message: I18n.t('errors.attributes.birth_date.blank')
+                },
+                presence: true, unless: -> { id_type == "no_id" }
+      validates :expiration_date,
+                inclusion: {
+                  in: Date.new(Time.now.year - 3)..Date.new(Time.now.year + 50), message: I18n.t('errors.attributes.birth_date.blank')
```
powersurge360, line 32: I think this message has to be in a lambda or else it will be set at class definition time and never change:

See [another example in our codebase](https://github.com/codeforamerica/vita-min/blob/main/app/models/user.rb#L70-L70)

```ruby
validates_confirmation_of :password, message: -> (_object, _data) { I18n.t("errors.attributes.password.not_matching") }
```

app/models/concerns/date_accessible.rb
```diff
@@ -72,8 +72,8 @@ def self.date_writer(*properties)
     # integer. Note that Date#change only accepts :year, :month, and :day as
     # keys, all other keys will be treated as nothing was passed at all.
     #
-    # Note that until all three fragments are passed; month, day, and year, the
-    # year is nonsense. This is expected to be caught by validation.
+    # Note that until all three fragments are passed; month, day, and year
+    # For year, a range must be indicated or else it is not validated.
```
powersurge360, line 76: I don't think this is quite right, this method doesn't ever receive a range directly. The range goes on the validator on the property itself.

powersurge360 (commented):
Nit on documentation comment, lazy load i18n required for approval.

powersurge360 (approved):
Looks good!

PR #5499, opened 2025-01-31
change submission status checking to handle out of order xml

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1733

## What was done?
- This year, we've seen MeF `submissions-status` API responses that include "out of order" submissions statuses for a few states. That is, the XML document we receive includes "Received by State" _after_ "Acknowledgement Received from State". Because we were previously assuming that we would always see the most "recent" or furthest-progress status at the top of the list, we were only looking at first node in the XML document corresponding to each submission ID we poll for. Thus, we have been failing to process a bunch of submissions out of the `transmitted` state.
- This PR changes our submission status checking to look through the whole list of submission statuses for each submission ID we are polling for, first looking for any status that would transition the submission to `ready-for-ack`, then any which would keep it in `transmitted` (and then failing if neither are found, though we should always find the node that marked the submission as `transmitted` in the first place).
## How to test?
- There are submissions on demo that are stuck in this state, so we can do acceptance by checking to see whether they proceed to `:ready-for-ack` and `:accepted` after this is merged to main and the cron job runs the poll service.

### DIFF
--- a/app/services/efile/poll_for_acknowledgments_service.rb
+++ b/app/services/efile/poll_for_acknowledgments_service.rb
@@ -1,4 +1,3 @@
-
 StatusRecordGroup = Struct.new(:irs_submission_id, :state, :xml)
 
 module Efile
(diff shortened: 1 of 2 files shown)


embarnard (approved):
Looks good to me!

PR #5503, opened 2025-02-03
Spike on updating AWS IP ranges

mrotondo:
This adds a webhook action which loads the current AWS IP ranges. The webhook will be called both at server initialization and by an AWS SNS subscription whenever the AWS IP ranges are updated. It also refactors the loading of IPs (from disk & from an AWS URL) into a service.


app/controllers/aws_ip_ranges_webhooks_controller.rb
```diff
@@ -0,0 +1,10 @@
+class AwsIpRangesWebhooksController < ActionController::Base
+  skip_before_action :verify_authenticity_token
+
+  def update_aws_ip_ranges
+    RemoteIpTrustedProxiesService.configure_trusted_proxies(RemoteIpTrustedProxiesService.load_current_aws_ip_ranges)
+    Rails.logger.info("LOGGING AWS SNS SUBSCRIPTION CONFIRMATION FOR UPDATE_AWS_IP_RANGES WEBHOOK: #{params}")
```
mrotondo, line 6: An alternative to logging this out so that we can confirm ownership manually would be for the webhook controller to visit the confirmation URL automatically. I didn't implement that because this work already got blocked by enough other things and I'd like to not make it any more complex.

app/jobs/configure_trusted_proxies_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      Rails.application.routes.url_helpers.url_for(
+        { host: MultiTenantService.gyr.host,
+          controller: 'aws_ip_ranges_webhooks',
+          action: 'update_aws_ip_ranges' }
+      ))
+    Net::HTTP.get_response(uri)
+  end
+
+  def priority
+    PRIORITY_MEDIUM
```
mpidcock, line 14: tiny dust - consider making this PRIORITY_LOW ? since it's running at startup there might be other jobs that are a higher priority, but idk if actually matters

config/routes.rb
```diff
@@ -395,6 +395,8 @@ def scoped_navigation_routes(context, navigation)
     post "/outgoing_email_status", to: "mailgun_webhooks#update_outgoing_email_status", as: :outgoing_email_status
     # OAuth login callback routes
     devise_for :users, path: "hub", only: :omniauth_callbacks, skip: [:session, :invitation], controllers: { omniauth_callbacks: "users/omniauth_callbacks" }
+    # AWS IP ranges update trigger
+    post "/update_aws_ip_ranges", to: "aws_ip_ranges_webhooks#update_aws_ip_ranges", as: :update_aws_ip_ranges
```
mpidcock, line 399: do we need a fyst route also? or only if we separate the deployments?

mpidcock (approved):
I'm not sure how much I can verify that this works, but it makes sense and seems reasonable

config/routes.rb
mrotondo: I decided against it, since they share a server/configuration and this is where webhook routes live currently.

PR #5512, opened 2025-02-04
[FYST-818] MD Military Subtraction

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-818

## What was done?
* Add Line U to calculator where the method is split into filer and spouse
* Line U is added to...
- 502SU XML and PDF
- 502 PDF

I tried to create the methods so that adding more letters would be easy, and so there is a constant with `Md502SuCalculator::CODE_LETTERS` including the valid letters.

## How to test?
* go through the flow for military retirment and see that the data 502SU "U" data has been added


app/lib/submission_builder/ty2024/states/md/documents/md502_su.rb
```diff
@@ -11,10 +11,13 @@ class Md502Su < SubmissionBuilder::Document
             def document
               build_xml_doc("Form502SU", documentId: "Form502SU") do |xml|
                 xml.Subtractions do |subtractions|
-                  if calculated_fields.fetch(:MD502_SU_LINE_AB).positive?
-                    subtractions.OtherDetail do |other_detail|
-                      other_detail.Code "AB"
-                      other_detail.Amount calculated_fields.fetch(:MD502_SU_LINE_AB)
+                  Efile::Md::Md502SuCalculator::CODE_LETTERS.each do |code_letter|
+                    code_letter_calculation = "MD502_SU_LINE_#{code_letter}".to_sym
+                    if calculated_fields.fetch(code_letter_calculation).positive?
```
anisharamnani, line 16: I've seen elsewhere in the XML checking if the `reitirement_ui` flag is enabled, & then showing these fields, but given that there will not be any follow up data without the UI flag enabled, this will not show up to begin with since the value for the letter will not be positive.

spec/lib/pdf_filler/md502_su_pdf_spec.rb
```diff
@@ -6,67 +6,52 @@
...
-        expect(pdf_fields["Your Last Name"]).to eq "Lando"
-        expect(pdf_fields["Your Social Security Number"]).to eq "123456789"
-        expect(pdf_fields["Spouses First Name"]).to eq ""
-        expect(pdf_fields["Text2"]).to eq ""
-        expect(pdf_fields["Spouses Last Name"]).to eq ""
-        expect(pdf_fields["Spouses Social Security Number"]).to eq ""
-        expect(pdf_fields["ab Income from US Government obligations See Instruction 13                         ab"]).to eq("2")
-        expect(pdf_fields["appropriate code letters                                                  TOTAL 1"]).to eq("2")
-      end
+    it "sets fields to the correct values" do
```
anisharamnani, line 39: i removed this, because the presence of interest reports is tested elsewhere in the calculator, so there's no need to have these two tests especially if the only difference is the "ab" field.

spec/lib/pdf_filler/md502_su_pdf_spec.rb
```diff
@@ -6,67 +6,52 @@
...
+        )
+    end
+    let(:spouse_ssn) { "987654321" }
+    let(:primary_ssn) { "345678901" }
     it 'uses field names that exist in the pdf' do
       missing_fields = pdf.hash_for_pdf.keys.map { |k| k.to_s.gsub("'", "&apos;").to_s } - pdf_fields.keys
       expect(missing_fields).to eq([])
     end
 
-    context 'without interest reports' do
```
anisharamnani, line 18: i removed this, because the presence of interest reports is tested elsewhere in the calculator, so there's no need to have these two tests especially if the only difference is the "ab" field.

spec/models/state_file_md_intake_spec.rb
```diff
@@ -223,13 +258,13 @@
   end
 
   describe "#address" do
-  context "a confirmed address" do
-    subject(:intake) { create :state_file_md_intake, :with_confirmed_address }
+    context "a confirmed address" do
```
anisharamnani, line 261: this is some good old formatting courtesy of the RubyMine linter.

app/lib/efile/md/md502_su_calculator.rb
```diff
@@ -3,28 +3,59 @@ module Md
     class Md502SuCalculator < ::Efile::TaxCalculator
       attr_reader :lines, :value_access_tracker
 
+      CODE_LETTERS = ["AB", "U"]
```
anisharamnani, line 6: since we will be adding "v" shortly i tried to loop through the code as much as i could to make things easier to test.

app/lib/efile/md/md502_su_calculator.rb
```diff
@@ -3,22 +3,37 @@ module Md
...
+        age_benefits = @intake.is_filer_55_and_older?(filer) ? 20_000 : 12_500
+        [@intake.sum_1099_r_followup_type_for_filer(:primary, :service_type_military?), age_benefits].min
+      end
+
+      def calculate_line_u_primary
+        calculate_military_per_filer(:primary)
+      end
+
+      def calculate_line_u_spouse
+        @intake.filing_status_mfj? && @intake.spouse_birth_date.present? ? calculate_military_per_filer(:spouse) : 0
```
arinchoi03: spoke about removing `@intake.spouse_birth_date` check here and instead updating specs to include a spouse birthday

app/models/state_file_md_intake.rb
```diff
@@ -158,6 +158,10 @@ def calculate_age(dob, inclusive_of_jan_1)
     super(dob, inclusive_of_jan_1: false)
   end
 
+  def is_filer_55_and_older?(filer)
```
arinchoi03, line 161: Thought about something like this, but as discussed, letting it fail seems rather than hiding this error seems preferable

```suggestion
  def is_filer_55_and_older?(filer)
     return false unless send(filer)&.birth_date&.present?
     
     calculate_age(send(filer).birth_ate, inclusive_of_jan_1: false) >= 55
```

spec/lib/efile/md/md502_su_calculator_spec.rb
```diff
@@ -29,11 +29,90 @@
     end
   end
 
+  describe "#calculate_military_per_filer" do
+    context "when the taxable amount is higher than the age calculation" do
```
arinchoi03, line 33: [dust] small description change -- same for the lower scenario
```suggestion
    context "when the taxable amount is higher than the age benefit amount" do
```

spec/lib/pdf_filler/md502_pdf_spec.rb
```diff
@@ -316,16 +316,16 @@
...
         end
 
         it "fills out subtractions fields correctly" do
           expect(pdf_fields["1SU"]).to eq "ab"
-          expect(pdf_fields["2SU"]).to eq ""
+          expect(pdf_fields["2SU"]).to eq "u"
           expect(pdf_fields["3SU"]).to eq ""
           expect(pdf_fields["4SU"]).to eq ""
-          expect(pdf_fields["13"].to_i).to eq 100
+          expect(pdf_fields["13"].to_i).to eq 200
```
arinchoi03, line 328: [question] why is 502SU pdf tested in this spec vs just in its own spec?

spec/models/state_file_md_intake_spec.rb
```diff
@@ -239,4 +274,58 @@
...
+          :state_file_md1099_r_followup,
+          service_type: "military",
+          state_file1099_r: create(:state_file1099_r, taxable_amount: 2_000, intake: intake, recipient_ssn: intake.spouse.ssn)
+        )
+      end
+      let!(:state_file_md1099_r_followup_without_military) do
+        create(
+          :state_file_md1099_r_followup,
+          service_type: "none",
+          state_file1099_r: create(:state_file1099_r, taxable_amount: 1_000, intake: intake)
```
arinchoi03, line 313: [dust] can we add the `recipient_ssn` just for completeness?

arinchoi03 (approved):
lgtm! ty for walking me through

spec/lib/efile/md/md502_calculator_spec.rb
```diff
@@ -3,7 +3,8 @@
 describe Efile::Md::Md502Calculator do
   let(:filing_status) { "single" }
   let(:county) { "Allegany" }
-  let(:intake) { create(:state_file_md_intake, filing_status: filing_status, residence_county: county) }
+  let(:includes_spouse) { filing_status == "married_filing_jointly" ? :with_spouse : nil}
```
anisharamnani, line 6: this felt like a good-ish way of when the filing status is set to "married_filing_jointly" adding the `:with_spouse` trait.

spec/lib/efile/md/md502_calculator_spec.rb
```diff
@@ -936,6 +938,7 @@
       create(
         :state_file_md_intake,
         filing_status: filing_status,
+        spouse_birth_date: MultiTenantService.statefile.end_of_current_tax_year - 40,
```
anisharamnani, line 941: not my favorite solution :/

arinchoi03: https://github.com/codeforamerica/vita-min/pull/5512/commits/f336cb7f8ffc9dd055b1f6bf2e6accab56af503a
opted to using `with_spouse` instead of hard-coding bday. The qss xml will still be used (overwrites factory xml)

spec/lib/efile/md/md502_calculator_spec.rb
arinchoi03: cool! ty for figuring this out

spec/lib/pdf_filler/md502_pdf_spec.rb
anisharamnani: the values for 502SU appear here and also appear here!

spec/lib/efile/md/md502_r_calculator_spec.rb
```diff
@@ -3,7 +3,13 @@
...
   let(:filing_status) { "single" }
   # df_data_2_w2s has $8000 in federal social security benefits
-  let(:intake) { create(:state_file_md_intake, :df_data_2_w2s, filing_status: filing_status) }
+  let(:intake) {
+    if filing_status == 'married_filing_jointly'
+      create(:state_file_md_intake, :df_data_2_w2s, :with_spouse, filing_status: filing_status)
+    else
+      create(:state_file_md_intake, :df_data_2_w2s, filing_status: filing_status)
+    end
+  }
```
arinchoi03, line 12: this one did _not_ like the conditional setting in MD502SU spec, not sure why -- maybe need to try again 😓

spec/lib/pdf_filler/md502_pdf_spec.rb
arinchoi03: ooh these were stubbed before, i see!

spec/lib/efile/md/md502_r_calculator_spec.rb
```diff
@@ -143,6 +149,7 @@
     context 'when filing MFJ with positive federal social security benefits' do
       let(:filing_status) { "married_filing_jointly" }
       before do
+        allow(intake.direct_file_data).to receive(:fed_ssb).and_return(100)
```
arinchoi03, line 152: i had to change this b/c now `:with_spouse` comes after which changes the `fed_ssb` value. But I thought being explicit here that we have a positive fed_ssb anyway is what we want

spec/factories/state_file_id_intakes.rb
```diff
@@ -153,6 +153,10 @@
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('id_lana_single') }
     end
 
+    trait :with_spouse do
+      :mfj_filer_with_json
+    end
+
```
arinchoi03, line 159: we need a cross-state solution for a standard `with_spouse` fixture

PR #5513, opened 2025-02-04
FYST-1288 Fix follow-up behavior for public school page

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1288

## What was done?
- moved the idea of "made contributions" from az 322 model to az intake model; ignored old column for now so no one sees an error
- made it so the yes/no question only shows up when there are no contribution records yet
## How to test?
- controller and feature spec
- tested the switch from old page to new locally using these steps:
  - check out main
  - fill out intake up to public school page
  - switch to feature branch
  - submit public school page
  - see that it doesn't break and also saves the contribution
- Before
    - it looked basically the same but the yes/no question was always there
- After
image
image
image
create page when there are already some 322's in the table:
image


app/views/state_file/questions/az_qualifying_organization_contributions/_form.html.erb
jenny-heath, line 1: i auto-formatted this page, no meaningful changes

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -24,20 +26,22 @@ def build_contribution
...
-        if current_intake.valid?(:az322) && @az322_contribution.valid?
+        if came_from_old_page || (current_intake.valid?(:az322_form_create) && @az322_contribution.valid?)
           @az322_contribution.save
           redirect_to action: :index, return_to_review: params[:return_to_review]
         else
           render :new
         end
       end
 
+      # remove this after it has been deployed
```
mrotondo, line 40: Can you make a ticket to remove this and add it to this sprint?

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
```diff
@@ -69,16 +68,22 @@ def destroy
...
-          :made_contribution,
           :school_name,
           :ctds_code,
           :district_name,
           :amount,
           :date_of_contribution_day,
           :date_of_contribution_month,
-          :date_of_contribution_year
+          :date_of_contribution_year,
+          state_file_az_intake_attributes: [:made_az322_contributions]
```
mrotondo, line 86: I don't think this is a problem since we only use it to validate the initial form submission, but will intakes from before this change have their `made_az322_contributions` set to the default `:unfilled` despite potentially having 1 or more `az322_contributions` associated with them?

app/controllers/state_file/questions/az_qualifying_organization_contributions_controller.rb
```diff
@@ -90,7 +90,7 @@ def az321_contribution_params
         params.require(:az321_contribution).permit(
             :charity_name, :charity_code, :amount,
             :date_of_contribution_day, :date_of_contribution_month,
-            :date_of_contribution_year, :made_az321_contributions,
```
mrotondo, line 93: Thanks for cleaning up!

app/models/state_file_az_intake.rb
```diff
@@ -103,13 +104,15 @@ class StateFileAzIntake < StateFileBaseIntake
...
   enum eligibility_529_for_non_qual_expense: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_529_for_non_qual_expense
   enum made_az321_contributions: { unfilled: 0, yes: 1, no: 2 }, _prefix: :made_az321_contributions
+  enum made_az322_contributions: { unfilled: 0, yes: 1, no: 2 }, _prefix: :made_az322_contributions
   enum eligibility_lived_in_state: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_lived_in_state
   enum eligibility_out_of_state_income: { unfilled: 0, yes: 1, no: 2 }, _prefix: :eligibility_out_of_state_income
 
   validates :made_az321_contributions, inclusion: { in: ["yes", "no"]}, on: :az321_form_create
+  validates :made_az322_contributions, inclusion: { in: ["yes", "no"]}, on: :az322_form_create
   validates :az321_contributions, length: { maximum: 10 }
+  validates :az322_contributions, length: { maximum: 10 }, on: :az322_form_create
```
mrotondo, line 114: This is non-blocking and maybe trivial, but do we need to restrict this validation to this context?

app/views/state_file/questions/az_public_school_contributions/_form.html.erb
```diff
@@ -9,79 +10,81 @@
...
+          <div class="reveal">
+            <button class="reveal__button"><%= t(".donations_this_year", current_year: current_tax_year + 1) %></button>
+            <div class="reveal__content">
+              <p><%= t(".donations_this_year_details_html", current_year: current_tax_year + 1) %></p>
+            </div>
+          </div>
         </div>
-      </div>
+      <% end %>
+    </div>
```
mrotondo, line 47: Shouldn't the `div.question-with-follow-up` contain both the `*__question` and `*__follow-up` elements? Maybe my diff viewer is messing this up, but it looks like you're closing the container div before the follow-up.

mrotondo: I just noticed that on heroku selecting Yes on the initial/blank page does show the form fields, but then selecting No doesn't hide it again. That might be because of this.

spec/controllers/state_file/questions/az_public_school_contributions_controller_spec.rb
```diff
@@ -82,36 +101,89 @@
...
+          expect(response.body).to include "must be greater than 0"
+        end
       end
     end
 
-    context "with invalid params" do
-      render_views
-
-      let(:invalid_params) do
+    context "DEPRECATED BEHAVIOR; keep tests until page is changed on prod" do
```
mrotondo, line 159: When you make a ticket to remove the old page/behavior, can you include file & line numbers of what should be deleted?

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
mrotondo: And do you think we should do a data migration to keep prior intakes consistent with future ones?

mrotondo (commented):
Couple clarifying questions & one UI issue I noticed that might be linked to a change to the html structure

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
jenny-heath: [done](https://codeforamerica.atlassian.net/browse/FYST-1767?atlOrigin=eyJpIjoiOTBjZDZiNDNlOTlmNDM2NmExMzI4YmJiNTA0YWViOTEiLCJwIjoiaiJ9)!

app/controllers/state_file/questions/az_public_school_contributions_controller.rb
jenny-heath: yes, i included a comment about backfilling the old intakes in that story you requested! do you see any issues with doing that as a follow-up (as opposed to as part of this story)?

app/models/state_file_az_intake.rb
jenny-heath: possibly not! i kept it there because it was already context-restricted (i just renamed the context). i guess the worry with not restricting it would be that someone would hit this validation error in a context where it can't be displayed and see a 500 or something like that. but those contexts don't exist as far as i know so happy to remove it

app/views/state_file/questions/az_public_school_contributions/_form.html.erb
jenny-heath: oops!

spec/controllers/state_file/questions/az_public_school_contributions_controller_spec.rb
jenny-heath: i haven't added those details yet but that's because i think it'll be a lot faster for me to just pick it up. will add all the info if someone else has to do it

mrotondo (approved):
🌵🏫

PR #5516, opened 2025-02-04
Nj 250 add NJ OOI to terms

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/250

## What was done?
- Just decided to check `current_state_code` instead of introducing something new to `StateInformationService` just for this one piece of text, but if any reviewer disagrees that would be fine too.
- Also took this opportunity to make the HTML in the YML files easier to follow

## How to test?
- Any profile in NJ will see `built by Code for America and the New Jersey Office of Innovation` on terms page

image


app/controllers/state_file/questions/terms_and_conditions_controller.rb
```diff
@@ -2,7 +2,10 @@ module StateFile
   module Questions
     class TermsAndConditionsController < QuestionsController
       def edit
-        @li_items = I18n.t('state_file.questions.terms_and_conditions.edit.list_items_html', privacy_policy_link: state_file_privacy_policy_path).split('</li>')
+        ownership = current_state_code == "nj" ?
+          "#{I18n.t('general.code_for_america')} #{I18n.t('general.and')} #{I18n.t('general.nj_ooi')}" :
+          I18n.t('general.code_for_america')
+        @li_items = I18n.t('state_file.questions.terms_and_conditions.edit.list_items_html', ownership: ownership, privacy_policy_link: state_file_privacy_policy_path).split('</li>')
```
jenny-heath, line 8: gotta admit i don't quite understand how it works that these are all missing a closing tag and yet it renders fine?

jenny-heath (approved):
pretty straightforward! could we get someone on the cfa side to do acceptance testing on this to make sure the content is otherwise the same? i tried to skim but would be good to have more 👀

app/controllers/state_file/questions/terms_and_conditions_controller.rb
mluedke2: great point -- the HTML is imperfect (all missing `</li>` because of the `split()` as you noted) but the browser is forgiving and knows how to render it anyway. but I'll do some timeboxed work to see if I can easily get it to proper HTML...

mluedke2: actually @jenny-heath  - the only reason any of these shenanigans with `split()` and the missing `</li>` exist is to support the feature flag just below, which was introduced 12 months ago. Any idea if we can drop support for one of the cases of that flag now that a long time has passed? Then we wouldn't need any of this workaround code at all!

Happy to make that a follow-up ticket too, if that requires a larger discussion. But just wanted to check in case it's easy.

mluedke2: Closing the loop - confirmed with Tiffany that both states of the flag are still needed. I switched the `split()` to a `scan()` with a regex so that the HTML is proper.

app/controllers/state_file/questions/terms_and_conditions_controller.rb
```diff
@@ -2,7 +2,14 @@ module StateFile
   module Questions
     class TermsAndConditionsController < QuestionsController
       def edit
-        @li_items = I18n.t('state_file.questions.terms_and_conditions.edit.list_items_html', privacy_policy_link: state_file_privacy_policy_path).split('</li>')
+        ownership = current_state_code == "nj" ?
+          "#{I18n.t('general.code_for_america')} #{I18n.t('general.and')} #{I18n.t('general.nj_ooi')}" :
```
aloverso, line 6: [dust/pebble] Similar to conversations we've had previously, it feels more correct for us to track a single key for "owner" that says either "Code for America" or "Code for America and New Jersey Office of Innovation" rather than keeping them as two keys joined by an "and" translation. I think changing it will be more resilient across languages so that we're not coding assumptions about grammatical structure but can instead let the language keys handle that.

mluedke2: oh thanks - yes it felt awkward to construct the string the way i did, but no better way came to mind. i'll do it that way...

mluedke2: now updated to use a single translation key with an `nj` option and a `default`

aloverso (approved):

PR #5517, opened 2025-02-04
Add hide suspended users button on users page

embarnard:
## Link to pivotal/JIRA issue
- e.g. [link](https://codeforamerica.atlassian.net/browse/GYR1-510?atlOrigin=eyJpIjoiMWIzMThlMDJiZTM2NDdiN2IzZmEyZWYwNTM0ZTMyNzEiLCJwIjoiaiJ9)

## What was done?
- Add button to hub/users page that will hide and unhide suspended users
- the filter should also apply if you go to the next page
- reordered users so that the suspended users always get shown at the bottom or last page
- made the number of users shown on a page go from 25 -> 50
## How to test?
- use heroku link and go to hub/users, click on "hide suspended users"
- Before
Screenshot 2025-02-04 at 4 15 43 PM

- After
Screenshot 2025-02-04 at 4 16 00 PM


app/views/hub/users/index.html.erb
```diff
@@ -65,3 +70,47 @@
...
+
+          applyFilter();
+
+          button.addEventListener("click", function () {
+              hidden = !hidden;
+              localStorage.setItem("hideSuspended", hidden);
+              applyFilter();
+          });
+
+          const observer = new MutationObserver(mutations => {
```
mrotondo, line 100: This is neat, I didn't know about MutationObserver. Do we actually mutate the contents of the `#users-table` while the page is loaded?

app/controllers/hub/users_controller.rb
```diff
@@ -29,7 +29,7 @@ def index
                    @users.search(params[:search])
                  end
       end
-
+      @users = @users.order(suspended_at: :desc)
```
mrotondo, line 32: TBH it seems like this is a better fix than the show/hide button, since without this line, hiding suspended users on a given page would end up with only a few users visible, and _with_ this line, you don't really need to hide them since they're all tucked away at the end. Maybe something to bring up with PM - ask them to choose between the two?

mrotondo (approved):
A few comments and questions, but looks good!

app/controllers/hub/users_controller.rb
embarnard: @mrotondo yes I agree! Thanks for bringing this up. After you commented, I brought it up to the PMs that the reordering means that the button isn't that useful anymore but they seemed to want to move ahead with it anyways, indicating there might be a change in the future. It seems to be a feature that hub users have been asking for a long time so maybe they want to see how users interact with it before changing implementation plans 🤷🏻‍♀️

PR #5529, opened 2025-02-05
NJ 79 - add page to select income source for 1099-R retirement income

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/79

## What was done?
- Add page to select income source for each 1099-R

## How to test?
- Use Zeus_Two_1099R
- (when running locally) use `Flipper.enable(:show_retirement_ui)` in rails console to enable the UI toggle
- (in heroku) sign into admin hub, then in the heroku app, navigate to `/en/hub/flipper/features` and enable `show_retirement_ui`

image
image

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -546,11 +546,9 @@
     line-height: 2.4rem;
   }
 
-  .review-header {
-    .with-top-separator {
-      padding-top: 25px;
-      border-top: 1px solid $color-state-file-separator;
-    }
+  .with-top-separator {
+    padding-top: 25px;
+    border-top: 1px solid $color-state-file-separator;
   }
 
   .review-section {
--- a/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
@@ -39,7 +39,7 @@ module StateFile
       end
 
       def self.show?(intake)
-        intake.state_file1099_rs.length.positive?
+        Flipper.enabled?(:show_retirement_ui) && intake.state_file1099_rs.length.positive?
       end
 
       def prev_path
--- a/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
@@ -1,0 +1,71 @@
+module StateFile
+  module Questions
+    class NjRetirementIncomeSourceController < QuestionsController
+      include ReturnToReviewConcern
+
+      before_action :load_1099r
+
+      def load_1099r
+        @index = params[:index].present? ? params[:index].to_i : 0
+        @state_file_1099r =
+          if params[:index].present?
+            current_intake.state_file1099_rs[params[:index].to_i]
+          else
+            current_intake.state_file1099_rs.first
+          end
+
+        @name_1099r = @state_file_1099r.payer_name
+        @taxpayer_name = @state_file_1099r.recipient_name
+        @amount = @state_file_1099r.taxable_amount
+      end
+
+      def initialized_edit_form
+        attribute_keys = Attributes.new(form_class.attribute_names).to_sym
+        state_specific_followup = retrieve_or_create_state_specific_followup
+        form_class.new(state_specific_followup, form_class.existing_attributes(state_specific_followup).slice(*attribute_keys))
+      end
+
+      def initialized_update_form
+        form_class.new(retrieve_or_create_state_specific_followup, form_params)
+      end
+
+      def retrieve_or_create_state_specific_followup
+        unless @state_file_1099r.state_specific_followup.present?
+          @state_file_1099r.state_specific_followup = StateFileNj1099RFollowup.create
+          @state_file_1099r.save
+        end
+
+        @state_file_1099r.state_specific_followup
+      end
+
+      def self.show?(intake)
+        intake.state_file1099_rs.length.positive?
+      end
+
+      def prev_path
+        options = {}
+        options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
+        prev_index = @index - 1
+        if prev_index.negative?
+          super
+        else
+          options[:index] = prev_index
+          NjRetirementIncomeSourceController.to_path_helper(options)
+        end
+      end
+
+      def next_path
+        options = {}
+        options[:return_to_review] = params[:return_to_review] if params[:return_to_review].present?
+        next_index = @index + 1
+        if next_index >= current_intake.direct_file_data.form1099r_nodes.length
+          super
+        else
+          options[:index] = next_index
+          NjRetirementIncomeSourceController.to_path_helper(options)
+        end
+      end
+    end
+  end
+end
+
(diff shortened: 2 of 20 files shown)


spec/forms/state_file/nj_retirement_income_source_form_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      it "is valid" do
+        expect(form.valid?).to eq true
+      end
+    end
+  end
+
+  describe ".save" do
+    context "when saving a new county" do
```
mmazanec22, line 33: [pebble] Copy paste error

app/views/state_file/questions/nj_retirement_income_source/edit.html.erb
```diff
@@ -0,0 +1,41 @@
+<%
+    title = t(".title" )
+    content_for :page_title, title
+%>
+
+<% content_for :card do %>
+  <h1 class="h2"><%= title %></h1>
+  <h2 class="h3"><%= t(".subtitle") %></h2>
```
mmazanec22, line 8: [pebble] Does it make sense to use an h2 here? It's not really organizing content. Maybe an [hgroup](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup)?  I'm not sure whether assistive tech will do anything with that but it's what the element is intended for

mmazanec22 (approved):
Looks good!

app/views/state_file/questions/nj_retirement_income_source/edit.html.erb
aloverso: I like the idea of the `hgroup`. It felt weird for it to be a `p` given that it is both visually & semantically distinct from the three `p` tags that follow it for the 1099-R details.

spec/controllers/state_file/questions/nj_retirement_income_source_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
+require "rails_helper"
+
+RSpec.describe StateFile::Questions::NjRetirementIncomeSourceController do
+  let(:intake) { create :state_file_nj_intake }
+  before do
+    sign_in intake
+    allow(Flipper).to receive(:enabled?).and_call_original
+    allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
```
jachan, line 8: [DUST] This is a cool way to do testing for unreleased features. Is Line 7 necessary?

app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        attribute_keys = Attributes.new(form_class.attribute_names).to_sym
+        state_specific_followup = retrieve_or_create_state_specific_followup
+        form_class.new(state_specific_followup, form_class.existing_attributes(state_specific_followup).slice(*attribute_keys))
+      end
+
+      def initialized_update_form
+        form_class.new(retrieve_or_create_state_specific_followup, form_params)
+      end
+
+      def retrieve_or_create_state_specific_followup
```
jachan, line 32: [DUST] What feature in the tests does state_specific_followup serve?

Right now this appears to default to the NJ followup unless a different state wants to implement their own followup. Is this a feature other states will be using or is it something we *think* they will use in the future?

jachan (approved):
Cool to learn about Flipper! One question about the future of the state_specific_followup code.

spec/controllers/state_file/questions/nj_retirement_income_source_controller_spec.rb
aloverso: It is necessary, because Flipper is also getting called by some other dependent pieces of the code for `show_intercom` so without Line 7, that will fail when testing

app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
jachan: Anne helped explain this in Slack!

This allows the user to edit the state_specific followup.

1. User visits page in the flow for the first time. No state_specific_followup is present on the model by default, so we create it.
2. User answers question, data gets saved to the state_file_1099_r_followup model(that we created) associated with the 1099-R
3. User visits review page and clicks edit on the 1099-R section to come back to the page
4. At this point, the state_specific_followup for that 1099-R already exists and has data in it, so we return that data to the page rather than creating a new empty data structure

aloverso: Discussed in [slack thread](https://njcio.slack.com/archives/C0590CD1EE8/p1738785229902119), resolving this question!

app/assets/stylesheets/_state-file.scss
```diff
@@ -546,11 +546,9 @@
     line-height: 2.4rem;
   }
 
-  .review-header {
-    .with-top-separator {
-      padding-top: 25px;
-      border-top: 1px solid $color-state-file-separator;
-    }
+  .with-top-separator {
```
tahsinaislam, line 549: (rock) i think we need to keep this because we use in the abstract review header partial

config/locales/en.yml
```diff
@@ -3524,6 +3524,21 @@ en:
           no_id: No driver's license or state ID / Prefer not to share
           subtitle: Details from your state-issued ID, if you have one, will help New Jersey combat stolen-identity tax fraud and protect you and your tax refund.
           title: Confirm Your Identity (Optional)
+      nj_retirement_income_source:
+        edit:
+          doc_1099r_label: '1099-R:'
+          helper_description_html: |
+            <p>U.S. military pensions result from service in the Army, Navy, Air Force, Marine Corps, or Coast Guard. Generally, qualifying military pensions and military survivor's benefit payments are paid by the U.S. Defense Finance and Accounting Service.</p>
+            <p>Civil service pensions and annuities do not count as military pensions, even if they are based on credit for military service. Generally, civil service annuities are paid by the U.S. Office of Personnel Management.</p>
+          helper_heading: What counts as a U.S. military pension?
```
tahsinaislam, line 3533: (pebble) is it possible to break out each line in its own key instead of using the <p> tag. i think it makes it simpler in transifex to process?

tahsinaislam (approved):
Looks good but just had a few comments

app/assets/stylesheets/_state-file.scss
aloverso: I think this change should still allow it to be used elsewhere? It just makes it available as its own class so that other elements can successfully use the `with-top-separator` class without being within `review-header`. But in my understanding, review-header should still be able to use the class without issues.

config/locales/en.yml
aloverso: Hm I didn't realize. Can we chat about this in our sync with CFA tomorrow rather than for this ticket? We are using this pattern (of `p` tags within description html blocks) _very_ extensively in our NJ content and I feel like this should be discussed as a more overall change to our approach

tahsinaislam: yeah that makes sense to me! this is also something new we realized. i'll approve this!

tahsinaislam (approved):

PR #5533, opened 2025-02-05
make nc review page use shared review header

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1735

## What was done?
- Remove NC's custom review header, and have them use the shared one.
## How to test?
- Go through the NC flow with various personas and verify that the income review box shows up on the final review page, and that the Unemployment Compensation box does not.
- Before
image
- After
image


spec/features/state_file/review_page_spec.rb
```diff
@@ -108,56 +120,6 @@
...
-      expect(page).to have_text strip_html_tags(I18n.t("state_file.questions.unemployment.index.lets_review"))
-
-      # delete a 1099G (there's only one)
-      recipient_name = intake.state_file1099_gs.last.recipient_name
-      click_on I18n.t("general.delete")
-      # redirects to new because there are no 1099Gs left, need to select "no" in order to continue
-      expect(page).to have_text I18n.t("state_file.questions.unemployment.destroy.removed", name: recipient_name)
-      choose I18n.t("general.negative")
-    end
-  end
```
arinchoi03, line 159: are we losing coverage of tests here where we test if we can edit/delete a 1099G? (or was that only a functionality that NC review page had?)

spec/features/state_file/review_page_spec.rb
```diff
@@ -1,15 +1,20 @@
 require "rails_helper"
 
-RSpec.feature "Completing a state file intake", active_job: true do
+RSpec.feature "Completing a state file intake", active_job: true, js: true do
   include MockTwilio
   include StateFileIntakeHelper
 
+  def wait_for_device_info
+    sleep(2)
+    expect(page.find_all('input[name="state_file_income_review_form[device_id]"]', visible: false).last.value).to be_present
```
arinchoi03, line 9: it looks like the tests pass with just the `sleep(2)` or (`sleep(1)` for me locally). I'm guessing the next `expect` is to just make sure we can still get to the next page correctly (and documentation purposes), which seems good.

I wonder if this is going to make our tests run slightly longer on circle ci? Is there any way to slow this down only on dev and not on circle ci?

also, where is the device info javascript that runs?

arinchoi03 (approved):
approved with some considerations, maybe `sleep(1)` to make this not so long?

spec/features/state_file/review_page_spec.rb
mrotondo: Thanks for poking on this. Kevin had actually linked me to a [blog post](https://nts.strzibny.name/avoid-sleep-rails-system-tests/) about sleeping in Capybara tests which includes a method for waiting for something in a tight loop with some timeout, and shows how to use Capybara's default timeout for the purpose. I switched over to that, so it should only add as much time is needed for the element to show up, which seems like it should be no additional time on CircleCI (given that these were already passing there).

The device info javascript is in app/javascript/lib/efile_security_information.js, and is called on app/javascript/listeners/index.js:57

spec/features/state_file/review_page_spec.rb
mrotondo: No, this was (mostly) duplicative of the code in `edit_unemployment` in this file, which executes for every state except NJ (which doesn't allow them to be edited)

arinchoi03 (approved):
Neato 💯

spec/features/state_file/review_page_spec.rb
arinchoi03: i missed that! ty

PR #5542, opened 2025-02-07
NJ 285 - return error if state_wages is 0 but wages is present

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/285

## What was done?
- Added validation to record to ensure that w2 is not valid if wages is present but state_wages is 0
- Added warning to W2 screen if federal wages is non-zero but state wages is 0.

## How to test?
- Devito Single
  - During W2 import, change XML
  - Set StateWagesAmt (Box 16) to 0 
  - Set StateIncomeTaxAmt (Box 17) to 0. If this is not done, the record will be invalid for the wrong reason (because it expects Box 17 to be less than Box 16)
  - Validate that warning exists and that you are unable to continue with the W2 submission.

- Before
- After

image

image
Warning appears

image
When attempting to submit without correcting

image
StateWagesAmount is now required to be greater than 0 when saving a W2

### DIFF
--- a/app/controllers/state_file/questions/income_review_controller.rb
+++ b/app/controllers/state_file/questions/income_review_controller.rb
@@ -22,9 +22,9 @@ module StateFile
 
       def update
         if current_intake.allows_w2_editing? && @w2s.any? do |w2|
-            w2.check_box14_limits = true
-            !w2.valid?(:state_file_edit)
-          end
+             w2.check_box14_limits = true
+             !w2.valid?(:state_file_edit)
+           end
           flash[:alert] = I18n.t("state_file.questions.income_review.edit.invalid_w2")
           render :edit
         else
@@ -35,6 +35,8 @@ module StateFile
       private
 
       def should_show_warning?(w2, w2_count_for_filer)
+        return true if w2.wages.present? && !w2.state_wages_amount.positive?
+
         return false if StateFile::StateInformationService
           .w2_supported_box14_codes(w2.state_file_intake.state_code)
           .none? { |code| code[:name] == "UI_WF_SWF" || code[:name] == "FLI" }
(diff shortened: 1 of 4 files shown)


mluedke2 (approved):
looks good to me!

app/controllers/state_file/questions/income_review_controller.rb
```diff
@@ -22,9 +22,9 @@ def set_sorted_vars
 
       def update
         if current_intake.allows_w2_editing? && @w2s.any? do |w2|
```
powersurge360, line 24: [pebble] I know not strictly part of this PR but I find this block sneaky and don't notice it's not literally part of the if block. Feels like a little much to do in an if check.

powersurge360 (approved):
Looks good. Left a comment on a dense if statement. Up to you to act on it now or later imo.

app/controllers/state_file/questions/income_review_controller.rb
jachan: Thanks! Looks like this will be addressed by FYST-1046 when that gets merged so I'm going to leave it as-is.

amurthy2:
I see the styling for [Added warning to W2 screen if federal wages is non-zero but state wages is 0] and worry that with multiple W2 errors, these warnings become visually overwhelming 🤔  

I would recommend something slimmer, like what we previously explored: 
image
which is similar to a styling [Gov.uk uses](https://design-system.service.gov.uk/components/warning-text/)

amanifarooque:
Screenshots in demo, for reference.

### New Jersey
* income review
  * image
* w2 page on attempt to save/continue
  * image
* xml checks out ✅ 

### Maryland
image
image
on attempt to save:
image
after resolution:
image

* put 10K in box 16, 200 in box 17
* demo intake ID: **md 967**

PR #5548, opened 2025-02-08
FYST-1762 Fix query for existing intakes WITH direct file data; if intake with ssn exists, log in with that

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1762

## What was done?
- the query to find an existing intake on verification code controller was accidentally returning intakes that do not have direct file data (we think the intention was the opposite)
- the sign in and redirect method was picking a random first intake to log in, but we should prioritize ones with an ssn (therefore, have done data import)
## How to test?
- Added one unit test for each change that failed before the change and passes after. Still feel like the general testing approach is not robust enough but a client was just reported to have this issue so it seems like maybe we should get a fix out and then look at making more code improvements later.


jnf (approved):
new tests close the gap well!

PR #5549, opened 2025-02-08
Revert "FYST-1288 Fix follow-up behavior for public school page (#5513)"

jenny-heath:
This reverts commit 4bb14f841df3dae871bd8473a393474500bf947c.

We found that you cannot advance past the AZ school credits page if you answer "no" to the initial question. We are reverting the whole feature for now and will revisit when we have more time so as not to block any filers on this page.


mpidcock (approved):

jenny-heath:
realized we shouldn't just revert this because it adds a database column

PR #5551, opened 2025-02-08
FYST-1774 Fix w2 current step error

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1774

## What was done?
- When logging in, a client is redirected to current_step on intake; when step is w2, it throws a 404 because that page requires an id in the params. This is a quick fix to send them to income review instead. We can talk about a better solution later imo
## How to test?
- Wrote tests for `controller_for_current_step` method


mrotondo (approved):
Looks good - we discussed possible resolutions & it seems like improving how we store `current_step` to include params would be good

PR #5553, opened 2025-02-10
[GYR] Fix a reference in the prod creds; we were trying to use a demo number when connecting calls instead of the more strongly verified prod number

jnf:
## Link to pivotal/JIRA issue
n/a

## What was done?
Replace an incorrect value in the prod creds related to how GYR connects phone calls.

## How to test?
Probably best way to test would be to pull the branch locally and starting a console as prod environment. The reference fetched for connecting calls should be the known-good GYR prod number (650) rather than the testing/demo number (415)


embarnard (approved):
If phone number is `+16504494147` looks good!

PR #5554, opened 2025-02-10
changed NJ intake to has_many analytics

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/281

## What was done?
- Added ability to attach multiple analytics records to one intake. This is important since we want 1 analytics record per SUBMISSION, rather than per intake.
- Example of when this is relevant:
  - Taxpayer submits their return
  - It gets rejected by the IRS
  - Taxpayer updates their return and resubmits
  - It gets accepted by the IRS
- This will result in 2 submissions associated with the intake.

## How to test?
- run tests in `vita-min/spec/jobs/state_file/create_nj_analytics_record_job_spec.rb`

- Before
- After

### DIFF
--- a/app/models/state_file_nj_intake.rb
+++ b/app/models/state_file_nj_intake.rb
@@ -114,7 +114,7 @@
 class StateFileNjIntake < StateFileBaseIntake
   self.ignored_columns += ["primary_signature_pin", "spouse_signature_pin"]
   encrypts :account_number, :routing_number, :raw_direct_file_data, :raw_direct_file_intake_data
-  has_one :state_file_nj_analytics
+  has_many :state_file_nj_analytics
 
   enum household_rent_own: { unfilled: 0, rent: 1, own: 2, neither: 3, both: 4 }, _prefix: :household_rent_own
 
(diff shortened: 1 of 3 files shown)


mmazanec22 (approved):

app/models/state_file_nj_intake.rb
```diff
@@ -114,7 +114,7 @@
 class StateFileNjIntake < StateFileBaseIntake
   self.ignored_columns += ["primary_signature_pin", "spouse_signature_pin"]
   encrypts :account_number, :routing_number, :raw_direct_file_data, :raw_direct_file_intake_data
-  has_one :state_file_nj_analytics
+  has_many :state_file_nj_analytics
```
aloverso, line 117: [dust] This might be my ignorance of ActiveRecord, but are we "allowed" to change the model file directly for this kind of relation, or this that something we should have a migration for?

aloverso (approved):
interesting edge case! (intake vs submission)
makes sense, one small question

app/models/state_file_nj_intake.rb
jnf: yeah, should be fine to change. the db structure is the same, we're just letting rails know to look for many potential associated records instead of strictly retrieving zero or one.

jnf (approved):

PR #5556, opened 2025-02-10
FYST-829 MD Public Safety Employee Subtraction

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-829

## Is PM acceptance required?

## What was done?
* Add Line U form to 502SU 
* This change also affects 502R Line 10 A and B. Since the primary and spouse values are used separately for those calculations.


app/lib/efile/md/md502_su_calculator.rb
jenny-heath, line 1: reminder to delete the comments in `calculate`!

spec/lib/efile/md/md502_su_calculator_spec.rb
```diff
@@ -94,6 +132,35 @@
     end
   end
 
+  describe "#calculate_line_v_primary" do
+    before do
+      allow(instance).to receive(:calculate_public_safety_employee).with(:primary).and_return 10_000
+    end
+
+    it "returns the value for #calculate_military_per_filer" do
```
jenny-heath, line 140: i assume this is a copy-paste—should it be "calculate_public_safety_employee"?

spec/lib/efile/md/md502_su_calculator_spec.rb
```diff
@@ -94,6 +132,35 @@
...
+    end
+  end
+
+  describe "#calculate_line_v_spouse" do
+    before do
+      allow(instance).to receive(:calculate_public_safety_employee).with(:spouse).and_return 10_000
+    end
+
+    context "when a single filer" do
+      it "returns the value for #calculate_military_per_filer" do
```
jenny-heath, line 151: same

spec/models/state_file_md_intake_spec.rb
```diff
@@ -329,4 +329,68 @@
...
+        )
+      end
+
+      context "when only one followup is present" do
+        it "returns 0" do
+          expect(intake.sum_two_1099_r_followup_types_for_filer(:primary, :income_source_pension_annuity_endowment?, :service_type_public_safety? )).to eq(0)
+        end
+      end
+
+      context "when both followups are present" do
```
jenny-heath, line 366: sorry if i'm missing something but it doesn't look like there's a difference in the setup between this block and the one on line 360, though it says that one has only one follow-up and the other has 2?

spec/models/state_file_md_intake_spec.rb
```diff
@@ -329,4 +329,68 @@
...
+        it "returns the sum of the taxable amount" do
+          expect(intake.sum_two_1099_r_followup_types_for_filer(:spouse, :income_source_pension_annuity_endowment?, :service_type_public_safety? )).to eq(2500)
+        end
+      end
+    end
+
+    context "when neither followup present" do
```
jenny-heath, line 373: i think this description is a little misleading—the followup exists in the database but (i think?) the reason it doesn't count towards the sum is that the income source is other? i would say something like `context "when none of the 1099-Rs are from a qualifying source"` or something like that

app/models/state_file_md_intake.rb
```diff
@@ -231,6 +231,17 @@ def sum_1099_r_followup_type_for_filer(primary_or_spouse, followup_type)
     end
   end
 
+  def sum_two_1099_r_followup_types_for_filer(primary_or_spouse, followup_type_1, followup_type_2)
```
jenny-heath, line 234: major nit but i think this would be clearer if the arguments were named like this instead:
`followup_type_1` -> `income_source`
`followup_type_2 ` -> `service_type`

spec/models/state_file_md_intake_spec.rb
jenny-heath: oh i see now, it's not that the followups are _present_, it's that they qualify them for the subtraction. i think i'd rename these contexts to "when the income source qualifies them" (or doesn't) or "when the followup has a qualifying service type" (or doesn't)

jenny-heath (approved):
i like your tests! left some non-blocking suggestions about naming stuff

app/models/state_file_md_intake.rb
anisharamnani: 🤔 ya i wanted to make this method re-usable for other follow up types, so that's why the generic-ish name. but, i can do that.

anisharamnani:
Thanks Jenny for the comments & review :)

PR #5561, opened 2025-02-11
NJ 80 - line 20a taxable retirement income

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/80

## What was done?
- add calculation for line 20a (based on income source) & display in XML/PDF
- control XML/PDF based on Flipper retirement UI
- Update calculation for line 27 to include line 20a in sum
- clean up a bit of XML tests to avoid duplicating logic with calculator tests

## How to test?
Use persona with 1099-Rs (zeus two 1099r)
Don't forget to enable Flipper flag

image
image

### DIFF
--- a/app/lib/pdf_filler/nj1040_pdf.rb
+++ b/app/lib/pdf_filler/nj1040_pdf.rb
@@ -269,6 +269,23 @@ module PdfFiller
                                                  ]))
       end
 
+      # line 20a
+      if @xml_document.at("PensAnnuitAndIraWithdraw") && Flipper.enabled?(:show_retirement_ui)
+        retirement_income = @xml_document.at("PensAnnuitAndIraWithdraw").text.to_i
+        answers.merge!(insert_digits_into_fields(retirement_income, [
+          "141",
+          "140",
+          "139",
+          "138",
+          "137",
+          "136",
+          "undefined_56",
+          "undefined_55",
+          "undefined_54",
+          "20a"
+        ]))
+      end
+
       if @xml_document.at("TotalIncome").present?
         total_income = @xml_document.at("TotalIncome").text.to_i
         answers.merge!(insert_digits_into_fields(total_income, [
--- a/app/lib/pdf_filler/nj1040_pdf.rb
+++ b/app/lib/pdf_filler/nj1040_pdf.rb
@@ -270,7 +270,7 @@ module PdfFiller
       end
 
       # line 20a
-      if @xml_document.at("PensAnnuitAndIraWithdraw") && Flipper.enabled?(:show_retirement_ui)
+      if @xml_document.at("PensAnnuitAndIraWithdraw")
         retirement_income = @xml_document.at("PensAnnuitAndIraWithdraw").text.to_i
         answers.merge!(insert_digits_into_fields(retirement_income, [
           "141",
--- a/spec/factories/state_file_nj_intakes.rb
+++ b/spec/factories/state_file_nj_intakes.rb
@@ -159,10 +159,6 @@ FactoryBot.define do
       end
     end
 
-    trait :with_1099_rs_synced do
-      after(:create, &:synchronize_df_1099_rs_to_database)
-    end
-
     trait :df_data_2_w2s do
       raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_zeus_two_w2s') }
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_zeus_two_w2s') }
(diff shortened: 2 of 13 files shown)


spec/factories/state_file_nj_intakes.rb
```diff
@@ -159,6 +159,10 @@
       end
     end
 
+    trait :with_1099_rs_synced do
+      after(:create, &:synchronize_df_1099_rs_to_database)
+    end
+
```
mluedke2, line 165: [pebble] - i see this in the AZ and MD calculator tests, but not in the NJ ones... do we really need this? Or should we be using it in our calculator tests too?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -422,15 +422,32 @@
       end
     end
 
+    describe "retirement income - line 20a" do
+      before do
+        allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
```
mluedke2, line 427: [boulder] - I believe the `show_retirement_ui` flag being `false` is the current state for users, so we'd want it covered by tests to make sure nothing is leaking out.

jnf (approved):
lgtm!

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -446,13 +463,9 @@
...
-          expected_line_15_w2_wages = 200_000
-          expected_line_16a = 500
-          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_15).and_return expected_line_15_w2_wages
-          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_16a).and_return expected_line_16a
-          expected_total = expected_line_15_w2_wages + expected_line_16a
-          expect(xml.at("WagesSalariesTips").text).to eq(expected_line_15_w2_wages.to_s)
-          expect(xml.at("GrossIncome").text).to eq(expected_total.to_s)
+          expected = 200_000
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_29).and_return expected
+          expect(xml.at("GrossIncome").text).to eq(expected.to_s)
```
mluedke2, line 468: +1 to mocking the expectation directly and not redoing the calculator test

mluedke2 (changes requested):
unless i'm mistaken we also need the flag-off tests, but other than that all looks good to me

spec/factories/state_file_nj_intakes.rb
aloverso: good catch!! this is from debugging the database setup issue, was one of the things I tried (but ultimately unnecessary) and forgot to clean up. thanks

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
aloverso: My rationale behind not testing the disabled flag state was that it felt like testing internal behavior of the flag rather than desired behavior of our code, but as I'm typing that out, I feel like my `if ... && enabled...` is in fact something OUR code does. Will add!

aloverso: Do you think we should go back and add flag-off tests to the #5529 controller as well?

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -270,7 +270,7 @@ def hash_for_pdf
       end
 
       # line 20a
-      if @xml_document.at("PensAnnuitAndIraWithdraw") && Flipper.enabled?(:show_retirement_ui)
+      if @xml_document.at("PensAnnuitAndIraWithdraw")
```
aloverso, line 273: @mluedke2 neat thing - adding the explicit Flipper disabled tests allowed me to clean up this line of code, because the test actually confirms that if the flipper is disabled, the XML will be nil, which means we don't need an extra feature flag check here in the PDF. pretty cool!

mluedke2: oh nice! 👍

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
mluedke2: oh good idea, thanks for adding that one too 👍

mluedke2 (approved):
thanks for adding those flagged-off tests! all good to me!

jachan (approved):
Makes sense to me! Appreciate the Flipper-off tests, and the example for 1099r setup. They are good patterns to keep in mind.

PR #5564, opened 2025-02-11
[FYST-819] MD Pension Exclusion

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-819

## What was done?
Line 11A and 11B to 502R were added. First, there were methods added to the `StateFileMdIntake` to determine if the filer qualified. Then, if qualified there is a `calculate_line_11` method that both filers can be referred to. The calculations are a bit gnarly. I'd recommend reading the ticket first before reviewing this PR. I tried to copy the calculations here, but it didn't work so well.


spec/lib/efile/md/md502_r_calculator_spec.rb
```diff
@@ -222,4 +222,114 @@
...
     end
   end
+
+  describe "#calculate_line_11a" do
+    context "when primary qualifies" do
+      before do
+        allow_any_instance_of(StateFileMdIntake).to receive(:qualifies_for_pension_exclusion?).with(:primary).and_return(true)
+        allow_any_instance_of(Efile::Md::Md502RCalculator).to receive(:calculate_line_1a).and_return(100_000)
+        allow_any_instance_of(Efile::Md::Md502RCalculator).to receive(:calculate_line_10a).and_return(80_000)
+        allow_any_instance_of(Efile::Md::Md502RCalculator).to receive(:calculate_line_9a).and_return(line_9_a_value)
```
jenny-heath, line 232: oh i like the approach of just having one of the 2 values being compared tied to a variable!

spec/lib/efile/md/md502_r_calculator_spec.rb
```diff
@@ -222,4 +222,114 @@
...
+        let(:line_9_b_value) { 39_501}
+        it "returns 0" do
+          main_calculator.calculate
+          expect(instance.lines[:MD502R_LINE_11B].value).to eq 0
+        end
+      end
+    end
+
+    context "when the spouse does not qualify" do
+      context "when the filer is not married filing jointly" do
```
jenny-heath, line 312: nit: could define `filing_status` within the block just to be clearer

app/lib/efile/md/md502_r_calculator.rb
```diff
@@ -79,6 +81,22 @@ def calculate_line_10a
...
+        result = [qualifying_pension_minus_ssn_or_railroad, tentative_exclusion].min
+        [result, 0].max
+      end
+
+      def calculate_line_11a
+        @intake.qualifies_for_pension_exclusion?(:primary) ? calculate_line_11(:primary) : 0
+      end
+
+      def calculate_line_11b
+        @intake.filing_status_mfj? && @intake.qualifies_for_pension_exclusion?(:spouse) ? calculate_line_11(:spouse) : 0
```
jenny-heath, line 98: i think this mfj check is redundant with the method on intake but also that doesn't really matter 🤷🏻‍♀️

jenny-heath (approved):
:chefs_kiss:

spec/lib/efile/md/md502_r_calculator_spec.rb
anisharamnani: that's a good idea! i thought about that too.

app/lib/efile/md/md502_r_calculator.rb
anisharamnani: i chatted with jenny! i need this check here because it's an OR check, so `qualifies_for_pension_exclusion?(:spouse)` could return a false positive.

PR #5565, opened 2025-02-11
FYST-1790 Fix intake clearing on landing page

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1790

## What was done?
- the code in landing page controller that is ostensibly supposed to clear the current intake has i guess never been doing that because it did not have access to the `current_intake` method for state file (just the one in application controller, which returned nothing). for in-state cases this didn't matter because logging in the same type of record would replace the old one.
- after several passes at this (not intentional, oops) i landed on this:
  - in production, it will log out in-state intakes only
  - in other environments, it will look for every type of state intake and log it out
- you can kind of see the journey i went on by looking at the individual commits
## How to test?
- updated the shared example to test the out of state case in different environments

### DIFF
--- a/app/controllers/state_file/landing_page_controller.rb
+++ b/app/controllers/state_file/landing_page_controller.rb
@@ -19,9 +19,7 @@ module StateFile
 
     def update
       all_current_intakes = StateFile::StateInformationService.active_state_codes.map { |c| send("current_state_file_#{c}_intake".to_sym) }.compact
-      if all_current_intakes.present?
-        all_current_intakes.each { |intake| sign_out intake }
-      end
+      all_current_intakes.each { |intake| sign_out intake }
       intake = StateInformationService.intake_class(params[:us_state]).new(
         visitor_id: cookies.encrypted[:visitor_id],
         source: session[:source],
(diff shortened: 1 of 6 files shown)


anisharamnani (approved):
I appreciated the tip to go through the PR commit by commit. 

🤔 I don't have any feedback or questions here, but I understand your approach. Thanks Jenny!

app/controllers/state_file/landing_page_controller.rb
```diff
@@ -18,7 +18,10 @@ def edit
     end
 
     def update
-      sign_out current_intake if current_intake.present?
+      all_current_intakes = StateFile::StateInformationService.active_state_codes.map { |c| send("current_state_file_#{c}_intake".to_sym) }.compact
+      if all_current_intakes.present?
```
mrotondo, line 22: non-blocking & minor, but i don't think you need this `.present?` check, since it'll be an array no matter what, and if there are no elements, the `.each` block will just not execute any times

mrotondo (approved):
looks good!

app/controllers/state_file/landing_page_controller.rb
jenny-heath: good point

PR #5568, opened 2025-02-12
Fyst 1780 md strip unexpected characters from drivers license number

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1780

## What was done?
- id number is now stripped of long dash characters
## How to test?
- Enter some of these dashes in the License Id field then check that they are absent in the XML

### DIFF
--- a/app/models/state_id.rb
+++ b/app/models/state_id.rb
@@ -36,6 +36,8 @@ class StateId < ApplicationRecord
     end
   end
 
+  private
+
   def remove_long_dashes(input)
     input.gsub(/[\u2013\u2014]/, '')
   end
--- a/app/models/state_id.rb
+++ b/app/models/state_id.rb
@@ -31,5 +31,15 @@ class StateId < ApplicationRecord
     if non_expiring_changed?(to: true)
       self.expiration_date = nil
     end
+    if id_number.present?
+      self.id_number = remove_long_dashes(id_number)
+    end
+  end
+
+  def remove_long_dashes(input)
+    input.gsub(/[\u2013\u2014]/, '')
   end
 end
+
+
+
--- a/spec/models/state_id_spec.rb
+++ b/spec/models/state_id_spec.rb
@@ -37,6 +37,7 @@ describe StateId do
          .and change(state_id, :state).to(nil)
       end
     end
+
     context "when non_expiring" do
       it " clears expiration_date" do
         state_id.non_expiring = true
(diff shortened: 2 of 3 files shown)


app/models/state_id.rb
```diff
@@ -31,5 +31,15 @@ class StateId < ApplicationRecord
...
     end
+    if id_number.present?
+      self.id_number = remove_long_dashes(id_number)
+    end
+  end
+
+  def remove_long_dashes(input)
+    input.gsub(/[\u2013\u2014]/, '')
   end
 end
```
arinchoi03, line 42: wondering if this should be a private method unless we're planning to use it elsewhere

app/models/state_id.rb
```diff
@@ -31,5 +31,15 @@ class StateId < ApplicationRecord
     if non_expiring_changed?(to: true)
       self.expiration_date = nil
     end
+    if id_number.present?
+      self.id_number = remove_long_dashes(id_number)
+    end
+  end
+
+  def remove_long_dashes(input)
+    input.gsub(/[\u2013\u2014]/, '')
```
squanto, line 40: <nit> I feel like there's a third dash `\u002D` but am unsure

spec/models/state_id_spec.rb
```diff
@@ -45,5 +45,11 @@
         }.to change(state_id, :expiration_date).to(nil)
       end
     end
+    context "when a long dash is in the id number" do
+      it "removes the dash" do
+        state_id.update(id_number: "MD —123–456")
+        expect(state_id.id_number).to eq("MD 123456")
+      end
+    end
```
arinchoi03, line 53: is there a test example of a regular dash (not em-dash or en-dash) that _will_ be accepted?

arinchoi03 (approved):
looks great! just a few non-blocking comments

app/models/state_id.rb
DrewProebstel: Thats the standard dash which we want to allow.

spec/models/state_id_spec.rb
DrewProebstel: good call I did it

app/models/state_id.rb
DrewProebstel: totally

arinchoi03 (approved):

PR #5570, opened 2025-02-12
[FYST-1704] Income/ donation forms: Inaccurate date selection

squanto:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1704

## What was done?
Update date accessible validations, so that all incoming parts of a date (day, month, and year) are validated together instead of piece by piece. This fixes cases where users enter invalid dates, like Feb 31st.

### DIFF
--- a/app/models/az321_contribution.rb
+++ b/app/models/az321_contribution.rb
@@ -32,6 +32,7 @@ class Az321Contribution < ApplicationRecord
   validates :amount, presence: true, numericality: { greater_than: 0 }
   validates :date_of_contribution,
     inclusion: {
-      in: TAX_YEAR.beginning_of_year..TAX_YEAR.end_of_year
+      in: TAX_YEAR.beginning_of_year..TAX_YEAR.end_of_year,
+      message: ->(_object, _data) {I18n.t('activerecord.errors.concerns.date_accessible.inclusion')}
     }
 end
--- a/app/models/az321_contribution.rb
+++ b/app/models/az321_contribution.rb
@@ -33,6 +33,5 @@ class Az321Contribution < ApplicationRecord
   validates :date_of_contribution,
     inclusion: {
       in: TAX_YEAR.beginning_of_year..TAX_YEAR.end_of_year
-    },
-    presence: true
+    }
 end
--- a/app/models/concerns/date_accessible.rb
+++ b/app/models/concerns/date_accessible.rb
@@ -4,7 +4,6 @@ module DateAccessible
   TAX_YEAR = Date.new(Rails.configuration.statefile_current_tax_year)
 
   included do
-    private
 
     # Calls `date_reader` and `date_writer` on specified date properties to set
     # getters and setters on the specified date properties. For use with
@@ -28,17 +27,7 @@ module DateAccessible
       properties = [properties] unless properties.is_a?(Enumerable)
 
       properties.each do |property|
-        self.define_method("#{property}_month") do
-          send(property)&.month
-        end
-
-        self.define_method("#{property}_year") do
-          send(property)&.year
-        end
-
-        self.define_method("#{property}_day") do
-          send(property)&.day
-        end
+        attr_reader :"#{property}_month", :"#{property}_year", :"#{property}_day"
       end
     end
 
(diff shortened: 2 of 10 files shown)


app/models/az321_contribution.rb
```diff
@@ -33,6 +33,5 @@ class Az321Contribution < ApplicationRecord
   validates :date_of_contribution,
     inclusion: {
       in: TAX_YEAR.beginning_of_year..TAX_YEAR.end_of_year
-    },
-    presence: true
+    }
```
arinchoi03, line 36: why is the presence check removed?

app/models/concerns/date_accessible.rb
```diff
@@ -4,7 +4,6 @@ module DateAccessible
   TAX_YEAR = Date.new(Rails.configuration.statefile_current_tax_year)
 
   included do
-    private
```
arinchoi03, line 7: what's the significance of making these not private methods?

spec/models/concerns/date_accessible_spec.rb
```diff
@@ -1,162 +1,93 @@
...
+      subject.expiration_date_day = 1
+      subject.expiration_date_month = 2
+      subject.expiration_date_year = year
+      subject.valid?
+      expect(subject.expiration_date).to eq(Date.new(year, 2, 1))
+
+      subject.expiration_date_day = 21
+      subject.expiration_date_year = nil
+      subject.valid?
+      expect(subject.expiration_date).to eq(Date.new(year, 2, 1))
```
arinchoi03, line 49: hmm i don't understand this example -- the year is `nil` but the date is still set to `year` -- is it b/c `year` `nil` is invalid so therefore it was never written to the expiration date?

arinchoi03 (commented):
I think someone else Kevin/Tahsina who have worked on this should review this but I'm trying to understand the code!

mrotondo:
One blocking issue I noticed: This breaks the _year, _month, and _day attr_readers, which means that the forms no longer show the previously entered dates when rendering. To see this, go through a state_id page like `/primary-state-id`, enter valid issue & expiration dates, continue, then go back and note that the date selects are blank. Happy to pair on this - there's definitely some ruby wizardry at work, since I can't tell what the attr_readers are backed by but they seem to work _when they've been used to set the date parts on an in-memory object_, so clearly they're backed by something.

app/models/az321_contribution.rb
squanto: for this the inclusion validation serves as a presence check as well, as `nil` values fall outside of the range. It made it so I didn't have to duplicate the error message

app/models/concerns/date_accessible.rb
```diff
@@ -52,42 +41,34 @@ def self.date_writer(*properties)
...
+          if send("#{property}_year").present? && send("#{property}_month").present? && send("#{property}_day").present?
+            send(
+              "#{property}=",
+              Date.new(
+                  send("#{property}_year").to_i,
+                  send("#{property}_month").to_i,
+                  send("#{property}_day").to_i,
+              )
+            )
+          end
```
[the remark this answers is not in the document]
squanto: if one of the values is missing, then the date won't get set / updated, and it'll fail the inclusion check in it's model

app/models/concerns/date_accessible.rb
squanto: The methods from the module are class methods, so they can't be private, and I believe this line has no effect

spec/models/concerns/date_accessible_spec.rb
squanto: that's right, the intent is to say that b/c ``year` `nil` is invalid, it won't write to the date. I'll split this out for better clarity on what it's testing

squanto:
With the update to the attr accessors, it appears we're now able to view incorrect fields when rerendering @mrotondo

app/models/concerns/date_accessible.rb
```diff
@@ -29,15 +28,13 @@ def self.date_reader(*properties)
...
         end
-
         self.define_method("#{property}_year") do
-          send(property)&.year
+          self.instance_variable_get("@#{property}_year") || send(property)&.year
         end
-
         self.define_method("#{property}_day") do
-          send(property)&.day
+          self.instance_variable_get("@#{property}_day") || send(property)&.day
```
powersurge360, line 37: This will sidestep any other setters, as opposed to calling `super`. Intended?

app/models/concerns/date_accessible.rb
```diff
@@ -52,42 +49,31 @@ def self.date_writer(*properties)
...
-      self.send(
-          "#{date_property}=",
-          existing_date.change(**args.transform_values(&:to_i))
-      )
-    rescue Date::Error
-      nil
+          define_method("#{property}_date_valid") do
+            date = send(property)
+            if date.present? && !Date.valid_date?(date.year, date.month, date.day)
+              errors.add(property, :invalid_date, message: "is not a valid calendar date")
```
powersurge360, line 72: If we're doing validation, should be translated.

powersurge360 (changes requested):
Looks great. I left two notes but the only one that I feel is a blocker is translating a validation string in the new default validator for the date.

app/models/concerns/date_accessible.rb
squanto: Yes that's the intent - for this setter to default to the instance variable, with a backup of parsing the persisted value

app/models/concerns/date_accessible.rb
squanto: absolutely, thanks for the catch

powersurge360 (approved):
Issues addressed. Approved.

PR #5581, opened 2025-02-13
GYR1-673 Updates the form to not accept HEIC directly.

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-673

## What was done?
- We were having failed HEIC image conversions happening due to iOS 18 having a more advanced encoding. Options for updating the library were unsatisfactory (see the engineering notes on the ticket for more information)
- Instead of updating the library, we are now relying on Safari's auto-re-encoding of HEIC images to handle it. On Safari on desktop, and all browsers on iOS mobile (since they are all safari under the hood), we should now receive a JPG instead. iMessage also performs this encoding over MMS

## How to test?
- Use an HEIC image to upload to any of the document forms. It must be taken with iOS 18. A test image will be provided on the ticket.

- Before
- After
image

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -228,7 +228,10 @@ describe Document do
   describe "after_create" do
     include ActiveJob::TestHelper
 
-    context "when the file extension is .heic" do
+    # iOS 18 uses a different codec. We are relying now on safari and MMS to
+    # handle this before it gets to us. For more information, see ticket
+    # GYR1-673
+    xcontext "when the file extension is .heic" do
       before do
         allow(HeicToJpgJob).to receive(:perform_later).and_call_original
       end
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -285,6 +285,9 @@ describe Document do
     end
   end
 
+  # iOS 18 uses a different codec. We are relying now on safari and MMS to
+  # handle this before it gets to us. For more information, see ticket
+  # GYR1-673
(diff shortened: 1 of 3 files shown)


spec/models/document_spec.rb
```diff
@@ -282,7 +285,7 @@
     end
   end
 
-  describe "#convert_heic_upload_to_jpg!" do
+  xdescribe "#convert_heic_upload_to_jpg!" do
```
mpidcock, line 288: minor - would it be ok to delete this? or explain what to do for this test to be brought back?

mpidcock (approved):
nice work!

spec/models/document_spec.rb
powersurge360: I thought about it but the core thing preventing it from working is a newer imagemagick. Because that is a fairly low code change to get it working again I was reluctant to remove it.

I am feel pretty that delegating to just requiring JPG and letting apple devices on safari re-encode on the fly is the most correct way to do things but it isn't necessarily the most inclusive way and we do leave out people uploading HEIC on Chrome and Firefox on desktop (but not on mobile on apple devices).

It may be that we want to flip it on once imagemagick is easier to get updated. Maybe we could split the difference and I'll set a reminder to remove it in 6 months if it hasn't been touched in the mean time?

mpidcock: i think a note like "this is disabled until we can update imagemagik to support HEIC" would be helpful, but also setting a reminder is a good idea!

PR #5584, opened 2025-02-14
FYST-1792 filing status is sometimes nil part 2

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1792

## What was done?
- Added a before action to catch people who try to load a page after data transfer without any df data (this can happen if they start intake over in a new tab and then refresh a page they used to have access to)
- Check depends on a new trait added to navigation section
- Skipped the before action on post data transfer because it's the only page in its section that already handles the no-df-data case
- We skipped adding this check for NJ because the way they have navigation sections set up is substantially different and seemed like a lot more work
## How to test?
- Added a shared example spec that checks for the redirect. Didn't add it to every controller spec, just a few that represent the different sections of intake and are shared between states (and looped over each state)


app/lib/navigation/state_file_az_question_navigation.rb
```diff
@@ -44,11 +44,11 @@ class StateFileAzQuestionNavigation < Navigation::StateFileBaseQuestionNavigatio
         Navigation::NavigationStep.new(StateFile::Questions::TaxesOwedController),
         Navigation::NavigationStep.new(StateFile::Questions::TaxRefundController),
         Navigation::NavigationStep.new(StateFile::Questions::EsignDeclarationController), # creates EfileSubmission and transitions to preparing
-      ]),
+      ], false, true),
```
arinchoi03, line 47: just highlighting something that i mentioned in the story that `StateFile::Questions::PostDataTransferController` can be currently in a state where df data doesnt exist -- it redirects the user to the data-import failed page if it failed

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -30,6 +30,16 @@ def redirect_if_no_intake
         end
       end
 
+      def redirect_if_df_data_required
+        return if current_state_code == "nj"
```
arinchoi03, line 34: hm, i get that we're excluding NJ but if so, I highly recommend asking NJ to fix this on their end, or maybe pairing with them to get this fixed together?

app/lib/navigation/state_file_nc_question_navigation.rb
```diff
@@ -44,11 +44,11 @@ class StateFileNcQuestionNavigation < Navigation::StateFileBaseQuestionNavigatio
         Navigation::NavigationStep.new(StateFile::Questions::NcTaxesOwedController),
         Navigation::NavigationStep.new(StateFile::Questions::NcTaxRefundController),
         Navigation::NavigationStep.new(StateFile::Questions::EsignDeclarationController), # creates EfileSubmission and transitions to preparing
-      ]),
+      ], true, true),
       Navigation::NavigationSection.new("state_file.navigation.section_6", [
         Navigation::NavigationStep.new(StateFile::Questions::SubmissionConfirmationController),
         Navigation::NavigationStep.new(StateFile::Questions::ReturnStatusController),
-      ]),
+      ], true, true),
```
arinchoi03, line 51: tru tru 🫶

spec/factories/state_file_id_intakes.rb
```diff
@@ -93,6 +93,7 @@
     raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.old_xml_sample }
     raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.old_json_sample }
     df_data_import_succeeded_at { DateTime.now }
+    state_file_analytics { StateFileAnalytics.create }
```
arinchoi03, line 96: why was this part of the change?

spec/controllers/state_file/questions/unemployment_controller_spec.rb
```diff
@@ -23,7 +28,33 @@
...
+      end
+
+      it "redirects to login on edit" do
+        get :new
+
+        expect(response).to redirect_to StateFile::StateFilePagesController.to_path_helper(action: :login_options)
+      end
+    end
+  end
+
```
arinchoi03, line 44: [q] is the reason we don't include `get :new` or `get :index` into the shared example is that some of these controllers just do not have these actions? 

this might be a bit going overboard but when if we just pass in the action to the `it_behaves_like` as an argument?

`it_behaves_like :df_data_required, true, :az, :action, :record_type`? Not super passionate about it but just a suggestion to make it easier to add to future controllers (altho idk if we'll ever add more)

arinchoi03 (approved):
just a few comments but looks great to me! 👏

spec/factories/state_file_id_intakes.rb
jenny-heath: this had to be present on intake in order for the test on `InitiateDataTransferController` to pass. we went with updating the factories because it seems realistic for this record to exist (since it's created at the same time as intake)

spec/controllers/state_file/questions/unemployment_controller_spec.rb
jenny-heath: answer to the initial question is yes, most controllers don't have anything beyond edit and update.

i thought briefly about passing in the action but we'd then have to have at least 2 `it_behaves_like` lines per controller spec (also the shared example would have to make some distinction between `get` and `post` which is a bit more complicated). i'm not sure which approach i prefer tbh so ended up leaving it as is.

arinchoi03: np! ty for listing out the possibilities. I'm good with current approach, especially as I don't think we'll be adding more controllers (at least at this time)

PR #5585, opened 2025-02-14
Nj 81 excludable retirement line 20b

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/81

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

## What was done?
- Add line 20b to XML and PDF

## How to test?
- Can use Zeus two 1099r, Bog turtle, or Eastern goldfinch to test
- Don't forget to enable retirement UI toggle

## Screenshots (for visual changes)
image
image

PR #5591, opened 2025-02-14
Nj 287 retirement other exclusion

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/287

## What was done?
Added the following lines for retirement income calculation

28A - always static 0 for now
28B - see ticket for details
28C - sum 28A and 28B
29 - update logic to subtract 28C

## How to test?
- Manual testing in Heroku
  - Springsteen MFJ, edit W2 so StateLocalWages is <= 3000. 1099-INT belongs to spouse, who is <62, so expect that income to be filtered out, while W2 wages are correctly included in line 28b.

- Before
- After
XML contains fields
image

PDF contains fields
image

WITHOUT modifying W2 XML to 3000, Springsteen MFJ is ineligible for the exclusion and the fields are not shown in the final return XML.

image

AND the fields are blank in the PDF

image

### DIFF
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -253,28 +253,6 @@ module Efile
         line_or_zero(:NJ1040_LINE_12_COUNT) * 1_000
       end
 
-      def calculate_line_15
-        @intake.state_file_w2s.sum do |w2|
-          w2.state_wages_amount.to_i
-        end
-      end
-
-      def calculate_line_20a
-        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
-          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
-        end
-
-        applicable_1099rs.sum(&:taxable_amount).round
-      end
-
-      def calculate_line_27
-        line_or_zero(:NJ1040_LINE_15) + line_or_zero(:NJ1040_LINE_16A) + line_or_zero(:NJ1040_LINE_20A)
-      end
-
-      def calculate_line_28a
-        0
-      end
-
       def line_53c_checkbox
         @intake.eligibility_all_members_health_insurance_yes?
       end
@@ -331,6 +309,12 @@ module Efile
           calculate_line_12
       end
 
+      def calculate_line_15
+        @intake.state_file_w2s.sum do |w2|
+          w2.state_wages_amount.to_i
+        end
+      end
+
       def calculate_line_16a
         @intake.direct_file_data.fed_taxable_income - interest_on_gov_bonds
       end
@@ -339,8 +323,24 @@ module Efile
         calculate_tax_exempt_interest_income if calculate_tax_exempt_interest_income.positive?
       end
 
+      def calculate_line_20a
+        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
+          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
+        end
+
+        applicable_1099rs.sum(&:taxable_amount).round
+      end
+
       def calculate_line_20b
         (non_military_1099rs.sum(&:gross_distribution_amount) - non_military_1099rs.sum(&:taxable_amount)).round
+      end
+
+      def calculate_line_27
+        line_or_zero(:NJ1040_LINE_15) + line_or_zero(:NJ1040_LINE_16A) + line_or_zero(:NJ1040_LINE_20A)
+      end
+
+      def calculate_line_28a
+        0
       end
 
       def calculate_line_28b
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -324,11 +324,7 @@ module Efile
       end
 
       def calculate_line_20a
-        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
-          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
-        end
-
-        applicable_1099rs.sum(&:taxable_amount).round
+        non_military_1099rs.sum(&:taxable_amount).round
       end
 
       def calculate_line_20b
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -21,6 +21,7 @@ module Efile
           intake: @intake,
           primary_or_spouse: :spouse
         )
+        @nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(@intake)
       end
 
       def calculate
@@ -343,19 +344,22 @@ module Efile
       end
 
       def calculate_line_28b
-        helper = Efile::Nj::NjRetirementIncomeHelper.new(@intake)
-        return 0 unless helper.eligible?
+        return 0 unless @nj_retirement_income_helper.retirement_exclusion_eligible?(
+          line_or_zero(:NJ1040_LINE_15), 
+          line_or_zero(:NJ1040_LINE_27), 
+          line_or_zero(:NJ1040_LINE_28A))
         
-        total_income = calculate_line_27
-        [helper.calculate_maximum_exclusion(total_income) - calculate_line_28a, helper.total_eligible_nonretirement_income].min
+        total_income = line_or_zero(:NJ1040_LINE_27)
+        [@nj_retirement_income_helper.calculate_maximum_exclusion(total_income) - line_or_zero(:NJ1040_LINE_28A),
+         @nj_retirement_income_helper.total_eligible_nonretirement_income].min
       end
 
       def calculate_line_28c
-        calculate_line_28a + calculate_line_28b
+        line_or_zero(:NJ1040_LINE_28A) + line_or_zero(:NJ1040_LINE_28B)
       end
 
       def calculate_line_29
-        line_or_zero(:NJ1040_LINE_27) - calculate_line_28c
+        line_or_zero(:NJ1040_LINE_27) - line_or_zero(:NJ1040_LINE_28C)
       end
 
       def calculate_line_31
--- a/app/lib/efile/nj/nj1040_calculator.rb
+++ b/app/lib/efile/nj/nj1040_calculator.rb
@@ -41,6 +41,9 @@ module Efile
         set_line(:NJ1040_LINE_20A, :calculate_line_20a)
         set_line(:NJ1040_LINE_20B, :calculate_line_20b)
         set_line(:NJ1040_LINE_27, :calculate_line_27)
+        set_line(:NJ1040_LINE_28A, :calculate_line_28a)
+        set_line(:NJ1040_LINE_28B, :calculate_line_28b)
+        set_line(:NJ1040_LINE_28C, :calculate_line_28c)
         set_line(:NJ1040_LINE_29, :calculate_line_29)
         set_line(:NJ1040_LINE_31, :calculate_line_31)
         set_line(:NJ1040_LINE_38, :calculate_line_38)
@@ -249,6 +252,28 @@ module Efile
         line_or_zero(:NJ1040_LINE_12_COUNT) * 1_000
       end
 
+      def calculate_line_15
+        @intake.state_file_w2s.sum do |w2|
+          w2.state_wages_amount.to_i
+        end
+      end
+
+      def calculate_line_20a
+        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
+          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
+        end
+
+        applicable_1099rs.sum(&:taxable_amount).round
+      end
+
+      def calculate_line_27
+        line_or_zero(:NJ1040_LINE_15) + line_or_zero(:NJ1040_LINE_16A) + line_or_zero(:NJ1040_LINE_20A)
+      end
+
+      def calculate_line_28a
+        0
+      end
+
       def line_53c_checkbox
         @intake.eligibility_all_members_health_insurance_yes?
       end
@@ -305,12 +330,6 @@ module Efile
           calculate_line_12
       end
 
-      def calculate_line_15
-        @intake.state_file_w2s.sum do |w2|
-          w2.state_wages_amount.to_i
-        end
-      end
-
       def calculate_line_16a
         @intake.direct_file_data.fed_taxable_income - interest_on_gov_bonds
       end
@@ -319,20 +338,24 @@ module Efile
         calculate_tax_exempt_interest_income if calculate_tax_exempt_interest_income.positive?
       end
 
-      def calculate_line_20a
-        non_military_1099rs.sum(&:taxable_amount).round
-      end
-
       def calculate_line_20b
         (non_military_1099rs.sum(&:gross_distribution_amount) - non_military_1099rs.sum(&:taxable_amount)).round
       end
 
-      def calculate_line_27
-        line_or_zero(:NJ1040_LINE_15) + line_or_zero(:NJ1040_LINE_16A) + line_or_zero(:NJ1040_LINE_20A)
+      def calculate_line_28b
+        helper = Efile::Nj::NjRetirementIncomeHelper.new(@intake)
+        return 0 unless helper.eligible?
+        
+        total_income = calculate_line_27
+        [helper.calculate_maximum_exclusion(total_income) - calculate_line_28a, helper.total_eligible_nonretirement_income].min
+      end
+
+      def calculate_line_28c
+        calculate_line_28a + calculate_line_28b
       end
 
       def calculate_line_29
-        line_or_zero(:NJ1040_LINE_27)
+        line_or_zero(:NJ1040_LINE_27) - calculate_line_28c
       end
 
       def calculate_line_31
--- a/app/lib/efile/nj/nj_retirement_income_helper.rb
+++ b/app/lib/efile/nj/nj_retirement_income_helper.rb
@@ -29,22 +29,23 @@ module Efile
       end
 
       def primary_62_and_older?
-        @intake.calculate_age(@intake.primary_birth_date, inclusive_of_jan_1: true) >= 62
+        @intake.calculate_age(@intake.primary_birth_date, inclusive_of_jan_1: false) >= 62
       end
 
       def spouse_62_and_older?
-        @intake.calculate_age(@intake.spouse_birth_date, inclusive_of_jan_1: true) >= 62 
+        @intake.calculate_age(@intake.spouse_birth_date, inclusive_of_jan_1: false) >= 62 
       end
 
-      def eligible?
+      def retirement_exclusion_eligible?(line_15, line_27, line_28a)
         if @intake.spouse_birth_date.present?
-          return false unless spouse_62_and_older? || primary_62_and_older?
+          spouse_or_primary_is_age_eligible = spouse_62_and_older? || primary_62_and_older?
+          return false unless spouse_or_primary_is_age_eligible
         elsif !primary_62_and_older?
           return false
         end
-        return false if @intake.calculator.lines[:NJ1040_LINE_15].value > 3_000
-        return false if @intake.calculator.lines[:NJ1040_LINE_27].value > 150_000
-        return false if @intake.calculator.lines[:NJ1040_LINE_28A].value > calculate_maximum_exclusion(@intake.calculator.lines[:NJ1040_LINE_27].value)
+        return false if line_15 > 3_000
+        return false if line_27 > 150_000
+        return false if line_28a > calculate_maximum_exclusion(line_27)
 
         true
       end
--- a/app/lib/efile/nj/nj_retirement_income_helper.rb
+++ b/app/lib/efile/nj/nj_retirement_income_helper.rb
@@ -19,10 +19,10 @@ module Efile
             total_income += w2.state_wages_amount.to_i
           end
         end
-        @intake.direct_file_json_data.interest_reports.each do |form1099|
-          sanitized_recipient_tin = form1099.recipient_tin.tr("-", "")
+        @intake.direct_file_json_data.interest_reports.each do |interest_report|
+          sanitized_recipient_tin = interest_report.recipient_tin.delete("-")
           if sanitized_recipient_tin.in?(eligible_ssn)
-            total_income += form1099.interest_on_government_bonds.to_i
+            total_income += (interest_report.amount_1099&.round || 0) + (interest_report.amount_no_1099&.round || 0)
           end
         end
         total_income
--- a/app/lib/efile/nj/nj_retirement_income_helper.rb
+++ b/app/lib/efile/nj/nj_retirement_income_helper.rb
@@ -1,0 +1,90 @@
+module Efile
+  module Nj
+    class NjRetirementIncomeHelper
+      def initialize(intake)
+        @intake = intake
+      end
+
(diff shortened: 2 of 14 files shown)


jachan:
Sending this back to draft as there are questions in the ticket about how 28b should be calculated.

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -137,6 +137,18 @@ def document
...
+                  end
+
+                  if calculated_fields.fetch(:NJ1040_LINE_28B).positive? && Flipper.enabled?(:show_retirement_ui)
+                    xml.OtherRetireIncomeExclus calculated_fields.fetch(:NJ1040_LINE_28B)
+                  end
+
+                  if calculated_fields.fetch(:NJ1040_LINE_28C).positive? && Flipper.enabled?(:show_retirement_ui)
+                    xml.TotalExclusionAmount calculated_fields.fetch(:NJ1040_LINE_28C)
+                  end
+
```
aloverso, line 151: [sand] Now that we're in the more-than-two case, I'd propose a refactoring here where we group all these fields into a single `if Flipper enabled` check (including lines 128-132) rather than redoing this conditional on each individual line

spec/factories/state_file_nj_intakes.rb
```diff
@@ -264,7 +264,7 @@
       raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_sinatra_qss') }
       filing_status { "qualifying_widow" }
     end
-
+
```
aloverso, line 267: [dust] whitespace added here should be removed

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -623,12 +623,58 @@ def over_65_birth_year
...
+      end
+    end
+
+    context 'when taxpayer is eligible' do
+      context 'when maximum_exclusion minus 28a is greater than total_eligible_nonretirement_income' do
+        it 'sets line 28b to total_eligible_nonretirement_income' do
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:eligible?).and_return true
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:total_eligible_nonretirement_income).and_return 1_000
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:calculate_maximum_exclusion).and_return 50_000
+          allow(instance).to receive(:calculate_line_28a).and_return 0
```
aloverso, line 643: A few small notes on this test:
- [the dustiest of dust] can we rearrange these three mock statements to be in the same order as the text? First maximum exclusion, then 28a, then total eligible. It makes it much easier to read and track what's going on against the behavior described
- [dust] Could we test an edge case regarding the "greater than"? ie 50,000 minus 1,000 is A LOT greater than, might be nicer to test 50,000 - 49,999 to really show the behavior.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -623,12 +623,58 @@ def over_65_birth_year
...
+          expect(instance.lines[:NJ1040_LINE_28B].value).to eq(1000)
+        end
+      end
+
+      context 'when maximum_exclusion minus 28a is less than total_eligible_nonretirement_income' do
+        it 'sets line 28b to maximum_exclusion minus 28a' do
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:eligible?).and_return true
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:total_eligible_nonretirement_income).and_return 100_000
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:calculate_maximum_exclusion).and_return 50_000
+          allow(instance).to receive(:calculate_line_28a).and_return 0
```
aloverso, line 654: [pebble] Right now, by mocking 28a to 0, we don't actually receive any confidence from this test that 28a is incorporated into the calculation at all - I think we could remove it from the code and the test would still pass. If we want to confirm that all parts are working as expected, something like max_exclusion=50,000, 28a=1,000, total_eligible=49,500. This would then ONLY pass if 28a is included in the calculation

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -623,12 +623,58 @@ def over_65_birth_year
...
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:eligible?).and_return true
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:total_eligible_nonretirement_income).and_return 1_000
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:calculate_maximum_exclusion).and_return 50_000
+          allow(instance).to receive(:calculate_line_28a).and_return 0
+          instance.calculate
+          expect(instance.lines[:NJ1040_LINE_28B].value).to eq(1000)
+        end
+      end
+
+      context 'when maximum_exclusion minus 28a is less than total_eligible_nonretirement_income' do
```
aloverso, line 649: [pebble] can we have a test for what happens when they are equal?

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -623,12 +623,58 @@ def over_65_birth_year
...
+  end
+
   describe 'line 29 - gross income' do
     let(:intake) { create(:state_file_nj_intake) }
 
-    it 'sets line 29 to the sum of all state wage amounts' do
+    it 'sets line 29 to the sum of all state wage amounts minus 28c' do
       allow(instance).to receive(:calculate_line_15).and_return 50000
       allow(instance).to receive(:calculate_line_16a).and_return 1_000
+      allow(instance).to receive(:calculate_line_28c).and_return 0
```
aloverso, line 677: [pebble] same comment as above - by mocking this to zero, I don't think we're gaining any confidence that this is actually being used in the calculation. I think this test would still pass even if we removed the changes from the calculate_line_29 method. Can we set it to a number and change the expected output to match?

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -1721,8 +1767,7 @@ def over_65_birth_year
         contribution_2 = instance.excess_ui_wf_swf_max - 2
 
         first_w2.update_attribute(:box14_ui_wf_swf, contribution_1)
-        second_w2.update_attribute(:box14_ui_wf_swf, contribution_2)
-        second_w2.update_attribute(:box14_ui_hc_wd, 1)
+        second_w2.update_attribute(:box14_ui_hc_wd, contribution_2)
```
aloverso, line 1770: [dust] is this an unrelated / mistake change?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -319,20 +338,24 @@ def calculate_line_16b
...
+        [helper.calculate_maximum_exclusion(total_income) - calculate_line_28a, helper.total_eligible_nonretirement_income].min
+      end
+
+      def calculate_line_28c
+        calculate_line_28a + calculate_line_28b
       end
 
       def calculate_line_29
-        line_or_zero(:NJ1040_LINE_27)
+        line_or_zero(:NJ1040_LINE_27) - calculate_line_28c
```
mmazanec22, line 358: [dust] Why not use the line_or_zero instead of calculate_line_28c for consistency and to avoid calculating things more than once?

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      def primary_62_and_older?
+        @intake.calculate_age(@intake.primary_birth_date, inclusive_of_jan_1: true) >= 62
+      end
+
+      def spouse_62_and_older?
+        @intake.calculate_age(@intake.spouse_birth_date, inclusive_of_jan_1: true) >= 62 
+      end
+
+      def eligible?
```
mmazanec22, line 39: [pebble] I think this method name would be clearer if it spelled out what a person is/not eligible for

spec/lib/pdf_filler/nj1040_pdf_spec.rb
```diff
@@ -1225,6 +1225,78 @@
...
+          expect(pdf_fields["199"]).to eq "2"
+          expect(pdf_fields["200"]).to eq "3"
+          # hundreds
+          expect(pdf_fields["undefined_85"]).to eq "4"
+          expect(pdf_fields["201"]).to eq "5"
+          expect(pdf_fields["202"]).to eq "6"
+          # decimals
+          expect(pdf_fields["undefined_86"]).to eq "0"
+          expect(pdf_fields["203"]).to eq "0"
+        end
```
aloverso, line 1296: [pebble] could we test the empty case as well for all of these? When the number is zero, all the pdf fields should be blank

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
```diff
@@ -137,6 +137,18 @@ def document
                     xml.TotalIncome calculated_fields.fetch(:NJ1040_LINE_27)
                   end
 
+                  if calculated_fields.fetch(:NJ1040_LINE_28A).positive? && Flipper.enabled?(:show_retirement_ui)
```
mmazanec22, line 140: [dust] In the spirit of data flowing from the calculator to the submission builder to the PDF, should the flipper check maybe be in the calculator instead?

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -497,6 +497,55 @@
...
+      
+      context "when filer is eligible" do
+        it 'sets OtherRetireIncomeExclus to the line 28b value' do
+          expected = 1_000
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_28b).and_return expected
+          expect(xml.at("OtherRetireIncomeExclus").text).to eq(expected.to_s)
+        end
+      end
+
+      context "when filer is ineligible" do
```
aloverso, line 527: [dust] Could we change the labels here to "when 28b has a value / does not have a value" or something similar? Nothing here is really dealing with the concept of eligibility, that's handled and tested elsewhere.

aloverso: Same for the other tests in this file above and below

spec/lib/submission_builder/ty2024/states/nj/documents/nj1040_spec.rb
```diff
@@ -497,6 +497,55 @@
...
+    describe "Other Retirement Income Exclusion - line 28a" do
+      before do
+        allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
+      end
+
+      context "when filer is eligible" do
+        it 'sets PensionExclusion to the line 28a value' do
+          expected = 5_000
+          allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_28a).and_return expected
+          expect(xml.at("PensionExclusion").text).to eq(expected.to_s)
```
aloverso, line 509: [pebble] Could we add a test for the 0 case where it does not get included in XML?

aloverso: Same with 28c

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
+require 'rails_helper'
+
+RSpec.describe Efile::Nj::NjRetirementIncomeHelper do
+  describe ".calculate_maximum_exclusion" do
+    [
+      { total_income: 99_000, expected: 75_000 },
```
aloverso, line 6: [pebble] for all these cases in this whole test file, could we test edge cases on both sides of each boundary? It'll be a lot more test cases to write but worth it rather than just having example cases from the middle of the range. Right now, if we mistakenly changed the magic numbers in our code, some of this could still pass.

I'm thinking:
- 100,000 (first column)
- 100,001 (second column edge 1)
- 125,000 (second column edge 2)
- 125,001 (third column edge 1)
- 150,000 (third column edge 2)
- 150,001 (beyond case)

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+          end
+          before do
+            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_15).and_return(test_case[:earned_income])
+            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_27).and_return(test_case[:total_income])
+            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_28a).and_return(test_case[:line_28a])
+            instance.calculate
+          end
+  
+          it "returns #{test_case[:expected]}" do
+            helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
```
aloverso, line 69: [dust] can we instantiate the helper in a before each (and still make use of lazy definition for the intake as-is) so that we don't have to duplicate this in every test?

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        end
+        it "returns #{test_case[:expected]}" do
+          helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
+          result = helper.calculate_maximum_exclusion(test_case[:total_income])
+          expect(result).to eq(test_case[:expected])
+        end
+      end
+    end
+
+    describe ".eligible?" do
```
aloverso, line 42: [dust] I think the indentation here is off, this seems like it's nested under the previous? Hard to tell in GH ui though

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      [
+        { traits: [:mfj_spouse_over_65], total_income: 99_000, earned_income: 3000, line_28a: 0, expected: true },
+        { traits: [:primary_over_65], total_income: 99_000, earned_income: 3000, line_28a: 0, expected: true },
+        { total_income: 99_000, earned_income: 3000, line_28a: 0, expected: false },
+        { traits: [:primary_over_65], total_income: 160_000, earned_income: 3000, line_28a: 0, expected: false },
+        { traits: [:primary_over_65], total_income: 99_000, earned_income: 4000, line_28a: 0, expected: false },
+        { traits: [:primary_over_65], total_income: 99_000, earned_income: 4000, line_28a: 100_000, expected: false },
```
aloverso, line 49: [pebble] Similar note about testing edge cases. Regarding age, I feel weird about using the over-65 trait when our logic here is checking the over-62 case. Could we add a new trait for over/under 62 so that we can verify that logic? If we changed the magic number in our code from 62 to 63, this test would still pass

aloverso: And similarly, we could gain more confidence by using 3,001 instead of 4,000 for the line-15 criteria, and $99,999, $150,000, and $150,001 for the total income criteria. This would provide a lot more confidence in the true logical behavior of the code, and also be a little more descriptive of the logic just from reading the cases

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        first_w2.update_attribute(:state_wages_amount, 1_000)
+        first_1099_int = intake.direct_file_json_data.interest_reports.first
+        first_1099_int.recipient_tin = primary_ssn.to_s
+        first_1099_int.interest_on_government_bonds = 2_000
+        second_1099_int = intake.direct_file_json_data.interest_reports.second
+        second_1099_int.recipient_tin = spouse_ssn.to_s
+        second_1099_int.interest_on_government_bonds = 4_000
+      end
+      context "when primary and spouse are both over 62" do
+        let(:intake) { create(:state_file_nj_intake, :df_data_one_dep, :primary_over_65, :mfj_spouse_over_65) }
```
aloverso, line 94: [pebble] same comment regarding over-65 vs over-62

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+
+        it 'sums both spouse and primary wages in total eligible income' do
+          setup_income_data(intake)
```
aloverso, line 97: [dust] you could do this duplicated line in a before-each for this context. In fact, I think you could replace this setup function and just make the whole block a before-each instead? It would still be able to reference intake through lazy loading, I believe

app/lib/submission_builder/ty2024/states/nj/documents/nj1040.rb
aloverso: [loosely held opinion] I prefer the check being in the XML as-is over the calculator. It feels like the calculator is more "pure" logic - ie, this is what the field calculates to given the input we have. Whereas the flipper check is a UI thing - "do we show this particular information to the user in the app" hence putting it in the XML builder

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -249,6 +252,28 @@ def calculate_line_12
...
+      end
+
+      def calculate_line_27
+        line_or_zero(:NJ1040_LINE_15) + line_or_zero(:NJ1040_LINE_16A) + line_or_zero(:NJ1040_LINE_20A)
+      end
+
+      def calculate_line_28a
+        0
+      end
+
```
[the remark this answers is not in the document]
jachan: Changed so that these methods could be mocked in the tests.

app/lib/efile/nj/nj_retirement_income_helper.rb
jachan: Renamed to retirement_exclusion_eligible?

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+        if @intake.spouse_birth_date.present?
+          return false unless spouse_62_and_older? || primary_62_and_older?
+        elsif !primary_62_and_older?
+          return false
+        end
+        return false if @intake.calculator.lines[:NJ1040_LINE_15].value > 3_000
+        return false if @intake.calculator.lines[:NJ1040_LINE_27].value > 150_000
+        return false if @intake.calculator.lines[:NJ1040_LINE_28A].value > calculate_maximum_exclusion(@intake.calculator.lines[:NJ1040_LINE_27].value)
```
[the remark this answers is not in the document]
jachan: Great point, updated.

app/lib/efile/nj/nj1040_calculator.rb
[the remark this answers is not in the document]
aloverso: Wait, really? I didn't realize they could only be mocked if public. I could have sworn we have other instances of mocking private methods in our tests?

spec/lib/efile/nj/nj1040_calculator_spec.rb
jachan: Yeah, I don't know how that got in there, but thanks for catching that before it got merged

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
jachan: This is a weird one, because of the each block on line 29 above, which needs to be nested under the describe block.

Rubocop then assumes that the test on line 42 must be aligned with the previous end on line 40. So, this indentation is necessary from a Rubocop point of view.

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -249,6 +253,28 @@ def calculate_line_12
...
+          w2.state_wages_amount.to_i
+        end
+      end
+
+      def calculate_line_20a
+        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
+          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
+        end
+
+        applicable_1099rs.sum(&:taxable_amount).round
```
mpidcock, line 267: [dust] there's a refactor opportunity here to combine this into one `.sum do` (you can search the codebase for a few multiline examples) but if you find this more readable I don't mind

(also i expect the `.present?` check could be removed if you use `&.`, as in `state_file_1099r.state_specific_followup&.income_source_none?` - but again, up to you)

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -0,0 +1,91 @@
...
+        total_income
+      end
+
+      def primary_62_and_older?
+        @intake.calculate_age(@intake.primary_birth_date, inclusive_of_jan_1: false) >= 62
+      end
+
+      def spouse_62_and_older?
+        @intake.calculate_age(@intake.spouse_birth_date, inclusive_of_jan_1: false) >= 62 
+      end
```
mpidcock, line 37: [dust] letting you know that `primary.birth_date` and `spouse.birth_date` are also valid, so you could have combined these, added a `:primary_or_spouse` parameter, and called `.send` - but i don't think it's worth updating it

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -623,14 +623,71 @@ def over_65_birth_year
...
+        expect(instance.lines[:NJ1040_LINE_28B].value).to eq(0)
+      end
+    end
+
+    context 'when taxpayer is eligible' do
+      context 'when maximum_exclusion minus 28a is greater than total_eligible_nonretirement_income' do
+        it 'sets line 28b to total_eligible_nonretirement_income' do
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:retirement_exclusion_eligible?).and_return true
+          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:calculate_maximum_exclusion).and_return 50_000
+          allow(instance).to receive(:calculate_line_28a).and_return 48_000
```
mpidcock, line 642: I agree with the comment above - you should be able to call `allow_any_instance_of(described_class).to receive(:calculate_line_28a)` without `calculate_line_28a` being public.

mpidcock: also, it looks like 28a returning 0 always (unless i'm missing something). is this the right way to test this scenario?

mpidcock (approved):
Ideally i'd like to see the public methods go back to being private, but none of my comments are blocking

spec/lib/efile/nj/nj1040_calculator_spec.rb
jachan: 28a is currently returning 0 but in the future will be filled in with different logic, so we're mocking it here to make sure that the arithmetic works out.

app/lib/efile/nj/nj1040_calculator.rb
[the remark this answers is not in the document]
jachan: I was wrong! Private methods can be mocked using `allow_any_instance_of()`. Fixed in latest revision.

app/lib/efile/nj/nj1040_calculator.rb
jachan: Thanks for the push to investigate this! Realized I was using old code that got updated in #5585, updated this method.

PR #5593, opened 2025-02-14
[GYR1-678] Clients who log in to GYR get stuck in a loop answering demographic questions

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-678

## Is PM acceptance required?

## What was done?

Summary of problem to address: when a client logs back in later, they get stuck 
in a loop of having to fill out the demographics questions over and over (which 
are asked *after* the "Success!" page). They never are able to get to a screen 
that lets them upload docs (the whole point of logging back in later).

Initial review of the business logic:
- First, note that the 3 required docs appear to be Id, selfie, and secondary ID 
  (e.g., SSN card). I base this on the logic in the already-existing 
  `has_all_required_doc?` method that was in 
  `app/controllers/documents/ssn_itins_controller.rb` (and which I refactored 
  into a new lib).
- I spent some time understanding the GYR documents flow (within the overall 
  intake flow), both examining code and experimenting with the app running 
  locally, to get a sense of intake flow behavior when the client reaches the 
  "Success!" page with versus without having uploaded the required docs.
- Seems like the issue is that, if no required docs get uploaded, the state will 
  always be `intake_in_progress` and there's no logic in the demographics 
  screens etc to change that. Additonally, somehow flow logic will always 
  'bookmark' their intake as being at the start of the demographics questions, 
  no matter if they've already filled/skipped them or not.

Based on the above:
- I refactored some code into a new GyrDocuments lib.
- In `app/controllers/questions/successfully_submitted_controller.rb`, I added 
  logic to transition the intake to a new state (either `intake_ready` or 
  intake_needs_doc_help); this was basically the same logic inside 
  `app/controllers/documents/ssn_itins_controller.rb` but here I put it inside a 
  `before_action` method (whereas in the latter controller it happens in 
  `after_update_succes`).

It seemed like there is some nuance around how the transition logic works. The 
changes I made here seem like just enough to solve the problem without affecting 
anything else.

## How to test?

Using Flow Explorer, test with uploading docs:
- Go all the way through the flow, and upload a (random) file for all 3 of id, 
  selfie, and seconary id, up to the "Success!" page. Then over in the Hub, the 
  newly created client should have a status (upper right hand corner) of Ready.

Using Flow Explorer, test by skipping the required doc uploads:
- Create another client, this time skipping all 3 of those documents. When you 
  get to the "Success!" page, you can try either going on through the 
  demographics page, or not (either way, the next behaviors should work the 
  same; wouldn't hurt to try this both ways).
  - Important: Go to the Hub, confirm this client has a status of Needs Doc 
    Help.
  - Now, If you're running locally:
    - go to a terminal and run `bin/rails jobs:work`.
    - Also, in a Rails console, you might need to do the following (you'll have 
      to figure out the right ID):

            Intake.find(14).update!({primary_consented_to_service: "yes"})
        
  - Then go home screen, click "Already started". Enter in the email address.
  - If you're using a heroku/other instance, use whatever method is in place to 
    check emails and grab the 6-digit code.
  - If you're running locally, pop back over to your jobs:work tail, and scroll 
    up to find that 6-digit code in the output.
  - After inputting the 6-digit code, you should land in the portal screen. The 
    button should say "Add missing documents" and when you click that, it should 
    take you to the "Here's a list of your documents." page.
  - This last bit of behavior was already in place, but it's worth testing it 
    again to be safe: Go through the process here of uploading the 3 requried 
    docs. Then in the Hub, the client should now have a status of Intake Ready.

### DIFF
--- a/app/lib/gyr_documents.rb
+++ b/app/lib/gyr_documents.rb
@@ -1,14 +1,16 @@
 module GyrDocuments
-  def self.has_all_required_docs?(intake)
-    submitted_doc_types = intake.documents.pluck(:document_type)
+  def has_all_required_docs?(intake)
+    intake_doc_types = intake.documents.pluck(:document_type)
+    required_docs = []
+    required_docs << DocumentTypes::Selfie.key unless IdVerificationExperimentService.new(intake).skip_selfies?
+    DocumentTypes::IDENTITY_TYPES.map(&:key).intersect?(intake_doc_types) &&
+      DocumentTypes::SECONDARY_IDENTITY_TYPES.map(&:key).intersect?(intake_doc_types) &&
+      required_docs.all? {|key| intake_doc_types.include?(key) }
+  end
 
-    has_photo_id = DocumentTypes::IDENTITY_TYPES.map(&:key).
-      intersect?(submitted_doc_types)
-    has_2ndary_id = DocumentTypes::SECONDARY_IDENTITY_TYPES.map(&:key).
-      intersect?(submitted_doc_types)
-
-    has_photo_id && has_2ndary_id &&
-      (IdVerificationExperimentService.new(current_intake).skip_selfies? ||
-        submitted_doc_types.include?(DocumentTypes::Selfie.key))
+  def advance_to(intake, next_state)
+    intake.tax_returns.each do |tax_return|
+      tax_return.transition_to(next_state) if tax_return.current_state.to_sym != next_state
+    end
   end
 end
--- a/app/lib/gyr_documents.rb
+++ b/app/lib/gyr_documents.rb
@@ -8,7 +8,7 @@ module GyrDocuments
       required_docs.all? {|key| intake_doc_types.include?(key) }
   end
 
-  def advance_to(intake, next_state)
+  def intake_transition_to(intake, next_state)
     intake.tax_returns.each do |tax_return|
       tax_return.transition_to(next_state) if tax_return.current_state.to_sym != next_state
     end
--- a/app/lib/gyr_documents.rb
+++ b/app/lib/gyr_documents.rb
@@ -1,0 +1,14 @@
+module GyrDocuments
+  def self.has_all_required_docs?(intake)
+    submitted_doc_types = intake.documents.pluck(:document_type)
+
+    has_photo_id = DocumentTypes::IDENTITY_TYPES.map(&:key).
+      intersect?(submitted_doc_types)
+    has_2ndary_id = DocumentTypes::SECONDARY_IDENTITY_TYPES.map(&:key).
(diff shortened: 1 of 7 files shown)


app/lib/gyr_documents.rb
```diff
@@ -0,0 +1,16 @@
...
+    required_docs = []
+    required_docs << DocumentTypes::Selfie.key unless IdVerificationExperimentService.new(intake).skip_selfies?
+    DocumentTypes::IDENTITY_TYPES.map(&:key).intersect?(intake_doc_types) &&
+      DocumentTypes::SECONDARY_IDENTITY_TYPES.map(&:key).intersect?(intake_doc_types) &&
+      required_docs.all? {|key| intake_doc_types.include?(key) }
+  end
+
+  def advance_to(intake, next_state)
+    intake.tax_returns.each do |tax_return|
+      tax_return.transition_to(next_state) if tax_return.current_state.to_sym != next_state
```
powersurge360, line 13: Why not `tax_return.advance_to`?

spec/features/web_intake/new_single_filer_spec.rb
```diff
@@ -457,6 +460,75 @@ def intake_after_triage_up_to_documents(intake)
       expect(intake.reload.current_step).to end_with("/questions/demographic-primary-race")
       expect(page).to have_selector("h1", text: I18n.t("portal.client_logins.new.title"))
     end
+
+    scenario "new client filing single without dependents AND without uploading *required* docs" do
+      intake = intake_up_to_documents
+
+      # IRS guidance
+      expect(page).to have_selector("h1", text: "First, we need to confirm your basic information.")
+      click_on "Continue"
```
powersurge360, line 469: Use i18n strings if possible to avoid copy changes breaking tests

powersurge360 (changes requested):
Code looks good. Had a minor question about transitioning and suggested moving to i18n strings rather than hard coded ones. I know we waffle about that in the code base but my understanding is the team is moving to using i18n exclusively after some copy-related headaches at launch.

app/lib/gyr_documents.rb
spompea-cfa: @powersurge360 and i tupled about this

spec/features/web_intake/new_single_filer_spec.rb
spompea-cfa: thanks for bringing this up. i've updated the spec test.

spompea-cfa:
@powersurge360, take another quick look when you can.

powersurge360 (approved):
Changes look good! Thanks for talking through it with me!

PR #5601, opened 2025-02-19
[FYST-1770] Update 1099G & income review behavior

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1770

## What was done?
- Adding a "final" vs "initial" (already existed as just "income_review") income review 
  - **allows users to move on to different `next_path` depending on whether 1099G controller was shown or not** 
  - no device info collection
  - final controller does not have a Form, b/c it's unnecessary
- We could've done this a different way -- for example, just keep one controller & pass in a param to make sure we were navigating to the correct `next_path`. However, this requires that we hard-code `next_path` for every state, which currently is all different - if we can reliably expect `retirement_subtraction` controllers (state-specific) we could probably use this approach; i'm not sure if we want to be tied to that approach.
## How to test?
  - Test data used
- Risk Assessment
- Before
- After


app/controllers/state_file/questions/final_income_review_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      include ReturnToReviewConcern
+      before_action :set_sorted_vars
+
+      def self.show?(intake)
+        intake.state_file_w2s.present? ||
+          intake.direct_file_data.fed_unemployment.positive? ||
+          intake.state_file1099_rs.present? ||
+          intake.direct_file_json_data.interest_reports.count.positive? ||
+          intake.direct_file_data.fed_ssb.positive? || intake.direct_file_data.fed_taxable_ssb.positive?
+      end
```
arinchoi03, line 13: should pull out shared methods into a shared controller that these inherit from with different `update` methods

app/controllers/state_file/questions/unemployment_controller.rb
```diff
@@ -3,98 +3,76 @@ module Questions
...
-        end
-
-        if @state_file1099_g.valid?
-          @state_file1099_g.save
-          redirect_with_review_param(:index)
-        else
-          render :new
-        end
-      end
-
```
arinchoi03, line 67: oops, broke some tests b/c i still point to `create` action in some places but it's been replaced

arinchoi03:
Need help on: 
general thoughts on the approach (two "income review" controllers vs one) -- should i do any of this differently? 😢

app/controllers/state_file/questions/retirement_income_controller.rb
```diff
@@ -31,6 +31,14 @@ def state_file1099_r_params
           :state_distribution_amount
         )
       end
+
+      def source_income_review_path
+        if params[:from_final_income_review] == "y" || params[:return_to_review] == "y"
```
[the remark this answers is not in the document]
embarnard: @arinchoi03 yes that makes sense. Also I think you are doing this but just to make sure all the states have initial income review but some of them have initial and final income review. So because device id collection for the initial state file device info object is happening on initial we are still always guaranteed to have collected it right?

mpidcock:
Oh goodness this is quite a lot. Two big thoughts here:
1. Maybe we could adjust 1099G form to be skipped if either 1099gs are present or there's no box 11 data. Then the continur button will just go to the next page in the flow and we don't need a separate page
2. Zooming out a bunch.... we get all the 1099G data we need from the json now, so at some point we will be importing them. It was deemed too much effort, but since this is growing in complexity, I would rather work on that than put in the effort to improve a flow that is going away. I'll bring this up with the PMs.

arinchoi03:
@mpidcock ooh yes, i think i for some reason missed the `had_box_11` aspect of it -- if it's answered (not `unfilled`) it means that page had been visited...just writing out the logic for myself:
income review -> unemployment controller (1099G edit) -> always go back to income_review controller -> next_path (still goes to 1099G edit but skip b/c `!had_box_11_unfilled?` which lands us back to whatever comes next. This is a lot simpler than what I had in mind, i for some reason thought that flag was not available 🙃 

Lmk how the convo with the PMs go! It should be a lot easier/simpler (ill probably start from scratch) but yea if we're going to get the data from DF I'd be interested to think through what kind of work we need to do for 1099G implementation.

arinchoi03:
Decided that we still want to pursue the story, will start from scratch without this double-income-review nonsense now that I realized i don't need it

PR #5604, opened 2025-02-19
NJ bugfix - check for nil state wages

mluedke2:
## Link to pivotal/JIRA issue

bugfix for https://codeforamerica.sentry.io/share/issue/a4cc8ba6916f4d0ba99b60c4a47ce20c/

## What was done?
- Test that considers the `nil` case
## How to test?
- Even if you edit the XML (either through UI or in code) to have an empty state wages, it still doesn't trigger the bug caught by Sentry. So I'm actually not sure how to reproduce. But since it is a clear fix I think the automated test covers it.


app/models/state_file_nj_intake.rb
```diff
@@ -213,7 +213,7 @@ def medical_expenses_threshold
 
   def validate_state_specific_w2_requirements(w2)
     super
-    if w2.wages.positive? && !w2.state_wages_amount.positive?
+    if w2.wages.positive? && (w2.state_wages_amount.nil? || w2.state_wages_amount <= 0)
```
aloverso, line 216: [dust] I was thinking about the switch from `!positive` to `<= 0`. I think it's less Ruby-ish but more clear to read conditionally - I'm in favor of this change

jachan (approved):

aloverso (approved):

mrotondo (approved):

PR #5607, opened 2025-02-19
FYST-1820 Remove duplicate w2s and dependents after importing

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1820

## What was done?
   - We're starting with a quick solution to prevent filers from submitting broken returns
   - This strategy for de-duplicating w2s would not work for 1099Rs, since we do not store and index for 1099Rs or anything thing else that is easy to detect a duplicate with
- Alternatives considered
   - Considered putting a database constraint on the w2 index column
      - one job would succeed, but the other would fail, and user should have no impact
      - still would want to cleanup as a backup
   - Considered trying to detect and suppress/delay/ignore duplicate api calls on the same intake
      - we are unsure of the approach and this is urgent
## How to test?
  - Added unit tests
- Risk Assessment
  - Doing just the cleanup is low risk
  - If the new code (including `destroy!` action) fails, user will see the page error, the job will fail & user will see data transfer failed page
  - Reproduction of the issue can't be done, only regression testing that this doesn't break data import

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,4 +78,3 @@ jmeter_test/results/*
 jmeter_test/jmeter.log
 
 .env
-
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,5 @@ jmeter_test/results/*
 jmeter_test/jmeter.log
 
 .env
+
+/config/credentials/test.key
--- a/.gitignore
+++ b/.gitignore
@@ -79,4 +79,3 @@ jmeter_test/jmeter.log
 
 .env
 
-/config/credentials/test.key
--- a/app/jobs/state_file/import_from_direct_file_job.rb
+++ b/app/jobs/state_file/import_from_direct_file_job.rb
@@ -35,7 +35,6 @@ module StateFile
         intake.synchronize_filers_to_database
 
         intake.update(df_data_import_succeeded_at: DateTime.now)
-
       rescue => err
         Rails.logger.error(err)
         intake.df_data_import_errors << DfDataImportError.new(message: err.to_s)
@@ -60,7 +59,7 @@ module StateFile
       collection.each do |record|
         value = record.send(identifying_attribute_name)
         if values.include?(value)
-          Rails.logger.info("ImportFromDirectFileJob removing duplicates #{record.class&.name} for #{intake&.state_code} #{intake&.id}")
+          Rails.logger.info("ImportFromDirectFileJob removing duplicate #{record.class&.name} for #{intake&.state_code} #{intake&.id}")
           record.destroy!
         else
           values.push(value)
--- a/app/jobs/state_file/import_from_direct_file_job.rb
+++ b/app/jobs/state_file/import_from_direct_file_job.rb
@@ -35,6 +35,11 @@ module StateFile
         intake.synchronize_filers_to_database
 
         intake.update(df_data_import_succeeded_at: DateTime.now)
+
+        # removing duplicate associations here because sometimes we create duplicate records during data import
+        # future work will prevent this issue from happening and this can be removed
+        remove_duplicate_w2s(intake)
+        remove_duplicate_dependents(intake)
       rescue => err
         Rails.logger.error(err)
         intake.df_data_import_errors << DfDataImportError.new(message: err.to_s)
@@ -46,5 +51,32 @@ module StateFile
     def priority
       PRIORITY_LOW
     end
+
+    private
+
+    def remove_duplicate_w2s(intake)
+      indices = []
+      intake.state_file_w2s.each do |w2|
+        current_index = w2.w2_index
+        if indices.include?(current_index)
+          w2.destroy!
+        else
+          indices.push(current_index)
+        end
+      end
+    end
+
+    def remove_duplicate_dependents(intake)
+      ssns = []
+      intake.dependents.each do |dependent|
+        ssn = dependent.ssn
+        if ssns.include?(ssn)
+          dependent.destroy!
+        else
+          ssns.push(ssn)
+        end
+      end
+    end
+
   end
 end
--- a/app/jobs/state_file/import_from_direct_file_job.rb
+++ b/app/jobs/state_file/import_from_direct_file_job.rb
@@ -55,29 +55,26 @@ module StateFile
 
     private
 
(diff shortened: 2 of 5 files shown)


mrotondo (commented):
The methods look good, but I think the method calls need to be at the bottom of the Import Job to make sure that they run after the second job finishes, since we'll proceed to this page as soon as the first one completes, and the duplicates might not have been created yet.

arinchoi03:
A reminder to self maybe we should log when we have to run these clean ups to keep track of the occurrence of these duplicate jobs?

arinchoi03:
> The methods look good, but I think the method calls need to be at the bottom of the Import Job to make sure that they run after the second job finishes, since we'll proceed to this page as soon as the first one completes, and the duplicates might not have been created yet.

oh b/c when the second job completes, we've already hit this `app/controllers/state_file/questions/post_data_transfer_controller.rb` 🤦 ty for the catch!

.gitignore
```diff
@@ -78,3 +78,5 @@ jmeter_test/results/*
 jmeter_test/jmeter.log
 
 .env
+
+/config/credentials/test.key
```
arinchoi03, line 82: why was this added, i guess fine?

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -35,6 +35,11 @@ def perform(authorization_code:, intake:)
         intake.synchronize_filers_to_database
 
         intake.update(df_data_import_succeeded_at: DateTime.now)
+
+        # removing duplicate associations here because sometimes we create duplicate records during data import
+        # future work will prevent this issue from happening and this can be removed
+        remove_duplicate_w2s(intake)
+        remove_duplicate_dependents(intake)
```
arinchoi03, line 42: I think i want to move these out of the `begin/rescue` block, like if the `destroy!` fails, idk if we want to make `df_data_import_errors` objects if we already set `df_data_import_succeeded_at` to a value.

.gitignore
mpidcock: oh, i thought i reverted this. we already have `/config/credentials/*.key` so i don't know why rubymine added it. i can remove.

mpidcock:
```suggestion
```

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -35,16 +35,45 @@ def perform(authorization_code:, intake:)
...
+
       rescue => err
         Rails.logger.error(err)
         intake.df_data_import_errors << DfDataImportError.new(message: err.to_s)
       end
 
+      # removing duplicate associations here because sometimes we create duplicate records during data import
+      # future work will prevent this issue from happening and this can be removed
+      remove_duplicate_w2s(intake)
+      remove_duplicate_dependents(intake)
```
arinchoi03, line 47: Should I wrap this in its own begin/rescue block so that if something raises in these methods, it blows up?

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -40,11 +40,39 @@ def perform(authorization_code:, intake:)
...
+      collection.each do |record|
+        value = record.send(identifying_attribute_name)
+        if values.include?(value)
+          Rails.logger.info("ImportFromDirectFileJob removing duplicate #{record.class&.name} for #{intake&.state_code} #{intake&.id}")
+          record.destroy!
+        else
+          values.push(value)
+        end
+      end
+    end
```
arinchoi03, line 68: i just did the refactor, not sure if this makes it more confusing 🤷

app/jobs/state_file/import_from_direct_file_job.rb
```diff
@@ -40,11 +40,39 @@ def perform(authorization_code:, intake:)
...
       DfDataTransferJobChannel.broadcast_job_complete(intake)
     end
 
     def priority
       PRIORITY_LOW
     end
+
+    private
+
+    def remove_duplicates(intake, collection, identifying_attribute_name)
```
arinchoi03, line 57: i did not want to send in the intake, but both `StateFileDependent` and `StateFileW2` have different names for the association, so I just pass it in

arinchoi03 (approved):
looks good! 

I just have one remaining q for you/us?:
https://github.com/codeforamerica/vita-min/pull/5607/files#r1964497075

app/jobs/state_file/import_from_direct_file_job.rb
mpidcock: I'm wondering if it actually makes sense to put this above the `intake.update(df_data_import_succeeded_at: DateTime.now)` after all. if the delete fails then the intake still has duplicates and we will need to intervene and the client should not continue the flow. Does that make sense to you too?

arinchoi03: yeah i think that makes sense! i can put it back

mpidcock:
To reproduce this issue locally, we needed to get multiple import jobs running at once on different threads. 
1. Update `Procfile.dev` to add more job workers (duplicate the existing worker a few times)
  ```
  worker: rails jobs:work
  worker: rails jobs:work
  worker: rails jobs:work
  worker: rails jobs:work
  ```
2. in one console run `foreman start`
3. in another console, run this to make a new intake and run the job multiple times. In this example, we are creating an NC intake and importing the "nc_nala_hoh" persona to it 20 times:
```
intake = StateFileNcIntake.create
20.times { StateFile::ImportFromDirectFileJob.perform_later(authorization_code: "nc_nala_hoh", intake: intake) }
```
4. Check the foreman logs to see the jobs complete. You should see multiple `RUNNING` and `COMPLETED` logs happening asynchronously. 
   * If every job is completing before another starts, make sure your `Procfile.dev` is properly updated to have multiple workers (and you started foreman since making the change)

Using this method to recreate the bug, we found that every job that started before any completed would create new `state_file_w2s`:
Screenshot 2025-02-21 at 3 03 12 PM
The number of workers and number of times running the job can be tweaked depending on how rare the race condition is. In this case the numbers are fairly low to reliably reproduce.

✅ I ran the same test on this branch and never saw the duplication, even when increasing the workers times running the job.

PR #5608, opened 2025-02-19
Nj 290 line 55 includes 1099-R

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/290

## What was done?
- Tests to check for W2s and 1099-Rs with taxes withheld
- There will eventually be more types of forms to be added in Line 55, but for now we only support these
- I also renamed `none` to `other` in the `income_source` enum of `state_file_nj1099_r_followup`, which feels more accurate

## How to test?
- Use `bog turtle`, which has 2000 withheld from W2 and 1000 withheld from 1099-R
- You'll have to adjust the UI Box 14 amount under the limit after you import
- It doesn't matter if Flipper is on for `show_retirement_ui` or not, or which income source you choose for the 1099-R. All withholdings count!
- XML and PDF show 3000 total withheld

image

image

### DIFF
--- a/spec/lib/efile/nj/nj1040_calculator_spec.rb
+++ b/spec/lib/efile/nj/nj1040_calculator_spec.rb
@@ -1607,8 +1607,8 @@ describe Efile::Nj::Nj1040Calculator do
 
     context 'when has only 1099-Rs' do
       let(:intake) { create(:state_file_nj_intake, :df_data_minimal)}
-      let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100 }
-      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :none }
+      let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100.1 }
+      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :other }
     
       it 'sets line 55 to total income tax withheld, rounded' do
         intake.reload
@@ -1619,8 +1619,8 @@ describe Efile::Nj::Nj1040Calculator do
 
     context 'when has both w2s and 1099-Rs' do
       let(:intake) { create(:state_file_nj_intake, :df_data_many_w2s)}
-      let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100 }
-      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :none }
+      let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100.1 }
+      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :other }
 
       it 'sets line 55 to total income tax withheld, rounded' do
         intake.reload
--- a/spec/lib/efile/nj/nj1040_calculator_spec.rb
+++ b/spec/lib/efile/nj/nj1040_calculator_spec.rb
@@ -1608,7 +1608,7 @@ describe Efile::Nj::Nj1040Calculator do
     context 'when has only 1099-Rs' do
       let(:intake) { create(:state_file_nj_intake, :df_data_minimal)}
       let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100.1 }
-      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :other }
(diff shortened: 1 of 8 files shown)


spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -1605,7 +1605,43 @@ def over_65_birth_year
...
+      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :military_pension }
+    
+      it 'sets line 55 to total income tax withheld, rounded' do
+        intake.reload
+        instance.calculate
+        expect(instance.lines[:NJ1040_LINE_55].value).to eq 100
+      end
+    end
+
+    # Remove when Flipper :show_retirement_ui is permanently enabled
```
jachan, line 1620: [DUST] Instead of having this ghost comment, would it be better to gate this test with Flipper? If :show_retirement_ui is enabled, then don't run this test?

I definitely appreciate your foresight to add this test, but I'm worried that we won't remember to remove this (which I suppose isn't that big of deal either)

jachan (approved):
One non-blocking comment about remembering to remove a test after Retirement UI is enabled.

spec/lib/efile/nj/nj1040_calculator_spec.rb
mluedke2: I don't want to actually flag tests on/off with feature flags-- I think they should all run. I thought about adding

```
before do
  allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(false)
end
```

to this test to try to make it clear that it's what would happen when flagged off, but that felt wrong because adding that code didn't actually have any effect on the test. so a comment felt like the best way to make clear that the test case will never happen anymore once the flag is off. Anyone doing a project-wide search for the flag to remove it would come across this comment, at least.

I also thought about not having this test at all, but that also felt wrong because the flag is off for users so I want it to be clear this behavior is out there now.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -1605,7 +1605,43 @@ def over_65_birth_year
       end
     end
 
-    context 'when no w2s' do
+    context 'when has only 1099-Rs' do
+      let(:intake) { create(:state_file_nj_intake, :df_data_minimal)}
+      let!(:state_file_1099r) { create :state_file1099_r, intake: intake, state_tax_withheld_amount: 100.1 }
```
jenny-heath, line 1610: [pebble] might be nice to have multiple 1099Rs here to check that it's summing something

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -1605,7 +1605,43 @@ def over_65_birth_year
...
+      let!(:state_specific_followup) { create :state_file_nj1099_r_followup, state_file1099_r: state_file_1099r, income_source: :military_pension }
```
jenny-heath, line 1611: is it necessary to have the followup here? it seems like the method doesn't take this into account

spec/lib/efile/nj/nj1040_calculator_spec.rb
jenny-heath: i think i'm not understanding the point of testing both cases (known income source and unknown income source), since the calculator method doesn't appear to look at it? i see `non_military_1099rs` takes income source into account, but that doesn't look like it affects line 55

jenny-heath (approved):
one question about what i think(?) is an unnecessary test but not blocking

spec/lib/efile/nj/nj1040_calculator_spec.rb
mluedke2: @jenny-heath  I totally agree that the followup isn't necessary for these tests at all.

I was building this from the context of all the other calculator lines thus far where we take 1099-R into account, and in those there specifically is a need to differentiate based on the `income_source`.

So the fact that this task is the first that **doesn't** feels novel enough that it warrants making clear in the tests. I actually initially coded this line like the others and incorrectly filtered out valid 1099-Rs, but did not incorporate `income_source` in the tests and thus everything passed despite being wrong. It wasn't until I re-read the ticket that I caught the detail to not filter, and decided to set up the tests so I my first attempt would've failed them.

I'm definitely open to just removing all the `followup` mentions in here and making it more concise, but if the above context feels valuable then I'll keep them as-is... Thoughts?

spec/lib/efile/nj/nj1040_calculator_spec.rb
mluedke2: done - i also combined this with the discussion about `income_source` so it sums from all different sources to make clear it works for all

spec/lib/efile/nj/nj1040_calculator_spec.rb
jenny-heath: ah, that makes sense. i'm fine with leaving the followups in there now that there aren't 2 separate test cases with and without them

PR #5611, opened 2025-02-20
[FYST-1830] remove the need for the full pdf download url to be in the page params

jnf:
## Link to pivotal/JIRA issue
[FYST-1830](https://codeforamerica.atlassian.net/browse/FYST-1830)

Yes - don't merge until JIRA issue is accepted! Needs to be tested on demo prior to release.

## What was done?
Shifts when in the flow the signed URL is generated. Before, the signed URL to the PDF was being generated in the action that rendered the 'download your pdf' button, and being passed as a param to a controller action that would log the activity and redirect to the URL. This flow exposed the signed PDF URL in the logs, and also created an avenue to siphon the signed url out of the client.

This PR moves the url generation to the end of the flow instead. Now there's no need for a param to hold the url between page loads; we log the request, generate the signed url, and redirect the client.

## How to test?
- Go through the GetYourPDF flow. Should be no noted difference in the client experience
- Inspect the logs (or ask eng to) to validate that the signed PDF url doesn't appear in the logged params
- Validate the appropriate access logs are created for each step in the retrieval

[FYST-1830]: https://codeforamerica.atlassian.net/browse/FYST-1830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/state_file/archived_intakes/pdfs_controller.rb
+++ b/app/controllers/state_file/archived_intakes/pdfs_controller.rb
@@ -14,14 +14,14 @@ module StateFile
       end
 
       def index
-        @prior_year_intake = StateFileArchivedIntake.find_by!(email_address: current_request.email_address)
-        @pdf_url = @prior_year_intake.submission_pdf.url(expires_in: pdf_expiration_time, disposition: "inline")
+        @state_code = StateFileArchivedIntake.find_by!(email_address: current_request.email_address).state_code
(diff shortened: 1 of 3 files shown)


DrewProebstel:
Went through the flow, lgtm. No pdf in the logs

PR #5612, opened 2025-02-20
FYST-1828 Allow redirect to pdf url

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1828

## What was done?
- It seems like you can't just redirect to one of these pdf urls or it returns a status 400. This doesn't show up in development because the pdf url is the same host (whereas in other envs it is a different host)
- I found `allow_other_host: true` on the internet, naturally
## How to test?
- Had to update test to make it pass but the test still doesn't catch this issue. Tried for a bit to write a test that generates a pdf url with a different host but gave up. Jey is working on some related changes and might find a way to test this.
- Before: tested on a different heroku app
image

- After: on this heroku app
image


anisharamnani (commented):
Thanks for the explanation Jenny!

anisharamnani (approved):

jenny-heath:
changes were incorporated in another branch

PR #5615, opened 2025-02-20
FYST-397 ensure that waiting to load data always proceeds correctly

anisharamnani:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-397

## What was done?

This is plagiarism from Mike's original description: 

Calls stream_for intake before broadcast_job_complete(intake). broadcast_to(intake, ...) is called by broadcast_job_complete(intake), but won’t do anything until stream_for is called. This should marginally speed up the data transfer page

## How to test?
We've written some tests. They are as good as they can get & don't actually test the problem. But, the [action cable testing documentation](https://github.com/palkan/action-cable-testing) is very limited.

### DIFF
--- a/app/channels/df_data_transfer_job_channel.rb
+++ b/app/channels/df_data_transfer_job_channel.rb
@@ -1,10 +1,10 @@
 class DfDataTransferJobChannel < ApplicationCable::Channel
   def subscribed
     intake = current_state_file_intake
+    stream_for intake
     if intake.raw_direct_file_data
       DfDataTransferJobChannel.broadcast_job_complete(intake)
     end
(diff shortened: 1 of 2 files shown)


arinchoi03 (approved):
Looks good, but would love some explanation/discussion soon about why we call `broadcast_job_complete` twice (once from within the `subscribed` method and once from within the `ImportFromDirectFileJob` and whether we need both.

anisharamnani:
@arinchoi03 I believe this is to check if the client already has `direct_file_data` & doesn't need to wait for the job to run.  as opposed to runninng the entire job.

PR #5616, opened 2025-02-20
FYST-1809 bounce from page when there is no request

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1809

## What was done?
- Right now if the URL is typed for any of our archived state file pages an error is thrown. This will stop that from happening so we can focus on the real errors.
- this does not solve all the IP related issues, see https://www.notion.so/cfa/PYA-accounts-Refactor-1a1373fd79b280a294fbc549d30a9345
## How to test?
-  try navigate to `/en/state_file/archived_intakes/verification_code/edit` without entering an email first.
- Risk Assessment
  - We could be bouncing people who are encountering whatever the unsolved problem is. But we are also logging that.

### DIFF
--- a/spec/controllers/state_file/archived_intake/archived_intake_controller_spec.rb
+++ b/spec/controllers/state_file/archived_intake/archived_intake_controller_spec.rb
@@ -88,5 +88,57 @@ describe StateFile::ArchivedIntakes::ArchivedIntakeController, type: :controller
         end
       end
     end
+
+    describe '#is_request_locked' do
+      before do
+        allow(controller).to receive(:current_request).and_return(request_instance)
+        allow(controller).to receive(:current_archived_intake).and_return(archived_intake)
+      end
+
+      context 'when the request is nil' do
+        before do
+          allow(controller).to receive(:current_request).and_return(nil)
+        end
+
+        it 'redirects to verification error page' do
+          expect(controller).to receive(:redirect_to).with(state_file_archived_intakes_verification_error_path)
+          controller.is_request_locked
+        end
+      end
+
+      context 'when the request is locked' do
+        before do
+          allow(request_instance).to receive(:access_locked?).and_return(true)
+        end
+
+        it 'redirects to verification error page' do
+          expect(controller).to receive(:redirect_to).with(state_file_archived_intakes_verification_error_path)
+          controller.is_request_locked
+        end
+      end
+
+      context 'when the archived intake is permanently locked' do
+        before do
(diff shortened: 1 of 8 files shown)


spec/controllers/state_file/archived_intake/archived_intake_controller_spec.rb
jenny-heath, line 1: [pebble] i think we typically use things like `.not_to be_redirect` instead of expecting it to receive redirect_to? seems a little more precise (because it checks the *response* as opposed to just checking that redirect_to was called, which it can be without actually redirecting) though i think they serve a very similar purpose. i will give a couple of examples below

jenny-heath: having said that i just realized that since there's no request being made in this spec, you can't really check the response :/ nvm

DrewProebstel: Yeah its def a little tricky to test, lmk if you have a better idea.

jenny-heath: but just to elaborate: i think a more typical (and you could argue, more thorough) test would involve testing an action on a controller (in this case, either a child of `ArchivedIntakeController` or a fake child controller created in the test file) and then having something like this:
```
it 'redirects to verification error page' do
  get :edit
  expect(response).to redirect_to state_file_archived_intakes_verification_error_path
end

it 'does not redirect' do
  get :edit
  expect(response).not_to be_redirect
end
```

jenny-heath (approved):
left some comments about an alternative testing approach and in writing out my argument for it, i looked into the other tests and realized that the before_action `is_request_locked` is covered in 3 or 4 different controller specs in the exact same way. if you're interested in a small refactor project, i would consolidate the test you added here with the request-locked specs for child controllers of `ArchivedIntakeController` (for example, spec/controllers/state_file/archived_intake/pdfs_controller_spec.rb:21) by making it a shared example. happy to provide more pointers if the suggestion is unclear!

i'm not sure if this is a priority change so i'm approving the PR. the refactor can be a follow-up if this is urgent but if not i'd just do it on this branch

spec/controllers/state_file/archived_intake/archived_intake_controller_spec.rb
jenny-heath: oh i totally missed your comment in the middle there. i was leaving several in a row so probably had to refresh the page. my real suggestion is in the approval comment!

PR #5618, opened 2025-02-21
Nj 117 retirement exclusion line 28a

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/117

## What was done?
- Update retirement income helper for 28a eligibility and total calculation
- Update 1040 calculator for 28a

## Logic

Eligibility:

- Line 27 <= $150,000
- 62 or older or blind/disabled

28a is min of:
- Line A = sum of 1099-R Box 2a amounts across eligible 1099-Rs only
- Line B = amount from max_exclusion table **using Line A as calculation value**
 
## How to test?
Springsteen MFJ


app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -1,27 +1,51 @@
 module Efile
   module Nj
     class NjRetirementIncomeHelper
+
+      attr_accessor :non_military_1099rs
       def initialize(intake)
         @intake = intake
+        @non_military_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
```
aloverso, line 8: This is duplicative logic from the NJ calculator, for now. I made a note in the story to do a fast-follow refactor to remove this from the calculator and ONLY reference the logic here in the helper instead for this.

Because of the urgency of getting this work tested, and because that refactor will require a bunch of updates to tests as well, figured it could be a fast-follow.

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -45,20 +69,38 @@ def retirement_exclusion_eligible?(line_15, line_27, line_28a)
...
+        true
+      end
+
+      def line_28a_primary_eligible?
+        primary_62_and_older? || @intake.primary_disabled_yes? || @intake.direct_file_data.is_primary_blind?
+      end
+
+      def line_28a_spouse_eligible?
+        spouse_62_and_older? || @intake.spouse_disabled_yes? || @intake.direct_file_data.is_spouse_blind?
+      end
```
aloverso, line 93: Compared to the 28b pattern, pulling out these methods prevents duplicative logic between the `eligible?` function and the total_eligible_income function. 

John and I discussed a fast-follow ticket where we would refactor 28b to use this pattern as well.

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -648,25 +690,14 @@ def over_65_birth_year
...
-      context 'when maximum_exclusion minus 28a is equal to total_eligible_nonretirement_income' do
-        it 'sets line 28b to total_eligible_nonretirement_income' do
-          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:retirement_exclusion_eligible?).and_return true
-          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:calculate_maximum_exclusion).and_return 50_000
-          allow(instance).to receive(:calculate_line_28a).and_return 1_000
-          allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:total_eligible_nonretirement_income).and_return 49_000
-          instance.calculate
-          expect(instance.lines[:NJ1040_LINE_28B].value).to eq(49_000)
-        end
-      end
```
aloverso: @jachan I apologize for the extra work you had to do because I know I suggested this in prior code review. But as I was looking more closely through these tests, it seems far more obvious to me that there's no real gain to a "when they're equal" test because of course they're equal so it comes out the same regardless of which "takes priority". My apologies!

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -1,39 +1,148 @@
...
+          expect(result).to eq(test_case[:expected])
+        end
+      end
+    end
+  end
+
+  describe "#calculate_maximum_exclusion for line 28b" do
     [
-      { total_income: 100_000, expected: 75_000 },
+      { traits: [:single], total_income: 100_000, expected: 75_000 },
```
aloverso, line 115: Adding the `:single` trait is just for readability sake

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -45,98 +154,164 @@
...
+
+        it "returns #{test_case[:expected]}" do
+          result = helper.line_28b_eligible?(test_case[:line_15], test_case[:line_27], test_case[:line_28a])
+          expect(result).to eq(test_case[:expected])
         end
       end
     end
+  end
+
+  describe "#total_eligible_nonretirement_income" do
```
aloverso, line 201: This diff got really messed up because there is both added code and also I fixed indentation on the existing tests. I did not update this describe block

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -45,98 +154,164 @@
...
-        end
+      it 'sums neither filers wages in total eligible income' do
+        expect(helper.total_eligible_nonretirement_income).to eq(0)
       end
+    end
+  end
 
-      context "when only spouse is over 62" do
-        let(:intake) { create(:state_file_nj_intake, :df_data_one_dep, :mfj_spouse_over_62) }
+  describe "#total_eligible_nonmilitary_1099r_income" do
```
aloverso, line 253: This describe block **is** new and should be reviewed

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -45,98 +154,164 @@
...
-            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_27).and_return(test_case[:line_27])
-            allow_any_instance_of(Efile::Nj::Nj1040Calculator).to receive(:calculate_line_28a).and_return(test_case[:line_28a])
-            instance.calculate
-          end
-
-          it "returns #{test_case[:expected]}" do
-            result = helper.retirement_exclusion_eligible?(test_case[:line_15], test_case[:line_27], test_case[:line_28a])
-            expect(result).to eq(test_case[:expected])
-          end
+  describe "#line_28b_eligible?" do
```
aloverso, line 164: This diff got really messed up because there is both added code and also I fixed indentation on the existing tests. I did not update this describe block

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -260,19 +260,20 @@ def calculate_line_15
       end
 
       def calculate_line_20a
-        applicable_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
-          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
-        end
-
-        applicable_1099rs.sum(&:taxable_amount).round
+        non_military_1099rs.sum(&:taxable_amount).round
```
aloverso, line 263: This is a small refactor to use the existing method in the calculator here rather than redoing the logic. This will get updated again in my fast-follow refactor for both this line and line_20b to use the helper's `@non_military_1099rs` instead

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -1,39 +1,148 @@
 require 'rails_helper'
 
 RSpec.describe Efile::Nj::NjRetirementIncomeHelper do
-  describe ".calculate_maximum_exclusion" do
```
aloverso, line 4: Note - in addition to renaming these, using a `#` which is the rspec pattern for instance methods, as opposed to the `.` which is the rspec pattern for class methods

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -45,20 +69,38 @@ def retirement_exclusion_eligible?(line_15, line_27, line_28a)
...
+
+      def calculate_maximum_exclusion(total_income, calculation_value)
```
jachan, line 95: [Pebble] Can we make `calculation_value` more descriptive? Perhaps `income_for_exclusion`?

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -336,17 +336,22 @@ def calculate_line_27
       end
 
       def calculate_line_28a
```
jachan, line 338: [Dust/Pebble] Right now, the display of 28a is not being gated behind Flipper at all. With the introduction of real 28a logic, do we need to revisit adding the feature flag to the XML generation code for 28a?

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -1,27 +1,51 @@
...
       def initialize(intake)
         @intake = intake
+        @non_military_1099rs = @intake.state_file1099_rs.select do |state_file_1099r|
+          state_file_1099r.state_specific_followup.present? && state_file_1099r.state_specific_followup.income_source_none?
+        end
+      end
+
+      def total_eligible_nonmilitary_1099r_income
+        total_eligible_income = 0
+        eligible_ssns = []
```
jachan, line 15: [DUST] Flagging another potential refactor for the future fast-follow:

Could move creation of `eligible_ssns` into the `initialize` method for the class, since we end up doing this in multiple places

spec/lib/efile/nj/nj1040_calculator_spec.rb
jachan: [DUST] That's what my original thought was, but I still thought it might be helpful to assert that in the tests, rather than, for example, throwing an error.

What do we gain from removing the test?

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
jachan: Neat, did not know this.

jachan (approved):
Two concerns I have, otherwise looks good:
- Flipper feature flag management for 28a
- A request to give the second argument to maximum_exclusion a more descriptive name

app/lib/efile/nj/nj_retirement_income_helper.rb
aloverso: I'm not sure we could, because they're not the exact same list. Eligible SSNs for 28a is different from the eligible SSNs for 28b because they have different eligibility logic (28a incorporates a blind / disabled check as well)

app/lib/efile/nj/nj_retirement_income_helper.rb
aloverso: I'll check with the PMs on a good idea for a name. I don't feel like I have a good understanding of what this value in the table represents contextually and why it's different between 28a and 28b, enough to feel confident about a descriptive name here. Do you know what it's supposed to represent?

app/lib/efile/nj/nj1040_calculator.rb
aloverso: Oh didn't realize, wasn't in the story. Will check with PMs whether this should be feature flagged

spec/lib/efile/nj/nj1040_calculator_spec.rb
aloverso: I know I'm flip-flopping on this point - I'm sorry! But I feel like at best, we don't gain anything from it at all - it doesn't give me any additional confidence in the behavior. I'd say that the potential gains from removing it are a) less code to maintain overall in the tests - why have tests we don't need? and b) it's not _misleading_ per se, but it feels random for us to describe the behavior as `sets line 28b to total_eligible_nonretirement_income` which is not actually fully true. Could be equally described as `sets line 28b to maximum_exclusion minus 28a`. The fact that it could be either but we have to pick one is what feels strange to me. [loosely held opinion]

app/lib/efile/nj/nj_retirement_income_helper.rb
jachan: I'm not sure, unfortunately. I wrote a long explanation below but then realized it was completely wrong. Definitely a PM question 😭

spec/lib/efile/nj/nj1040_calculator_spec.rb
jachan: a) I think we do gain something from the test though, we are clarifying what the behavior of the method is in this edge case

b) perhaps this is just as simple as changing the test name to `it sets line 28b to the value shared by total_eligible_nonretirement_income and maximum_exclusion minus 28a`? I hate how wordy it is haha, but that IS an accurate and descriptive name.

app/lib/efile/nj/nj_retirement_income_helper.rb
mpidcock: i agree that this is awkward and unintuitive, but don't have any insight unfortunately!

mpidcock (approved):
Looks good, I appreciate the comments 🙏

spec/lib/efile/nj/nj1040_calculator_spec.rb
aloverso: 👍 will add the test back in!

PR #5629, opened 2025-02-22
[FYST-1842] Do not allow clients to update W2 money amount with a nil value

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1842

## What was done?
- Added validations if user enters `nil` into these fields
- they were _not_ folded into existing `validates` lines (using `allow_blank: false`) b/c it got too messy (i.e. need to evaluate local boxes & box14 conditional, easier to see in one place, i think)
- instead, they are only run from the `state_file_edit` context (w2 editing screen; not from the income review page), since we believe this page is the source of the error that we're seeing.
- Furthermore, we should ensure any calculations we do guard against potential `nil` values -- this was communicated to Tiffany & I'm flagging places where I'm seeing these errors pop up https://codeforamerica.atlassian.net/browse/FYST-1842
- Ideally, we would allow users to proceed without the client having to put in the `0` -- however, this is not how our money fields work currently across the board (user always has to enter a value to proceed), so we're going to stay ✨ consistent until we determine a better approach for handling. Discussed this with Jey & Erica 
## How to test?
  - manual tests
  - Test data used: 
    - MD zeus w2s
    - NC Bert 1099r (no w2 edit; checking nothing broke)
    - NJ Zeus box 14
    - AZ Tycho single 1099R
- Specify any relevant testing environments used: dev
- Risk Assessment
  - should be completely isolated to the w2 edit page
- Before
- After
Screenshot 2025-02-24 at 9 50 21 AM


app/views/state_file/questions/w2/edit.html.erb
```diff
@@ -26,7 +26,7 @@
               <%= f.vita_min_money_field(
                     field_name.to_sym,
                     t(".box14_#{code['name'].downcase}_html"),
-                    options: { value: value || 0 },
+                    options: { value: value },
```
arinchoi03, line 29: removed b/c we were pre-filling with 0 and then showing them an error when they saved to "put in 0"

spec/models/state_file_w2_spec.rb
```diff
@@ -35,20 +35,23 @@
...
+           local_wages_and_tips_amount: 8000,
+           locality_nm: "NYC",
+           state_file_intake: intake,
+           state_income_tax_amount: 600,
+           state_wages_amount: 8000,
+           box14_fli: 0,
+           box14_stpickup: 0,
+           box14_ui_hc_wd: 0,
+           box14_ui_wf_swf: 0,
+           w2_index: 0
```
arinchoi03, line 50: a `valid` state_file_w2 needs the values at set to non-nil value

mpidcock (approved):

app/models/state_file_w2.rb
```diff
@@ -70,12 +71,38 @@ class StateFileW2 < ApplicationRecord
     validate :validate_tax_amts
     validate :state_specific_validation
   end
+
+  validate :validate_nil_tax_amounts, on: :state_file_edit
```
arinchoi03, line 75: re: conversation with @powersurge360 we could do a `presence` check here instead; maybe I could've added another block for `state_file_edit` form instead of creating a method:
```
with_options on :state_file_edit do
supported_box14_codes
    validates :state_wages_amount, :presence => {message: ->(_object, _data) { I18n.t("state_file.questions.w2.edit.no_money_amount") }},
    validates :state_income_tax_amount,  :presence => {message: ->(_object, _data) { I18n.t("state_file.questions.w2.edit.no_money_amount") }},
    validates :local_wages_and_tips_amount,  :presence => {message: ->(_object, _data) { I18n.t("state_file.questions.w2.edit.no_money_amount") }}, if: -> StateFile::StateInformationService.w2_include_local_income_boxes(state_file_intake.state_code) }
    validates :local_income_tax_amount,  :presence => {message: ->(_object, _data) { I18n.t("state_file.questions.w2.edit.no_money_amount") }}, if: -> StateFile::StateInformationService.w2_include_local_income_boxes(state_file_intake.state_code) }
    validates :box14_fli,  :presence => {message: ->(_object, _data) { I18n.t("state_file.questions.w2.edit.no_money_amount") }}, if: -> { supported_box14_codes.includes("box14_fli")}
...etc
``` 
I think the pros of the pulled out logic is that I know exactly what were doing in the `state_file_edit` form only and it's easy to see when the validations run (the if blocks get gnarly toward the end of the lines)

arinchoi03: We also chatted about using https://api.rubyonrails.org/v7.1/classes/ActiveRecord/Normalization/ClassMethods.html
if we decided to default `nil` values to 0. 

Something like this?
```
  normalizes :state_wages_amount, with: -> state_wages_amount { state_wages_amount.to_d }
```

PR #5634, opened 2025-02-24
Added a concept of default values to forms. Also includes using models &

powersurge360:
## Link to pivotal/JIRA issue
- No JIRA ticket

## What was done?
When you use a form's `set_attributes_for` on fields with defaults and don't set a value, you can get database level failures. This provides two possible fixes; either:
  a. Specifically provide a `defaults` dictionary or
  b. Use a model rather than a symbol for the first argument to `set_attributes_for`

When doing the latter, we inspect the `column_defaults` on the model and populate the default values with which we initialize the form, so they will never be nil without being explicitly set to nil.

This behavior is opt in and uses new arguments and reflection on the model. It should change no other behavior.

This resolves issues like the ones seen on [update_13614c_form_page1](https://github.com/codeforamerica/vita-min/blob/7409348c052b5440b8d387d59a7ceee946291c9c/app/forms/hub/update_13614c_form_page1.rb#L108-L117).

It will also set us up well for a default `QuestionsForm#save` method; if all keys are models, then we can assume that their respective attributes are available and we can save it out.

## How to test?
- It's pretty high touch. Doing basically anything in the application will either succeed or fail.
- There's a chance that this will need to be touched up a little more when it gets first used, I didn't really want to explicitly change any forms, but I look forward to playing with it in the future.

- Before
- After

### DIFF
--- a/app/forms/concerns/form_attributes.rb
+++ b/app/forms/concerns/form_attributes.rb
@@ -11,7 +11,7 @@ module FormAttributes
       default_attrs = self.class.scoped_defaults.values.reduce(&:merge)
 
       self.class.scoped_attributes.values.flatten.each do |attribute_string|
-        next unless send(attribute_string).nil?
+        next unless send(attribute_string).blank?
         send("#{attribute_string}=", default_attrs&.fetch(attribute_string, nil))
       end
     end
--- a/app/forms/concerns/form_attributes.rb
+++ b/app/forms/concerns/form_attributes.rb
@@ -5,8 +5,8 @@ module FormAttributes
     extend AutoStripAttributes
     class_attribute :attribute_names
 
-    def initialize
-      super
+    def initialize(*args, **kwargs)
+      super(*args, **kwargs)
 
       default_attrs = self.class.scoped_defaults.values.reduce(&:merge)
 
--- a/app/forms/concerns/form_attributes.rb
+++ b/app/forms/concerns/form_attributes.rb
@@ -4,20 +4,46 @@ module FormAttributes
   included do
     extend AutoStripAttributes
     class_attribute :attribute_names
+
+    def initialize
+      super
+
+      default_attrs = self.class.scoped_defaults.values.reduce(&:merge)
+
+      self.class.scoped_attributes.values.flatten.each do |attribute_string|
+        next unless send(attribute_string).nil?
+        send("#{attribute_string}=", default_attrs&.fetch(attribute_string, nil))
+      end
+    end
   end
 
   class_methods do
-    def set_attributes_for(model, *attributes)
+    def set_attributes_for(model, *attributes, **options)
+      if options[:defaults].present?
+        scoped_defaults[model] = options[:defaults].compact
+      elsif model.respond_to?(:column_defaults)
+        scoped_defaults[model] = model.column_defaults
+          .compact
+          .transform_keys(&:to_sym)
+          .slice(*attributes)
+      end
+
       scoped_attributes[model] = attributes
+
       self.attribute_names = scoped_attributes.values.flatten
       attribute_strings = Attributes.new(attributes).to_s
 
       attr_accessor(*attribute_strings)
+
       auto_strip_attributes *attribute_strings, virtual: true
     end
 
     def before_validation_squish(*attributes)
       auto_strip_attributes *attributes, virtual: true, squish: true
+    end
+
+    def scoped_defaults
+      @scoped_defaults ||= {}
     end
 
     def scoped_attributes
--- a/app/forms/hub/update_13614c_form_page1.rb
+++ b/app/forms/hub/update_13614c_form_page1.rb
@@ -1,6 +1,6 @@
 module Hub
   class Update13614cFormPage1 < ClientForm
-    set_attributes_for :intake,
+    set_attributes_for Intake::GyrIntake,
                        :primary_first_name,
                        :primary_last_name,
                        :primary_middle_initial,
@@ -107,7 +107,7 @@ module Hub
     def save
       return false unless valid?
 
-      modified_attributes = attributes_for(:intake)
+      modified_attributes = attributes_for(Intake::GyrIntake)
                               .except(:primary_birth_date_year, :primary_birth_date_month, :primary_birth_date_day, :spouse_birth_date_year, :spouse_birth_date_month, :spouse_birth_date_day)
                               .merge(
                                 primary_birth_date: parse_date_params(primary_birth_date_year, primary_birth_date_month, primary_birth_date_day),
@@ -115,21 +115,6 @@ module Hub
                               )
       modified_attributes[:ever_married] = modified_attributes.delete(:never_married) == "yes" ? "no" : "yes"
       modified_attributes[:dependents_attributes] = formatted_dependents_attributes
-
-      # we are getting null violations from the database; the below lines fix that.
-      modified_attributes[:multiple_states] ||= 'unfilled'
-      modified_attributes[:primary_owned_or_held_any_digital_currencies] ||= 'unfilled'
-      modified_attributes[:spouse_issued_identity_pin] ||= 'unfilled'
-      modified_attributes[:spouse_owned_or_held_any_digital_currencies] ||= 'unfilled'
-      modified_attributes[:balance_pay_from_bank] ||= 'unfilled'
-      modified_attributes[:presidential_campaign_fund_donation] ||= 'unfilled'
-      modified_attributes[:receive_written_communication] ||= 'unfilled'
-      modified_attributes[:savings_split_refund] ||= 'unfilled'
-      modified_attributes[:register_to_vote] ||= 'unfilled'
-      modified_attributes[:primary_visa] ||= 'unfilled'
-      modified_attributes[:spouse_visa] ||= 'unfilled'
-      modified_attributes[:refund_other_cb] ||= 'unfilled'
-      modified_attributes[:married_for_all_of_tax_year] ||= 'unfilled'
 
       @client.intake.update(modified_attributes)
       @client.touch(:last_13614c_update_at)
--- a/spec/forms/concerns/form_attributes_spec.rb
+++ b/spec/forms/concerns/form_attributes_spec.rb
@@ -86,7 +86,7 @@ RSpec.describe FormAttributes do
         expect(attributes).to be_empty
       end
 
-      it 'should create initialize with nil if defaults are not passed' do
+      it 'should initialize with nil if defaults are not passed' do
         attributes = subject.attributes_for(:without_defaults)
         expect(attributes).to match(attribute_one: nil, attribute_two: nil)
       end
--- a/spec/forms/concerns/form_attributes_spec.rb
+++ b/spec/forms/concerns/form_attributes_spec.rb
@@ -1,7 +1,28 @@
 require 'rails_helper'
 
 class ExampleModel
-  # This was copied from OutgoingEmail to have a real example, but not literally tie to an existing model (which may change)
+  # This was copied from OutgoingEmail to have a real example, but not literally
+  # tie to an existing model (which may change)
+  def self.column_defaults
+    {
+      "id" => nil,
+      "body" => "some body text",
+      "client_id" => nil,
+      "created_at" => nil,
+      "mailgun_status" => "sending",
+      "message_id" => nil,
+      "sent_at" => nil,
+      "subject" => nil,
+      "to" => nil,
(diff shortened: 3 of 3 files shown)


spompea-cfa:
Cool, I'm glad you're tackling this!

I had a question about the behavior when a defaults hash is specified. I did a 
little experiment, and it seems like the behavior will be: if a column name 
isn't mentioned in the defaults hash, then even if that column has a default 
value set in the model, it will still get set to nil. Assuming I am intepreting 
that right (maybe I'm not), my first thought is: that feels unintutive to me; it 
seems like it would be better to fall back on a model-defined default value here? 
Although, I suppose all that's happening really is that if the column isn't
mentioned in the defaults hash, then basically it's treated just the same as 
it would already get treated today with the current production behavior. Still,
could be confusing for someone who comes along later and
makes use fo the defaults hash capability.  I guess
I'm wondering if your defaults hash idea could be taken even further.

---

My little experiement was to modify the spec test like so:

```
class SecondExampleModel
  # This was copied from OutgoingEmail to have a real example, but not literally
  # tie to an existing model (which may change)
  def self.column_defaults
    {
      "id" => nil,
      "body" => "some body text",
      "client_id" => nil,
      "created_at" => nil,
      "mailgun_status" => "sending",
      "message_id" => nil,
      "sent_at" => nil,
      "subject" => nil,
      "to" => nil,
      "updated_at" => nil,
      "user_id" => 'snoopy'       # <------------ ADDED THIS
    }
  end
end

```

And then... way down at the bottom:

```
      it 'should not use column defaults when a defaults hash is specified' do
        expect(subject.mailgun_status).not_to eq 'sending'
        expect(subject.mailgun_status).to eq 'not_sent'
        expect(subject.user_id).to eq 'snoopy'   # <------- AND ADDED THIS
      end
```

Gets me this:

```
Failures:

  1) FormAttributes#set_attributes_for when passing a model should not use column defaults when a defaults hash is specified
     Failure/Error: expect(subject.user_id).to eq 'snoopy'
     
       expected: "snoopy"
            got: nil
     
       (compared using ==)
```

powersurge360:
I actually chose to implement it this way after thinking about it for a bit. It might be surprising, but it is surprising in an immediately caught way, imo, whereas the opposite is opaque. In my mind, someone who expects the defaults from the database to be there when a defaults hash is provided is likely to immediately notice when they test it (as you have just done). Someone who expects the defaults hash to override the column_defaults (but doesn't test it) might not notice, as even `column_defaults` is not really common working rails knowledge. The way it's coded currently, the buck stops w/ the defaults hash that gets passed in, if you opt into it, you take full ownership of what the defaults are with no hidden surprises changing those values, which I think is the most reasonable for providing defaults. Otherwise you have to square your user provided defaults with _default_ defaults.

It is debatable for sure though and I think I'd like some more opinions before I decide. Ultimately, it's as simple as doing a `#reverse_merge` of the column_defaults into defaults passed, when available.

spompea-cfa:
ok, very interesting to here your thoughts. cool. are you wanting to merge this now, or do you want to let it marinate some more?

spompea-cfa (approved):
approved 🚂

spec/forms/concerns/form_attributes_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    context "when passing a symbol" do
+      subject { SymbolExampleForm.new }
+
+      it 'should create an empty hash by default' do
+        attributes = subject.attributes_for(:empty)
+        expect(attributes).to be_a Hash
+        expect(attributes).to be_empty
+      end
+
+      it 'should create initialize with nil if defaults are not passed' do
```
jenny-heath, line 89: i feel like "create" and "initialize" are accidentally redundant here?

spec/forms/concerns/form_attributes_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+        attributes = subject.attributes_for(:without_defaults)
+        expect(attributes.keys).to eq [:attribute_one, :attribute_two]
+        expect(attributes.values).to be_all nil
```
jenny-heath, line 92: i think this also works! i think what's nice about it is seeing it as a hash but i don't really have a preference 
```suggestion
        expect(attributes).to match({attribute_one: nil, attribute_two: nil})
```

jenny-heath: (same goes for the following test blocks, ofc)

app/forms/concerns/form_attributes.rb
```diff
@@ -4,22 +4,48 @@ module FormAttributes
   included do
     extend AutoStripAttributes
     class_attribute :attribute_names
+
+    def initialize(*args, **kwargs)
+      super(*args, **kwargs)
```
jenny-heath, line 9: i think i'm getting a little confused about where `super` is, since this is a concern not a class? to use an example: `QuestionsForm` includes `FormAttributes` and initialize looks like this:
```
def initialize(intake = nil, params = {})
    @intake = intake
    super(params)
  end
```
where `super` is the initialize on `ActiveModel`, i believe. so in that context:
- it seems that `super` in `QuestionsForm` would now refer to `FormAttributes#initialize` (i checked this locally)
- Q: would `super` in the concern then refer to `QuestionsForm#initialize`? it seems like that should be the case but i couldn't figure out how to verify that

powersurge360: That's right. `*args` and `**kwargs` receive and then pass on the call without doing anything to the values. I think in the case of `#include` it will actually call into `Form#initialize` as `QuestionsForm#initialize` should have already run.

In any case, I don't think order _particularly_ matters as I'm only dealing with values that are nil. Screenshot of the output of `QuestionsForm.ancestors` for reference:

image

jenny-heath (commented):
ok i'm pretty sure i understand what's going on here but was wondering if this PR could include an example of this in action—maybe in both ways? (the model-level defaults and the defaults hash passed to `set_attributes_for`) i think that would give me more confidence that it works because the tests are a bit abstract.

i think i have reasonable confidence that it at least doesn't break the existing form usage because the tests are passing.

also, why does the branch name start with `nj/`...?

powersurge360:
Oh sorry, old habit. I use `nj` to mean `no-jira` but I should have made a ticket or not used that name heh.

spec/forms/concerns/form_attributes_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      "created_at" => nil,
+      "mailgun_status" => "sending",
+      "message_id" => nil,
+      "sent_at" => nil,
+      "subject" => nil,
+      "to" => nil,
+      "updated_at" => nil,
+      "user_id" => nil
+    }
+  end
```
arinchoi03, line 20: just verifying that this `column_defaults` exist on a model by default -- using the database definitions of what the default is for these fields, is that correct?

app/forms/concerns/form_attributes.rb
```diff
@@ -4,22 +4,48 @@ module FormAttributes
...
+
+      default_attrs = self.class.scoped_defaults.values.reduce(&:merge)
```
arinchoi03, line 11: i know this is lazy & I'm going to pull down a branch to see what this is doing, but wanted an explanation on what this is up to 🙈

app/forms/concerns/form_attributes.rb
```diff
@@ -4,22 +4,48 @@ module FormAttributes
...
+        send("#{attribute_string}=", default_attrs&.fetch(attribute_string, nil))
+      end
+    end
   end
 
   class_methods do
-    def set_attributes_for(model, *attributes)
+    def set_attributes_for(model, *attributes, **options)
+      if options[:defaults].present?
+        scoped_defaults[model] = options[:defaults].compact
```
arinchoi03, line 23: is there an example of when we would want to override the `model.column_defaults` in favor of the `options[:default]` passed in?

arinchoi03 (commented):
few comments but I think i mostly understand it

spec/forms/concerns/form_attributes_spec.rb
arinchoi03: i agree this would be clearer

app/forms/concerns/form_attributes.rb
arinchoi03: ok it's doing this:
```
(byebug) self.class.scoped_defaults.values
[{:default_attribute_one=>"foo", :default_attribute_two=>"bar"}]
(byebug) self.class.scoped_defaults.values.reduce(&:merge)
{:default_attribute_one=>"foo", :default_attribute_two=>"bar"}
```
so basically taking it out of the array?

app/forms/concerns/form_attributes.rb
arinchoi03: also a test specifying this preference

app/forms/hub/update_13614c_form_page1.rb
jenny-heath, line 1: just to clarify: do we think the fact that this works is tested by `edit_13614c_spec.rb`?

jenny-heath: my read is yes but would love a second opinion

spec/forms/concerns/form_attributes_spec.rb
jenny-heath: yes, i didn't know about this either! https://apidock.com/rails/ActiveRecord/Base/column_defaults/class

app/forms/concerns/form_attributes.rb
jenny-heath: i think it is [intentionally prioritizing the defaults hash over column_defaults](https://github.com/codeforamerica/vita-min/pull/5634#issuecomment-2685530364) and think [this test](http://github.com/codeforamerica/vita-min/pull/5634/files#diff-5881b61ee7e02c6ca3340a03af7dd2de98c8feb14aebb3e59cc4fc28ba1cf4cdR123-R126) specifies the preference

jenny-heath: (idk about an example of when the defaults hash would override the model defaults with something different, afaik it's theoretical for now)

jenny-heath (approved):
idk why i had to hype myself up to approve this, i think i am still scared of how high-touch it is but i also have reasonable confidence that it's not breaking anything

would like to re-up my testing suggestion: https://github.com/codeforamerica/vita-min/pull/5634/files#r1974302500
also, speaking of testing, would like to get confirmation that you feel the 13614c form is sufficiently tested

app/forms/concerns/form_attributes.rb
powersurge360: There could be, it mimics the way the scoped_attributes work. Although we scope at where we 'register' our attributes, we don't have any kind of scoping on the attribute accessor. In other words if we have:

```ruby
scoped_defaults = {
  some_model: {foo: 'foo', bar: 'bar'},
  some_other_model: { baz: 'baz' }
}
```

What this code would do is call '#values' to get the values of each key, wrapping in a list:
```ruby
scoped_defaults.values
# => [{foo: 'foo', bar: 'bar'}, {baz: 'baz'}]
```

And then finally `#reduce(&:merge)` will pop off the initial value in the list to use as the collection point, merge the next item into it, and return it as the next call's merge. In other words, each hash will be merged into a single hash left to right. If there is one hash this is the same as calling `scoped_defaults.values.pop` but if there are multiple it is equivalent to calling `{}.merge(*scoped_default.values)`, if that makes sense.

powersurge360:
Sorry! I had missed your suggested change, initially. If you would feel more comfortable we can tuple and do a test refactor in a couple of forms to see that things still work.

app/forms/concerns/form_attributes.rb
powersurge360: Yep, to succinctly re-iterate, I think there are two options:

1. Overwrite `column_defaults` completely (for this form)
2. Merge `column_defaults`

Both can be reasonably said to be surprising but I feel option 1. is more easily discoverable. I still could be swayed, though.

powersurge360: Oh, also I kind of mostly intend on folks either passing a model as the key to opt into using `column_defaults` or passing in a defaults hash. I would expect doing _both_ to be kind of niche. I imagine it would only be done if you wanted the `defaults` behavior but also wanted the free default `#save` (which doesn't exist yet, but I'll have another PR for in the next month-ish, probably)

spompea-cfa:
@powersurge360,  @jenny-heath - hey, just fyi, in case it's relevant: I just merged the PR for the reenabled 14-C mapping spec tests.

PR #5402, opened 2025-01-22
Fyst 1636 social security benefits data model and UI

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1636

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

## What was done?
- Add social security benefits page with controller and form with validations 

## How to test?
- Added in unit tests for the controller and form

## Screenshots (for visual changes)
Screenshot 2025-01-21 at 6 20 50 PM

PR #5636, opened 2025-02-25
Fyst 1798 use df json data for mfj ssa income

mrotondo:
_TBD (asked product & program): Are we supposed to filter out SSB reports based on whether they are of type SSB-1099 or RRB-1099?_

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1798

## What was done?
- Removes the page (implemented in https://github.com/codeforamerica/vita-min/pull/5402) that asks MD filers for the primary & spouse filers' SSB amounts
- Removes the corresponding columns on `state_file_md_intakes`
- Adds a DF JSON parser for the new socialSecurityReports items, and uses values from those items in the calculator in place of the now-removed columns
## How to test?
- Go through the flow for a persona (Tiger 55 should work, now that i re-imported them from DF) who has `socialSecurityReports` in their DF JSON. Verify that the output XML & PDF have the correct social security benefit amounts for the primary and spouse filers.


app/models/state_file_md_intake.rb
```diff
@@ -112,6 +110,9 @@
 #
 class StateFileMdIntake < StateFileBaseIntake
   include MdResidenceCountyConcern
+
+  self.ignored_columns += %w[primary_ssb_amount spouse_ssb_amount]
```
arinchoi03, line 114: this is data that clients self-reported. we can't really "backfill" this data back into the json files, so do we just keep it forever indefinitely?

app/models/direct_file_json_data.rb
```diff
@@ -111,4 +129,11 @@ def filers
   def dependents
     data["familyAndHousehold"]&.map { |dependent| DfJsonDependent.new(dependent) } || []
   end
+
+  private
+
+  def social_security_benefit_amount_for(filer)
+    social_security_reports.filter { |social_security_report| social_security_report.recipient_tin == filer.tin }.sum(&:net_benefits)
```
arinchoi03, line 136: sorry this is on my mind, but is there any likelihood at all we could get a `null` for the `net_benefits` amount?

app/lib/efile/md/md502_r_calculator.rb
```diff
@@ -61,16 +61,16 @@ def calculate_line_8
...
             @intake.direct_file_data.fed_ssb.round
           end
         end
       end
 
       def calculate_line_9b
-        if @intake.filing_status_mfj? && @intake.direct_file_data.fed_ssb.positive? && @intake.spouse_ssb_amount.present?
-          @intake.spouse_ssb_amount.round
+        if @intake.filing_status_mfj? && @intake.direct_file_data.fed_ssb.positive?
+          @intake.direct_file_json_data.spouse_filer_social_security_benefit_amount&.round
```
arinchoi03, line 73: just noticing that `calculate_line_9a` returns `0` if the amount is nil` but `calculate_line_9b` returns `nil` if it's nil, probably should choose one or the other, although it'll end up in the same place since this ultimately uses `add_element_if_present`

arinchoi03 (commented):
looks good but just one q about keeping the self-reported data

app/models/direct_file_json_data.rb
mpidcock: The `type: :money_amount` on the `net_benefits` json accessor defaults to zero if the key is nil, and DF has committed to always send the keys, even if there is no data.

On that note though, we should probably establish a `tin` type that strips hyphens, because they send us tins with hyphens in the json, and the xml sends them without, so comparisons like this break.

arinchoi03: oh yea i always forget about the json data having a default 0 for `money_amount` type! phew

app/models/state_file_md_intake.rb
mrotondo: I undid the change to drop the columns, and made the calculator attempt to use them if they're present. I also made them nullable in the db with no default value, so that they should not be `.present?` for intakes created after this change.

app/lib/efile/md/md502_r_calculator.rb
mrotondo: Changed so that they both return nil if no value is present

arinchoi03 (approved):

jnf (approved):

PR #5495, opened 2025-01-31
Nj 249 ineligible content

mluedke2:
## Link to pivotal/JIRA issue

https://github.com/newjersey/affordability-pm/issues/249

## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

**Reminder**: merge main into this branch and get green tests before merging to main
## What was done?
- Updated content for EN and ES, and did the usual tricks to put in NJ-specific content without directly checking for `nj` in code
- The ineligibility reason strings were structured in a way that the translators were only getting half the sentence and some English and some Spanish reasons didn't make complete proper sentences. So I fixed those.
- Noticed that the partial was just rendering four sibling elements and was not reused, and this was confusing the nesting of strings in the YML. So I brought that content back into the parent ERB file and restructured the YML.
- i18n health checks were failing, thinking that the NJ strings were unused
  - I decided to use the linter comments and mention NJ directly for the ones where we are the only exception
  - There was one spot where MD also has custom content, so I told the i18n config to ignore
## How to test?
- Use a persona like Streep and select that not all had health insurance (just one of the many ineligibility reasons)
## Screenshots (for visual changes)

image

image

image

PR #5641, opened 2025-02-26
Nj 249 bugfix non nj

mluedke2:
## Link to pivotal/JIRA issue

Followup to
https://github.com/codeforamerica/vita-min/pull/5495

## What was done?
- The check for NJ content was checking on info in the StateInformationService that was present in other states and thus the new NJ-specific button was appearing in other states
- I added tests to the two related controllers that verified that NJ content was NOT showing for those states. (These tests failed before the implementation code in the PR, meaning the issue was reproduced).

## How to test?
- For NJ, use a persona like Streep and select that not all had health insurance (just one of the many ineligibility reasons)
- For AZ, MD, etc., steps TBD

### DIFF
--- a/app/views/state_file/questions/data_transfer_offboarding/edit.html.erb
+++ b/app/views/state_file/questions/data_transfer_offboarding/edit.html.erb
@@ -13,15 +13,11 @@
     <p><%= vita_introduction %></p>
   <% end %>
 
-  <% if current_state_code == "nj" %>
-    <% vita_link = I18n.locale == :en ? StateFile::StateInformationService.vita_link_en(current_state_code) : StateFile::StateInformationService.vita_link_es(current_state_code) %>
-    <% if vita_link.present? %>
-      <p><%= link_to t("state_file.questions.eligible.faq_link"), state_faq_path(us_state: current_state_code) %></p>
-      <%= link_to(vita_link, class: "button button--primary button--wide spacing-below-15", target: "_blank", rel: "noopener noreferrer") do %>
-        <%= t('state_file.questions.eligible.vita_option.connect_to_vita') %>
-      <% end %>
-    <% end %>
-  <% end %>
+  <% vita_link_key = "state_file.questions.eligible.vita_option.connect_to_vita.#{current_state_code}" %>
+  <% vita_link_href = StateFile::StateInformationService.send("vita_link_#{I18n.locale}", current_state_code) %>
+  <% faq_link_href = state_faq_path(us_state: current_state_code) %>
+  <% vita_link = t(vita_link_key, faq_link_href: faq_link_href, vita_link_href: vita_link_href, default: '') %>
+  <%= vita_link.html_safe if vita_link.present? %>
 
   <%= render 'state_file/questions/eligible/other_filing_options' %>
 <% end %>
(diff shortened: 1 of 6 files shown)


app/views/state_file/questions/data_transfer_offboarding/edit.html.erb
```diff
@@ -13,11 +13,13 @@
     <p><%= vita_introduction %></p>
   <% end %>
 
-  <% vita_link = I18n.locale == :en ? StateFile::StateInformationService.vita_link_en(current_state_code) : StateFile::StateInformationService.vita_link_es(current_state_code) %>
-  <% if vita_link.present? %>
-    <p><%= link_to t("state_file.questions.eligible.faq_link"), state_faq_path(us_state: current_state_code) %></p>
-    <%= link_to(vita_link, class: "button button--primary button--wide spacing-below-15", target: "_blank", rel: "noopener noreferrer") do %>
-      <%= t('state_file.questions.eligible.vita_option.connect_to_vita') %>
+  <% if current_state_code == "nj" %>
```
mmazanec22, line 16: [dust] For this case and the other one, I wonder if instead of checking for a state code or for the link itself, you could move this content to the yml files as something like nj_data_transfer_offboarding_links (or split it up as needed) and pass the vita link as a param?

mluedke2: @mmazanec22 I moved the content to yml as you suggested and I do think I like it better now. Thoughts?

app/views/state_file/questions/data_transfer_offboarding/edit.html.erb
```diff
@@ -13,13 +13,12 @@
...
   <% end %>
 
-  <% vita_link = I18n.locale == :en ? StateFile::StateInformationService.vita_link_en(current_state_code) : StateFile::StateInformationService.vita_link_es(current_state_code) %>
-  <% if vita_link.present? %>
-    <p><%= link_to t("state_file.questions.eligible.faq_link"), state_faq_path(us_state: current_state_code) %></p>
-    <%= link_to(vita_link, class: "button button--primary button--wide spacing-below-15", target: "_blank", rel: "noopener noreferrer") do %>
-      <%= t('state_file.questions.eligible.vita_option.connect_to_vita') %>
-    <% end %>
-  <% end %>
+  <% # i18n-tasks-use t('state_file.questions.eligible.vita_option.connect_to_vita.nj') # hint for the i18n linter that, yes, we are using this key (sometimes) %>
```
embarnard, line 16: @mluedke2 is `i18n-tasks-use` a command? Sorry I'm confused about this comment, could you help me understand?

mluedke2: it's a linter pattern i copied from elsewhere in the app (if you text search for the comment you will see many of them throughout). without it, the dynamic key will fail the automated i18n health checks because it won't know the key was used.

embarnard: Oooooh I see right because there will be a failure in the i18n health checkup if we are not using a key. I hadn't realized we had moved away from adding it to the ignore file to mentioning it as a comment but you're right I do see it elsewhere in the app.

embarnard (approved):

PR #5642, opened 2025-02-26
[GYR1-636] Reenable pending spec tests in f13614c_pdf_spec.rb

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-636

## Is PM acceptance required?

## What was done?

Reënabled four spec tests that got turned off while we updated code for the 
ty2024 updates.

### 1) it "can successfully write everything that comes out of #hash_for_pdf to 
the PDF"

Originally this test checked that the filler class's fields matched exactly the 
fields in a generated PDF. But with our current approach (ty2024), the filler 
class is not doing anything with fields that won't get filled in, so this test 
won't quite work as-is. So, I updated the logic slightly to ensure that at least 
the intersection of fields between the two aforementioned sources is equal to 
the generated pdf's fields (i.e., we're not trying to fill in any fields that 
don't exist).

In the process, it was necessary to clear out all out-of-date field mappings 
that still lingered in the filler class.

### 2) it "fills out answers from the DB into the pdf"

Replaced all the old PDF field keys with keys from ty2024 pdf.

### 3) testing gated questions/values

We don't use the 'gated' field mechanism anymore.

- I removed this spec test.
- I also removed the `fetch_gated_value` from the filler class.
- Note: I left the `GATES` data structure in place at the top fo the filler 
  class. Why? It's still used by `VitaMinFormBuilderFor13614` for some very 
  bespoke logic around setting default values in maybe one or a handful of form 
  elements in the Hub editable 14-C (otherwise, I would've deleted `GATES` too).

### 4) it "15080 fields are nil" do

Small changes were needed to update this. Note that I made the related updates 
in the filler class's `vita_consent_to_disclose_info` method too; however, to 
the best of my knowledge, this is not currently used.

## How to test?

- Run the spec tests :O


app/lib/pdf_filler/f13614c_pdf.rb
```diff
@@ -141,26 +138,6 @@ def hash_for_pdf
 
       answers["form1[0].page2[0].receivedMoneyFrom[0].howManyJobs[0]"] = @intake.job_count.to_s
 
-      # PAGE 2: INCOME
-      answers.merge!(
-        yes_no_checkboxes("form1[0].page2[0].Part3[0].q3Scholarships[0]", @intake.had_scholarships, include_unsure: true),
```
embarnard, line 146: why were these removed? Are they outdated from previous years 13614c?

embarnard (approved):
This was a lot of work to add all these new lines into the spec! Good job, approved

app/lib/pdf_filler/f13614c_pdf.rb
spompea-cfa: @embarnard yeah, these are holdovers from the ty2023 form. one of the spec tests howled about them when i turned it back on.

PR #5643, opened 2025-02-26
[FYST-1583] ID 1099R Subtraction Card to Final Screen & Review/Edit feature

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1583

## What was done?
- Added a card on final review for Qualified Retirement Income Deduction (show if 8e.positive, but show 8f value)
- Added functionality to return to one of two retirement-related controllers, depending on the age of the filer(s) -- if any one of them are 62-65, we take them to the `IdDisabilityController`, otherwise to the `IdRetirementAndPensionController`
- Added concept of `review_all_items_before_returning_to_review` for a `RepeatedQuestionConcern` -- currently we "review and edit" into a single item in a repeated collection, and then return to final review. For this use case, we need to go through all eligible 1099Rs (via `IdRetirementAndPensionController`) and _then_ return to final review
- Moved look up of `eligible_1099rs` to intake level from the controller since they are shared in two controllers in slightly different ways
- See `#clean_up_followups` method in `IdDisabilityForm`; before we were not destroying followups if client answered "no" for disability (if they're 62-65, their 1099R amount is no longer eligible for deduction). Without this fix, there could be a bug where if we select "No" and proceed, the 1099R amount will still be seen as an eligible amount to sum up.
## How to test?
  - Lots of tests
  - Test data used:
    - Kingsleigh: Primary filer over 65
    - Potato_mfj: primary over 65 & spouse 62-65 (63 yo)
    - Pangolin: primary 62-65, spouse under 62
- Risk Assessment
  - There are some usability concerns both Martha & I have about this, one of which is that when you click Review & Edit, you have to go through all the eligible 1099Rs again (since this number can change based on disability answer; also even if it doesnt for 65+ year olds, they still have to revisit all 1099Rs), which is a departure from other editing income functionalities.
Screenshot 2025-02-27 at 11 23 26 AM

### DIFF
--- a/app/controllers/concerns/state_file/repeated_question_concern.rb
+++ b/app/controllers/concerns/state_file/repeated_question_concern.rb
@@ -22,13 +22,16 @@ module StateFile
     end
 
     def next_path
-      return super if params[:return_to_review].present?
+      return super if params[:return_to_review].present? && !review_all_items_before_returning_to_review
 
       next_index = current_index + 1
       if next_index >= num_items
         super
       else
         options = {index: next_index}
+        if params[:return_to_review].present? && review_all_items_before_returning_to_review
+          options[:return_to_review] = params[:return_to_review]
+        end
         self.class.to_path_helper(options)
       end
     end
@@ -50,5 +53,9 @@ module StateFile
       raise NotImplementedError
     end
 
+    def review_all_items_before_returning_to_review
+      # define in the controller if needed review all items instead of just one
+      false
+    end
   end
 end
--- a/app/controllers/state_file/questions/id_disability_controller.rb
+++ b/app/controllers/state_file/questions/id_disability_controller.rb
@@ -23,7 +23,7 @@ module StateFile
           if @eligible_1099rs.any?
             StateFile::Questions::IdRetirementAndPensionIncomeController.to_path_helper(return_to_review: params[:return_to_review])
           else
-            StateFile::Questions::IdReviewController.to_path_helper(return_to_review: params[:return_to_review])
+            StateFile::Questions::IdReviewController.to_path_helper
           end
         else
           super
--- a/app/controllers/state_file/questions/id_disability_controller.rb
+++ b/app/controllers/state_file/questions/id_disability_controller.rb
@@ -4,17 +4,7 @@ module StateFile
       def self.show?(intake)
         Flipper.enabled?(:show_retirement_ui) &&
           intake.state_file1099_rs.any? { |form1099r| form1099r.taxable_amount&.to_f&.positive? } &&
-          !intake.filing_status_mfs? && meets_age_requirements?(intake)
-      end
-
-      def self.meets_age_requirements?(intake)
-        primary_age = intake.calculate_age(intake.primary_birth_date, inclusive_of_jan_1: true)
-        if intake.filing_status_mfj? && intake.spouse_birth_date.present?
-          spouse_age = intake.calculate_age(intake.spouse_birth_date, inclusive_of_jan_1: true)
-          (primary_age >= 62 && primary_age < 65) || (spouse_age >= 62 && spouse_age < 65)
-        else
-          primary_age >= 62 && primary_age < 65
-        end
+          !intake.filing_status_mfs? && intake.meets_age_requirements_for_disability_check?
       end
 
       def next_path
--- a/app/controllers/state_file/questions/id_disability_controller.rb
+++ b/app/controllers/state_file/questions/id_disability_controller.rb
@@ -17,6 +17,27 @@ module StateFile
         end
       end
 
+      def next_path
+        has_disability_in_household =
+          if current_intake.filing_status_mfj?
+            form_params[:primary_disabled] == "yes" || form_params[:spouse_disabled] == "yes"
+          else
+            form_params[:primary_disabled] == "yes"
+          end
+
+        has_over_65_senior_in_household = current_intake.primary_senior? || current_intake.spouse_senior?
+
+        if params[:return_to_review] == "y"
+          if has_over_65_senior_in_household || has_disability_in_household
+            StateFile::Questions::IdRetirementAndPensionIncomeController.to_path_helper(return_to_review: params[:return_to_review])
+          else
+            StateFile::Questions::IdReviewController.to_path_helper(return_to_review: params[:return_to_review])
+          end
+        else
+          super
+        end
+      end
+
       private
 
       def form_params
--- a/app/controllers/state_file/questions/id_disability_controller.rb
+++ b/app/controllers/state_file/questions/id_disability_controller.rb
@@ -18,17 +18,8 @@ module StateFile
       end
 
       def next_path
-        has_disability_in_household =
-          if current_intake.filing_status_mfj?
-            form_params[:primary_disabled] == "yes" || form_params[:spouse_disabled] == "yes"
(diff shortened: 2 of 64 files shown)


spec/features/state_file/review_page_spec.rb
```diff
@@ -335,6 +335,102 @@
...
+
+            click_on I18n.t("general.review_and_edit")
+          end
+
+          expect(page).to have_text I18n.t("state_file.questions.id_disability.edit.title")
+          choose "No"
+          click_on I18n.t("general.continue")
+
+          expect(page).to have_text I18n.t("state_file.questions.shared.abstract_review_header.title")
+          expect(page).not_to have_text I18n.t("state_file.questions.id_review.edit.qualified_retirement_benefits_deduction")
```
arinchoi03, line 420: This spec tests the destroying of followups in `IdDisabilityForm` if the user chooses to select "No" for disability. They already have a followup (created at the top of the spec) which should be destroyed -- leading to the calculations for the review card to **not** sum up the value of the 1099Rs (both are primary's)

app/forms/state_file/id_disability_form.rb
```diff
@@ -27,10 +27,24 @@ def save
...
+      primary_followups =  @intake.filer_1099_rs(:primary).map(&:state_specific_followup).compact
+      if primary_disabled == "no" || ["spouse", "none"].include?(mfj_disability)
+        primary_followups.each(&:destroy)
+      end
+
+      if @intake.filing_status_mfj? && ["me", "none"].include?(mfj_disability)
+        spouse_followups =  @intake.filer_1099_rs(:spouse).map(&:state_specific_followup).compact
+        spouse_followups.each(&:destroy)
+      end
+    end
```
arinchoi03, line 46: previously, we kept the followups around if they submitted that they were no longer disabled. The followup only contains what they answer on the next controller (`IdRetirementAndPensionController` and whether `eligible_income_source`). If we do not update the `eligible_income_source` to at least `"unfilled"` or `"no"` the calculator will still consider the 1099R amount as an eligible income (see `calculate_sec_b_line_8e` in ID39R calculator)

app/models/state_file_id_intake.rb
```diff
@@ -118,4 +118,26 @@ def has_blind_filer?
...
   end
+
+
+  def eligible_1099rs
+    state_file1099_rs.select do |form1099r|
+      form1099r.taxable_amount&.to_f&.positive? && person_qualifies?(form1099r)
+    end
+  end
+
+  private
```
arinchoi03, line 129: add line here

app/views/state_file/questions/id_review/edit.html.erb
```diff
@@ -2,6 +2,20 @@
...
 
+  <% if Flipper.enabled?(:show_retirement_ui) && current_intake.calculator.line_or_zero(:ID39R_B_LINE_8e).positive? %>
+    <div id="qualified-retirement-benefits-deduction" class="white-group">
+      <div class="spacing-below-5">
+        <p class="text--bold spacing-below-5"><%= t(".qualified_retirement_benefits_deduction") %></p>
+        <p class="text spacing-below-2"><%= t(".qualified_retirement_benefits_deduction_explain") %></p>
+
+        <p class="text--bold spacing-below-5"><%= t(".qualified_disabled_retirement_benefits") %></p>
+        <p><%= number_to_currency(current_intake.calculator.line_or_zero(:ID39R_B_LINE_8f), precision: 2) %></p>
+        <% id_disability_or_retirement_link = StateFile::Questions::IdDisabilityController.meets_age_requirements?(current_intake) ?  StateFile::Questions::IdDisabilityController.to_path_helper(return_to_review: "y") : StateFile::Questions::IdRetirementAndPensionIncomeController.to_path_helper(return_to_review: "y", index: 0) %>
```
arinchoi03, line 13: if they see this card, they initially entered the IdDisability/IdRetirementAndPensionIncomeController.
The logic to return to the edit screen is much simpler, we just ask if they are between 62-65 (`meets_age_requirements?`), and if so - go to disability controller, or go straight to IdRetirementAndPensionIncomeController (for over 65+)

app/forms/state_file/id_disability_form.rb
```diff
@@ -27,10 +27,24 @@ def save
...
+
     def eligible?(primary_or_spouse)
```
arinchoi03, line 48: I have an open question about the `eligible?` method, it looks like we don't update `_disabled` attribute for primary/spouse unless they're `age_eligible`. Not sure why this and `has_taxable_1099r` has anything to do with whether a person is disabled or not?

app/views/state_file/questions/id_review/edit.html.erb
arinchoi03, line 13: `meets_age_requirements` can maybe move to a shared place?

app/controllers/state_file/questions/id_disability_controller.rb
```diff
@@ -17,6 +17,19 @@ def self.meets_age_requirements?(intake)
         end
       end
 
+      def next_path
+        if params[:return_to_review] == "y"
+          @eligible_1099rs ||= current_intake.eligible_1099rs
+          if @eligible_1099rs.any?
+            StateFile::Questions::IdRetirementAndPensionIncomeController.to_path_helper(return_to_review: params[:return_to_review])
+          else
+            StateFile::Questions::IdReviewController.to_path_helper(return_to_review: params[:return_to_review])
```
arinchoi03, line 26: remove params[:return_to_review], it's useless

app/controllers/concerns/state_file/repeated_question_concern.rb
```diff
@@ -22,13 +22,16 @@ def prev_path
     end
 
     def next_path
-      return super if params[:return_to_review].present?
+      return super if params[:return_to_review].present? && !review_all_items_before_returning_to_review
```
arinchoi03, line 25: this is the default action that happens if `return_to_review` is present except for this `review_all_items_before` case

app/forms/state_file/id_disability_form.rb
arinchoi03: File a bug for this!

mpidcock (approved):
This seems good to me and I don't want to hold things up over my usability concerns. If there are adjustments needed we can follow up. (Also it will be easier to demonstrate the complexity with it implemented.)

arinchoi03:
@mpidcock one weird usability thing I just noticed
For pangolin (filer btwen 62-65 & spouse under 62), we ask them about disability, and if we enter "Yes, both are", we end up at final review with this card about retirement benefits. If we click "review & edit" and answer "Yes, my spouse is" (filer 62-65 no longer eligible; spouse always not eligible b/c they're <62) they return to the reivew page and the card is just ✨ gone ✨ . I mean, it's working as intended, but weird b/c they can't nav back to change the answers anymore.

mpidcock:
@arinchoi03 You make a good point - although I have no idea what would be better. Can you comment this in the ticket? (and also follow up about your previous comments) I don't think we'll be able to fix this soon, but it will be good to warn CS and perhaps plan out a reveal that can explain the logic and calculations better. if/when we start hearing from CS about confusion we will have more info for how to improve and iterate

mpidcock (approved):
🚢

PR #5649, opened 2025-02-26
[FYST-1873] ID 1099R should still save their disability status as answered even though ineligible

arinchoi03:
Fast follow to: https://github.com/codeforamerica/vita-min/pull/5643 to separate out the concern

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1873

## What was done?
- Save the `_disabled` attributes as it was answered, instead of only setting it to "yes" if the filer is eligible (dependent on age/income). This is for data integrity, the eligibility of the filer does not have anything to do with disability. Thankfully these attributes are not used for anything other than to determine which 1099Rs are eligible (`eligible_1099rs` method on `StateFileIdIntake`)
## How to test?
  - Removed test set up that makes the filer "eligible"
- Risk Assessment
  - Audited that the disabled attributes are not being used elsewhere
  - Chatted with Margaret to confirm we're not using that value except to progress through the 1099R followup flow

### DIFF
--- a/app/forms/state_file/id_disability_form.rb
+++ b/app/forms/state_file/id_disability_form.rb
@@ -27,9 +27,23 @@ module StateFile
       else
         @intake.update(attributes_for(:intake))
       end
+
+      clean_up_followups
     end
 
     private
+
+    def clean_up_followups
+      primary_followups =  @intake.filer_1099_rs(:primary).map(&:state_specific_followup).compact
+      if primary_disabled == "no" || ["spouse", "none"].include?(mfj_disability)
+        primary_followups.each(&:destroy)
+      end
+
+      if @intake.filing_status_mfj? && ["me", "none"].include?(mfj_disability)
+        spouse_followups =  @intake.filer_1099_rs(:spouse).map(&:state_specific_followup).compact
+        spouse_followups.each(&:destroy)
+      end
+    end
 
     def eligible?(primary_or_spouse)
       person = intake.send(primary_or_spouse)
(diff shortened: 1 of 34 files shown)


app/forms/state_file/id_disability_form.rb
```diff
@@ -8,19 +8,13 @@ class IdDisabilityForm < QuestionsForm
 
     def save
       if intake.filing_status_mfj?
-        primary_eligible = eligible?(:primary)
-        spouse_eligible = eligible?(:spouse)
-
         case mfj_disability
         when "me"
-          @intake.update(primary_disabled: primary_eligible ? "yes" : "no", spouse_disabled: "no")
+          @intake.update(primary_disabled: "yes", spouse_disabled: "no")
```
anisharamnani, line 13: [dust] the dustiest of dust, but we renamed this to `primary` for maryland for clarity & since it's not actually "me" but that doesn't super matter.

embarnard (approved):

anisharamnani:
possibly blocking comment: i believe that the pension exclusion show method needs to be updated since the eligibility check is being removed from this form, that logic needs to appear in the `app/controllers/state_file/questions/id_retirement_and_pension_income_controller.rb`

anisharamnani:
update: @arinchoi03 explained that the `id_disability_controller.rb` will only be shown if the filer is between 62 and 65, so please disregard my comment!

anisharamnani (approved):

arinchoi03:
@mpidcock i'm going to hold off on merging this until the base PR review is done!

app/forms/state_file/id_disability_form.rb
arinchoi03: ooh i'm going to try to change it

PR #5656, opened 2025-02-27
NJ 148 - Add retirement warning screen

jachan:
## Link to pivotal/JIRA issue
- e.g. https://github.com/newjersey/affordability-pm/issues/148#issue-2630021954

## What was done?
- In NJ, we want taxpayers to acknowledge that their retirement exclusion may be less if they use our tool. We set a heuristic to capture a superset of taxpayers who will be affected and show them the retirement warning screen.
- Heuristic logic is in the ticket
- When we show the user the retirement income warning page, they can select whether they want to continue or not. If they select that they don't want to continue, we send them to an offboarding screen, otherwise we send them to the next page in the flow.

## How to test?
  - Ensure that Flipper(:show_retirement_ui) is ON, otherwise the Retirement Warning screen will not show up. 
  - Testing with localhost, use Giudice persona, validate that the Retirement Warning screen is shown
     - Logic: total_box_1_1099 income is 160_000, which is greater than the single threshold of 75_000
  - Validate that tool offboards users who select "No"
  - Validate that tool allows users to continue if they select "Yes, continue"

- Testing with localhost, use Manzos persona, validate that the Retirement Warning screen is not shown
    - Logic: total_box_1_1099 income is 44_000, which is less than the MFJ threshold of 100_000.

- Risk Assessment
  - If this page is shown to too many people, we risk losing users as they choose to offboard.
 
- Before
- After
image

Offboarding
image

### DIFF
--- a/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
@@ -4,8 +4,8 @@ module StateFile
       include EligibilityOffboardingConcern
 
       def self.show?(intake)
-        line_15 = intake.calculator.line_or_zero(:NJ_1040_LINE_15)
-        line_16a = intake.calculator.line_or_zero(:NJ_1040_LINE_16A)
+        line_15 = intake.calculator.line_or_zero(:NJ1040_LINE_15)
+        line_16a = intake.calculator.line_or_zero(:NJ1040_LINE_16A)
         nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
         nj_retirement_income_helper.show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
       end
--- a/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
@@ -1,0 +1,15 @@
+module StateFile
+  module Questions
+    class NjRetirementWarningController < QuestionsController
+      include EligibilityOffboardingConcern
+
+      def self.show?(intake)
+        line_14 = intake.calculator.line_or_zero(:NJ_1040_LINE_14)
+        line_15 = intake.calculator.line_or_zero(:NJ_1040_LINE_15)
+        line_16a = intake.calculator.line_or_zero(:NJ_1040_LINE_16A)
+        nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
+        nj_retirement_income_helper.show_retirement_income_warning?(line_14, line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+      end
+    end
+  end
+end
--- a/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
@@ -4,11 +4,10 @@ module StateFile
       include EligibilityOffboardingConcern
 
       def self.show?(intake)
-        line_14 = intake.calculator.line_or_zero(:NJ_1040_LINE_14)
         line_15 = intake.calculator.line_or_zero(:NJ_1040_LINE_15)
         line_16a = intake.calculator.line_or_zero(:NJ_1040_LINE_16A)
         nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
-        nj_retirement_income_helper.show_retirement_income_warning?(line_14, line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+        nj_retirement_income_helper.show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
       end
     end
   end
--- a/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_warning_controller.rb
@@ -6,8 +6,11 @@ module StateFile
       def self.show?(intake)
         line_15 = intake.calculator.line_or_zero(:NJ1040_LINE_15)
         line_16a = intake.calculator.line_or_zero(:NJ1040_LINE_16A)
-        nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
-        nj_retirement_income_helper.show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+        retirement_helper(intake).show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+      end
+
+      def self.retirement_helper(intake)
+        @retirement_helper ||= Efile::Nj::NjRetirementIncomeHelper.new(intake)
       end
     end
   end
--- a/app/lib/efile/nj/nj_retirement_income_helper.rb
+++ b/app/lib/efile/nj/nj_retirement_income_helper.rb
@@ -10,7 +10,7 @@ module Efile
         end
       end
 
-      def show_retirement_income_warning?(line_14, line_15, line_16a)
+      def show_retirement_income_warning?(line_15, line_16a)
         if non_military_1099r_box_1_total.zero?
           return false
         end
@@ -27,7 +27,7 @@ module Efile
           return false
         end 
 
-        if non_military_1099r_box_1_total <= max_exclusion_threshold && (line_14 > 3_000 || all_filers_under_62?)
+        if non_military_1099r_box_1_total <= max_exclusion_threshold && (line_15 > 3_000 || all_filers_under_62?)
           return false
         end
 
--- a/app/lib/efile/nj/nj_retirement_income_helper.rb
+++ b/app/lib/efile/nj/nj_retirement_income_helper.rb
@@ -10,6 +10,38 @@ module Efile
         end
       end
 
+      def show_retirement_income_warning?(line_14, line_15, line_16a)
+        if non_military_1099r_box_1_total.zero?
+          return false
+        end
+
+        box_1_gross_income = non_military_1099r_box_1_total + line_15 + line_16a
+        max_exclusion_threshold = if @intake.filing_status_mfs? 
+                                    50_000
+                                  elsif @intake.filing_status_mfj?
+                                    100_000
+                                  else
+                                    75_000
+                                  end
+        if box_1_gross_income <= max_exclusion_threshold
+          return false
+        end 
+
+        if non_military_1099r_box_1_total <= max_exclusion_threshold && (line_14 > 3_000 || all_filers_under_62?)
+          return false
+        end
+
+        true
+      end
+
+      def non_military_1099r_box_1_total
+        box_1_totals = 0
+        @non_military_1099rs.each do |non_military_1099r|
+          box_1_totals += non_military_1099r.gross_distribution_amount.round
+        end
+        box_1_totals
+      end
+
       def total_eligible_nonmilitary_1099r_income
         total_eligible_income = 0
         eligible_ssns = []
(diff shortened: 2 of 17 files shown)


spec/forms/state_file/nj_retirement_warning_form_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it "requires radio answer" do
+      form = described_class.new(intake, invalid_params)
+      form.valid?
+
+      expect(form.errors[:eligibility_retirement_warning_continue]).to include "Can't be blank."
+    end
+  end
+
+  describe "#save" do
+    context "when all tax household members had health insurance" do
```
mluedke2, line 23: [pebble] this and the below contexts are for a different form. maybe `when taxpayer chooses to continue` and `when taxpayer chooses not to continue`?

db/migrate/20250227182343_add_retirement_warning_continue_confirmation.rb
```diff
@@ -0,0 +1,5 @@
+class AddRetirementWarningContinueConfirmation < ActiveRecord::Migration[7.1]
+  def change
+    add_column :state_file_nj_intakes, :eligibility_retirement_warning_continue, :integer, default: 0
+  end
+end
```
mluedke2, line 5: [potential boulder] - is the user's choice whether to continue or not really something that we want to save in the DB and on the model? or should it merely affect which screen they see after the warning (offboarding or not), and not be saved in state?

I may be unaware of a secondary usage of the choice as a data point, but if it's only for navigation then might we opt for the lighter touch here?

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe StateFile::Questions::NjRetirementWarningController do
+  describe ".show?" do
+    let (:intake) { create :state_file_nj_intake }
+    before do
+      allow_any_instance_of(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
```
mluedke2, line 7: [pebble] i know it's tedious but we should have the tests for the flag being off too... the price for using flags in production

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -10,6 +10,38 @@ def initialize(intake)
...
+        true
+      end
+
+      def non_military_1099r_box_1_total
+        box_1_totals = 0
+        @non_military_1099rs.each do |non_military_1099r|
+          box_1_totals += non_military_1099r.gross_distribution_amount.round
+        end
+        box_1_totals
+      end
```
mluedke2, line 43: [pebble] this can be simplified to

```
def non_military_1099r_box_1_total
  @non_military_1099rs.sum { |non_military_1099r| non_military_1099r.gross_distribution_amount.round }
end
```

app/controllers/state_file/questions/nj_retirement_warning_controller.rb
```diff
@@ -0,0 +1,15 @@
+module StateFile
+  module Questions
+    class NjRetirementWarningController < QuestionsController
+      include EligibilityOffboardingConcern
+
+      def self.show?(intake)
+        line_14 = intake.calculator.line_or_zero(:NJ_1040_LINE_14)
+        line_15 = intake.calculator.line_or_zero(:NJ_1040_LINE_15)
+        line_16a = intake.calculator.line_or_zero(:NJ_1040_LINE_16A)
```
mluedke2, line 9: [dust] - i think this is fine, but noting that i think it's a new patten to use `intake.calculator.line_or_zero` in the controller for a question, right? but maybe that's just because this task is unique in its dependency on calculations?

spec/factories/state_file_nj_intakes.rb
```diff
@@ -283,6 +284,18 @@
...
+      raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_giudice_hoh_1099_r') }
+      raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_giudice_hoh_1099_r') }
+      filing_status { "head_of_household" }
+    end
+
+    trait :df_data_napolitano do
+      raw_direct_file_data { StateFile::DirectFileApiResponseSampleService.new.read_xml('nj_napolitano_single_1099_r') }
+      raw_direct_file_intake_data { StateFile::DirectFileApiResponseSampleService.new.read_json('nj_napolitano_single_1099_r') }
+      filing_status { "single" }
+    end
```
aloverso, line 297: [dust/pebble] are these being used anywhere? I didn't see it but I might have missed it. If not being used, can we remove?

spec/lib/efile/nj/nj_retirement_income_helper_spec.rb
```diff
@@ -250,14 +250,100 @@
...
+      { traits: [:head_of_household], box_1_total: 0, expected: false },
+      { traits: [:qualifying_widow], box_1_total: 0, expected: false },
+      { traits: [:married_filing_jointly], box_1_total: 0, expected: false },
+      { traits: [:married_filing_separately], box_1_total: 0, expected: false },
+
+      { traits: [:single], box_1_total: 25_000, line_15: 25_000, line_16a: 24_999, expected: false },
+      { traits: [:head_of_household], box_1_total: 25_000, line_15: 25_000, line_16a: 24_999, expected: false },
+      { traits: [:qualifying_widow], box_1_total: 25_000, line_15: 25_000, line_16a: 24_999, expected: false },
+      { traits: [:married_filing_jointly], box_1_total: 33_333, line_15: 33_333, line_16a: 33_333, expected: false },
+      { traits: [:married_filing_separately], box_1_total: 20_000, line_15: 20_000, line_16a: 9_999, expected: false },
```
aloverso, line 265: [dust] An idea I have here to make this test behave more as tests-as-documentation could be to break these up into a couple different describe blocks, to each describe their own case for why we expect the result.

I'm thinking something like:
- "when sum of box_1 totals is 0" (includes first 5 cases here)
- "when box_1_total + line_15 + line_16a is less than max_exclusion for filing status"
- "when box_14 is over 3000"
- ... and so on

that way it emphasizes in text what the important part of the test setup is, and helps describe to someone in words what leads to the actual final output. Right now, this is an effective test for our confidence levels, but doesn't help someone who's new to the code read it and understand what's going on or why.

On this note, might be a good followup refactor for us on the other lengthy test-case blocks in this file.

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -60,11 +92,18 @@ def spouse_62_and_older?
...
         elsif !primary_62_and_older?
+          return true
+        end
+        false
+      end
+
+      def line_28b_eligible?(line_15, line_27, line_28a)
+        if all_filers_under_62?
           return false
         end
```
aloverso, line 108: [dust] for cleanliness I like the pattern on the line below `return false if ...` instead of the three-line if statement here

db/migrate/20250227182343_add_retirement_warning_continue_confirmation.rb
jachan: It's purely for navigation, so your point is well-taken! I was under the impression that eligibility checks needed to reference the model (following the `eligibility_health_insurance` model), is there a lighter touch way here to determine which option the user has selected?

app/controllers/state_file/questions/nj_retirement_warning_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+        nj_retirement_income_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
```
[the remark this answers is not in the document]
jachan: Thanks. Had mentally flagged this for a refactor after getting tests green, but forgot to do it in the rush to get this out before EOD.

db/migrate/20250227182343_add_retirement_warning_continue_confirmation.rb
jachan: Talked with PMs about this, and it's important to them to track # of filers seeing this page, which is an argument in favor of writing this to the model.

spec/factories/state_file_nj_intakes.rb
jachan: unused, removed.

jachan:
Had to update the logic for Show Retirement Warning - PMs were testing and realized that the requirements had a typo in them! The Retirement Warning check no longer depends on line 14, it instead depends on line 15 and line 16a. While I was refactoring the tests, I went ahead and broke them out into sections per @aloverso's suggestion. They should be much more readable.

arinchoi03 (commented):
some comments 🙏

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -60,13 +88,18 @@ def spouse_62_and_older?
         @intake.calculate_age(@intake.spouse_birth_date, inclusive_of_jan_1: false) >= 62 
       end
 
-      def line_28b_eligible?(line_15, line_27, line_28a)
+      def all_filers_under_62?
         if @intake.spouse_birth_date.present?
           spouse_or_primary_is_age_eligible = spouse_62_and_older? || primary_62_and_older?
-          return false unless spouse_or_primary_is_age_eligible
+          return true unless spouse_or_primary_is_age_eligible
```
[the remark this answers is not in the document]
jachan: Appreciate the callout, a good thing to check!!

Gabriel confirmed that MFS filers effectively do not have a spouse for retirement exclusion eligibility purposes. So this method should work as written.

app/lib/efile/nj/nj_retirement_income_helper.rb
```diff
@@ -10,6 +10,34 @@ def initialize(intake)
...
+
+        if non_military_1099r_box_1_total <= max_exclusion_threshold && (line_14 > 3_000 || all_filers_under_62?)
+          return false
+        end
+
+        true
+      end
+
+      def non_military_1099r_box_1_total
+        @non_military_1099rs.sum { |non_military_1099r| non_military_1099r.gross_distribution_amount.round }
```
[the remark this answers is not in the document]
jachan: Based on what I'm seeing in `app/lib/submission_builder/state1099_r.rb` I assume no? Most other fields that can be `nil` have `.present?` but checking the schema for the source of truth here.

```
          xml.PayerEIN form1099r.payer_identification_number
          xml.RecipientSSN sanitize_for_xml(form1099r.recipient_ssn) if form1099r.recipient_ssn.present?
          xml.RecipientNm sanitize_for_xml(form1099r.recipient_name) if form1099r.recipient_name.present?
          xml.GrossDistributionAmt form1099r.gross_distribution_amount&.round
```

app/controllers/state_file/questions/nj_retirement_warning_controller.rb
jachan: HUGE callout on the incorrect line keys. A THOUSAND thank yous for catching this. Fixed in the latest commit, cannot believe I missed that...

I'm going to vote we do a follow up refactor to move this logic out of the calculator. DirectFile needs this to be launched by Tuesday so I want to merge this logic in, as long as it's correct.

app/lib/efile/nj/nj_retirement_income_helper.rb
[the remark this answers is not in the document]
jachan: > but I’m also personally willing to say it’s never correct for box 1 to be missing (you can’t have a retirement distribution without a distribution!), and so if that errored out i’d be fine with it bc other stuff would be wrong

Can't find a schema for this...Talked with Gabriel, the consensus here is that it doesn't make sense for `gross_distribution_amount` to be nil in any case.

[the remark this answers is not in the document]
arinchoi03: the `minOccurs` is `0` 
```
			<xsd:element name="GrossDistributionAmt" type="USAmountType" minOccurs="0">
				<xsd:annotation>
					<xsd:documentation>
						<Description>Gross distribution amount</Description>
						<LineNumber>1</LineNumber>
						<ELFFieldNumber>0110</ELFFieldNumber>
					</xsd:documentation>
				</xsd:annotation>
			</xsd:element>
```
I guess I'd say, it doesn't hurt to guard against the `nil` anyways but maybe we _want_ things to break in this instance? (which it will anyways unless we fix the 1099R)

No need to fix, but i'll make a note to look into these kinds of cases overall across the app.

db/migrate/20250227182343_add_retirement_warning_continue_confirmation.rb
mluedke2: works for me- thanks for verifying that

arinchoi03 (approved):
ty for working through the comments so quickly John!

mluedke2 (approved):

PR #5664, opened 2025-02-28
[FYST-1882] Update ID disability logic to only ask questions about filers who fall within 62-65 age range

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1882
## Is PM acceptance required? (delete one)
- Yes - don't merge until JIRA issue is accepted!

**Reminder**: merge main into this branch and get green tests before merging to main
## What was done?
- Only showing the four options view "Yes, I am", "Yes, my spouse is", "Yes, we both are", and "No, neither of us is" if _both_ filers are between 62-65. Previously, we showed it for any mfj filer who had eligible 1099Rs
- added a bunch of `_between_62_and_65_years_old?` methods. Seems a bit overkill but they all do get used 🤷 shorter naming suggestions will be welcome

## How to test?
- Describe the testing approach taken to verify the changes, including:
  - Unit/integration/manual tests
  - Test data used:
    -  Pangolin: primary between 62-65; spouse filer under 62
    - Potato MFJ: primary 65+; only spouse between 62-65
    - Charlie Swann: Single filer, 62-65 yo
    - **TBD** MFJ persona that has two filers both between 62-65 yo
- Specify any relevant testing environments used (e.g., development, staging, demo, Heroku).
- Risk Assessment
  - This was a quick change (quick request, quick implementation)
## Screenshots (for visual changes)
Potato MFJ (only shows q about spouse):
Screenshot 2025-02-28 at 3 28 12 PM

Pangolin (only shows q about primary):
Screenshot 2025-02-28 at 3 29 37 PM

Charlie Swan (only q about primary):
Screenshot 2025-02-28 at 3 30 52 PM

Barrel Roll (both filers are 62-65): 
Screenshot 2025-03-03 at 8 59 35 AM

PR #5665, opened 2025-03-03
[FYST-1868] ID refactor logic

arinchoi03:
## Link to pivotal/JIRA issue
- followup to: https://github.com/codeforamerica/vita-min/pull/5664
- will address: https://codeforamerica.atlassian.net/browse/FYST-1868

## What was done?
- Mainly just refactoring based on convo here: https://github.com/codeforamerica/vita-min/pull/5664/files#r1976948511
  - sum up only `eligible_1099rs` income for ID39R calculations
  - only `show` retirement subtractions controllers for all states if `eligible_1099rs` exist for the intake instead of if any 1099rs are present. notably for nc currently, there is no `eligible_1099rs` method (it uses the base method, which looks at all 1099rs), so the logic should still be the same.
    - may need to talk to Hugo about https://codeforamerica.atlassian.net/browse/FYST-1868, as that story may just involve creating `eligible_1099rs` method in the intake as well.

## How to test?
- Same run through for ID & that the _show_ logic works for all states (AZ/ID/NC/MD) as currently expected.
  - make sure line 8e calculates as expected with this change (should be same as before) 
- Risk Assessment
  - the factory was changed to set filer to `disabled` to make the 1099R income eligible for id intakes, maybe this was a bad idea?


app/lib/efile/id/id39_r_calculator.rb
```diff
@@ -85,7 +85,7 @@ def calculate_sec_b_line_8d
       end
 
       def calculate_sec_b_line_8e
-        @intake.state_file1099_rs.sum do |form1099r|
+        @intake.eligible_1099rs.sum do |form1099r|
           if form1099r.state_specific_followup&.eligible_income_source_yes? && form1099r.taxable_amount.present?
```
arinchoi03, line 89: presumably with this refactor, `taxable_amount` will be always present already

spec/models/state_file_base_intake_spec.rb
```diff
@@ -313,4 +313,15 @@
     end
   end
 
+  describe "#eligible_1099rs" do
+    %w[az md nc nj].each do |state_code|
```
arinchoi03, line 317: although nj is not using this method, I wanted to make sure it was documented that they will have it

arinchoi03:
closing this, in favor of:
1. ID refactor: https://github.com/codeforamerica/vita-min/pull/5681
2. eligible_1099rs refactor: https://github.com/codeforamerica/vita-min/pull/5680

PR #5667, opened 2025-03-03
Fyst 1853/money changes 1099 r g

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1853

## What was done?
- Adds presence check for the following money amounts:
  - 1099-R
    - Box 14, State tax withheld
    - Box 15, State distribution
  - 1099-G
    - Box 4, Federal income tax withheld
    - Box 11, State income tax withheld

Notably, I did _not_ add validation for 0 values for unemployment on 1099-G because there already was a validation for 1 dollar, which made more sense to me. Reaching out to Tiffany to make sure this is correct and will change if necessary.

## How to test?
- Tests are provided
- Step through the flow and test the 1099-G and 1099-R pages to make sure the values have the correct information.

- Before
- After
- 1099-R
image

- 1099-G
image

### DIFF
--- a/app/models/state_file1099_g.rb
+++ b/app/models/state_file1099_g.rb
@@ -53,18 +53,21 @@ class StateFile1099G < ApplicationRecord
   validates :recipient_zip, zip_code: { zip_code_lengths: [5, 9, 12].freeze }
   validates :unemployment_compensation_amount,
     numericality: {
-      greater_than_or_equal_to: 1,
-      message: proc { I18n.t('state_file.questions.w2.edit.no_money_amount') }
+      greater_than_or_equal_to: 1
     }
   validates :federal_income_tax_withheld_amount,
     numericality: {
       greater_than_or_equal_to: 0,
-      message: proc { I18n.t('state_file.questions.w2.edit.no_money_amount') }
+    },
+    presence: {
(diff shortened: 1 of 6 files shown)


app/models/state_file1099_g.rb
jenny-heath, line 1: am i correct in my reading that the only meaningful change to this file is the addition of error messages to `federal_income_tax_withheld_amount` and `state_income_tax_withheld_amount`?

jenny-heath: i think you said that but just wanted to make sure

spec/models/state_file1099_g_spec.rb
```diff
@@ -82,6 +82,16 @@
       )
     }
 
+    it do
```
jenny-heath, line 85: omg i didn't know you could just `it do` gonna use this all the time

jenny-heath (approved):

spec/models/state_file1099_g_spec.rb
powersurge360: yea, shoulda matchers are intended to be one-liners but w/ some customization it gets a little bulky and needs to be `it do ... end` rather than the prettier `it { ... }`

PR #5674, opened 2025-03-03
sketch out an example of what it might look like to delete blank nodes while building instead of while bundling

jenny-heath:
an experiment to present at Computer Gardening Club


arinchoi03:
@jenny-heath i think the specs are failing here but seems handy

jenny-heath:
this is old and maybe fyst will be sunsetting 🤷🏻‍♀️

PR #5679, opened 2025-03-04
Nj 301 - Add NJ retirement analytics

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/301

## What was done?
PMs want to be able to track 1) attrition when the Retirement Income Warning screen is shown to users and 2) verify that the Retirement Income Warning screen is displayed to users correctly. In order to do this, we are adding additional analytics.

1) We are adding a :shown status to the eligibility_retirement_warning_continue enum to capture the population of users who see the screen but do not select an option on the form.
2) We are adding additional calculator values to the StateFileNjAnalytics record for debugging purposes

- Update the enum for `eligibility_retirement_warning_continue` to include the `:shown` status, which is used when the retirement warning screen is shown to the user
- Add logic in the `nj_retirement_warning_controller` to set `eligibility_retirement_warning_continue` to `:shown` if the page is shown to the user
- Add logic to ensure that we never overwrite a `:yes` or `:no` value with `:shown`
- Add additional fields to the StateFileNjAnalytics record that are used to validate and debug NJ Retirement Warning fields
- Add backfill for data migration to populate existing StateFileNjAnalytics records with 0s

## How to test?
For engineers with access to the rails console, you can verify that the Intake state is changing correctly as you move through the flow. See below for example.

Demo (NJ OOI only)
https://drive.google.com/file/d/17ncF3GjyEYZOif6QQ0pOAgpCQyenf3nB/view?usp=sharing


app/controllers/state_file/questions/nj_retirement_warning_controller.rb
```diff
@@ -7,7 +7,11 @@ def self.show?(intake)
         line_15 = intake.calculator.line_or_zero(:NJ1040_LINE_15)
         line_16a = intake.calculator.line_or_zero(:NJ1040_LINE_16A)
         retirement_helper = Efile::Nj::NjRetirementIncomeHelper.new(intake)
-        retirement_helper.show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+        should_show = retirement_helper.show_retirement_income_warning?(line_15, line_16a) && Flipper.enabled?(:show_retirement_ui)
+        if should_show && intake.eligibility_retirement_warning_continue_unfilled?
+          intake.eligibility_retirement_warning_continue = :shown
```
jenny-heath, line 12: i don't think this works. i tested it locally and the value goes back to `unfilled` when i put a break point in `edit`

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -8,17 +8,31 @@
...
           expect(described_class.show?(intake)).to eq false
+          expect(intake.eligibility_retirement_warning_continue).to eq("unfilled")
         end
       end
   
       context "when the helper decides a income warning should be shown" do
         it "does show" do
           allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:show_retirement_income_warning?).and_return(true)
           expect(described_class.show?(intake)).to eq true
+          expect(intake.eligibility_retirement_warning_continue).to eq("shown")
```
jenny-heath, line 35: if you call `.reload` on the intake here the test will fail

db/data/20250304181048_backfill_nj_retirement_analytics_values.rb
jenny-heath, line 1: i think i don't quite understand what this backfill does...if these columns are new and they have a default value, then the default value should populate for existing records when you run the migration. am i missing something?

jenny-heath (commented):
a few things, ascending in scope:
- the way `shown` is set will not currently work. you have to use `update` or `save` to get it to persist.
- the backfill migration seems unnecessary if the whole goal is to save the default values of new columns for existing rows? but again maybe there's some other purpose i'm missing
- setting `shown` on `show?` feels like an antipattern because it's actually invoked every time any controller tries to find the next or previous controller (see `ControllerNavigation#seek`), so you'll end up with `shown` as the value in cases where they actually still haven't seen the page. i think it would make more sense to save it on `edit`.
- re: adding a new value to the enum, if i'm interpreting this right, `shown` just captures people who dropped off on this page (because it requires them to answer yes or no to continue). in my mind it's a little unusual to use the database to capture information like this, since we use tools like mixpanel and datadog to track progression through the app and see where people drop off. that said, i know y'all have limited access to these tools so maybe this is the best solution in the end. just wanted to ask whether other options had been discussed!

jachan:
> in my mind it's a little unusual to use the database to capture information like this, since we use tools like mixpanel and datadog

Thanks Jenny! Yes, since we don't have access to Datadog, the PMs are asking for this to be written to the analytics record.

> i think it would make more sense to save it on edit.

Thank you so much for your point about putting this on `edit` rather than `show?`. Will move the logic around.

> the backfill migration seems unnecessary if the whole goal is to save the default values of new columns for existing rows? but again maybe there's some other purpose i'm missing

That is the goal. I thought since the constraint for the field is NOT NULL, they will fail validation without some sort of backfill script. Will read up on this today.

> the way shown is set will not currently work. you have to use update or save to get it to persist.

Thank you for this! Will look into this, the failing unit test is extremely helpful.

Will have another revision out to address your concerns!

jachan:
@jenny-heath @aloverso should be ready for re-review

@jenny-heath I decided that refactoring the `show_retirement_warning?` method should be out of scope for this ticket, as IRS released Retirement Income today and I don't want to block this feature any longer.

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -1,25 +1,50 @@
...
+
+        it "does show" do
           expect(described_class.show?(intake)).to eq true
         end
+
+        it "updates eligibility_retirement_warning_continue" do
+          expect {
+            get :edit
+          }.to change { intake.reload.eligibility_retirement_warning_continue }.from("unfilled").to("shown")
+        end
```
jenny-heath, line 47: [sand] usually controller specs are split up by action, which can make them easier to skim quickly. if you have time maybe move this into it's own `#edit` block and also test the case where it doesn't update the column

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -31,13 +56,15 @@
         it "does NOT show" do
           allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:show_retirement_income_warning?).and_return(false)
           expect(described_class.show?(intake)).to eq false
+          expect(intake.eligibility_retirement_warning_continue).to eq("unfilled")
```
jenny-heath, line 59: [pebble] since `show?` no longer updates this column, i don't think there's a point in testing the value here

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -31,13 +56,15 @@
...
         end
       end
   
       context "when the helper decides a income warning should be shown" do
         it "does NOT show" do
           allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:show_retirement_income_warning?).and_return(true)
           expect(described_class.show?(intake)).to eq false
+          expect(intake.eligibility_retirement_warning_continue).to eq("unfilled")
```
jenny-heath, line 67: ^same

jenny-heath: just realized this also applies to lines 22 and 30

jenny-heath (approved):
couple of small test nitpicks that don't require re-review. otherwise looks good!

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -42,4 +46,41 @@
...
+  describe "#edit" do
+    context "when the controller is shown to the user" do
+      before do
+        allow_any_instance_of(Efile::Nj::NjRetirementIncomeHelper).to receive(:show_retirement_income_warning?).and_return(true)
+      end
+      context "and the existing value for eligibility_retirement_warning_continue is UNFILLED" do
+        it "sets eligibility_retirement_warning_continue on the intake to SHOWN" do
+          expect {
+            get :edit
+          }.to change { intake.reload.eligibility_retirement_warning_continue }.from("unfilled").to("shown")
```
aloverso, line 59: [dust] Cool! `to change` is a new rspec matcher to me, cool to learn! I assume this is the same as a three-liner like so?
```
expect to be X
get edit
expect to be Y
```

aloverso (approved):

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
jachan: Yep! Jenny showed me this! One important quality is that the code for the `change` portion needs to be inside `{ }` for the matcher to work

PR #5680, opened 2025-03-04
[FYST-1868] NC: only ask about 1099rs with taxable amt

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1868

## What was done?
- Consolidated logic for `eligible_1099rs` into `StateFileBaseIntake` since all states now (except NJ) requires `taxable_amount` to show a `RetirementIncomeSubtractionController` descendant page. NC previously had no assumption about what was an eligible 1099r, so it utilized the base intake method for `eligible_1099rs` which returned `state_file1099_rs` (basically all 1099rs from df)
- Updated `self.show?` logic on `RetirementIncomeSubtractionController` to be consistent across all states as well
## How to test?
  - unit tests written
  - Test data used: Elphaba, should only see 1 1099R subtractions question
  - Since this was refactored & impacts all states, double check that the _show_ logic works for all states (AZ/ID/NC/MD) as currently expected -- only eligible (>0 taxable_amount) 1099rs should be shown
    - for ID, the disability logic for 62-64 will factor into the show logic


spec/models/state_file_base_intake_spec.rb
```diff
@@ -313,4 +313,15 @@
     end
   end
 
+  describe "#eligible_1099rs" do
+    %w[az md nc nj].each do |state_code|
```
arinchoi03, line 317: nj is not using this method, but wanted to document that this method does exist for them

arinchoi03:
you can look at the individual commits to make sense of what's happening:
1. 93ad0ec45f88afc0b1740bb1dfc2976679e0ab3d fixing the bug
2. e7e20db140c0a6f10bcfe0c217188b8c4c6613c8 making NC use the partial like AZ/ID (MD to come, didn't want to add even more to this PR)
3. 80e94767614ff9229ff55cc257b0f817c10f282e spec tweaked to guard against showing ineligible 1099Rs on this page, as it had been before

spec/models/state_file_base_intake_spec.rb
tahsinaislam: i'm guessing idaho already has a separate test for this because of its additional logic with mfs

tahsinaislam (approved):
lgtm!

PR #5685, opened 2025-03-05
FYST-1893 Do not send notifications for filers in cancelled status

tahsinaislam:
## Link to pivotal/JIRA issue

## What was done?
- Updated the conditions for the following notifications (survey_notification, still_processing) to not send when submission is in a cancelled state 
- The following already do not send to intakes because the submission has not been built yet so it would not even be in a cancelled state: pre_deadline_reminder, post_deadline_reminder, finish_return
  - all of these messages have a query `  .where(efile_submissions: { id: nil })`

## How to test?
Updated unit tests

### DIFF
--- a/app/jobs/state_file/send_still_processing_notice_job.rb
+++ b/app/jobs/state_file/send_still_processing_notice_job.rb
@@ -2,7 +2,7 @@ module StateFile
   class SendStillProcessingNoticeJob < ApplicationJob
     def perform(submission)
       return if submission.data_source.sms_notification_opt_in_no?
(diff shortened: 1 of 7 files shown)


arinchoi03 (commented):
Can you let me know how we know the below? I can't find where the logic is that would prevent these from being sent for a submission in cancelled state:

>The following already do not send to submission in any state: pre_deadline_reminder, post_deadline_reminder, finish_return

are we saying that these reminders are never sent in any state (meaning states like az/nc/id OR cancelled/accepted/etc)?

arinchoi03 (approved):
ty for clarifying!

PR #5688, opened 2025-03-06
GYR1-690 allow site coordinators to edit site coordinator and team member accounts

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-690

## What was done?
- Added permission for site coordinators to manage other site coordinators and team members in their site(s)
- One of the tests was mocking a method (`accessible_vita_partners`) that wasn't being used in the relevant part of the ability file so we replaced the mocking with just setting up the actual db associations
## How to test?
- One spec in ability file and one on the users controller to see that it shows on the page
- Before
image

- After
image


app/lib/ability.rb
```diff
@@ -189,6 +188,9 @@ def initialize(user)
     end
 
     if user.site_coordinator?
+      can :manage, User, id: User.where(role: SiteCoordinatorRole.assignable_to_sites(user.role.sites)).pluck(:id)
```
embarnard, line 191: I would just check with product that we want to include all these actions, for example I'm not sure if we want to allow them to destroy users (we use suspend/unsuspend instead) so something like might be more accurate: 
```suggestion
      can [:update, :create, :update, :unsuspend, :suspend (and any other custom actions)], User, id: User.where(role: SiteCoordinatorRole.assignable_to_sites(user.role.sites)).pluck(:id)
```

app/lib/ability.rb
```diff
@@ -189,6 +188,9 @@ def initialize(user)
...
+      can :manage, User, id: User.where(role: TeamMemberRole.assignable_to_sites(user.role.sites)).pluck(:id)
```
embarnard, line 192: I guess instead of including all those custom actions you could also just add a cannot after the lines like for example
```suggestion
      can :manage, User, id: User.where(role: TeamMemberRole.assignable_to_sites(user.role.sites)).pluck(:id)
      cannot :destroy, User
```

embarnard (approved):
Looks good! But I would check on actions, see comment

app/lib/ability.rb
```diff
@@ -189,6 +188,9 @@ def initialize(user)
     end
 
     if user.site_coordinator?
+      can [:suspend, :unsuspend, :update, :unlock, :resend_invitation], User, id: User.where(role: SiteCoordinatorRole.assignable_to_sites(user.role.sites)).pluck(:id)
```
anisharamnani, line 191: Thanks @embarnard for flagging the actions. These are the updated ones.

app/lib/pdf_filler/id39r_pdf.rb
```diff
@@ -8,7 +8,6 @@ def source_pdf_name
 
     def initialize(submission)
       @submission = submission
-      @intake = submission.data_source
```
anisharamnani: 🙈 i was switching between branches and did the wrong thing

PR #5691, opened 2025-03-06
[FYST-1900] Make MD retirement subtraction card consistent with other states'

arinchoi03:
Followup to https://github.com/codeforamerica/vita-min/pull/5680; noticed that all states now use the partial in the review_header but only MD does not

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1900

## What was done?
- Uses the same pattern as other states
  - only showing the information for subtractions for `eligible_1099rs`: previously we were showing this card if any 1099rs exist, and _then_ only showing the relevant sub-elements if followup existed (which could only exist if the 1099r was qualified) which is a bit less straightforward.
- Left the disability card in the `md_review` b/c it needs to show if 1099Rs exist at all (at this point anyway, that's the logic); there is a followup ticket to refactor this to show if it matches the `include` logic on MD502R; Anyways, disability needs to be shown if 502R exists at all, so `eligible_1099rs` logic does not factor into showing disability-related information on the final review. I can move the whole section into the partial, if desired.
- Alternatives considered:
  - I don't like how we're putting not shared html into the `shared` folder, i think they should all move back into the `#{stsate_code}_review` folders as specific partials. However, that'll involve me moving around more copy & not sure if i want to make Aaron/September do more testing for all states 😭 
## How to test?
  - Test data used
- Risk Assessment
- Before
- After


app/views/state_file/questions/shared/_md_retirement_income_deductions_review_header.erb
```diff
@@ -0,0 +1,39 @@
+<% if current_intake.eligible_1099rs.length.positive? %>
+  <section id="retirement-income-source" class="white-group">
+    <div class="spacing-below-5">
+      <h2 class="text--body text--bold spacing-below-15"><%=t(".retirement_income_deductions") %></h2>
+
+      <% current_intake.eligible_1099rs.each_with_index do |state_file1099_r, index| %>
```
arinchoi03, line 6: note that we're changing to `eligible_1099rs` instead of `state_file1099_rs` here -- problem navigating back to a specific 1099R unless we have the same indices or a given 1099r!

anisharamnani (approved):
⭐ Thanks Arin for taking on this work!

app/views/state_file/questions/shared/_md_retirement_income_deductions_review_header.erb
arinchoi03: in fact, we _do_ have a problem. I used frank rollover & tried to navigate to the 2nd retirement deduction shown -- it takes me to oops page (on demo; fixed on this PR)

PR #5692, opened 2025-03-06
FYST-1376 Pull suffix to final pdf for all states

tahsinaislam:
## Link to pivotal/JIRA issue

## What was done?
- Added prefixes to last names on all pdfs across states that were missing. The following is the list:
  - id40
  - md502cr
  - md502
  - md502r
  - md502b
  - md el101
  - nc d400
  - nc d400 schedule s
- refactored all pdfs to pull name and ssn from intake rather than xml

## How to test?
- Added in units tests for adding prefixes

Screenshot 2025-03-06 at 3 35 35 PM
Screenshot 2025-03-06 at 3 35 25 PM

Screenshot 2025-03-06 at 3 35 42 PM


jenny-heath (commented):
question! why did you change it to pull from the intake instead of the xml? afaik we intentionally read as much data as possible from the xml when filling out the pdfs so we don't have to manually make sure they match up

jenny-heath (approved):
lgtm!

PR #5693, opened 2025-03-06
Make the drop-down arrow for multi-select vita-partners on hub/invitations clickable

embarnard:
## Link to pivotal/JIRA issue
- e.g. [link](https://codeforamerica.atlassian.net/browse/GYR1-693?atlOrigin=eyJpIjoiYjljNTI2OWFjMzJlNGUzZjlkYjg1NTQ4OTNlM2NmNjIiLCJwIjoiaiJ9)

## What was done?
- made the arrow next to the multi select vita-min partner drop down in hub/invitations clickable 
## How to test?
- Go to `/hub/users/invitations` or Tools->Invitations in the hub
- select Site Coordinator or Team member to test multi select 
https://github.com/user-attachments/assets/fbc2ada9-07e2-4b83-afb0-2fb13e8c1e88


app/helpers/tagging_helper.rb
```diff
@@ -15,7 +15,8 @@ def taggable_vita_partners(vita_partners)
   end
 
   def taggable_sites(vita_partners)
-    taggable_vita_partners = vita_partners.sites.includes(:parent_organization).map do |site|
+    taggable_vita_partners = vita_partners.sites.includes(:parent_organization).filter_map do |site|
```
embarnard, line 18: this will crash the drop down if there is any site that doesn't have a parent org (aka parent org got deleted at some point); adding this next will avoid this error

app/views/hub/sites/edit.html.erb
```diff
@@ -2,7 +2,9 @@
 <% content_for :page_title, @title %>
 <% content_for :card do %>
   <section class="slab slab--padded">
-    <%= link_to(@site.parent_organization.name, edit_hub_organization_path(id: @site.parent_organization)) %>
+    <% if @site.parent_organization %>
```
embarnard, line 5: same with this, will crash for a site with no parent org

app/views/devise/invitations/new.html.erb
```diff
@@ -38,7 +38,7 @@
       <% end %>
 
       <% if (params[:role] == SiteCoordinatorRole::TYPE) || (params[:role] == TeamMemberRole::TYPE) %>
-        <div class="form-group">
+        <div class="form-group tagify-clickable-dropdown-wrapper">
```
embarnard, line 41: had to add this b/c the drop down arrow is inserted through css via an ::after classification, this will wrap the whole select box to ensure we include it

arinchoi03: is the drop down arrow inserted somewhere outside our code via tagify?

embarnard: for the single select (used for coalition and org leads) Tagify inserts it but for this one, multiselect, its inserted via honey crisp @arinchoi03

arinchoi03 (commented):
there are other places where this `.multi-select-vita-partner` is used -- do we need to update the behavior there too? (add `.tagify-clickable-dropdown-wrapper`)?

I see this on `hub/state_routings/edit.html.erb`, `admin/experiments/edit.html.erb`, `hub/users/edit_role`, `_client_filters.html.erb`

would love to chat on Monday about how this works, it's been a while since I've been looking at javascript 😵

anisharamnani:
+1 to @arinchoi03’s comment if you wouldn’t mind adding me to the chat. it feels like it’s been awhile since i’ve touched JS & i didn’t feel comfortable approving only due to my own lack of knowledge.

app/javascript/lib/tagging/index.js
```diff
@@ -57,30 +59,57 @@ export function initMultiSelectVitaPartner() {
...
-        let { value } = event.target
+        let { value } = event.target;
         if (value) {
-            value = JSON.stringify(JSON.parse(event.target.value).map(v => v.id))
+            value = JSON.stringify(JSON.parse(event.target.value).map(v => v.id));
         }
-        document.querySelector("#vita_partners").value = value
-    })
+
+        const hiddenInput = document.querySelector("#vita_partners");
```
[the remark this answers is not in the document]
embarnard: My understanding of this story is that the existing box has an arrow that is confusing for users that when they try to click it and there is nothing that comes down. Experiments is an admin tool so only used by CFA folks and additionally along with hub/state_routings and client filters there is no drop down arrow by those boxes so nothing to make clickable. I'll make a note to Ryan about these pages though
Screenshot 2025-03-12 at 3 08 04 PM
Screenshot 2025-03-12 at 3 06 47 PM
Screenshot 2025-03-12 at 3 07 02 PM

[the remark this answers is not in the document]
embarnard: Yeah sorry that other piece of code was trying to solve this null vitapartners issue I was seeing on that page occassionally but I'll try to make that clearer or take it out

spec/features/hub/filtered_clients_bulk_action_spec.rb
```diff
@@ -35,7 +35,7 @@
     perform_enqueued_jobs
     visit current_path
 
-    expect(page).to have_text "Successful Bulk Client Organization Update"
-    expect(page).to have_text "You successfully moved 30 clients to Rooster Brew."
+    expect(page).to have_text("Successful Bulk Client Organization Update", wait: 5)
```
embarnard, line 38: this test is a little flakey sometimes because it doesn't allot enough time for the job to go through to get a success message. Adding this will allow for the test to wait up to 5 seconds to look for the text before failing

app/javascript/lib/tagging/index.js
```diff
@@ -57,30 +59,53 @@ export function initMultiSelectVitaPartner() {
...
     input.addEventListener("change", (event) => {
-        let { value } = event.target
+        // extracts just the vita partner ids and stores in hidden field
+        let { value } = event.target;
         if (value) {
-            value = JSON.stringify(JSON.parse(event.target.value).map(v => v.id))
+            value = JSON.stringify(JSON.parse(event.target.value).map(v => v.id));
+        }
+
+        const hiddenField = document.querySelector("#vita_partners");
```
embarnard, line 74: I tried to delete this but then ran into all sorts of issues because some of the forms that use `multi-select-vita-partner` have a hidden field (something that might look like `<%= f.hidden_field(:vita_partners, class: "multi-select-vita-partner") %>`) where they will store these vita partner ids for the form like on admin/experiments or client filters

app/javascript/lib/tagging/index.js
```diff
@@ -57,30 +59,53 @@ export function initMultiSelectVitaPartner() {
...
+    if (clickableContainer) {
+        clickableContainer.addEventListener("click", (event) => {
+            event.stopPropagation();
+            if (tagify.state.dropdown.visible) {
+                tagify.dropdown.hide();
+            } else {
+                tagify.dropdown.show();
+            }
+        });
+        document.addEventListener("click", (event) => {
```
embarnard, line 90: the big change here for the re-review is that this is now inside of the `if (clickableContainer)` statement so that we won't have issues if we have `multi-select-vita-partner` elements that don't have `tagify-clickable-dropdown-wrapper`

spec/features/hub/filtered_clients_bulk_action_spec.rb
arinchoi03: oof 5 seconds is kind of long, can it be shorter, or is 5 seconds what we need? and do we need to wait for both expect for up to 5 seconds, or can we just put it on this one & remove from the second?

app/javascript/lib/tagging/index.js
arinchoi03: i think this refactor makes things clearer, ty!

arinchoi03 (approved):
I think it's a lot easier to digest now, ty Em!

embarnard:
@arinchoi03 yes I agree 5 seconds is long but I just doubled checked and this method will only wait **up to** five seconds if it can't find the text, if it finds the text in less than 5 seconds it will move on. I agree that its also potentially a long time before failing the test if the test should truly fail but I still think it will save more time in the long run from not having to re-run the whole test suite if it fails incorrectly.

spec/features/hub/filtered_clients_bulk_action_spec.rb
embarnard: oops posted this below but posting here as well for continuity:
@arinchoi03 yes I agree 5 seconds is long but I just doubled checked and this method will only wait up to five seconds if it can't find the text, if it finds the text in less than 5 seconds it will move on. I agree that its also potentially a long time before failing the test if the test should truly fail but I still think it will save more time in the long run from not having to re-run the whole test suite if it fails incorrectly.

PR #5694, opened 2025-03-06
FYST-1796 twilio webhook error

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1796

## What was done?
- We send the status_callback arg to twilio when we send state file messages even though we don't have that endpoint set up, which causes an error
## How to test?
- unit test whether the argument is passed
- N/A

### DIFF
--- a/app/services/multi_tenant_service.rb
+++ b/app/services/multi_tenant_service.rb
@@ -1,4 +1,5 @@
 class MultiTenantService
+  include Rails.application.routes.url_helpers
   attr_accessor :service_type
 
   SERVICE_TYPES = [:gyr, :ctc, :statefile]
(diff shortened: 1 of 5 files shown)


mrotondo (approved):
Thanks for doing this! Non-blocking but if you want to: I think it might be good to, rather than branching on `service_type` in `TextMessageVerificationCodeService`, add a method named something like `twilio_status_webhook_url` to `MultiTenantService` (and make it `when :statefile then nil`), and then just remove the conditional from the `TextMessageVerificationCodeService` entirely and let the `.compact` call take care of removing the `nil` from the `message_arguments` hash

jenny-heath:
@mrotondo done!

mrotondo (approved):
nice, thanks again!

PR #5700, opened 2025-03-07
add w2 show page

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1874

## What was done?
- Added a w2 show page and link to it from the income review page on states that don't allow w2 editing (only NC right now)
- Edited the recently-added 1099-R to more closely match its designs, using the same semantic html structure as the w2 show page 
- I used [`<dl>, <dt>, <dd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) for the show pages' HTML, as this was the closest I could fine to a "right answer" on how to structure read-only labeled numerical/textual data in a way that would be interpreted correctly by accessibility tools. I tried it out with a screen reader and, at least on mac, it didn't seem to have any benefit over just putting everything in `<p>` tags, but maybe it's better in some way I don't know how to assess. Typical accessibility tools like `<label>` and `aria-labeledby` are only available on input elements, so I couldn't use those unless I changed the HTML to use readonly input fields instead of normal tags, which I felt was not the intent of the page.
## How to test?
- Go through the NC flow for a persona with some w2s and verify that they show up on the income review page, can be clicked and that doing so goes to a non-interactive page where the w2 info can be viewed, and that the continue and back buttons work as expected
- Verify that the 1099-R show pages in NC still work
- Try adding `/edit` to the URL of a w2 page while going through the NC flow and verify that it redirects you to the income review page
#### Before
image

#### After
image

### W2 show screen
image

### DIFF
--- a/app/controllers/state_file/questions/w2_controller.rb
+++ b/app/controllers/state_file/questions/w2_controller.rb
@@ -12,9 +12,10 @@ module StateFile
       end
 
       def edit
-        @state_code = current_state_code
         @w2.valid?(:state_file_edit)
       end
+
+      def show; end
 
       def update
         @w2.assign_attributes(form_params)
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -366,12 +366,24 @@
       list-style-type: disc;
       padding-left: 2rem;
       margin: 1rem 0;
+
+      .unbulleted {
+        list-style-type: none;
+      }
     }
 
     li {
       font-size: 1.6rem;
       line-height: 2.4rem;
       margin-bottom: 0.5rem;
+    }
+
+    dt {
+      font-size: 1.3rem;
+    }
+
+    dd {
+      margin-left: 0;
     }
   }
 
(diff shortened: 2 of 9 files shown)


app/controllers/state_file/questions/w2_controller.rb
```diff
@@ -12,10 +13,11 @@ def prev_path
       end
 
       def edit
-        @state_code = current_state_code
         @w2.valid?(:state_file_edit)
       end
 
+      def show; end
+
       def update
```
tahsinaislam, line 21: is it bad to put some of the logic into the controller like this: 

```
@box14_codes_with_values = []
        
        if @box14_codes.present?
          @box14_codes.each do |code|
            field_name = "box14_#{code['name'].downcase}"
            
            value = if code['name'] == "UIWFSWF"
                      @w2.get_box14_ui_overwrite
                    else
                      @w2.send(field_name)
                    end
            
            @box14_codes_with_values << {
              name: code['name'],
              translation_key: "state_file.questions.w2.edit.box14_#{code['name'].downcase}_html",
              value: value
            }
          end
```
and then iterate through `  @box14_codes_with_values` in the view?

tahsinaislam (approved):
lgtm!

PR #5701, opened 2025-03-07
[FYST-1359] Add Names and SSN to ID39 PDF

anisharamnani:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-1359

## What was done?
Added Names & SSN to ID39. 
* If mfj & both filers have the same last name, the names appear as “Helena & Mark Eagan"
* If mfj & both filers have different last names, the names appear as “Helena Eagan & Mark Scout"
* If non-mfj, the primary filer’s name appears as “Helena Eagan"

## How to test?
* added test to `id39r_pdf_spec.rb` 
* went through flow & saw that the PDF is correctly filled in

- Before
image
- After
Screenshot 2025-03-06 at 5 21 47 PM


app/lib/pdf_filler/id39r_pdf.rb
```diff
@@ -40,6 +43,18 @@ def hash_for_pdf
       answers
     end
 
+    def formatted_display_name
+      if @intake.filing_status_mfj?
+        if @intake.primary.last_name == @intake.spouse.last_name
```
arinchoi03, line 48: [super dust] i dont think this will be a likely scenario, but what if for some reason they had different capitalizations? Should we downcase & match?

arinchoi03 (approved):
👏

app/lib/pdf_filler/id39r_pdf.rb
anisharamnani: 🤔 it’s not a bad idea, i pushed up a fix for this.

PR #5706, opened 2025-03-07
NJ 235 - fix calculator negative value bug

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/235#issue-2763737917

## What was done?
- Problem reported was that review screen contained a negative value for "NJ taxable income" (see screenshot)
- Looked in the calculator and we were allowing negative values for many of the subtractions.
- Added checks for all subtraction calculations in the NJ calculator. If line value was negative, set it to zero instead

## How to test?
- Use Eastern Goldfinch persona continue to end of the flow and look at NJ taxable income on the review screen

- Before
image

- After
image

### DIFF
--- a/spec/lib/efile/nj/nj1040_calculator_spec.rb
+++ b/spec/lib/efile/nj/nj1040_calculator_spec.rb
@@ -724,19 +724,17 @@ describe Efile::Nj::Nj1040Calculator do
   describe 'line 29 - gross income' do
     let(:intake) { create(:state_file_nj_intake) }
 
-    it 'sets line 29 to the sum of all state wage amounts minus 28c' do
-      allow(instance).to receive(:calculate_line_15).and_return 50_000
-      allow(instance).to receive(:calculate_line_16a).and_return 1_000
-      allow(instance).to receive(:calculate_line_28c).and_return 3_000
+    it 'sets line 29 to line 27 minus line 28c' do
+      allow(instance).to receive(:calculate_line_27).and_return 3_000
+      allow(instance).to receive(:calculate_line_28c).and_return 1_000
       instance.calculate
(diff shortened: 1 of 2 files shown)


spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -731,6 +731,16 @@ def over_65_birth_year
...
       expect(instance.lines[:NJ1040_LINE_29].value).to eq(48_000)
     end
+
+    context 'when 28c is larger than 29' do
+      it 'sets line 29 to zero' do
+        allow(instance).to receive(:calculate_line_15).and_return 1_000
+        allow(instance).to receive(:calculate_line_16a).and_return 1_000
+        allow(instance).to receive(:calculate_line_28c).and_return 2_001
+        instance.calculate
+        expect(instance.lines[:NJ1040_LINE_29].value).to eq(0)
```
aloverso, line 741: [pebble] I think the test title is misleading here? It should be "when 28c is greater than line 27" (from looking at the code) right?

Additionally, if that is the case, I'd suggest/prefer to see the test simplified by just mocking Line 27 and Line 28c rather than lines 15 and 16a, which seems more complicated and harder to read/parse.

aloverso (approved):
One small test improvement, otherwise looks good! Appreciate you finding all the relevant cases here & futureproofing

spec/lib/efile/nj/nj1040_calculator_spec.rb
jachan: That makes sense. The logic for line 27 has also changed in the meantime so this makes it more accurate.

jnf (approved):
lgtm!

PR #5719, opened 2025-03-11
Fyst 1905 md disability proof not saving correctly

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1905

## What was done?
- When users clicked through to the softboarding page, any selections they had made on the `md-permanently-disabled` page were discarded because they never clicked Continue to submit the form.
- This removes the warning box (and the included link to the softboarding page), and instead makes non-mfj filers go through the same flow as mfj filers, where the softboarding page is shown after the `md-permanently-disabled` page if they satisfy the conditions (filer under 65, permanently disabled, no proof of disability submitted)
- Alternatives considered were to submit the form when clicking the link to the softboarding page, but I estimated that would be more awkward to implement
## How to test?
- Several of the examples in the `md_pension_exclusion_offboarding_controller_spec` controller test weren't actually testing anything (in that you could change their `before` or `let` expressions to set the opposite value and the tests would still pass. To address this, I changed the test so that it sets all necessary preconditions for the page to show, verifies that it does show, and then verifies that it doesn't show when each of the preconditions is individually set to false.
- Before
image

- After
image

### DIFF
--- a/app/views/state_file/questions/md_permanently_disabled/edit.html.erb
+++ b/app/views/state_file/questions/md_permanently_disabled/edit.html.erb
@@ -35,20 +35,10 @@
 
       <div class="question-with-follow-up__follow-up" id="primary-disability-proof">
         <div class="white-group">
-          <div class="question-with-follow-up">
-            <div class="question-with-follow-up__question">
-                <% show_disability_warning =  !current_intake.filing_status_mfj? ? "#disability-warning" : nil %>
-                <%= f.cfa_radio_set(:primary_proof_of_disability_submitted, label_text: t(".primary_proof_question_html"), collection: [
-                  { value: "yes", label: t("general.affirmative") },
-                  { value: "no", label: t("general.negative"), input_html: { "data-follow-up": show_disability_warning }},
-                ]) %>
-              </div>
-            <div class="question-with-follow-up__follow-up" id="disability-warning">
-                <div class="notice">
-                  <p><%= t(".unfortunately_dont_support_html", path: questions_md_pension_exclusion_offboarding_path) %></p>
-                </div>
-              </div>
-          </div>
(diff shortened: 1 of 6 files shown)


app/views/state_file/questions/md_permanently_disabled/edit.html.erb
```diff
@@ -38,20 +38,10 @@
...
-                  { value: "yes", label: t("general.affirmative") },
-                  { value: "no", label: t("general.negative"), input_html: { "data-follow-up": show_disability_warning }},
-                ]) %>
-              </div>
-            <div class="question-with-follow-up__follow-up" id="disability-warning">
-                <div class="notice">
-                  <p><%= t(".unfortunately_dont_support_html", path: questions_md_pension_exclusion_offboarding_path) %></p>
-                </div>
-              </div>
-          </div>
```
DrewProebstel, line 54: Was it just to much of a pain to have the notice buried so deep in question with follow ups?

DrewProebstel (commented):
Looks good.

anisharamnani (approved):
md disability strikes again!

app/views/state_file/questions/md_permanently_disabled/edit.html.erb
mrotondo: No, it's that clicking off the page here would cause this page's form to never get submitted. There were various shenanigans proposed about having this link also submit the form, or having the Continue button on the softboarding page somehow remember and set the fields to what the user had selected on this page, but I spoke with design and we felt that just having all users consistently see softboarding as the next page in the flow, if it was applicable, was reasonable from a user perspective and easier to implement.

DrewProebstel: Cool thanks!

PR #5721, opened 2025-03-11
NJ 274 - a11y fix, buttons have role link

jachan:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/274#issue-2813739879

## What was done?
- Fix to be compliant with [WCAG 3.2.4: Consistent Identification (Level AA)](https://w3.org/TR/WCAG21#consistent-identification) 
- Currently, we use button styling for both links and buttons. This can be confusing to users, as it's unclear that the "Submit" and "Continue" buttons will take them to the next page when they are coded as `button` elements rather than `a` elements

```
<a
    class="button button--primary button--wide" 
    id="firstCta" href="/en/questions/contact-preference">
    Continue
</a>
```
and

```
<input 
    type="submit" name="commit" 
    value="Continue" 
    class="button button--primary button--wide" 
    data-disable-with="Continue">
```

While one way to solve this is to change all Continue and Submit buttons to <a> tags, but that felt too disruptive to a code-base of this size. Instead, I implemented the alternate solution of using the `role="link"` attribute to input/button elements to signify that the button will link out to a different page. I did this by modifying the shared component in the VitaMin form builder library.

## How to test?
- navigate to `notifications-preference` screen or any screen with a continue or submit button that takes you to the next page and inspect HTML to validate that there is now a `role="link"` attribute on any input element.

- Before
image

- After (role attribute on the input element)
image

### DIFF
--- a/app/helpers/vita_min_form_builder.rb
+++ b/app/helpers/vita_min_form_builder.rb
@@ -478,6 +478,7 @@ class VitaMinFormBuilder < Cfa::Styleguide::CfaFormBuilder
   def submit(value, options = {})
     options[:data] ||= {}
     options[:data][:disable_with] = value
+    options[:role] = "link"
     super(value, **options)
   end
 


app/helpers/vita_min_form_builder.rb
```diff
@@ -478,6 +478,7 @@ def vita_min_checkbox_set(
   def submit(value, options = {})
     options[:data] ||= {}
     options[:data][:disable_with] = value
+    options[:role] = "link"
```
jachan, line 481: The continue button below this uses the `submit` method defined here so we update both `f.continue` and `f.submit`.

aloverso:
@jachan this doesn't seem to address https://github.com/newjersey/accessibility-pm/issues/18 — is that intentional to leave for another ticket? The inconsistent behavior of these non-submits: "Email me a code" is still a button and the "Sign in with email" is still a link

jachan:
Closed in favor of https://github.com/codeforamerica/vita-min/pull/5748

PR #5726, opened 2025-03-11
FYST-1486 withdrawal date refactor

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1486

## What was done?
- Refactored the way we track when payment is due for taxes owed in each state
   - When taxes are owed and filers choose to pay with direct deposit, they can schedule when the payment is withdrawn
   - They can choose any date between today and the deadline (with exceptions in NC)
   - Once the form is viewed on or after the deadline, they are no longer able to set the withdrawal date
      - In this case we set the date automatically to the current day
      - Future work will adjust this
   - If the form is submitted after the payment deadline with a scheduled withdrawal date, the date is ignored
- Since we have States in three different timezones on April 15th (EDT, MDT, MST), the functionality for the payment deadline is now in the `StateInformationService`
   - Each state's deadline is stored as a month and a day, and then `payment_deadline_date(state_code)` will return the full date in the current filing year
   - Also `before_payment_deadline?(datetime, state_code)` will use the State's government timezone to detect if the given datetime is before midnight "locally" in that State
      - Noting that this is the _government_ timezone, because in Arizona the Navajo nation follows a different timezone than the State government
- While I was looking at the code I also refactored some other pieces such as extracting the NC validations into separate methods and reorganized the tests in `taxes_owed_form_spec`. Apologies that this makes it hard to review, I'm happy to pair to review.

## How to test?

- I added `TimeCop` tests to the taxes owed controller that look for form changes before and after the deadline
- The form tests also change the `app_time` in order to test when/if the validations run correctly

- Personas in each state that owe:
   - AZ: Martha_MFJ_Owe, Shell HOH
   - ID: Grey_no_1099r, Grey_1099R, Spud_Single_Blind, Ida_HoH, Yukon_Child_Dep_Care
   - MD: Drum_MFS_over1K, Felicity_Single, Felicity_Single_w2, Eels_1099R
   - NC: Pork E Pig with 1099-R
- Use session toggles to manually go to one minute before/after April 15th, 2025 in the timezone of the state you are testing (you'll have to type it into the date selector)
   - In MD, the deadline is April 30th instead
- Look for the date selector to appear and disappear, as well as the text to change, and display the correct deadline

- Before
- After


spec/controllers/state_file/questions/taxes_owed_controller_spec.rb
```diff
@@ -48,5 +48,65 @@
       end
     end
 
+    StateFile::StateInformationService.active_state_codes.excluding("ny").each do |state_code|
```
arinchoi03, line 51: should we exclude `"nc"` as well since this controller is not used by that state?

spec/controllers/state_file/questions/taxes_owed_controller_spec.rb
```diff
@@ -48,5 +48,65 @@
...
+          end
+
+          it "does not explain when your payment will be withdrawn" do
+            get :edit
+            expect(response).to be_successful
+            expect(response_html).not_to have_text(
+              "Because you are submitting your return on or after #{stringified_deadline}, " \
+              "the state will withdraw your payment as soon as they process your return."
+            )
+          end
```
arinchoi03, line 82: can the two `it` statements just be one? same for the context below

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -1,63 +1,49 @@
...
+    private
 
-    def date_electronic_withdrawal_nc_validations
-      five_pm_et = app_time_eastern.change(hour: 17, min: 0, sec: 0)
-      if (app_time_eastern >= five_pm_et) && less_than_two_business_days_in_future?
-        errors.add(:date_electronic_withdrawal, I18n.t("errors.attributes.nc_withdrawal_date.post_five_pm"))
-      end
-      if date_electronic_withdrawal <= app_time_eastern.to_date
+    def withdrawal_date_is_not_today
+      unless date_electronic_withdrawal.after?(@form_submitted_time.to_date)
```
arinchoi03, line 14: i'm struggling to see how this logic means `withdrawal_date_is_not_today`?

we add an error unless date_electronic_withdrawal is after form is submitted...so it more means that `withdrawal_date_is_in_the_future` or something like that?

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -1,63 +1,49 @@
...
-      current_date
     end
 
-    def less_than_two_business_days_in_future?
-      two_business_days_later = add_business_days(2)
-      date_electronic_withdrawal <= two_business_days_later
+    def add_business_days_to_date(date, num_days)
+      while num_days.positive?
+        date += 1.day
+        num_days -= 1 if date.wday.between?(1, 5)
```
arinchoi03, line 44: i kind of like the comment:
`# Check if the current date is a business day (Mon-Fri)` here for obscure `wday` method

app/services/state_file/state_information_service.rb
```diff
@@ -49,6 +50,19 @@ def department_of_taxation(state_code)
         I18n.t("state_file.state_information_service.#{state_code}.department_of_taxation")
       end
 
+      def payment_deadline_date(state_code)
```
arinchoi03, line 53: i wonder if leaving a comment that this returns date based on UTC time might be good to note?

spec/forms/state_file/taxes_owed_form_spec.rb
```diff
@@ -1,142 +1,172 @@
...
+      expect(intake.account_number).to eq nil
+      expect(intake.account_type).to eq "unfilled"
+      expect(intake.withdraw_amount).to eq nil
+      expect(intake.date_electronic_withdrawal).to eq nil
     end
+  end
 
-    context "when params valid and payment type is deposit" do
-      let(:bank_info_params) do
+  StateFile::StateInformationService.active_state_codes.each do |state_code|
```
arinchoi03, line 30: again, I think we should exclude NC here (and maybe even NY too?) to further serve as documentation that this form is unused for NC.

spec/forms/state_file/taxes_owed_form_spec.rb
```diff
@@ -1,142 +1,172 @@
...
-            expect(intake.account_number).to eq "12345"
-            expect(intake.date_electronic_withdrawal).to eq Date.parse("April 30th, #{current_year}")
+        context "when the withdrawal date is in the past" do
+          let(:withdrawal_month) { (app_time - 1.days).month }
+          let(:withdrawal_day) { (app_time - 1.days).day }
+
+          it "is not valid" do
+            form = described_class.new(intake, params)
+            expect(form).not_to be_valid
+            expect(form.errors).to include :date_electronic_withdrawal
```
arinchoi03, line 140: should we be more explicit about what error is attached? for example, this one vs the one below it are maybe different errors & i'd expect to see that in the tests

arinchoi03 (commented):
ok i gave it my best shot! nothing concerning on my end but a bunch of comments to talk through before going to acceptance

app/controllers/state_file/questions/tax_refund_controller.rb
```diff
@@ -9,11 +9,6 @@ def refund_amount
         current_intake.calculated_refund_or_owed_amount
       end
       helper_method :refund_amount
-
-
-      private
-
-      def card_postscript; end
```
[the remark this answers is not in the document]
mpidcock: Yeah, I think it was left out by error, but it would be better to consult product first and separate this out.

app/controllers/state_file/questions/taxes_owed_controller.rb
```diff
@@ -3,17 +3,25 @@ module Questions
     class TaxesOwedController < QuestionsController
 
       def self.show?(intake)
-        intake.calculated_refund_or_owed_amount.negative? # what happens if zero?
```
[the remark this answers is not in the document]
mpidcock: correct. this comment bothers me everytime i see it because it makes it seem like we don't handle this situation (but we do)

app/services/state_file/state_information_service.rb
```diff
@@ -147,7 +163,8 @@ def state_intake_class_names
         voucher_form_name: "Form PV",
         voucher_path: "/pdfs/md-pv-TY2024.pdf",
         w2_supported_box14_codes: [{name: "STPICKUP"}],
-        w2_include_local_income_boxes: true
+        w2_include_local_income_boxes: true,
+        payment_deadline: { month: 4, day: 30 } # April 30th every year
```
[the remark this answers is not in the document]
mpidcock: I put it there because I know we sometimes do global searches for the word "April" but I'm ok with deleting it

app/forms/state_file/taxes_owed_form.rb
```diff
@@ -41,52 +50,52 @@ def self.existing_attributes(intake)
...
 
-    def withdrawal_date_intake_validations
-      @intake.date_electronic_withdrawal = date_electronic_withdrawal
-      @intake.validate(:date_electronic_withdrawal)
-      if @intake.errors[:date_electronic_withdrawal].present?
-        errors.add(:date_electronic_withdrawal, @intake.errors[:date_electronic_withdrawal])
-      end
+    def withdraw_amount_does_not_exceed_owed_amount
+      owed_amount = intake&.calculated_refund_or_owed_amount&.abs
+      return unless self.withdraw_amount.to_i > owed_amount
```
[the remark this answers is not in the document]
mpidcock: if the user has gotten here without a `calculated_refund_or_owed_amount` then I think the argument error would be appropriate. I wouldn't want to swallow the error and continue with a default of zero or something, because clearly something wrong is happening.

app/forms/state_file/taxes_owed_form.rb
```diff
@@ -41,52 +50,52 @@ def self.existing_attributes(intake)
...
-      end
+    def form_submitted_before_payment_deadline?
+      StateInformationService.before_payment_deadline?(@form_submitted_time, intake.state_code)
     end
 
-    def date_electronic_withdrawal_post_deadline
-      Date.parse(post_deadline_withdrawal_date)
-    end
+    def date_electronic_withdrawal
+      # Future story: set post-deadline withdrawal date during submission bundle instead (use submission date)
```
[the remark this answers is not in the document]
mpidcock: I'm waiting on product to gather the requirements

spec/controllers/state_file/questions/taxes_owed_controller_spec.rb
mpidcock: The NC controller inherits from this controller, so it still makes sense to me to include it here. There are additional tests for the NC controller that handle the exceptions, but the base validations are largely the same and it could be problematic to duplicate the tests.

spec/controllers/state_file/questions/taxes_owed_controller_spec.rb
mpidcock: They are on different parts of the page so I think of them as different elements, but combining is ok with me

app/forms/state_file/nc_taxes_owed_form.rb
mpidcock: the `@form_submitted_time` is essentially the current time in the given timezone when this code is run. By calling `to_date` we get a day, which is today. In NC, we not only want this to be in the future, but not the same day as "today". I can imagine the specifics of this being changed a little in the future though to accommodate if the intake is submitted on a future day. (we haven't had this issue yet though, so might not matter)

app/services/state_file/state_information_service.rb
mpidcock: It returns a date without any time or timezone (but becomes UTC if you convert it to something with a time). I'll add a comment.

spec/forms/state_file/taxes_owed_form_spec.rb
mpidcock: I would rather keep the shared test cases in one place instead of copying them

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -1,63 +1,49 @@
 module StateFile
   class NcTaxesOwedForm < TaxesOwedForm
-    include DateHelper
-    set_attributes_for :intake,
-                       :payment_or_deposit_type,
-                       :routing_number,
-                       :account_number,
-                       :account_type,
-                       :withdraw_amount
```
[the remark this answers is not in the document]
mpidcock: turns out I was wrong that `set_attributes_for` can be inherited, but the tests don't specifically cover it. looks like there is a need for a few more tests to cover things I thought were inherited.

app/forms/state_file/nc_taxes_owed_form.rb
arinchoi03: I still think this method should be renamed to be in-line with what the code says:
`withdrawal_date_is_in_the_future?` (so not today & not in the past)

app/services/state_file/state_information_service.rb
```diff
@@ -59,10 +59,10 @@ def survey_link(state_code, locale: nil)
       # Returns the state-specific date in the current filing year only - no time or timezone.
       # If later converted to a DateTime, the time will be in UTC.
       # Ex: 2025-04-15
-      def payment_deadline_date(state_code)
-        current_filing_year = MultiTenantService.statefile.current_tax_year.to_i + 1
+      def payment_deadline_date(state_code, filing_year: nil)
```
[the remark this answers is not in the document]
mpidcock: mostly, yeah

PR #5727, opened 2025-03-11
[FYST-1553] Fix links to return to a previous page which were breaking when the language is changed

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1553

## What was done?
- Most states have one eligibility controller and one offboarding controller
  - MD & NJ have two eligibility screening controllers which can both lead to offboarding
    - MD: `md_eligibility_filing_status` & `md_permanent_address` pages
    - NJ: `NjEligibilityHealthInsuranceController`, `NjRetirementWarningController`
  - Less important:
     - NY also does too (we will ignore them for now)
     - AZ does not utilize the offboarding controller at all
- When the language is changed on the `eligibility_offboarding_controller`, the `session[:offboarded_from]` param which saves which controller to go back to is lost (i.e. for NJ, `NjRetirementWarningController` can lead to offboarding page, but it will "go back" to `NjEligibilityHealthInsuranceController` b/c that's the controller before it, which is the default behavior if no `offboarded_from` param exists)
- `session[:offboarded_from]` gets deleted at the offboarding controller when the page renders❗
```
        def set_prev_path
        @prev_path = session.delete(:offboarded_from)
      end
````
- I guess the solution is to _not_ delete it at the offboarding controller (is it a problem if we never delete it?) -- so I'm opting to delete it when the user returns (before `edit` action) to the one of the eligibility controllers, since `session[:offboarded_from]` value can be set again from the `next_path`. It is possible for the client to keep this in their session if they nav away from this screen without going back to one of the eligibility controllers, but if they hit the eligibility screen again, they'll have to re-set this value.

## How to test?
  - Test data used:
    - MD: Frodo hoh to test the permanent address offboarding page; any other persona works for the other eligiblity page
    - NJ: I didn't test these as I'm not certain of which persons to use to get into these offboarding experiences, but did test the controllers & have reasonable expectation they will be fine
    - For all states (except AZ which doesn't have offboarding) make sure that the "Go back to correct" button actually does go back to one of the eligibility pages it came from.
- Risk Assessment
  - After the language changes, the `offboarded_from` data still persists a link back to the original language, instead of the language being currently shown on the page. If there's a way to elegantly change the url of the link, I'll prefer to do that. I'm not even sure exactly what would be desired from product perspective, so probably need a conversation if we want to make it happen.
  - Probably want NJ team to review some of this as I didn't test this locally for their flow.
- Before
**English: see link at the bottom left corner**
Screenshot 2025-03-11 at 10 50 07 PM

**Spanish: see link at the bottom left corner**
Screenshot 2025-03-11 at 10 50 14 PM

- After
**English: see link at the bottom left corner**
Screenshot 2025-03-11 at 10 47 54 PM

**Spanish: see link at the bottom left corner**
Screenshot 2025-03-11 at 10 47 49 PM

### DIFF
--- a/app/lib/navigation/state_file_md_question_navigation.rb
+++ b/app/lib/navigation/state_file_md_question_navigation.rb
@@ -5,6 +5,7 @@ module Navigation
     SECTIONS = [
       Navigation::NavigationSection.new("state_file.navigation.section_1", [
         Navigation::NavigationStep.new(StateFile::Questions::MdEligibilityFilingStatusController),
+        Navigation::NavigationStep.new(StateFile::Questions::EligibilityOffboardingController, false),
         Navigation::NavigationStep.new(StateFile::Questions::EligibleController),
       ]),
       Navigation::NavigationSection.new("state_file.navigation.section_2", [
--- a/spec/controllers/concerns/state_file/eligibility_offboarding_concern_spec.rb
+++ b/spec/controllers/concerns/state_file/eligibility_offboarding_concern_spec.rb
@@ -7,6 +7,8 @@ RSpec.describe StateFile::EligibilityOffboardingConcern, type: :controller do
     def index
       redirect_to next_path
     end
+
+    def edit; end
   end
 
   let(:intake) { create(:state_file_az_intake) }
--- a/spec/controllers/state_file/questions/eligibility_offboarding_controller_spec.rb
+++ b/spec/controllers/state_file/questions/eligibility_offboarding_controller_spec.rb
@@ -61,7 +61,6 @@ RSpec.describe StateFile::Questions::EligibilityOffboardingController do
             get :edit
 
             expect(Nokogiri::HTML.parse(response.body)).to have_link(href: eligibility_controller_path)
-            expect(session[:offboarded_from]).to be_nil
           end
         end
 
@@ -72,7 +71,6 @@ RSpec.describe StateFile::Questions::EligibilityOffboardingController do
             get :edit
 
             expect(Nokogiri::HTML.parse(response.body)).to have_link(href: default_controller_path)
-            expect(session[:offboarded_from]).to be_nil
           end
         end
       end
--- a/spec/controllers/state_file/questions/eligibility_offboarding_controller_spec.rb
+++ b/spec/controllers/state_file/questions/eligibility_offboarding_controller_spec.rb
@@ -36,7 +36,6 @@ RSpec.describe StateFile::Questions::EligibilityOffboardingController do
         get :edit
 
         expect(Nokogiri::HTML.parse(response.body)).to have_link(href: offboarded_from_path)
-        expect(session[:offboarded_from]).to be_nil
       end
     end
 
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: app/lib/navigation/state_file_md_question_navigation.rb [hunk 1]
```diff
@@ -5,6 +5,7 @@ class StateFileMdQuestionNavigation < Navigation::StateFileBaseQuestionNavigatio
     SECTIONS = [
       Navigation::NavigationSection.new("state_file.navigation.section_1", [
         Navigation::NavigationStep.new(StateFile::Questions::MdEligibilityFilingStatusController),
+        Navigation::NavigationStep.new(StateFile::Questions::EligibilityOffboardingController, false),
```
arinchoi03, line 8: my first initial solve was to add this back, but realized that this was not the full extent of the issue. All other state navigations have the EligibilityOffboardingController in the navigation -- in the odd case this controller does not have an `offboarded_from` param, it will default to the `form_navigation.prev` to determine what to set the path for "Go back to correct" link.

spec/controllers/concerns/state_file/eligibility_offboarding_concern_spec.rb
```diff
@@ -7,6 +7,8 @@
     def index
       redirect_to next_path
     end
+
+    def edit; end
```
arinchoi03, line 11: needed since the concern has a before_action on the edit action and if it didn't exist on the mock controller here, it would error

spec/controllers/state_file/questions/eligibility_offboarding_controller_spec.rb
```diff
@@ -36,40 +36,39 @@
         get :edit
 
         expect(Nokogiri::HTML.parse(response.body)).to have_link(href: offboarded_from_path)
-        expect(session[:offboarded_from]).to be_nil
```
arinchoi03, line 39: we no longer delete the `offboarded_from` from session in this controller

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
```diff
@@ -83,4 +83,28 @@
       end
     end
   end
+
+  describe "eligibility_offboarding_concern" do
+    # the disqualifying param here is permanent_address_outside_md but this is set in the form based on:
+    # * whether they are using the address from DF
+    # * if so, whether that address is in MD
+    # so we have to mock the DF data
+    it_behaves_like :eligibility_offboarding_concern, intake_factory: :state_file_nj_intake do
```
anisharamnani, line 92: i don’t think this comment applies here because it’s nj and not md? but i may be wrong.

anisharamnani (approved):
✅ i have one minor comment. navigation is _so_ tricky! thanks for taking this on @arinchoi03

spec/controllers/state_file/questions/nj_retirement_warning_controller_spec.rb
arinchoi03: oops! i thought i deleted the comment from all places but obviously forgot one.

PR #5733, opened 2025-03-12
FYST-1917-add-an-interstitial-screen-to-submission-flow

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1917

## What was done?
- Creates a channel for managing the showing and hiding of the download link button
- Hands the dowloand pdf button until the pdf is ready.
## How to test?
- fill out and intake and watch the button at the end

- Before
- Screenshot 2025-03-17 at 5 25 25 PM

- After
Screenshot 2025-03-17 at 5 00 34 PM


spec/support/helpers/javascript_helpers.rb
```diff
@@ -17,4 +17,10 @@ def wait_for_device_info(form_name)
     end
   end
 
+  def simulate_submission_pdf_ready
```
anisharamnani, line 20: why does NJ need it’s own javascript helper?

anisharamnani (approved):
interesting work! i reviewed this to the best of my ability, but i would wait for another reviewer if possible? i am not very familiar with channels & someone else might have additional feedback.

spec/support/helpers/javascript_helpers.rb
DrewProebstel: I think this is used by every state

spec/channels/state_file_submission_pdf_status_channel_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it "returns ready after the BuildSubmissionPdfJob runs" do
+      subscribe
+      StateFile::BuildSubmissionPdfJob.perform_now(submission.id)
+      intake.reload
+
+      result = perform(:status_update)
+      expect(result).to eq({ status: :ready })
+    end
```
jnf, line 33: how long does this take to run? is it really building a pdf here? would it be reasonable to mock that process instead?

DrewProebstel: Good call!

jnf (approved):
lgtm. appreciate the thorough tests.

spec/channels/state_file_submission_pdf_status_channel_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        expect(result).to eq({ status: :ready })
+      end
+    end
+
+    context "already has a pdf attached" do
+      before do
+        allow(intake.submission_pdf).to receive(:attached?).and_return(true)
+      end
+
+      it "automatically broadcasts `:ready` upon subscription if the PDF is attached" do
```
mrotondo, line 44: could you move this into the above context?

spec/features/state_file/complete_intake_spec.rb
```diff
@@ -155,16 +155,26 @@
...
       wait_for_device_info("esign_declaration")
 
       click_on I18n.t('state_file.questions.esign_declaration.edit.submit')
 
-      expect(page).to have_text I18n.t("state_file.questions.submission_confirmation.edit.title", state_name: "Arizona", filing_year: filing_year)
+      expect(page).not_to have_text I18n.t("state_file.questions.submission_confirmation.edit.title", state_name: "Arizona", filing_year: filing_year)
+      expect(page).not_to have_link I18n.t("state_file.questions.submission_confirmation.edit.download_state_return_pdf")
+      expect(page).to have_text I18n.t("state_file.questions.submission_confirmation.edit.just_a_moment", state_name: "Arizona")
+
+      allow(StateFileSubmissionPdfStatusChannel).to receive(:broadcast_status) do
```
mrotondo, line 170: Why does this (line 170-174) need to be simulated?

app/javascript/packs/state_file.js
```diff
@@ -1,22 +1,62 @@
...
+      received(data) {
+        document.querySelector('[data-after-data-transfer-button]').click()
+      }
+    }
+  },
+  "StateFile::Questions::SubmissionConfirmationController#edit": {
+    channelName: { channel: "StateFileSubmissionPdfStatusChannel" },
+    channelEvents: {
+      initialized () {
+        this.pdfStatus = null
```
mrotondo, line 25: Do you think it would be reasonable to put a timer here that shows the button (and disconnects the channel) after a few (5+) seconds, regardless of what the channel has reported? Most just as a safety measure in case ActionCable fails to connect or the job doesn't get picked up? At worst, clicking the button when the job hasn't completed will result in falling back to the synchronous PDF generation in `StateFile::Questions::SubmissionPdfsController#show`

mrotondo (commented):
I think this change needs one more structural piece, which is to never show the `loading-container` if the intake already has an attached `submission_pdf`. 

Currently, the intended behavior seems to be that we will always show the `loading-container` (which incorrectly tells the user "Just a moment, we’re submitting your Idaho state tax return" even if we already did) until we hear back on the channel that a `submission_pdf` is attached (forcing an update at 500ms). This means that reloading the page will show the spinner and incorrect message for a short amount of time, every time. In my opinion, we shouldn't do this and should instead just show the button immediately if there is an attached `submission_pdf` (& this check should be made in the controller).

More concerningly, it seems like sometimes the cable simply doesn't update the page. On the heroku deployment, by refreshing `/submission-confirmation`, going back to the previous page and then forward again using the browser buttons, or by going back and resubmitting, I can regularly get the page into a state where the spinner is never replaced by the download button. I don't know why this is happening, but it's definitely a situation I want to avoid ever happening in production, so we should do some debugging to figure out what's going on.

mrotondo:
Continuing to look into this locally, and it seems like at least one of the issues I was hitting isn't with this PR, but seems to be with our resubmission logic - I'm not seeing it attach a new submission pdf after detaching the old one. Continuing to investigate and will update further here, but there might be a separate ticket that needs to be created for that issue.

app/javascript/packs/state_file.js
DrewProebstel: that a great idea.

spec/features/state_file/complete_intake_spec.rb
DrewProebstel: it doesn't

mrotondo (approved):
Looks good, thanks! As we discussed in Slack, you should go to acceptance to get that QA rigor but this shouldn't get merged til we've also fixed the resubmit button IMO (willing to be overridden by PM)

PR #5735, opened 2025-03-12
[FYST-1926] Return to income_review if current step was at retirement_income

arinchoi03:
co-authored with @tahsinaislam 

## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1926
- Sentry: https://codeforamerica.sentry.io/issues/6391479077/?referrer=sentry-issues-glance

## What was done?
- Borrowed logic from w2 where if user was at a specific income form edit page, we route them back to the income-review page instead (https://github.com/codeforamerica/vita-min/pull/5551)
- In a similar manner, retirement_income edit/show need an id. I think we get a 500 instead of 404 here b/c of `load_warnings` method which attempts to look up values on a `@state_file1099_r` that is `nil`, although the Sentry error instead throws: `No route matches {:action=>"edit", :controller=>"state_file/questions/retirement_income", :locale=>:en} (ActionController::UrlGenerationError)`
## How to test?
  - Unit/manual tests
  - Test data used: any persona that has 1099R that could be edited/shown (All states except NC can edit/North Carolina can _show_ the income form, if they are on this page at the time of logging out, they should also not be stuck on 500 page)

### DIFF
--- a/app/models/state_file_base_intake.rb
+++ b/app/models/state_file_base_intake.rb
@@ -418,7 +418,7 @@ class StateFileBaseIntake < ApplicationRecord
       StateFile::Questions::ReturnStatusController
     else
       step_name = current_step.split('/').last
-      if step_name == "w2"
+      if ["w2", "retirement-income"].include?(step_name)
         StateFile::Questions::IncomeReviewController
       else
         "StateFile::Questions::#{step_name.underscore.camelize}Controller".constantize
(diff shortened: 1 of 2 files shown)


arinchoi03:
forgot to `git duet` but i paired with @tahsinaislam on this

anisharamnani (approved):
⭐ even though y’all paired, i, anisha, was here.

PR #5737, opened 2025-03-14
FYST 1899 - ID Retirement Benefits Deductions UI Update

tahsinaislam:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-1899

## What was done?
- Altered the logic for the id deductions pages to now look at the specifics of the income source to determine if the income is eligible for the deduction
- Added a new controller and second page for the 1099R Followup flow for idaho where users will be offboarded to if they select that the income source is civil service and the account number ends with 8
- Updated logic for calculations in the id39r form 

## How to test?
  - Updated unit tests 
  - Added new factory traits for 1099Rs for idaho
 - to be tested on heroku

- Before
Screenshot 2025-03-14 at 4 03 50 PM

- After
Screenshot 2025-03-14 at 2 23 25 PM
Screenshot 2025-03-14 at 2 23 20 PM
Screenshot 2025-03-14 at 2 23 12 PM
Screenshot 2025-03-14 at 2 23 05 PM
Screenshot 2025-03-14 at 2 22 57 PM
Screenshot 2025-03-14 at 2 29 27 PM

Screenshot 2025-04-03 at 12 10 42 PM

**Design proposal & writeup of the `ControllerNavigation` changes are [written up in Notion](https://www.notion.so/cfa/Subflows-architecture-brainstorm-1c0373fd79b280ecb2d2e31e2a910f31?showMoveTo=true&saveParent=true)**


app/controllers/state_file/questions/id_ineligible_retirement_and_pension_income_controller.rb
```diff
@@ -0,0 +1,47 @@
+module StateFile
+  module Questions
+    class IdIneligibleRetirementAndPensionIncomeController < RetirementIncomeSubtractionController
+      include OtherOptionsLinksConcern
+      def self.show?(intake)
```
anisharamnani, line 5: why is the `show` method set to false?

spec/controllers/state_file/questions/id_ineligible_retirement_and_pension_income_controller_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  let!(:state_specific_followup) do
+    create :state_file_id1099_r_followup,
+           state_file1099_r: primary_1099r,
+           civil_service_account_number: "eight"
+  end
+
+  before do
+    sign_in intake
+    allow(Flipper).to receive(:enabled?).and_call_original
```
anisharamnani, line 22: 🤔 this controller doesn’t check for the retirement UI flipper. i commented the flipper lines out & the test still passed. 

do you need the flipper to be enabled here? or, can it be skipped because this page is only linked to if they meet certain criteria?

spec/lib/efile/id/id39_r_calculator_spec.rb
```diff
@@ -240,15 +240,47 @@
   end
 
   describe "Section B Line 8e: Retirement Benefits" do
-    context "when eligible retirement benefits exist" do
-      let(:intake) { create(:state_file_id_intake, :with_eligible_1099r_income) }
+    context "when eligible civil service retirement benefits exist" do
```
anisharamnani, line 243: 🤔 i think all of these tests are only for the positive cases where the client qualifies. what about the case where they don’t?

anisharamnani (approved):
i  _think_ my comments are fairly minor except i want to clarify that the `show?` method for  the `id_ineligible_retirement_and_pension_income_controller.rb` 

i believe that method should be checking if the intake is ineligible for the retriement & pension income? but, i don’t see that happening. i may be misunderstanding and the check is happening elsewhere. 

also, i wanted to add that this is very impressive! you put this PR together very quickly and it’s still covering all of our bases. 🎉

config/locales/en.yml
```diff
@@ -3066,13 +3086,26 @@ en:
...
+          none_apply: None of these were true
+          police_officer: Service as an Idaho city Police officer
+          police_officer_follow_up: Were any of the following true for your retirement income as a police officer of an Idaho city?
+          police_persi: Benefits were paid out of the Public Employee Retirement System of Idaho (PERSI)
+          police_retirement_fund: Benefits were paid from the Policemen’s Retirement Fund that no longer admits new members as of Jan 2012
+          question_html: 'Select the source of this retirement income:'
+          seven_or_nine: 7 or 9
+          subtitle: Some 1099-Rs with a taxable amount listed may qualify for this deduction. We need more information about this 1099-R to check your eligibility and file your state return.
+          subtitle_2: If you reported multiple 1099-Rs, you’ll be asked to complete this step for each one with a taxable amount.
+          zero_to_four: 0, 1, 2, 3, 4
```
arinchoi03, line 3108: why is this needing a translation?

spec/features/state_file/review_page_spec.rb
```diff
@@ -416,7 +416,7 @@
             expect(page).to have_text I18n.t("state_file.questions.id_retirement_and_pension_income.edit.subtitle")
             expect(page).to have_text("Couch Potato Cafe")
             expect(page).to have_text("$50")
-            choose "No"
+            choose "None of the above"
```
arinchoi03, line 419: can we add an example where we choose the softboarding option?

app/controllers/state_file/questions/id_retirement_and_pension_income_controller.rb
```diff
@@ -5,7 +5,23 @@ def self.show?(intake)
...
+        0
+      end
+
+      def prev_question_controller_class
+        IdIneligibleRetirementAndPensionIncomeController
+      end
+
+      def next_question_controller_class
+        IdIneligibleRetirementAndPensionIncomeController
+      end
```
arinchoi03, line 24: i'm confused by this & maybe i need just a walkthrough, are we saying the previous controller and next controller are `IdIneligibleRetirementAndPensionIncomeController` if your'e in the `IdRetirmenetAndPensionIncomeController`? What if you're at the **first** `IdRetirmenetAndPensionIncomeController`, what should the `prev_question_controller_class` be?

app/controllers/state_file/questions/id_ineligible_retirement_and_pension_income_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+        false
+      end
+
+      def edit
+        unless @state_file_1099r.state_specific_followup.civil_service_account_number_eight?
```
arinchoi03, line 10: feel like this logic should be in the `show?` method instead of in the edit

spec/controllers/state_file/questions/id_ineligible_retirement_and_pension_income_controller_spec.rb
tahsinaislam: good catch i took this out

spec/features/state_file/review_page_spec.rb
tahsinaislam: this is a great idea. i implemented it

app/views/state_file/questions/shared/_id_retirement_income_deductions_review_header.html.erb
```diff
@@ -1,59 +1,59 @@
-<% if current_intake.calculator.line_or_zero(:ID39R_B_LINE_8e).positive? %>
+<% if current_intake.eligible_1099rs.length.positive? %>
```
anisharamnani, line 1: i believe you can do `.present?` here, but very minor and only for readability.

app/models/state_file_id1099_r_followup.rb
```diff
@@ -2,14 +2,38 @@
...
 
   enum eligible_income_source: { unfilled: 0, yes: 1, no: 2}, _prefix: :eligible_income_source
+  enum income_source: { unfilled: 0, civil_service_employee: 1, police_officer: 2, firefighter: 3, military: 4, none: 5}, _prefix: :income_source
+  enum civil_service_account_number: { unfilled: 0, zero_to_four: 1, seven_or_nine: 2, eight: 3}, _prefix: :civil_service_account_number
+  enum police_retirement_fund: { unfilled: 0, yes: 1, no: 2}, _prefix: :police_retirement_fund
+  enum police_persi: { unfilled: 0, yes: 1, no: 2}, _prefix: :police_persi
+  enum firefighter_frf: { unfilled: 0, yes: 1, no: 2}, _prefix: :firefighter_frf
+  enum firefighter_persi: { unfilled: 0, yes: 1, no: 2}, _prefix: :firefighter_persi
+
+  def qualifying_retirement_income?
```
anisharamnani, line 28: i _think_ you may have moved this here, but not updated the calculator or tests to include this shift. 

i know this has been kind of a big PR, & idk what your energy levels are like since these are a lot of changes, but i’m happy to pair on using this method and testing it if that is helpful.

app/controllers/state_file/questions/id_retirement_and_pension_income_controller.rb
tahsinaislam: when youre at the first 1099R it will be the previous controller before it. its hard to see how this works but its more a part of the repeated questions concern

app/lib/efile/id/id39_r_calculator.rb
```diff
@@ -86,14 +86,29 @@ def calculate_sec_b_line_8d
...
-            form1099r.taxable_amount.round
-          else
-            0
-          end
+          next 0 unless qualifying_retirement_income?(form1099r)
+          form1099r.taxable_amount.round
         end
       end
 
+      def qualifying_retirement_income?(form)
```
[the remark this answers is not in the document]
arinchoi03: I'd say rename to `state_file1099_r` or something more representative of exactly what the object is

app/models/state_file_id1099_r_followup.rb
tahsinaislam: done!

spec/lib/efile/id/id39_r_calculator_spec.rb
tahsinaislam: i did have a test for this but now that i moved the method to the model decided to keep old test as is that test qualified very unqualified 1099rs

app/lib/efile/id/id39_r_calculator.rb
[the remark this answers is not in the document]
tahsinaislam: this as been moved to the model!

app/views/state_file/questions/shared/_id_retirement_income_deductions_review_header.html.erb
tahsinaislam: done!

app/controllers/state_file/questions/retirement_income_subtraction_controller.rb
```diff
@@ -15,28 +19,28 @@ def initialized_update_form
...
-        if @state_file_1099r.nil?
+        @state_file_1099r = self.class.load_1099r(current_intake, index)
+        if state_file_1099r.nil?
           render "public_pages/page_not_found", status: 404
         end
       end
 
+      def self.load_1099r(intake, index)
+        intake.eligible_1099rs[index]
+      end
```
arinchoi03, line 37: why is this a class method?

app/lib/navigation/state_file_nj_question_navigation.rb
```diff
@@ -75,6 +78,13 @@ def self.show_progress?(controller_class)
...
     ].freeze
-    FLOW = SECTIONS.map(&:controllers).flatten.freeze
+
+    def self.controllers
+      sections.flat_map(&:controllers)
+    end
+
+    def self.pages(visitor_record)
+      sections.flat_map { |section| section.pages(visitor_record) }
+    end
```
arinchoi03, line 88: can the methods be refactored to be defined once or is this b/c we need these to be class methods specifically on these navigations?

app/models/state_file_az_intake.rb
```diff
@@ -147,6 +147,8 @@ def disqualifying_eligibility_rules
   end
 
   def disqualified_from_excise_credit_df?
+    return false if raw_direct_file_data.blank?
```
arinchoi03, line 150: is this a fix for something else? would have been nice to split this off into its own story

app/lib/concerns/controller_navigation.rb
```diff
@@ -1,54 +1,134 @@
...
+      # We should not be using arity to decide how to call `show?`
+      case controller_class.method(:show?).arity
+      when 2
         controller_class.show?(
           controller_class.model_for_show_check(current_controller),
-          @current_controller
+          current_controller
         )
-      else
+      when -1, 1
```
arinchoi03, line 121: noting for myself that `-1` means variable number or arguments, but one is required

app/lib/concerns/controller_navigation.rb
```diff
@@ -1,54 +1,134 @@
...
         controller_class.show?(
           controller_class.model_for_show_check(current_controller)
         )
+      when -2
```
arinchoi03, line 125: `IdIneligibleRetirementAndPensionIncomeController` is one example of `-2` arity? why isn't it `2`?
`IdRetirementAndPensionIncomeController` is another one.

app/views/state_file/questions/shared/_id_retirement_income_deductions_review_header.html.erb
```diff
@@ -1,13 +1,84 @@
-<% if current_intake.calculator.line_or_zero(:ID39R_B_LINE_8e).positive? %>
+<% if current_intake.show_disability_question? %>
+  <section id="disability-info" class="white-group">
+    <p class="text--bold spacing-below-5"><%= t("state_file.questions.shared.id_disability_review_header.meets_qualifications") %></p>
+    <p class="spacing-below-15"><%= (current_intake.primary_disabled_yes? || current_intake.spouse_disabled_yes?) ? t("general.affirmative") : t("general.negative") %></p>
+    <p class="text--small spacing-below-15"><%= t("state_file.questions.shared.id_disability_review_header.editing_may_affect_deductions") %></p>
+    <div class="spacing-below-5">
+      <%= link_to t("general.review_and_edit"), StateFile::Questions::IdDisabilityController.to_path_helper(
+        return_to_review_before: StateFile::Questions::IdDisabilityController.name.demodulize.underscore,
+        return_to_review_after: "retirement_income_deduction"), class: "button--small" %>
```
arinchoi03, line 9: the way I understand this after looking at it -- `return_to_review_before` and `return_to_review_after` are kind of bookend markers: if you reach at this controller when you click to go prev/back, return to review & if you reach end of "retirement_income_deduction" when you click to proceed/forward, return to review.

arinchoi03 (commented):
I've looked over it and still wrapping my head around it, and this definitely requires more eyes than mine -- i had some initial questions about when you choose class (`self.`) methods over instance.

app/lib/concerns/controller_navigation.rb
mrotondo: This is a refactor I'd like to make outside the scope of this PR, but the gist is that arity becomes negative when there are optional parameters - like the negative sign is like a bit that indicates the presence of optional params

mrotondo: [Returns an indication of the number of arguments accepted by a method. Returns a nonnegative integer for methods that take a fixed number of arguments. For [Ruby](https://apidock.com/ruby/RDoc/Parser/Ruby) methods that take a variable number of arguments, returns -n-1, where n is the number of required arguments. Keyword arguments will be considered as a single additional argument, that argument being mandatory if any keyword argument is mandatory. For methods written in C, returns -1 if the [call](https://apidock.com/ruby/Method/call) takes a variable number of arguments.](https://apidock.com/ruby/Method/arity)

app/models/state_file_az_intake.rb
mrotondo: Sorry - yes, there was a time when this PR called `show?` on _every_ controller in the flow, rather than just until it finds one that can be shown. That surfaced that we have a bunch of controllers whose `show?` methods crash when intakes are in certain states (typically when there's no DF data yet). I changed how `show?` is called, so we no longer hit these crashes (because I didn't want to also fix them across GYR & CTC) but I left these fixes in since we _shouldn't_ crash on any input to these methods

app/lib/navigation/state_file_nj_question_navigation.rb
mrotondo: these have to be here so that they're defined _after_ (and therefore override) the versions defined in the module that's included at the top of this file. barring a larger refactor, they have to be here

app/controllers/state_file/questions/retirement_income_subtraction_controller.rb
mrotondo: so that it can be called by `self.show?`, which (in the case of `IdIneligibleRetirementAndPensionIncomeController`) needs to grab the correct 1099-R to determine whether the associated followup indicates that we should show the page.

app/controllers/state_file/questions/retirement_income_subtraction_controller.rb
```diff
@@ -5,6 +5,10 @@ class RetirementIncomeSubtractionController < QuestionsController
```
DrewProebstel, line 5: Do we still need to include `RepeatedQuestionConcern`

mrotondo: We don't _need_ it, no - we could move that code into this file, since it's the only place we _currently_ use `RepeatedMultiPageStep`. My reason not for doing so is that while it's a tiny amount of code, it'll hopefully be a helpful starting point (that'll help us avoid some code duplication) for whatever controller we create next that isn't 1099R-related but is used within a `RepeatedMultiPageStep`. Do you have an idea for an alternative approach, or a reason you'd like to see it removed?

mpidcock (commented):
Struggling to approve since I don't feel certain I grasp all the impacts, but don't let that stop acceptance

app/controllers/state_file/questions/retirement_income_subtraction_controller.rb
DrewProebstel: That makes sense. I think it is fine here.

app/controllers/ctc/questions/questions_controller.rb
```diff
@@ -37,23 +37,27 @@ def parent_class
...
 
       def next_path
-        next_step = form_navigation.next
+        next_page_info = form_navigation.next
+        return unless next_page_info.present?
+        next_page_controller = next_page_info[:controller]
+
         options = {}
-        if next_step.resource_name.present? && next_step.resource_name == self.class.resource_name
+        if next_page_controller.resource_name.present? && next_page_controller.resource_name == self.class.resource_name
```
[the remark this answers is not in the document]
mrotondo: I think you're suggesting that I replace `next_page_controller.resource_name.present?` with `next_page_controller&.resource_name` - if not, please ignore my response :)

The `&.` operator provides safety in the case that the thing on the left of it is nil, which you're right that we should be protecting against here (though GetCTC is down, so it shouldn't matter). I'll add that check above here, as I did in the other `QuestionsController#next_path` overrides. However, I'm not going to remove the `#present?` call and just leave it as a nil check (which is what `if _some_variable_` implicitly does), since they don't do the same thing: `#present?` returns false on values like empty strings, which would be evaluated as `true` if we just put the value itself in the `if` clause.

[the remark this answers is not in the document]
mrotondo: Reading the line more closely, I see that because we do check the value of `resource_name` against something else later on this line, the fact that empty strings would make it through the first clause of the `if` is fine. Sorry for missing that earlier! I'll make the suggested change here, as well as adding the nil check for `next_page_controller` above.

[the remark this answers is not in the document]
mrotondo: Alright, I found a way to combine the nil check for page info, the presence of the `:controller` key, and the nil check on the value associated with the `:controller` key into one clause that will lead to an early return, added some more tests, and removed the unnecessary `present?` calls in the lines you pointed out.

app/controllers/state_file/questions/questions_controller.rb
```diff
@@ -15,6 +15,10 @@ def ip_for_irs
         end
       end
 
+      def review_controller
+        "StateFile::Questions::#{current_state_code.titleize}ReviewController".constantize
```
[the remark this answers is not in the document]
mrotondo: good call - that's much nicer. switched over to using it

PR #5738, opened 2025-03-14
NJ 311 - estimated tax payments redesign

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/311

## What was done?
- Redesign estimated tax payments page - switch to yes/no radio button and only accept numeric inputs if they choose yes
- Add overpayments input
- Update review page with new yes/no style and amount
- Update line 57 calculation to sum estimated payments + overpayments

## How to test?
Any persona

image
image
image


config/locales/en.yml
```diff
@@ -3522,23 +3522,28 @@ en:
...
-                <ul class="list--bulleted">
-                  <li><strong>This does not include withholdings in your paycheck</strong></li>
-                </ul>
-              </li>
-              <li>Overpayments on your %{prior_year} return that you chose to apply to your %{filing_year} New Jersey taxes
-                <ul class="list--bulleted">
-                  <li>Note: you would have selected this option when filing your %{prior_year} return</li>
-                </ul>
-              </li>
+              <li><a href="https://www.nj.gov/treasury/taxation/njit20.shtml" target="_blank" rel="noopener nofollow">Estimated quarterly payments</a> that you made. These were due on April 15, June 16, and September 15 last year and January 15 this year. You would have paid these <a href="https://www1.state.nj.us/TYTR_RevTaxPortal/jsp/IndTaxLoginJsp.jsp" target="_blank" rel="noopener nofollow">online</a> or by mail.</li>
```
mluedke2, line 3530: [dust] I wonder if the `April 15, June 16, and September 15 last year and January 15 this year` dates are specific to this year or always the same? If they are likely to change maybe the string should have `2024` in the key to make it clear to revisit this string next year? but if it's low chance of change then nevermind

mluedke2 (approved):
looks good to me!

app/forms/state_file/nj_estimated_tax_payments_form.rb
```diff
@@ -1,10 +1,15 @@
...
+                       :has_estimated_payments,
+                       :estimated_tax_payments,
+                       :overpayments
 
+    validates :has_estimated_payments, presence: true
     validates_numericality_of :estimated_tax_payments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
-    validates :estimated_tax_payments, presence: true, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }
+    validates_numericality_of :overpayments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
+    validates :estimated_tax_payments, presence: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
+    validates :overpayments, presence: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
```
jenny-heath, line 12: seems odd that this is required to be greater than 0 even though the copy says "if applicable"?

app/forms/state_file/nj_estimated_tax_payments_form.rb
```diff
@@ -1,10 +1,15 @@
...
     set_attributes_for :intake,
-                       :estimated_tax_payments
+                       :has_estimated_payments,
+                       :estimated_tax_payments,
+                       :overpayments
 
+    validates :has_estimated_payments, presence: true
     validates_numericality_of :estimated_tax_payments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
-    validates :estimated_tax_payments, presence: true, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }
+    validates_numericality_of :overpayments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
```
jenny-heath, line 10: why does this only need to be a number if they had estimated tax payments?

app/forms/state_file/nj_estimated_tax_payments_form.rb
aloverso: `0` is allowed — it's greater than or equal to. Or are you saying the field should be optional altogether?

app/forms/state_file/nj_estimated_tax_payments_form.rb
aloverso: If they answer `No` to the radio question, we don't need to validate anything about this field at all, because the user won't see it and doesn't fill it in. Behaves pretty similarly to Use Tax

aloverso: Sorry, no, I misread this. This is actually a typo. My bad! Will fix, thanks.

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -477,7 +477,8 @@ def calculate_line_56
       end
 
       def calculate_line_57
-        @intake.estimated_tax_payments&.round
+        return nil if @intake.estimated_tax_payments.nil? && @intake.overpayments.nil?
```
jachan, line 480: [PEBBLE/BOULDER?] What happens when the user fills in `estimated_tax_payments` AND `overpayments`, and then changes their answer to `has_estimated_payments`? Do we need to add logic either in the form or here to treat the fields like `nil` if `has_estimated_payments` is No?

jachan (commented):
One blocking question about what happens when user changes their answer to `has_estimated_payments` from `yes` to `no`. Otherwise looks good!

app/forms/state_file/nj_estimated_tax_payments_form.rb
jenny-heath: ah yeah i guess i did mean "do they need to write 0" but i appreciate the context, john! no further questions

spec/lib/efile/nj/nj1040_calculator_spec.rb
```diff
@@ -1781,14 +1781,28 @@ def over_65_birth_year
...
-        expect(instance.lines[:NJ1040_LINE_57].value).to eq 401
+      let(:intake) { create(:state_file_nj_intake, estimated_tax_payments: 400.44, overpayments: 400.44 )}
+      it 'sets line 57 to the rounded sum of estimated_tax_payments and overpayments' do
+        expect(instance.lines[:NJ1040_LINE_57].value).to eq 801
       end
     end
 
-    context 'when estimated_tax_payments is nil' do
-      let(:intake) { create(:state_file_nj_intake, estimated_tax_payments: nil)}
+    context 'when estimated_tax_payments is nil but overpayments exist' do
```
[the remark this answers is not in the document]
aloverso: For one, I think there's a possibility for existing users who filled out the form last week, say, and come back this week after this PR goes out. We want to handle that case. For other, I think that if the value CAN be nil, we should elegantly handle it in the calculator. Even if the validations make that state not possible for the user, it could happen in the code and better to have a check for it than to have a runtime error

spec/forms/state_file/nj_estimated_tax_payments_form_spec.rb
```diff
@@ -6,28 +6,91 @@
...
+            has_estimated_payments: "yes",
+            overpayments: nil,
+            estimated_tax_payments: 0
+          }
+        end
+
+        it "is invalid" do
+          expect(form.valid?).to eq false
+          expect(form.errors[:overpayments]).to include "Can't be blank."
+        end
```
jenny-heath, line 68: [pebble] pretty sure these are redundant with the shared example `nj_money_field_concern`

app/lib/efile/nj/nj1040_calculator.rb
aloverso: Hmm... so this actually DOES work, the fields do get set back to nil when you change your answer to No. The thing is, I'm not actually sure _why_ that behavior is happening by default?

spec/forms/state_file/nj_estimated_tax_payments_form_spec.rb
jenny-heath, line 1: [small boulder] this file doesn't appear to cover the `has_estimated_payments == "yes"` condition (i.e. if it's "no", the other fields are allowed to be missing). would like to see a test for that

app/lib/efile/nj/nj1040_calculator.rb
tristanvanech: this was one of the test cases and it seems to be handling it well!

> yes, enter amount, edit to no from Review + Edit screen, amount isn't included in calculations (like Use Tax)

jenny-heath: the question with follow-up classes in the view should make it so that if you change your answer it will wipe out any answers to the follow-up question. so, when they submit the form with a "no" answer it should overwrite the other fields with nil

app/lib/efile/nj/nj1040_calculator.rb
```diff
@@ -477,7 +477,8 @@ def calculate_line_56
...
+        ((@intake.estimated_tax_payments || 0) + (@intake.overpayments || 0)).round
```
jenny-heath, line 481: [pebble] if they answered "yes" to `has_estimated_payments`, they're required to at least type in 0 for both these fields, right? so there shouldn't be a case where only one of them is `nil`. might be clearer to actually reference `has_estimated_payments` in this method, like this:
```
if @intake.has_estimated_payments
  (@intake.estimated_tax_payments + @intake.overpayments).round
end
```

jenny-heath: actually, per [this comment](https://github.com/codeforamerica/vita-min/pull/5738#discussion_r2002039912), i take it back

jenny-heath (approved):
couple of comments re: testing. happy to re-review those before merge if that's helpful!

PR #5739, opened 2025-03-14
FYST-1767 remove code that handles people who were on the old version of the school credits page

anisharamnani:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-1767

## What was done?

backfilling the old column data into the new columns. for any az intake where made_az322_contributions is "unfilled"
* when az322_contribution record has a value of "yes" for made_contribution, made_az322_contributions on intake should be updated to "yes". All of the az322_contribution will have a value of yes since they are only created when the value is yes. 
* every intake that was created before [this pull request](https://github.com/codeforamerica/vita-min/pull/5513/files) (and has an efile submission, as a proxy for “saw the public school page”) that has no az322_contributions associated should have made_az322_contribution set to "no"

## How to test?
* running the queries on production: 34 intakes have az322_contributions, 1431 do not


db/data/20250312183451_migrate_made_contribution_az322_contribution_to_made_az322_contributions_on_state_file_az_intake.rb
```diff
@@ -0,0 +1,34 @@
...
+    Sentry.capture_message "There are #{az_intakes_with_contributions_created_before_feb_7.count} intakes created before February 7, 2025 at 6:26 PM EST that have az322 contributions"
+    az_intakes_with_contributions_created_before_feb_7.update_all(made_az322_contributions: 1)
+    Sentry.capture_message "There are #{az_intakes_with_contributions_created_before_feb_7.count} intakes created before February 7, 2025 that have been updated to register az322 contributions"
+
+    submitted_az_intakes_without_contributions_created_before_feb_7 = StateFileAzIntake
+                                                      .left_joins(:efile_submissions)
+                                                      .left_joins(:az322_contributions)
+                                                      .where.not(efile_submissions: { id: nil })
+                                                      .where('state_file_az_intakes.created_at < ?', time_of_release_of_contribution_update)
+                                                      .where(made_az322_contributions: 0)
```
jenny-heath, line 23: since this is identical to the first part of the other query (i think?) you could save it off to a variable like so:
```
submitted_az_intakes_created_before_feb_7 = StateFileAzIntake
                                                      .left_joins(:efile_submissions)
                                                      .left_joins(:az322_contributions)
                                                      .where.not(efile_submissions: { id: nil })
                                                      .where('state_file_az_intakes.created_at < ?', time_of_release_of_contribution_update)
                                                      .where(made_az322_contributions: 0)
```

and then do
```
with_contributions = submitted_az_intakes_created_before_feb_7.where.not(az322_contributions: { id: nil })
without_contributions = submitted_az_intakes_created_before_feb_7.where(az322_contributions: { id: nil })
```

and i think that makes it a little easier to read?

jenny-heath: not 100% sure that works so would have to try it

jenny-heath (approved):
i think (???) we tried these queries on production and there were very few records, right? if so, this looks good but i would try it locally and on demo (checking the records before and after). if i've misremembered the small number of records, it should not be an `update_all` but do something in batches instead.

tbh i feel like the really responsible thing to do would be write a test? but i always feel torn about this for one-offs and mike and i didn't write a test for our recent data migration (haven't run it yet) so it would feel unfair to make you write one

db/data/20250312183451_migrate_made_contribution_az322_contribution_to_made_az322_contributions_on_state_file_az_intake.rb
anisharamnani: there are two differences! 
1. the first also does not check if there is an efile submission. this because regardless if the intake has an efile submission, if they do have az322_contributions, then the answer should be set to yes. i am going to update the first query to not join efile submissions since the submission status doesn't reflect that. 
2. & the second is if they do have az contributions.

with that said, i will try to pull out the similarities to make the differences clearer if possible.

anisharamnani:
running the queries on production: 34 intakes have az322_contributions, 1431 do not, so i don't think batching is needed. 

however, i will run this on demo first just to make sure there are no syntax errors & to check that the records have been updated.

PR #5740, opened 2025-03-14
Nj 320 - Convert w2 box 16 error to warning

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/320

## What was done?
- For NJ, allow state wage to be 0 even if federal wages are nonzero, as long as user visits the W2 edit screen

## How to test?
- Use bad W2s persona
- Validate that you cannot proceed past income review before editing individual w2s

- Before
- After

### DIFF
--- a/app/controllers/state_file/questions/w2_controller.rb
+++ b/app/controllers/state_file/questions/w2_controller.rb
@@ -15,6 +15,11 @@ module StateFile
       end
 
       def edit
+        if StateFile::StateInformationService.check_box_16(current_state_code)
+          current_intake.confirmed_w2_indexes.append(@w2.w2_index)
+          current_intake.save
+        end
+
         @w2.valid?(:state_file_edit)
       end
 
@@ -23,11 +28,6 @@ module StateFile
       def update
         @w2.assign_attributes(form_params)
         @w2.check_box14_limits = true
-
-        if StateFile::StateInformationService.check_box_16(current_state_code)
-          current_intake.confirmed_w2_indexes.append(@w2.w2_index)
-          current_intake.save
-        end
 
         if @w2.valid?(:state_file_edit)
           @w2.box14_ui_hc_wd = nil
--- a/app/controllers/state_file/questions/w2_controller.rb
+++ b/app/controllers/state_file/questions/w2_controller.rb
@@ -23,10 +23,13 @@ module StateFile
       def update
         @w2.assign_attributes(form_params)
         @w2.check_box14_limits = true
+        @w2.taxpayer_reviewed = true
+        @w2.errors.clear
 
         if @w2.valid?(:state_file_edit)
           @w2.box14_ui_hc_wd = nil
-          @w2.save(context: :state_file_edit)
+          @w2.save(context: [:state_file_edit, :state_file_income_review])
+          binding.pry # w2 is valid in both contexts here, taxpayer_reviewed is true
           redirect_to StateFile::Questions::IncomeReviewController.to_path_helper(return_to_review: params[:return_to_review])
         else
           render :edit
--- a/app/controllers/state_file/questions/w2_controller.rb
+++ b/app/controllers/state_file/questions/w2_controller.rb
@@ -31,7 +31,7 @@ module StateFile
 
         if @w2.valid?(:state_file_edit)
           @w2.box14_ui_hc_wd = nil
-          @w2.save(context: [:state_file_edit, :state_file_income_review])
+          @w2.save(context: :state_file_edit)
           redirect_to StateFile::Questions::IncomeReviewController.to_path_helper(return_to_review: params[:return_to_review])
         else
           render :edit
--- a/app/controllers/state_file/questions/w2_controller.rb
+++ b/app/controllers/state_file/questions/w2_controller.rb
@@ -4,7 +4,7 @@ module StateFile
       before_action :allows_w2_editing, only: [:edit, :update]
       before_action :load_w2
 
-      helper_method :box_14_codes_and_values
+      helper_method :box_14_codes_and_values, :state_wages_invalid?
 
       def self.show?(intake) # only accessed via button, not navigator
         false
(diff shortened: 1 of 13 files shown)


config/locales/en.yml
```diff
@@ -4292,6 +4292,7 @@ en:
           box14_ui_wf_swf_html: "<strong>UI/WF/SWF</strong>"
           box15_html: "<strong>Box 15,</strong> Employer’s state ID number"
           box16_html: "<strong>Box 16,</strong> State wages, tips, etc."
+          box16_warning_nj: '⚠️ Warning: If this is incorrect, you could end up underpaying state taxes. If box 16 has a number in it, enter that number here. If it''s blank or 0, click "Chat with us" and let''s make sure this is correct.'
```
arinchoi03, line 4295: [pebble] double `''` here in "let's"

app/controllers/state_file/questions/w2_controller.rb
```diff
@@ -15,6 +15,11 @@ def prev_path
       end
 
       def edit
+        if StateFile::StateInformationService.check_box_16(current_state_code)
+          current_intake.confirmed_w2_ids.append(@w2.id)
```
arinchoi03, line 19: [pebble] if this page gets visited multiple times, will you add the id multiple times into the column as well? I think we should not add if it already exists

spec/controllers/state_file/questions/w2_controller_spec.rb
```diff
@@ -75,6 +75,74 @@
...
+          
+        end
+
+        it "adds w2 index to the list of verified W2s and removes the state_wages_amount error when user opens the page" do
+          expect(state_file_w2.valid?(:state_file_edit)).to eq false
+          expect(intake.confirmed_w2_ids.size).to eq 0
+          expect(state_file_w2.errors).to include(:state_wages_amount)
+          get :edit, params: params
+          state_file_w2.reload
+          state_file_w2.valid?(:state_file_edit)
```
arinchoi03, line 95: [q] should `expect(state_file_w2.valid?(:state_file_edit)).to eq true` here or are other fields invalid for this w2? not a huge deal but wanted to clarify

spec/models/state_file_nj_intake_spec.rb
```diff
@@ -295,15 +296,32 @@
...
-      w2.state_wages_amount = 1000000
-      intake.validate_state_specific_w2_requirements(w2)
-      expect(w2).to be_valid
-      expect(w2.errors[:state_wages_amount]).not_to be_present
+      it "does not permit state_wages_amount to be 0 if federal wages is non-zero" do
+        intake.confirmed_w2_ids = []
+        w2.state_wages_amount = 0
+        intake.validate_state_specific_w2_requirements(w2)
+        expect(w2.errors[:state_wages_amount]).to be_present
+        expect(w2.valid?(:state_file_edit)).to eq false
```
arinchoi03, line 310: [dust] noticing we use two different contexts (`:state_file_edit`, `:state_file_income_review`) for validating here (above and below rspec contexts), should it be `false` no matter what context is used?

We should maybe test both contexts if it doesn't matter which one they're in

arinchoi03 (commented):
couple pebbles & we probably all need to chat about the upcoming income review at some point 😅

arinchoi03:
also need to i18n normalize for the green build!

config/locales/en.yml
jachan: image

Renders properly on the heroku app! `''` serves as an escape sequence for `'`. Recognize that this can be confusing, though, happy to change back to `/'` to escape if its unclear.

spec/controllers/state_file/questions/w2_controller_spec.rb
jachan: Other fields are invalid since the default `box14_fli` and `box14_ui` fields are 0. Didn't want to tie the test to other validation rules, but calling `state_file_w2.valid?(:state_file_edit)` is necessary to recalculation what the `state_file_w2.errors` are here.

app/controllers/state_file/questions/w2_controller.rb
```diff
@@ -15,6 +15,11 @@ def prev_path
...
+          current_intake.save
+        end
```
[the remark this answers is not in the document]
jachan: Yeah, I imagine we will need to refactor with the W2 design changes, this was meant to be a fix for the P2 blocking issue. Let's definitely do a joint design review of the new W2 changes from an eng perspective and make sure we're aligned on implementation before beginning the work!

spec/models/state_file_nj_intake_spec.rb
jachan: Added an assert here below line 310 to `expect(w2.valid?(:state_file_income_review)).to eq false`. 

However, checking the `state_file_edit` context below on line 323 shows that while `expect(w2.valid?(:state_file_income_review)).to eq true` passes, `expect(w2.valid?(:state_file_edit)).to eq true` fails due to box14 validations. Is it alright to leave the validation as is? Don't want to include setup for box 14 validations in this test which is meant to be a box 16 feature.

`[#<ActiveModel::Error attribute=box14_ui_wf_swf, type=Please enter an amount. If amount is blank, enter $0., options={}>, #<ActiveModel::Error attribute=box14_fli, type=Please enter an amount. If amount is blank, enter $0., options={}>]`

config/locales/en.yml
arinchoi03: ooh lol i didn't realize it worked that way -- it's fine

spec/models/state_file_nj_intake_spec.rb
arinchoi03: hmm i get that but since we're honing in on the box16 validation, i'd prefer if everything else was valid. I think it's confusing to read this test and see that you're validating two contexts in one test & not both in the other?
If it's truly difficult to add the box14 values, I'm fine with leaving

arinchoi03 (approved):

spec/models/state_file_nj_intake_spec.rb
jachan: What you said about test clarity convinced me! It's a pretty simple fix. Went ahead and added box14 values to the intake above.

PR #5742, opened 2025-03-15
[TBE-141] Fix flaky test that depended on run order

jnf:
## Link to pivotal/JIRA issue
[TBE-141](https://codeforamerica.atlassian.net/browse/TBE-141)

## What was done?
- tweaks how we look for existing intakes at login to only try to match intake id if the intake state is the same
- some tidying 🧹  

## ytho
There are 3 context blocks in the tests for #update, and each does a little bit of setup, creating 2 intakes (the details of which vary by context). After each context block finishes, the intakes created for it are destroyed so that the database is in a “clean” state for the next context block.

I put “clean” in quotes because there is a side-effect to how rspec is making and removing the intakes, and that side effect bleeds between contexts, causing this inconsistent behavior. The tl;dr is that while the intakes are destroyed, the indexes are not reset, therefore when intakes are inserted in the next context block, the database ids for those records continue to increment.

So, when you run the whole suite, the first context block makes AZ1 and AZ2. It deletes them at the end of its tests. When the second context block runs, it makes ID1 and AZ3 (because we delete the records but don’t reset the table indexes). This is important.

All of the tests in the file are eventually calling VerificationCodeController#get_existing_intake. In that method, we attempt to match the current_intake to any pre-existing intakes by comparing the email or phone number (whichever is being used to sign in). To do so we generate a query that starts with intake_class.where.not(id: intake.id)… meaning we’ll not consider an existing intake with the same ID as the current_intake as an additional intake.  That makes sense if you have an Arizona intake and are looking for other Arizona intakes, but this method is intentionally checking every state for potential duplicates.

The failing test is trying to prove that if you have an AZ and ID intake using the same email address, that’s a problem. But it’s failing to match the ID intake if it is the first (or only) test run in the suite. This is because of id collision. If it’s the first/only context run, the two intakes are ID1 and AZ1, so the first clause of the ActiveRecord query (.where.not(id: intake.id)) isn’t going to select the ID1 record (because it and the AZ intake are both id 1). If this context block isn’t run first, it will create intakes ID1 and AZn (where n is however many AZ intakes were created in the contexts run previously). In that case, the query will correctly select the ID record and the test will pass.

This is bananas and I love it. We need to refactor #get_existing_intake to be smarter about comparing records between states. It’s too friday for me to know what that looks like at the moment, but i bet on monday we can figure it out.

## How to test?
- Green build
- Able to log in to an existing intake when one exists

[TBE-141]: https://codeforamerica.atlassian.net/browse/TBE-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anisharamnani (approved):
⭐ i appreciate the comments here since login logic is tricky to follow.

jenny-heath (approved):
pair reviewed, looks great!

PR #5743, opened 2025-03-17
[TBE-141] Save the matching intakes search as it's getting refined

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-141

## What was done?
- Save off the query as it's getting more refined; i think what was happening was that the search was just returning the first intake that does not have a direct file data, and in the specs this wasn't surfacing b/c we clean up the intake every time the tests run
- The Oops error occurs after `validate_token` runs and redirects to `:create` action (no idea _why_ create action runs, instead of a more elegant offboarding -- it might be that the `ClientLoginsController` is kind of from legacy code from CTC?)
## How to test?
  - We should revamp the tests to expose the issue better in the controller specs.. to fix up demo I think we need to patch it to unblock releases (as i know NJ wants to release today); maybe make a followup story. I added a `existing_intake_with_df_data` in the second commit of this PR to demonstrate how the error manifests, but I wonder if we need a more robust re-write
    - funny thing is that even with the re-write, the tests pass in the "match" scenarios (although it matched incorrectly) b/c the login hash is based on the **current** intake, not the existing intake. The test fails for the non-matching one b/c it ends up matching when it doesnt expect to and goes to `login` path instead of the plain `code-verified` path (for intakes with no existing matches)

### DIFF
--- a/app/controllers/state_file/questions/verification_code_controller.rb
+++ b/app/controllers/state_file/questions/verification_code_controller.rb
@@ -55,8 +55,8 @@ module StateFile
         state_intake_classes.each do |intake_class|
           key = intake.contact_preference == "text" ? :phone_number : :email_address
           search = intake_class.where.not(raw_direct_file_data: nil) # has a successful df import...
-          search.where(key => intake[key]) # ...using the same contact info
-          search.where.not(id: intake.id) if intake_class == intake.class # ...unless it's literally the same intake
+          search = search.where(key => intake[key]) # ...using the same contact info
+          search = search.where.not(id: intake.id) if intake_class == intake.class # ...unless it's literally the same intake
 
           existing_intake = search.first
           return existing_intake if existing_intake
--- a/spec/controllers/state_file/questions/verification_code_controller_spec.rb
+++ b/spec/controllers/state_file/questions/verification_code_controller_spec.rb
@@ -75,6 +75,7 @@ RSpec.describe StateFile::Questions::VerificationCodeController do
           )
         )
         expect(response).to redirect_to(login_location)
+        expect(StateFileAzIntake.where(id: intake.id)).to be_empty
       end
     end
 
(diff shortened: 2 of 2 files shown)


app/controllers/state_file/questions/verification_code_controller.rb
```diff
@@ -55,8 +55,8 @@ def get_existing_intake(intake)
         state_intake_classes.each do |intake_class|
           key = intake.contact_preference == "text" ? :phone_number : :email_address
           search = intake_class.where.not(raw_direct_file_data: nil) # has a successful df import...
-          search.where(key => intake[key]) # ...using the same contact info
-          search.where.not(id: intake.id) if intake_class == intake.class # ...unless it's literally the same intake
+          search = search.where(key => intake[key]) # ...using the same contact info
+          search = search.where.not(id: intake.id) if intake_class == intake.class # ...unless it's literally the same intake
```
squanto, line 59: may be worth a spec that the clause is picked up

arinchoi03: hmm yea, i mean i make one of the tests fail (pre-this code change) by adding the `first_existing_intake` in the spec file, but not sure if that's sufficient.

app/controllers/state_file/questions/verification_code_controller.rb
```diff
@@ -55,8 +55,8 @@ def get_existing_intake(intake)
...
 
           existing_intake = search.first
```
mpidcock, line 61: This might be a personal preference, and there might be legitimate reasons to have the code the way it is, but I actually find it more readable and maintainable to combine this search into one whole query. I can try to make an example.

arinchoi03: let's do that in the follow up bug you created to match on contact preference!

app/controllers/state_file/questions/verification_code_controller.rb
arinchoi03: @mrotondo helped me write more tests 🙏

PR #5748, opened 2025-03-18
Added role="button" tags to state_file links

jachan:
## Link to pivotal/JIRA issue
(visible to CFA) https://github.com/newjersey/accessibility-pm/issues/14
(OOI internal) https://github.com/newjersey/affordability-pm/issues/274

## What was done?
- Change all state_file occurences of link_to elements styled as button--wide to have role="button"

## How to test?
- Step through flow, validate that all Continue buttons are either `<input type="submit">` or `<a role="button">` elements.

Videos illustrate how a button="role" element appear to users using a screen reader, and how this creates consistency between <input type="submit"> and <a role="button"> elements

https://github.com/user-attachments/assets/3402315d-be17-4e24-8204-7a9fddc96295

https://github.com/user-attachments/assets/7f3cfdec-773c-4cdd-93b7-6b78a54f076e

### DIFF
--- a/app/views/state_file/questions/return_status/edit.html.erb
+++ b/app/views/state_file/questions/return_status/edit.html.erb
@@ -7,7 +7,7 @@
   <% end %>
 
   <% if !@error&.auto_wait || app_time.after?(Rails.configuration.state_file_end_of_in_progress_intakes) && @submission_to_show.present? %>
-    <%= link_to t('.download_state_return_pdf'), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: @submission_to_show.id), class: "button button--primary button--wide spacing-above-60" %>
+    <%= link_to t('.download_state_return_pdf'), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: @submission_to_show.id), class: "button button--primary button--wide spacing-above-60", role: "button" %>
   <% end %>
 
   <% if show_xml? %>
--- a/app/views/state_file/questions/submission_confirmation/edit.html.erb
+++ b/app/views/state_file/questions/submission_confirmation/edit.html.erb
@@ -38,7 +38,7 @@
 
   <%= render partial: "state_file/questions/submission_confirmation/#{current_state_code}_additional_content" rescue nil %>
 
-  <%= link_to t(".download_state_return_pdf"), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: current_intake.latest_submission), class: "button button--primary button--wide" %>
+  <%= link_to t(".download_state_return_pdf"), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: current_intake.latest_submission), class: "button button--primary button--wide", role: "button" %>
 
   <% if show_xml? %>
     <p>
--- a/app/views/state_file/landing_page/edit.html.erb
+++ b/app/views/state_file/landing_page/edit.html.erb
@@ -26,7 +26,7 @@
                 <%= t(".#{@state_code}.supported_by") %>
               </div>
             </div>
-            <%= link_to StateFile::StateFilePagesController.to_path_helper(action: :login_options), class: "button button--primary button--wide", id: "firstCta" do %>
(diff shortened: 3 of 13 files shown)


app/views/state_file/questions/return_status/edit.html.erb
```diff
@@ -7,7 +7,7 @@
   <% end %>
 
   <% if !@error&.auto_wait || app_time.after?(Rails.configuration.state_file_end_of_in_progress_intakes) && @submission_to_show.present? %>
-    <%= link_to t('.download_state_return_pdf'), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: @submission_to_show.id), class: "button button--primary button--wide spacing-above-60" %>
+    <%= link_to t('.download_state_return_pdf'), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: @submission_to_show.id), class: "button button--primary button--wide spacing-above-60", role: "button" %>
```
aloverso, line 10: [dust] I wonder if this one should stay a link, given that it opens in a new tab?

app/views/state_file/questions/submission_confirmation/edit.html.erb
```diff
@@ -38,7 +38,7 @@
 
   <%= render partial: "state_file/questions/submission_confirmation/#{current_state_code}_additional_content" rescue nil %>
 
-  <%= link_to t(".download_state_return_pdf"), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: current_intake.latest_submission), class: "button button--primary button--wide" %>
+  <%= link_to t(".download_state_return_pdf"), StateFile::Questions::SubmissionPdfsController.to_path_helper(action: :show, id: current_intake.latest_submission), class: "button button--primary button--wide", role: "button" %>
```
aloverso, line 41: [dust] same as other comment, should download PDF be a button or link?

aloverso (approved):

mluedke2 (commented):
I have to confess I still don't really understand the general rule as described in the ticket that `Generally, buttons do things and links go places`, because clicking `Continue` in any of these goes to the next screen and changes URL, meaning by that rule they should be classified as links.

So I guess what we are saying is "we are actually going to just ignore that rule and call them buttons anyway, but at least we are ignoring it consistently and not just some of the time"? Is that how I should be understanding the approach?

If so then I can mark as Approve, but it felt like a contradiction to me so I want to clarify so I understand

jachan:
@mluedke2 thanks for taking a look! Yes, it would be too much of an overhaul to switch all `link_to` instances to `<input type="submit">` elements. Most of the time, when we use `link_to` in the CFA app, it's because the page is not a CFA form element and so should not be submitting any data. 

Form Builder:
https://github.com/codeforamerica/vita-min/blob/9f0aa0df7f40940d065f2035a93ad4e548ea1d38/app/helpers/vita_min_form_builder.rb#L1

Given that we're ignoring the general rule, we still want users with screen readers to have a consistent experience with continue buttons throughout the application. That's why we're tagging all elements styled `button--wide` with `role="button"`.

mluedke2:
got it. thanks @jachan , makes sense to me!

app/views/state_file/questions/return_status/edit.html.erb
mrotondo: IMO if the intention of this role annotation is to make the audible description of the page as-identical-as-possible to the visual appearance, then anything that looks like a button (i.e. has the `button` class) should also have `role: "button"`, but I'm no expert on html semantics or accessibility to so I'm very willing to be wrong

app/views/state_file/questions/submission_confirmation/edit.html.erb
mrotondo: Same comment as above from me too

mrotondo (approved):
This looks good - if you don't mind, can you write up a sentence or two on this change with the link you shared with me in Slack, and I can bring that back to our eng channel to make sure folks know to add this role annotation to links in the future? And/or - do you know of a way to automatically enforce a rule as part of a git commit hook that will make flag PRs which add a link with class="button" but don't include this annotation?

app/views/state_file/questions/return_status/edit.html.erb
jachan: Agree that for consistency's sake, we should add `role="button"` to all `button--wide` elements.

PR #5749, opened 2025-03-18
[FYST-1933] Add logging on the possible matching intakes when IntakeLoginForm is invalid

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1933

## What was done?
- `current_intake` does not seem to exist (at least in tests/localhost) at the point where we're incrementing failed attempts after invalid ssn was entered, so the best we can do to correlate the current attempt with the token which is generated in `VerificationCodeController` with below code:
```
        hashed_verification_code = VerificationCodeService.hash_verification_code_with_contact_info(
          contact_info, @form.verification_code
        )
``` 
We can use the `token` to do:
`EmailAccessToken.lookup(raw_token).pluck(:email_address)` or `TextMessageAccessToken.lookup(raw_token).pluck(:sms_phone_number)` 
can't just do the above b/c the `lookup` limits by time (created within 10 mins)
```
  scope :lookup, ->(raw_token) do
    where(token: Devise.token_generator.digest(TextMessageAccessToken, :token, raw_token)).where("created_at > ?", Time.current - 10.minutes)
  end
```
so do:
```
raw_token = <token from url/log>
phone_number = TextMessageAccessToken.where(token: Devise.token_generator.digest(TextMessageAccessToken, :token, raw_token)).pluck(:sms_phone_number)
email_address = EmailAccessToken.where(token: Devise.token_generator.digest(EmailAccessToken, :token, raw_token)).pluck(:email_address)
```

which should give you the contact info that's being attempted to log into. Then we can confirm:
1. Whether the contact information that the client is attempting to log into is indeed the one they have started filing with us
2. Whether the email address/phone number being used does in fact match the records in the log
3. If they do match, the only explanation is something about the matching process is broken (or the wrong ssn IS being entered)

## How to test?
  - Test data used
- Risk Assessment
  - Security concerns: Is the token sensitive data to log? Is the matching intakes' ids sensitive data to log? The token is already exposed in the logs (we see the token in `/en/login/<token>` edit/put requests). I guess it's possible that an attacker may be able to find out that the token generated with a certain email/phone number matches records in our db.
  - After we figure out what the SSN issues are, I think we should remove the logs

- Before
- After
Screenshot 2025-03-18 at 3 05 35 PM


app/controllers/state_file/intake_logins_controller.rb
```diff
@@ -39,7 +39,13 @@ def update
       if @form.valid?
         sign_in_and_redirect
       else
-        @records.each(&:increment_failed_attempts)
+        if @records&.any?
```
arinchoi03, line 42: will `@records` be `nil`? probably not, but since this logic is complicated, not going to take anything for granted. `.present?` alone was not a good check, since i want to make sure you can loop over the collection

jenny-heath (approved):
looks good! one thought: for the expectations where you `expect(Rails.logger).not_to receive(:error)`, i think you can leave it at that and not include the `with`. the reason i say that is: it could log an error with a message that's just slightly different and the test would pass, even though it's not supposed to log anything at all. it's basically just an easier test to get a false positive on, if that makes sense. happy to discuss further!

spec/rails_helper.rb
```diff
@@ -135,7 +135,8 @@ def filling_in_ssn_without_dashes(args)
         intercom_access_token: "fake_access_token",
         secure_mode_secret_key: "a-fake-key-to-use-for-hashing",
         statefile_secure_mode_secret_key: "a-fake-key-to-use-for-hashing"
-      }
+      },
+      mixpanel_token: "fake_mixpanel_token"
```
arinchoi03, line 139: Rails.logger _did_ receive an error without this change, b/c the mixpanel service was not able to instantiate a `@tracker` object without a `mixpanel_token` 

```
    mixpanel_key = Rails.application.credentials.dig(:mixpanel_token)
    return if mixpanel_key.nil?

    @buffer = []
    @mutex = Mutex.new

    @tracker = # from here it instantiates a tracker depending on env
```

PR #5751, opened 2025-03-19
[FYST-1935] Keep record of unfinished intake if its being destroyed during redirect_into_login from verification code

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1935

## What was done?
- Made sure to keep a record of the intake id of the intake that's being destroyed (which is only destroyed if there's a different intake that matches the contact information provided for log in)
## How to test?
  - Unit tests
  - I ran into some odd behavior in the tests (similar to what Jey saw in a [previous PR ](https://github.com/codeforamerica/vita-min/pull/5742)on this spec) where we create an intake in AZ an ID which could both have id of 1 if we only run this test `"with an intake matching an existing intake in a different state"` scenario). Although I doubt this happens, we were under the assumption that it as the same intake if the ids matched. This obviously is not the case as the id can be the same but have a different state_code. Why I say this is unlikely is that the `existing_intake` comes from searching by contact info (email/phone) and the likelihood of a client creating two intakes in different states that end up having the same id seems unlikely...anyway, good to patch.
    - we do the same "is this the same intake" check in the `IntakeLoginsController`, but I am less familiar with that bit of code to fix it, so currently mulling over it
- Risk Assessment
  - I _am_ changing logic on when we're deleting intakes (only if they match id _and_ state_code) so I'd just want a sanity check here for general log in behavior to have not changed. The `unfinished_intake_ids` can't really be checked by QA so eng review should be sufficient here.

### DIFF
--- a/app/assets/stylesheets/_state-file.scss
+++ b/app/assets/stylesheets/_state-file.scss
@@ -991,3 +991,17 @@ $state-colors: (
     .hub-searchbar__input {}
   }
 }
+
+@media (forced-colors: active) {
+  .text-input:focus,
+  a:focus,
+  .reveal__button:focus,
+  .button:focus,
+  .button:hover,
+  .radio-button input[type="radio"]:focus,
+  .select__element:focus,
+  .checkbox input[type="checkbox"]:focus {
+    box-shadow: 0 0 0 2px #ffffff, 0 0 0 7px Highlight;
+    outline: 2px Highlight solid;
+  }
+}
--- a/app/controllers/hub/state_file/automated_messages_controller.rb
+++ b/app/controllers/hub/state_file/automated_messages_controller.rb
@@ -31,12 +31,12 @@ module Hub::StateFile
     def message_params(locale)
       intake = get_intake
       state_code = intake.state_code
-      locale = locale || I18n.locale
+      locale ||= I18n.locale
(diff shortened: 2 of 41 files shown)


arinchoi03:
talked through warden logic with @powersurge360 

```
      unfinished_logged_in_intake_ids = StateFile::StateInformationService.active_state_codes.filter_map do |state_code|
        session.dig("warden.user.state_file_#{state_code}_intake.key", 0)
      end
```

I could not for the life of me find where this data was even getting set, and Kevin let me know this gets set when the user logs into an intake. That set up is in the `routes.rb`
```
  active_state_codes.each do |code|
    devise_for "state_file_#{code}_intakes"
  end
```

jenny-heath (commented):
since we're adding the same_state check, i'm wondering if we can have a test that checks the case where the state is different but id is the same. is it possible to update the id manually in a test? maybe to a number that's so absurdly high it wouldn't have a collision with other test data? if you do try this, i'd time-box it to 15 min and if you can't get it to work don't worry about it since, as you said, it's very unlikely

jenny-heath (approved):

arinchoi03:
i used `allow` to return the same id as the other intake, hope that's sufficient!

PR #5754, opened 2025-03-19
FYST-1934 Change SocialSecurityNumberValidator to allow SSN/ITINs to start with 9

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1934

## What was done?
- Update the SSN/ITIN validator to allow numbers to start with a 9
## How to test?
- Updated one form that uses the validator to have an example ITIN that starts with 9 (taken from Atticus fixture)

### DIFF
--- a/app/validators/social_security_number_validator.rb
+++ b/app/validators/social_security_number_validator.rb
@@ -1,6 +1,6 @@
 class SocialSecurityNumberValidator < ActiveModel::EachValidator
-  REAL_SSN_REGEX = /\A(?!(000|666|9))\d{3}-?(?!00)\d{2}-?(?!0000)\d{4}\z/ # Real SSNS cannot have 00 in 4th and 5th position
-  LOOSE_SSN_REGEX = /\A(?!(000|666|9))\d{3}-?\d{2}-?(?!0000)\d{4}\z/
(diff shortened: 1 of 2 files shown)


arinchoi03 (commented):
one question i have about this, re: ITIN format requirements from https://www.irs.gov/pub/irs-pdf/p4757.pdf: 

> All valid ITINs are a nine-digit number in the same format as the SSN (9XX-8X-XXXX), begins with a “9” and the 4th and 5th digits range from 50 to 65, 70 to 88, 90 to 92, and 94 to 99.

In the case of FYST log-in, it doesn't really matter too much whether the itin is the "correct" format, b/c we just need it to match what is in the DF data; but in the other cases we take it in (GYR client forms?), maybe we _do_ need to make sure it falls in line with the expected valid values. Maybe worth asking if strict adherence to the format for ITIN is a requirement for this story.

arinchoi03:
also another random thing:
why do we match on dashes in the regex and not just on the numbers?
Im wondering if the client is using different kind of dashes (copy pasted SSN from another site/pdf) which would make their ssn "invalid"

arinchoi03 (approved):
approved!

PR #5835, opened 2025-03-25
Gyr1 519 remove deprecated get ctc statuses from the hub

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-519

## What was done?
- Creates a new method that removes certain status from view in the gyr hub
## How to test?
- Login in as an admin and look at the filters
- Before
Screenshot 2025-03-20 at 1 17 07 PM

- After
Screenshot 2025-03-20 at 1 16 33 PM

### DIFF
--- a/app/helpers/tax_return_status_helper.rb
+++ b/app/helpers/tax_return_status_helper.rb
@@ -1,6 +1,6 @@
 module TaxReturnStatusHelper
   def grouped_status_options_for_select
-    TaxReturnStateMachine.available_states_for(role_type: current_user.role_type).map do |stage, statuses|
+    TaxReturnStateMachine.states_to_show_for_client_filter(role_type: current_user.role_type).map do |stage, statuses|
       translated_stage = TaxReturnStatusHelper.stage_translation(stage)
       translated_statuses = statuses.map { |status| [TaxReturnStatusHelper.status_translation(status), status.to_s] }
       [translated_stage, translated_statuses]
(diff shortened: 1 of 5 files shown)


app/helpers/tax_return_status_helper.rb
```diff
@@ -1,6 +1,6 @@
 module TaxReturnStatusHelper
   def grouped_status_options_for_select
-    TaxReturnStateMachine.available_states_for(role_type: current_user.role_type).map do |stage, statuses|
+    TaxReturnStateMachine.states_to_show_for_client_filter(role_type: current_user.role_type).map do |stage, statuses|
```
arinchoi03, line 3: noticing that `available_states_for` is still used in some places, `#refresh_filterable_properties` in `search_indexer.rb` and `#available_stage_and_states` in `returns_by_status_presenter.rb`. 
 Are these places that need to show the ctc states?
 
Just wondering if we should update these to use `states_to_show_for_client_filter` as well (or just update `available_states_for` to be `states_to_show_for_client_filter`

arinchoi03: moving from slack convo to here:
I'd just note it in the story --
1. that we are including these statuses (["file_fraud_hold", "file_needs_review"]) in the hub dashboard still when we show returns_by_status , if it's available to the current user. Maybe that's actually the intent, to preserve historical data (if the hub user had access to ctc data previously, they will still be able to view them in the dashboard).
2. Not as sure about the index searcher, but doesn't seem super critical, as it tells the user whether the client is "greetable" based on current status. Maybe it's ok for the user to see the client as a "greetable" client if they're in ["file_fraud_hold", "file_needs_review"]

PR #5837, opened 2025-03-26
NJ 314 content updates

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/314#issue-2910837398

## What was done?
- Content updates

## How to test?
- Devito persona, pick veterans exemption

- Before

Eligibility
image
image
image

Veterans
image

Submission Confirmation
image

Review
image
image

- After
Eligibility
image
image

Veterans
image
image

Submission Confirmation
image
image
image

Review
image
image

image
image

### DIFF
--- a/app/views/state_file/questions/eligible/_vita_option.html.erb
+++ b/app/views/state_file/questions/eligible/_vita_option.html.erb
@@ -1,15 +1,9 @@
 <p><%=t('state_file.questions.eligible.vita_option.want_to_claim_vita_1', state_name: current_state_name) %></p>
 
-<% vita_link_key = "state_file.questions.eligible.vita_option.connect_to_vita.#{current_state_code}" %>
+<% # i18n-tasks-use t('state_file.questions.eligible.vita_option.vita_eligibility_reveal.nj') # hint for the i18n linter that, yes, we are using this key %>
+<% vita_link_key = "state_file.questions.eligible.vita_option.vita_eligibility_reveal.#{current_state_code}" %>
 <% vita_link_href = StateFile::StateInformationService.send("vita_link_#{I18n.locale}", current_state_code) %>
 <% faq_link_href = state_faq_path(us_state: current_state_code) %>
 <% vita_link = t(vita_link_key, faq_link_href: faq_link_href, vita_link_href: vita_link_href, default: '') %>
 <%= vita_link.html_safe if vita_link.present? %>
-
-<ul class="list--bulleted">
-  <% unless vita_link.present? %>
-    <li><%= link_to t("state_file.questions.eligible.vita_option.vita_link"), @vita_link, target: "_blank", rel: "noopener noreferrer" %></li>
-  <% end %>
-  <li><%=t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: @faq_other_options_link) %></li>
-</ul>
-
+<%=t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: @faq_other_options_link) %>
--- a/app/views/state_file/questions/eligible/_vita_option.html.erb
+++ b/app/views/state_file/questions/eligible/_vita_option.html.erb
@@ -6,4 +6,9 @@
 <% faq_link_href = state_faq_path(us_state: current_state_code) %>
 <% vita_link = t(vita_link_key, faq_link_href: faq_link_href, vita_link_href: vita_link_href, default: '') %>
 <%= vita_link.html_safe if vita_link.present? %>
-<%=t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: @faq_other_options_link) %>
+<ul class="list--bulleted">
+  <% unless vita_link.present? %>
+    <li><%= link_to t("state_file.questions.eligible.vita_option.vita_link"), @vita_link, target: "_blank", rel: "noopener noreferrer" %></li>
+  <% end %>
+    <li><%=t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: @faq_other_options_link) %></li>
+</ul>
--- a/spec/controllers/state_file/questions/eligible_controller_spec.rb
+++ b/spec/controllers/state_file/questions/eligible_controller_spec.rb
@@ -17,5 +17,29 @@ describe StateFile::Questions::EligibleController do
       expect(response).to be_successful
       expect(response_html).to have_text I18n.t("state_file.questions.eligible.edit.title1", year: MultiTenantService.statefile.current_tax_year, state: "Arizona")
     end
+
+    context "AZ" do
+      let(:intake) { create :state_file_az_intake }
+  
+      it "does not show NJ content, links to AZ FAQ" do
+        get :edit
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/az/faq/other_state_filing_options")
+        expect(response.body.html_safe).not_to include("Get started with VITA now.")
+        expect(response.body.html_safe).not_to include("Visit our FAQ")
+        expect(response.body.html_safe).not_to include("Get connected now")
+      end
+    end
+
+    context "NJ" do
+      let(:intake) { create :state_file_nj_intake }
+  
+      it "shows vita_eligibility_reveal content and not connect_to_vita content" do
+        get :edit
+        expect(response.body.html_safe).to include("Get started with VITA now.")
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/nj/faq/other_filing_options")
+        expect(response.body.html_safe).not_to include("Visit our FAQ")
+        expect(response.body.html_safe).not_to include("Get connected now")
+      end
+    end
   end
 end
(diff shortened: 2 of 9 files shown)


spec/controllers/state_file/questions/submission_confirmation_controller_spec.rb
```diff
@@ -23,8 +23,8 @@
   end
 
   describe "nj veteran content" do
-    let(:mfj_body_html) { I18n.t('state_file.questions.submission_confirmation.nj_additional_content.body_mfj_html')[0..30] }
-    let(:body_html) { I18n.t('state_file.questions.submission_confirmation.nj_additional_content.body_html')[0..30] }
+    let(:mfj_body_html) { I18n.t('state_file.questions.submission_confirmation.nj_additional_content.body_mfj_html')[3..30].delete("\n") }
+    let(:body_html) { I18n.t('state_file.questions.submission_confirmation.nj_additional_content.body_html')[3..30].delete("\n") }
```
jenny-heath, line 27: i don't totally understand this? is it because of the html in the copy? if so you can look for `CGI.escapeHTML(` to find some examples where we match on strings from the locale file that contain html

jachan: Originally, the [0..30] was meant to check the beginning of the translation string, which differs for MFJ and non-MFJ filers.

MFJ - "You said you or your spouse are filing for the veteran exemption..."
Non-MFJ - "You said you are filing for the veteran exemption."

After making the content for the `body_html` and `body_mfj_html` keys include HTML <p> tags, I changed [0..30] to [3..30] to avoid the <p> tag, which didn't exist in the previous content.

Understood that this isn't the most legible. Will look into `CGI.escapeHTML`.

jachan:
@jenny-heath Looks like the tests are failing for good reason: Changing the `state_file.questions.eligible.vita_option.connect_to_vita.nj` copy is affecting the offboarding pages, since they use it as part of `other_filing_options`.

https://github.com/codeforamerica/vita-min/blob/814f77d6bde5131a2f59a1b26158ccf3ab445acd/app/views/state_file/questions/eligibility_offboarding/edit.html.erb#L26

https://github.com/codeforamerica/vita-min/blob/814f77d6bde5131a2f59a1b26158ccf3ab445acd/app/views/state_file/questions/eligible/_other_filing_options.html.erb#L4

Confirmed with Tristan that we want the offboarding pages to stay the same, and we want to update only the eligibility page. I may end up just adding another translation key that gets used only by the eligibility page to solve this.

aloverso (approved):
Looks good, but feel free to ping me again to look at the revised code when the tests are passing

spec/controllers/state_file/questions/eligible_controller_spec.rb
```diff
@@ -17,5 +17,29 @@
...
+
+    context "AZ" do
+      let(:intake) { create :state_file_az_intake }
+  
+      it "does not show NJ content, links to AZ FAQ" do
+        get :edit
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/az/faq/other_state_filing_options")
+        expect(response.body.html_safe).not_to include("Get started with VITA now.")
+        expect(response.body.html_safe).not_to include("Visit our FAQ")
+        expect(response.body.html_safe).not_to include("Get connected now")
```
aloverso, line 29: [dust] Could these tests reference the translation keys rather than the direct strings?

app/views/state_file/questions/eligible/_vita_option.html.erb
```diff
@@ -1,15 +1,9 @@
...
+<% vita_link_key = "state_file.questions.eligible.vita_option.vita_eligibility_reveal.#{current_state_code}" %>
 <% vita_link_href = StateFile::StateInformationService.send("vita_link_#{I18n.locale}", current_state_code) %>
 <% faq_link_href = state_faq_path(us_state: current_state_code) %>
 <% vita_link = t(vita_link_key, faq_link_href: faq_link_href, vita_link_href: vita_link_href, default: '') %>
 <%= vita_link.html_safe if vita_link.present? %>
-
-<ul class="list--bulleted">
-  <% unless vita_link.present? %>
-    <li><%= link_to t("state_file.questions.eligible.vita_option.vita_link"), @vita_link, target: "_blank", rel: "noopener noreferrer" %></li>
-  <% end %>
```
aloverso, line 12: [dust] a little confused here, is this getting removed altogether? I don't see it getting added somewhere else

spec/controllers/state_file/questions/eligible_controller_spec.rb
jachan: For the negative assertions, I actually prefer testing the text rather than the translation key, since it's more general. Using the translation key results in testing the presence of specific HTML, while testing the plaintext accounts for cases where the FAQ or VITA links change or are incorrect.

Updated the positive assertion, though!

app/views/state_file/questions/eligible/_vita_option.html.erb
jachan: I think removing this code was a mistake - thanks for catching this.

spec/controllers/state_file/questions/eligible_controller_spec.rb
aloverso: @jachan Hm. I feel like specifically the negative test needs that translation key to make give me confidence in the test. Otherwise, if the text is incorrect or out of date, the test will always pass and we'd never know.

This test will likely never get updated if the content in the translation file changes. If we change the text from "Get started with VITA now." to something like "Chat now with VITA" or whatever, and then there was a bug where this text WAS showing when it wasn't supposed to, this test would not catch that because it's looking for text that would never exist on the page anyways (because it's out of date)

spec/controllers/state_file/questions/eligible_controller_spec.rb
```diff
@@ -17,5 +17,27 @@
...
       expect(response_html).to have_text I18n.t("state_file.questions.eligible.edit.title1", year: MultiTenantService.statefile.current_tax_year, state: "Arizona")
     end
+
+    context "AZ" do
+      let(:intake) { create :state_file_az_intake }
+  
+      it "does not show NJ content, links to AZ FAQ" do
+        get :edit
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/az/faq/other_state_filing_options")
+        expect(response.body.html_safe).not_to have_text ActionController::Base.helpers.strip_tags(I18n.t('state_file.questions.eligible.vita_option.vita_eligibility_reveal.nj')).delete("n")
```
jenny-heath, line 27: 🤔

```
pry(#<RSpec::ExampleGroups::StateFileQuestionsEligibleController::Edit::AZ>)> ActionController::Base.helpers.strip_tags(I18n.t('state_file.questions.eligible.vita_option.vita_eligibility_reveal.nj')).delete("n")
=> "\n  Get started with VITA ow.\n\n"
```

spec/controllers/state_file/questions/eligible_controller_spec.rb
```diff
@@ -17,5 +17,27 @@
...
+    end
+
+    context "NJ" do
+      let(:intake) { create :state_file_nj_intake }
+  
+      it "shows vita_eligibility_reveal content and not connect_to_vita content" do
+        get :edit
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.vita_eligibility_reveal.nj', vita_link_href: "https://airtable.com/appqG5OGbTLBiQ408/pag9EUHzAZzfRIwUn/form")
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/nj/faq/other_filing_options")
+        expect(response.body.html_safe).not_to have_text ActionController::Base.helpers.strip_tags(I18n.t('state_file.questions.eligible.vita_option.connect_to_vita.nj')).delete("n")
```
jenny-heath, line 39: [pebble] i don't think this check is really necessary? outside of the context of this PR it doesn't really make sense

spec/controllers/state_file/questions/eligible_controller_spec.rb
```diff
@@ -17,5 +17,27 @@
...
     end
+
+    context "AZ" do
+      let(:intake) { create :state_file_az_intake }
+  
+      it "does not show NJ content, links to AZ FAQ" do
+        get :edit
+        expect(response.body.html_safe).to include I18n.t('state_file.questions.eligible.vita_option.want_to_claim_learn_more_html', link: "/en/az/faq/other_state_filing_options")
+        expect(response.body.html_safe).not_to have_text ActionController::Base.helpers.strip_tags(I18n.t('state_file.questions.eligible.vita_option.vita_eligibility_reveal.nj')).delete("n")
+        expect(response.body.html_safe).not_to have_text ActionController::Base.helpers.strip_tags(I18n.t('state_file.questions.eligible.vita_option.connect_to_vita.nj')).delete("n")
```
jenny-heath, line 28: [pebble] same as comment below

jenny-heath (approved):
one more request to refactor tests but shouldn't require re-review

spec/controllers/state_file/questions/eligible_controller_spec.rb
jachan: It was supposed to be `/n`, not `n`! That is very funny. Will just take your suggestion here.

PR #5844, opened 2025-03-27
[FYST-1988] Fix flakey verification_code_controller spec

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1988

## What was done?
- Fixing a test which is actually failing locally when the entire file is run (passes if only the single spec is run)
## How to test?
  - run multiple times in circle ci -- see that it does not fail on the first run (`bundle exec parallel_rspec -n 12 || touch /tmp/re-run`)
- Alternative considered:
  - Tried setting the id to a high number (see the change): https://github.com/codeforamerica/vita-min/pull/5844/commits/081c241adaed2638385ba81eea204eb2ce6780da
  - initially just set the id to something high like Jenny suggested: https://github.com/codeforamerica/vita-min/pull/5751#pullrequestreview-2704377810 but that felt brittle, so added some logic to grab the `last + 1` id from both id/az intakes test db tables. At the end of the day, it felt really odd to update an intake's id like this but difficult to recreate this scenario without it, so decided to go with this approach instead.

### DIFF
--- a/spec/controllers/state_file/questions/verification_code_controller_spec.rb
+++ b/spec/controllers/state_file/questions/verification_code_controller_spec.rb
@@ -100,7 +100,7 @@ RSpec.describe StateFile::Questions::VerificationCodeController do
         )
         expect(response).to redirect_to(login_location)
         expect(StateFileAzIntake.where(id: intake.id)).to be_empty
-        expect(StateFileIdIntake.find(existing_intake.id).unfinished_intake_ids).to include(intake.id.to_s)
+        expect(existing_intake.reload.unfinished_intake_ids).to include(intake.id.to_s)
       end
 
       context "with the same intake ids" do
@@ -118,7 +118,7 @@ RSpec.describe StateFile::Questions::VerificationCodeController do
           )
           expect(response).to redirect_to(login_location)
           expect(StateFileAzIntake.where(id: intake.id)).to be_empty
-          expect(StateFileIdIntake.find(existing_intake.id).unfinished_intake_ids).to include(intake.id.to_s)
+          expect(existing_intake.reload.unfinished_intake_ids).to include(intake.id.to_s)
         end
       end
     end
--- a/spec/controllers/state_file/questions/verification_code_controller_spec.rb
+++ b/spec/controllers/state_file/questions/verification_code_controller_spec.rb
@@ -100,7 +100,7 @@ RSpec.describe StateFile::Questions::VerificationCodeController do
         )
         expect(response).to redirect_to(login_location)
         expect(StateFileAzIntake.where(id: intake.id)).to be_empty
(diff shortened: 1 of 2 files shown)


spec/controllers/state_file/questions/verification_code_controller_spec.rb
```diff
@@ -85,6 +85,7 @@
       let(:intake) do
         build(:state_file_az_intake, contact_preference: "email", email_address: "someone@example.com", visitor_id: "v1s1t1n9").tap do |intake|
           intake.raw_direct_file_data = nil
+          intake.id = 52322
```
arinchoi03, line 88: just really want to set the id intake's id to be the same as the az intake, but setting to a high value so it doesn't accidentally clash with another id intake in test db

spec/controllers/state_file/questions/verification_code_controller_spec.rb
```diff
@@ -105,7 +106,7 @@
 
       context "with the same intake ids" do
         before do
-          allow(existing_intake).to receive(:id).and_return(intake.id)
```
arinchoi03, line 108: i think this introduced some weird behavior when i i called:

`expect(existing_intake.reload.unfinished_intake_ids).to include(intake.id.to_s)` b/c the reload uses the id of the `existing_intake` to reload from the db, but idk if the above `allow` stubbing was working for the reload. So we would sometimes get the same ids for the id/az intake and sometimes not 🙃

spec/controllers/state_file/questions/verification_code_controller_spec.rb
```diff
@@ -85,6 +85,8 @@
       let(:intake) do
         build(:state_file_az_intake, contact_preference: "email", email_address: "someone@example.com", visitor_id: "v1s1t1n9").tap do |intake|
           intake.raw_direct_file_data = nil
+          unused_id = ([StateFileAzIntake.last&.id, StateFileIdIntake.last&.id].compact.max || 0) + 1
+          intake.id = unused_id
```
arinchoi03, line 89: just want to grab an id that won't clash with any existing ones on either AZ/ID intakes table

arinchoi03:
recent runs all are green & do not include the verification code controller spec as the failed one
Screenshot 2025-03-27 at 11 38 11 PM

spec/controllers/state_file/questions/verification_code_controller_spec.rb
jachan: Ah that makes sense! What a tricky testing scenario...

jachan (approved):
Thanks for fixing!

anisharamnani (commented):
this has already been merged, but i also reviewed this!

PR #5857, opened 2025-04-02
Nj 321 extension payments

jachan:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/321#issue-2921165506

## What was done?
- During extension season, we want to ask taxpayers how much they've made in extension payments

## How to test?
  - Test in heroku with `extension_period` feature flag OFF, question should not appear on Estimated Payments page.
  - Test in heroku with `extension_period` feature flag ON, question should appear on Estimated Payments page and Lines 57 and 66 should include the extension_payments amount.

- Before (feature flag off)
image

- After (feature flag on)
image

### DIFF
--- a/app/forms/state_file/nj_estimated_tax_payments_form.rb
+++ b/app/forms/state_file/nj_estimated_tax_payments_form.rb
@@ -3,13 +3,16 @@ module StateFile
     set_attributes_for :intake,
                        :has_estimated_payments,
                        :estimated_tax_payments,
+                       :extension_payments,
                        :overpayments
 
     validates :has_estimated_payments, presence: true
     validates_numericality_of :estimated_tax_payments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
     validates_numericality_of :overpayments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
+    validates_numericality_of :extension_payments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
     validates :estimated_tax_payments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
     validates :overpayments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
+    validates :extension_payments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
 
     def save
       @intake.update(attributes_for(:intake))
--- a/app/forms/state_file/nj_estimated_tax_payments_form.rb
+++ b/app/forms/state_file/nj_estimated_tax_payments_form.rb
@@ -9,7 +9,7 @@ module StateFile
     validates :has_estimated_payments, presence: true
     validates_numericality_of :estimated_tax_payments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
     validates_numericality_of :overpayments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
-    validates_numericality_of :extension_payments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
+    validates_numericality_of :extension_payments, message: I18n.t("validators.not_a_number"), if: -> { extension_payments.present? }
     validates :estimated_tax_payments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
     validates :overpayments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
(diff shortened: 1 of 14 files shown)


aloverso (changes requested):
[boulder] Seems like the new field is missing all its validation. It will allow me to enter a negative number or a string without error. I think we're missing the code and tests that validate its numericality and the like

app/forms/state_file/nj_estimated_tax_payments_form.rb
```diff
@@ -3,13 +3,16 @@ class NjEstimatedTaxPaymentsForm < QuestionsForm
...
+                       :extension_payments,
                        :overpayments
 
     validates :has_estimated_payments, presence: true
     validates_numericality_of :estimated_tax_payments, message: I18n.t("validators.not_a_number"), if: -> { estimated_tax_payments.present? }
     validates_numericality_of :overpayments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
+    validates_numericality_of :extension_payments, message: I18n.t("validators.not_a_number"), if: -> { overpayments.present? }
     validates :estimated_tax_payments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
     validates :overpayments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
+    validates :extension_payments, allow_blank: true, numericality: { greater_than_or_equal_to: 0 }, if: -> { has_estimated_payments == "yes" }
```
tahsinaislam, line 16: do you need to have these validations also check for the flipper flag?

spec/forms/state_file/nj_estimated_tax_payments_form_spec.rb
```diff
@@ -79,6 +82,7 @@
         expect(intake.has_estimated_payments).to eq "no"
         expect(intake.estimated_tax_payments).to eq nil
         expect(intake.overpayments).to eq nil
+        expect(intake.extension_payments).to eq nil
       end
     end
   end
```
tahsinaislam, line 88: [dust] maybe would consider having tests for the conditions to see if there are validation errors only if `overpayments` are present and if has_estimated_payments == "yes"?

tahsinaislam (approved):
just had a question on the flipper and adding some tests for the form!

app/forms/state_file/nj_estimated_tax_payments_form.rb
jachan: I don't believe so! I think of these validations as checks on the content of the field, rather than checking that it was displayed correctly. I added additional tests in the `nj_estimated_tax_payments_controller_spec` to validate that the Flipper flag correctly shows/hides the extension payments field!

spec/forms/state_file/nj_estimated_tax_payments_form_spec.rb
jachan: I believe I addressed this with the

`it_behaves_like :nj_money_field_concern, field: :extension_payments, can_be_empty: true do`

tests in the latest commit! Let me know if you still have concerns!

aloverso (approved):
looks good! appreciate the flipper tests as well

app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
```diff
@@ -44,6 +44,16 @@
                 options: { placeholder: "0" },
                 help_text: t('.overpayments_input_helper_html', prior_year: current_tax_year-1)
               ) %>
+
+          <% if Flipper.enabled?(:extension_period) %>
+            <%= f.vita_min_money_field(
+                  :extension_payments,
+                  t('.extension_payments_input_label_html'),
+                  classes: ["form-width--long"],
+                  options: { placeholder: "0" },
```
mmazanec22, line 53: [pebble] Do we want to use the placeholder given how bad the contrast is between placeholder and regular text, and how confusing it is to realize this field doesn't default to 0? Is it adding anything to the user experience?

mmazanec22 (approved):

app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
tristanvanech: what do you mean it doesn't default to 0?

jachan: We treat nil like 0 in the application, so if a user continues past this screen without inputting anything, the value used for extension payments is 0.

What Melanie is referring to is a design question about whether the placeholder text of "0" here is helping the user or not. I defer that decision to Hannah.

tristanvanech: > We treat nil like 0 in the application, so if a user continues past this screen without inputting anything, the value used for extension payments is 0.

so doesn't that mean the field *does* default to 0?

I get the design decision, and agree Hannah should make the call here, just want to make sure I'm not missing something

PR #5860, opened 2025-04-02
Add indexes; move intake query out of join

powersurge360:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-712

## What was done?
The explain query seemed to do most of the work in the intake join. This code separates that into a new query and then uses it as a cache to recombine in ruby. There was also some low hanging fruit in some indexes that could be added.

## How to test?
- It's a little difficult; the only time this code is really an issue is in production. Probably the best thing to do would be to test that sorting retains its functionality in the heroku instance. If so, it should be fine to promote to production as it is unlikely to be worse than existing behavior.

- Before
- After


spec/controllers/hub/clients_controller_spec.rb
```diff
@@ -1526,7 +1526,12 @@
     let(:product_year) { Rails.configuration.product_year }
     let(:intake) { build(:intake, product_year: product_year) }
     let(:client) { create(:client, intake: intake, tax_returns: tax_returns) }
-    let(:presenter) { Hub::ClientsController::HubClientPresenter.new(client) }
+    let(:presenter) do
+      Hub::ClientsController::HubClientPresenter.new(
+        client,
+        {intakes: [client.intake]}
```
squanto, line 1532: < nit >looks like this passes, but doesn't the implementation have a different shape? something like
 `{ intakes: { client_id: intake } }`

powersurge360: You're right. When I made the argument optional I think I don't need this at all anyhow.

powersurge360: Since the tests take eons, here's this one spec turning green w/ that most recent change

image

squanto (approved):

spompea-cfa:
love it 👍

PR #5867, opened 2025-04-03
turn on show_retirement_ui flipper flag in review page spec to be more like prod

mrotondo:

## What was done?
- Previously, the review page spec didn't enable the `show_retirement_ui` flipper flag, which has now been turned on in prod for a while and is likely to stay on forever.
- Because of this, our automated tests failed to catch a change in another PR which would have caused the review page to crash for all NJ filers.
- Weirdly, we _ already do_ test 1099-R editing behavior in the review page spec - it seems like we may have been inconsistent in what was hidden behind the feature flag?
## How to test?
- As long as the review spec continues to pass, this change should be harmless

### DIFF
--- a/spec/features/state_file/review_page_spec.rb
+++ b/spec/features/state_file/review_page_spec.rb
@@ -6,6 +6,10 @@ RSpec.feature "Completing a state file intake", active_job: true, js: true do
 
   before do
     allow_any_instance_of(Routes::StateFileDomain).to receive(:matches?).and_return(true)
+
+    # Turn on 1099-R support to be more like prod
+    allow(Flipper).to receive(:enabled?).and_call_original
+    allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
   end
 
   StateFile::StateInformationService.active_state_codes.without("ny").each do |state_code|


spec/features/state_file/review_page_spec.rb
```diff
@@ -6,6 +6,10 @@
 
   before do
     allow_any_instance_of(Routes::StateFileDomain).to receive(:matches?).and_return(true)
+
+    # Turn on 1099-R support to be more like prod
+    allow(Flipper).to receive(:enabled?).and_call_original
+    allow(Flipper).to receive(:enabled?).with(:show_retirement_ui).and_return(true)
```
arinchoi03, line 12: the rest of the specs appear to do this -- this spec is only the income review & editing directly from that income review to the 1099R edit (which was outside the `show_retirement_ui` scope -- i think `show_retirement_ui` is more about the followups?

mrotondo: You're right, false alarm on the flipper flag not hiding the right stuff. But I do still think we should either turn the flag on for this test, or create an NJ feature test that includes 1099-R followups and a review page visit, because apparently there is no test that hits the NJ review page with the show_retirement_ui flag on (or it would have failed when I accidentally broke that page)

PR #5868, opened 2025-04-04
GYR1-700 add mixpanel click events for homepage filing options

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-700

## What was done?
- Imported from `@rails/ujs` and set `window.Rails`, per this post: https://www.reddit.com/r/rails/comments/hz9v3y/why_is_the_rails_object_not_defined_in_javascript/
- It seemed maybe nice to have the tests check that this works and it looked to me like the reason this wasn't caught in tests was that the test mocked the Rails object. The thing is, if I remove the mock it fails because it doesn't seem to take `application.js` into account and I don't know how to include that. One way to get it to pass is to import Rails in `mixpanel_event_tracking.js` but that seems redundant
## How to test?
- see above

### DIFF
--- a/app/javascript/lib/mixpanel_event_tracking.js
+++ b/app/javascript/lib/mixpanel_event_tracking.js
@@ -1,5 +1,6 @@
 import "../vendor/navigator.sendbeacon.min.js";
-
+import Rails from "@rails/ujs";
+window.Rails = Rails;
 
 const MixpanelEventTracking = (function () {
     const addClickTrackingToOutboundLinks = function () {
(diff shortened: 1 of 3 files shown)


jnf:
i'd be a fan of lifting the CSRF param/token combo directly out of the page like `document.querySelector('meta[name=csrf-token]')?.content` as described in the [Rails Guide](https://guides.rubyonrails.org/security.html#required-security-token).

app/javascript/lib/mixpanel_event_tracking.js
```diff
@@ -1,5 +1,6 @@
 import "../vendor/navigator.sendbeacon.min.js";
-
+import Rails from "@rails/ujs";
+window.Rails = Rails;
```
jnf, line 3: we can eliminate this dependency from this file (and potentially solve the issue w/ test scope you mention in the pr description) by lifting the csrf information from the DOM directly.

jnf (approved):
yay!

PR #5875, opened 2025-04-07
Nj 334 remove worksheet g renters

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/334#issue-2977477424

## What was done?
- Research found that Worksheet G was confusing for renters, and we worked with taxation to create different language that was more legible for users, instead of directing them to the worksheet.
- This change removes Worksheet G from the renters' flow, removes questions from the eligibility screen that previously directed users to the Worksheet G page, and updates content for the rent paid and renter eligibility screens

## How to test?
- Using Danny Devito Single persona
- Claim to be a renter for property tax credit purposes
- Verify that Eligibility Screen has updated content and fewer checkboxes
- Verify that Rent Paid screen has new content

- Before
image
image

- After
image
image


app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
```diff
@@ -31,8 +31,6 @@
         document.getElementById("state_file_nj_tenant_eligibility_form_tenant_home_subject_to_property_taxes"),
         document.getElementById("state_file_nj_tenant_eligibility_form_tenant_building_multi_unit"),
         document.getElementById("state_file_nj_tenant_eligibility_form_tenant_access_kitchen_bath"),
-        document.getElementById("state_file_nj_tenant_eligibility_form_tenant_more_than_one_main_home_in_nj"),
-        document.getElementById("state_file_nj_tenant_eligibility_form_tenant_shared_rent_not_spouse")
```
aloverso, line 35: [dust] I know Tristan wants to test & experiment with this feature with the possibility of rolling back if it makes things more complicated for people. So because of that, I don't think we should include a database migration in *this* PR that removes these fields from the database altogether in case of that situation. That said, I do think we should ensure a ticket gets put in the backlog to do so in the future, because if these fields are never being used, they *should* be removed from the database and keep a clean data model

app/views/state_file/questions/nj_tenant_eligibility/edit.html.erb
```diff
@@ -6,7 +6,7 @@
 <% content_for :card do %>
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder, html: { class: "form-card form-card--long" } do |f| %>
     <h1 class="h2"><%= title %></h1>
-
+  <p><%= t('.instructions_html') %></p>
```
aloverso, line 9: [pebble] This will nest a `<p>` within a `<p>` because the content also contains a p. It should only be one or the other

app/views/state_file/questions/nj_tenant_rent_paid/edit.html.erb
```diff
@@ -7,6 +7,7 @@
   <h1 class="h2"><%= t(".title", filing_year: current_tax_year) %></h1>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
+    <p><%= t('.instructions_html') %></p>
```
aloverso, line 10: [pebble] This will nest a `<p>` within a `<p>` because the content also contains a p. It should only be one or the other

config/locales/en.yml
```diff
@@ -3842,32 +3842,25 @@ en:
               <li>A residence on which Payments-In-Lieu-of-Tax (P.I.L.O.T) payments were made. (If you think your building may have been subject to P.I.L.O.T payments, you can contact your building owner or the municipal tax office to confirm.)</li>
             </ul>
           helper_heading: How do I know if the home I rented was subject to property taxes?
+          instructions_html: "<p>Many people are surprised by this, but New Jersey has a law on the books that gives you a  property tax deduction or credit as a renter.</p> \n<p><strong>We will walk you through it.</strong></p>\n"
```
aloverso, line 3845: [pebble] There's no need for the `\n`s if we're using paragraph tags to break lines

jenny-heath (approved):

spec/helpers/state_file/nj_tenant_eligibility_helper_spec.rb
```diff
@@ -22,22 +22,6 @@
...
-      end
-
-      context "when YES to tenant_shared_rent_not_spouse" do
-        let(:intake) {  create :state_file_nj_intake, tenant_shared_rent_not_spouse: "yes" }
-        it "returns worksheet required" do
-          expect(described_class.determine_eligibility(intake)).to eq(described_class::WORKSHEET)
-        end
-      end
-    end
-
```
aloverso, line 40: [pebble/boulder if you want re-review] GitHub isn't letting me select the right lines of code to put this comment on, but this whole file should be reviewed & revised. Several of the other tests in here (like the "advance state" tests) reference the `tenant_shared_rent_not_spouse` and `tenant_more_than_one_main_home_in_nj` checkboxes, and since those are no longer part of the code for this helper component, they should be cleaned up in the tests as well

spec/helpers/state_file/nj_tenant_eligibility_helper_spec.rb
```diff
@@ -111,8 +95,8 @@
                  tenant_shared_rent_not_spouse: "yes", # worksheet required
                  tenant_more_than_one_main_home_in_nj: "no" # advance
         }
-        it "determines the taxpayer should fill out the worksheet" do
-          expect(described_class.determine_eligibility(intake)).to eq(described_class::WORKSHEET)
+        it "determines the taxpayer should advance" do
+          expect(described_class.determine_eligibility(intake)).to eq(described_class::ADVANCE)
```
aloverso, line 99: [pebble] Similar to my comment above, this test should be reviewed for what value it is providing us rather than just change the words. The test context "taxpayer selects checkboxes that correspond to both worksheet required and advance states" is no longer possible, so I'd argue this should probably be deleted

mmazanec22 (commented):
Wow, what a dramatic usability improvement

hannahchatham:
reviewed, looks great! Thank you @jachan

PR #5881, opened 2025-04-08
[FYST-2000] Update md direct debit behavior post april 15

squanto:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-2000

## What was done?
- hardcode a April 15th at midnight deadline for setting a payment date when md state taxes are owed
- Set default to 
## How to test?
- Use session toggle to update the app time to after the april 15th deadline for md
- Start a md intake, selecting a personal that owes taxes (like Drum or Frank rollover)
- Verify that a date is not editable on the taxes owed screen

- After
Screenshot 2025-04-08 at 3 08 02 PM

### DIFF
--- a/app/controllers/state_file/questions/md_taxes_owed_controller.rb
+++ b/app/controllers/state_file/questions/md_taxes_owed_controller.rb
@@ -4,6 +4,33 @@ module StateFile
       def self.form_key
         "state_file/taxes_owed_form"
       end
+
+      def edit
+        super
+        @after_filing_deadline = app_time.in_time_zone(timezone).to_date.after?(md_submission_deadline)
+      end
+
+      private
+
+      def md_payment_deadline
+        @md_payment_deadline ||= StateFile::StateInformationService.payment_deadline_date("md", filing_year: current_year)
+      end
+
+      def md_submission_deadline
+        @md_submission_deadline ||= Date.new(current_year, 4, 15).in_time_zone(timezone).at_midnight
+      end
+
+      def current_year
+        current_tax_year + 1
+      end
+
+      def current_tax_year
+        MultiTenantService.new(:statefile).current_tax_year
+      end
+
+      def timezone
+        StateFile::StateInformationService.timezone(current_intake.state_code)
+      end
     end
   end
 end
--- a/app/controllers/state_file/questions/md_taxes_owed_controller.rb
+++ b/app/controllers/state_file/questions/md_taxes_owed_controller.rb
@@ -1,0 +1,9 @@
+module StateFile
+  module Questions
+    class MdTaxesOwedController < TaxesOwedController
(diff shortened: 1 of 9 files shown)


app/controllers/state_file/questions/md_taxes_owed_controller.rb
```diff
@@ -0,0 +1,36 @@
+module StateFile
+  module Questions
+    class MdTaxesOwedController < TaxesOwedController
+      def self.form_key
+        "state_file/taxes_owed_form"
+      end
+
+      def edit
+        super
+        @after_filing_deadline = app_time.in_time_zone(timezone).to_date.after?(md_submission_deadline)
```
mpidcock, line 10: We started discussing this, but I'll try to capture it all here.

* Calling `Date.new()` will get you a date without a time stamp
   * e.g. `Tue, 15 Apr 2025`
* `DateTime.new()` with no time params will default to midnight GMT
   * e.g. `Tue, 15 Apr 2025 00:00:00 +0000`
* Calling `in_time_zone` will look at the difference in offset of the object's timezone and the timezone param, and apply that offset to the time. 
   * if there is no time on the object, it adds a time of midnight in the new timezone
      * (using `Date`) => `Tue, 15 Apr 2025 00:00:00.000000000 EDT -04:00`
   * if there is a time but no timezone, it defaults to GMT
      * (using `DateTime`) => `Mon, 14 Apr 2025 20:00:00.000000000 EDT -04:00`
* Something I learned for myself is that calling 

So, it was just a note of caution about applying `in_time_zone` if there is a time but no timezone, because it will change the date to be a day earlier (for US timezones). You aren't doing this, so that's ok.

Additionally, by calling `to_date`, the `after?` comparison is only comparing the date. This means it only returns true if the date is later (not equal), and that the time doesn't matter (since midnight is when the date changes anyway)

One more thing to be aware of is that other stories are also needing april 15th as a standalone value (not using the `payment_deadline`. 

Here's what I would like to suggest:
1. Have @md_submission_deadline return a date with no time or timezone (it's just "April 15th")
2. Set the timezone for both dates on this line for the `after?` comparison
   * I would actually switch to use `before?` since we want the logic to change on the 15th, when they are equal
   * since the comparison is being done in the same timezone, it doesn't matter what the actual time values are
   
I can help on this later today if needed!

app/views/state_file/questions/md_taxes_owed/_md_bank_details.html.erb
mpidcock, line 1: I know this isn't clear, but the same partial in `md_tax_refund` also has the logic for `tax_refund`, so this isn't needed. :/

squanto:
closing in favor of https://github.com/codeforamerica/vita-min/pull/5889

PR #5886, opened 2025-04-09
NJ 340 - reduce W2 box 16 amount from DF import when greater than DB max

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/340

## What was done?
- If state wages amount received from DF is >$9,999,999,999.99, overwrite it to $9,999,999,999.99
- Note that the $9,999,999,999.99 amount will be rounded to an even 10,000,000,000 in the submitted NJ return XML because of our rounding rules. No matter what the DF original value was (be it 10,000,000,000 or anything else) it will get submitted as the rounded 10mil if the user does not edit it on the income editing screen. Because this is a rounding thing on the submission side, it doesn't cause any database error in the app)

## How to test?
Use newly created NJ persona `Irs test box 16 more than 10 digits`
(note that this fix will apply to all states, not just NJ though)

**W2 XML in DirectFile fake data:**
image

**After importing into FYST:**
image

**In NJ Return XML:**
image

### DIFF
--- a/app/models/state_file_base_intake.rb
+++ b/app/models/state_file_base_intake.rb
@@ -172,6 +172,7 @@ class StateFileBaseIntake < ApplicationRecord
   def synchronize_df_w2s_to_database
     direct_file_data.w2s.each_with_index do |direct_file_w2, i|
       state_file_w2 = state_file_w2s.where(w2_index: i).first || state_file_w2s.build
+      db_numeric_max = 9_999_999_999.99
       box_14_values = {}
       direct_file_w2.w2_box14.each do |deduction|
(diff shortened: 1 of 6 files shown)


jachan (approved):
This change looks necessary, approving. Do we need to worry about other fields such as federal wages?

aloverso:
@jachan we have a [lower-rush ticket](https://github.com/newjersey/affordability-pm/issues/341) to investigate how it might be necessary to do so for other fields

mpidcock (approved):
💥

mmazanec22 (approved):

PR #5861, opened 2025-04-02
[FYST-2013] Update logic for RequestedPaymentDate to use the intake's submission date if after April 15th

squanto:
## Link to pivotal/JIRA issue
https://codeforamerica.atlassian.net/browse/FYST-2013

## Is PM acceptance required? (delete one)
- No - merge after code review approval

## What was done?
update logic for setting `RequestedPaymentDate` in xml. after deadline, use intake submission time

PR #5888, opened 2025-04-09
Fyst 2009 nc ensure direct debit payment requested date is correct on after 4 15

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2009
- ## Is PM acceptance required? (delete one)

## What was done?
- Hides the date selector for the NC taxes owed page and uses the next available day instead.
## How to test?
- Use session toggles to check before and after 4/15
- Look at the XML for RequestPaymentDate
- Before
Screenshot 2025-04-09 at 4 02 02 PM

- After
Screenshot 2025-04-09 at 4 01 32 PM


app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -35,7 +48,7 @@ def withdrawal_date_is_not_on_a_weekend
     end
 
     def withdrawal_date_is_not_a_federal_holiday
-      if Holidays.on(date_electronic_withdrawal, :us, :federalreservebanks, :observed).any?
+      if  holiday?(date_electronic_withdrawal)
```
arinchoi03, line 51: [super dust] extra space

arinchoi03 (commented):
a few things need revisiting, here if you need to tuple/chat... 

Big picture thought here -- looking at this PR https://github.com/codeforamerica/vita-min/pull/5861 -- after April 15th, only the `@submission.created_at` is used to fill in `RequestedPaymentDate` in the xml. So even if you save the correct `date_electronic_withdrawal` date, it will not be used to fill the xml. You rather need to move the next business day logic to where we fill out the xml for NC and leave `date_electronic_withdrawal` blank in the form/intake on save. 

Currently on your branch, I'm seeing that even if i use session toggles to after April 15, 2025 (`{"app_time"=>"2032-08-23T21:29:00.000-07:00"} UTC`) that `StateFile::StateInformationService.before_payment_deadline?(current_time, self.state_code)` returns `true`, and uses the logic you've updated in this PR to fill in the date, so it **appears** to work correctly b/c _before_ deadline, it uses the `date_electronic_withdrawal` date. This is b/c `@submission.created_at` is used to fill the field & the created_at date does not take `app_time` into consideration, so you're currently "before_payment_deadline" even if you've used the session toggle to set to after April 15th.
```
# state_file/esign_declaration_form.rb
        new_efile_submission = EfileSubmission.create!(data_source: @intake)
```

I'm kind of unsure how you're supposed to let QA test this -- maybe if there was a way to set the heroku instance time to on or after April 15??

spec/forms/state_file/nc_taxes_owed_form_spec.rb
```diff
@@ -152,5 +152,98 @@
         end
       end
     end
+
+    context "when the form is submitted after the payment deadline, month is nil and day is nil" do
+      let(:withdrawal_month) { nil }
+      let(:withdrawal_day) { nil }
```
mpidcock, line 158: The withdrawal_month and withdrawal_day are not relevant after the payment deadline since the selector is gone, and even if they are populated, they are ignored. I think adding them here is distracting since they have no impact on the logic of this test (it's the current date that triggers the condition).

spec/models/state_file_base_intake_spec.rb
mpidcock, line 1: Also move this test to `state_file_nc_intake_spec.rb`

spec/forms/state_file/nc_taxes_owed_form_spec.rb
```diff
@@ -152,5 +152,98 @@
...
+
+      context "when it is before 5pm and the next day is a valid day" do
+        # 4pm Tuesday
+        let(:app_time) { DateTime.new(filing_year, 4, 16, 15, 0, 0)}
```
mpidcock, line 162: (formatting - missing space)

mpidcock, line 162: This would be better to set under the header for `context "when the form is submitted after the payment deadline"` because this line is what makes that context true. Other tests can additionally set other dates, but having a default of the 15th would work. Also, this logic should start applying on the 15th.

spec/forms/state_file/nc_taxes_owed_form_spec.rb
```diff
@@ -152,5 +152,98 @@
...
+        it "is valid and saves the intake with the next day" do
+          form = described_class.new(intake, params)
+          # needed because we want to keep the year the same in the tests for holiday/day of the week.
+          allow(form).to receive(:form_submitted_before_payment_deadline?).and_return(false)
```
mpidcock, line 166: I don't like this being here because it hijacks an important piece of the logic - detecting if it is currently before or after the payment deadline. I can work with you to get this to trigger accurately instead of mocking critical logic.

mpidcock: I ended up playing around with this and have a fix to recommend

spec/controllers/state_file/questions/nc_taxes_owed_controller_spec.rb
```diff
@@ -9,12 +9,31 @@
...
+        expect(response).to be_successful
+        expect(response_html).to have_text "You owe"
+        expect(response_html).to have_text "Routing Number"
+        expect(response_html).to have_text "When would you like the funds withdrawn from your account?"
+      end
+    end
+
+    context 'after intake closes' do
+      render_views
+      let(:app_time) { DateTime.new(Time.now.year, 4, 20, 12, 0, 0) }
```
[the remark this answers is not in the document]
DrewProebstel: Ended up not needing this test

spec/forms/state_file/nc_taxes_owed_form_spec.rb
```diff
@@ -152,5 +152,98 @@
...
+    context "when the form is submitted after the payment deadline, month is nil and day is nil" do
+      let(:withdrawal_month) { nil }
+      let(:withdrawal_day) { nil }
+
+      context "when it is before 5pm and the next day is a valid day" do
+        # 4pm Tuesday
+        let(:app_time) { DateTime.new(filing_year, 4, 16, 15, 0, 0)}
+        it "is valid and saves the intake with the next day" do
+          form = described_class.new(intake, params)
+          # needed because we want to keep the year the same in the tests for holiday/day of the week.
```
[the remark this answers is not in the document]
DrewProebstel: Ended up not needing this test

app/helpers/date_helper.rb
```diff
@@ -62,6 +62,40 @@ def valid_text_date(date_year, date_month, date_day, key = :date)
...
+    # if you submit your bank draft payment after 5:00 pm EST, the earliest draft date available will be two business days in the future
+    after_5pm = after_5pm_in_nc(@form_submitted_time)
+    two_business_days_away = add_business_days_to_date(@form_submitted_time.to_date, 2)
+    if after_5pm && !date_electronic_withdrawal.to_date.after?(two_business_days_away)
+      errors.add(:date_electronic_withdrawal, I18n.t("errors.attributes.nc_withdrawal_date.post_five_pm"))
+    end
+  end
+
+  def next_available_date(current_time)
+    initial_days_to_add = after_5pm_in_nc(current_time).hour >= 17 ? 2 : 1
```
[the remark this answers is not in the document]
DrewProebstel: was a wip

app/views/state_file/questions/nc_tax_refund/_nc_bank_details.html.erb
```diff
@@ -9,7 +9,7 @@
     <% if owe_taxes %>
       <%= form.hidden_field :app_time, value: app_time %>
       <%= form.vita_min_money_field(:withdraw_amount, t('.withdraw_amount', owed_amount: taxes_owed), classes: ["form-width--long"]) %>
-      <% if current_time_before_payment_deadline? %>
+      <% if current_time_before_payment_deadline?%>
```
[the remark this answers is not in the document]
DrewProebstel: artifact from some unneeded changes before

spec/forms/state_file/nc_taxes_owed_form_spec.rb
DrewProebstel: Ended up not needing this test

spec/forms/state_file/nc_taxes_owed_form_spec.rb
DrewProebstel: Ended up not needing this test

spec/models/state_file_nc_intake_spec.rb
```diff
@@ -181,4 +181,29 @@
...
+    let(:payment_deadline_date) { StateFile::StateInformationService.payment_deadline_date("nc") }
+    let(:filing_year) { MultiTenantService.new(:statefile).current_tax_year }
+
+    context "when submitted after the payment deadline" do
+      let(:current_time) { payment_deadline_date + 1.day }
+
+      it "returns next available date" do
+        expect(intake).to receive(:next_available_date)
+        intake.calculate_date_electronic_withdrawal(current_time: current_time)
+      end
```
mpidcock, line 197: My only concern is that this test doesn't ensure that the timezone is handled correctly, which could result in the `next_available_date` being calculated from the day in UTC instead of EST day (they are different from 8pm-midnight, which is a common filing time).

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -13,15 +13,27 @@ class NcTaxesOwedForm < TaxesOwedForm
...
-    with_options unless: -> { payment_or_deposit_type == "mail" || !form_submitted_before_payment_deadline? } do
-      validate :withdrawal_date_is_after_today
-      validate :withdrawal_date_is_not_on_a_weekend
-      validate :withdrawal_date_is_not_a_federal_holiday
-      validate :withdrawal_date_is_at_least_two_business_days_in_the_future_if_after_5pm
+    validate :validate_withdrawal_date_fields
+
+    def save
+      attrs = attributes_for(:intake)
+      date = form_submitted_before_payment_deadline? ? date_electronic_withdrawal : next_available_date(Time.parse(app_time))
```
mpidcock, line 20: FYI `app_time` is in UTC, but `@form_submitted_time` is in the local time. Also, this does get overwritten by `calculate_date_electronic_withdrawal` once the intake is submitted, this value is here as a placeholder so that it isn't nil between submitting the form and submitting the whole intake, but afaik it isn't actually used until the intake is submitted. 

also, this is the one change to the form that warrants adding a test, but the stakes are very low since this value is likely never looked at

app/helpers/date_helper.rb
```diff
@@ -62,6 +62,40 @@ def valid_text_date(date_year, date_month, date_day, key = :date)
...
+    date.to_date
+  end
+
+  def holiday?(date)
+    Holidays.on(date, :us, :federalreservebanks, :observed).any?
+  end
+
+  def after_5pm_in_nc(time)
+    time.hour >= 17
+  end
```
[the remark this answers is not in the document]
arinchoi03: that was my recommendation 😅 i want to say I do think it's helpful to have the logic extracted out & some kind of business logic comment. It took me a second to piece together why 1700 was an important number here. Maybe it should be renamed to:

```
def after_business_hours
```

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -17,13 +17,23 @@ class NcTaxesOwedForm < TaxesOwedForm
...
       attrs = attributes_for(:intake)
-      date = form_submitted_before_payment_deadline? ? date_electronic_withdrawal : next_available_date(Time.parse(app_time))
+      date = form_submitted_before_payment_deadline? ? date_electronic_withdrawal : @intake.next_available_date(Time.parse(app_time))
 
       @intake.update(attrs.merge(date_electronic_withdrawal: date))
     end
 
     private
 
+    def withdrawal_date_is_at_least_two_business_days_in_the_future_if_after_5pm
```
mpidcock, line 27: you're gonna hate me for being so picky, but imo this method should go at the bottom of the file because it is called last by `validate_withdrawal_date_fields` (and that's where it was before). Do we follow this convention consistently? absolutely not. but we can always strive to be better

app/helpers/date_helper.rb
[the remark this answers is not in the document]
mpidcock: works for me!

spec/models/state_file_nc_intake_spec.rb
```diff
@@ -184,25 +184,69 @@
...
     context "when submitted after the payment deadline" do
-      let(:current_time) { payment_deadline_date + 1.day }
-
       it "returns next available date" do
-        expect(intake).to receive(:next_available_date)
-        intake.calculate_date_electronic_withdrawal(current_time: current_time)
+        Time.use_zone(timezone) do
+          Timecop.freeze(payment_deadline_date + 1.hour) do
+            expect(intake).to receive(:next_available_date)
+            intake.calculate_date_electronic_withdrawal(current_time: Time.zone.now)
```
mpidcock, line 199: I hate to continue to be really picky about this, but our server time is UTC, not in the local timezone, so you can't expect the `current_time` parameter to be pre-converted. I don't think this tests that it is converted correctly.

mpidcock (commented):
I think this looks really good, but I am going to be a stickler about the tests

mpidcock (approved):
Approved! (as long as tests pass)

app/forms/state_file/nc_taxes_owed_form.rb
```diff
@@ -13,15 +13,37 @@ class NcTaxesOwedForm < TaxesOwedForm
...
-    with_options unless: -> { payment_or_deposit_type == "mail" || !form_submitted_before_payment_deadline? } do
-      validate :withdrawal_date_is_after_today
-      validate :withdrawal_date_is_not_on_a_weekend
-      validate :withdrawal_date_is_not_a_federal_holiday
-      validate :withdrawal_date_is_at_least_two_business_days_in_the_future_if_after_5pm
+    validate :validate_withdrawal_date_fields
+
+    def save
+      attrs = attributes_for(:intake)
+      date = form_submitted_before_payment_deadline? ? date_electronic_withdrawal : @intake.next_available_date(Time.parse(app_time))
```
arinchoi03, line 20: @mpidcock @DrewProebstel I'm going to ask again why we'd need to update the `date_electronic_withdrawal` on this intake if the user is not really updating that after payment deadline? I guess at the end of the day it doesn't matter too much, but seems unnecessary. No need to block on this though

mpidcock: It's a good question! We still have to provide a `date_electronic_withdrawal` on the xml, even after the deadline, but the filer can no longer select the date. For all other States we set this to the current day, but NC has special logic where we have to advance it to the next available day.

arinchoi03: ohhh i see what we're doing now. i thought we were just setting that in the xml but this is ok too

PR #5889, opened 2025-04-10
FYST-2000 Make MD payment deadline change on the 15th

mpidcock:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2000

## What was done?
- Updated the way state-specific payment deadlines are decided to use app_time, since MD's deadline changes on tax day
- MD is the only one with special logic, and it's even more special than I originally thought
   - Before tax day the deadline is april 30th
   - On and after tax day, the deadline is april 15th
   - All other states the deadline is always april 15th
- This change does end up hardcoding "April 15" in a few places as "tax day", but I'm not very concerned about this for now
## How to test?
- Updated the unit tests for this logic
- Manually test using filers that owe taxes
- Use session toggles to test times within an hour or so of April 15th to ensure timezone math is correct
- Risk Assessment
  - This affects scheduling payments, and if implemented incorrectly can result in invalid xml
  - Testing with session toggles is the only way to verify this will work in preparation for tax day


squanto (approved):
no template changes needed, very nice

app/services/state_file/state_information_service.rb
```diff
@@ -57,20 +56,28 @@ def survey_link(state_code, locale: nil)
...
-        datetime.in_time_zone(timezone).to_date.before?(payment_deadline.in_time_zone(timezone).to_date)
+        datetime.in_time_zone(timezone).to_date.before?(payment_deadline_date)
+      end
+
+      # Maryland has different payment deadline logic from all our other States
+      # 1. If filing before April 16: payment can be scheduled until April 30th
+      # 2. If filing on or after April 16: payment cannot be scheduled (same as other States)
+      def get_md_payment_deadline(datetime)
+        timezone = StateInformationService.timezone("md")
+        before_april_16 = datetime.in_time_zone(timezone).to_date.before?(Date.new(datetime.year, 4, 16))
```
mrotondo, line 78: IMO we shouldn't be adding dates to the codebase but should put them in configuration, ideally using the dates that are already in `config.tax_year_filing_seasons`. I think we could make those into date ranges rather than datetime ranges, as I think it would simplify their use here (and in the one place we use them elsewhere).

mpidcock: I will make a ticket for a follow-up refactor

PR #5890, opened 2025-04-10
NJ 330 - update content for estimated payments,  and for extensions feature flag content

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/330

## What was done?
- Update extensions description content behind feature flag
- General content updates (eg, overpayments label, review screen)

## How to test?
Toggle `extension_period` feature flag to see changes

#### English
image
image
image

#### Spanish
image
image
image

#### English
image
image
image

#### Spanish
image
image
image

### DIFF
--- a/app/views/state_file/questions/eligible/_nj_credits_unsupported.html.erb
+++ b/app/views/state_file/questions/eligible/_nj_credits_unsupported.html.erb
@@ -11,7 +11,11 @@
     <div class="spacing-below-25">
       <b><%=t('.also_unsupported_title') %></b>
       <ul class="list--bulleted">
-        <%=t('.also_unsupported_html') %>
+        <% if Flipper.enabled?(:extension_period) %>
+          <%=t('.also_unsupported_extensions_html') %>
+        <% else %>
+          <%=t('.also_unsupported_html') %>
+        <% end %>
       </ul>
     </div>
     <p>
--- a/app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
+++ b/app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
@@ -6,7 +6,11 @@
 <% content_for :card do %>
   <h1 class="h2"><%= title %></h1>
 
-  <%= t(".description_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% if Flipper.enabled?(:extension_period) %>
+    <%= t(".description_extensions_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% else %>
+    <%= t(".description_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% end %>
 
   <%= form_with model: @form, url: { action: :update }, local: true, method: "put", builder: VitaMinFormBuilder do |f| %>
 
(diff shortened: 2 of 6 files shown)


jachan (approved):

app/views/state_file/questions/nj_review/edit.html.erb
```diff
@@ -225,8 +225,10 @@
       <% if current_intake.has_estimated_payments_yes? %>
         <p class="text--bold spacing-below-5"><%=t(".estimated_tax_payments") %></p>
         <p><%=number_to_currency(current_intake.estimated_tax_payments || 0, precision: 0)%></p>
-        <p class="text--bold spacing-below-5"><%=t(".extension_payments") %></p>
-        <p><%=number_to_currency(current_intake.extension_payments || 0, precision: 0)%></p>
+        <% if Flipper.enabled?(:extension_period) %>
+          <p class="text--bold spacing-below-5"><%=t(".extension_payments") %></p>
```
mmazanec22, line 229: [dust] Is this a paragraph masquerading as a header? (Genuine question, can't tell from context)

mmazanec22 (approved):

app/views/state_file/questions/nj_review/edit.html.erb
aloverso: @mmazanec22 oooh nice catch. I'm actually not sure. Looking at the page further, we are using two different patterns here. I could see an argument for either `p` or `h3` for these labels, but we should at least be consistent. Which do you think?

image

mmazanec22: I think each box should have one header and then the content below, like the property tax deduction or credit box. The other heading pattern, seen in the estimated or other payments box, would seem to screen reader users that it labels the whole box, but then the "yes" underneath is maybe kind of context-less and doesn't follow the pattern.

Not for this PR but: the mismatch between visual design and markup might also be confusing for sighted and partially sighted screen reader users. It's unusual to have header elements that aren't visually distinct. It looks like the [WCAG 3 draft specifically recommends visually distinct headings](https://www.w3.org/WAI/GL/WCAG3/2020/methods/visually-distinct/).

Also not for this PR, but I wonder if using a header + a `dl` element would make the most sense.

app/views/state_file/questions/eligible/_nj_credits_unsupported.html.erb
```diff
@@ -11,7 +11,11 @@
     <div class="spacing-below-25">
       <b><%=t('.also_unsupported_title') %></b>
       <ul class="list--bulleted">
-        <%=t('.also_unsupported_html') %>
+        <% if Flipper.enabled?(:extension_period) %>
+          <%=t('.also_unsupported_extensions_html') %>
+        <% else %>
+          <%=t('.also_unsupported_html') %>
+        <% end %>
       </ul>
```
tahsinaislam, line 19: [dust] could this logic live in the controller with an instance variable? `@also_unsupported_text = Flipper.enabled?(:extension_period) ? t('.also_unsupported_extensions_html') : t('.also_unsupported_html')`

app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
```diff
@@ -6,7 +6,11 @@
 <% content_for :card do %>
   <h1 class="h2"><%= title %></h1>
 
-  <%= t(".description_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% if Flipper.enabled?(:extension_period) %>
+    <%= t(".description_extensions_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% else %>
+    <%= t(".description_html", filing_year: current_tax_year, prior_year: current_tax_year-1) %>
+  <% end %>
```
tahsinaislam, line 14: same with this logic using an instance variable

tahsinaislam (approved):
looks good just have some suggestions!

app/views/state_file/questions/eligible/_nj_credits_unsupported.html.erb
aloverso: @tahsinaislam Do you see a concern here where that would live in the shared controller for every state (this is the `eligible` controller) whereas in this case, the logic is limited to the NJ component

app/views/state_file/questions/nj_estimated_tax_payments/edit.html.erb
aloverso: @tahsinaislam I experimented with this and ultimately I think it's a little clearer & shorter to keep this in the erb file rather than in the controller, keeps all the translation references in one place. Going to merge this as-is. Let me know if that's not okay and I'll do a followup!

PR #4501, opened 2024-04-16
Modify the check for verified emails to also look for verification on…

mpidcock:
… a matching intake (using ssn & address)

PR #5891, opened 2025-04-10
Fyst 2004 update post deadline reminder fyst notification behavior

anisharamnani:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-2004

## What was done?
* extract deadline reminder query into `state_file_base_intake` to be used by pre and post deadline reminders
* change post deadline reminder to be sent this year with two schedules cron jobs on April 17 and 23
## How to test?
- rspec tests
- we’ll also test this on demo for both email & text


app/services/state_file/send_post_deadline_reminder_service.rb
```diff
@@ -1,46 +1,13 @@
...
-    BATCH_SIZE = 10
-    HOURS_AGO = 24
+    BATCH_SIZE = 10 # change batch size
 
     def self.run
-      cutoff_time_ago = HOURS_AGO.hours.ago
-      intakes_to_notify = StateFile::StateInformationService.state_intake_classes.map do |class_object|
+      intakes_to_notify = []
 
-        # First we get all intake ids by email address
```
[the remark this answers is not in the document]
anisharamnani: but, if there is something vital i’m missing, please let me know :)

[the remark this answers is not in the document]
arinchoi03: I think maybe this has to do with checking multiple intakes that share a `hashed_ssn`/ and matching on email & if any of them have already been submitted (has efile_submissions)
```
          # New criteria - gonna see if any associated intakes have submissions
```
here's the PR that added the logic: https://github.com/codeforamerica/vita-min/pull/4501

Currently, I think it _is_ possible that a user could create multiple intakes with different verified contact info (with the same hashed_ssn). And if they submitted one and not the other, we probably don't want to send a notification to the user if they have another intake submitted.

@mpidcock could probably explain this better as she worked on the original PR linked. We don't do this level of checking for any other jobs & I wonder if we need to do this for all the other messages...

spec/tasks/send_post_deadline_reminder_service_spec.rb
```diff
@@ -2,71 +2,30 @@
 require 'rails_helper'
 
 describe 'state_file:post_deadline_reminder' do
+  include_context "rake"
```
arinchoi03, line 5: what does this allow us to do?

spec/tasks/send_pre_deadline_reminder_service_spec.rb
```diff
@@ -14,73 +14,19 @@
...
-           message_tracker: { "messages.state_file.finish_return" => (Time.now - 2.hours).utc.to_s }
-  }
-  let!(:md_intake_disqualifying_df_data) {
-    create :state_file_md_intake,
-           df_data_imported_at: 2.minutes.ago,
-           email_address: 'test+01@example.com',
-           email_address_verified_at: 5.minutes.ago,
-           email_notification_opt_in: 1
-  }
-
```
arinchoi03, line 69: wondering if these fixtures were the test for the hashed_ssn/shared email scenarios

arinchoi03 (commented):
Left some comments!

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
anisharamnani: ah, do you know if it’s possible for there to be multiple intakes with df data?

spec/tasks/send_pre_deadline_reminder_service_spec.rb
anisharamnani: this is for the pre-deadline reminder, but the post deadline reminder did not test for the hashed ssn. that was also confusing.

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
arinchoi03: I think so -- since we ourselves can use sample df data over and over to create new intakes

spec/tasks/send_pre_deadline_reminder_service_spec.rb
arinchoi03: o yeah that's not great

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
anisharamnani: yeah 🤔 i am wondering if it is possible to create multiple intakes with df data. does that serve the same purpose of the hashed ssn? if it doesn’t i’m happy to add this to the query as well. 

& i also wonder if it is needed for pre deadline reminder as well?

app/services/state_file/send_post_deadline_reminder_service.rb
```diff
@@ -1,46 +1,13 @@
 module StateFile
   class SendPostDeadlineReminderService
-    BATCH_SIZE = 10
-    HOURS_AGO = 24
+    BATCH_SIZE = 10 # change batch size
```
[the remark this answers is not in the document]
anisharamnani: another good call out, thank you! i put the note and forgot it.

app/models/state_file/automated_message/post_deadline_reminder.rb
```diff
@@ -10,7 +10,7 @@ def self.after_transition_notification?
     end
 
     def self.send_only_once?
-      true
+      false
```
mpidcock, line 13: I think this should stay true - we only want to send it to a client one time

app/models/state_file_base_intake.rb
```diff
@@ -73,6 +73,23 @@ def self.state_code
     state_code.to_s
   end
 
+  def self.selected_intakes_for_deadline_reminder_notifications
+    self.left_joins(:efile_submissions)
+      .where(efile_submissions: { id: nil })
+      .where.not(df_data_imported_at: nil)
+      .has_verified_contact_info
```
mpidcock, line 80: Remove this check - this isn't needed for outreach messages

app/models/state_file_base_intake.rb
```diff
@@ -73,6 +73,23 @@ def self.state_code
...
+      .select(&:has_not_recently_received_finish_return_message_or_includes_disqualifying_df_data)
+  end
+
+  def has_not_recently_received_finish_return_message_or_includes_disqualifying_df_data
```
mpidcock, line 84: This is a pretty confusing method name - could you consider separating these conditions?

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
mpidcock: Hi, so sorry this took me awhile to answer, but you do not need to do this check anymore. Intakes will not have the same DF data (aka the same hashed_ssn), but last year we did see this happen and had to handle it. Thank you for checking!

app/models/state_file_base_intake.rb
anisharamnani: this is needed for demo. we’ve found that many demo intakes don’t include data for phone numbers or emails, and so this scopes the query down to check if there is at least a phone number or email present.

app/models/state_file_base_intake.rb
anisharamnani: ya, i thought about that but i want to be able to add a select to the above query & check for both these conditions at once. otherwise i will have to loop through twice. i can think of a more succinct name, but i also like that it is explicit. the logic is complicated and therefore the name is.

app/models/state_file/automated_message/post_deadline_reminder.rb
anisharamnani: the ticket says to send the notification on April 17 (4/17/2025) and April 23 (4/23/2025)

app/models/state_file_base_intake.rb
anisharamnani: if you have a better name i’m happy to change it to that. :)

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
arinchoi03: No worries! Just making sure that it is fine to send messages if they make intakes with same df data (hashed ssn) and DIFFERENT contact info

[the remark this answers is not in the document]
mpidcock: Yes that's fine, should be very rare because it means someone forgot they filed and decided to file again. For the record, the second intake would be rejected because the ssn already filed, but that would be expected behavior.

app/models/state_file/automated_message/post_deadline_reminder.rb
mpidcock: Oh, I see, we want people to get both reminders if they still haven't filed

app/models/state_file_base_intake.rb
mpidcock: oh, I see. there's `require_verification: false` in the service that handles my concern. thanks!

app/services/state_file/send_post_deadline_reminder_service.rb
```diff
@@ -1,46 +1,13 @@
 module StateFile
   class SendPostDeadlineReminderService
     BATCH_SIZE = 10
```
mpidcock, line 3: We probably don't need this small of a batch size, but since it was here before it can stay if you don't want to mess with it.

app/services/state_file/send_post_deadline_reminder_service.rb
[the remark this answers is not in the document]
mpidcock: I did a check on prod and there are no duplicate hashed_ssns in any state

mpidcock (approved):
Thanks for your patience!

app/models/state_file/automated_message/post_deadline_reminder.rb
anisharamnani: yep!

app/services/state_file/send_post_deadline_reminder_service.rb
anisharamnani: i was thinking the same. i am hoping in the follow up audit ticket can discover a better batch size.

PR #5892, opened 2025-04-10
NJ 338 - Added federal extension checkbox to NJ PDF

jachan:
## Link to pivotal/JIRA issue
- https://github.com/newjersey/affordability-pm/issues/338#issue-2983009669

## What was done?
- NJ doesn't have an XML schema field for the federal extension checkbox
- Decided to still fill in the PDF field if the user says they made federal extension payments so the user is not confused: see https://njcio.slack.com/archives/C07GYAQ4E1E/p1744293998466119

## How to test?
- Set extension_period feature flag
- Answer Yes when asked if you paid NJ extension payments
- Confirm in PDF that the checkbox is selected

- Before
- After
image

### DIFF
--- a/app/lib/pdf_filler/nj1040_pdf.rb
+++ b/app/lib/pdf_filler/nj1040_pdf.rb
@@ -100,7 +100,10 @@ module PdfFiller
         
         # signature fields
         'Date1_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Primary DateSigned")&.text,
-        'Date2_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Secondary DateSigned")&.text
+        'Date2_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Secondary DateSigned")&.text,
+
+        # Federal extension payment checkbox
+        'Check Box18': Flipper.enabled?(:extension_period) && @submission.data_source.paid_federal_extension_payments_yes? ? "Yes" : "Off"
       }
 
       dependents = get_dependents
--- a/spec/lib/pdf_filler/nj1040_pdf_spec.rb
+++ b/spec/lib/pdf_filler/nj1040_pdf_spec.rb
@@ -1041,6 +1041,7 @@ RSpec.describe PdfFiller::Nj1040Pdf do
 
     describe "disabled show_retirement_ui flag" do
       before do
+        allow(Flipper).to receive(:enabled?).and_call_original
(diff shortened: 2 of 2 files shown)


mmazanec22 (approved):

app/lib/pdf_filler/nj1040_pdf.rb
```diff
@@ -100,7 +100,10 @@ def hash_for_pdf
         
         # signature fields
         'Date1_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Primary DateSigned")&.text,
-        'Date2_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Secondary DateSigned")&.text
+        'Date2_es_:signer:date': @xml_document.at("ReturnHeaderState Filer Secondary DateSigned")&.text,
+
+        # Federal extension payment checkbox
+        'Check Box18': Flipper.enabled?(:extension_period) && @submission.data_source.paid_federal_extension_payments_yes? ? "Yes" : "Off"
```
aloverso, line 106: [dust] is it necessary to check the Flipper here? It seems like `paid_federal_extension_payments` will only ever be "Yes" if it's enabled anyways.

aloverso (approved):

app/lib/pdf_filler/nj1040_pdf.rb
jachan: I prefer to think of it this way:

If we are not in the extension period, we should NEVER check the extension box.
If we are in the extension period, we should then also evaluate the user's answer.

I think it's better to tie the logic to the flipper flag rather than relying on the `show` method of a controller to be implemented correctly.

aloverso: @jachan I'm down with this explanation! I'd like it if we tested it exhaustively (ie, if Flipper is disabled but `paid_federal_extension_payments` is "Yes") but it worries me a bit (hence my original dust comment) to be testing a situation that isn't technically possible

arinchoi03 (approved):

app/lib/pdf_filler/nj1040_pdf.rb
arinchoi03: @jachan 
> I think it's better to tie the logic to the flipper flag rather than relying on the show method of a controller to be implemented correctly.

I kind of think the same way & wasn't able to articulate it -- thanks for that!

PR #5895, opened 2025-04-10
Add `role: button` to small buttons and some others

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1963
- Maybe?

## What was done?
- Basically just went through state file views and tried to find buttons that were `link_to`s and added `role: button` to them if they didn't have it already
## How to test?
- Not sure if it's worth doing this in acceptance testing but the thing to do would be to go to all these pages and see that the buttons now behave the expected way with a screen reader? And also aren't broken? Maybe not worth it for how tedious it is though, would love some feedback
- visually looks the same


jachan (approved):
Tested some of these using VoiceOver and the elements tagged with "role: button" are being correctly read as buttons!

My personal opinion is that PM acceptance testing is not necessary here.

arinchoi03 (approved):
just one i wasn't sure about but rest looks fine to me!

PR #5897, opened 2025-04-10
GYR1-706 Twilio NoMethodError

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-706

## What was done?
- Put the attachments-to-documents step in a job that retries on NoMethodError in the case that we can't retrieve the attachments yet
- I just realized this might mean we still get the Sentry exception...is there a way to swallow it after it triggers the retry? If not I can always catch it and kick off the job manually instead of using `retry_on`, but I kind of like using `retry_on`...
## How to test?
- Not sure if there's any way to acceptance test that the retrying works honestly, but it probably makes sense to test that the attachments still show up as documents


app/services/incoming_text_message_service.rb
```diff
@@ -55,7 +40,7 @@ def self.process(params)
           phone_number: contact_record.from_phone_number,
           body: contact_record.body,
           client: contact_record.client,
-          has_documents: contact_record.documents.present?
+          has_documents: contact_record.documents.present? || params["NumMedia"].to_i > 0,
```
jenny-heath, line 43: not 100% sure about this but i wanted to have a fallback in case the job hadn't finished running and this seems like it would be accurate?

arinchoi03:
I'm just responding to the question in the description:
> I just realized this might mean we still get the Sentry exception...is there a way to swallow it after it triggers the retry? If not I can always catch it and kick off the job manually instead of using retry_on, but I kind of like using retry_on...

From the ruby doc https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
> Catch the exception and reschedule job for re-execution after so many seconds, for a specific number of attempts. If the exception keeps getting raised beyond the specified number of attempts, the exception is allowed to bubble up to the underlying queuing system, which may have its own retry mechanism or place it in a holding queue for inspection.

So it seems like at least until the retries have been exhausted, the error won't get raised, which I think is what we want here.

arinchoi03 (commented):
few comments, hoping @mrotondo comes to comment a bit more

app/jobs/process_text_message_attachments_job.rb
```diff
@@ -0,0 +1,26 @@
+class ProcessTextMessageAttachmentsJob < ApplicationJob
+  retry_on NoMethodError
```
mrotondo, line 2: this is nice! if we wanted to be even more narrow (so that we don't retry on other things weirdly being nil) we could define an empty exception class (`MissingAttachmentError`?) that inherits from StandardError and raise that when attachment[:body] is empty.

app/jobs/process_text_message_attachments_job.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def perform(incoming_text_message_id, client_id, params)
+    attachments = TwilioService.new.parse_attachments(params)
+
+    documents = attachments.map do |attachment|
+      Document.new(
+        client: Client.find(client_id),
```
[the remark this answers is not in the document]
mrotondo: I believe it's because the arguments passed to a job have to be serializable (since they need to go into a database and used later) and so we use IDs instead of full objects

app/jobs/process_text_message_attachments_job.rb
```diff
@@ -0,0 +1,26 @@
...
+        upload: {
+          io: StringIO.new(attachment[:body]),
+          filename: attachment[:filename],
+          content_type: attachment[:content_type],
+          identify: false
+        }
+      )
+    end
+
+    IncomingTextMessage.find(incoming_text_message_id).update(documents: documents)
```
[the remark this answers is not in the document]
mrotondo: IMO yes, this should blow up if it can't find the message it's supposed to be attaching things to

app/services/incoming_text_message_service.rb
```diff
@@ -22,31 +22,16 @@ def self.process(params)
...
-          }
-        )
-      end
-
       contact_record = IncomingTextMessage.create!(
         body: params["Body"],
         received_at: DateTime.now,
         from_phone_number: phone_number,
         client: client,
-        documents: documents
```
mrotondo, line 47: what did this do, and is it ok that we're losing it?

app/services/incoming_text_message_service.rb
mrotondo: (i just confirmed that a job run with `perform_now` will still retry _asynchronously_ if it fails with an error named by `retry_on`), so yeah, we may need to specify here that there are attachments even if documents is blank, and i think this is a reasonable way to do it

mrotondo: it looks like the only way this is currently used is to decide to show a link to the client's documents tab in intercom, so this should be fine.

spec/jobs/process_text_message_attachments_job_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    it "retries the job when the attachment body is nil and doesn't persist any documents" do
+      client_id = create(:client).id
+      incoming_text_message_id = create(:incoming_text_message).id
+      params = { some: "params" }
+      allow_any_instance_of(TwilioService).to receive(:parse_attachments).and_raise(NoMethodError)
+
+      expect {
+        described_class.perform_now(incoming_text_message_id, client_id, params)
+      }.to have_enqueued_job(described_class).with(incoming_text_message_id, client_id, params)
```
mrotondo, line 46: nice :)

mrotondo (approved):
looks good! my only question is about the change to the contact record, but that's just because i don't know what it is. 

i think we may need to watch this and follow up if we still see the error in sentry, probably by increasing the number of retries for this job. with exponential falloff, i expect that we'd be able to handle whatever delay twilio is imposing on the body availability with just 5 or 10 more retries on top of the current 5. but because we currently have no data on exactly how long it's taking them to make that body available, i'm fine with this going in as-is and us modifying it later if necessary

app/jobs/process_text_message_attachments_job.rb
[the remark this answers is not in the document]
arinchoi03: ok, then what will happen if `Client` is not find-able (unlikely to happen but just asking to be devil's advocate); the document will be created without a client...is that ok?

app/services/incoming_text_message_service.rb
arinchoi03: I think we're doing:
```
    IncomingTextMessage.find(incoming_text_message_id).update(documents: documents)
```
in the job instead of setting it on create

app/jobs/process_text_message_attachments_job.rb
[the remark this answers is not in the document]
arinchoi03: actually, I guess `Client.find` will likely blow up & error

mrotondo (approved):
looks good!

PR #5901, opened 2025-04-11
NJ 336 - Added headers to review screen

jachan:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/336#issue-2977593519

## What was done?
- Added headers and rearranged final review screen to match design: 
image

- aligned header structure with Shared Header, previous NJ used h2 for group headers, now follows standard structure below

```
  - <h2> section header (e.g. Deductions and Credits)
     - <white group>
        - <h3> group header (e.g. Veterans Exemption) 
```
## How to test?
- Go to review screen and verify that headers have been added and content appears correct

- Before
image
image

- After
image
image

### DIFF
--- a/spec/features/state_file/nj/complete_intake_spec.rb
+++ b/spec/features/state_file/nj/complete_intake_spec.rb
@@ -272,7 +272,8 @@ RSpec.feature "Completing a state file intake", active_job: true do
       expect(page).to be_axe_clean.within "main"
 
       groups = page.all(:css, '.white-group').count
-      expect(groups).to eq(0)
+      expect(groups).to eq(11)
+      # one white group per exemption/section
 
       h2s = page.all(:css, 'h2').count
       # one h2 per section header (e.g.Household Information)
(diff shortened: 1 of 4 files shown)


aloverso (approved):
Looks good!

spec/features/state_file/nj/complete_intake_spec.rb
```diff
@@ -272,10 +272,12 @@ def check_xml_results
...
       groups = page.all(:css, '.white-group').count
+      # one white group per exemption/section
+      expect(groups).to eq(11)
+
       h2s = page.all(:css, 'h2').count
-      expect(groups).to eq(h2s - 2)
-      # total should be h2s -1 to account for the h2 reveal button
-      # there's also an extra h2 in the veteran exemption box
+      # one h2 for each of 5 section headers (e.g Household Information), "Your refund amount" is also an h2
+      expect(h2s).to eq(5 + 1)
```
mluedke2, line 280: [dust|sand] I wonder if there's a way to make this test clearer to read? I don't know from looking at just this file that `11` or `5 + 1` are the correct numbers to be asserted against (the comment definitely helps, but then I have to also trust that the `5` in the comment is right too).

I don't have a great alternative to suggest, though. We could give them names as constants instead of using comments, but that doesn't really change anything. A comment with a link to a design document would only work if we can trust that doc to always be up to date. A more comprehensive check for the presence of each section header's text seems like way too much for now. So if nothing else comes to mind, I'm fine with it as-is.

mluedke2 (approved):

mpidcock (approved):

PR #5902, opened 2025-04-11
NJ 339 - remove payment amount box from taxes_owed screen, fill automatically instead

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/339

## What was done?
- Rather than making this just NJ-specific, add new `auto_calculate_withdraw_amount` flag to StateInformationService, any state that implements it will receive this behavior
- When `auto_calculate_withdraw_amount` is true, on the taxes_owed screen, remove the input box for `withdraw_amount` and instead auto-fill the amount on save based on taxes owed. Show an authorization text instead.

## How to test?
Use `jones mfj` for a persona that will owe taxes

image

### DIFF
--- a/app/forms/state_file/taxes_owed_form.rb
+++ b/app/forms/state_file/taxes_owed_form.rb
@@ -16,7 +16,7 @@ module StateFile
                        :app_time
 
     with_options unless: -> { payment_or_deposit_type == "mail" } do
-      validates :withdraw_amount, presence: true, numericality: { greater_than: 0 }
+      validates :withdraw_amount, presence: true, numericality: { greater_than: 0 }, if: -> { !StateFile::StateInformationService.auto_calculate_withdraw_amount(intake.state_code) }
       validate :withdraw_amount_does_not_exceed_owed_amount
       with_options if: -> { form_submitted_before_payment_deadline? } do
         validate :date_electronic_withdrawal_is_valid_date
@@ -33,6 +33,7 @@ module StateFile
 
     def save
       attrs = attributes_for(:intake)
+      attrs[:withdraw_amount] = owed_amount if StateFile::StateInformationService.auto_calculate_withdraw_amount(intake.state_code)
       @intake.update(attrs.merge(date_electronic_withdrawal: date_electronic_withdrawal))
     end
 
@@ -88,8 +89,11 @@ module StateFile
       end
     end
 
+    def owed_amount
+      intake&.calculated_refund_or_owed_amount&.abs
+    end
+
     def withdraw_amount_does_not_exceed_owed_amount
-      owed_amount = intake&.calculated_refund_or_owed_amount&.abs
       return unless self.withdraw_amount.to_i > owed_amount
 
       self.errors.add(:withdraw_amount,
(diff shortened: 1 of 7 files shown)


arinchoi03 (approved):
What if the user _does not_ want to pay the full amount, is the only other route to pay by check/mail?

app/forms/state_file/taxes_owed_form.rb
```diff
@@ -88,8 +89,11 @@ def date_electronic_withdrawal_within_range
       end
     end
 
+    def owed_amount
+      intake&.calculated_refund_or_owed_amount&.abs
```
mluedke2, line 93: [dust|sand] - I know this was the previous syntax as well so I'm presuming all is well in practice. But it still feels off to use `abs` when only owing taxes is valid for this screen. Would multiplying by `-1` be more precise? `abs` could coalesce a refund into a positive value too. I'm guessing having a refund by the point we get to this line is impossible because of upstream logic, but if that's true then `* -1` is what we want anyway, right?

mluedke2 (approved):
had a non-blocking suggestion for a follow-up to consider, but otherwise looks good to me!

app/services/state_file/state_information_service.rb
```diff
@@ -28,6 +28,7 @@ class StateInformationService
       :voucher_path,
       :w2_supported_box14_codes,
       :w2_include_local_income_boxes,
+      :auto_calculate_withdraw_amount,
```
jachan, line 31: [DUST] Great choice to put this in the `state_information_service` rather than make it NJ-specific

spec/controllers/state_file/questions/taxes_owed_controller_spec.rb
```diff
@@ -96,6 +96,36 @@
             )
           end
         end
+
+        shared_examples "withdraw amount is user-entered" do
+          it "shows withdraw_amount field" do
+            get :edit
+            expect(response.body).to include("How much do you authorize to be withdrawn from your account?")
```
jachan, line 103: [DUST] Should this reference the i18n module instead of being a static string? (same for all of these shared example tests)

app/views/state_file/questions/tax_refund/_bank_details.html.erb
```diff
@@ -4,7 +4,11 @@
   <div>
     <% if owe_taxes %>
       <%= form.hidden_field :app_time, value: app_time %>
-      <%= form.vita_min_money_field(:withdraw_amount, t('.withdraw_amount', owed_amount: taxes_owed), classes: ["form-width--long"]) %>
+      <% if StateFile::StateInformationService.auto_calculate_withdraw_amount(current_intake.state_code) %>
+        <p><%= t(".authorized_amount", owed_amount: taxes_owed) %></p>
```
jachan, line 8: [PEBBLE] Design in ticket specifies that this text should appear within a blue alert box.

jachan (commented):
Love the choice of approach, minor comments on tests + implementation doesn't appear to match the design

app/views/state_file/questions/tax_refund/_bank_details.html.erb
aloverso: [See thread](https://njcio.slack.com/archives/C079N39B74P/p1744383487220629) - it will be redesigned

jachan (approved):

aloverso:
@arinchoi03 Here's Tristan's response to your question:
> yes they need to choose an alternate payment method. I believe credit card also works. We are adding more details about alternate payment methods in a followup ticket

PR #5909, opened 2025-04-12
fully migrate nj 1099R followups to subflows, and preserve their intended functionality of showing followup pages for all 1099Rs instead of just ones with a non-zero taxable amount

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2041

## What was done?
- NJ's 1099-R followup pages were partially migrated onto the subflows code - I changed their question navigation class and what params were carried from page to page by the continue button partial, but didn't update the controller, tests or review page link.
- This PR updates their controller to use the same `RetirementIncomeSubtractionController` superclass as the other controllers doing the same job, while overriding the `self.show?` and `self.load_1099r` methods to use the full list of 1099rs instead of just those with taxable amounts, as other states do.
- It also updates the review page link to return to review before or after the full `retirement_income_deduction` step, which was already their implemented functionality, but which now is something they get for free with subflows.
## How to test?
- Go through the NJ flow using a persona with multiple 1099-Rs and verify that you can proceed past the second one. Verify that you can return to retirement income deduction from the final review page and that you return to review after using the back button from the first 1099-R or the continue button from the last 1099-R.

### DIFF
--- a/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
+++ b/app/controllers/state_file/questions/nj_retirement_income_source_controller.rb
@@ -1,68 +1,14 @@
 module StateFile
   module Questions
-    class NjRetirementIncomeSourceController < QuestionsController
+    class NjRetirementIncomeSourceController < RetirementIncomeSubtractionController
+      def followup_class = StateFileNc1099RFollowup
 
-      before_action :load_1099r
-
-      def load_1099r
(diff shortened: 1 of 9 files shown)


arinchoi03 (approved):

tahsinaislam (approved):
Lgtm!

PR #5920, opened 2025-04-14
Add NJ 1099R feature tests

jachan:
## Link to pivotal/JIRA issue
See https://njcio.slack.com/archives/C07GVSC8X3P/p1744655902159469

## What was done?
 - Add NJ tests to 1099R feature spec

## How to test?
- run test suite

- N/A


jachan:
@aloverso Definitely something I'd like @mrotondo 's perspective on. I understood the bug to be that the taxpayer could not proceed past the retirement pages due to the way that indexing was done in NJ. Mike, would this test have caught the bug?

This test lives in the `spec/features/state_file/review_page_spec.rb` file rather than in the `spec/features/state_file/nj/complete_intake_spec.rb` file because that was the file Mike suggested for this. Understand your point that this does not step through the complete flow for a user, but rather just tests the review -> edit -> review flow specifically.

Happy to rewrite these tests in the complete intake spec if that's a better place for these tests to live.

jachan:
> we didn't catch an issue that this change introduced where NJ filers with multiple 1099-Rs would be unable to proceed past the second one while answering the "retirement income source" questions.

Never mind, I'm wrong. I see two possible fixes for this:
1) Add a third 1099R 🤣 
2) Do the end to end flow in `complete_intake_spec` to more closely match how a user would experience this

Leaning towards the 2nd

mrotondo:
Anne is right - adding/modifying a complete intake spec to go through multiple (at least 2) 1099-R followups/"retirement income source" pages would be the best approach. I was a bit shortsighted with suggesting that the test follow the examples of the other review page specs. Thanks for your suggestion, Anne!

mrotondo (approved):
looks good, thanks!

aloverso (approved):
looks good!

PR #5924, opened 2025-04-15
NJ 343 - add federal extensions response to review screen

aloverso:
## Link to pivotal/JIRA issue
https://github.com/newjersey/affordability-pm/issues/343

## What was done?
- Add federal extension to review screen
- Update federal extension page to use continue button with "return to review" functionality

## How to test?
Any persona with `:extension-period` enabled

image

### DIFF
--- a/app/views/state_file/questions/federal_extension_payments/edit.html.erb
+++ b/app/views/state_file/questions/federal_extension_payments/edit.html.erb
@@ -21,6 +21,6 @@
       ]) %>
     </div>
 
-    <%= f.continue %>
+    <%= render "state_file/questions/shared/continue_button", f: f %>
   <% end %>
 <% end %>
--- a/app/views/state_file/questions/nj_review/edit.html.erb
+++ b/app/views/state_file/questions/nj_review/edit.html.erb
@@ -230,6 +230,19 @@
     </div>
   </section>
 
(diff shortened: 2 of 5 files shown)


app/views/state_file/questions/federal_extension_payments/edit.html.erb
```diff
@@ -21,6 +21,6 @@
       ]) %>
     </div>
 
-    <%= f.continue %>
+    <%= render "state_file/questions/shared/continue_button", f: f %>
```
jachan, line 24: [DUST] Can you explain why this change is necessary?

jachan (approved):
Looks good! Quick question about changing `f.continue` to a `render` block

mluedke2 (approved):

app/views/state_file/questions/federal_extension_payments/edit.html.erb
aloverso: @jachan The shared continue button includes this:
```
  <%= hidden_field_tag "return_to_review", params[:return_to_review] %>
```

Without this hidden field present, then the controller doesn't know that the `return_to_review` param is present, and doesn't know to send the user back to the review page when they click continue (versus just moving them along to the next controller in the flow)

PR #5935, opened 2025-04-16
clean up extension payments xmls

DrewProebstel:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-2030

## What was done?
- Only show the extension payment fields if they are positive.
## How to test?
- flip extension_period tag then submit returns with 0 extension payments.

### DIFF
--- a/app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
+++ b/app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
@@ -151,7 +151,7 @@ module SubmissionBuilder
               end
               xml.TotalPaymentAndCredits do
                 xml.AzIncTaxWithheld calculated_fields.fetch(:AZ140_LINE_53)
-                xml.ExtPayment calculated_fields.fetch(:AZ140_LINE_55) if Flipper.enabled?(:extension_period)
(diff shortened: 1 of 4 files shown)


app/lib/submission_builder/ty2022/states/az/az_return_xml.rb
```diff
@@ -151,7 +151,7 @@ def documents_wrapper
               end
               xml.TotalPaymentAndCredits do
                 xml.AzIncTaxWithheld calculated_fields.fetch(:AZ140_LINE_53)
-                xml.ExtPayment calculated_fields.fetch(:AZ140_LINE_55) if Flipper.enabled?(:extension_period)
+                xml.ExtPayment calculated_fields.fetch(:AZ140_LINE_55) if Flipper.enabled?(:extension_period) && calculated_fields[:AZ140_LINE_55].positive?
```
arinchoi03, line 154: I'm seeing the requirement is that we don't include this if it's `0`. I don't think any of these values are going to be negative, if they _happen to be_ (b/c the calculations have a bug or we accidentally allow negative values from the front-end) we should include it to flag issues with this submission.

I'd say change the logic to `.zero?`

DrewProebstel: ok good point.

arinchoi03 (approved):

PR #5941, opened 2025-04-19
FYST-2063 tighten requirements for sending deadline reminders

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2063
- ???

## What was done?
- 2 things:
  1. filter out intakes that have a duplicate (same ssn ~~and same contact~~) that has a submission
  2. fix the existing method because it was doing an if/else that shouldn't have been an if/else
  3. moved some state file base intake method testing specs from az intake spec to the base intake specs
## How to test?
- Added to and refactored the az intake spec
- Not sure how acceptance testing for these things has worked


app/models/state_file_base_intake.rb
```diff
@@ -78,16 +78,21 @@ def self.selected_intakes_for_deadline_reminder_notifications
       .where(efile_submissions: { id: nil })
       .where.not(df_data_imported_at: nil)
       .has_verified_contact_info
-      .select(&:should_not_be_sent_reminder)
```
jenny-heath, line 81: i renamed this because we are selecting the people who should be sent the reminder? i think??

app/models/state_file_base_intake.rb
```diff
@@ -78,16 +78,21 @@ def self.selected_intakes_for_deadline_reminder_notifications
...
-  end
-
-  def should_not_be_sent_reminder
-    if message_tracker.present? && message_tracker["messages.state_file.finish_return"]
-      finish_return_msg_sent_time = Time.parse(message_tracker["messages.state_file.finish_return"])
-      finish_return_msg_sent_time < 24.hours.ago
-    else
-      !disqualifying_df_data_reason.present?
-    end
```
jenny-heath, line 90: having this as an if/else meant that if you had received a message more than 24 hours ago but also had disqualifying data, you'd still receive the message, which is wrong

app/models/state_file_base_intake.rb
```diff
@@ -78,16 +78,21 @@ def self.selected_intakes_for_deadline_reminder_notifications
...
+    received_reminder_recently = if message_tracker.present? && message_tracker["messages.state_file.finish_return"]
+                                   finish_return_msg_sent_time = Time.parse(message_tracker["messages.state_file.finish_return"])
+                                   finish_return_msg_sent_time > 24.hours.ago
+                                 else
+                                   false
+                                 end
+    matching_ssn_with_submission = self.class.where(hashed_ssn: hashed_ssn).excluding(self)
+    matching_email = matching_ssn_with_submission.where.not(email_address: nil).where(email_address: email_address)
+    matching_phone = matching_ssn_with_submission.where.not(phone_number: nil).where(phone_number: phone_number)
+    has_duplicate_with_submission = (matching_email + matching_phone).any? { |intake| intake.efile_submissions.any? }
```
jenny-heath, line 94: i kind of hate this please refactor if you have better ideas

arinchoi03: want to note that @mrotondo worked on a similar story to prevent matching hashed_ssn submissions & the difference between the two is that this PR cares whether the other intake has the same contact info while the other PR only cares if the intake has the same `hashed_ssn`.

Do we want to send duplicate messages for pre/post deadline reminders if a person has two intakes, both with same `hashed_ssn` but different contact info (email/phone_number), if one of the intakes have a submission? I think, probably not, but maybe we should clarify

app/models/state_file_base_intake.rb
```diff
@@ -78,16 +78,21 @@ def self.selected_intakes_for_deadline_reminder_notifications
...
+  def should_be_sent_reminder
+    received_reminder_recently = if message_tracker.present? && message_tracker["messages.state_file.finish_return"]
+                                   finish_return_msg_sent_time = Time.parse(message_tracker["messages.state_file.finish_return"])
+                                   finish_return_msg_sent_time > 24.hours.ago
+                                 else
+                                   false
+                                 end
+    matching_ssn_with_submission = self.class.where(hashed_ssn: hashed_ssn).excluding(self)
+    matching_email = matching_ssn_with_submission.where.not(email_address: nil).where(email_address: email_address)
+    matching_phone = matching_ssn_with_submission.where.not(phone_number: nil).where(phone_number: phone_number)
```
arinchoi03, line 93: i kind of wonder if this `.where.not(email_address: nil)` and `.where.not(phone_number: nil)` is really saving us much -- we already scoped down the collection to `matching_ssn_with_submission` which shouldn't be huge 🤞

app/models/state_file_base_intake.rb
arinchoi03: discussed in standup - will take over and add some refactoring commits to use the same logic from the above linked PR

app/models/state_file_base_intake.rb
```diff
@@ -72,26 +72,37 @@ def self.state_code
     end
     state_code.to_s
   end
+  delegate :state_code, to: :class
```
arinchoi03, line 75: just moving the code closer to relevant spot

app/models/state_file_base_intake.rb
```diff
@@ -72,26 +72,37 @@ def self.state_code
...
+    !received_reminder_recently && !disqualifying_df_data_reason.present? && !other_intake_with_same_ssn_has_submission?
+  end
+
+  def other_intake_with_same_ssn_has_submission?
+    return false if hashed_ssn.nil?
+    StateFile::StateInformationService.state_intake_classes.excluding(StateFileNyIntake).any? do |intake_class|
+      intakes = intake_class
+                  .where(hashed_ssn: hashed_ssn)
+                  .where.associated(:efile_submissions)
+      intakes = intakes.where.not(id: id) if intake_class == self.class
```
arinchoi03, line 101: except(self) wasn't working here, wrote a test that should break if we use that instead

spec/models/state_file_az_intake_spec.rb
```diff
@@ -496,45 +496,88 @@
...
+    }
+    let!(:az_intake_submitted_duplicate) {
+      create :state_file_az_intake,
+             email_address: "test@example.com",
+             email_address_verified_at: 1.hour.ago,
+             email_notification_opt_in: 1,
+             phone_number: nil,
+             df_data_imported_at: 2.minutes.ago,
+             hashed_ssn: "111443333"
+    }
```
arinchoi03, line 538: Removing one of the duplicate intakes, since I don't understand the need for both `az_intake_has_submitted_duplicate` and `az_intake_submitted_duplicate`

spec/models/state_file_base_intake_spec.rb
```diff
@@ -442,4 +441,81 @@
       end
     end
   end
+
+  describe "#other_intake_with_same_ssn_has_submission?" do
+    let(:intake) { create :state_file_nc_intake, hashed_ssn: hashed_ssn }
```
arinchoi03, line 446: this is a method on every state, but only testing with nc intake, is that ok? Also noting that the specs were added to `state_file_az_intake` only for the `should_be_sent_reminder` methods, but maybe it should be moved into the base intake since this method should apply to every state?

mrotondo: Agreed, I think - behavior in `StateFileBaseIntake` should be tested in `state_file_base_intake_spec.rb`

mrotondo: And to your other question - as long as it's not overridden on any of the subclasses, I think it's fine to only test it on an instance of one of the subclasses.

mrotondo (approved):
Looks good, with a non-blocking agreement with your idea about moving the `should_be_sent...` tests into the base intake spec.

mrotondo (approved):

PR #5943, opened 2025-04-22
check if we capped w2 state wages and show an error if so

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2048

## What was done?
- We already have a way to show a blocking error on the income review page with a "please review" box next to the offending income form
- I reused this for a new error on W-2s that should only show up if we modified their state wages by capping it to a maximum with the code added in https://github.com/codeforamerica/vita-min/commit/06ec557be385797650cf6b3506f20de2537cde98
## How to test?
- Go through the flow for any state with a persona with a W2
- Modify the W2 in the XML editor to set state wages to a number greater than $9999999999.99
- On the income review screen, verify that a blocking error shows up, and that there is a warning next to the correct W2
- On the W2 edit form, verify that the  state wages has been capped
- Verify that you cannot submit the form without modifying the state wages amount
- Verify that if you modify it to a lower amount, you can succesfully submit the W2 form
- Verify that if you modify it to a higher amount, it gets capped again and fails to submit

### DIFF
--- a/app/models/state_file_base_intake.rb
+++ b/app/models/state_file_base_intake.rb
@@ -1,4 +1,6 @@
 class StateFileBaseIntake < ApplicationRecord
+  DB_NUMERIC_MAX = 9_999_999_999.99
+
   self.ignored_columns = [:df_data_import_failed_at, :bank_name]
 
   devise :lockable, :unlock_strategy => :time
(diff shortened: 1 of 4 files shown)


mrotondo:
The last verification (that the number is re-capped if the form is submitted with a value higher than our max) is actually bugged at the moment (and has been) - I still need to implement that.

mrotondo:
> The last verification (that the number is re-capped if the form is submitted with a value higher than our max) is actually bugged at the moment (and has been) - I still need to implement that.

Belay that - This is a problem beyond the scope of this story, so I'm going to make a new story to deal with this issue more broadly.

DrewProebstel (approved):

PR #5946, opened 2025-04-22
Fyst 2064 md answer to disability proof question not automatically selected

DrewProebstel:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2064

## What was done?
- Moves the data fields in the MD disability page to conditional partials to eliminate the bug where radio buttons were not filled when returning to the page
- Changes HoneyCrisp to allow multiple follow ups. Definitely a little risky to do.
## How to test?
- navigate to the  MD disability page. fill out "yes, yes" then use the review page to return to the permanent disability page


app/views/state_file/questions/md_permanently_disabled/edit.html.erb
```diff
@@ -36,39 +36,17 @@
...
+      <% end %>
 
-        <div class="white-group">
-          <%= f.cfa_radio_set(:spouse_proof_of_disability_submitted, label_text: t(".spouse_proof_question_html"), collection: [
-            { value: "yes", label: t("general.affirmative") },
-            { value: "no", label: t("general.negative") },
-          ]) %>
-        </div>
-      </div>
+      <% if should_warn_about_pension_exclusion && current_intake.filing_status_mfj? %>
```
mrotondo, line 47: It looks like this condition is the same as the one on line 43 - is that intended?

DrewProebstel: Good catch. Totally not needed.

arinchoi03:
does this work for `spouse_proof_of_disability_submitted` answer persisting? I assume it's broken b/c the spouse & both partials are being shown at the same time. 

I remember when we discussed this similar ticket for ID, we were talking about creating separate controllers, and now I feel like maybe that's what is needed to happen here :(

app/views/state_file/questions/md_permanently_disabled/_both.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="question-with-follow-up__follow-up" id="both-disability-proof">
+  <div class="white-group">
+    <%= f.cfa_radio_set(:primary_proof_of_disability_submitted, label_text: t("state_file.questions.md_permanently_disabled.edit.primary_proof_question_html"), collection: [
+      { value: "yes", label: t("general.affirmative") },
+      { value: "no", label: t("general.negative") },
+    ]) %>
+  </div>
```
arinchoi03, line 7: if you put these as partials `_primary` here it may make more sense

arinchoi03 (approved):
👏 stellar work and this edit html looks a lot better than it used to be!

app/views/state_file/questions/md_permanently_disabled/edit.html.erb
```diff
@@ -36,39 +35,13 @@
...
-
-        <div class="white-group">
-          <%= f.cfa_radio_set(:spouse_proof_of_disability_submitted, label_text: t(".spouse_proof_question_html"), collection: [
-            { value: "yes", label: t("general.affirmative") },
-            { value: "no", label: t("general.negative") },
-          ]) %>
-        </div>
-      </div>
+      <% if should_warn_about_pension_exclusion && current_intake.filing_status_mfj? %>
+        <%= render partial: "state_file/questions/md_permanently_disabled/spouse", locals: { f: f } %>
```
arinchoi03, line 43: i don't know if these need to be partials anymore but ambivalent

DrewProebstel: I like them as partials but I am happy to change it back.

arinchoi03: Eh just keep them 🤷🏻‍♀️

mrotondo (approved):
approving again with some non-blocking refactoring suggestions

app/forms/state_file/md_permanently_disabled_form.rb
```diff
@@ -27,6 +27,27 @@ def save
...
+    end
+
+    def self.mfj_disability_to_disabled_attributes_hash
+    {
+      both: {primary: "yes", spouse: "yes"},
+      none: {primary: "no", spouse: "no"},
+      primary: {primary: "yes", spouse: "no"},
+      spouse: {primary: "no", spouse: "yes"},
+    }
+  end
```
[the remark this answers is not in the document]
mrotondo: you could also use tuples (e.g. `("yes", "yes"), ("no", "no")`) as instead of hashes as keys instead if you feel uncomfortable using hashes as keys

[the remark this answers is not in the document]
DrewProebstel: This is how Arin did it in the ID disability page so I think I want to keep it as is for consistency.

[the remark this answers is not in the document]
arinchoi03: This was brought from the ID implementation! And I had it this way bc I needed things in the current format for saving the mfj_disability to the intake appropriately (and we have one source of truth in terms of mfj_disability to intake attr conversion -- but maybe I made things more complicated? If this hash isn't used again yes I'd say flip it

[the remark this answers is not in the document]
mrotondo: recording for posterity that i've retracted the suggestion to use hashes as hash keys, and instead recommended a method that takes the primary & spouse disability values as args and has a case statement where the two are combined into an array, as so:
```
def xy(x, y)
  case [x,y]
  when [1, 1]
    :first
  when [0, 0]
    :second
    ... etc ...
  end
end

File: app/forms/state_file/id_disability_form.rb [hunk 5]
```diff
@@ -37,12 +37,20 @@ def save
...
+          mfj_disability = "both"
+        when ["no", "no"]
+          mfj_disability = "none"
+        when ["yes", "no"]
+          mfj_disability = "primary"
+        when ["no", "yes"]
+          mfj_disability = "spouse"
+        else
+          mfj_disability = nil
+        end
```
arinchoi03, line 52: nittiest of nits
```
        mfj_disability = case [intake.primary_disabled, intake.spouse_disabled]
               when ["yes", "yes"]
                 "both"
               when ["no", "no"]
                  "none"
               end
```

PR #5947, opened 2025-04-22
Add positive money field validation for /extension-payments and /apply-refund pages

embarnard:
## Link to pivotal/JIRA issue

## What was done?
- Alternatives considered
## How to test?
  - Test data used
- Risk Assessment
- Before
- After

### DIFF
--- a/app/validators/positive_money_field_validator.rb
+++ b/app/validators/positive_money_field_validator.rb
@@ -1,0 +1,15 @@
+class PositiveMoneyFieldValidator < ActiveModel::EachValidator
+  def validate_each(form, attribute, value)
+    if value.blank?
+      form.errors.add(attribute, form.payment_msg)
+      return
+    end
+
+    unless value.to_s.match?(/\A-?\d+(?:\.\d+)?\z/)
+      form.errors.add(attribute, I18n.t("validators.not_a_number"))
+      return
+    end
+
+    form.errors.add(attribute, form.payment_msg) if value.to_d <= 0
+  end
+end
--- a/spec/validators/positive_money_field_validator_spec.rb
+++ b/spec/validators/positive_money_field_validator_spec.rb
@@ -1,0 +1,83 @@
+require "rails_helper"
+
+describe PositiveMoneyFieldValidator do
+  before do
+    @validatable = Class.new do
+      include ActiveModel::Validations
+      validates_with PositiveMoneyFieldValidator, attributes: :extension_payments_amount
+      attr_accessor  :extension_payments_amount
+
+      def payment_msg
+        "class specific error message"
+      end
+    end
+  end
+
+  let(:extension_payments_amount) { nil }
+
+  subject { @validatable.new }
+
+  context 'is a positive decimal' do
+    before do
+      allow(subject).to receive(:extension_payments_amount).and_return '1.2'
+    end
+
+    it 'is valid' do
+      expect(subject).to be_valid
+    end
+  end
+
+  context 'value is 0.2' do
+    before do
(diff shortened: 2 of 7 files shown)


app/validators/positive_money_field_validator.rb
```diff
@@ -0,0 +1,15 @@
+class PositiveMoneyFieldValidator < ActiveModel::EachValidator
+  def validate_each(form, attribute, value)
+    if value.blank?
+      form.errors.add(attribute, form.error_msg_if_blank_or_zero)
+      return
+    end
+
+    unless value.to_s.match?(/\A-?\d+(?:\.\d+)?\z/)
```
embarnard, line 8: This is the numeric character check will match values such as "234", "1.2", "5.0" and "-30.45". Will not match values such as "1,000", "jdkfjsd", "1x", ".5" and "5."

arinchoi03 (commented):
i'm guessing we want to use this positive money validator in a lot more places in the future, if we ever add more pages or if we refactor existing ones -- maybe worth flagging to the team something like this exists now.

added some questions about the values not considered valid, otherwise the code looks good to me!

app/validators/positive_money_field_validator.rb
```diff
@@ -0,0 +1,15 @@
...
+      form.errors.add(attribute, I18n.t("validators.not_a_number"))
+      return
+    end
+
+    form.errors.add(attribute, form.error_msg_if_blank_or_zero) if value.to_d <= 0
```
arinchoi03, line 13: wonder if there should be a default `error_msg_if_blank_or_zero` if the form does not define one. I guess there is no case like in real life use & maybe every money field that is validated for this will always have a very unique validation message.

app/validators/positive_money_field_validator.rb
embarnard: Yes `-30.45` should not be a valid input but we should show the other error message and not the "this is not a number" validation message, so it'll be caught but the other statement below.

I'll ask prod about these values "5. or .5 or 1,000"

app/validators/positive_money_field_validator.rb
embarnard: Good idea! I'll add "This field can not be blank. Please enter a number that is greater than 0." since its more general

app/validators/positive_money_field_validator.rb
embarnard: @arinchoi03 chatted with product and decided to allow these values as well, made the change here: https://github.com/codeforamerica/vita-min/pull/5947/commits/9b5062bc373b002b9ce174cc7f05e09852134d58 ; thanks for lifting that!

app/validators/positive_money_field_validator.rb
embarnard: also added the default here: https://github.com/codeforamerica/vita-min/pull/5947/commits/f57400c574aa3ab3ec9c7f9ffdb07090e464b6f3

arinchoi03 (approved):
one suggestion for test but otherwise lgtm!

spec/validators/positive_money_field_validator_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+
+  context 'value is .5' do
+    before do
+      allow(subject).to receive(:extension_payments_amount).and_return '.5'
+    end
+
+    it 'is valid' do
+      expect(subject).to be_valid
+    end
+  end
```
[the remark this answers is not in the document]
embarnard: Good idea! just added

PR #5949, opened 2025-04-23
FYST-1911 Remove trailing whitespace from email address when generating token

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1911
- yes

## What was done?
- This was way more confusing than I thought it was going to be! A summary of what I found, not including all of the wrong turns 🤦🏻‍♀️:
  - The issue seems to be with trailing whitespace on the emails being entered in the PYA flow. The form validation that happens in the flow (`StateFile::ArchivedIntakes::EmailAddressForm`) allows trailing whitespace but the validation that happens on generation of the access token (`EmailAccessToken#one_or_more_valid_email_addresses`) does not allow it, so when the token is generated with a `!` from `ArchivedIntakeEmailVerificationCodeJob`, it throws an error.
  - Incidentally, the `StateFileArchivedIntake` model also allows trailing whitespace on emails so we have some in the db like this. Some of them could be old but in at least 2 cases we have created a new archived intake record when an existing one with the stripped email address already exists. Not great! These people probably couldn't access their PDFs
- I've done 2 things:
  - stripped whitespace from email addresses during the PYA flow. it seems silly to show a validation error when all you did was add a space on the end
  - made the email validation on the access token model match the email validation everywhere else. it probably wasn't intentional to have only this one validator care about whitespace around emails.
- open question: should we A) go into the DB and strip whitespace around emails that made it in with whitespace and/or B) try to identify clients who couldn't get their pdf due to including whitespace around their email and do remediation?
## How to test?
- I think all you'd have to do is fill out the PYA flow with a trailing whitespace on the email and see that you actually do receive the verification code email. Try it on another review app to see that you don't get the email. EDIT: i tried this and it works

### DIFF
--- a/spec/forms/state_file/archived_intakes/email_address_form_spec.rb
+++ b/spec/forms/state_file/archived_intakes/email_address_form_spec.rb
@@ -1,6 +1,15 @@
 require "rails_helper"
 
 RSpec.describe StateFile::ArchivedIntakes::EmailAddressForm do
+  describe "before validation" do
+    it "strips the whitespace from the email" do
+      form = StateFile::ArchivedIntakes::EmailAddressForm.new(email_address: "test@example.com ")
+      form.valid?
+
+      expect(form.email_address).to eq "test@example.com"
+    end
+  end
+
   describe "#valid?" do
     context "when the email address is valid" do
       it "returns true" do
--- a/app/forms/state_file/archived_intakes/email_address_form.rb
+++ b/app/forms/state_file/archived_intakes/email_address_form.rb
@@ -1,7 +1,10 @@
 module StateFile
   module ArchivedIntakes
     class EmailAddressForm < Form
+      include FormAttributes
+
       attr_accessor :email_address
+      before_validation_squish :email_address
 
       validates :email_address, presence: true, 'valid_email_2/email': true
     end
--- a/app/forms/state_file/archived_intakes/email_address_form.rb
+++ b/app/forms/state_file/archived_intakes/email_address_form.rb
@@ -4,7 +4,6 @@ module StateFile
       attr_accessor :email_address
 
       validates :email_address, presence: true, 'valid_email_2/email': true
-
     end
   end
 end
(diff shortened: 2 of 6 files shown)


spec/forms/state_file/archived_intakes/email_address_form_spec.rb
```diff
@@ -1,6 +1,15 @@
 require "rails_helper"
 
 RSpec.describe StateFile::ArchivedIntakes::EmailAddressForm do
+  describe "before validation" do
+    it "strips the whitespace from the email" do
+      form = StateFile::ArchivedIntakes::EmailAddressForm.new(email_address: "test@example.com ")
+      form.valid?
```
embarnard, line 7: what is this line doing? do we expect the form to be valid

jnf:
yes, we should update the existing records so that they are consistent with the current validation.
I am interested in how many people have run into this issue. Can we tell that from the audit logs?

jenny-heath:
@jnf i'm not sure how to use the logs for this but i think we can tell from the database. we'd just have to look for records that, once whitespace is stripped from email, have a record with a matching email.

should we be looking into that before merging this change...?

powersurge360 (approved):
Seems alright. I think jey had a comment though before it can be merged.

mrotondo:
> @jnf i'm not sure how to use the logs for this but i think we can tell from the database. we'd just have to look for records that, once whitespace is stripped from email, have a record with a matching email.
> 
> should we be looking into that before merging this change...?

there are 8 intakes that this change would affect on prod:
```
irb(main):003> StateFileArchivedIntake.where("email_address LIKE '% '").count
=> 7
irb(main):004> StateFileArchivedIntake.where("email_address LIKE ' %'").count
=> 1
```

I can strip those email addresses along with this change going in

mrotondo:
Leading & trailing spaces in existing email addresses on `StateFileArchivedIntake` have all been stripped on prod as of now.

PR #5952, opened 2025-04-23
[FYST-2023] Refactor filing_status (state vs. federal) and audit MD

arinchoi03:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-2023

## What was done?
- Using `state_filing_status` for calculators/xml/pdf
  - For all states, `state_filing_status` is same as `filing_status` (federal)
    - MD: if `direct_file_data.claimed_as_dependent`, the state filing status changes to `:dependent`
      - the change is more benign -- the "dependent" status was only in the calculator anyways. BUT running into some situations where we use filing status for filling out things in xml/pdf and wondering if we need to rely on federal filing for those...it's confusing 🙃 
    - AZ: if filing_status is `:qualifying_widow`, we consider their status `:head_of_household` 
      - bit more dangerous b/c AZ flow had basically operated with `filing_status` that is overwritten (unless `direct_file.filing_status` was directly called instead). it's actually "safer" to use `state_filing_status` since it's exactly what `filing_status` used to be. But now I need to think through on the flip side of the coin if this will have adverse effects where we use `intake.filing_status`
- do we want to use state_filing_status for controller logic?  
## How to test?
  - Test data used
- Risk Assessment
- Before
- After

### DIFF
--- a/app/lib/efile/md/md502_calculator.rb
+++ b/app/lib/efile/md/md502_calculator.rb
@@ -74,7 +74,7 @@ module Efile
         @md502r.calculate # depends on 502SU
         set_line(:MD502_LINE_10A, :calculate_line_10a) # depends on 502R
         set_line(:MD502_LINE_13, :calculate_line_13)
-        if filing_status_mfj?
+        if state_filing_status_mfj?
           @two_income_subtraction_worksheet.calculate
         end
         set_line(:MD502_LINE_14, :calculate_line_14)
@@ -155,19 +155,19 @@ module Efile
       end
 
       def calculate_line_a_spouse
-        filing_status_mfj? ? "X" : nil
+        state_filing_status_mfj? ? "X" : nil
       end
 
       def calculate_exemption_amount
         # Exemption amount
-        income_ranges = if filing_status_single? || filing_status_mfs?
+        income_ranges = if state_filing_status_single? || state_filing_status_mfs?
                           [
                             [-Float::INFINITY..100_000, 3200],
                             [100_001..125_000, 1600],
                             [125_001..150_000, 800],
                             [150_001..Float::INFINITY, 0]
                           ]
-                        elsif filing_status_hoh? || filing_status_mfj? || filing_status_qw?
+                        elsif state_filing_status_hoh? || state_filing_status_mfj? || state_filing_status_qw?
                           [
                             [-Float::INFINITY..100_000, 3200],
                             [100_001..125_000, 3200],
@@ -198,7 +198,7 @@ module Efile
       end
 
       def calculate_line_b_spouse_senior
-        return nil unless filing_status_mfj? || filing_status_qw?
+        return nil unless state_filing_status_mfj? || state_filing_status_qw?
 
         @intake.spouse_senior? ? "X" : nil
       end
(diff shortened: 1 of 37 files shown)


app/lib/efile/tax_calculator.rb
```diff
@@ -58,28 +58,28 @@ def set_line(line_id, value_fn_or_data_source, field = nil)
...
 
-    def filing_status_qw?
+    def state_filing_status_qw?
       @filing_status == :qualifying_widow
     end
 
-    def md_filing_status_dependent?
-      @direct_file_data.claimed_as_dependent?
+    def state_filing_status_dependent?
+      @filing_status == :dependent
```
arinchoi03, line 82: just want to nuke all these methods and use `intake.state_filing_status_{filing_status}` directly instead of relying on the calculators...

app/models/state_file_az_intake.rb
```diff
@@ -185,11 +193,6 @@ def disqualified_from_excise_credit_fyst?
     all_filers_incarcerated || whole_credit_already_claimed || direct_file_data.claimed_as_dependent?
   end
 
-  def filing_status
-    return :head_of_household if direct_file_data&.filing_status == 5 # Treat qualifying_widow as hoh
-    super
-  end
```
arinchoi03, line 191: since we override the `filing_status` of an intake here for a `qualifying_widow` to become `head_of_household`, by leaving the `filing_status` as-is (same as what's in the `direct_file_data`) may be problematic & we need to ensure we update all downstream places where we reference `filing_status` (even controllers/forms)

app/lib/efile/md/md502_calculator.rb
```diff
@@ -155,19 +155,19 @@ def calculate_line_a_primary
...
-        income_ranges = if filing_status_single? || filing_status_mfs?
+        income_ranges = if state_filing_status_single? || state_filing_status_mfs?
                           [
                             [-Float::INFINITY..100_000, 3200],
                             [100_001..125_000, 1600],
                             [125_001..150_000, 800],
                             [150_001..Float::INFINITY, 0]
                           ]
-                        elsif filing_status_hoh? || filing_status_mfj? || filing_status_qw?
+                        elsif state_filing_status_hoh? || state_filing_status_mfj? || state_filing_status_qw?
```
arinchoi03, line 170: with this change, a filer could get into the `else` for being a `:dependent` state_filing_status... is that what we want?

app/lib/efile/md/md502_calculator.rb
```diff
@@ -74,7 +74,7 @@ def calculate
         @md502r.calculate # depends on 502SU
         set_line(:MD502_LINE_10A, :calculate_line_10a) # depends on 502R
         set_line(:MD502_LINE_13, :calculate_line_13)
-        if filing_status_mfj?
+        if state_filing_status_mfj?
```
arinchoi03, line 77: do we still want to do the worksheet if they're `:dependent` state_filing_status?

app/lib/efile/md/md502_calculator.rb
```diff
@@ -155,19 +155,19 @@ def calculate_line_a_primary
       end
 
       def calculate_line_a_spouse
-        filing_status_mfj? ? "X" : nil
+        state_filing_status_mfj? ? "X" : nil
```
arinchoi03, line 158: do we want to fill this out if they are mfj & claimed as dpenedent?

arinchoi03:
chatted with Erica, it seems that we can't just change all the `filing_status` to `state_filing_status` wholesale like this, each line _could_ have different tax laws/logic that depends on federal filing status vs state filing status. And given the recent events, it seems not worthwhile to do a big change like this & potentially disrupt smooth filing

PR #5956, opened 2025-04-24
GYR1-713 archive document getting clicked by mistake

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-713

## What was done?
- There are 2 different approaches in this PR, separated by commit
- The second one seems cleaner to me because I hate single-use css classes buuuut obviously comes with the risk of accidentally changing the look of other checkboxes. But they are hub checkboxes so that seems...lower risk? Maybe we can get around that with thorough acceptance testing
- Please weigh in on which approach you like better!
## How to test?
- Visual/manual testing seems fine for this as it's just styling
- Before
image

- After
image

### DIFF
--- a/app/assets/stylesheets/components/_form-elements.scss
+++ b/app/assets/stylesheets/components/_form-elements.scss
@@ -49,6 +49,10 @@
   font-weight: 400;
 }
 
+.checkbox-fit-content {
+  width: fit-content;
+}
+
 .text-input--small-label {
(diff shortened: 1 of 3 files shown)


arinchoi03 (approved):
i agree this approach looks better with some detailed notes for QA where these `hub_checkbox`es are

PR #5962, opened 2025-04-25
pin selenium webdriver browser version to 113 explicitly

mrotondo:


arinchoi03 (commented):
shouldn't we pin to 133, not 113?

mrotondo:
> shouldn't we pin to 133, not 113?

maybe! i don't know what we've been running on, so i just tried an old-ish version that seemed supported. but i didn't read the chromium bug reports you linked closely enough - i'll try 133 if it sounds like these issue are addressed there

powersurge360 (approved):
Approved! Merging imminent.

PR #5969, opened 2025-04-28
[GYR1-659] Do not install rack-mini-profiler in production

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-659

## What was done?
- not installing rack-mini-profiler in production, which would lead `Rack::MiniProfiler` to be not defined
- `application_controller.rb` has this code:
```
  before_action do
    if defined?(Rack::MiniProfiler) && current_user&.admin?
      Rack::MiniProfiler.authorize_request
    end
  end
```
and b/c `Rack::MiniProfiler` is not defined without the gem installation, this code should not run. 
- Other than that Rack::MiniProfiler is referred to in the initializer. I removed `Rails.env.production?` check there & excluding the rest of the set up unless its in an env with REDIS _and_ rack-mini-profiler
## How to test?
  - Since this impacts production specifically, we won't really be able to test it 
  - I tested this by treating `:development` like `:production` env in the `Gemfile` (you have to restart server after you make changes) and visiting statefile.localhost:3000 and not seeing the profiler result in the corner anymore.
- Risk Assessment
  - I don't think there's a risk, since we're just removing a gem??? but i tried to minimize potential errors on our side

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,9 +87,10 @@ group :demo, :development, :test do
 end
 
 group :demo, :development, :heroku, :staging, :production do
-  # for storing results of rack-mini-profiler
   gem 'redis'
+end
 
+group :demo, :development, :heroku, :staging do
   gem 'rack-mini-profiler'
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,7 @@ gem 'ordinalize_full'
 gem 'awesome_print'
 gem 'rack-attack'
 gem 'holidays'
+gem 'redis'
 
 # Use Flipper for feature flagging
 gem 'flipper'
@@ -84,10 +85,6 @@ gem 'flipper-ui'
 
 group :demo, :development, :test do
   gem 'factory_bot_rails' # added to demo for creating fake data
-end
-
-group :demo, :development, :heroku, :staging, :production do
-  gem 'redis'
 end
 
 group :demo, :development, :heroku, :staging do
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -87,9 +87,10 @@ group :demo, :development, :test do
 end
 
 group :demo, :development, :heroku, :staging, :production do
-  # for storing results of rack-mini-profiler
   gem 'redis'
+end
```
arinchoi03, line 91: redis is now used not only for minifiler, but for action cable cache

Gemfile
```diff
@@ -87,9 +87,10 @@ group :demo, :development, :test do
...
 
+group :demo, :development, :heroku, :staging do
```
arinchoi03, line 93: so not production and not test

Gemfile
```diff
@@ -87,9 +87,10 @@ group :demo, :development, :test do
 end
 
 group :demo, :development, :heroku, :staging, :production do
```
jnf, line 89: isn't this all our environments? do we need this grouping for redis or can it go into the 'regular' list of gems?

jnf (approved):
1 wee question, but not blocking.

Gemfile
arinchoi03: all except test...but yes, i guess it's fine if we install it for test env like we do with most of our gems since there's no good reason to exclude it 🤔

PR #5982, opened 2025-05-07
GYR1-729 Not Ready Status Bug

tahsinaislam:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-729

## Is PM acceptance required? 

## What was done?
This fixes a bug where the status “Not Ready” (internally called intake_in_progress) was missing from status dropdowns. That caused confusion because when a client is in the “Not Ready” status, on their profile it showed a different status like “Needs Doc Help”. This caused clients in this status not to be able finish their intake
This was fixed this by adding “Not Ready” back to the dropdown, but made sure it’s disabled, so it can’t be selected manually. This way the status is visible and the status is consistent across all pages but still protected from manual updates.

## How to test?
Client in intake_in_progress status:
- Should see "Not Ready" on All Clients page and in client profile
- Status dropdown (e.g., Take Action) should include "Not Ready" as a disabled option
Verified by updating the tax_return_status_helper_spec.rb spec with expectations for a disabled option

- Before
Screenshot 2025-05-07 at 6 04 32 PM

- After
Screenshot 2025-05-07 at 5 46 54 PM

### DIFF
--- a/app/helpers/tax_return_status_helper.rb
+++ b/app/helpers/tax_return_status_helper.rb
@@ -2,8 +2,16 @@ module TaxReturnStatusHelper
   def grouped_status_options_for_select(unwanted_statuses: [])
     TaxReturnStateMachine.states_to_show_for_client_filter(role_type: current_user.role_type).map do |stage, statuses|
       translated_stage = TaxReturnStatusHelper.stage_translation(stage)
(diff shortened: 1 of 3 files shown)


jnf (approved):
clever approach. i like it.

PR #5989, opened 2025-05-19
Remove '16, '22, '23 needs help columns and add backfill rake task for 2022 intakes

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/TBE-152

## What was done?
- Remove `needs_help_2016`, `needs_help_2022` and `needs_help_2023` columns since they don't contain any real data
- backfill intakes from 2022 so their needs help data is represented by the dynamic `needs_help_current_year`, `needs_help_previous_year_1` etc fields
- once we run the backfill task on prod we can make another migration to remove the rest of the year specific needs_help columns
- update Mix panel service to send relevant data, I checked with Max to make sure the change wouldn't break any of his mixpanel data
## How to test?
- `needs_help_data_clean_up_spec.rb`
- checked the prod database to make sure `needs_help_2016`, `needs_help_2022` and `needs_help_2023` didn't have data and that 2022 product intakes are the only ones that need to be backfilled


app/services/mixpanel_service.rb
```diff
@@ -309,12 +309,10 @@ def data_from_gyr_intake(intake)
...
-        needs_help_2022: intake.needs_help_2022,
-        needs_help_2021: intake.needs_help_2021,
-        needs_help_2020: intake.needs_help_2020,
-        needs_help_2019: intake.needs_help_2019,
-        needs_help_2018: intake.needs_help_2018,
+        needs_help_current_year: intake.needs_help_current_year,
+        needs_help_previous_year_1: intake.needs_help_previous_year_1,
+        needs_help_previous_year_2: intake.needs_help_previous_year_2,
+        needs_help_previous_year_3: intake.needs_help_previous_year_3,
         needs_help_backtaxes: intake.needs_help_with_backtaxes? ? "yes" : "no",
```
arinchoi03, line 316: just curious -- was max (or anyone) using this mixpanel `needs_help` data?

arinchoi03 (commented):
lgtm! is there a followup story to drop the 2018-2021 columns?

app/models/intake.rb
```diff
@@ -619,7 +616,12 @@ def had_dependents_under?(yrs)
   end
 
   def needs_help_with_backtaxes?
-    MultiTenantService.new(:gyr).backtax_years.any? { |year| send("needs_help_#{year}_yes?") }
+    if (product_year == 2022)
+      # TODO: eventually we will remove this when we remove the needs_help_2021, needs_help_2020, needs_help_2019, needs_help_2018 fields
+      MultiTenantService.new(:gyr).backtax_years.any? { |year| send("needs_help_#{year}_yes?") }
+    else
+      needs_help_previous_year_1_yes? || needs_help_previous_year_2_yes? || needs_help_previous_year_3_yes?
+    end
```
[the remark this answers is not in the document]
arinchoi03: are there any tests for this method?

lib/tasks/needs_help_data_clean_up.rake
```diff
@@ -0,0 +1,29 @@
...
+
+        intake.update!(updated_attrs)
+      rescue => e
+        Rails.logger.error("----*Failed to update intake #{intake.id}: #{e.message}*----")
+      end
+    end
+
+    puts "*****Done!*****"
+  end
+end
```
arinchoi03, line 29: why this instead of a `data_migration`?

app/services/mixpanel_service.rb
embarnard: I asked Max if it was okay to make this change, I didn't ask specifically if he was using the needs_help data but might be a good idea to mention it to him that this data wasn't here before and the other needs_help columns were basically trash. I doubt anyone would have been using it without it raising red flag as it would have just returned unfilled for every intake

app/models/intake.rb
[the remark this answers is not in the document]
embarnard: Just added a test! I caught it in the first place because it was causing a test on the mixpanel service spec to fail but I think its good to have an explicit test as well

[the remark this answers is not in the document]
embarnard: I also changed the method, since we are not calling this method for intakes in the previous product years just for sending mixpanel data about the intake during the flow so we don't need to cover the case for 2022 intakes

lib/tasks/needs_help_data_clean_up.rake
embarnard: no real reason, it just feels easier to delete this after running rather than the migration and based off a conversation in comp garden a few months ago it didn't seem like people had a preference for rake tasks vs data migrations. Would you prefer data migration?

PR #5993, opened 2025-05-23
[FYST-2089] Show correct es return status links

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-2089

## What was done?
- Removed the escape characters (`\` ) from the yml string. b/c the `"` was being escaped, `"%{return_status_link}"` (i.e. `"https://demo.fileyourstatetaxes.org/es/questions/return-status"`) `""` was seen as part of a url string (instead of being just necessary for interpolating a variable) and the default behavior for a random string without hostname is to concatenate it to the current host url + append
## How to test?
  - No tests - maybe there could be something written to check if `\` are being used in yml string, but sometimes we _do_ need them, so need to figure out when to check.
  - To reproduce: follow steps in Jira ticket & hover over the links, they should all look as expected
- Before 

rejected:
Screenshot 2025-05-23 at 9 02 25 AM

accepted_refund:
Screenshot 2025-05-23 at 9 02 39 AM

accepted_owed:
Screenshot 2025-05-23 at 9 02 45 AM

- After (note the bottom left corner previewing the link I hovered over)

rejected:
Screenshot 2025-05-23 at 9 00 33 AM

accepted_refund:
Screenshot 2025-05-23 at 9 00 18 AM

accepted_owed:
Screenshot 2025-05-23 at 9 00 00 AM

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1522,7 +1522,7 @@ es:
 
               Si aún no has pagado mediante débito directo al presentar tu declaración, por favor haz tu pago lo antes posible para evitar multas y costos adicionales en %{state_pay_taxes_link}.
 
-              Descarga tu declaración en <a href=\"%{return_status_link}\" target=\"_blank\" rel=\"noopener nofollow\">%{return_status_link}</a>.
+              Descarga tu declaración en <a href="%{return_status_link}" target="_blank" rel="noopener nofollow">%{return_status_link}</a>.
 
               ¿Necesitas ayuda? Responde a este correo electrónico.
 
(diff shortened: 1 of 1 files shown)


config/locales/es.yml
```diff
@@ -1522,7 +1522,7 @@ es:
 
               Si aún no has pagado mediante débito directo al presentar tu declaración, por favor haz tu pago lo antes posible para evitar multas y costos adicionales en %{state_pay_taxes_link}.
 
-              Descarga tu declaración en <a href=\"%{return_status_link}\" target=\"_blank\" rel=\"noopener nofollow\">%{return_status_link}</a>.
```
arinchoi03, line 1525: b/c the `"` was being escaped, `"%{return_status_link}"` was seen as part of a url string and the default behavior for a random string without hostname is to concatenate it to the current host url + append?

PR #6002, opened 2025-06-05
GYR1-721 Remove selfie requirement from GYR flow

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/GYR1-721

## What was done?
- Deleted controllers and views for selfies and selfie instructions
- Removed selfie from docs overview page (client-facing)
- Returning `false` from `relevant_to?` in Selfie doc type class for now. Not sure if we'll eventually delete the whole class or not, but since there will still be clients with Selfie docs I figured we should keep it.
- Also returning `true` from `skip_selfies?` in experiment file
  - ~~Honestly might have been able to do a better job with this file? Not sure if the experiment is even under way anymore so maybe we can delete it but maybe that is better as a follow-up task to reduce complexity~~ Asked Erica and it looks like the whole ID experiment is not being run anymore and we don't have plans to run it again in the near future. I think we should remove all of the code relating to this experiment but it might be better to do that in a follow-up commit in order to keep this review simple and expedite the story. Open to removing it now if someone thinks of a good reason!
- Removed selfie from required doc types
- Removed every feature spec section that expects to see the selfie pages or to upload a selfie doc from the portal
- Updated various specs to no longer expect selfie to be required/uploaded/etc
## How to test?
- Go through the flow and see that the pages don't show up and that selfies are no longer listed on the documents overview 
- In the hub, make sure the % document calculation looks right
- Before
Screenshot 2025-06-05 at 10 42 06 AM
Screenshot 2025-06-05 at 10 51 54 AM
Screenshot 2025-06-05 at 11 04 57 AM
- After
Screenshot 2025-06-05 at 10 43 13 AM
Screenshot 2025-06-05 at 10 51 39 AM
Screenshot 2025-06-05 at 10 58 33 AM


app/controllers/documents/selfies_controller.rb
```diff
@@ -1,19 +0,0 @@
...
-
-    def self.document_type
-      DocumentTypes::Selfie
-    end
-
-    def after_update_success
-      current_intake.tax_returns.each do |tax_return|
-        tax_return.advance_to(:intake_needs_doc_help)
-      end
-    end
```
arinchoi03, line 13: what is the significance of this `after_update_success`? Do we still need to advance to `intake_needs_doc_help` in the previous controller now?

arinchoi03 (commented):
`.selfie-instructions-list` scss can probably be removed as well?
app/assets/stylesheets/components/_instructions-list.scss

also I'm not super familiar with this code so I'm not sure if im the best person to review it -- i have one question on the `selfies_controller`

app/controllers/documents/selfies_controller.rb
jenny-heath: it looks like the [previous controller](https://github.com/codeforamerica/vita-min/blob/main/app/controllers/documents/ids_controller.rb) is already advancing to this status, as is the [next controller](https://github.com/codeforamerica/vita-min/blob/main/app/controllers/documents/ssn_itins_controller.rb)

anisharamnani (approved):
⭐ bye bye bye selfies! i have the tiniest comment, but considering that this is mostly deleting code, i don’t have any feedback.

PR #6008, opened 2025-06-12
Add pundit, application, efile-error and user policy

embarnard:
## Link to pivotal/JIRA issue
- e.g. [https://codeforamerica.atlassian.net/browse/GYR1-797](https://codeforamerica.atlassian.net/browse/GYR1-797)

## What was done?
- Add pundit gem and pundit policies for application, efile-error and user
## How to test?
- Test that user and efile error objects [still work as expected](https://www.notion.so/cfa/Hub-User-Permissions-208373fd79b2809088a9e0ad43f27428?source=copy_link#209373fd79b280669a34d53d28a66191)


app/policies/user_policy.rb
```diff
@@ -0,0 +1,77 @@
+class UserPolicy < ApplicationPolicy
```
embarnard, line 1: converted these CanCan rules:

# # Anyone can manage their name & email address (roles are handled separately)
# can :manage, User, id: user.id
#
# # Anyone can read info about users that they can access
# can :read, User, id: user.accessible_users.pluck(:id)
#
# Organization leads can view and edit users who are organization leads, site coordinators, and team members in their coalition
# can :manage, User, id: user.accessible_users.pluck(:id) if user.org_lead?
#
# if user.site_coordinator?
# can [:suspend, :unsuspend, :update, :unlock, :resend_invitation],
# User, id: User.where(role: SiteCoordinatorRole.assignable_to_sites(user.role.sites)).pluck(:id)
# or id: User.where(role: TeamMemberRole.assignable_to_sites(user.role.sites)).pluck(:id)

powersurge360 (changes requested):
The double negative stuff is the only stuff I'm marking 'Request Changes' for. I'd like to talk some bitwise stuff and the other comments are at your discretion.

app/policies/user_policy.rb
powersurge360:
```ruby
module Perms
  SHOW = 0b0001
  CREATE = 0b0010
  UPDATE = 0b0100
  DESTROY = 0b1000
  MANAGE = SHOW | CREATE | UPDATE | DESTROY
end

def can?(*perms, role:)
  !perms.reduce(:|).nobits?(perms_for_role(role))
end

def perms_for_role(role)
  case role
  when :admin
    Perms::CREATE | Perms::UPDATE | Perms::DESTROY | Perms::SHOW
  when :greeter
    Perms::SHOW | Perms::UPDATE 
  end
end

IRB.start
```
Here is a small sample script to play with the concepts above.

I talked with em about this and demonstrated the script. Bitwise operations are somewhat arcane and maybe not ergonomic for our needs (notably, it is difficult to deal with binary representations of numbers in ruby). I think my anxiety is rooted in the ambiguity of combining or subclassing permissions. I think the takeaway here is I think that either:

1. We should flatly disallow permissions that inform other permissions. For example, we should not have a `can_manage?` that overrides a `can_update?`. *or*
1. We should strictly formalize what composable/inherited permissions look like. This way it is harder to overlook when a permission implies another and policies will have a similar architecture for different resources/controllers.

PR #6009, opened 2025-06-12
update our state api credentials to the newly-generated keys and certs, replacing the old ones which will expire mid-season

mrotondo:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/FYST-1966

I've verified that the demo certs work locally via twingate, but we should manually go through the import flow on demo once this is merged just to make sure. Once that's done, we unfortunately have no way to test the prod certs other than to release them and monitor for errors, and roll this back if we see any.


arinchoi03:
confirmed that `cert_base64` and `private_key_base64` values for each states seem to be updated with new values for dev/demo/heroku/staging/production

arinchoi03 (approved):

PR #6017, opened 2025-07-03
Add TY-24 state file intake archiver service

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-2143?atlOrigin=eyJpIjoiYjkzZWI3N2FjYjIyNDczNTgwZDk2OGIwOWYxNDJjNzEiLCJwIjoiaiJ9

## What was done?
- Add new archiver service for archiving the state file intakes from the 2024 tax year
- uploaded new az_addresses.csv, id_addresses.csv, nj_addresses.csv, md_addresses.csv and nc_addresses.csv into vita-min-prod-docs in S3 so that when these new intakes are created they will have new batch of addresses to pick from, different from last years
## How to test?
1. create accepted intakes for each state with unique contact methods
2. transition them to accepted
3. run `bundle exec rake state_file:ty24:archive_#{state_code}`
4. check for the existence of StateFileArchivedIntakes with the correctly copied over data


lib/tasks/state_file_archive_intakes.rake
```diff
@@ -1,27 +1,38 @@
...
-      archiver = StateFile::Ty23ArchiverService.new(state_code: 'az', batch_size: batch_size)
-      archiver.find_archiveables # sets `current_batch` on the archiver instance
-      while archiver.current_batch.count.positive? # keep archiving, in batches, until the archiver doesn't find anything else
-        archiver.archive_batch # process the batch
-        archiver.find_archiveables # set the next batch
-      end
+      # we batch these since archiving involves copying the submission pdf to a new location in s3
+      StateFile::Ty24ArchiverService.archive!(
+        state_code: args[:state_code],
+        batch_size: 50
```
embarnard, line 9: last year the batch number was 10, it seemed a little stingy to me so I increased it but happy to go down again

app/services/state_file/ty24_archiver_service.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+      Rails.logger.info("*****Completed archive for state: #{state_code}*****")
+    end
+
+    def archive_all
+      Rails.logger.info("*****Archiving for state: #{state_code}, tax_year: #{tax_year}, batch_size: #{batch_size}*****")
+
+      et = Time.find_zone('America/New_York')
+      start_date = et.parse('2025-01-15 00:00:00') # state_file_start_of_open_intake
+      end_date = et.parse('2025-10-25 23:59:59') # state_file_end_of_in_progress_intakes
```
anisharamnani, line 43: how many times will this be run? i’m wondering if these dates might need to be changed?

anisharamnani: ah, because of the query for the already archived intakes. this may need not change.

app/services/state_file/ty24_archiver_service.rb
```diff
@@ -0,0 +1,111 @@
...
+                                                         created_at: start_date..end_date,
+                                                         to_state: :accepted,
+                                                         most_recent: true
+                                                       }
+                                                     ).pluck('efile_submissions.data_source_id')
+
+      # remove intakes that have already been archived
+      archived_emails = StateFileArchivedIntake.where(state_code: state_code, tax_year: tax_year)
+                                               .pluck(:email_address)
+      archived_phones = StateFileArchivedIntake.where(state_code: state_code, tax_year: tax_year)
```
anisharamnani, line 59: why query by the email and phone as opposed to the hashed ssn?

anisharamnani: i guess they are both unique so it doesn’t really matter but i was just curious

app/services/state_file/ty24_archiver_service.rb
```diff
@@ -0,0 +1,111 @@
...
+        contact_preference: intake.contact_preference,
+        mailing_street: intake.direct_file_data.mailing_street,
+        mailing_apartment: intake.direct_file_data.mailing_apartment,
+        mailing_city: intake.direct_file_data.mailing_city,
+        mailing_state: intake.direct_file_data.mailing_state,
+        mailing_zip: intake.direct_file_data.mailing_zip,
+      )
+      archived.save!
+
+      if intake.submission_pdf.attached?
```
anisharamnani, line 98: why would we archive intakes where there is no submission pdf?

🤔 feels like a weird state to be in considering the query checks if the submission has been accepted.

could we just query to guarantee that the submission PDF  is present?

app/services/state_file/ty24_archiver_service.rb
embarnard: Since this is unique identifier that they use to sign in with, they either pick email or phone number. Last year they only did this check on emails since we didn't have sms working in the previous season before that. But actually I was thinking about it more and I don't know why we didn't just add a key reference for that state file intake like `data_source_id` instead wdyt

app/services/state_file/ty24_archiver_service.rb
embarnard: Okay so the plan would have been to note all the ones that didn't have pdfs via the rails logger warning and then investigate why there wasn't a submission pdf

mpidcock: yeah, I think that's ok, we were in a rush to delete the old records before so I was prioritizing archiving anything we might need so that the og records could be safely cleared, but we don't have that concern this year, so skipping + logging should be fine!

anisharamnani (approved):
⭐

anisharamnani:
great work em! thank you ❤️

PR #6023, opened 2025-07-15
Add mailgun creds for production (GYR, FYST, CTC) and heroku

embarnard:
## Link to pivotal/JIRA issue
- e.g. last PR for this set https://codeforamerica.atlassian.net/browse/TBE-107?atlOrigin=eyJpIjoiZmEwMTlkMWNiZTg2NDYwMzk4YjUxZTg1MDM3YmI0NGEiLCJwIjoiaiJ9

## What was done?
- Rotated mailgun creds for production GYR, FYST and CTC (although it seems like CTC creds could be removed as a part of that remove CTC messaging story later in this sprint)
- Rotated mailgun creds for heroku; I think sending will work for heroku but the webhook technically won't work unless someone puts in their url with the basic auth password under mailgun domains, but added it in anyways in case someone needs to do this in the future
## How to test?
- `bin/gyr credentials_diff --base main --environment production` and `bin/gyr credentials_diff --base main --environment heroku`
- can check webhook URLs in mailgun UI under "Send → Sending → Webhooks"
- these are the steps I took to change this stuff: https://www.notion.so/cfa/Updating-Mailgun-API-Credentials-22a373fd79b28012aea5c0e4397947d2


embarnard:
> Is there a chore/task to remove the extraneous keys in each of the domains? I see 3 keys for each domain.

Yes! I'm going to do it as a part of this task once I do the deploy and confirm its working with the new keys

embarnard:
> could you explain why ctc and statefile do not require `api_id` (like `ctc_api_id` or `statefile_api_id`):
> 
> ```
>   api_key: <key>
>   api_id: <id>
>   domain: <domain>
>   ctc_api_key: <ctc_api_key>
>   ctc_domain: <ctc_domain>
>   statefile_api_key: <state_file_key>
>   statefile_domain: <state_file_domain>
> ```

Yeah that's a good point! I think originally I added it for the demo one particularly since it seemed like the demo domain was shared between multiple environments so it was important to differentiate between the sending keys to make sure we didn't delete the wrong ones. But I think you're right we should add it for these other ones to be comprehensive

PR #6029, opened 2025-07-23
Send email notifications to assigned users for incoming client messages

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/GYR1-739

## What was done?
- notify the Hub user(s) assigned to that client via email when a client sends a message through SMS, email, or the client portal
## How to test?
**there is no heroku deploy and since mailgun doesn't work in heroku I didn't bother to fix it for this PR but it should be on staging 
- Attach your email or phone number to a client and send a message to GYR
- make sure the `hub_email_notifications` flipper flag is turned on
- attach another email to a user that is assigned to that client
- see if you get an email from that client (locally in the rails s log or on [staging](https://staging.getyourrefund.org/) which has mailgun configured to get the actual email)
view on `/en/hub/automated_messages`
Screenshot 2025-07-24 at 3 50 56 PM
`/hub/flipper`
Screenshot 2025-07-28 at 12 32 57 PM
view in email
Screenshot 2025-07-28 at 12 39 25 PM


app/views/layouts/mailer.html.erb
```diff
@@ -35,8 +35,8 @@
           width: 80%;
       }
       .email-contents button a {
-        text-decoration: none;
-        color: #FFFFFF;
+          text-decoration: none;
```
anisharamnani, line 38: i believe these are duplications.

embarnard: Oh oops! thanks for pointing that out, must have messed up when i pulled in y'alls button stuff

embarnard: Actually its just adding some spacing so I'm not gonna remove it if thats okay since it matches the spacing of the above block

app/models/user.rb
```diff
@@ -81,6 +82,8 @@ class User < ApplicationRecord
   scope :active, -> { where(suspended_at: nil) }
   scope :suspended, -> { where.not(suspended_at: nil) }
 
+  enum email_notification: { unfilled: 0, yes: 1, no: 2}, _prefix: :email_notification
```
arinchoi03, line 85: since the default is `1`, do we ever get into a state where we will have an `unfilled` value? If so, should we just have `yes` or `no`?

spec/services/interaction_tracking_service_spec.rb
```diff
@@ -11,4 +11,143 @@
...
+        described_class.record_incoming_interaction(client, set_flag: true, message_received_at: fake_time)
+        expect(InternalEmail).not_to have_received(:create!)
+        expect(SendInternalEmailJob).not_to have_received(:perform_later)
+      end
+    end
+
+    context "when only one tax return has an assigned user" do
+      before do
+        tax_return_2.update!(assigned_user: nil)
+      end
```
arinchoi03, line 79: can you explain what this case is testing?
the tax return 2 no longer has an assigned user, but then we check if we have created InternalEmail for the other tax return?

spec/services/interaction_tracking_service_spec.rb
```diff
@@ -11,4 +11,143 @@
...
+      it "touches only last_incoming_interaction_at and first_unanswered_incoming_interaction_at, but leaves flagged_at nil" do
+        Timecop.freeze(fake_time) do
+          InteractionTrackingService.record_incoming_interaction(client, set_flag: false, message_received_at: fake_time)
+          client.reload
+
+          expect(client.last_incoming_interaction_at).to eq fake_time
+          expect(client.first_unanswered_incoming_interaction_at).to eq fake_time
+          expect(client.flagged_at).to be_nil
+        end
+      end
```
arinchoi03, line 150: i appreciate you beefing up the test suite here!

app/mailers/user_mailer.rb
```diff
@@ -33,4 +32,28 @@ def assignment_email(
 
     mail(to: @assigned_user.email, subject: @subject)
   end
+
+  def incoming_interaction_notification_email(client:, user:, message_received_at:)
+    @client_id = client.id
```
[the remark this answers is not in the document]
embarnard: we really shouldn't send a message if the client isn't present and I don't think we have to be worried about it not being called in this case since the `record_incoming_interaction` method that calls it won't run unless we can find tax returns off this client
Screenshot 2025-07-28 at 3 31 00 PM

spec/services/interaction_tracking_service_spec.rb
embarnard: Ooo yeah you're right it's not clear, I guees I could add an expect(InternalEmail).not_to have_received(:create!) with the user_no_notifications argument? I guess I wanted to test this line `users_to_contact` but not sure about this best way to do that

anisharamnani (approved):
thanks em! i have read through, i don’t have comments like arin does, but i approve!

arinchoi03 (approved):

PR #6030, opened 2025-07-24
[TBE-243] No longer reference ctc_root_path from the maintenance page

arinchoi03:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-243

## What was done?
- Noticed during audit hour that [we saw increased 500 errors (from `hund.io` source param)](https://www.notion.so/cfa/07-21-2025-237373fd79b28079af15df564087d030?source=copy_link#237373fd79b28018889cf661de71d387)
- Looked into datadog logs (and the surrounding context) and saw that the requests to `/maintenance` page was leading to the 500 error
- Fixed the maintenance page to not refer to `ctc_root_path` which was breaking the page, and instead use get your refund `root_path` as the default. Now the "at capacity" message (for the `maintenance` page) will be branded for GYR by default.
## How to test?
- Added a test to make sure the page can load
- Manually navigate to `/maintenance` page
- Before
Screenshot 2025-07-23 at 11 26 02 PM

- After
Screenshot 2025-07-23 at 11 44 57 PM

### DIFF
--- a/app/views/public_pages/maintenance.html.erb
+++ b/app/views/public_pages/maintenance.html.erb
@@ -1,6 +1,6 @@
 <% content_for :page_title, t("views.ctc.questions.at_capacity.title") %>
 <% content_for :header do %>
-  <%= render 'shared/header', variation: "ctc", home_link: ctc_root_path %>
+  <%= render 'shared/header', variation: "gyr", home_link: root_path %>
 <% end %>
 <% content_for :footer do %>
   <footer class="footer">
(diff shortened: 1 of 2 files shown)


app/views/public_pages/maintenance.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :page_title, t("views.ctc.questions.at_capacity.title") %>
 <% content_for :header do %>
-  <%= render 'shared/header', variation: "ctc", home_link: ctc_root_path %>
```
arinchoi03, line 3: this was the cause of the error (500)

app/views/public_pages/maintenance.html.erb
```diff
@@ -12,7 +12,7 @@
           </p>
           <p>
             <% cfa_link = link_to("Code for America", "https://www.codeforamerica.org", target: "_blank", rel: "noopener", class: "footer__link") %>
-            <%= t("views.shared.footer.ctc_service_description_html", link: cfa_link) %>
```
arinchoi03, line 15: making the footer match the header

app/views/public_pages/maintenance.html.erb
embarnard: Ooof nice good catch

embarnard (approved):
Nice! thanks for adding that test in too

PR #6034, opened 2025-08-07
Add batching for client interaction notification emails

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/GYR1-793

## What was done?
- Adds batching for user notification emails which means if a client sends a message within the 10 minute window then we should only send one message to the user notifying of the message
## How to test?
- Test on staging
- make sure flipper is on and you have not responded to the message on the user side 
- Send a message as a client and then send another one within the 10 minute window and one outside of the ten minute window => confirm that you got the first message 10 minutes after the second message and one message 10 minutes after the third
The only visual change is now the client message notification is pluralized so that we note the number of messages sent within that 10 minute window
- Before
Screenshot 2025-08-12 at 11 45 57 AM

- After (added message count)
Screenshot 2025-08-12 at 11 45 02 AM

### DIFF
--- a/app/services/interaction_tracking_service.rb
+++ b/app/services/interaction_tracking_service.rb
@@ -13,7 +13,6 @@ class InteractionTrackingService
       users_to_contact = User.where(id: users_to_contact, email_notification: "yes")
       unless users_to_contact.empty?
         users_to_contact.each do |user|
-          next unless user.email_notification == "yes"
           interaction = ClientInteraction.create!(client: client, interaction_type: interaction_type)
           ClientInteractionNotificationEmailJob.set(wait: 10.minutes).perform_later(interaction, user, received_at: attrs[:received_at])
         end
--- a/spec/jobs/state_file/client_interaction_notification_email_job_spec.rb
+++ b/spec/jobs/state_file/client_interaction_notification_email_job_spec.rb
@@ -10,9 +10,11 @@ RSpec.describe ClientInteractionNotificationEmailJob, type: :job do
     let(:user) { create :user }
     let!(:interaction) { create(:client_interaction, client: client, interaction_type: "client_message", created_at: fake_time) }
 
+
     before do
-      allow(UserMailer).to receive(:assignment_email).and_return(mailer)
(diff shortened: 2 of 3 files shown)


app/models/client_interaction.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+#  index_client_interactions_on_client_id  (client_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (client_id => clients.id)
+#
+class ClientInteraction < ApplicationRecord
+  belongs_to :client
+  enum interaction_type: { unfilled: 0, document_upload: 1, client_message: 2 }, _prefix: :interaction_type
```
powersurge360, line 21: Learned from Mike a while ago that `prefix: true` is equivalent if you'd like to avoid the repetition.

app/jobs/client_interaction_notification_email_job.rb
```diff
@@ -0,0 +1,40 @@
+class ClientInteractionNotificationEmailJob < ApplicationJob
+  retry_on Mailgun::CommunicationError
+
+  def perform(interaction, user, **attrs)
```
powersurge360, line 4: The only `attrs` that seems to be acknowledged is `:received_at`. Why not be explicit about it?

app/services/interaction_tracking_service.rb
```diff
@@ -4,28 +4,22 @@ def self.update_last_outgoing_communication_at(client)
     client.touch(:last_outgoing_communication_at)
   end
 
-  # When a client contacts us, update last incoming & last interaction and send a email notification to assigned users
-  def self.record_incoming_interaction(client, set_flag: true, message_received_at: nil)
-    if message_received_at.present? && Flipper.enabled?(:hub_email_notifications)
+  def self.record_incoming_interaction(client, set_flag: true, **attrs)
```
powersurge360, line 7: Similar to above, there seems to only be two `attrs` that get used. Why not be explicit?

powersurge360 (approved):
Seems good. Just had a couple of nits.

PR #6039, opened 2025-08-13
[FYST-2211] Guard against copying in S3 if no key exists for the given intake

arinchoi03:
https://codeforamerica.atlassian.net/browse/FYST-2211 continued

Saw errors like:

```
was not able to copy 358 blob  due to error: missing required parameter params[:key]
```

when we ran this on demo, b/c we were trying to copy objects without giving it a key.


anisharamnani (approved):
we paired!

PR #6040, opened 2025-08-13
Add Puma::HttpParserError to the list of ignored errors in sentry initializer

jenny-heath:
## Link to pivotal/JIRA issue
- https://codeforamerica.atlassian.net/browse/TBE-222

## What was done?
- we upgraded puma so now we're getting a ton of distinct sentry issues (where they used to all be grouped under one issue). let's ignore all of them for now so they don't clog up the alerts channel and the sentry queue


arinchoi03 (approved):
I'm ok with ignoring it for now, but would love a story to figure out how we can stop these errors from happening in the first place (if we can?). I dont know if ignoring the errors is a good long-term strategy, and would like a story to address the root cause

PR #6081, opened 2025-10-03
Don't send FYST reminder notifications in demo

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/FYST-2293
- No (not sure what PM would test)

## What was done?
- Don't send State-file reminder notifications in demo
- Added tests for year and environment differences in rake task specs
- replaced all instances of `return` with `next` since I was getting a `LocalJumpError: unexpected return` error during testing which happens because a Rake task’s block is a proc, not a method/lambda and returns can’t exit from a proc.
## How to test?
- not sure the best way to test this, i think the tests suffice?

### DIFF
--- a/crontab
+++ b/crontab
@@ -12,7 +12,7 @@
 0 18 13,23 4 * bundle exec rake send_reject_resolution_reminder_notifications:send
 0 18 08,15 4 * bundle exec rake state_file:pre_deadline_reminder
 0 18 17,23 4 * bundle exec rake state_file:post_deadline_reminder
-0 21 27 10 * bundle exec rake users:suspend_non_admins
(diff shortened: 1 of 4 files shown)


crontab
```diff
@@ -12,7 +12,7 @@
 0 18 13,23 4 * bundle exec rake send_reject_resolution_reminder_notifications:send
 0 18 08,15 4 * bundle exec rake state_file:pre_deadline_reminder
 0 18 17,23 4 * bundle exec rake state_file:post_deadline_reminder
-0 21 27 10 * bundle exec rake users:suspend_non_admins
```
embarnard, line 15: just reordered these by date

PR #6264, opened 2026-04-09
GYR1-917 Handle password-protected PDFs gracefully

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-917

## Is PM acceptance required?

## What was done?

First, a quick summary of PDF encryption: a PDF can be 'encrypted' for a few 
reasons:

- password-protected
- printing disallowed
- copy disallowed (as in, copy-paste)
- commenting disallowed

There's a bit more to it, but that's the gist. We want PDFs that are
encrypted-but-not-password-protected to still be uploadable.

What was done:
- Enhanced the Document class's validation of PDF uploads to display a message 
  in the UI when a password-protected PDF is uploaded.
- Also enabled the upload validation to apply to all uploads (was previously 
  limited to `UnsignedForm8879` only, it seems?)
- The exception being caught is PDF::Reader's EncryptedPDFError. Made sure that 
  encrypted-but-not-password-protected PDFs were still accepted via a distinct 
  spec test for each scenario.
- The new PDF fixtures came from GCF :) 

## How to test?

Using the flow tester, try out the document upload screen using each of the two 
new fixture PDFs added in this PR. The password-protected one should show an 
informative message on the screen (like the screenshot below). The other should 
work O.K.

## Screenshots

upload-message


app/models/document.rb
```diff
@@ -53,7 +53,7 @@ class Document < ApplicationRecord
   validate :tax_return_present_sometimes
   validate :tax_return_absent_sometimes
   validate :upload_must_have_data
-  validate :upload_must_be_readable, if: -> { document_type == DocumentTypes::UnsignedForm8879.key }
+  validate :upload_must_be_readable
```
spompea-cfa, line 57: i didn't see any reason to restrict this to `UnsignedForm8879` only. but lmk if there's an historical reason for this that still applies; if so, i might need to rework the logic here a bit

embarnard (approved):

PR #6267, opened 2026-04-13
Implement image compression for large uploads and a 20-page limit for PDFs to prevent Bedrock/Anthropic payload errors

embarnard:
## Link to pivotal/JIRA issue
- e.g. https://codeforamerica.atlassian.net/browse/GYR1-982

## What was done?
- compress uploads that are larger than the 3.7 megabytes limit
- only take the first 20 pages of a pdf if it is longer than that to prevent the 20 image limit error


app/services/bedrock_doc_screener.rb
```diff
@@ -109,6 +115,23 @@ def self.screen_document!(document:)
...
+    end
+
+    result = image.to_blob
+
+    if result.bytesize > MAX_RAW_IMAGE_SIZE
+      image.resize "50%"
+      result = image.to_blob
+    end
+
+    result
```
spompea-cfa, line 132: down the line it might be useful to add some sort of retry logic for a case where the file is still over the max

spompea-cfa (approved):
looks great! just one comment below

PR #6268, opened 2026-04-14
Fix heroku deploys

embarnard:
Fix broken heroku deploys:

upgrade to 7.2 method `ActiveRecord::Base.connection.schema_migration` were removed


spompea-cfa (approved):
🥳

PR #6272, opened 2026-04-14
GYR1-966 SmartScan feature into Ability

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-966

## Is PM acceptance required?

## What was done?

Ensured that access to the SmartScan feature is managed through the Ability
class. This will make it easy to enable the feature for other groups of users as 
needed.

## How to test?

Functionality should be unchanged: as before, right now only admin users should 
be able to see SmartScan-related UI.

Testing involves uploading a document to a client's account in the Hub, and 
then:
- as an admin, checking that the SmartScan columns appear in the Documents page
- checking that the SmartScan card appears on the document's details page (in a 
  non-prod environment, this can be simulated via the Rerun Screener button; 
  there will be an error, but when you navigate back to the page, you'll see 
  that a card will now be present -- edit: correction, in this non-prod scenario,
  it's not the standard card that's displayed, but rather the assessment output in a table form, 
  which should suffice for testing purposes here).
- Then, logout and login as a non-admin user; the above UI pieces should now be 
  hidden.

## Screenshots

Admin:

a - Screenshot 2026-04-14 at 5 59 43 PM

Non-admin:

b - Screenshot 2026-04-14 at 6 01 23 PM

### DIFF
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -47,9 +47,11 @@ class Ability
         cannot :manage, :flipper_dashboard
       end
 
-      can :manage, :smartscan_doc_screener
+      can :read, DocAssessment
+      can :update, DocAssessment
       can :read, DocAssessmentFeedback
       can :create, DocAssessmentFeedback
+
       return
     end
 
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -46,6 +46,9 @@ class Ability
       unless user.email.downcase.include?("@codeforamerica.org")
         cannot :manage, :flipper_dashboard
       end
+
+      can :manage, :smartscan_doc_screener
+      can :manage, DocAssessmentFeedback
       return
     end
 
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -48,7 +48,8 @@ class Ability
       end
 
       can :manage, :smartscan_doc_screener
-      can :manage, DocAssessmentFeedback
+      can :read, DocAssessmentFeedback
+      can :create, DocAssessmentFeedback
       return
     end
 
--- a/app/views/hub/documents/_ai_screener.html.erb
+++ b/app/views/hub/documents/_ai_screener.html.erb
@@ -1,4 +1,4 @@
-<% if current_user.admin? && !@document.latest_assessment&.confirmed? %>
+<% if can? :manage, :smartscan_doc_screener && !@document.latest_assessment&.confirmed? %>
   <% verdict = @document.latest_assessment&.matches_doc_type_verdict %>
   <% unless verdict.blank? %>
     <%= render "hub/documents/smart_scan_card",
--- a/app/views/hub/documents/_ai_screener.html.erb
+++ b/app/views/hub/documents/_ai_screener.html.erb
@@ -1,4 +1,4 @@
-<% if can? :read, :DocAssessmentFeedback && !@document.latest_assessment&.confirmed? %>
+<% if can?(:read, :DocAssessmentFeedback) && !@document&.latest_assessment&.confirmed? %>
   <% verdict = @document.latest_assessment&.matches_doc_type_verdict %>
   <% unless verdict.blank? %>
     <%= render "hub/documents/smart_scan_card",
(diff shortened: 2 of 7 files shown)


app/lib/ability.rb
```diff
@@ -46,6 +46,9 @@ def initialize(user)
       unless user.email.downcase.include?("@codeforamerica.org")
         cannot :manage, :flipper_dashboard
       end
+
+      can :manage, :smartscan_doc_screener
+      can :manage, DocAssessmentFeedback
```
embarnard:
```suggestion
      can :read, DocAssessmentFeedback
      can :create, DocAssessmentFeedback
```

app/views/hub/documents/_ai_screener_dev_info.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if current_user.admin? && !acts_like_production? %>
+<% if can? :manage, :smartscan_doc_screener && !acts_like_production? %>
```
embarnard:
```suggestion
<% if can? :create, :DocAssessmentFeedback && !acts_like_production? %>
```

app/views/hub/documents/_ai_screener.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if current_user.admin? %>
+<% if can? :manage, :smartscan_doc_screener %>
```
embarnard:
```suggestion
<% if can? :read, :DocAssessmentFeedback %>
```

app/views/hub/documents/_feedback_collection.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if current_user.admin? && (@document.latest_assessment.present? && @document.latest_assessment&.feedbacks.blank?) %>
+<% if can? :manage, :smartscan_doc_screener && (@document.latest_assessment.present? && @document.latest_assessment&.feedbacks.blank?) %>
```
embarnard:
```suggestion
<% if can? :read, :DocAssessmentFeedback && (@document.latest_assessment.present? && @document.latest_assessment&.feedbacks.blank?) %>
```

app/views/hub/documents/index.html.erb
```diff
@@ -41,7 +41,7 @@
           <%= render "shared/column_sort_link", title: t("general.document_type"), column_name: "document_type" %>
         </th>
 
-        <% if current_user.admin? %>
+        <% if can? :manage, :smartscan_doc_screener %>
```
embarnard:
```suggestion
        <% if can? :read, :DocAssessmentFeedback %>
```

app/views/hub/documents/index.html.erb
```diff
@@ -59,7 +59,7 @@
         <th scope="col" class="index-table__header document-column__upload-date">
           <%= render "shared/column_sort_link", title: t(".upload_date"), column_name: "created_at" %>
         </th>
-        <% if current_user.admin? %>
+        <% if can? :manage, :smartscan_doc_screener %>
```
embarnard:
```suggestion
        <% if can? :read, :DocAssessmentFeedback %>
```

app/views/hub/documents/index.html.erb
```diff
@@ -84,7 +84,7 @@
               <%= id_prefix + document.document_type_label.to_s %>
             </td>
 
-            <% if current_user.admin? %>
+            <% if can? :manage, :smartscan_doc_screener %>
```
embarnard:
```suggestion
            <% if can? :read, :DocAssessmentFeedback %>
```

embarnard (commented):
This looks good! But I think we can scope it down more specifically to what we are doing to the objects also we should do something like load_and_authorize_resource :doc_assessment_feedback in the controller so we don't even allow loading the object if they don't have access on documents controller and then pass it to the view

app/views/hub/documents/index.html.erb
```diff
@@ -59,11 +59,6 @@
         <th scope="col" class="index-table__header document-column__upload-date">
           <%= render "shared/column_sort_link", title: t(".upload_date"), column_name: "created_at" %>
         </th>
-        <% if current_user.admin? %>
-          <th scope="col" class="index-table__header document-column__screener_verdict">
-            <%= render "shared/column_sort_link", title: "Matches doc-type", column_name: "screener_verdict" %>
-          </th>
-        <% end %>
```
spompea-cfa, line 66: ancillary tidying -- this column was removed under GYR1-872 and then the header (but not the row-by-row cells) reappeared in GYR-935; no need for it, re-removing

DrewProebstel (approved):
This looks great! I appreciate the clean up on the columnin the index page.

PR #6283, opened 2026-04-23
GYR1-1005 SmartScan UI updates

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-966

## Is PM acceptance required?

## What was done?

Style and placement updates were made to SmartScan-related UI.

## How to test?

Confirm:
- Documents page/tab: in the SmartScan column, the text "Warning" should be used 
  now instead of "Fail"
- In the Document Edit page, SmartScan card:
  - updated corner 'radius' to match design scheme of rest of app
  - no longer has a horizontal bar in it
  - is placed near the top (just under the title)

## Screenshots

pass

warn

### DIFF
--- a/app/views/hub/documents/_smart_scan_card.html.erb
+++ b/app/views/hub/documents/_smart_scan_card.html.erb
@@ -1,14 +1,16 @@
-<div class="smart-scan-card smart-scan-card--<%= state %>">
+<div class="eligibility-card smart-scan-card smart-scan-card--<%= state %>">
   <div class="smart-scan-card__header">
     <span class="smart-scan-card__icon" aria-hidden="true"></span>
     <strong><%= t("hub.documents.smart_scan_card.validoc_notes") %></strong>
   </div>
 
-  <div class="smart-scan-card__divider"></div>
+  &nbsp;<br>
 
   <div>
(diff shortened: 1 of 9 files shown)


app/views/hub/documents/_smart_scan_card.html.erb
```diff
@@ -1,15 +1,17 @@
...
     <span class="smart-scan-card__icon" aria-hidden="true"></span>
     <strong><%= t("hub.documents.smart_scan_card.validoc_notes") %></strong>
   </div>
 
-  <div class="smart-scan-card__divider"></div>
+  &nbsp;<br>
 
   <div>
-    <strong><%= state == "success" ? t("hub.documents.smart_scan_card.validation_passed") : t("hub.documents.smart_scan_card.validation_fail") %></strong>
+    <strong><%= state == "success" ? t("hub.documents.smart_scan_card.validation_passed") : 'Validation Warning' %></strong>
```
spompea-cfa, line 10: my understanding is the Hub does not need to support languages other than English, so my sense is to unwind these t() calls as we update the code. but if we'd prefer to continue using them, just lmk

embarnard (approved):

PR #6291, opened 2026-04-27
GYR1-1031 (1/2) Refactor AiScreenerMetricsService

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1031

## Is PM acceptance required?

## What was done?

As a preliminary step to adding caching to this class (later PR), I refactored the class:
- the public `call` method is now a class method
- some basic tidying for readability
- there was some unused functionality (the optional arg into `initialize`) to run the calculations for just a 
  specific doc type. this has been removed and, along with that, was able to 
  remove a chunk of spec tests as well. 

a couple other thoughts

- let's keep thinking about ways to remove/avoid unused functionality/code -- 
  it's never just the code itself, but also the additional spec tests that are 
  then needed
- in this particular case, running analysis on specific doc types is not a bad 
  idea, but i'd suggest e.g. metabase/etc for that kind fo analysis

### DIFF
--- a/app/controllers/hub/ai_screener_metrics_controller.rb
+++ b/app/controllers/hub/ai_screener_metrics_controller.rb
@@ -3,7 +3,7 @@ module Hub
     before_action :require_admin
 
     def show
-      @metrics = AiScreenerMetricsService.new.call
+      @metrics = AiScreenerMetricsService.call
     end
   end
 end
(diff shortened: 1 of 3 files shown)


DrewProebstel (approved):
looks good, thanks for cleaning this up.

PR #6296, opened 2026-05-01
SECURITY: dependency cooldowns

robdiciuccio:
## What was done?
- Upgrade `yarn` to latest stable version and implement minimum package age and disable package scripts
  - Add explicit exception for `percy` postinstall script
- Add minimum package age to dependabot config

**NOTE:** requires changes to developer environment (yarn upgrade)

Relevant docs: https://www.notion.so/cfa/Dependency-Management-Best-Practices-WIP-341373fd79b28043862ef8995b817f1d

## How to test?
- Does is build and run properly?


.github/dependabot.yml
```diff
@@ -11,10 +11,20 @@ updates:
     # Dependabot external code execution allows gems/gemspec files to possibly read dependabot secrets. The only secret is a
     # GitHub token. It has read-only permissions and is a machine user.
     # https://docs.github.com/en/developers/overview/managing-deploy-keys#machine-users
-    insecure-external-code-execution: allow
+    # insecure-external-code-execution: allow
```
robdiciuccio: There are several GitHub secrets in this repo. This should be disabled unless there is a specific use case for it.

package.json
```diff
@@ -2,16 +2,16 @@
...
   "dependencies": {
-    "@babel/core": "7",
-    "@babel/plugin-transform-runtime": "7",
-    "@babel/preset-env": "7",
-    "@babel/runtime": "7",
-    "@rails/actioncable": "7.2.3.1",
-    "@rails/actiontext": "7.2.3.1",
-    "@rails/activestorage": "7.2.3.1",
-    "@rails/ujs": "7.2.3.1",
+    "@babel/core": "^7",
```
robdiciuccio, line 5: These are all yarn v4 compatibility changes

package.json
```diff
@@ -72,12 +73,17 @@
     ]
   },
   "resolutions": {
-    "webpack-dev-server/compression/on-headers": "1.1.0",
-    "**/load-nyc-config/js-yaml": "^3.14.2",
```
robdiciuccio, line 76: Need more info here, if these are necessary.

.github/dependabot.yml
martha-brandt: I'm in favor of deleting this (and the comment above it) as long as we know dependabot can still open PRs for this repo

package.json
spompea-cfa: assuming this is some syntax sugar mostly for documentation purposes, and that this change is ok. (`npm ls` can also provide this info). cc @mrotondo

spompea-cfa (approved):

package.json
mrotondo: They were overrides to force updates on transitive dependencies. If dependabot doesn't complain when they're removed, it probably means that the direct dependencies have been updated to point at newer versions of these.

mrotondo: see https://github.com/codeforamerica/vita-min/pull/6130

PR #6297, opened 2026-05-01
GYR1-1024 Fix 'I Don't Know' button

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1024

## Is PM acceptance required?

## What was done?

Fixed a bug that was causing a NotNullViolation (resulting in a 500 error / 
"Oops" screen in prod) when users clicked on the "I Don't Know" button (Yes/No 
question screens).

E.g., in `had_dependents_form`, this bug could be observed like so (upon 
clicking "I Don't Know"):

    Inside of app/forms/had_dependents_form.rb ...
        4: def save
        5:   pry
     => 6:   @intake.update(attributes_for(:intake))
        7: end

    [1] pry(#<HadDependentsForm>)> attributes_for(:intake)
    => {had_dependents: nil}

## How to test?

Check any Yes/No page and make sure things work smoothly when "I Don't Know" is 
chosen.


app/views/layouts/yes_no_question.html.erb
```diff
@@ -44,7 +44,7 @@
                 </div>
                 <% if has_unsure_option? %>
                   <div class="i-dont-know">
-                    <%= f.button :submit, name: "#{f.object_name}[#{method_name}", value: "unsure", class: "button button--link", "aria-describedby": "main-question" do %>
+                    <%= f.button :submit, name: "#{f.object_name}[#{method_name}]", value: "unsure", class: "button button--link", "aria-describedby": "main-question" do %>
```
spompea-cfa, line 47: seems like just the one missing bracket char was the problem

embarnard (approved):

PR #6315, opened 2026-05-19
GYR1-1034 SmartScan feature for partners

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1034

## Is PM acceptance required?

## What was done?

- Added a new boolean field to VitaPartner called `show_smartscan_ui`
- Updated Ability class to use this field to allow particular sites to see the 
  SmartScan UI (when `show_smartscan_ui` is true)
- Added a new Admin Tools screen called SmartScan Admin that let's admins 
  enable/disable the SmartScan UI on a site-by-site basis
- Small change to seeder.rb

## How to test?

- As an admin, navigate to to the new screen, and select a site using the 
  checkbox; then click Save.
- Log out and then log in using a test user for that particular site (see 
  seeder.rb); go to the Hub Documents page ...
  - confirm you can see the SmartScan column on the main Docs page
  - Edit a particular Doc; make sure you see the appropriate elements (SmartScan 
    card) on the the page
  - Ensure you can successfully click the green checkmark / red x on an
     "Is the Smart Scan label correct?" widget. 

- After

New admin menu item named "SmartScan Admin":

image

New screen:

image


app/lib/seeder.rb
```diff
@@ -63,7 +63,6 @@ def run
     vp1 = first_org = VitaPartner.find_or_create_by!(
       name: "Oregano Org",
       allows_greeters: true,
-      coalition: koalas,
```
spompea-cfa, line 66: the combination of coaliaion and StateRoutingTarget on vp1 seemed to apparently be tripping a validation error as i was doing some testing - organization.rb - "Since the organization has states configured, it cannot also be part of a coalition."

app/controllers/hub/admin/smartscan_admin_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Hub
+  module Admin
+    class SmartscanAdminController < Hub::BaseController
```
embarnard: sorry this is soooo picky but could we name it SmartScanAdminController instead?
```suggestion
    class SmartScanAdminController < Hub::BaseController
```

app/lib/ability.rb
```diff
@@ -207,5 +208,13 @@ def initialize(user)
         user.role.sites.map.any? { |site| role.sites.map(&:id).include? site.id }
       end
     end
+
+    if accessible_groups.any?(&:show_smartscan_ui)
+      can :read, DocAssessment
```
embarnard:
```suggestion
      can [:read, :update, :create], DocAssessment
```

app/lib/ability.rb
```diff
@@ -207,5 +208,13 @@ def initialize(user)
...
+      can :update, DocAssessment
+      can :create, DocAssessment
+      can :read, DocAssessmentFeedback
```
embarnard:
```suggestion
      can [:read, :create], DocAssessmentFeedback
```

embarnard (approved):
Looks good! but maybe missing a test were we check which different types of users and what is accessible to them on the doc page

app/controllers/hub/admin/smartscan_admin_controller.rb
spompea-cfa: sure

app/lib/ability.rb
spompea-cfa: 👍

app/lib/ability.rb
spompea-cfa: 👍

app/lib/ability.rb
```diff
@@ -207,5 +208,11 @@ def initialize(user)
         user.role.sites.map.any? { |site| role.sites.map(&:id).include? site.id }
       end
     end
+
+    if accessible_groups.any?(&:show_smartscan_ui)
+      can [:read, :update, :create], DocAssessment
+      can [:read, :create], DocAssessmentFeedback
+      can [:record_feedback, :rerun_screener], Document
```
spompea-cfa, line 215: 🌟

app/views/hub/documents/_ai_screener.html.erb
```diff
@@ -1,12 +1,14 @@
...
     <% unless verdict.blank? %>
       <%= render "hub/documents/smart_scan_card",
                  document: @document,
                  state: verdict == "pass" ? "success" : "fail" %>
 
-      <%= render "hub/documents/feedback_collection" %>
+      <% if !@document.latest_assessment&.confirmed? %>
+        <%= render "hub/documents/feedback_collection" %>
+      <% end %>
     <% end %>
```
spompea-cfa, line 12: it seems like it would be better to check `confirmed?` down here, but if that's seems unnecessary or not-quite-right, lmk

app/lib/ability.rb
spompea-cfa: just to add some notes here ... one piece of functionality that i discovered wasn't quite working when the pr was initially put up (but was only revealed when additional spec tests were being added) is: on the documents edit screen, the smart scan-related UI was visible, but when upon submitting feedback, the submit failed, and a 'not allowed' screen was presented. i spent a lot of time reviewing the CanCanCan docs and experimenting with various ability-related directives inside of DocumentsController, without success. finally, based on intuition, i tried this line:`can [:record_feedback, :rerun_screener], Document` (inside Ability.rb), and it worked.

PR #6319, opened 2026-05-22
Fix SmartScan feedback box bug

embarnard:
## Link to Jira issue

- e.g. https://codeforamerica.atlassian.net/browse/GYR1-1043

## Is PM acceptance required?

## What was done?

- fixed bug where when we were submitting the feedback for the SmartScan it was crashing
- it was breaking because when we moved around the box from the bottom to the top we put it inside of another form causing the params to break
- also added a feature test so that tests will break if it breaks in the future

### DIFF
--- a/app/views/hub/documents/_feedback_collection.html.erb
+++ b/app/views/hub/documents/_feedback_collection.html.erb
@@ -1,9 +1,9 @@
-<% if can? :read, :DocAssessmentFeedback && @document.latest_assessment&.feedbacks&.blank? %>
+<% if can?(:read, :DocAssessmentFeedback) && @document.latest_assessment&.feedbacks&.blank? %>
   <div class="feedback-box spacing-above-15">
     <div id="top-question">
       <span>Is the Smart Scan label correct?</span>
 
-      <%= button_to record_feedback_hub_client_document_path(client_id: @document.client_id, id: @document.id, document_id: @document.id),
+      <%= button_to record_feedback_hub_client_document_path(client_id: @document.client_id, id: @document.id),
                     params: { feedback: "correct" },
                     style: "background:none; border:none; padding:0; cursor:pointer;" do %>
         <%= image_tag('icons/checkmark-green-circle.svg', alt: "correct") %>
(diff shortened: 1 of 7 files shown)


app/views/hub/documents/_form.html.erb
```diff
@@ -17,8 +19,6 @@
       </div>
     <% end %>
 
-    <%= render 'hub/documents/ai_screener' %>
```
embarnard, line 20: main issue was this, putting this form inside of another form caused the document params to be blank

spompea-cfa: snaaaaap

app/views/hub/documents/_feedback_collection.html.erb
```diff
@@ -1,9 +1,9 @@
-<% if can? :read, :DocAssessmentFeedback && @document.latest_assessment&.feedbacks&.blank? %>
+<% if can?(:read, :DocAssessmentFeedback) && @document.latest_assessment&.feedbacks&.blank? %>
   <div class="feedback-box spacing-above-15">
     <div id="top-question">
       <span>Is the Smart Scan label correct?</span>
 
-      <%= button_to record_feedback_hub_client_document_path,
+      <%= button_to record_feedback_hub_client_document_path(client_id: @document.client_id, id: @document.id),
```
spompea-cfa: i'm wondering if these new args are needed?  based on `bin/rails routes | grep feedback`, seems like they could be redundant

app/views/hub/documents/_feedback_collection.html.erb
```diff
@@ -18,7 +18,7 @@
 
     <div data-feedback-notes class="hidden spacing-above-15 spacing-below-0">
       <p class="spacing-below-5">Provide additional feedback</p>
-      <%= form_with url: record_feedback_hub_client_document_path, method: :post do |f| %>
+      <%= form_with url: record_feedback_hub_client_document_path(client_id: @document.client_id, id: @document.id), method: :post do |f| %>
```
spompea-cfa: same q as above

spompea-cfa (approved):
👍

app/views/hub/documents/_feedback_collection.html.erb
spompea-cfa: actually i'm not sure. but, yeah, just want to understand this better :)

embarnard: @spompea-cfa i think you're right! removed them and they still work

PR #6325, opened 2026-05-29
GYR-1012 Update service selection income page

spompea-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1012

## Is PM acceptance required?

## What was done?

Updates to the eiligibity-wages page:
- updated/expanded the income bands in the dropdown
- added 3 new checkbox items to the 'scenarios that apply' section; was able to 
  use existing fields for these. there is some nuanced logic around these fields; assumption
  is that the new checkboxes should *not* affect the determination of an 'ineligible' status
- added a new question: "Do you have your income tax documents", which involved 
  adding a new field to Intake
- various copy changes
- spec test updates, specifically around 'triage', related to the above changes.

## How to test?

- Exercise the fields in this page and ensure values are properly saved.
- Exercise DIY vs GYR triage logic as it pertains to all the form choices on this page

- Before

image

- After

image

image


spec/support/helpers/feature_helpers.rb
```diff
@@ -66,7 +66,7 @@ def answer_gyr_triage_questions(screenshot_method: nil, **options)
       triage_feature_helper.maybe_screenshot do
         # Personal Info
         expect(page).to have_selector("h2", text: I18n.t('questions.eligibility_wages.edit.title'))
-        select I18n.t("questions.eligibility_wages.edit.income_level.options.#{options[:triage_income_level]}")
+        find("option[value='" + options[:triage_income_level] + "']").select_option
```
spompea-cfa, line 69: backed out the I18n slugs (for income bands) up in `app/views/questions/eligibility_wages/edit.html.erb`, so this capybara select statement needed to be changed

app/forms/eligibility_wages_form.rb
```diff
@@ -17,12 +21,20 @@ class EligibilityWagesForm < QuestionsForm
   validate :answered_vita_income_ineligible
 
   def save
+    attributes = attributes_for(:intake)
+
+    had_rental_income = attributes[:had_rental_income]
+    has_crypto_income = attributes[:has_crypto_income]
+    if had_rental_income == 'no' && has_crypto_income == '0'
+      attributes.update(triage_vita_income_ineligible: 'no')
+    end
```
spompea-cfa, line 30: there're some pre-existing, idiosyncratic semantics around the triage_vita_income_ineligible field  (labelled "None of these apply" in the UI) -- e.g., in the html form, unchecked is 'yes', while checked is 'no',  and it gets referenced in several places later on ... with the addition of three new checkboxes in this PR, this new additional logic preserves the correct business logic around this field. (edit: fixed a small typo)

app/forms/eligibility_wages_form.rb
```diff
@@ -41,7 +53,10 @@ def save
 
   def answered_vita_income_ineligible
     if had_rental_income == "no" &&
-       has_crypto_income != "true" &&
+       has_crypto_income != "1" &&
```
spompea-cfa: pretty sure this was a bug before (rails sends up booleans as '0' / '1' typically)

spompea-cfa: nevermind, i rolled back the change i'd made here. i misunderstood how the `has_crypto_income` field is working here. apparently, when checked, it comes back as `'true'`, and when unchecked, it comes back as `'0'`. (i wonder why it works like that?)

app/forms/eligibility_wages_form.rb
spompea-cfa: further clarifying: my understanding is that only when had_rental_income and/or has_crypto_income get checked does it then follow that triage_vita_income_ineligible needs to become 'yes'.

app/models/intake/gyr_intake.rb
```diff
@@ -514,6 +517,11 @@ class Intake::GyrIntake < Intake
     "unfilled" => 0,
     "zero" => 1,
     "1_to_69000" => 2,
+    "1_to_10000" => 5,
```
embarnard, line 520: ooof I hate how this is out of order now but I guess the alternative is that we would have to migrate all the old records right?

embarnard (approved):

app/models/intake/gyr_intake.rb
spompea-cfa: yeah 😶‍🌫️

PR #6347, opened 2026-06-26
Revert "Remove logic from gemfile. (#6343)"

powersurge360:
Reverts the removal of logic from the gemfile in anticipation of updating our codebase's dependencies.


spompea-cfa (approved):
👍

PR #6350, opened 2026-07-08
[closed w/o merging] GYR1-1050 Client portal improvements for statuses of "Intake, Not Ready" & "Intake, Needs Doc Help"

spompea-cfa:
> [!NOTE]
> _Closed w/o merging in favor of another PR:_ https://github.com/codeforamerica/vita-min/pull/6358

## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1050

## Is PM acceptance required?

## What was done?

- Flipper flag added: `:client_portal_improvements`; most changes in this PR 
  (where it made sense) are wrapped with this flag
- TaxReturnCardHelper logic update: slight change to how the state 
  `intake_needs_doc_help` is reflected/handled by the client portal home page: 
  per the Figma document, the client portal should now look the same (in terms 
  of copy and button etc) for both `intake_in_progress` or 
  `intake_needs_doc_help`.  (The current behavior is for `intake_needs_doc_help` 
  to invoke the complete_intake_documents button and associated copy, etc.).
  - the button itself still should take the user to the correct url, so some 
    additional logic was added to ensure this happens in the case of 
    `intake_needs_doc_help`
- simple copy changes
  - removed the "Client ID" bit at the top above the card
  - removed the "here's a snapshot" copy (also above the card)
  - added "Your" (i.e., "Your 2025 return")
  - added new body text (just above card)
  - updated button text
- UI widgets
  - removed circle/pie chart progress widget
  - added an "In Progress" yellow badge
  - added new CTA yellow box (this replaces the red text) inside card + new copy
- mixpanel
  - client portal home now sends up a 'client_portal_visited' event along with a 
    key-value pair of `{return_status: current_state}`
  - button inside card now has a `"data-track-click": 
    "client_portal_finish_intake"` added to it (for versions of that button 
    which are in scope for this ticket/PR).

## How to test?

Toggle the Flipper flag on/off as well as flip between the 'en' and 'es' 
versions and ensure that the text, UI, and layout aspects are looking good as 
you do so.

## Note to reviewer

Would appreciate a close review around ... 
- the new logic around intake_needs_doc_help inside TaxReturnCardHelper
- the flipper flag being tucked into various nooks and crannies
- and just the overall number of details that i want to ensure we get right

- Before

Before, the `intake_in_progress` (that is, 'Intake, Not Ready') and
  `intake_needs_doc_help` statuses invoked slightly different versions of the page:

Intake, Not Ready:

image

Intake, Need Doc Help:

image

- After

Now, the page is the same for both statuses:

image

Spanish version:

image

### DIFF
--- a/app/assets/stylesheets/components/_portal-status-line-items.scss
+++ b/app/assets/stylesheets/components/_portal-status-line-items.scss
@@ -27,3 +27,26 @@
     }
   }
 }
+
+.portal-status-badge {
+  background-color: #fece68;
+  border-color: #fece68;
+  border-radius: 0.5rem;
+  color: black;
+  display: inline-block;
+  font-size: 1.7rem;
+  font-weight: 500;
+  line-height: 1.5rem;
+  margin-left: 0.5rem;
+  padding: 0.75rem 1rem;
+  position: relative;
+  top: -.1em;
+}
+
+.portal-info-box {
+  background-color: #fff4d1;
+  border: 3px solid #fab323;
+  color: black;
+  margin-bottom: 2.5rem;
+  padding: 1rem 2rem;
+}
--- a/app/controllers/portal/portal_controller.rb
+++ b/app/controllers/portal/portal_controller.rb
@@ -9,6 +9,9 @@ module Portal
     def home
       @tax_returns = current_client.tax_returns.order(year: :desc).to_a
       @tax_returns << PseudoTaxReturn.new(intake: current_intake, time: app_time) if @tax_returns.empty?
+
+      current_state = current_intake.tax_returns.last.current_state
+      send_mixpanel_event(event_name: 'client_portal_visited', data: {return_status: current_state})
     end
 
     def current_intake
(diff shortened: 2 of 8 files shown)


embarnard (commented):
We talked about this in one of the sprint planning sessions but looking at the JIRA ticket I realize it was never added there. But I think since there are going to be so many of these client portal changes having so many of the `(Flipper.enabled?(:client_portal_improvements)` if/elses incorporated into the existing portal files will be confusing and harder to disentangle once we get rid of the old pages. And I think it would make more sense to just have two portal pages and we route to each one depending on the flipper flag and obviously if it supports that status currently. And then when we can just delete the old pages when its no longer relevant. What do you think?

spompea-cfa:
> have two portal pages and we route to each one depending on the flipper flag

that sounds good to me. i'll take another look

spompea-cfa:
_Closed w/o merging in favor of another PR:_ https://github.com/codeforamerica/vita-min/pull/6358

PR #6359, opened 2026-07-15
GYR-1008 Create env feature flag

powersurge360:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1008

## Is PM acceptance required?

## What was done?

This is part of a process to move to doppler.
- [x] Take an inventory of doppler secrets
- [x] Implement a thin class to swap between credentials and env var style variables (you are here)
- [ ] Move development to `fnox` and/or `doppler run`
- [ ] Add tofu to support env vars and make sure doppler has valid values for these env vars
- [ ] Turn on a percentage of credential access to use env vars
- [ ] Rotate credentials

## How to test?

This change is mostly cosmetic at this stage, until the flag is turned on. There are tests that should pass and in many cases the application won't boot unless these variables are set. Clicking around where the variables are used should indicate whether or not this was successful.

## Additional notes
Bulk of renaming was done with this script, which wasn't included in source control due to it being a one time thing:

```ruby
#!/usr/bin/env ruby -i

ARGF.each_line do |line| 
  captures = line.match(/Rails.application.credentials.dig\(([a-zA-Z,:_\s)]+)\)/)&.captures

  if captures
    env_var = captures[0].delete(":").gsub(", ", '_').upcase

    puts line.gsub(/Rails.application.credentials.dig\(.+\)/, "EnvironmentCredentials['#{env_var}']")
  else
    puts line
  end
end
```

- Before

- After

### DIFF
--- a/app/lib/environment_credentials.rb
+++ b/app/lib/environment_credentials.rb
@@ -18,7 +18,7 @@ class EnvironmentCredentials
       else
         credentials_value(name)
       end
-    rescue ActiveRecordError
+    rescue ActiveRecord::ActiveRecordError
       # Necessary when starting from scratch because flipper tables are not
       # established yet
       ENV.fetch(name, nil) || credentials_value(name)
--- a/app/lib/environment_credentials.rb
+++ b/app/lib/environment_credentials.rb
@@ -18,7 +18,7 @@ class EnvironmentCredentials
       else
         credentials_value(name)
       end
-    rescue ActiveRecord::NoDatabaseError, ActiveRecord::StatementInvalid
+    rescue ActiveRecordError
       # Necessary when starting from scratch because flipper tables are not
       # established yet
       ENV.fetch(name, nil) || credentials_value(name)
--- a/app/lib/environment_credentials.rb
+++ b/app/lib/environment_credentials.rb
@@ -12,11 +12,16 @@ class EnvironmentCredentials
 
   class << self
     def [](name)
+      
       if Flipper.enabled?(:use_env_secrets)
         ENV.fetch(name, nil)
       else
         credentials_value(name)
       end
+    rescue ActiveRecord::NoDatabaseError, ActiveRecord::StatementInvalid
+      # Necessary when starting from scratch because flipper tables are not
+      # established yet
+      ENV.fetch(name, nil) || credentials_value(name)
     end
 
     def dig(*keys)
--- a/app/lib/environment_credentials.rb
+++ b/app/lib/environment_credentials.rb
@@ -1,10 +1,5 @@
 class EnvironmentCredentials
-  # Maps environment variable names to Rails credential dig paths for keys
-  # where the service name doesn't match the first segment of the ENV var name.
-  #
-  # Most keys are resolved automatically by trying 2-level and 3-level
-  # credential paths (e.g. MAILGUN_API_KEY -> [:mailgun, :api_key]).
-  # Only truly non-obvious mappings go here.
+  # Some variables don't map obviously.
   SECRET_KEYS = {
     'VITA_MIN_EFIN' => [:irs, :efin],
     'MD_SIN' => [:irs, :md_sin],
--- a/app/lib/environment_credentials.rb
+++ b/app/lib/environment_credentials.rb
@@ -1,5 +1,29 @@
 class EnvironmentCredentials
+  # Maps environment variable names to Rails credential dig paths for keys
+  # where the service name doesn't match the first segment of the ENV var name.
+  #
+  # Most keys are resolved automatically by trying 2-level and 3-level
+  # credential paths (e.g. MAILGUN_API_KEY -> [:mailgun, :api_key]).
+  # Only truly non-obvious mappings go here.
+  SECRET_KEYS = {
+    'VITA_MIN_EFIN' => [:irs, :efin],
+    'MD_SIN' => [:irs, :md_sin],
+    'VITA_MIN_SIN' => [:irs, :sin],
+    'GYR_EFILER_CERT' => [:irs, :efile_cert_base64],
+    'GYR_EFILER_ETIN' => [:irs, :etin],
+    'GYR_EFILER_APP_SYS_ID' => [:irs, :app_sys_id],
+    'INTERCOM_ACCESS_TOKEN' => [:intercom, :intercom, :access_token]
+  }.freeze
+
   class << self
+    def [](name)
+      if Flipper.enabled?(:use_env_secrets)
+        ENV.fetch(name, nil)
+      else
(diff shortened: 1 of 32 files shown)


app/models/airtable/organization.rb
```diff
@@ -1,9 +1,9 @@
 module Airtable
   class Organization < Airrecord::Table
-    self.base_key = Rails.application.credentials.dig(:airtable, :base_key)
-    self.table_name = Rails.application.credentials.dig(:airtable, :table_name)
+    self.base_key = EnvironmentCredentials['AIRTABLE_BASE_KEY']
+    self.table_name = EnvironmentCredentials['AIRTABLE_TABLE_NAME']
```
powersurge360, line 4: It is notable that things like this get set at class eval time and therefore are not subject to `#percentage_of_time` on each call. Rather, what you get at assignment is what you will get perpetually. You could add something to the `ApplicationHelper` like

```ruby
  def lazy_config(obj, property, value)
    obj.define_singleton_method(obj, property, lambda do
      EnvironmentCredentials[value]
    end)
  end
```

And then set here with
```ruby
lazy_config(self, 'base_key', 'AIRTABLE_BASE_KEY')
```

but I think there are enough things elsewhere to signal our system isn't working. The secret swap as a percentage of time doesn't need to be comprehensive and this workaround won't work for where we set env vars via keyword arguments.

app/services/intercom_service.rb
```diff
@@ -40,23 +40,27 @@ def self.inform_client_of_handoff(client:, send_sms:, send_email:)
...
-      user_id.to_s
-    ) unless user_id.nil?
+    cred = EnvironmentCredentials['INTERCOM_SECURE_MODE_SECRET_KEY']
+
+    unless user_id.nil?
+      OpenSSL::HMAC.hexdigest(
+        'sha256',
+        cred,
+        user_id.to_s
+      )
```
powersurge360, line 50: This and changes of the same shape are automatic editor refactors.

app/services/intercom_service.rb
```diff
@@ -160,12 +164,11 @@ def self.intercom_api(collection, verb, params)
 
     begin
       result = intercom.send(collection).send(verb, params)
-    rescue
-      Intercom::AuthenticationError => e
-        retry_counts += 1
-        DatadogApi.increment("intercom.api.authentication_failure_retry")
-        retry if retry_counts <= MAX_RETRY_COUNT
-        raise e, "Failed #{MAX_RETRY_COUNT} times to authenticate with Intercom"
+    rescue Intercom::AuthenticationError => e
```
powersurge360, line 167: I think this is equivalent. Another auto editor refactor. Please double check this.

app/lib/environment_credentials.rb
```diff
@@ -1,5 +1,24 @@
...
+    'VITA_MIN_SIN' => [:irs, :sin],
+    'GYR_EFILER_CERT' => [:irs, :efile_cert_base64],
+    'GYR_EFILER_ETIN' => [:irs, :etin],
+    'GYR_EFILER_APP_SYS_ID' => [:irs, :app_sys_id],
+    'INTERCOM_ACCESS_TOKEN' => [:intercom, :intercom, :access_token]
+  }.freeze
+
   class << self
+    def [](name)
+      if Flipper.enabled?(:use_env_secrets)
```
powersurge360, line 16: The idea is that we will slowly turn this on with Flipper's `percentage_of_time` [functionality](https://www.flippercloud.io/docs/features/percentage-of-time) and observe errors to mitigate explosions.

app/lib/environment_credentials.rb
```diff
@@ -15,5 +34,36 @@ def irs(key)
...
       ENV[env_var_names[key]].presence || dig(:irs, key)
     end
+
+    private
+
+    # Imperfectly attempts to find keys within credentials, if the credential
+    # isn't found in the SECRET_KEYS hash map. Note, this means that FOO_BAR_BAZ
+    # will first match dig(:foo, :bar_baz) if it is present and will therefore
+    # not match dig(:foo, :bar, :baz)
+    def credentials_value(name)
```
powersurge360, line 49: This function is basically recursive except refactored as a loop. For when you prefer potential infinite loops to potential stack overflows. The rest of the info relevant is in the comment.

app/lib/environment_credentials.rb
```diff
@@ -1,5 +1,29 @@
 class EnvironmentCredentials
+  # Maps environment variable names to Rails credential dig paths for keys
+  # where the service name doesn't match the first segment of the ENV var name.
+  #
+  # Most keys are resolved automatically by trying 2-level and 3-level
+  # credential paths (e.g. MAILGUN_API_KEY -> [:mailgun, :api_key]).
+  # Only truly non-obvious mappings go here.
+  SECRET_KEYS = {
```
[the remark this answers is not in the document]
powersurge360: Intentionally left IRS alone to be backwards compatible since it already freely switches between envs and credentials.

app/lib/environment_credentials.rb
spompea-cfa: * disregard if not applicable, but would using a double underscore  -- eg FOO__BAR_BIZ to distinguish between FOO__BAR__BIZ -- be something we could do?
* is the 'infinite loops' comment a joke or ... ? i'm a little worried lol

app/services/intercom_service.rb
spompea-cfa: what editor/tool?

spompea-cfa (approved):
thank you for moving this work forward.

app/services/intercom_service.rb
powersurge360: It uses rubocop, with the application specific rules. You would get the same output running `bundle exec rubocop` on the file.

app/lib/environment_credentials.rb
powersurge360: I don't think the infinite loops aren't strictly possible without a bug. `FOO__BAR__BIZ` would be a solution but it is the existing secret storage informing the way we do secrets forever, so I am reluctant. I think it would be better and easier to disambiguate in the `credentials.yml` file when necessary.

powersurge360: And to be clear, the bug would be happen if someone forgets to modify the `string` variable in the loop and thereby stops consuming the string.

app/services/intercom_service.rb
spompea-cfa: 👍

app/lib/environment_credentials.rb
spompea-cfa: cool 👍

PR #6364, opened 2026-07-21
Adds fnox for secrets from doppler

powersurge360:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1008

## Is PM acceptance required?

## What was done?

- This adds `fnox.toml` configuration to the application, for use in development.

## How to test?

- Install `fnox` and [follow the instructions for shell integration](https://fnox.jdx.dev/guide/quick-start.html)
- `cd` into the directory and see that the secrets are being loaded
- Enable the `use_env_secrets` flag by going into the console. `bin/rails c`. Once inside, issue this command:

```ruby
Flipper.enable(:use_env_secrets)
```

- Before

- After

### DIFF
--- a/fnox.toml
+++ b/fnox.toml
@@ -1,0 +1,46 @@
+default_provider = "doppler"
+
+[providers.doppler]
+type = "doppler"
+project = "tax-get-your-refund"
+config = "dev"
+
+[secrets]
+ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY = { provider = "doppler", value = 'ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY' }
+ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT = { provider = "doppler", value = 'ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT' }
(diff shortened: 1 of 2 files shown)


fnox.toml
```diff
@@ -0,0 +1,46 @@
+default_provider = "doppler"
```
embarnard, line 1: default is already doppler so don't need to declare it every line right? I don't really mind though

spec/controllers/concerns/state_file/fyst_sunset_redirect_concern_spec.rb
```diff
@@ -30,7 +37,12 @@ def index
 
   context "when the feature flag is disabled" do
     before do
-      allow(Flipper).to receive(:enabled?).with(:fyst_sunset_pya_live).and_return(false)
+      allow(Flipper).to receive(:enabled?) do |arg|
```
embarnard, line 40: thanks for fixing this!

embarnard (approved):
Looks good!

fnox.toml
powersurge360: It seems like you don't need to. I was setting these with `fnox set SOME_VAR SOME_VAR` cause I was curious what would happen, and it set it in this way so I thought I'd leave it alone. I imagine if you need multiple profiles or multiple providers it would become important.

PR #6379, opened 2026-07-29
GYR1-1085-client-portal-quality-review-signature-requested

DrewProebstel:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1085

## Is PM acceptance required?

## What was done?

- Adds the signature required box for client portal

- After
image


app/javascript/lib/mixpanel_event_tracking.js
```diff
@@ -12,7 +12,7 @@ const MixpanelEventTracking = (function () {
       document.querySelectorAll("a[data-track-click]").forEach((link) => {
         link.addEventListener("click", function (e, options) {
           const pageData = document.querySelector("#mixpanelData").dataset;
-          const clickedElement = e.target;
+          const clickedElement = e.currentTarget;
```
spompea-cfa, line 15: was curious what this change does?

sp-cfa (commented):
Looks good. The additions to the TaxReturns controller/model are nice. Looking 
around, a few things I wanted to check with you on before we move fwd with 
everything:
- Copy differences: the copy I see (when i run this branch locally) seems  
  different from the Figma screenshot on the ticket? (As well as the screenshot 
  on the PR description, though i think that's just b/c it's old.) See screenshot
  below.
- "Download final tax papers" link -- the ticket's not clear on what this is 
  supposed to do, but i have a local test account that has both a 'final tax 
  document' as well as an 8879 document. When the flipper flag is off, i get two 
  distinct links (1 for each document), but when the flipper flag is on, the 
  link is only downloading a single doc (the 'final tax document' one). it's 
  possible that the way i have my test set-up locally is not quite right, but 
  wondering if can we double check what this link is supposed to do when there 
  are multiple documents?
- The change to `mixpanel_event_tracking.js` -- was curious what this change 
  is doing? (since it's code that's more low-level)

Image

app/javascript/lib/mixpanel_event_tracking.js
DrewProebstel: e.target is the deepest element actually clicked, which can be a child of the <a>
e.currentTarget is always the element the listener is attached to, i.e. the <a> itself.

`  <a href="/portal/documents/1" class="link" data-track-click="client_portal_download_tax_papers" data-track-attribute-page="client_portal">
    <span class="link__border">Download your 2024 tax papers</span><i class="icon-keyboard_arrow_right link__icon"></i>
  </a>`

so here it makes it so the entire <a> tag is sending the correct event.

spompea-cfa: ok

sp-cfa (approved):

PR #6388, opened 2026-08-03
create tax returns index page

DrewProebstel:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1058

## Is PM acceptance required?

## What was done?

- Creates the Tax Return index page

image


spompea-cfa (commented):
looks great, just a couple things i noticed that i think are worth addressing; lmk if you want to chat about these.

- i think only the latest two tax years needs to be shown, but tax_returns#index is pulling all years (ticket: "Display client's tax return documents for prior and current year")

- seems like we should remove the tax returns from the portal documents page (ticket: "Do not display tax returns on the document tabs page " -- lmk if you're interpreting that differently). (also, would need to wrap this bit of logic in the flipper flag)

Image

Image

sp-cfa (approved):
👍

PR #6389, opened 2026-08-04
GYR1-1120 (1 of 2) Migrate from Rails 7.0 defaults to 7.1 defaults

sp-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1120

(This is the first PR of two.)

## Is PM acceptance required?

## What was done?

When upgrading to Rails 7.1 (which was done some time ago) a new file is introduced, 
`config/initializers/new_framework_defaults_7_1.rb`; initially all new 7.1 
settings/defaults are commented out. Basically, uncommenting all the settings in 
this file is equivalent to simply setting `config.load_defaults` to 7.1 in 
application.rb (with 1 or 2 caveats).

The process here is to uncomment the settings one by one (or a few at a time, 
as seems reasonable), and address any spec tests issues that come up; in 
cases where we end up needing to diverge from a 7.1 default, or the 7.1 guidance 
otherwise indicates, new lines of logic have been added to application.rb.  

Lastly, removed the temporary defaults file and updated `config.load_defaults` 
(in application.rb) from 7.0 to 7.1.

## How to test?

This process relies heavily on the spec tests. For each new default, if the 
tests are all green, it was assumed OK.


config/application.rb
```diff
@@ -166,5 +166,45 @@ class Application < Rails::Application
...
+    # our `Rails.application.config.active_support.key_generator_hash_digest_class`
+    # currently has a value of `OpenSSL::Digest::SHA256` and (b) we *do*
+    # currently use field-level encryption, these two directives are therefore
+    # included here.
+    Rails.application.config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256
+    # (Guidance says set to false if *not* already using field-level encryption ...)
+    Rails.application.config.active_record.encryption.support_sha1_for_non_deterministic_encryption = true
+
+    # Needs to be included explicitly (even after setting load_defaults to 7.1).
+    config.active_support.cache_format_version = 7.1
```
embarnard: I don't think this is getting set, when I run it in the console it returns `nil`

config/application.rb
```diff
@@ -166,5 +166,45 @@ class Application < Rails::Application
...
+    # E.g., spec/controllers/questions/consent_controller_spec.rb:70
+    # (7.1 default is to set this to `true`.)
+    Rails.application.config.precompile_filter_parameters = false
+
+    # Per Rails 7.1 defaults guidance, since (a)
+    # our `Rails.application.config.active_support.key_generator_hash_digest_class`
+    # currently has a value of `OpenSSL::Digest::SHA256` and (b) we *do*
+    # currently use field-level encryption, these two directives are therefore
+    # included here.
+    Rails.application.config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256
```
embarnard: I think this already gets set when we put the default to 7.1, in anycase it returns still `OpenSSL::Digest::SHA256` when i comment this line out

embarnard (approved):
A few minor comments but looks good! 😊

config/application.rb
spompea-cfa: thanks. the guidance is a little vague at times; guess this is in fact being set once `load_defaults` is 7.1.

config/application.rb
spompea-cfa: i think you can check it via: `ActiveSupport::Cache.format_version` .... that said, this seems to be sort of the same situation as my other comment, i.e., it looks like this is, in fact, set once `load_defaults` is 7.1, so i backed it out

config/application.rb
```diff
@@ -166,5 +166,40 @@ class Application < Rails::Application
         "#{log_line.to_json}\n"
       end
     end
+
+    # ------------------------------------------------ #
+    # BEGIN additions for Rails 7.1 defaults migration #
+    # ------------------------------------------------ #
+
+    # For reference, latest version of 7.1 defaults guidance is here:
+    # https://github.com/rails/rails/blob/v7.1.6/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_7_1.rb.tt
```
spompea-cfa, line 175: the guidance is housed in the very file we want to remove once finished; for reference, i'm including a link to a copy in the Rails repo

PR #6397, opened 2026-08-11
heroku has EOL'd the 22 stack; 26 is the current

jnf:
I had a couple minutes; just enough time to delete the latest deprecation reminder email from Heroku.


sp-cfa (approved):
👍

PR #6403, opened 2026-08-14
use doppler vars for credentials

DrewProebstel:
## Is PM acceptance required?

Removes database url for credentials files and ensures that the doppler env var is used


config/database.yml
```diff
@@ -11,7 +11,7 @@ deploy_default: &deploy_default
   adapter: postgis
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + 1 %>
   timeout: 5000
-  url: <%= Rails.application.credentials[:database_url] %>
+  url: <%= ENV.fetch("DATABASE_URL", "") %>
```
powersurge360: I think merely deleting it will get you the functionality.

embarnard: @powersurge360 because the default will get pulled from aptible?

powersurge360: We found out yesterday that you _do_ need to set `DATABASE_URL` in aptible, but rails has a hardcoded check to look for that var without any configuration being required. It seems like the priority works like this:

1. Configured pieces in the yaml file (eg, `user`, `password`, etc properties)
2. The full `DATABASE_URL`, if specified
3. the `url` property in yaml.

The rails code seems to merge these three together with the further down overwriting the higher up. Because this was set to pull from credentials, it was overwriting the `DATABASE_URL` being passed in.

embarnard (changes requested):
Can we test in demo as well?

config/database.yml
embarnard: Okay so we should remove it from database.yml as to not confuse ourselves in the future on where this is getting set right? Maybe we should add a comment in here saying "don't set the database url in here it will overwrite the value in doppler" for our future selves

powersurge360: Yea, I agree. I think just the standard combo of using the broken up, individual config key value pairs like `user: <some_user>` and `DATABASE_URL` is the right way to go. That third tier I find very surprising.

embarnard (approved):

PR #6412, opened 2026-08-21
GYR1-1168 Migrate from annotate gem to annotaterb gem

sp-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1168

## Is PM acceptance required?

## What was done?

- old annotate gem: https://github.com/ctran/annotate_models
- new annotaterb https://github.com/drwl/annotaterb
- annotaterb Migration Guide:
  https://github.com/drwl/annotaterb/blob/main/MIGRATION_GUIDE.md

annotaterb is a community-forked replacement for the annotate gem. annotate 
isn't maintained anymore, and the GitHub issues on the annotate repo imply it’s 
not fully compatible with Ruby 8. 

There's a migration guide that was followed; see the link above.

Using version 4.23 for now, instead of 4.24, due to an enum regression bug in 4.24.
More info: https://github.com/drwl/annotaterb/issues/373

A new rake task was added; it gets created when running: ` bin/rails g 
annotate_rb:install`

### Bringing custom settings over

Compared the settings in the legacy ` auto_annotate_models.rake` with the new 
`.annotaterb.yml`, the aim being to preserve differing settings that needed to 
be carried over into the new yml file. The two files were mostly the same, but a 
few differences listed here:

- Skipped/ignored:
  - `format_bare: true` -- this was in the old defaults but not in the 
    new .annotaterb.yml; however, this is unused now in annotaterb, see:
    https://github.com/drwl/annotaterb/blob/a4a52ca836cd374f82d5b3f529977761af6f4da4/lib/annotate_rb/options.rb#L39

- Changed (.annotaterb.yml had both these set to empty):
  - `:hide_default_column_types: 'json,jsonb,hstore'`
  - `:hide_limit_column_types: 'integer,bigint,boolean'`

- Added: 

        :ignore_model_sub_dir:
        - archived

## Testing

Ensure checks are green -- particularly the annotate check.


DrewProebstel (approved):
Bundles just fine on my end. Looks good to me!

PR #6416, opened 2026-08-25
Take out websocket-driver from Gemfile

sp-cfa:
websocket-driver doesn't need to be pulled into the Gemfile anymore (though it is a transitive dependency of actioncable).


Noaj (approved):
+1 
Confirmed I will be pulled in transitively by Action Cable

PR #6414, opened 2026-08-24
Show new version of portal states design if Flipper flag is enabled

tmccoy-cfa:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1157?atlOrigin=eyJpIjoiYzQ0YzFkMWFlYmNlNDc0OWE0ZmQ1YTk4NWE2MDJmOGMiLCJwIjoiaiJ9

## Is PM acceptance required?

- Yes - don't merge until Jira issue is accepted!

## What was done?

- Update to show new design for portal states tax return status cards when Flipper flag is enabled

## How to test?

- Manual Testing
  - Log in to the Hub with user with permissions to view Portal States page and update Flipper flags
  - If 'client_portal_improvements' flag is not enabled, enable it
  - View Portal States page to see updated designs for tax return status cards
  - Disable 'client_portal_improvements' flag to view previous design and confirm Flipper flag functionality

- Risk Assessment
  - N/A

## Screenshots (visual changes)

- Before
- 
image

- After
image

PR #6419, opened 2026-08-26
GYR1-1156 Backfill some new spec tests for client portal improvements

sp-cfa:
## Jira ticket

- https://codeforamerica.atlassian.net/browse/GYR1-1156

## Is PM acceptance required?

- No

## What was done?

- Backfilled some new spec tests for the new client portal work.
- Removed some legacy CTC spec tests 
- Added a `pseudo?` method -- which always returns false -- to the Intake model 
  class; the purpose of this is to sync up with a recent change made to 
  PortalStatesController::PseudoTaxReturn (under PR #6414) which these new spec 
  tests leverage.


Noaj (approved):
Looks good!
🚢 it!

PR #6420, opened 2026-08-27
Gyr1 479 gyr clients cannot upload images on portal upload documents page

Noaj:
## Link to Jira issue

  - https://codeforamerica.atlassian.net/browse/GYR1-479

## Is PM acceptance required?

## What was done?

  - Fixed the issue related to upload documents in portal/still-need-help/upload-documents. The issue was the hidden document_type field that was only rendered when document_type is set, but document_type is only set when a the param ?document_type is present in the URL. When visiting /en/portal/upload-documents directly (no param), document_type is nil, no hidden field is rendered, and the form submits without a document_type that causes the “Can't be blank.” error message.
   
  - Another issue that I found: When an error happens, it will render the Edit view but with an empty list, but it should display the already uploaded documents(if document exist). This is fixed as well.

## How to test?

  - Manual tests
    - Log in as GYR Client
    - Visit [portal/still-need-help/upload-documents](http://localhost:3000/en//portal/still-need-help/upload-documents)
    - Click “Upload Additional Documents”
    - Click “Select a file”

- Before
Before_updates2

Before_updates

- After
before_upload

after_upload

### DIFF
--- a/app/controllers/portal/upload_documents_controller.rb
+++ b/app/controllers/portal/upload_documents_controller.rb
@@ -40,6 +40,12 @@ module Portal
         redirect_to portal_upload_documents_path(document_type: form_params[:document_type])
       else
         flash.now[:error] = I18n.t("portal.upload_documents.error")
+        if form_params[:document_type].present?
+          @documents = current_client.documents.active.where(document_type: form_params[:document_type])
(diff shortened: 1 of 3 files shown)


sp-cfa (approved):
👍

PR #6424, opened 2026-08-31
Gyr1 1175 add missing indexes for columns to filter sort

Noaj:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-1175

## Is PM acceptance required?

## What was done?

- We had some fields use for sorting and filter that were not indexed. Fields affected: locale, preferred_interview_language, state_of_residence, with_general_navigator, with_incarcerated_navigator, with_limited_english_navigator, with_unhoused_navigator.

## How to test?

  - Go to the client search(in /en/hub/clients) and use the filter and sorts.
  - Another way to test is doing some EXPLAIN in the DB, something like: EXPLAIN ANALYZE SELECT * FROM intakes
  WHERE type = 'Intake::GyrIntake' AND state_of_residence = 'CA';

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -388,17 +388,24 @@
 #  index_intakes_on_email_address                          (email_address)
 #  index_intakes_on_email_domain                           (email_domain)
 #  index_intakes_on_hashed_primary_ssn                     (hashed_primary_ssn)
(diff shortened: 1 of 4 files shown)


sp-cfa (approved):
👍

PR #6435, opened 2026-09-10
Gyr1 733 turn active hub view to default opt out

Noaj:
## Link to Jira issue

- https://codeforamerica.atlassian.net/browse/GYR1-733

## Is PM acceptance required?

## What was done?

- We are allowing users to unselect “display active returns only” filter and display all returns, not just active ons. By turning it into an opt out, rather than opt-in features; more users will discover, use, and benefit from this feature.

## How to test?

   - Go to the hub(/hub/sign_in) and log in as any role who has clients. 
   - Click on My Clients or All Clients
   - You should see the filter "Display active returns only" selected by default and users has the ability to unselect it and display all returns.

- Before
before_optout

- After

after_optout


sp-cfa (approved):
👍

PR #6443, opened 2026-09-11
Add intercom token call

embarnard:
## Is PM acceptance required?

## What was done?

`EnvironmentCredentials::SECRET_KEYS['INTERCOM_ACCESS_TOKEN']` was mapped to `[:intercom, :intercom, :access_token]`, but the actual creds structure (staging, production, demo and dev) stores it flat as intercom.intercom_access_token. This mismatch means that the creds file lookup returned nil silently.

This went unnoticed because `use_env_secrets` (added in #6359) is enabled in staging and production, so `EnvironmentCredentials` normally reads `ENV['INTERCOM_ACCESS_TOKEN'`] instead of the credentials file bypassing the bug. But Doppler secret backing that env var is itself misnamed INTERCOM_INTERCOM_ACCESS_TOKEN, so ENV.fetch('INTERCOM_ACCESS_TOKEN', nil) also returns nil.


martha-brandt (approved):
🙏

